PR #4951, opened 2026-07-16
feat: Implement TimeType support - Infrastructure - sort, min/max

### LINKED ISSUE #4379: Implement TimeType support: Infrastructure - sort, min/max

This is #4288 sub-issue 

These are the core infrastructure pieces that make TimeType columns usable in
real-world queries. Without them, any query that shuffles, sorts, or aggregates
a time column falls back to Spark.

Since TimeType is physically a fixed-width i64 (same as LongType/TimestampType),
all of these should be straightforward additions to existing type lists.

**Sort**
spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala:833-843

supportedScalarSortElementType does not include TimeType. Add alongside
DateType | TimestampType | TimestampNTZType. The Rust sort implementation
already handles i64 types, so no native changes should be needed.

**Min/Max aggregates**
spark/src/main/scala/org/apache/comet/serde/aggregates.scala:742-751

minMaxDataTypeSupported has DateType | TimestampType but not TimeType.
Add it. The Rust min/max implementation works on any orderable Arrow type,
so Time64(Nanosecond) should work without native changes.

### Additional context

_No response_

YutaLin:
Which issue does this PR close?
Closes #4379 

Rationale for this change
Support sort, min/max for TimeType, include native and JVM columnar

What changes are included in this PR?
support sort and add sql tests.

How are these changes tested?
run sql test in spark4.1, spark4.2


YutaLin:
Hi @parthchandra, can you help me review this? Thanks!

spark/src/test/resources/sql-tests/expressions/datetime/make_time_sort.sql
```diff
@@ -0,0 +1,47 @@
...
+CREATE TABLE test_make_time_sort(id int, hours int, minutes int, secs decimal(16,6)) USING parquet
+
+statement
+INSERT INTO test_make_time_sort VALUES
+    (1, 12, 30, 45.123456),
+    (2, 0, 0, 0.000000),
+    (3, 23, 59, 59.999999),
+    (4, 1, 2, 3.500000),
+    (5, 0, 0, 0.000001),
+    (6, NULL, NULL, NULL)
```
parthchandra: Maybe also test for some boundary values `make_time(0,0,0.000000)` and `make_time(23,59,59.999999)`

spark/src/test/resources/sql-tests/expressions/datetime/make_time_sort.sql
```diff
@@ -0,0 +1,47 @@
...
+)
+SORT BY t ASC NULLS FIRST
+
+query
+SELECT id, t
+FROM (
+    SELECT id, make_time(hours, minutes, secs) AS t
+    FROM test_make_time_sort
+)
+SORT BY t DESC NULLS LAST
```
parthchandra, line 48: what about also checking for `ASC NULLS LAST` and `DESC NULLS FIRST`

parthchandra: Also maybe add a case with multi column sort - `ORDER BY time_col, id`

parthchandra (commented):
Do we need to handle Literal `Time64(Nanosecond)` in the planner (`create_expr`)? 
Maybe try a query with `ORDER BY time_col LIMIT 10`

spark/src/test/resources/sql-tests/expressions/datetime/make_time_sort.sql
YutaLin: It's already in test cases.

YutaLin:
> Do we need to handle Literal `Time64(Nanosecond)` in the planner (`create_expr`)?
> Maybe try a query with `ORDER BY time_col LIMIT 10`

Hi @parthchandra, thanks for review! CometLiteral already serializes a non-null TimeType literal as LongVal, but create_expr only handled LongVal for Int64 and timestamps. I added the missing conversion to ScalarValue::Time64Nanosecond(Some(value)) and a focused Rust unit test. 
I also added more sql tests, please help me review again, appreicate!

spark/src/test/resources/sql-tests/expressions/datetime/make_time_sort.sql
```diff
@@ -0,0 +1,81 @@
...
+
+query
+SELECT id, t
+FROM (
+    SELECT id, make_time(hours, minutes, secs) AS t
+    FROM test_make_time_sort
+)
+SORT BY t ASC NULLS LAST, id DESC
+
+query
```
parthchandra, line 74: Optional: a test like - 
```
query
  SELECT id, t
  FROM (
      SELECT id, make_time(hours, minutes, secs) AS t
      FROM test_make_time_sort
  )
  ORDER BY t ASC NULLS FIRST
  LIMIT 3
```
would be an additional check for `supportedScalarSortElementType` but reached via a different path.

parthchandra (commented):
@YutaLin could you rebase? I have one optional comment, otherwise this is looking good.

PR #4950, opened 2026-07-16
build: enable Spark SQL tests for released Spark 4.2.0

andygrove:
## Which issue does this PR close?

Part of #4142.

Supersedes #4208, which was opened against `4.2.0-preview4` and went stale.

## Rationale for this change

Spark 4.2.0 is now available in Maven Central. This is stage 3 of the bring-up described in `adding_a_new_spark_version.md`: turn on Spark's own SQL tests for 4.2. The profile bump to the released 4.2.0, the `CometInternalRowShim` split it forced, and the regenerated plan stability goldens went in separately as #4960, so this PR is now only the Spark SQL test diff, the CI wiring, and the Comet bugs those test runs exposed.

Spark 4.2.0 is not simply `4.2.0-preview4` with the qualifier dropped. It restructures the test harness and adds `FILTER (WHERE ...)` support for window aggregates, so several real reconciliations were needed.

## What changes are included in this PR?

**Comet fixes exposed by the 4.2 test runs**

- `CometNativeScan` rejects a scan whose data or partition schema contains a type with no proto representation (GEOMETRY / GEOGRAPHY on 4.2), which otherwise threw `NoSuchElementException` during planning instead of falling back. The native scan serializes the full data schema, not just the projected columns, so the check cannot be limited to the required schema. `CometScanRule` now runs the projected/partition schema type check *before* this one so that a column whose type is unsupported is still reported as such (`Unsupported s of type VariantType`) rather than as the coarser serialization failure. New fixture `expressions/misc/geospatial_types.sql` covers the GEOMETRY / GEOGRAPHY fallback, including the case where the geospatial column is not projected.
- Spark 4.2 allows `FILTER (WHERE ...)` on a window aggregate. Comet dropped the filter silently -- `aggExprToProto` only serialized it for `Partial` mode aggregates and window aggregates are `Complete` mode -- and DataFusion window expressions have no filter support anyway, so `window.sql` got **wrong results**. `CometWindowExec` now declines window expressions whose aggregate carries a filter, and `aggExprToProto` falls back instead of silently dropping a filter for any non-`Partial` mode. New fixture `expressions/window/window_filter.sql` fails (wrong results) without the fix.

**Spark SQL test diff**

- Adds `dev/diffs/4.2.0.diff`, seeded from the current `4.1.3.diff` against the `v4.2.0` tag. (#4960 already deleted the stale `4.2.0-preview4.diff`.)
[...]

Issue #4142, opened 2026-04-29
Track Spark 4.2 test failures


andygrove:
### Describe the bug

```
CometExec3_4PlusSuite:
*** RUN ABORTED *** (4 minutes, 50 seconds)
  java.lang.AssertionError: assertion failed: Function identifier must be fully qualified (3-part): might_contain
  at scala.Predef$.assert(Predef.scala:279)
```

```
- unary negative integer overflow test *** FAILED *** (388 milliseconds)
  "[ARITHMETIC_OVERFLOW] overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22003" did not contain "integer overflow" (CometExpressionSuite.scala:2136)
```

```
- ANSI support for add *** FAILED *** (184 milliseconds)
  "[ARITHMETIC_OVERFLOW] overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22003
  == SQL (line 2, position 1) ==
  SELECT
  ^^^^^^
    _1 + _2
  ^^^^^^^^^
    from tbl
  ^^^^^^^^^^
  " did not contain "[ARITHMETIC_OVERFLOW] integer overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error" (CometExpressionSuite.scala:2731)
- ANSI support for subtract *** FAILED *** (163 milliseconds)
  "[ARITHMETIC_OVERFLOW] overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22003
  == SQL (line 2, position 1) ==
  SELECT
  ^^^^^^
    _1 - _2
  ^^^^^^^^^
    from tbl
  ^^^^^^^^^^
  " did not contain "[ARITHMETIC_OVERFLOW] integer overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error" (CometExpressionSuite.scala:2750)
- ANSI support for multiply *** FAILED *** (152 milliseconds)
  "[ARITHMETIC_OVERFLOW] overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22003
  == SQL (line 2, position 1) ==
  SELECT
  ^^^^^^
    _1 * _2
  ^^^^^^^^^
    from tbl
  ^^^^^^^^^^
[...]

PR #4960, opened 2026-07-17
chore: bump spark-4.2 profile to the released 4.2.0

andygrove:

Part of #4142.

## Rationale for this change

Spark 4.2.0 is now available in Maven Central, so the `spark-4.2` profile no longer has to target a preview. This PR moves the profile onto the released version and makes the main sources compile against it, without yet wiring up Spark's own SQL tests for 4.2. It is a smaller slice of the larger bring-up in #4950, intended to land the version bump on its own.

Spark 4.2.0 is not simply `4.2.0-preview4` with the qualifier dropped: it changes `SpecializedGetters` and repackages a dependency, so a couple of real reconciliations are needed to compile.

**Build**

- `spark.version` moves from `4.2.0-preview4` to `4.2.0`, and the WIP compile-only comment on the `spark-4.2` profile is dropped.
- Spark 4.2.0 bundles its own copy of `org.apache.datasketches.memory.internal.ResourceImpl` inside `spark-catalyst`, colliding with the transitive `datasketches-memory` jar and failing the `BanDuplicateClasses` enforcer rule. That class is now ignored.
- The stale Jetty comment that claimed the test-scope pin matched preview4's Jetty version is corrected. The pin itself is unchanged.

**Shims**

- Spark 4.2 removes `GeographyVal` / `GeometryVal` and replaces `getGeography` / `getGeometry` on `SpecializedGetters` with a single `getBinaryView` returning `BinaryView`. `CometInternalRowShim` can no longer be shared from the `spark-4.1+` source root, because the major, minor-plus and minor shim directories are all compile source roots (a `spark-4.2` copy would be a duplicate class, not an override). It is split into per-version `spark-4.1` and `spark-4.2` copies.

**Plan stability**

- Regenerated with `dev/regenerate-golden-files.sh --spark-version 4.2` against the released 4.2.0. The `q2`, `q5`, `q54` (v1_4) and `q5a` (v2_7) goldens are pruned because their plans now match the shared fallback on 4.2.0, and `q77a` (v2_7) is added to record a degraded plan.

This PR intentionally does not enable the Spark SQL tests or add CI jobs for 4.2; that work is tracked separately in #4950.

`./mvnw -Pspark-4.2 -DskipTests test-compile` builds the main and test sources against the released Spark 4.2.0. `./mvnw -Pspark-4.1 -DskipTests test-compile` still passes, confirming the shim split did not regress the 4.1 profile. The plan-stability suites (`CometTPCDSV1_4_PlanStabilitySuite` and `CometTPCDSV2_7_PlanStabilitySuite`) pass on the regenerated goldens.


parthchandra (approved):
lgtm. Thanks @andygrove

mbutrovich:
Great to see!

PR #4962, opened 2026-07-17
chore: codeql dependabot fix

comphead:

Closes #.

## Rationale for this change

CodeQL is split into diff PRs which break the CI, example
#4954 
#4953 

The PR combines codeql changes into single group to be in the same PR

Similar change in DF https://github.com/apache/datafusion/pull/23561


mbutrovich (approved):
Thanks @comphead!

Issue #4942, opened 2026-07-15
Ranked backlog: native scalar expressions with per-row optimization opportunities


andygrove:
Follow-up to #4936. A scan of the native scalar expression implementations in `datafusion-comet-spark-expr` (`native/spark-expr/`) for per-row antipatterns: per-element builder loops that could be vectorized Arrow kernels, `iter().map().collect()` over Option/Result, per-row heap allocation (`String`/`format!`/`to_string`), per-row `value(i)` allocation, per-element `MutableArrayData::extend`, and constants recomputed inside the per-row loop.

Each is a small self-contained change. Methodology (benchmark-first, bit-identical, no-regression gate): see the [Optimizing Scalar Expressions](https://github.com/apache/datafusion-comet/blob/main/docs/source/contributor-guide/optimizing_expressions.md) guide (PR #4933). Ranked by potential impact = per-row cost severity x usage frequency.

## High impact (decimal aggregation hot path, common in TPC-DS)

- [ ] **`spark_unscaled_value`** (`math_funcs/internal/unscaled_value.rs:37`) - per-element `Int64Builder` loop casting Decimal128 -> Int64, rebuilding an identical null buffer. Fix: `arr.unary::<_, Int64Type>(|v| v as i64)`. Injected by Spark's `DecimalAggregates` rule for `sum`/`avg` over narrow decimals.
- [ ] **`spark_make_decimal`** (`math_funcs/internal/make_decimal.rs:45`) - per-element `Decimal128Builder` loop, Int64 -> Decimal128. Fix: `unary_opt` (`long_to_decimal` already returns `Option`). Paired with `UnscaledValue` on the output side of the same rewrite.

## Medium impact

- [ ] **`list_extract`** (`array_funcs/list_extract.rs:285`) - per-row `MutableArrayData::extend(0, start+i, start+i+1)` (one element/row) plus a `Box<dyn Fn>` index adjuster called per row. Fix: build a gather-index `UInt32Array` and issue one `arrow::compute::take`; monomorphize the index adjustment. Backs `GetArrayItem` (`arr[i]`) and `element_at`.
- [ ] **`cast_int_to_timestamp`** (`conversion_funcs/numeric.rs:1006`) - per-element `TimestampMicrosecondBuilder` loop over an infallible `(v as i64) * MICROS_PER_SECOND`. Fix: `unary` + `with_timezone_opt`.
[...]

PR #4972, opened 2026-07-18
perf: vectorize `spark_unscaled_value` (9x faster)

u70b3:

Part of #4942.

## Rationale for this change

`spark_unscaled_value` is injected by Spark's `DecimalAggregates` rule for every `sum`/`avg` over narrow decimals, so it sits on the TPC-DS decimal aggregation hot path. The array branch used a per-element `Int64Builder` loop.

- Replace the per-element builder loop with the vectorized Arrow `unary` kernel: the Decimal128 values buffer is mapped to Int64 in a single pass and the null buffer is carried over untouched (zero-copy). The scalar branch is unchanged.
- Add a criterion benchmark (`benches/unscaled_value.rs`) covering no-null / sparse-null (1 in 10) / dense-null (1 in 2) shapes at 8192 rows, following the scalar expression optimization guide.

Benchmark results (8192 rows, Decimal128(20,2), median):

| shape | before | after | speedup |
|---|---|---|---|
| no nulls | 29.5 µs | 3.2 µs | ~9.3x |
| sparse nulls (1/10) | 35.5 µs | 3.0 µs | ~11.7x |
| dense nulls (1/2) | 30.5 µs | 2.9 µs | ~10.4x |

- Added unit tests for the array path (values and null distribution are bit-exact, including nulls) and the scalar path; the function previously had no direct tests.
- Full `datafusion-comet-spark-expr` suite passes (497 tests), plus `clippy --all-targets -D warnings` and `fmt`.
- Benchmark shows a strict improvement on all three shapes with no regression (criterion change −89% to −92%, p=0.00).

### DIFF
--- a/native/spark-expr/Cargo.toml
+++ b/native/spark-expr/Cargo.toml
@@ -167,3 +167,7 @@ harness = false
 [[bench]]
 name = "check_overflow"
 harness = false
+
+[[bench]]
+name = "unscaled_value"
+harness = false
--- a/native/spark-expr/benches/unscaled_value.rs
+++ b/native/spark-expr/benches/unscaled_value.rs
@@ -1,0 +1,62 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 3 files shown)


andygrove (approved):
LGTM. Thanks @u70b3

Issue #4978, opened 2026-07-20
Config naming audit: rename inconsistent spark.comet.* keys before 1.0


andygrove:
Tracker for [#4082](https://github.com/apache/datafusion-comet/issues/4082) "Review all configuration options and consider renaming some for consistency."

An audit of the 54 keys in `spark/src/main/scala/org/apache/comet/CometConf.scala` surfaced six categories of inconsistency. This issue is the umbrella; individual rename PRs should each cover one cluster and use the `.withAlternative(...)` mechanism to keep old keys working as deprecated aliases (see the [Configuration Conventions](https://datafusion.apache.org/comet/contributor-guide/config_conventions.html) contributor-guide page).

## Category 1 — Boolean flags missing the `.enabled` suffix

Current convention: boolean flags end in `.enabled`. Exceptions to rename:

- [ ] `spark.comet.nativeLoadRequired` → `spark.comet.nativeLoadRequired.enabled`
- [ ] `spark.comet.exceptionOnDatetimeRebase` → `spark.comet.datetimeRebase.exceptionOnRead.enabled` (or similar; the current name is also very long)
- [x] `spark.comet.exec.replaceSortMergeJoin` → `spark.comet.exec.forceShuffledHashJoin` (done as the pilot rename)

## Category 2 — Word separator inside a segment

Most segments use `camelCase`; a handful use dots or dashes to separate words. Fix:

- [ ] `spark.comet.columnar.shuffle.async.max.thread.num` → `spark.comet.columnar.shuffle.async.maxThreadNum`
- [ ] `spark.comet.columnar.shuffle.async.thread.num` → `spark.comet.columnar.shuffle.async.threadNum`
- [ ] `spark.comet.parquet.read.parallel.io.thread-pool.size` → `spark.comet.parquet.read.parallel.io.threadPoolSize`
- [ ] `spark.comet.exec.pyarrowUdf.enabled` vs `spark.comet.exec.scalaUDF.codegen.enabled` — pick one casing for UDF and apply consistently.

## Category 3 — `.explain.*` group has orphaned siblings

`spark.comet.explain.*` already exists (`format`, `native.enabled`, `rules`). These related keys live at the top level:

- [ ] `spark.comet.explainCodegen.enabled` → `spark.comet.explain.codegen.enabled`
- [ ] `spark.comet.explainFallback.enabled` → `spark.comet.explain.fallback.enabled`
[...]

Issue #4082, opened 2026-04-25
Comet 1.0.0 Release


andygrove:
### What is the problem the feature request solves?

I would like to start planning ahead for a 1.0.0 release. Let's use this issue to discuss what that means.

Suggestions for criteria for 1.0.0

- Demonstrated cost savings for TPC-H and TPC-DS at SF1000 (1TB)
- Consistent logging
- Thorough documentation for compatibility status
- Review all configuration options and consider renaming some for consistency
- Documented policy for how long we will support Spark versions
- Documented process for preventing major performance regressions
- Documented process for versioning (semver) moving forward. What does that mean for Comet?

parthchandra:
I'd like to see iceberg metadata columns support make it https://github.com/apache/datafusion-comet/issues/4082 (depends on a bunch of PRs in iceberg-rust though).

mbutrovich:
I will get Iceberg 1.11 support into shape https://github.com/apache/datafusion-comet/pull/4840.

andygrove:
We plan on cutting the 1.0.0 release candidate this Friday, 7/31

andygrove:
I would like to backport https://github.com/apache/datafusion-comet/pull/5079 to increase TPC-DS coverage

andygrove:
I would also like to back port https://github.com/apache/datafusion-comet/pull/5110

andygrove:
another one https://github.com/apache/datafusion-comet/pull/5132

andygrove:
We need a few more days for the release. I'm going to stop back porting PRs to branch-1.0 and will do hard reset on main branch next week and then create fresh PRs to bump version, add changeling, etc.

Goal is to have the release candidate available by Friday 8/7.

PR #4979, opened 2026-07-20
feat: add withAlternative alias mechanism for CometConf

andygrove:
## Summary

Kicks off the "Review all configuration options and consider renaming some for consistency" item on the 1.0.0 tracker (#4082). This PR lands the machinery + first rename, and files #4978 to track the full rename backlog.

- **Alias mechanism.** Adds `.withAlternative(oldKey, ...)` to `ConfigBuilder`. Reads check the primary key first, then any registered alternatives in order. Reading a value from an alternative logs a one-time deprecation warning per JVM per alternative key, naming the current key.
- **Threaded through** `TypedConfigBuilder` → `ConfigEntryWithDefault` and `OptionalConfigEntry`, so configs with defaults, env-var overrides, and optional configs all support aliases.
- **Pilot rename.** `spark.comet.exec.replaceSortMergeJoin` → `spark.comet.exec.forceShuffleHashJoin.enabled` (Scala symbol `COMET_REPLACE_SMJ` → `COMET_FORCE_SHJ`). Old key still works as a deprecated alias.
- **Contributor-guide doc.** New "Configuration Conventions" page under `docs/source/contributor-guide/` documenting key/symbol naming rules and the rename workflow.
- **Tests.** New `CometConfSuite` covers primary-wins, alternative fallback, default, multiple alternatives, `OptionalConfigEntry`, type conversion through the alias, and the pilot rename end-to-end.

Related:
- Broader rename backlog: #4978
- 1.0.0 tracker: #4082

## Test plan

- [ ] `./mvnw -pl spark test -Dtest=none -Dsuites="org.apache.comet.CometConfSuite"` — 7 tests pass locally
- [ ] `./mvnw -pl spark -am compile -DskipTests` — clean
- [ ] `npx prettier "**/*.md" --write` — clean

### DIFF
--- a/spark/src/main/scala/org/apache/comet/CometConf.scala
+++ b/spark/src/main/scala/org/apache/comet/CometConf.scala
@@ -372,8 +372,9 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(false)
 
-  val COMET_REPLACE_SMJ: ConfigEntry[Boolean] =
-    conf(s"$COMET_EXEC_CONFIG_PREFIX.replaceSortMergeJoin")
+  val COMET_FORCE_SHJ: ConfigEntry[Boolean] =
+    conf(s"$COMET_EXEC_CONFIG_PREFIX.forceShuffleHashJoin.enabled")
+      .withAlternative(s"$COMET_EXEC_CONFIG_PREFIX.replaceSortMergeJoin")
       .category(CATEGORY_EXEC)
       .doc("Experimental feature to force Spark to replace SortMergeJoin with ShuffledHashJoin " +
         s"for improved performance. This feature is not stable yet. $TUNING_GUIDE.")
@@ -1072,7 +1073,8 @@ private class TypedConfigBuilder[T](
       parent._doc,
       parent._category,
       parent._public,
-      parent._version)
+      parent._version,
+      parent._alternatives)
     CometConf.register(conf)
     conf
   }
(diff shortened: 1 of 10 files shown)


spark/src/main/scala/org/apache/comet/CometConf.scala
```diff
@@ -372,8 +372,9 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(false)
 
-  val COMET_REPLACE_SMJ: ConfigEntry[Boolean] =
-    conf(s"$COMET_EXEC_CONFIG_PREFIX.replaceSortMergeJoin")
+  val COMET_FORCE_SHJ: ConfigEntry[Boolean] =
+    conf(s"$COMET_EXEC_CONFIG_PREFIX.forceShuffledHashJoin")
+      .withAlternative(s"$COMET_EXEC_CONFIG_PREFIX.replaceSortMergeJoin")
```
andygrove, line 377: This is an example of renaming a config and preserving the old deprecated name

kazuyukitanimura: Should we mention that replaceSortMergeJoin is deprecated somewhere?

kazuyukitanimura (approved):
pending ci

andygrove:
@kazuyukitanimura since CI is already running, I'll follow up with a docs PR for the one renamed config

Issue #4842, opened 2026-07-07
(Informational) AI driven performance analysis of scan implementation across plain Spark, Velox, and Comet


schenksj:
> Built with great assistance from Anthropic Fable

# Comet Scan Performance: Root-Cause Analysis vs Plain Spark and Gluten

---

## Executive summary

**The report "Comet scans are slower than plain Spark" pertains to both plain Parquet and Iceberg, but the causes are different — and a large share of the plain-Parquet reports trace to bugs that were diagnosed and fixed across Comet 0.15.0–0.17.0 (Apr–Jun 2026), with two scan-IO fixes merged to main but still unreleased as of 2026-07 (0.17.0 tagged 2026-06-16).** The canonical public case is the AWS EKS 3TB TPC-DS benchmark: Comet 0.14.0 was **11% slower** than vanilla Spark, and after the fixes 0.16.0 was **32% faster** (37% on Iceberg). So the first diagnostic question for any new report is *which Comet version and which storage backend*.

**The report "much slower than Gluten" is structural and still true**: Comet's own published comparison (0.16.0 vs Gluten 1.6.0) shows Gluten ahead on TPC-H (2.8× vs 2.4× speedup), and the scan-side mechanisms responsible are identifiable in the Velox source and absent (or default-off) in Comet.

Current-code causes, one line each (detail + evidence below):

| # | Cause | Format | vs Spark | vs Gluten |
|---|-------|--------|----------|-----------|
| 1 | No IO/decode overlap by default: data cache + async prefetch exist but default **off**; demand-driven object-store reads | both | ✗ (loses on cold S3) | ✗✗ |
| 2 | Row-level filter pushdown default **off**, and known to *regress* when enabled (#3457) → no late materialization | both | ✗ (Spark has page skip + lazy dict) | ✗✗ (Velox: full late materialization) |
| 3 | Iceberg `dataFileConcurrencyLimit` default **1** — files read serially within a task | iceberg | ✗ | ✗ |
| 4 | Iceberg path is a separate IO stack (iceberg-rust/opendal): bypasses cache/prefetch/s3a translation; per-file overhead (~30% CPU in worst cases, iceberg-rust #2172) | iceberg | ✗ | ✗ |
| 5 | Per-batch schema-adaptation casts (INT96/ts-ms/unsigned/decimal/UUID; Iceberg field-ID re-projection) | both | ✗ (when triggered) | ✗ |
[...]

PR #4981, opened 2026-07-20
chore: remove dead Parquet parallel-IO configs

andygrove:

Partially addresses #4842 (Item #9 in the remediation section: "Delete or implement the dead IO configs (`parallel.io.*`, `mergeRanges`) so benchmarking isn't misled").

## Rationale for this change

The following configs are defined in `CometConf.scala` but referenced nowhere in current native or JVM code. They were consumed by the JVM-side Parquet reader that has since been removed:

- `spark.comet.parquet.read.parallel.io.enabled` (default `true`)
- `spark.comet.parquet.read.parallel.io.thread-pool.size` (default `16`)
- `spark.comet.parquet.read.io.mergeRanges` (default `true`)
- `spark.comet.parquet.read.io.mergeRanges.delta` (default `8 MB`)
- `spark.comet.parquet.read.io.adjust.readRange.skew` (default `false`)

These are a trap: they appear tunable but change nothing. Anyone benchmarking "Comet with IO tuning" is tuning a no-op.

- Delete the five dead config entries from `CometConf.scala`.
- Delete the stale `ParquetReadSuite` "test merge scan range" test, which only asserted that the (dead) config propagated to the Hadoop conf.

The auto-generated `configs.md` will be regenerated on the next docs build and will no longer list these entries.

Compile-checked against `spark-3.5`. No behavior change — these configs were no-ops in current code.


mbutrovich:
Can you nab "spark.comet.use.lazyMaterialization" as well?

mbutrovich:
Actually, CI is green, we can get it in a followup.

mbutrovich (approved):
Thanks @andygrove!

Issue #3756, opened 2026-03-21
[EPIC] Iceberg Feature Matrix: Spark-Iceberg vs iceberg-rust vs datafusion-comet


Shekharrajak:
We should use below matrix to check for any missing implementations that could accelerate Spark Iceberg pipeline using comet

### READ PATH

| Feature | Iceberg Java | iceberg-rust | datafusion-comet (via iceberg-rust) |
|---|---|---|---|
| Basic Parquet scan | Yes | Yes | Yes - IcebergScanExec |
| Positional deletes (V2 MoR) | Yes | Yes - ArrowReader + DeleteVector (RoaringTreemap) | Yes - delegates to ArrowReader with row_selection_enabled(true) |
| Equality deletes (V2 MoR) | Yes | Yes - ArrowReader builds equality delete predicates | Yes - delegates to ArrowReader |
| Deletion vectors (V3) | Yes - DVUtil, DVFileWriter, DVIterator | Yes - DeleteVector + Puffin deletion-vector-v1 blob support | Not wired - Comet doesn't pass DV metadata via protobuf |
| Schema evolution | Yes | Yes | Yes - IcebergStreamWrapper adapts batches to target schema |
| Partition pruning (static) | Yes | Yes | Yes - partitions serialized in protobuf |
| Dynamic partition pruning | Yes (Spark) | N/A (engine-level) | Yes - CometIcebergNativeScanExec defers serialization for DPP |
| Row-group filtering (residuals) | Yes | Yes | Yes - residual predicates converted to iceberg::expr::BoundPredicate |
| Identity partition columns | Yes | Yes | Yes |
| Object stores (S3/GCS/OSS) | Yes (Hadoop FS) | Yes (OpenDAL) | Yes (OpenDAL via FileIOBuilder) |
| V1 spec | Yes | Yes | Yes |
| V2 spec | Yes | Yes | Yes |
| V3 spec metadata | Yes | Yes (FormatVersion::V3, next_row_id, row lineage) | Not used - Comet doesn't handle V3-specific metadata |

### WRITE PATH

| Feature | Iceberg Java | iceberg-rust | datafusion-comet |
|---|---|---|---|
| Data file writing | Yes - DataWriter | Yes - DataFileWriter | No - uses raw parquet crate, not iceberg-rust |
| Partitioned writes (sorted) | Yes - ClusteredDataWriter | Yes - ClusteredWriter | No - writes single file per Spark partition |
[...]

PR #2447, opened 2025-09-23
feat: Parquet Modular Encryption with Spark KMS for native readers

mbutrovich:
## Which issue does this PR close?

Closes #.

## Rationale for this change

We want to add Parquet Module Encryption support for the native readers when using a Spark KMS. We use the encryption factory features added in DataFusion 50 to register an encryption factory that uses JNI to get decryption keys from Spark.

## How are these changes tested?

- Existing PME tests with new readers added.
- New tests that exercise PME options like plaintext footer, etc.

PR #4991, opened 2026-07-21
feat: Iceberg table format V3: native table decryption, fall back for other V3 features

mbutrovich:

Part of #3756. Closes #.

## Rationale for this change

Comet's native Iceberg scan rejected V3 tables. This adds native reads of encrypted Iceberg tables (a V3 feature).

Iceberg-Java unwraps the key envelope on the driver during planning and stores each data file's plaintext data encryption key in the manifest's `key_metadata` field. So `DataFile.keyMetadata()` already holds the plaintext key when Comet reads the scan tasks. Comet forwards those bytes to iceberg-rust, which decodes `StandardKeyMetadata` and uses the key directly for Parquet decryption. No native KMS or JNI is needed (unlike plain-Parquet Modular Encryption in #2447, where the reader unwraps at read time).

- Bump arrow-rs to `58.4.0` and iceberg-rust to `4532da8` for the encryption read path (apache/iceberg-rust#2584 adds encrypted-data-file reads and `FileScanTask::key_metadata`; apache/iceberg-rust#2878 adds AES-256 data keys).
- proto: `optional bytes key_metadata` on `IcebergFileScanTask` (field 6) and `IcebergDeleteFile` (field 8; 5 to 7 reserved for a separate deletion-vector branch).
- serde (`CometIcebergNativeScan`): forward `ContentFile.keyMetadata()` for data and delete files. Unset when null.
- planner: populate `FileScanTask`/`FileScanTaskDeleteFile` `key_metadata` from the proto.
- `CometScanRule` / `IcebergReflection`: allow format V3 (reject V4+); fall back for V3 default values, for unsupported column types (allow-list over the full table schema), and for 192-bit data keys (no AES-192-GCM in the underlying crypto; 128- and 256-bit are supported); keep the delete gate Parquet-only so deletion-vector tables still fall back. Table encryption itself is not a fallback reason.
- Tests: `CometTestKMS` (in-memory `KeyManagementClient`, spark-4.1 source set since `KeyManagementClient` is public only from Iceberg 1.10 and encryption needs 1.11); `spark-hive` added as a test dependency to stand up an embedded Hive metastore.
- `dev/diffs/iceberg/1.11.0.diff`: `testMissingRequiredWithoutDefault` tolerates Comet's exception wrapping.

- Rust `standard_key_metadata_roundtrips`: `StandardKeyMetadata` encode/decode round-trips (API-stability guard after the bump).
- Rust `decodes_java_produced_key_metadata`: decodes a real Iceberg-Java blob and recovers the identical 16-byte DEK and AAD prefix (cross-language wire-format guard).
- Scala `CometIcebergEncryptionSuite.data file key_metadata carries a plaintext DEK that round-trips`: `StandardEncryptionManager` stores a plaintext (not wrapped) DEK in `key_metadata` and the blob round-trips.
- Scala `CometIcebergEncryptionSuite.encrypted V3 table is read natively and matches Spark`: end-to-end. A `type=hive` Iceberg catalog (embedded Derby metastore) with `CometTestKMS` creates an encrypted V3 table; the test asserts `key_metadata` is populated, that results match Spark, and that the read runs through `CometIcebergNativeScanExec`.
- Scala `CometIcebergEncryptionSuite.encrypted V3 table with a 256-bit data key is read natively and matches Spark`: same, with `encryption.data-key-length=32`.
- Scala `CometIcebergEncryptionSuite.encrypted V3 table with a 192-bit data key falls back to Spark`: a table with `encryption.data-key-length=24` falls back and still returns correct results.
- Iceberg's own `TestTableEncryption` also runs under Comet in the iceberg-spark CI (via `dev/diffs/iceberg/1.11.0.diff`, which enables native Iceberg scan), reading a real encrypted V3 table.
- Ran `org.apache.iceberg.spark.sql.TestTableEncryption` locally against Iceberg 1.11 with the diff applied (13 tests, all passing) and inspected the captured logs: the `INFO org.apache.comet.serde.operator.CometIcebergNativeScan - IcebergScan: ...` lines confirm the encrypted data reads (`testSelect` and `testInsertAndDelete`) engage `CometIcebergNativeScan` rather than falling back. Those tests assert exact result sets, so the reads are decrypted correctly through the native path. The remaining fallbacks are metadata-table scans (`ALL_DATA_FILES`, `MANIFESTS`, `all_manifests`), which are unrelated to the data path.


docs/source/user-guide/latest/iceberg.md
```diff
@@ -63,7 +63,14 @@ The native Iceberg reader supports the following features:
 
 **Table specifications:**
 
-- Iceberg table spec v1 and v2 (v3 will fall back to Spark)
+- Iceberg table spec v1, v2, and v3
+
+**Encryption:**
+
+- Encrypted v3 tables using 128-bit AES-GCM data keys (requires Iceberg 1.11 or newer). Iceberg-Java
```
hsiang-c: (nit) I think we could mention 192-bit and 256-bit are allowed in Iceberg Java: https://github.com/apache/iceberg/blob/apache-iceberg-1.11.0/core/src/main/java/org/apache/iceberg/encryption/EncryptionUtil.java#L117-L120

spark/pom.xml
```diff
@@ -107,6 +107,13 @@ under the License.
       <artifactId>spark-hadoop-cloud_${scala.binary.version}</artifactId>
       <classifier>tests</classifier>
     </dependency>
+    <dependency>
+      <!-- Test-only: provides an embedded Hive metastore so Iceberg's type=hive catalog can
+           create encrypted tables (only Hive wires the KMS-backed EncryptionManager). Version and
+           scope come from the parent pom's dependencyManagement. -->
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-hive_${scala.binary.version}</artifactId>
```
hsiang-c, line 115: Do we need 

```xml
<scope>test</scope>
```

spark/src/test/spark-4.1/org/apache/comet/CometIcebergEncryptionSuite.scala
```diff
@@ -0,0 +1,211 @@
...
+      assert(
+        java.util.Arrays.equals(ByteBuffers.toByteArray(dek), roundTripped),
+        "DEK did not survive the StandardKeyMetadata encode/parse round-trip")
+    }
+  }
+
+  // The native Parquet reader (arrow-rs 58.3) decrypts only 128-bit AES-GCM keys, so a table with a
+  // larger data key must fall back to Spark and still return correct results. Flip this to a native
+  // assertion when arrow-rs 58.4 (256-bit support, arrow-rs#10349) is picked up.
+  test("encrypted V3 table with a 256-bit data key falls back to Spark") {
```
hsiang-c: 👍

spark/src/test/spark-4.1/org/apache/comet/CometIcebergEncryptionSuite.scala
```diff
@@ -0,0 +1,211 @@
...
+        TBLPROPERTIES ('format-version' = '3', 'encryption.key-id' = '${CometTestKMS.MasterKeyId}')
+      """)
+      spark.sql(s"""
+        INSERT INTO $table VALUES (1, 'Alice', 10.5), (2, 'Bob', 20.3), (3, 'Charlie', 30.7)
+      """)
+
+      // Guard against a silently-plaintext table (e.g. catalog not wiring encryption): a genuinely
+      // encrypted table records a plaintext DEK blob per data file in key_metadata.
+      val keyMetadatas = spark
+        .sql(s"SELECT key_metadata FROM $table.files WHERE content = 0")
```
hsiang-c, line 126: The `file_path` columns is part of the `.files` table and one check I have is reading the first 4 bytes of the Parquet file see if it is the `PARE` magic string.

hsiang-c (approved):
LGTM

docs/source/user-guide/latest/iceberg.md
mbutrovich: Good call. Updated the docs to note that Iceberg-Java also permits 192- and 256-bit data keys, and that those tables fall back to Spark for now since the native Parquet reader decrypts only 128-bit AES-GCM until arrow-rs 58.4.

spark/pom.xml
mbutrovich: The scope is inherited from the parent pom's dependencyManagement, which declares spark-hive as test scope. That is why it is omitted here, matching the spark-sql and spark-hadoop-cloud test deps just above. The inline comment notes this. Happy to make it explicit if you would prefer it spelled out on the dependency itself.

spark/src/test/spark-4.1/org/apache/comet/CometIcebergEncryptionSuite.scala
mbutrovich: Thanks, that is a clean way to confirm encryption at rest. I left it out here because the suite already proves the same thing end to end: it asserts key_metadata is populated and that the read goes through CometIcebergNativeScanExec while matching Spark. For that native result to be correct, arrow-rs had to decrypt the file with the DEK from key_metadata. A plaintext PAR1 file read with decryption properties set would error rather than return correct rows, so a passing native answer already implies the bytes were encrypted and decrypted correctly. Reading the PARE magic would check the same property one layer lower, inside a detail Iceberg and arrow-rs own, so I would rather keep the test at the API boundary.

mbutrovich:
> LGTM

Thanks for the review @hsiang-c!

.github/actions/java-test/action.yaml
```diff
@@ -50,6 +50,18 @@ runs:
         cd native
         cargo build --release
 
+    # Evaluate hashFiles now, on the clean checkout, and reuse the result in the
+    # cache key below. The cache step's inputs are re-evaluated when its post
+    # (cache-save) phase runs, and by then the test run has left trees under the
+    # workspace (e.g. the Hive-catalog warehouse from CometIcebergEncryptionSuite)
+    # that make a fresh '**/pom.xml' walk throw "Fail to hash files under
+    # directory". A precomputed string keeps the save phase from re-walking them.
```
mbutrovich, line 58: I also added some cleanup logic for the Hive-catalog, but just in case I'm also addressing the footgun here. It fixed a failure in the `Post Java test steps`:

```
Post job cleanup.
Error: The template is not valid. /home/runner/work/datafusion-comet/datafusion-comet/./.github/actions/java-test/action.yaml (Line: 63, Col: 14): hashFiles('**/pom.xml') failed. Fail to hash files under directory '/home/runner/work/datafusion-comet/datafusion-comet'
```
that I saw after adding the Hive catalog https://github.com/apache/datafusion-comet/actions/runs/29912607672/job/88905543943

native/proto/src/proto/operator.proto
```diff
@@ -277,6 +282,13 @@ message IcebergDeleteFile {
 
   // Equality field IDs (empty for positional deletes)
   repeated int32 equality_ids = 4;
+
+  // Serialized StandardKeyMetadata for an encrypted delete file, from DeleteFile.keyMetadata().
+  // Absent for unencrypted tables. Decoded directly by iceberg-rust; no KMS unwrap needed.
+  // Field 8: fields 5-7 are reserved for the deletion-vector coordinates on the dv-read branch
```
andygrove: Should we add an explicit `reserved 5, 6, 7` ?

andygrove (approved):
LGTM with one suggested change. Thanks @mbutrovich

spark/src/test/spark-4.1/org/apache/comet/CometIcebergEncryptionSuite.scala
```diff
@@ -0,0 +1,261 @@
...
+
+      assert(
+        java.util.Arrays.equals(ByteBuffers.toByteArray(dek), roundTripped),
+        "DEK did not survive the StandardKeyMetadata encode/parse round-trip")
+    }
+  }
+
+  // A 256-bit (32-byte) AES-GCM data key is supported natively, so the read must run through Comet
+  // and match Spark.
+  test("encrypted V3 table with a 256-bit data key is read natively and matches Spark") {
```
hsiang-c, line 194: This updated test LGTM, thanks @mbutrovich

Issue #4993, opened 2026-07-21
Object store cache key drops the ABFS container, so two containers in one storage account share a store instance


andygrove:
## Describe the bug

`prepare_object_store_with_configs` in `native/core/src/parquet/parquet_support.rs` caches object store instances under `(url_key, config_hash)`, where `url_key` is built from:

```rust
let url_key = format!(
    "{}://{}",
    scheme,
    &url[url::Position::BeforeHost..url::Position::AfterPort],
);
```

`Position::BeforeHost` starts *after* the URL userinfo. For ABFS, the container is the userinfo, not the host:

```
abfss://container@account.dfs.core.windows.net/path/to/file.parquet
```

So `abfss://c1@acct.dfs.core.windows.net/...` and `abfss://c2@acct.dfs.core.windows.net/...` both produce the key `abfss://acct.dfs.core.windows.net`. `config_hash` is computed over the per-session Hadoop config map, which does not vary by container, so it does not disambiguate them either.

Meanwhile `MicrosoftAzureBuilder::parse_url` bakes the container into the store instance from the URL userinfo, and the `Path` handed to that store is container-relative (the bucket/container is stripped by `Path::from_url_path(url.path())` in `PhysicalPlanner::get_partitioned_files`).

The consequence: within one executor process, reading `abfss://c2@acct/data/f.parquet` after `abfss://c1@acct/data/f.parquet` should hit the cached `c1` store and read **c1's data**.

## Steps to reproduce

Not yet reproduced. This was found by code reading during review of an unrelated change. Confirming it needs an Azure storage account with two containers holding same-named paths, read from a single executor in one application.

## Expected behavior

Each ABFS container resolves to its own object store instance, and a read from one container never returns another container's data.

## Additional context

- S3, GCS, and HDFS are unaffected: their bucket/host lives in the URL host component, which `url_key` does keep.
- The same key derivation is now also used to key a shared Parquet metadata cache, where the impact is narrower (entries are additionally validated against file size and modification time). That is documented as a known limitation there; this issue is about the store cache itself, where no such validation exists.
- The object store cache is process-wide and unbounded by design, so a wrong entry persists for the executor's lifetime.

peterxcli:
let me take a stab at this.

PR #4995, opened 2026-07-21
feat: share Parquet metadata cache across tasks in an executor

### LINKED ISSUE #4994: Share Parquet footer and page index metadata across tasks in an executor

Comet's native Parquet scan caches each file's footer and page index in DataFusion's `FileMetadataCache`, which it obtains from the task's `RuntimeEnv` (`native/core/src/parquet/parquet_exec.rs`). Comet builds a fresh `RuntimeEnv` for every Spark task, so that cache is discarded when the task ends.

Two tasks on the same executor therefore re-fetch and re-parse the same footer, once each. This recurs in two common shapes:

- a large Parquet file split into many Spark partitions, each read by a different task;
- the same table scanned again by a later stage or query in the same application.

Each miss costs a round trip to object storage plus a Thrift parse. Spark's own reader has the same per-task behavior, so this is parity with Spark, but Velox caches file handles and footers process-wide, and this is one of the gaps identified as item P5 in #4842.

Sharing the cache also reduces memory: today each concurrent task gets its own 50MB budget.

Hold the metadata caches in a process-wide registry keyed by object store URL, and use it at the scan's single call site, falling back to the per-task cache when sharing is disabled.

This requires the cache validity check to actually work first. DataFusion validates cached metadata against `(size, last_modified)`, but Comet builds its `PartitionedFile` with `PartitionedFile::new_with_range`, which hardcodes `last_modified` to epoch 0, and `SparkPartitionedFile` carries no modification time. Validation is therefore size-only today. That is harmless for a cache that lives for one task and a wrong-results hazard for one that lives for the executor, so Spark's `PartitionedFile.modificationTime` needs to reach `object_meta.last_modified`.

## Additional context

Item P5 of #4842.

andygrove:

Closes #4994. This is item P5 of #4842.

## Rationale for this change

Comet's native Parquet scan takes its `FileMetadataCache` from the task's `RuntimeEnv`, and Comet builds a fresh `RuntimeEnv` for every Spark task, so the cached footer and page index are discarded when the task ends. Two tasks on one executor re-fetch and re-parse the same footer, which costs a round trip to object storage plus a Thrift parse per task. This happens whenever a large file is split across tasks, and again whenever a later stage or query rescans the same table.

Sharing the cache across tasks also lowers memory use: the 50MB budget was previously allocated per task, so an executor with 8 task slots could hold 8 budgets at once.

Making the cache outlive a task requires its validity check to work first. DataFusion validates cached metadata against `(size, last_modified)`, but Comet builds `PartitionedFile` with `PartitionedFile::new_with_range`, which hardcodes `last_modified` to epoch 0, and `SparkPartitionedFile` carried no modification time. Validation was therefore size-only. That is harmless for a cache that lives for one task and a wrong-results hazard for one that lives for the executor process, so the modification time is plumbed through as part of this change rather than separately.

Carry the file modification time to the native side:

- `SparkPartitionedFile` gains a `modification_time` field, populated from Spark's `PartitionedFile.modificationTime` in `partition2Proto`.
- `PhysicalPlanner::get_partitioned_files` sets `object_meta.last_modified` from it, so DataFusion's `(size, last_modified)` validation is real.

Share the cache:

- New `native/core/src/parquet/metadata_cache.rs` holds a process-wide registry of `FileMetadataCache` instances, keyed by object store URL. It parallels the existing process-wide `object_store_cache` in `parquet_support.rs`, including its rationale for process lifetime. Keying per object store matters because `DefaultFilesMetadataCache` keys entries on a bare `object_store::path::Path`, and Comet builds that path from `url.path()`, which drops the bucket.
- `init_datasource_exec` uses the registry, falling back to the per-task `RuntimeEnv` cache when sharing is disabled.

Configuration and docs:

- `spark.comet.scan.metadataCache.enabled` (default `true`) and `spark.comet.scan.metadataCache.memoryLimit` (default 50MB, matching DataFusion's own default). Both are injected explicitly in `serializeCometSQLConfs`, since `SQLConf.getAllConfs` carries only explicitly-set values.
- A tuning guide section covering the memory scope, eviction, and validation.

Two known limitations are documented rather than fixed here. The memory limit bounds each object store's cache within an executor, so an executor reading several buckets can hold a multiple of it. And for `abfs://` URLs the container is URL userinfo rather than host, so Comet's object store URL key does not separate two containers in one storage account; those entries rely on `(size, last_modified)` validation. The second limitation stems from a pre-existing issue in the object store cache itself, filed as #4993.

New Rust tests:

- `metadata_cache.rs`: the same URL returns one instance, different buckets get separate instances, a disabled registry returns `None`, the configured limit reaches new and live caches, and concurrent callers on one URL all receive the same instance.
- `parquet_exec.rs`: two independent `SessionContext`s (standing in for two tasks) over one file, asserting the second sees a cache hit through the entry's hit counter; and a file whose `last_modified` advances, asserting the cached entry is replaced rather than served stale. The existing `caches_full_metadata_with_page_index` regression test now targets the shared cache.
- `planner.rs`: the modification time from the protobuf reaches `object_meta.last_modified`.

New Scala tests:

- `CometFilePartitionSerdeSuite`: the modification time is serialized into the protobuf.
- `CometMetadataCacheSuite`: both configs cross JNI with their defaults; a repeated read of the same files exercises a shared-cache hit across queries in one process; and a scan with sharing disabled exercises the per-task fallback path.

Also verified locally: `make` build, `cargo clippy --all-targets --workspace -- -D warnings`, `cargo fmt`, `spotless:check`, the native `parquet` (92) and `planner` (16) test modules, and `ParquetReadV1Suite` (51).

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -303,6 +303,8 @@ jobs:
           - name: "scans"
             value: |
               org.apache.comet.parquet.CometParquetWriterSuite
+              org.apache.comet.CometMetadataCacheSuite
+              org.apache.comet.serde.CometFilePartitionSerdeSuite
               org.apache.comet.parquet.ParquetReadV1Suite
               org.apache.comet.parquet.ParquetReadV2Suite
               org.apache.comet.parquet.ParquetReadFromFakeHadoopFsSuite
--- a/.github/workflows/pr_build_macos.yml
+++ b/.github/workflows/pr_build_macos.yml
@@ -119,6 +119,8 @@ jobs:
           - name: "scans"
             value: |
               org.apache.comet.parquet.CometParquetWriterSuite
+              org.apache.comet.CometMetadataCacheSuite
+              org.apache.comet.serde.CometFilePartitionSerdeSuite
               org.apache.comet.parquet.ParquetReadV1Suite
               org.apache.comet.parquet.ParquetReadV2Suite
               org.apache.comet.parquet.ParquetReadFromFakeHadoopFsSuite
--- a/docs/source/user-guide/latest/tuning.md
+++ b/docs/source/user-guide/latest/tuning.md
@@ -110,11 +110,12 @@ different splits of the same file, or later stages re-reading the same table, do
 and re-parse it. The cache is shared by all tasks in an executor process and is enabled by
 default.
 
-`spark.comet.scan.metadataCache.memoryLimit` (default 50MB) bounds the cache in each executor,
-with least-recently-used eviction. Note that this budget is per executor rather than per task,
-so it does not scale with the number of task slots. Entries are validated against each file's
-size and modification time, so an overwritten file is re-read rather than served from the
-cache.
+`spark.comet.scan.metadataCache.memoryLimit` (default 50MB) bounds, with least-recently-used
+eviction, the cache for each distinct object store an executor reads from, so an executor
+reading from several object stores, or several buckets, holds a multiple of this limit. It does
+not scale with the number of task slots, however. Entries are validated against each file's size
+and modification time (validation falls back to size alone if the source reports no modification
+time), so an overwritten file is normally re-read rather than served from the cache.
 
 Set `spark.comet.scan.metadataCache.enabled=false` to give each task its own cache instead.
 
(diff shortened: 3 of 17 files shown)


mbutrovich:
I remember @comphead tried this and we always ran into a Spark SQL test failure where it creates a file, reads it, deletes it, then reads it again. We'd fail for serving the query from metadata (i.e., cache invalidation wasn't managed).

andygrove:
> I remember @comphead tried this and we always ran into a Spark SQL test failure where it creates a file, reads it, deletes it, then reads it again. We'd fail for serving the query from metadata (i.e., cache invalidation wasn't managed).

Yup, I hit the same issue.

PR #4821, opened 2026-07-03
feat: support Spark 4.1 TIME type and expressions via codegen dispatch

andygrove:
## Summary

- Wires the Spark 4.1 `TIME` data type end-to-end and adds JVM codegen dispatch for the new TIME expressions (`HoursOfTime`, `MinutesOfTime`, `SecondsOfTime`, `SecondsOfTimeWithFraction`, `TimeAddInterval`, `SubtractTimes`, `TimeDiff`, `TimeTrunc`) and casts touching `TimeType`. No native implementations added.
- Teaches the batch-kernel input/output emitters, `CometScalaUDFCodegen`, and `ArrowWriters` to handle `TimeNanoVector` as a Long-backed vector.
- Parquet scans still opt out of `TimeType` (no native decoder yet); the fallback reason is now explicit.

## Details

Every new TIME expression in Spark 4.1 is a `RuntimeReplaceable` that rewrites to `StaticInvoke(DateTimeUtils, ...)` before physical planning. The shim (`spark-4.1+/CometExprShim.scala`) matches those `StaticInvoke` forms and routes them through `CometScalaUDF.emitJvmCodegenDispatch`, which runs Spark's own `doGenCode` inside the Comet batch kernel. `CurrentTime` needs no work — the optimizer folds it to a literal. `CometCast` routes any cast touching `TimeType` through the same dispatcher.

Existing `make_time` / `to_time` / `try_to_time` continue to use the native scalar-function paths already wired up.

`LocalTableScan` for `TimeType` columns now works natively via the new `TimeNanoWriter` in `ArrowWriters`. The pre-existing "falls back when schema contains TimeType" test flips to "handles TimeType column".

Known limitation: `time - time` (result type `DayTimeIntervalType`) still falls back at the projection level because Comet's generic type-support list doesn't cover `DayTimeIntervalType` outputs. Pre-existing, not TIME-specific — left as a `checkSparkAnswer` regression test in the suite.

## Test plan

- [x] `./mvnw test -Dsuites="org.apache.spark.sql.CometTimeTypeSuite" -Dtest=none` — 15/15 pass on Spark 4.1 (new suite)
- [x] `./mvnw test -Dsuites="org.apache.comet.exec.CometExecSuite" -Dtest=none` — 131/131 pass (LocalTableScan regression check)
- [x] `./mvnw test -Dsuites="org.apache.comet.CometCastSuite" -Dtest=none` — 161/161 pass (cast regression check)
[...]

PR #4997, opened 2026-07-21
feat: extract TimeType to int/decimal

### LINKED ISSUE #4983: Implement TimeType support - Time extraction (from TimeType to Int/Decimal)

What is the problem the feature request solves?
This is https://github.com/apache/datafusion-comet/issues/4288 sub-issue

These are the dedicated TimeType expressions from timeExpressions.scala. All are
RuntimeReplaceable and resolve to StaticInvoke(DateTimeUtils, methodName, ...).
They follow the same shim pattern as make_time.

time extraction (from TimeType to Int/Decimal)

| SQL | Spark expression | StaticInvoke method | Return type |
|---|---|---|---|
| `extract(HOUR FROM t)` | `HoursOfTime` | `getHoursOfTime` | IntegerType |
| `extract(MINUTE FROM t)` | `MinutesOfTime` | `getMinutesOfTime` | IntegerType |
| `extract(SECOND FROM t)` | `SecondsOfTime` | `getSecondsOfTime` | IntegerType |
| `extract(SECOND FROM t)` (with fraction) | `SecondsOfTimeWithFraction` | `getSecondsOfTimeWithFraction` | DecimalType |

**Shim pattern**: Match `StaticInvoke` where `staticObject == classOf[DateTimeUtils.type]`
and `functionName` is one of the above.

**Rust implementation**: Simple arithmetic on nanoseconds since midnight:
- hours = nanos / NANOS_PER_HOUR
- minutes = (nanos % NANOS_PER_HOUR) / NANOS_PER_MINUTE
- seconds = (nanos % NANOS_PER_MINUTE) / NANOS_PER_SECOND

### Additional context

_No response_

YutaLin:

Closes #4983 

## Rationale for this change

Support time extraction (from TimeType to Int/Decimal)

| SQL | Spark expression | StaticInvoke method | Return type |
|---|---|---|---|
| `extract(HOUR FROM t)` | `HoursOfTime` | `getHoursOfTime` | IntegerType |
| `extract(MINUTE FROM t)` | `MinutesOfTime` | `getMinutesOfTime` | IntegerType |
| `extract(SECOND FROM t)` | `SecondsOfTime` | `getSecondsOfTime` | IntegerType |
| `extract(SECOND FROM t)` (with fraction) | `SecondsOfTimeWithFraction` | `getSecondsOfTimeWithFraction` | DecimalType |

* Adds native Comet support for extracting hour, minute, and second from Spark TimeType values.
* Adds serde handling for Spark’s DateTimeUtils StaticInvoke expressions used by TimeType extraction.
* Implements fractional-second extraction as Decimal(8, 6) from Arrow Time64(Nanosecond) values.
* Registers the new second_with_fraction native scalar function.
* Adds Rust unit tests and Spark SQL coverage for hour, minute, second, extract(SECOND FROM ...), and date_part('SECOND', ...).

- Added Rust unit tests for extracting hour, minute, integer second, and fractional second from `Time64(Nanosecond)`, including multiple precision values.
- Added Spark SQL tests in `time_extract.sql` covering `hour`, `minute`, `second`, `extract(SECOND FROM ...)`, and `date_part('SECOND', ...)`.
- Verified the SQL results against Spark while requiring native Comet execution.
- Ran the SQL test suite with Spark 4.1.


YutaLin:
Hi @parthchandra, could you help me review this? Thanks!

cc @andygrove

andygrove (commented):
Thanks for this, the PR is clean and well tested. I read the four `DateTimeUtils` methods and the `timeExpressions.scala` expression classes in Spark to check the behavior, and the acceleration path matches Spark closely. A few notes below, mostly one scope question and a couple of small suggestions.

**Integral extraction looks correct.** The three integral parts map cleanly to `nanosToLocalTime(nanos).getHour/getMinute/getSecond`, and routing them through the existing `date_part` kernels with `timezone = "UTC"` is right. The nice detail is timezone independence: for a `Time64(Nanosecond)` input, `array_with_timezone` falls through its final `_ => Ok(array)` arm, so the array passes through untouched and arrow's `date_part` reads the time-of-day fields directly. The unit test checking `hour`/`minute`/`second` across UTC, LA, and Tokyo is a good guard for exactly that.

**The fractional-second math matches Spark's truncation.** Spark's `getSecondsOfTimeWithFraction` does `(nanos % NANOS_PER_SECOND) * scaleFactor / NANOS_PER_SECOND` with integer division, so it truncates. The native `second_with_fraction` truncates the same way with pure `i128` arithmetic. Using integers instead of Spark's `Double` intermediate is arguably more robust, and at scale 6 the results are identical. The unit test at precisions 0, 3, 6 confirms it.

**Main thing worth confirming: the fractional path only supports `TIME(6)`.** Spark's `SecondsOfTimeWithFraction` declares its return type as `DecimalType(2 + p, p)` where `p` is the child `TimeType` precision, and `TimeType` allows `p` in `[0, 9]` (`MAX_PRECISION = NANOS_PRECISION = 9`). The shim guard is `s.dataType == DecimalType(8, 6)`, which only admits `p = 6`, and the native function reinforces that by always emitting `Decimal128(8, 6)`. So `extract(SECOND FROM t)` over a `TIME(3)` or `TIME(9)` column will not match the shim and falls back to Spark. That is safe and correct, just not accelerated. It happens to cover the common case because `make_time` always produces `TIME(MICROS_PRECISION)` = `TIME(6)`, which is why all the tests hit the native path. Is limiting to `TIME(6)` the intended scope for this PR? If so, a one-line comment on the `DecimalType(8, 6)` guard explaining the restriction would help, and a follow-up issue to generalize the other precisions would be nice to link. The integral `hour`/`minute`/`second` path already handles every precision, so only the fractional case is limited.

**Minor guard inconsistency.** `isValidTimePrecision` bounds the literal by `TimeType.MAX_PRECISION` (9), but the native `second_with_fraction` returns an `Execution` error for any precision above 6 and always writes scale 6. Today this never bites because the `s.dataType == DecimalType(8, 6)` guard pins precision to 6 before `isValidTimePrecision` is reached. If that dataType guard is ever relaxed, the two checks would disagree and `TIME(7..9)` would produce a native error rather than a clean fallback. It might be clearer to bound `isValidTimePrecision` by `MICROS_PRECISION` (6) with a comment, so the Scala and Rust limits agree.

**Test coverage suggestion.** All the SQL cases build on `make_time`, so every value is `TIME(6)`. It might be worth adding a `CAST(make_time(...) AS TIME(3))` (and maybe `TIME(9)`) case with `spark_answer_only` to document that lower and higher precisions still return correct results through the fallback path. An integral `hour`/`minute` case at a non-6 precision would also exercise the all-precision native path.

No `expressions.md` change is needed since `hour`, `minute`, `second`, `extract`, and `date_part` are already documented and `second_with_fraction` is an internal kernel rather than a user-facing function.

Overall this looks good. The only real discussion point is whether the `TIME(6)`-only restriction on fractional seconds is intended, and making that boundary explicit in the code.

native/spark-expr/src/datetime_funcs/extract_date_part.rs
```diff
@@ -16,13 +16,21 @@
...
-use std::fmt::Debug;
+use std::{fmt::Debug, sync::Arc};
+
+const NANOS_PER_SECOND: i64 = 1_000_000_000;
+const MICROS_PER_SECOND: i128 = 1_000_000;
+const SECOND_DECIMAL_PRECISION: u8 = 8;
+const SECOND_DECIMAL_SCALE: i8 = 6;
 
 /// Returns true when the type is a timestamp without a timezone (Spark's TimestampNTZType),
 /// including when wrapped in a dictionary. Such values store local wall-clock time and must not
```
parthchandra, line 36: Nit: `MICROS_PER_SECOND` is technically correct (1M µs/s) but here it serves as the Decimal(8,6) unscaled-unit factor. A name like `UNSCALED_UNITS_PER_SECOND` or just inlining `10_i128.pow(SECOND_DECIMAL_SCALE as u32)` would make the intent clearer — currently readers have to reason about why "micros" appears in a nanosecond-based function.

native/spark-expr/src/datetime_funcs/extract_date_part.rs
```diff
@@ -117,10 +125,113 @@ extract_date_part!(SparkHour, "hour", Hour);
 extract_date_part!(SparkMinute, "minute", Minute);
 extract_date_part!(SparkSecond, "second", Second);
 
+fn second_with_fraction(nanos: i64, precision: i32) -> Result<i128> {
+    if !(0..=SECOND_DECIMAL_SCALE as i32).contains(&precision) {
+        return Err(DataFusionError::Execution(format!(
+            "second_with_fraction: precision must be between 0 and 6, found {precision}"
```
parthchandra, line 131: Defensive suggestion: `second_with_fraction` accepts `i64` but silently produces wrong results for negative nanos (Rust `%` preserves sign). TimeType values are always non-negative today, so this is not a live bug, but a `debug_assert!(nanos >= 0)` would guard against future misuse.

parthchandra (commented):
Minor nits only — Andy's review covers the substantive design points well.

andygrove:
Triage note: this and my #4821 both add a `StaticInvoke` arm for `getHoursOfTime`, `getMinutesOfTime`, `getSecondsOfTime` and `getSecondsOfTimeWithFraction` to `sparkVersionSpecificExprToProtoInternal` in the same `spark-4.1+/CometExprShim.scala`, so they conflict on exactly those four expressions. Mine routes them through the JVM codegen dispatcher, which was only ever meant to get the whole Spark 4.1 TIME family off Spark fallback quickly — a native implementation is the better end state.

I would rather this one land and shrink #4821 to the expressions it does not cover (`TimeAddInterval`, `SubtractTimes`, `TimeDiff`, `TimeTrunc`) plus the vector plumbing. Does that work for you?

PR #4999, opened 2026-07-22
chore: disable ORC in Iceberg Spark tests

### LINKED ISSUE #4586: ci: Remove ORC and Avro data file testing from Iceberg Java tests
Currently we run the entirety of the Iceberg Java tests against Comet. Comet doesn't support accelerated ORC reads, and has no plans to do so. We should modify our Iceberg diffs to change the parameterized formats exercised to remove ORC, which we already fall back for. I would say we could skip table format V3 at the moment, but we'll hopefully start picking up support for V3 some features soon.

hsiang-c:

Closes #4586

## Rationale for this change

 - Reduce CI resource consumption by skipping ORC related Iceberg Spark tests b/c Comet doesn't accelerate Iceberg ORC scans.

  - Scope: Iceberg Spark 4.1 tests only
  - Summarized with the help of Claude

  | # | Pattern | What to do | Applied to |
  |---|---------|-----------|------------|
  | 1 | Drop ORC rows from parameterized `parameters()` providers | Delete the `Object[]` / `Object[][]` entries that pass `FileFormat.ORC` or `"orc"` | `SparkRowLevelOperationsTestBase`, `TestRowLevelOperationsWithLineage`, `TestFilteredScan`, `TestPartitionPruning`, `TestPartitionValues`, `TestSparkReadProjection`, `TestSparkMetadataColumns`, `TestSparkReaderDeletes`, `TestSparkScan`, `TestSparkDataWrite`, `TestIdentityPartitionData`, `TestMetadataTablesWithPartitionEvolution`, `TestPositionDeletesTable`, `TestDataFrameWriterV2Coercion`, `TestCompressionSettings` |
  | 2 | Disable whole ORC-only test classes | Add `import org.junit.jupiter.api.Disabled;` and annotate the class with `@Disabled` | `TestOrcWrite`, `TestSparkOrcReader`, `TestSparkOrcReadMetadataColumns`, `TestSparkRecordOrcReaderWriter`, `TestVectorizedOrcDataReader`, `TestORCDataFrameWrite` |
  | 3 | Disable individual ORC test methods | Replace `@TestTemplate` / `@Test` with `@Disabled` on the single ORC case |   `TestRewriteTablePathsAction#testPositionDeletesOrc()` |
  | 4 | Remove ORC-specific assertions / setup / imports inline | Strip the ORC statements and now-unused static imports from tests that mix ORC with other formats | `TestSparkCompressionUtil` (delete `testOrcCompressionRatios()` + ORC branches, remove `import static …FileFormat.ORC`); `TestSparkWriteConf` (delete ORC compression cases + `ORC_COMPRESSION`, `ORC_COMPRESSION_STRATEGY`, `DELETE_ORC_COMPRESSION`, `DELETE_ORC_COMPRESSION_STRATEGY` imports); `TestCompressionSettings` (drop `ORC_COMPRESSION`/`DELETE_ORC_COMPRESSION` `put()` calls + imports); `TestStructuredStreamingRead3` (remove `orcFileRecords`, its `appendData(…, "orc")`, and the ORC part of the expected-result assertion) |

Running Iceberg Spark 4.1 with the new `1.11.0.diff`


mbutrovich:
One thing we might want to add is a test to CometIcebergNativeSuite that uses ORC as the data file type, that way we at least have a self-contained test that the fallback is working correctly.

mbutrovich:
Interestingly, this doesn't seem to save much time in CI.

andygrove:

Skipping tests for a format Comet does not accelerate is a reasonable way to buy back CI time, and the pattern table in the description is a good way to make a 600-line diff reviewable. The new `CometIcebergNativeSuite` test asserting that an ORC-backed Iceberg table falls back with a specific reason is the right companion, since it pins the premise the whole change rests on.

Four things.

**What does this actually save?**

The stated goal is reducing CI resource consumption, but there is no number. How much wall clock does the Iceberg Spark 4.1 job drop by? That is the entire justification for taking on 600 lines of diff maintenance, and it should be in the description. If it turns out to be two minutes out of seventy, the trade is different than if it is twenty.

This is also worth weighing against #5459, which is sharding the same job across four runners. If that lands, wall clock stops being the constraint and this becomes purely about runner minutes.

**Diff maintenance cost**

`dev/diffs/iceberg/1.11.0.diff` grows by 600 lines, and every one of those hunks has to be rebased whenever Iceberg changes those test files. Iceberg 1.11 will get patch releases, and 1.12 will come. My memory of these files is that they are already painful to carry forward.

Is there a lighter mechanism? A JUnit tag filter, a Gradle exclusion by class name, or a system property that the workflow passes, would express "skip ORC" in a handful of lines that do not conflict on every upstream change. `@Disabled` on six whole classes in particular seems like something a `-PexcludeTags` or an exclusion pattern could do from outside the source.

**Scope is 1.11 only**

The description says Iceberg Spark 4.1 tests only. The other three Iceberg versions still run the ORC tests. Is the intent to extend this later, or is 1.11 the only version that runs on every PR (with 1.8, 1.9, and 1.10 gated behind the label) so it is the only one that matters? A sentence saying which is the case would help.

**One correctness question on the removals**

Pattern 4 strips ORC statements inline from tests that mix formats, including removing `orcFileRecords` and part of an expected-result assertion from `TestStructuredStreamingRead3`. That is the one category where a mistake changes what the remaining test asserts rather than just skipping something. Did you diff the surviving assertions to confirm the Parquet and Avro expectations are unchanged? A note confirming that would be reassuring, given how many files pattern 4 touches.

hsiang-c:
Close this one as we don't see significant time saved by disabling ORC in parameterized tests.

Issue #5002, opened 2026-07-23
perf: optimize native shuffle writer (redundant copies, per-block allocations)


andygrove:
### What is the problem the feature request solves?

A review of the native shuffle writer (`native/shuffle/`) identified several performance optimization opportunities. The partitioning and interleave logic is in good shape (reused scratch buffers, counting-sort partition assignment, `interleave_record_batch`, dedup'd buffer accounting), but the serialization/write side has redundant copies, per-block allocations, and avoidable per-partition churn.

The crate already has benchmarks (`native/shuffle/benches/shuffle_writer.rs` and `native/shuffle/src/bin/shuffle_bench.rs`), so each item below can be validated in isolation before running Spark-level benchmarks.

### High impact

- [ ] **Every batch is copied an extra time through `BatchCoalescer`, even when already full.** `BufBatchWriter::write` (`writers/buf_batch_writer.rs`) pushes every batch through a `BatchCoalescer`. In arrow 58.3, the normal path always does `copy_rows` into in-progress builders, even when the pushed batch is exactly `target_batch_size` and the buffer is empty (the zero-copy bypass only activates when `with_biggest_coalesce_batch_size` is explicitly set, which we never do). `PartitionedBatchIterator` already emits exactly `batch_size`-row batches from `interleave_record_batch` (except the tail), so in the multi-partition spill and finish paths essentially all shuffle data is copied twice: once by interleave, once by the coalescer. Fix: bypass the coalescer when its buffer is empty and `batch.num_rows() >= batch_size`, writing the batch directly (or configure `biggest_coalesce_batch_size`). This removes one full copy of the entire shuffle payload with bit-identical output.

- [ ] **The single-partition path coalesces twice.** `SinglePartitionShufflePartitioner` (`partitioners/single_partition.rs`) buffers small batches and `concat_batches` them to `batch_size`, then hands the result to the long-lived `BufBatchWriter`, whose own `BatchCoalescer` copies all the rows again. The concat layer appears to predate the coalescer in the writer and is now redundant. Dropping it removes another full copy in single-partition shuffles.

[...]

PR #5003, opened 2026-07-23
perf: bypass shuffle BatchCoalescer for already-sized batches

andygrove:

Addresses the "Every batch is copied an extra time through `BatchCoalescer`, even when already full" item in #5002. That issue is an umbrella tracking several shuffle-writer optimizations, so this PR does not close it.

## Rationale for this change

`BufBatchWriter::write` pushes every batch through a `BatchCoalescer`. In arrow 58.3 the normal push path always does `copy_rows` into the in-progress builders, even when the pushed batch is exactly `target_batch_size` and the buffer is empty. The zero-copy bypass in `BatchCoalescer` only activates when `biggest_coalesce_batch_size` is explicitly set, which Comet never does.

`PartitionedBatchIterator` already emits exactly `batch_size`-row batches from `interleave_record_batch` (except the tail), so in the multi-partition spill and finish paths, and the single-partition path, essentially all shuffle data was copied twice: once by interleave, once by the coalescer.

`BufBatchWriter` now configures its `BatchCoalescer` with `biggest_coalesce_batch_size = batch_size - 1`, enabling Arrow's zero-copy passthrough for batches that are already at least `batch_size` rows. Those batches are serialized verbatim instead of being copied into the in-progress builders. Smaller (tail) batches still flow through the coalescer as before, and the passthrough does not fire while the coalescer holds buffered rows, so output ordering is preserved.

For the multi-partition spill and finish paths, `PartitionedBatchIterator` never emits batches larger than `batch_size`, so the passthrough fires only for exactly-`batch_size` batches and block boundaries are unchanged. The single-partition path routes `>= batch_size` batches to the writer directly, so an oversized batch there is now written as one oversized block instead of being split at `batch_size`. Rows are written in the same order either way. This removes one full copy of the shuffle payload.

An end-to-end microbenchmark (`shuffle_writer` bench, 8192-row batches, 16 partitions, Lz4) shows roughly a 3% improvement, with a larger relative win expected for the uncompressed path where the copy is a bigger fraction of total work.

Existing `datafusion-comet-shuffle` unit tests continue to pass. A new `test_full_batches_bypass_coalescer` test feeds a sequence of full, oversized, and partial batches and asserts the resulting block boundaries. The oversized (150-row) batch is what distinguishes the passthrough: the coalescer alone would split it into a 100-row block plus 50 buffered rows, so it can only appear as a single 150-row block when the passthrough fires. The test also asserts that a partial batch is still coalesced with the following full batch, and that all rows roundtrip in order.

### DIFF
--- a/native/shuffle/src/shuffle_writer.rs
+++ b/native/shuffle/src/shuffle_writer.rs
@@ -924,6 +924,80 @@ mod test {
             .sum()
     }
 
+    /// Batches that are already at least `batch_size` rows should bypass the
+    /// coalescer, being written one-block-per-batch, while smaller batches are
+    /// still coalesced. In both cases the data must roundtrip exactly, in order.
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn test_full_batches_bypass_coalescer() {
+        use crate::writers::BufBatchWriter;
+        use arrow::array::Int32Array;
+
+        let schema = Arc::new(Schema::new(vec![Field::new("v", DataType::Int32, false)]));
+        let batch_size = 100;
+
+        let make_batch = |start: i32, len: i32| {
+            let values: Vec<i32> = (start..start + len).collect();
+            RecordBatch::try_new(
+                Arc::clone(&schema),
+                vec![Arc::new(Int32Array::from(values)) as Arc<dyn Array>],
+            )
+            .unwrap()
+        };
+
+        // Sequence: two full batches (bypass), a partial batch (buffered), then a
+        // full batch (must not bypass because the coalescer still holds buffered rows).
+        let inputs = vec![
+            make_batch(0, batch_size),          // rows 0..100   -> bypass
+            make_batch(batch_size, batch_size), // rows 100..200 -> bypass
+            make_batch(200, 30),                // rows 200..230 -> buffered
+            make_batch(230, batch_size),        // rows 230..330 -> coalesced with buffered 30
+        ];
+
+        let codec = CompressionCodec::Lz4Frame;
+        let encode_time = Time::default();
+        let write_time = Time::default();
+
+        let mut output = Vec::new();
+        {
+            let mut writer = ShuffleBlockWriter::try_new(schema.as_ref(), codec.clone()).unwrap();
+            let mut buf_writer = BufBatchWriter::new(
+                &mut writer,
+                Cursor::new(&mut output),
+                1024 * 1024,
+                batch_size as usize,
+            );
+            for batch in &inputs {
+                buf_writer.write(batch, &encode_time, &write_time).unwrap();
+            }
+            buf_writer.flush(&encode_time, &write_time).unwrap();
+        }
+
+        let blocks = read_all_ipc_batches(&output);
+
+        // The two full batches are emitted verbatim as their own blocks, then the
+        // partial+full pair coalesces into a full block plus a 30-row tail block.
+        let block_rows: Vec<usize> = blocks.iter().map(|b| b.num_rows()).collect();
+        assert_eq!(
+            block_rows,
+            vec![100, 100, 100, 30],
+            "unexpected block boundaries"
+        );
+
+        // All 330 rows must roundtrip in order (0..330).
+        let mut roundtripped = Vec::new();
+        for block in &blocks {
+            let col = block
+                .column(0)
+                .as_any()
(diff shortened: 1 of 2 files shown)


mbutrovich:
Doesn't `BatchCoalescer` already have an optimization for this? Are we just not using it? See `biggest_coalesce_batch_size`. Elsewhere in DataFusion we usually initialize it to `target_batch_size` / 2.

andygrove:
> Doesn't `BatchCoalescer` already have an optimization for this? Are we just not using it? See `biggest_coalesce_batch_size`. Elsewhere in DataFusion we usually initialize it to `target_batch_size` / 2.

Thanks, updated to use that config. Much simpler.

comphead (approved):

native/shuffle/src/shuffle_writer.rs
```diff
@@ -924,6 +924,80 @@ mod test {
...
+        }
+
+        let blocks = read_all_ipc_batches(&output);
+
+        // The two full batches are emitted verbatim as their own blocks, then the
+        // partial+full pair coalesces into a full block plus a 30-row tail block.
+        let block_rows: Vec<usize> = blocks.iter().map(|b| b.num_rows()).collect();
+        assert_eq!(
+            block_rows,
+            vec![100, 100, 100, 30],
```
0lai0: Could we add a case with a batch larger than `batch_size`, like 150 rows with `batch_size = 100` ?
I think `vec![100, 100, 100, 30]` is what the coalescer alone would produce, so this test would still pass without the fast path. 
But I'm not really sure about this. Thanks @andygrove

0lai0 (commented):
Just one question on the test, otherwise this looks nice

native/shuffle/src/shuffle_writer.rs
andygrove: Thanks for the review @0lai0. I added a test based on your feedback.

native/shuffle/src/writers/buf_batch_writer.rs
```diff
@@ -28,7 +28,9 @@ use std::io::{Cursor, Seek, SeekFrom, Write};
 ///
 /// Small batches are coalesced using Arrow's [`BatchCoalescer`] before serialization,
 /// producing exactly `batch_size`-row output batches to reduce per-block IPC schema overhead.
```
mbutrovich: "producing exactly `batch_size`-row output batches" is not accurate once a `>= batch_size` batch is bypassed verbatim, which the single-partition path already does today. Could soften to "at least `batch_size`" and mention the passthrough.

mbutrovich:
> output bytes are identical to before

True for multi-partition and spill. For single-partition, a `>= batch_size` batch on an empty coalescer buffer now yields one oversized block instead of `batch_size`-row blocks, because `single_partition.rs:107-108` already routes such batches straight to the coalescer on `main`. Worth qualifying the claim so it does not read as unconditional. Rows still roundtrip in order, so this is a block-boundary and buffer-size change, not a correctness issue.

mbutrovich (commented):
Minor discussion points, thanks @andygrove!

native/shuffle/src/writers/buf_batch_writer.rs
andygrove: You are right, the passthrough emits an oversized batch as its own block, so "exactly" was wrong. Reworded in fb86cd613: the doc comment now says output batches hold at least `batch_size` rows apart from the flush remainder, and that an oversized input batch is written as a single oversized block. The inline comment inside `write` had the same problem with its "bit-identical output" claim and has been corrected too.

andygrove:
> True for multi-partition and spill. For single-partition, a `>= batch_size` batch on an empty coalescer buffer now yields one oversized block instead of `batch_size`-row blocks

Confirmed, `single_partition.rs` routes `num_rows >= batch_size` straight to the writer, so those batches now pass through as a single block. I have updated the description to scope the unchanged-boundaries claim to the `PartitionedBatchIterator` paths and to spell out the single-partition behavior. The description also still described the earlier hand-rolled bypass rather than `biggest_coalesce_batch_size`, so that section is refreshed as well.

The code comments in `buf_batch_writer.rs` carried the same overstated claim and are fixed in fb86cd613.

mbutrovich (approved):
Approved pending CI. Thanks for looking at this, @andygrove!

PR #5004, opened 2026-07-23
refactor: drop redundant concat layer in single-partition shuffle

andygrove:

Addresses the "The single-partition path coalesces twice" item in #5002. That issue is an umbrella tracking several shuffle-writer optimizations, so this PR does not close it.

## Rationale for this change

`SinglePartitionShufflePartitioner` buffered small batches and `concat_batches`'d them to `batch_size` before handing the result to the long-lived `BufBatchWriter`, whose own `BatchCoalescer` then batches the rows into IPC blocks. That concat layer predates the coalescer and duplicates its job: the coalescer already batches the row stream to `batch_size`, so the partitioner was maintaining a second, redundant buffering mechanism in front of it.

The primary motivation is removing that redundant layer. There is also a performance win, but it is narrower than "one less copy of every row" — see below.

`SinglePartitionShufflePartitioner` now streams every incoming batch straight to the `PartitionWriter` and lets the writer's `BatchCoalescer` do the batching. The internal `buffered_batches` / `num_buffered_rows` state, the `add_buffered_batch` / `concat_buffered_batches` helpers, and the now-unused `batch_size` field are removed, and `try_new` becomes an infallible `new`. This also brings single-partition in line with the multi-partition path, which already relies on the writer's coalescer.

Since #5003 (merged), the coalescer passes a batch that is already `>= batch_size` straight through when it lands on an empty buffer, writing it verbatim as a single block that may exceed `batch_size`. So block boundaries in single-partition shuffle depend on how the input is chunked: sub-`batch_size` batches are coalesced to `batch_size`, and an oversized batch on an empty buffer is emitted as one oversized block. Either way every row is written exactly once, in order.

### How much does this actually save?

Whether a copy is eliminated depends on how the input batch size relates to `batch_size`, because that decides whether the old concat output qualified for the #5003 passthrough:

- **Input batch size does not divide `batch_size` evenly** (e.g. 3000-row batches with `batch_size` 8192): the old flush condition emitted a concatenated batch of 6000 rows, which is `< batch_size`, so it missed the passthrough and the coalescer copied those rows again. Old = two copies, new = one. **This is the case that improves.**
- **Input batch size divides `batch_size` evenly** (e.g. 8192, 1024 or 256 rows): the old flush condition tripped at *exactly* `batch_size` buffered rows, so the concat output did take the passthrough. Both old and new do exactly one copy, and this PR removes nothing. Note this includes the exactly-`batch_size` case that a native scan typically produces.

Measured with a throwaway in-process A/B harness running both implementations interleaved, min-of-60, total rows held constant, aggregated over four runs (positive = this PR is faster):

| rows/batch | vs `main` | notes |
| --- | --- | --- |
| 8192 | ~-1% | control: code path is identical in both, so this is the noise floor (spread ±6%) |
| 1024 | ~+2% | divides evenly; within noise |
| 256 | ~-4% | divides evenly; mild regression from one `push_batch` per input batch |
| 1000 | ~+10% | does not divide evenly |
| 3000 | ~+13% | does not divide evenly; consistently positive across all runs |

These were taken on a loaded machine, so treat the magnitudes as indicative rather than precise; the direction and the split between the two groups were stable across runs. The mild regression at 256 rows/batch looks real rather than noise: 1280 `push_batch` calls replace cheap `Vec` buffering plus 40 `concat_batches` calls.

Net: a solid win for input that is not `batch_size`-aligned (common downstream of filters, joins and limits, and for JVM-produced batches), a wash when it is aligned, and a simpler partitioner in all cases.

Existing `datafusion-comet-shuffle` unit tests, including `test_single_partition_shuffle_writer`, continue to pass. The `single_partition_passthrough_oversized_and_coalesces_small` test drives the partitioner directly, leading with a larger-than-`batch_size` batch on an empty buffer so it exercises the coalescer passthrough rather than leaving the outcome to buffer state. It reads the output data file back and asserts the blocks are `[250, 100, 30]` (an oversized passthrough block, then a coalesced `batch_size` block, then the tail) and that every row roundtrips in order.

The `shuffle_writer` bench previously only covered `Hash(16)` and `RangePartitioning(16)`, so the path this PR changes had no benchmark coverage at all. This PR adds two `SinglePartition` cases that differ only in input chunking (`rows_per_batch` of 8192 and 3000), covering the passthrough and copy paths described above.


native/shuffle/src/partitioners/single_partition.rs
```diff
@@ -19,66 +19,27 @@ use crate::metrics::ShufflePartitionerMetrics;
 use crate::partitioners::ShufflePartitioner;
 use crate::writers::PartitionWriter;
 use arrow::array::RecordBatch;
-use datafusion::common::DataFusionError;
 use std::iter;
 use tokio::time::Instant;
 
-/// A partitioner that writes all shuffle data to a single file and a single index file
+/// A partitioner that writes all shuffle data to a single file and a single index file.
```
mbutrovich, line 25: The "independent of input chunking" in the PR description reasoning holds only while `biggest_coalesce_batch_size` is unset. Given #5003 sets it, would it be worth dropping the chunk-independence phrasing so this comment and #5003 do not disagree about what the coalescer guarantees?

native/shuffle/src/shuffle_writer.rs
```diff
@@ -924,6 +923,86 @@ mod test {
             .sum()
     }
 
+    /// The single-partition partitioner streams batches straight to the writer's
+    /// `BatchCoalescer` without any concat/buffer layer of its own. Feeding batches
+    /// of varying sizes must still produce `batch_size`-row IPC blocks (plus a tail)
+    /// with every row roundtripping in order.
+    #[tokio::test]
+    #[cfg_attr(miri, ignore)]
+    async fn single_partition_coalesces_to_batch_size_in_order() {
```
mbutrovich: Asserts `[100, 100, 100, 50]`. I traced it under #5003's bypass too and it still passes, because the 250-row batch arrives on a 60-row buffer and takes the "small buffer, coalesce normally" branch (`coalesce.rs:471-475`) rather than the passthrough. So neither PR's test exercises the divergent case, a `>= batch_size` batch on an empty buffer. A test with the large batch first would pin down the intended post-#5003 behavior instead of leaving it to buffer state.

native/shuffle/src/partitioners/single_partition.rs
```diff
@@ -19,66 +19,27 @@ use crate::metrics::ShufflePartitionerMetrics;
...
-    /// Number of rows in the concatenating batches
-    num_buffered_rows: usize,
     /// Metrics for the repartitioner
     metrics: ShufflePartitionerMetrics,
-    /// The configured batch size
-    batch_size: usize,
 }
 
 impl<T: PartitionWriter> SinglePartitionShufflePartitioner<T> {
-    pub(crate) fn try_new(
```
mbutrovich, line 40: Good cleanup, `new` is infallible now that concat is gone. Quick check that `DataFusionError` is unused elsewhere in the file, though lint will catch it.

mbutrovich (commented):
After #5003, single-partition shuffle can emit an IPC block larger than `batch_size` when a `>= batch_size` batch lands on an empty coalescer buffer. Is that the intended tradeoff (one fewer copy, variable block sizes, larger read-side batch and compression buffer), or should single-partition still produce uniform `batch_size` blocks? If intended, #5003's description should drop the unconditional "identical bytes" wording and #5004's comment should drop "independent of input chunking." If not, #5003's bypass needs a guard, and note this is independent of #5004 since `main` already routes oversized batches to the coalescer.

native/shuffle/src/partitioners/single_partition.rs
andygrove: Reworded the comment. It now says a `>= batch_size` batch on an empty buffer is passed through verbatim as one (possibly oversized) block and that block boundaries depend on how the input is chunked, so it no longer disagrees with #5003. Dropped the same chunk-independence phrasing from the PR description.

native/shuffle/src/shuffle_writer.rs
andygrove: Reworked the test to lead with a 250-row batch on an empty buffer, so it now takes the passthrough branch. It asserts blocks `[250, 100, 30]`: the 250-row block can only appear if the bypass fires (plain coalescing would split it into 100/100/50), which pins down the post-#5003 behavior instead of leaving it to buffer state.

native/shuffle/src/partitioners/single_partition.rs
andygrove: Confirmed. The `use datafusion::common::DataFusionError` import was removed along with `concat_buffered_batches`, so nothing else in the file references it, and clippy is clean.

andygrove:
On the design question: this is the intended tradeoff, and as you noted it is a property of #5003 rather than this PR (`main` already routes oversized batches through the coalescer). A single-partition shuffle feeds one reducer, so passing an already-materialized `>= batch_size` batch through as one block just means a single larger IPC message instead of re-splitting it, which is exactly what #5003's bypass is for. Rather than guard it, I rebased this PR onto #5003 and made it agree with that behavior:

- The doc comment and `native_shuffle.md` drop the uniform-`batch_size` / chunk-independence wording and now describe the passthrough and the resulting variable block sizes.
- The single-partition test leads with an oversized batch on an empty buffer so it exercises the passthrough and asserts `[250, 100, 30]`.

I left #5003's merged description as-is. Happy to open a quick follow-up to drop its "identical bytes" wording if you think it is worth it.

mbutrovich (approved):

Issue #4655, opened 2026-06-15
pref: Use builtin compression for arrow ipc writer


wForget:
### What is the problem the feature request solves?

I benchmarked several shuffle IPC write strategies and compared five different approaches. The results show that Arrow IPC's built-in compression provides both better performance and a higher compression ratio than the other tested approaches.

The evaluated strategies are:

1. **`outer-zstd-stream`** — a single ZSTD stream wrapping the entire IPC stream.
2. **`ipc-builtin-zstd`** — Arrow IPC `StreamWriter` with built-in ZSTD compression.
3. **`per-batch-zstd-stream`** — each IPC batch written into its own ZSTD frame (**current comet shuffle style**).
4. **`per-batch-bulk-zstd`** — per-batch IPC buffer compressed with bulk ZSTD and written as length-prefixed frames.
5. **`per-batch-ipc-builtin-zstd`** — per-batch IPC stream using Arrow IPC's built-in ZSTD compression.

Based on the benchmark results, the built-in Arrow IPC compression (`ipc-builtin-zstd`) achieves the best overall balance of throughput and compression efficiency.

Run [ipc_write_bench](https://github.com/wForget/arrow-datafusion-comet/commit/07c2253871b816af56554f40f84999958fa7c311) benchmark:

```
cargo run --features shuffle-bench -p datafusion-comet-shuffle --bin ipc_write_bench -- \
  --input benchmark_data/customer.parquet \
  --output-dir /tmp/ipc_write_bench \
  --iterations 3 \
  --warmup 1
```

result:

```
=== Column Batch Write Benchmark ===
Input:       benchmark_data/customer.parquet
Batches:     210
Total rows:  1,500,000
Batch size:  8,192
Zstd level:  3
Iterations:  3 (warmup: 1)

--- outer-zstd-stream ---
  write avg:      15.853s
  write min/max:  15.702s / 16.099s
  write rows/s:   94,620
  read avg:       4.818s
  read min/max:   4.568s / 5.112s
  read rows/s:    311,306
  file size:      319.96 MiB
  verify rows:    1,500,000 (ok)

--- ipc-builtin-zstd ---
  write avg:      13.864s
  write min/max:  13.829s / 13.901s
  write rows/s:   108,195
  read avg:       4.755s
  read min/max:   4.730s / 4.804s
  read rows/s:    315,465
  file size:      318.94 MiB
  verify rows:    1,500,000 (ok)

--- per-batch-zstd-stream ---
[...]

PR #5005, opened 2026-07-23
perf: reuse zstd context and pre-encode IPC schema in shuffle blocks [experimental]

andygrove:
> **Experimental / not ready for review.** Opening as a draft to run larger-scale benchmarks before this is ready. Please do not review yet.

Addresses two of the "high impact" items in #5002: the per-block zstd compression-context allocation and the per-block IPC schema re-encoding. That issue is an umbrella tracking several shuffle-writer optimizations, so this PR does not close it.

## Rationale for this change

`ShuffleBlockWriter::write_batch` allocated a fresh zstd context (a `ZSTD_CCtx` plus its workspace) and re-serialized the schema flatbuffer (via `StreamWriter::try_new`) for every block. `read_ipc_compressed` likewise allocated a fresh `ZSTD_DCtx` per block. With small blocks (high partition counts), per-block context setup is a meaningful fraction of compression time.

- **Reuse the zstd compression/decompression context across blocks** via a per-thread context. Compression/decompression is streamed (not buffered whole), so large blocks are not copied an extra time. An earlier attempt using zstd's one-shot bulk API regressed large blocks by ~3% because it had to buffer the entire uncompressed stream; the streaming approach avoids that.
- **Pre-encode the IPC schema message once** in `try_new` and write it verbatim per block, instead of re-serializing it per block. Schemas that contain dictionary types keep using `StreamWriter`, whose dictionary-id bookkeeping ties schema and batch encoding together.

Output framing is unchanged: blocks remain standalone Arrow IPC streams in the same length-prefixed format, so the read side and the wire format are unaffected.

### Preliminary benchmark (local, zstd, 5M-row input; `shuffle_bench`)

Higher partition count means smaller blocks, where per-block context setup dominates.

| Partitions | ~rows/block | Encode time (main → this PR) |
| ---: | ---: | :--- |
| 2,000 | ~5,000 | 0.301s → 0.303s (flat) |
| 20,000 | ~500 | 0.492s → 0.449s (~9% faster) |
| 50,000 | ~200 | 0.752s → 0.669s (~11% faster) |

No regression at large block sizes; the win grows as blocks shrink. Larger-scale benchmarks are still to be run, which is why this is a draft.

Existing `datafusion-comet-shuffle` and `datafusion-comet` shuffle tests pass. New tests were added for a dictionary-typed column roundtrip (exercising the `StreamWriter` fallback and dictionary-id matching) and for reading multiple distinct zstd blocks on one thread (exercising cross-frame reuse of the thread-local compression/decompression contexts, which a dirty context would corrupt).

### DIFF
--- a/native/shuffle/src/ipc.rs
+++ b/native/shuffle/src/ipc.rs
@@ -19,34 +19,47 @@ use arrow::array::RecordBatch;
 use arrow::ipc::reader::StreamReader;
 use datafusion::common::DataFusionError;
 use datafusion::error::Result;
+use std::cell::RefCell;
+use std::io::Read;
 
 pub fn read_ipc_compressed(bytes: &[u8]) -> Result<RecordBatch> {
     match &bytes[0..4] {
-        b"SNAP" => {
-            let decoder = snap::read::FrameDecoder::new(&bytes[4..]);
-            let mut reader =
-                unsafe { StreamReader::try_new(decoder, None)?.with_skip_validation(true) };
-            reader.next().unwrap().map_err(|e| e.into())
-        }
-        b"LZ4_" => {
-            let decoder = lz4_flex::frame::FrameDecoder::new(&bytes[4..]);
-            let mut reader =
-                unsafe { StreamReader::try_new(decoder, None)?.with_skip_validation(true) };
-            reader.next().unwrap().map_err(|e| e.into())
-        }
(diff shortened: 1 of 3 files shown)


mbutrovich:
Does this incorporate the evaluation done in https://github.com/apache/datafusion-comet/issues/4655?

PR #5006, opened 2026-07-23
perf: encode shuffle IPC schema once per writer instead of per block

andygrove:

Addresses the "IPC schema is re-encoded for every block" item in #5002. That issue is an umbrella tracking several shuffle-writer optimizations, so this PR does not close it.

## Rationale for this change

`ShuffleBlockWriter::write_batch` created a `StreamWriter` for every block, and `StreamWriter::try_new` re-serializes the schema flatbuffer into the stream each time. Every block in a shuffle shares one schema, so this repeats identical work per block. It matters for wide or deeply nested schemas (larger schema flatbuffer) and for shuffles with many small blocks (high partition counts), where per-block fixed costs dominate.

- Pre-encode the schema IPC message once in `try_new` and write it verbatim at the start of every block; the record batch is written via `IpcDataGenerator::encode` plus the end-of-stream marker, instead of a per-block `StreamWriter`.
- Schemas that contain dictionary types keep using `StreamWriter`. Dictionary encoding requires the schema and the record batch to share a dictionary tracker (the batch's dictionary ids come from the schema encoding), so those two cannot be separated; a `contains_dictionary` check selects the fallback.

Output framing is unchanged: blocks remain standalone Arrow IPC streams in the same length-prefixed format, so the read side and wire format are unaffected. This is the schema-encoding half of #5002's compression items; zstd context reuse is a separate change.

### Benchmark

A new criterion benchmark, `shuffle_block_schema_encoding`, encodes an 8192-row batch with the `None` codec (to isolate schema + IPC encoding from compression), for a wide flat schema (50 `Int32` columns) and a deeply nested schema (6-deep nested structs). Local results, main vs this PR:

| Schema | main | this PR | |
| --- | --- | --- | --- |
| flat (50 cols) | ~80µs | ~75µs | ~7% faster |
| nested (6-deep structs) | ~42µs | ~32µs | ~24% faster |

The win scales with schema size relative to the batch's data: modest for a wide flat schema whose column data dominates, larger for a nested schema whose flatbuffer is expensive to re-encode. It also grows with block count (high partition counts), where the fixed per-block schema cost is paid more often.

Existing `datafusion-comet-shuffle` and `datafusion-comet` shuffle tests pass, including the roundtrip and dictionary-encoded shuffle-block tests. A new `roundtrip_ipc_dictionary` test covers a dictionary-typed column across all codecs, exercising the `StreamWriter` fallback and confirming the dictionary ids still resolve.

### DIFF
--- a/native/shuffle/src/writers/shuffle_block_writer.rs
+++ b/native/shuffle/src/writers/shuffle_block_writer.rs
@@ -47,11 +47,16 @@ pub enum CompressionCodec {
 /// How a writer encodes the Arrow IPC schema at the start of each block. The two arms are mutually
 /// exclusive by schema shape, decided once in [`ShuffleBlockWriter::try_new`], so the retained
 /// state never carries both a pre-encoded message and a schema at the same time.
+///
+/// Both arms hold their payload behind an `Arc` because a `ShuffleBlockWriter` is cloned once per
+/// output partition (see `LocalPartitionWriter`), and a shuffle can request millions of partitions.
+/// Deep-copying the pre-encoded schema per clone would allocate gigabytes for a large partition
+/// count; sharing it makes cloning O(1).
 #[derive(Clone)]
 enum SchemaEncoding {
     /// Dictionary-free schema: the IPC schema message, pre-encoded once, is written verbatim at the
     /// start of every block instead of being re-serialized per block.
-    Precoded(Vec<u8>),
+    Precoded(Arc<[u8]>),
     /// Schema containing dictionary types: the schema and record batch must share a dictionary
     /// tracker, so each block is encoded with `StreamWriter`, which re-serializes the schema.
     Fallback(SchemaRef),
@@ -68,7 +73,9 @@ enum SchemaEncoding {
 #[derive(Clone)]
 pub struct ShuffleBlockWriter {
     codec: CompressionCodec,
-    header_bytes: Vec<u8>,
+    /// Shared behind an `Arc` so cloning the writer per output partition stays O(1); see the note
+    /// on [`SchemaEncoding`].
+    header_bytes: Arc<[u8]>,
     /// IPC options shared by the schema and record-batch encoders so the two can never diverge;
     /// pinned to metadata version V5, which [`IPC_EOS`] depends on.
     write_options: IpcWriteOptions,
@@ -123,12 +130,12 @@ impl ShuffleBlockWriter {
             );
             let mut buf = Vec::new();
             write_message(&mut buf, encoded_schema, &write_options)?;
-            SchemaEncoding::Precoded(buf)
+            SchemaEncoding::Precoded(Arc::from(buf))
         };
 
         Ok(Self {
             codec,
-            header_bytes,
+            header_bytes: Arc::from(header_bytes),
             write_options,
             schema_encoding,
         })
@@ -228,5 +235,39 @@ impl ShuffleBlockWriter {
         timer.stop();
 
         Ok((end_pos - start_pos) as usize)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use arrow::datatypes::{DataType, Field};
+
+    /// A `ShuffleBlockWriter` is cloned once per output partition (see `LocalPartitionWriter`), and
+    /// a shuffle can request millions of partitions (e.g. the SPARK-48037 test uses more than 16
+    /// million). Cloning must share the immutable header and pre-encoded schema buffers rather than
+    /// deep-copying them; otherwise a large partition count allocates gigabytes and stalls the
+    /// executor.
+    #[test]
+    fn clone_shares_buffers() {
+        let schema = Schema::new(vec![
+            Field::new("a", DataType::Int64, false),
+            Field::new("b", DataType::Int64, false),
+        ]);
+        let writer = ShuffleBlockWriter::try_new(&schema, CompressionCodec::None).unwrap();
+        let cloned = writer.clone();
+
+        assert!(
+            Arc::ptr_eq(&writer.header_bytes, &cloned.header_bytes),
+            "header bytes should be shared across clones, not deep-copied"
+        );
+
+        match (&writer.schema_encoding, &cloned.schema_encoding) {
+            (SchemaEncoding::Precoded(a), SchemaEncoding::Precoded(b)) => assert!(
+                Arc::ptr_eq(a, b),
+                "pre-encoded schema should be shared across clones, not deep-copied"
+            ),
+            _ => panic!("dictionary-free schema should use the pre-encoded fast path"),
+        }
     }
 }
--- a/native/shuffle/src/writers/shuffle_block_writer.rs
+++ b/native/shuffle/src/writers/shuffle_block_writer.rs
@@ -16,12 +16,20 @@
 // under the License.
 
 use arrow::array::RecordBatch;
-use arrow::datatypes::Schema;
-use arrow::ipc::writer::StreamWriter;
+use arrow::datatypes::{DataType, Schema, SchemaRef};
+use arrow::ipc::writer::{
+    write_message, CompressionContext, DictionaryTracker, IpcDataGenerator, IpcWriteOptions,
+    StreamWriter,
+};
 use datafusion::common::DataFusionError;
 use datafusion::error::Result;
 use datafusion::physical_plan::metrics::Time;
-use std::io::{Cursor, Seek, SeekFrom, Write};
+use std::io::{Seek, SeekFrom, Write};
+use std::sync::Arc;
+
+/// Arrow IPC stream end-of-stream marker: the continuation marker (`0xFFFFFFFF`) followed by a
+/// zero message length, matching what `StreamWriter::finish` emits for metadata version V5.
+const IPC_EOS: [u8; 8] = [0xff, 0xff, 0xff, 0xff, 0x00, 0x00, 0x00, 0x00];
 
 /// Compression algorithm applied to shuffle IPC blocks.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -32,42 +40,119 @@ pub enum CompressionCodec {
...
 #[derive(Clone)]
 pub struct ShuffleBlockWriter {
     codec: CompressionCodec,
     header_bytes: Vec<u8>,
+    schema: SchemaRef,
+    /// Pre-encoded Arrow IPC schema message, written at the start of every block. Only used when
+    /// the schema has no dictionary types.
+    schema_message: Vec<u8>,
+    /// Whether the schema contains any dictionary types (see [`Self::encode_ipc_stream`]).
+    has_dictionaries: bool,
```
comphead:
```suggestion
/// Pre-encoded Arrow IPC schema message.
///
/// `None` indicates the schema contains dictionary types and therefore
/// cannot be reused across blocks.
schema_message: Option<Vec<u8>>,
```

maybe we can end up with a single field?

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -32,42 +40,119 @@ pub enum CompressionCodec {
...
+            // tracker, so `StreamWriter` (which re-encodes the schema per block) is used here.
+            let mut stream_writer = StreamWriter::try_new(out, &self.schema)?;
+            stream_writer.write(batch)?;
+            stream_writer.finish()?;
+            return Ok(());
+        }
+
+        // Fast path: reuse the pre-encoded schema message and write the record batch manually.
+        let options = IpcWriteOptions::default();
+        let data_gen = IpcDataGenerator::default();
```
comphead: would it make sense to cache and reuse these, instead of per batch allocation?

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -32,42 +40,119 @@ pub enum CompressionCodec {
...
+    schema_message: Vec<u8>,
+    /// Whether the schema contains any dictionary types (see [`Self::encode_ipc_stream`]).
+    has_dictionaries: bool,
 }
 
 impl ShuffleBlockWriter {
     pub fn try_new(schema: &Schema, codec: CompressionCodec) -> Result<Self> {
-        let header_bytes = Vec::with_capacity(20);
-        let mut cursor = Cursor::new(header_bytes);
+        let mut header_bytes = Vec::with_capacity(20);
```
comphead, line 89: we may want to comment why 20 is here

native/shuffle/src/writers/shuffle_block_writer.rs
andygrove: Good call to look at these. Checking each:

- `IpcDataGenerator {}` is a zero-sized struct, so constructing it is free.
- `DictionaryTracker::new(true)` is an empty `HashMap` + `Vec` with no heap allocation until first use, and it's never populated on this fast path. It also has to be fresh per block, since each block is a self-contained IPC stream.
- `CompressionContext` (an alias of `IpcWriteContext`) is the one that genuinely benefits from reuse: it holds a `FlatBufferBuilder` and, for zstd, a `Compressor`, and arrow's docs describe it as intended to be reused across writes.

Reusing the `CompressionContext` isn't free to wire in here, though: it's not `Clone` (the zstd `Compressor` isn't), while `ShuffleBlockWriter` is `Clone` and shared across partitions via `Borrow`, and `write_batch` takes `&self`. Reuse means threading a `&mut CompressionContext` down from the per-partition `BufBatchWriter` (the actual mutable owner) through `write_batch` and its call sites. That's worth doing, but it's an API change orthogonal to encoding the schema once, so I'd prefer to do it as a follow-up rather than grow this PR. Filed as #5008.

native/shuffle/src/writers/shuffle_block_writer.rs
andygrove: Done

native/shuffle/src/writers/shuffle_block_writer.rs
andygrove: Done

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -32,42 +40,128 @@ pub enum CompressionCodec {
     Snappy,
 }
 
+/// Returns true if `data_type` is, or nests, a dictionary type.
+fn contains_dictionary(data_type: &DataType) -> bool {
```
mbutrovich: arrow-rs already provides this. `Schema::flattened_fields()` walks the full nested field tree (via `Field::fields()` / `Field::_fields`), so the detection at 103-106 becomes:

```rust
let has_dictionaries = schema
    .flattened_fields()
    .iter()
    .any(|f| matches!(f.data_type(), DataType::Dictionary(_, _)));
```

and this function can be deleted. Beyond removing the hand-maintained match arms, it closes a gap: `contains_dictionary` omits `ListView` and `LargeListView`, but arrow's own dictionary traversal (`_encode_dictionaries`, arrow-ipc/src/writer.rs:438,450) recurses into them. A `ListView<Dictionary<..>>` column would be classified dictionary-free, take the fast path, and trip the `encoded_dictionaries.is_empty()` assert (or drop dictionary batches in release). Unreachable today since Comet emits no ListView, but keying off arrow's own field walk keeps this from drifting on a version bump, and it makes the `debug_assert` at line 157 sound rather than dependent on this recursion staying in sync.

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -32,42 +40,128 @@ pub enum CompressionCodec {
...
+/// batch message, end-of-stream marker). For the common case of a schema with no dictionary types,
+/// the schema flatbuffer is encoded once in [`Self::try_new`] and written verbatim at the start of
+/// every block, rather than being re-serialized per block as `StreamWriter::try_new` would do.
+/// Schemas that contain dictionary types fall back to `StreamWriter`, whose dictionary-id
+/// bookkeeping ties schema and batch encoding together.
 #[derive(Clone)]
 pub struct ShuffleBlockWriter {
     codec: CompressionCodec,
     header_bytes: Vec<u8>,
+    schema: SchemaRef,
```
mbutrovich: Following up on the single-field suggestion: `schema` and `schema_message` are mutually exclusive by use (fast path reads only `schema_message`, fallback reads only `schema`), so an enum states the invariant and drops the dead field:

```rust
enum SchemaEncoding {
    Precoded(Vec<u8>),   // dictionary-free: written verbatim per block
    Fallback(SchemaRef), // dictionary schemas: StreamWriter re-encodes
}
```

Today `try_new` runs `Arc::new(schema.clone())` at line 130 unconditionally, a deep clone of every field that is never read on the dictionary-free path. With the enum the schema is only retained in the `Fallback` arm, and as an `Arc::clone` rather than a deep clone.

native/shuffle/src/writers/shuffle_block_writer.rs
mbutrovich: Checked this against arrow-ipc 58.4: none of these need caching.

- `CompressionContext` is `#[derive(Default)]` over `Option<zstd::bulk::Compressor>` (compression.rs:30-34), so `::default()` is `None` and the zstd context is built lazily only when arrow's internal IPC compression runs. We pass `IpcWriteOptions::default()` with `batch_compression_type: None` (block compression is done by our own frame encoders), so arrow never constructs it. No allocation here. (It was an eager `ZSTD_createCCtx` in 58.1 and earlier, made lazy in 58.3.)
- `IpcDataGenerator` is zero-sized. `IpcWriteOptions::default()` is a few scalar assignments, no allocation.
- `DictionaryTracker` must stay fresh per block, since each block is a standalone stream and the tracker records which dict ids were already emitted.

So reuse would add interior mutability for no measurable gain. Worth revisiting after the arrow 59.x bump: there `CompressionContext` becomes `IpcWriteContext` with `reserve_scratch_with_capacity`, and reusing one instance lets the `arrow_data` staging buffer grow once instead of per block. That pairs with the zstd-context-reuse item still open in #5002.

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -32,42 +40,128 @@ pub enum CompressionCodec {
...
+            // Dictionary encoding requires the schema and record batch to share a dictionary
+            // tracker, so `StreamWriter` (which re-encodes the schema per block) is used here.
+            let mut stream_writer = StreamWriter::try_new(out, &self.schema)?;
+            stream_writer.write(batch)?;
+            stream_writer.finish()?;
+            return Ok(());
+        };
+
+        // Fast path: reuse the pre-encoded schema message and write the record batch manually.
+        let options = IpcWriteOptions::default();
```
mbutrovich: The schema is encoded with an `IpcWriteOptions` built in `try_new` (line 114) and the batch with a second one built here. They are identical today, but storing one `IpcWriteOptions` on the writer and referencing it in both places removes the chance of the schema and batch being encoded under divergent options if a default is ever changed.

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -16,12 +16,20 @@
...
 use datafusion::common::DataFusionError;
 use datafusion::error::Result;
 use datafusion::physical_plan::metrics::Time;
-use std::io::{Cursor, Seek, SeekFrom, Write};
+use std::io::{Seek, SeekFrom, Write};
+use std::sync::Arc;
+
+/// Arrow IPC stream end-of-stream marker: the continuation marker (`0xFFFFFFFF`) followed by a
+/// zero message length, matching what `StreamWriter::finish` emits for metadata version V5.
+const IPC_EOS: [u8; 8] = [0xff, 0xff, 0xff, 0xff, 0x00, 0x00, 0x00, 0x00];
```
mbutrovich, line 36: Shuffle blocks are always V5 (`IpcWriteOptions::default()`), so hardcoding the EOS is fine here. The comment already notes the V5 provenance; worth making the constraint explicit so a future reader knows it is an assumption, not a universal constant: these bytes are the continuation marker plus zero length only for `MetadataVersion::V5` with `write_legacy_ipc_format = false` (V4-legacy emits 4 zero bytes with no marker, per `write_continuation` in arrow-ipc/src/writer.rs:1893-1922). If Comet ever uses arrow IPC elsewhere and applies the same pre-encode trick under different options, this constant must be revisited. A one-line `debug_assert!(matches!(options.metadata_version, MetadataVersion::V5))` next to the write at line 161 pins the assumption at low cost.

mbutrovich (commented):

native/shuffle/src/writers/shuffle_block_writer.rs
andygrove: Done in ca61876. `contains_dictionary` is deleted; detection is now:

```rust
let has_dictionaries = schema
    .flattened_fields()
    .iter()
    .any(|f| matches!(f.data_type(), DataType::Dictionary(_, _)));
```

Confirmed `flattened_fields` is present in the arrow-schema we pin (58.3.0, `schema.rs:364`). This also closes the `ListView`/`LargeListView` gap you pointed at, so the `encoded_dictionaries.is_empty()` debug_assert no longer depends on our own recursion staying in sync with arrow's dictionary traversal across version bumps.

native/shuffle/src/writers/shuffle_block_writer.rs
andygrove: Done in ca61876. Both fields are replaced by:

```rust
enum SchemaEncoding {
    Precoded(Vec<u8>),   // dictionary-free: written verbatim per block
    Fallback(SchemaRef), // dictionary schemas: StreamWriter re-encodes
}
```

The `Arc::new(schema.clone())` now happens only in the `Fallback` arm, so the dictionary-free path no longer deep-clones a schema it never reads.

One correction to the detail: `try_new` takes `&Schema`, not a `SchemaRef`, so the fallback arm is still a deep clone rather than an `Arc::clone`. It is now paid only on the rare dictionary path, and once per writer rather than per block, so I left the signature alone.

native/shuffle/src/writers/shuffle_block_writer.rs
andygrove: Done in ca61876. A single `write_options: IpcWriteOptions` is stored on the writer and used by all three encode sites: the schema encode in `try_new`, the batch encode in `encode_ipc_stream`, and the dictionary fallback (switched from `StreamWriter::try_new` to `try_new_with_options`). There is no longer a second options value that could drift from the first.

native/shuffle/src/writers/shuffle_block_writer.rs
andygrove: Took the intent here but not the literal form, since the suggested assert does not compile: `IpcWriteOptions::metadata_version` is a private field (arrow-ipc 58.3 `writer.rs:65`) and there is no public getter on `IpcWriteOptions`, so it cannot be inspected from outside the crate.

Pinned V5 at the single construction site instead:

```rust
let write_options = IpcWriteOptions::try_new(64, false, MetadataVersion::V5)?;
```

`(64, false, V5)` is exactly what `IpcWriteOptions::default()` builds (`writer.rs:148-157`), so behavior is unchanged, but V5 becomes a deliberate choice rather than an inherited default, and the constraint holds in release builds rather than only under `debug_assertions`. Since that one value now feeds every encode site, the options that produce the stream and the hardcoded `IPC_EOS` cannot diverge.

Also expanded the `IPC_EOS` doc comment to state the V4-legacy contrast explicitly (four zero bytes, no continuation marker) and to point at the pin as the reason the constant is valid.

mbutrovich (approved):
Approved pending CI, thanks for the quick revision @andygrove!

wForget (approved):
Thanks @andygrove , LGTM

wForget:
The CI seems to be hanging, and it may be related to this PR.

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -32,42 +44,129 @@ pub enum CompressionCodec {
...
+            batch,
+            &mut dictionary_tracker,
+            &self.write_options,
+            &mut compression_context,
+        )?;
+        debug_assert!(encoded_dictionaries.is_empty());
+
+        out.write_all(schema_message)?;
+        write_message(&mut *out, encoded_batch, &self.write_options)?;
+        out.write_all(&IPC_EOS)?;
```
wForget, line 173: Should we call `out.flush()` here, since we need to get `ipc_length` later? 

https://github.com/andygrove/datafusion-comet/blob/ca61876bb3dbc0d0ff747f8b047f61a2fab76bbc/native/shuffle/src/writers/shuffle_block_writer.rs#L214-L215

I'm not sure if the CI hang is due to this.

andygrove: Good question, but flushing isn't needed here: for the compressed codecs, `write_batch` calls `.finish()`/`.into_inner()` on the encoder after `encode_ipc_stream` returns and before reading `stream_position()`, which fully flushes into `output`.

The hang turned out to be unrelated. `ShuffleBlockWriter` is cloned once per output partition, and this PR made that clone deep-copy the pre-encoded schema `Vec`. `SPARK-48037` uses 16M+ partitions, so that allocated GBs and stalled the executor (only the shard running `AdaptiveQueryExecSuite` hangs). Fixed in the last commit by sharing the buffers via `Arc`.

PR #5014, opened 2026-07-23
fix: raise CAST_INVALID_INPUT for invalid calendar dates in ANSI cast to date

### LINKED ISSUE #5012: ANSI mode: cast(string as date) returns NULL instead of throwing for invalid calendar dates
### Describe the bug

With `spark.sql.ansi.enabled=true`, `CAST(string AS date)` must raise `CAST_INVALID_INPUT` for a malformed input. Comet returns `NULL` instead when the string has the shape of a date but is not a real calendar date, e.g. `'2016-13-01'`, `'2016-02-30'`, `'2016-01-32'`.

Strings that fail to parse structurally (`'abc'`) do throw correctly, and `CAST(string AS timestamp)` throws correctly for the same inputs, so this is specific to the date target.

Root cause looks like the fast path in `date_parser` (`native/spark-expr/src/conversion_funcs/string.rs:1948-1962`): for a canonical 10-byte `yyyy-mm-dd` shape it returns `Ok(ymd_to_epoch_day(..).map(..))` directly, so an invalid calendar date yields `Ok(None)` → null, bypassing `return_result`, which is the only place the `EvalMode::Ansi` error is raised. The comment above that block asserts that "Comet already returns null for that in every eval mode ... matching what the general parser does here", which is what the repro contradicts.

This is the Tier 4 item from #4180 (does `spark.sql.ansi.enabled` propagate beyond the paths where it was originally wired up).

### Steps to reproduce

The expression has to run over a real column, otherwise Spark constant-folds it and Comet never executes.

```scala
val path = "/tmp/ansi_date"
spark.sql("SELECT s FROM VALUES ('2016-13-01'),('2016-02-30'),('2016-01-01') AS v(s)")
spark.sql(s"CREATE OR REPLACE TEMPORARY VIEW t AS SELECT * FROM parquet.`$path`")

spark.conf.set("spark.sql.ansi.enabled", "true")

spark.sql("SELECT cast(s as date) FROM t").collect()
// org.apache.spark.SparkDateTimeException: [CAST_INVALID_INPUT] The value '2016-13-01'
// of the type "STRING" cannot be cast to "DATE" because it is malformed.

spark.sql("SELECT cast(s as date) FROM t").show()
// +-------------+
// |CAST(s AS DATE)|
// +-------------+
// |         NULL|
// |         NULL|
// |   2016-01-01|
// +-------------+
```

`to_date(s)` goes through the same cast and behaves the same way.

The consequence is not limited to a missing exception. A filter on the cast silently returns rows where Spark fails the query:

```sql
SELECT count(*) FROM t WHERE cast(s as date) > date'2000-01-01'
-- spark: SparkDateTimeException
-- comet: 1
```

Behaviour that is already correct and should stay that way:

* `spark.sql.ansi.enabled=false` — both return `NULL` (verified, matches).
* `try_cast(s as date)` under ANSI — both return `NULL` (verified, matches).
* `cast(s as timestamp)` / `cast(s as timestamp_ntz)` under ANSI — both throw (verified).

### Expected behavior

Under `EvalMode::Ansi`, `cast(string as date)` should raise `CAST_INVALID_INPUT` for an invalid calendar date, matching the behaviour for structurally malformed strings and for the timestamp target.

### Additional context

* Reproduced on `apache/main` @ `0761e549a`, default Maven profile (Spark 4.1.2, where ANSI is on by default), macOS aarch64, debug build. Applies to Spark 3.x too whenever ANSI is enabled explicitly.
* An adjacent, much smaller discrepancy found in the same sweep: `cast(string as boolean)` under ANSI throws the right error class and message but as `SparkNumberFormatException` where Spark throws `SparkRuntimeException`. Not filed separately; mentioning here in case it is worth folding into the same fix.

andygrove:

Closes #5012.

## Rationale for this change

With `spark.sql.ansi.enabled=true`, `CAST(string AS date)` must raise `CAST_INVALID_INPUT` for malformed input. Comet returned `NULL` instead when the string parsed structurally but did not name a real calendar date, e.g. `'2016-13-01'`, `'2016-02-30'`, `'2016-01-32'`.

Beyond the missing exception, this silently changes results: a filter on the cast returns rows for a query that Spark fails outright.

```sql
SELECT count(*) FROM t WHERE cast(s as date) > date'2000-01-01'
-- spark: SparkDateTimeException
-- comet: 1
```

Spark's `SparkDateTimeUtils#stringToDate` returns `None` whenever `LocalDate.of` throws or the narrowing of the epoch day to an `Int` overflows (`localDateToDays` uses `Math.toIntExact`), and `stringToDateAnsi` turns every `None` into `CAST_INVALID_INPUT`. Comet's port had paths that produced `Ok(None)` without going through `return_result`, which is the only place the `EvalMode::Ansi` error is raised.

The underlying reason one of those paths was missed is structural. `date_parser` has a fast path for the canonical `yyyy-mm-dd` shape alongside the general byte-scanning parser, and each one independently decided what a parsed `(year, month, day)` triple *means*. The fast path implemented only the calendar-validity rule and elided the rest, with the argument for why that was safe living in a prose comment rather than in code.

In `date_parser` (`native/spark-expr/src/conversion_funcs/string.rs`):

* A new `finish` helper holds the whole decision for a parsed `(year, month, day)` triple: `ymd_to_epoch_day`, the `i32` narrowing, the chrono representability guard, and the `return_result` call that raises under ANSI. Both parsing paths now call it, so the fast path is decoding only and the semantics exist in exactly one place. Widening the fast path's trigger in future can no longer silently reintroduce this bug class.
* As part of that consolidation, the previous `-262143..=262142` year guard is replaced by an explicit `i32::try_from` check on the computed epoch day, so years whose epoch day overflows an `Int` fail exactly where Spark fails. The chrono representability limit is applied afterwards.

Years that Spark accepts but chrono cannot represent (roughly `262143` to `5881580` and their negative counterparts) still return `NULL` in every eval mode. Raising for those would diverge from Spark in the other direction, and Comet's downstream date kernels cannot handle those values, so that limitation is preserved and now documented in the code.

Behaviour that was already correct is unchanged: non-ANSI mode returns `NULL`, `try_cast` returns `NULL` under ANSI, and structurally malformed strings continue to throw.

The `cast(string as boolean)` exception-type discrepancy also mentioned in #5012 is not addressed here and remains open.

* `CometCastSuite` "cast StringType to DateType" gains ten invalid calendar dates in its `invalidDates` list. That test compares against Spark in both non-ANSI mode (values) and ANSI mode (exception parity), so it owns per-string coverage in both directions. The existing fuzz generator uses a `0123456789/` plus whitespace alphabet and so never produced a hyphenated date shape, which is why this went unnoticed.
* `date_parser_test` in `native/spark-expr/src/conversion_funcs/string.rs` asserts an error under `EvalMode::Ansi` for invalid calendar dates in both the fast-path and general-parser shapes, and for years whose epoch day overflows `i32`. The chrono-limit cases keep asserting `Ok(None)` in all three eval modes. The repeated per-eval-mode loops are collapsed into two helpers that report the offending input on failure.
* New SQL file test `spark/src/test/resources/sql-tests/expressions/cast/cast_string_to_date_ansi.sql` covering the plan shapes `CometCastSuite` does not exercise: the cast over a column projection, the cast inside a filter predicate from the issue, and `to_date`. It includes non-error sentinel queries over valid input so a silent fallback to Spark cannot let the `expect_error` cases pass vacuously. Verified that this fixture fails against the unpatched native library and passes with the fix.
* Full `CometCastSuite` (162 passed), full `CometSqlFileTestSuite` (419 passed), `CometTemporalExpressionSuite` (32 passed), and the full `datafusion-comet-spark-expr` Rust test suite (511 passed) are green, with clippy clean.
* The refactor is not a hot-path regression: benchmarking `native/spark-expr/benches/cast_string_to_date.rs` against the pre-change parser shows the new code marginally faster (canonical input 54.8 vs 56.6 us, mixed input 86.6 vs 87.4 us).


native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1930,6 +1929,31 @@ fn date_parser(date_str: &str, eval_mode: EvalMode) -> SparkResult<Option<i32>>
...
+        // Spark builds a `LocalDate` and narrows its epoch day to an `Int`, so an invalid
+        // calendar date or an epoch day that overflows `i32` both yield `None` there, which
+        // `stringToDateAnsi` turns into CAST_INVALID_INPUT.
+        let Some(days) = ymd_to_epoch_day(year, month, day).and_then(|d| i32::try_from(d).ok())
+        else {
+            return return_result(date_str, eval_mode);
+        };
+        // Spark accepts years beyond what chrono can represent, and downstream date kernels
+        // cannot handle those values, so Comet keeps returning null for them in every eval mode
+        // rather than raising. This is a Comet limitation, not a malformed input.
```
peterxcli, line 1951: It would be good to add the coresponding chrono issue or its code snippet here.

peterxcli (approved):
Great work, LGTM!

peterxcli:
failed ci looks unrelated
https://github.com/apache/datafusion-comet/actions/runs/30050199441/job/89355580948?pr=5014
```
  #
  # A fatal error has been detected by the Java Runtime Environment:
  #
  #  SIGSEGV (0xb) at pc=0x0000000000000000, pid=22862, tid=59915
  #
  # JRE version: OpenJDK Runtime Environment Zulu17.68+17-CA (17.0.20+8) (build 17.0.20+8-LTS)
  # Java VM: OpenJDK 64-Bit Server VM Zulu17.68+17-CA (17.0.20+8-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
  # Problematic frame:
  # C  0x0000000000000000
  #
  # No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
  #
  # An error report file with more information is saved as:
  # /Users/runner/work/datafusion-comet/datafusion-comet/spark/hs_err_pid22862.log
  #
  # If you would like to submit a bug report, please visit:
  #   http://www.azul.com/support/
  #
```

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1930,6 +1929,31 @@ fn date_parser(date_str: &str, eval_mode: EvalMode) -> SparkResult<Option<i32>>
             Ok(None)
         }
     }
+
+    /// Turns parsed year/month/day segments into an epoch day. Shared by both parsing paths so
+    /// that the decision of what a segment triple *means* lives in exactly one place.
+    fn finish(
```
comphead: maybe we can rename `finish` ?

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1930,6 +1929,31 @@ fn date_parser(date_str: &str, eval_mode: EvalMode) -> SparkResult<Option<i32>>
...
+        year: i64,
```
comphead, line 1936: `chrono` usually uses `u32`
https://docs.rs/chrono/latest/chrono/trait.Datelike.html#tymethod.with_day

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -2798,6 +2809,25 @@ mod tests {
         );
     }
 
+    /// Malformed input: null in legacy and try mode, CAST_INVALID_INPUT in ANSI mode.
```
comphead:
```
fn assert_dates(
    dates: &[&str],
    modes: &[EvalMode],
    expect_ansi_error: bool,
) {
    for &date in dates {
        for &mode in modes {
            if expect_ansi_error && mode == EvalMode::Ansi {
                assert!(date_parser(date, mode).is_err(), "{date}");
            } else {
                assert_eq!(date_parser(date, mode).unwrap(), None, "{date}");
            }
        }
    }
}

/// Malformed input: null in legacy and try mode, CAST_INVALID_INPUT in ANSI mode.
fn assert_null_or_ansi_error(dates: &[&str]) {
    assert_dates(
        dates,
        &[EvalMode::Legacy, EvalMode::Try, EvalMode::Ansi],
        true,
    );
}

/// Input Spark parses successfully but Comet cannot represent: null in every eval mode.
fn assert_null_in_all_modes(dates: &[&str]) {
    assert_dates(
        dates,
        &[EvalMode::Legacy, EvalMode::Try, EvalMode::Ansi],
        false,
    );
}
```

to remove some duplication

comphead (approved):
Thanks @andygrove some non blocking comments

mbutrovich:
> https://github.com/apache/datafusion-comet/actions/runs/30050199441/job/89355580948?pr=5014



> failed ci looks unrelated https://github.com/apache/datafusion-comet/actions/runs/30050199441/job/89355580948?pr=5014
> 
> ```
>   #
>   # A fatal error has been detected by the Java Runtime Environment:
>   #
>   #  SIGSEGV (0xb) at pc=0x0000000000000000, pid=22862, tid=59915
>   #
>   # JRE version: OpenJDK Runtime Environment Zulu17.68+17-CA (17.0.20+8) (build 17.0.20+8-LTS)
>   # Java VM: OpenJDK 64-Bit Server VM Zulu17.68+17-CA (17.0.20+8-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
>   # Problematic frame:
>   # C  0x0000000000000000
>   #
>   # No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
>   #
>   # An error report file with more information is saved as:
>   # /Users/runner/work/datafusion-comet/datafusion-comet/spark/hs_err_pid22862.log
>   #
>   # If you would like to submit a bug report, please visit:
>   #   http://www.azul.com/support/
>   #
> ```

I've seen this a few times lately. I will file a tracking issue.

andygrove:
@comphead I went ahead and addressed some of the feedback

PR #5020, opened 2026-07-24
fix: apply 1.11 Iceberg diff changes from #4991 to other versions

mbutrovich:

Closes #.

## Rationale for this change

`main` builds fail the Iceberg builds for 1.8, 1.9, and 1.10 because I neglected to apply the same changes I made to 1.11 in #4991.

Apple the 1.11 Iceberg diff changes from #4991 to 1.8, 1.9, and 1.10.

Existing CI. I'll stick the `run-iceberg-tests` label on here so we should run all versions like `main` builds do. I should have done that on #4991.

### DIFF
--- a/dev/diffs/iceberg/1.10.0.diff
+++ b/dev/diffs/iceberg/1.10.0.diff
@@ -2513,6 +2513,31 @@ index daf4e29ac0..0c7d376f80 100644
              .enableHiveSupport()
              .getOrCreate();
  
+diff --git a/spark/v3.5/spark/src/test/java/org/apache/iceberg/spark/data/AvroDataTestBase.java b/spark/v3.5/spark/src/test/java/org/apache/iceberg/spark/data/AvroDataTestBase.java
+index 0db6a65fd3..9ee8b9e084 100644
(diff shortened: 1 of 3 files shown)


dev/diffs/iceberg/1.10.0.diff
```diff
@@ -2513,6 +2513,31 @@ index daf4e29ac0..0c7d376f80 100644
...
+-import org.assertj.core.api.Condition;
+ import org.junit.jupiter.api.Test;
+ import org.junit.jupiter.api.io.TempDir;
+ import org.junit.jupiter.params.ParameterizedTest;
+@@ -307,12 +306,6 @@ public abstract class AvroDataTestBase {
+                 .build());
+ 
+     assertThatThrownBy(() -> writeAndValidate(writeSchema, expectedSchema))
+-        .has(
+-            new Condition<>(
```
mbutrovich, line 2533: We don't throw this exact class, but the check for the error message is still valid.

andygrove (approved):

mbutrovich:
Thanks for the quick review, @andygrove! I'll get this in once CI goes green to fix the `main` builds.

PR #5025, opened 2026-07-24
refactor: centralize data type support predicates

### LINKED ISSUE #5021: Centralize data-type support predicates so new types don't have to be added in multiple places
## Background

While reviewing #4976 (interval codegen dispatch for nested values and native shuffle), it became clear that Comet now has several independent data-type support predicates that each had to learn about `YearMonthIntervalType` and `DayTimeIntervalType` separately:

- `supportedSinkDataType` in `spark/src/main/scala/org/apache/comet/serde/operator/CometSink.scala`
- `supportedSerializableDataType` in `spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometShuffleExchangeExec.scala`
- `isTypeSupported` in `spark/src/main/scala/org/apache/spark/sql/comet/CometLocalTableScanExec.scala`
- `supportedDataType` in `spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala`

In #4976, `supportedSinkDataType` is effectively a copy of `supportedDataType(_, allowComplex = true)` with intervals added at each nesting level. That was the least invasive option for that PR, but it means these predicates can drift out of sync as new types are added.

## Proposal

Consider consolidating so that adding support for a new data type only requires touching one place. Options to explore:

- Give `supportedDataType` an additional flag (for example `allowIntervals`) rather than forking the traversal in `CometSink`.
- Or converge the sink / shuffle / local-scan predicates onto a shared helper with explicit, named capability flags for the differences that genuinely need to vary per boundary (for example, interval file scans staying unsupported while LocalTableScan supports them).

## Notes

This is a maintainability cleanup, not a correctness bug. The predicates are correct today. Filed as follow-up from the #4976 review.

peterxcli:

Closes #5021.

Depends on #4976 (merged).

## Rationale for this change

Data-type support checks are duplicated across sinks, local scans, and shuffle paths, making new type support error-prone.

Centralize recursive capability checks in `QueryPlanSerde.supportedDataType` and reuse them across existing call sites while preserving their boundary-specific behavior.

### Per-call-site equivalence

The flag defaults reproduce the pre-refactor shared `supportedDataType` exactly, so existing expression-side callers are unchanged. Each boundary opts in/out via named arguments:

| Boundary | Old predicate accepted | New call | Delta |
|---|---|---|---|
| Expression-side callers (`literals`, `hash`, `arithmetic`, `CometScalarSubquery`) | old shared `supportedDataType` | defaults | none |
| Native shuffle (`nativeShuffleFailureReasons`) | scalars, strings, time types, YM/DT + calendar intervals, complex | `allowComplex = true, allowIntervals = true` | none |
| Columnar shuffle (`columnarShuffleFailureReasons`) | scalars, strings, time types, complex with distinct field names — no intervals of any kind | `allowComplex = true, allowCalendarInterval = false, allowDuplicateStructFieldNames = false` | none |
| Sinks (`CometSink.convert`, shuffle direct read) | `supportedSinkDataType`: YM/DT intervals + complex recursion over the shared predicate | `allowComplex = true, allowIntervals = true` | none |
| Local table scan (`CometLocalTableScanExec.isTypeSupported`) | `NullType`/YM/DT at top level, everything else via `DataTypeSupport` | `allowComplex = true, allowIntervals = true, allowTimeType = false, allowAnyStringType = false`, then `super` only to record fallback reasons | **intended**: intervals and `NullType` are now also accepted *nested* inside complex types, enabled by the nested interval codegen from #4976 |

Added recursive capability coverage (including `CalendarIntervalType` gating at the columnar shuffle boundary) and ran focused Spark 4.1 and Spark 3.5 tests, interval SQL tests, Spotless, and Scalastyle.


andygrove (commented):
Nice refactor. Folding the four duplicated capability checks into one `supportedDataType` with opt-in flags is a real improvement, and I traced the callers and flag defaults to convince myself it is behavior-preserving. All new capabilities are gated so existing default callers see identical behavior, and the callers in `CometScalarSubquery`, `hash.scala`, `arithmetic.scala`, and `literals.scala` all use bare defaults or named `allowComplex`, so none of them misbind the newly inserted positional parameters.

**The one spot worth a look is `CometLocalTableScanExec.isTypeSupported`:**

```scala
val supported = supportedDataType(dt, allowComplex = true, allowIntervals = true,
  allowTimeType = false, allowAnyStringType = false)
if (!supported) super.isTypeSupported(dt, name, fallbackReasons)
supported
```

The `super.isTypeSupported(...)` result is discarded and the method always returns `supported`, which on first read looks like a dropped return value. I think it is actually safe. `DataTypeSupport.isTypeSupported` only returns true for default scalars, the default `StringType` object, and nested-of-supported. It rejects `TimeType`, collated strings, and intervals. Every type `supportedDataType` rejects here is also rejected by super, so there is no type where super would say true while `supported` is false. Super is being called only to record the fallback reason.

Could we make that intent explicit, either as `if (supported) true else super.isTypeSupported(dt, name, fallbackReasons)`, which is behavior-identical for the reasons above, or a short comment noting super is called only for its fallback-reason side effect? Right now a future reader could easily "fix" this into an actual behavior change.

**On the new unit test:** the recursive coverage is a nice addition, and a pure Scala test is the right call since this is a predicate rather than an executed expression. One small gap: it does not cover the `allowTimeType` flag, which is the one capability only local scan relies on. A case asserting a time type is rejected when `allowTimeType = false` and accepted when true would round it out.

**Minor:** since this is stacked on #4976 and carries the duplicate `classOf[DurationVector]` lines in `CometBatchKernelCodegenInput.scala`, a rebase after #4976 merges will keep those cleanups from reappearing here.

I could not find any behavior change beyond the intended addition of interval support to the local-scan and sink paths. The only thing I would want addressed before merge is clarifying the discarded `super` call, and even that is about readability rather than correctness.

andygrove:

Collapsing five near-identical recursive type predicates into one is worth doing. That duplication is exactly why new types keep getting supported in one place and not another.

That said, I do not think this is behavior-preserving, and for a refactor that gates shuffle eligibility that matters a lot.

**Columnar shuffle now accepts `CalendarIntervalType`**

The old `supportedSerializableDataType` for columnar shuffle listed:

```scala
case _: BooleanType | ... | _: DateType | _: NullType => true
```

with no `CalendarIntervalType`. The new call is `supportedDataType(dt, allowComplex = true, allowDuplicateStructFieldNames = false)`, and `CalendarIntervalType` is in the unconditional base arm of `supportedDataType`, not behind `allowIntervals`. So calendar intervals are now eligible for columnar shuffle where they previously were not.

Columnar shuffle converts unsafe rows to Arrow in `shuffle/row.rs`. Does that path handle `CalendarInterval`? If it does not, this turns a clean fallback into a runtime failure. If it does, then the change is fine but it is a functional change that should be called out and tested rather than arriving as a side effect of a refactor.

**`CometLocalTableScanExec` no longer consults `super` for complex types**

The old override was:

```scala
case _: NullType | _: YearMonthIntervalType | _: DayTimeIntervalType => true
case _ => super.isTypeSupported(dt, name, fallbackReasons)
```

so a `StructType` went to `DataTypeSupport.isTypeSupported`. The new version returns `true` from `supportedDataType(..., allowComplex = true, ...)` for any nested struct of supported leaves and never reaches `super`. Whatever `DataTypeSupport` was doing for complex types at that boundary, including any config gating and any `fallbackReasons` messages it appended, is now skipped. Is that intended? If it is, the comment should say that local scans deliberately no longer honour `DataTypeSupport`'s complex-type rules.

There is a related side effect: when the new check returns `true`, nothing is appended to `fallbackReasons`, which is correct, but when it returns `false` and `super` also returns `false`, the reason now comes only from `super`. Worth confirming the messages users see are unchanged.

**Six boolean parameters is a lot of surface**

`supportedDataType(dt, allowComplex, allowIntervals, allowTimeType, allowAnyStringType, allowDuplicateStructFieldNames)` has defaults that pull in different directions: `allowComplex` and `allowIntervals` default to restrictive, while `allowTimeType`, `allowAnyStringType`, and `allowDuplicateStructFieldNames` default to permissive. A caller who forgets a flag gets stricter behavior in two cases and looser behavior in three, and at a call site `supportedDataType(dt, allowComplex = true, allowIntervals = true)` gives no hint that three other switches exist and are silently on.

Would named per-boundary entry points be better? Something like `supportedForNativeShuffle(dt)`, `supportedForColumnarShuffle(dt)`, `supportedForSink(dt)`, each a one-liner over the core function. Then the boundary's policy lives in one named place, the call sites are unambiguous, and adding a sixth switch does not require auditing every caller.

**The description should show the equivalence**

For a refactor across shuffle, sinks, and local scans, "preserving their boundary-specific behavior" is a claim that needs evidence. Could the description include a small table, one row per call site, listing the old predicate's accepted set and the flags chosen to reproduce it? That is what a reviewer needs in order to check the claim, and it is how the `CalendarIntervalType` difference above would have been caught before review.

**One process question**

The description says this depends on #4976. Is that merged? If not, this should presumably wait, and it would be worth marking as such.

peterxcli:
Thanks for the detailed pass @andygrove — one of the findings was real.

**Columnar shuffle + `CalendarIntervalType`: confirmed and fixed.** The old columnar predicate rejected it and the refactor accidentally accepted it via the unconditional base arm. The native unsafe-row-to-Arrow converter (`native/shuffle/src/spark_unsafe/row.rs`) has no interval support, so this would have surfaced as a native `unreachable!` panic instead of a clean Spark fallback. Fixed by adding an `allowCalendarInterval` capability (default `true` — every other boundary and the default expression-side callers accepted it pre-refactor), with the columnar shuffle call site passing `false`, plus unit-test coverage for the top-level and nested cases.

**Local scan and `super`:** the code is `if (supported) true else super.isTypeSupported(...)`, so `super` *is* consulted on every rejection and its fallback messages are unchanged. When the new predicate accepts, `super` would also have accepted and appends nothing — and `DataTypeSupport.isTypeSupported` has no config gating for complex types, so nothing is skipped. The one behavioral delta is intended: intervals nested inside structs/arrays/maps are now accepted, which is what this stack (with #4976) adds. I added a comment making the super call's reason-recording role explicit.

**Flag surface:** the defaults reproduce the pre-refactor shared `supportedDataType` exactly, which is why they point in different directions — restrictive flags opt boundaries *into* more (complex, intervals), permissive ones let boundaries opt *out* of what the legacy predicate already accepted (time types, collations, duplicate field names). Flipping everything restrictive would mean touching every legacy expression-side caller. All boundary call sites use named arguments and each boundary has exactly one call site, so the per-boundary policy already lives in one named place; happy to add named wrappers (`supportedForColumnarShuffle` etc.) as a follow-up if you'd prefer that shape.

**Description:** updated with a per-call-site equivalence table showing the old accepted set, the flags chosen, and the deltas.

**#4976** merged on July 28 and this branch has since been updated from main; the `DurationVector` duplication is gone.

andygrove:
> **Note on this review:** this was drafted by an LLM (Claude Code) at my request. Unlike my previous pass, the findings here were checked against a local build of this branch rather than read off the diff, but please still push back on anything that is wrong or already handled.

Thanks for turning the last round around so quickly. I checked out the branch, rebuilt native, and confirmed the `CalendarIntervalType` fix. With `spark.comet.shuffle.mode=jvm` and `spark.comet.shuffle.enabled=true`:

```scala
spark.sql("select id, make_interval(1,2,3,4,5,6,7) as i from range(100)").repartition(4).collect()
```

now plans a plain Spark `Exchange` and returns 100 rows, where before it failed with `Comet Internal Error: Unsupported type: Interval(MonthDayNano)` out of `Native.writeSortedFileNative`. Defaulting the new flag to `true` is the right call given every other boundary accepted calendar intervals pre-refactor.

I also re-traced all four boundaries against the six-flag signature and they reproduce the old predicates exactly, including the local scan case where `DataTypeSupport`'s `StructType` arm recurses through the unqualified `isTypeSupported` and dispatches back into the override. Your reading of the `super` call was right and the comment you added makes it clear. I tried to find a matching latent failure on the native shuffle boundary, which accepts `CalendarIntervalType` both before and after this PR, and could not produce one, so I am not raising it here.

**Nothing locks the fix in place.** The new unit test asserts that the predicate returns `false`, which pins the predicate but not the call site. If someone drops `allowCalendarInterval = false` from the columnar shuffle call, the unit test still passes and the bug comes straight back. Could we add the query above as a regression test, asserting the plan contains a Spark `Exchange` rather than a `CometColumnarExchange`? Right now the only thing standing between us and a repeat is a comment.

**The test shape cannot catch the bug we just found.** More generally the test asserts each flag in isolation, but the defect was a call site passing the wrong flag set, and flag-level assertions structurally cannot catch that. Since the whole value of this refactor is that four boundaries share one predicate, would you consider restructuring it as a per-boundary table? One row per call site, each listing types that must be accepted and rejected, so the columnar shuffle row pins down that calendar intervals and duplicate field names are out while the native shuffle row pins down that intervals are in. That is cheap to write and it is the artifact that would have caught this before review.

**The `allowAnyStringType` assertion is vacuous on two of the five CI versions.** `assert(!supportedDataType(CharType(1), allowAnyStringType = false))` only exercises the flag where `CharType` is a `StringType` subclass. Checking the Spark sources, `v3.5.8` has `case class CharType(length: Int) extends AtomicType` and it only became `extends StringType(...)` in `v4.0.1`, so on Spark 3.4 and 3.5 it hits the default `false` arm regardless of the flag. Since the comment in `CometLocalTableScanExec` says the intent is rejecting non-default collations, could we use a collated `StringType` instead? I verified on 4.1 that `supportedDataType(StringType("UTF8_LCASE"), allowAnyStringType = false)` is `false` and `true` by default, so it exercises the flag directly. `QueryPlanSerde.isStringCollationType` returns `false` on the Spark 3.x shim if that helps with gating.

While we are in that test, it is a pure predicate test rather than an expression correctness test, and `CometExpressionSuite` is already very large. Would `spark/src/test/scala/org/apache/comet/serde/` be a better home, alongside `CometScalarFunctionSuite`?

**`supportedDataType` still has no scaladoc.** The inline comments you added at the columnar shuffle call site are exactly right, particularly the `row.rs` pointer. Could we go one step further and document the function itself, covering what each of the six flags gates and which boundary needs it? I am not reopening the named-wrappers question, your answer on that was fair. But the function is now the single chokepoint for four boundaries plus the default expression-side callers in `hash.scala`, `arithmetic.scala`, `literals.scala`, and `CometScalarSubquery.scala`, and it carries no documentation at all. The native shuffle rationale that used to sit above the deleted predicate, that it relies on the Arrow IPC writer and so should support everything Comet supports, is gone with nowhere to land. That missing knowledge is the mechanism that produced the bug. Nothing in the function records which boundary can handle which type, so a type added to the unconditional base arm silently widens all four at once. It would also be worth noting in the doc that the polarity is mixed, since `allowComplex` and `allowIntervals` default restrictive while the other four default permissive.

**On CI:** `Verify TPC-H Results` is red at the `Build project` step. I could not read the log, but the same step passed in `Verify TPC-DS Results` and every other build, lint, and test-group job is green, which rules out a compile error. Looks like flake. Could you kick off a re-run so we have a clean board before merge?

peterxcli:
@andygrove Addressed your review. please take another look, thanks!

peterxcli:
@andygrove I've address the review comment, could you please take another look? thanks!

PR #4986, opened 2026-07-20
refactor: unify shuffle configs under `spark.comet.shuffle.*` prefix

andygrove:
## Summary

Part of #4978 (Category 2 + Category 4).

Shuffle-related configs live under four disjoint prefixes today:
`spark.comet.exec.shuffle.*`, `spark.comet.columnar.shuffle.*`,
`spark.comet.native.shuffle.*`, and `spark.comet.shuffle.*`. This PR
consolidates all shuffle configs under `spark.comet.shuffle.*` with
`.jvm.` and `.native.` sub-namespaces for mode-specific settings,
and fixes the dots-in-segment naming for the four JVM-shuffle keys.

The sub-namespaces (`.jvm.` / `.native.`) match the values users already
set on `spark.comet.shuffle.mode` (`jvm` / `native` / `auto`). Note that
both modes are columnar (both use Arrow); the distinction is where the
shuffle logic runs, so `.jvm.` is the honest name.

Every renamed key is registered via `.withAlternative(...)`, so
existing user configurations keep working (a one-time deprecation
warning is logged per old key). No behavior change.

## Rename table

**Master / shared (both modes):**

| Old key | New key |
| --- | --- |
| `spark.comet.exec.shuffle.enabled` | `spark.comet.shuffle.enabled` |
| `spark.comet.exec.shuffle.mode` | `spark.comet.shuffle.mode` |
| `spark.comet.exec.shuffle.directRead.enabled` | `spark.comet.shuffle.directRead.enabled` |
| `spark.comet.exec.shuffle.convertFromSparkPlan.enabled` | `spark.comet.shuffle.convertFromSparkPlan.enabled` |
| `spark.comet.exec.shuffle.revertRedundantColumnar.enabled` | `spark.comet.shuffle.revertRedundantColumnar.enabled` |
| `spark.comet.exec.shuffle.compression.codec` | `spark.comet.shuffle.compression.codec` |
| `spark.comet.exec.shuffle.compression.zstd.level` | `spark.comet.shuffle.compression.zstd.level` |

**Native-shuffle-only** (moved from `spark.comet.native.shuffle.*` and one leaf from `spark.comet.exec.shuffle.*`):

| Old key | New key |
| --- | --- |
| `spark.comet.exec.shuffle.writeBufferSize` | `spark.comet.shuffle.native.writeBufferSize` |
| `spark.comet.native.shuffle.partitioning.hash.enabled` | `spark.comet.shuffle.native.partitioning.hash.enabled` |
[...]

PR #5026, opened 2026-07-24
refactor: group orphan `spark.comet.explain.*` configs under one prefix

andygrove:
## Summary

Part of #4978 (Category 3).

Three explain-related boolean flags lived at the top level while their close
siblings (`format`, `native.enabled`, `rules`) already sat under
`spark.comet.explain.*`. This PR moves them under the group so the config
namespace is discoverable and consistent.

## Rename table

| Old key | New key |
| --- | --- |
| `spark.comet.explainFallback.enabled` | `spark.comet.explain.fallback.enabled` |
| `spark.comet.explainCodegen.enabled` | `spark.comet.explain.codegen.enabled` |
| `spark.comet.logFallbackReasons.enabled` | `spark.comet.explain.fallback.log.enabled` |

For the placement question raised by the issue on `logFallbackReasons`
(either `explain.fallback.log.enabled` or `fallback.log.enabled`), I went with
`explain.fallback.log.enabled`. `logFallbackReasons` and `explainFallback` are
the two fallback-reason reporters — one aggregated WARN vs one WARN per
reason — so nesting the "log every reason" variant under the same
`.explain.fallback.*` parent makes the pairing visible in tab-completion and
docs.

Every renamed key is registered via `.withAlternative(...)`, so existing user
configurations keep working (a one-time deprecation warning is logged per old
key). No behavior change.

## Not in this PR

- The `ENABLE_COMET_LOG_FALLBACK_REASONS` env var is preserved as-is. There is
  no alias mechanism for env vars, and renaming it would silently break users
  who set it in their launcher configs.
- The `dev/diffs/iceberg/*.diff` files still reference the old string keys.
  They keep working through aliases, and the shuffle rename PR (#4986) took
  the same approach — regenerating those diffs is out of scope here.

## Val rename

`val COMET_LOG_FALLBACK_REASONS` is renamed to
`COMET_EXPLAIN_FALLBACK_LOG_ENABLED` to match the new key (the two sibling
vals already had matching names, so only this one moved).

## Test plan

- [x] `./mvnw -Pspark-3.5 test -pl spark -Dsuites="org.apache.comet.CometConfSuite"` — 10 tests pass (including 3 new alias tests)
- [x] `./mvnw -Pspark-3.5 -pl spark spotless:check` — clean
- [x] `./mvnw -Pspark-3.5 -pl spark test-compile` — clean
- [ ] CI runs the broader test matrix


comphead (approved):
Thanks @andygrove it makes sense to structure explain configs!

comphead:
We would need to start Upgrade guide to cover migration path for this PR changes and also https://github.com/apache/datafusion-comet/pull/4986

andygrove:
> We would need to start Upgrade guide to cover migration path for this PR changes and also #4986

The old config names are still respected. Users do not have to make changes. Maybe we should consider upgrade guides though.

PR #4267, opened 2026-05-08
feat(experimental): ScalaUDF and Java UDF support via Janino codegen

mbutrovich:
## Which issue does this PR close?

Closes #.

## Rationale for this change

Builds on the JVM UDF bridge (#4232) and the per-task-attempt-id bridge cache (#4345). Adds `CometScalaUDFCodegen`, a `CometUDF` that compiles a specialized Arrow batch kernel per bound `ScalaUDF` and input schema via Janino. Without it, any plan containing a `ScalaUDF` falls back to Spark for the enclosing operator.

- Any `ScalaUDF` whose argument and return types are in the supported surface routes through native, no hand-written `CometUDF` required.
- The whole argument tree binds together, so Catalyst sub-expressions (`upper(s)`, `concat(c1, c2)`, `monotonically_increasing_id()`, HOFs like `transform` / `filter` / `array_max`) compile into the same per-row loop as the user function.
- Surrounding native operators stay native. The UDF is no longer a whole-operator fallback boundary.

Opt-in via `spark.comet.exec.scalaUDF.codegen.enabled` (default `false`, experimental).

Iceberg ships ScalaUDFs that real workloads hit: `IcebergSpark.registerBucketUDF` / `registerTruncateUDF` for partition-aligned predicates, and `RewriteDataFiles` with `sort-strategy=zorder` for compaction. With this PR enabled, those run natively and the surrounding project / exchange / sort stay on the Comet path. Without it, the operator falls back and the shuffle gets demoted from `CometExchange` to `CometColumnarExchange`.

The dispatcher is one of potentially many `CometUDF` implementations the bridge can route to. Hand-written `CometUDF`s for specific expression families remain a parallel path, dispatched by class name from the proto.

## What changes are included in this PR?

| Area | Where |
| --- | --- |
| Core codegen + dispatcher + serde | `spark/src/main/scala/org/apache/comet/codegen/`, `.../udf/codegen/`, `.../serde/CometScalaUDF.scala` |
| Tests | 4 suites + shared assertions under `spark/src/test/scala/org/apache/comet/CometCodegen*.scala` |
| Cross-version shims | `spark/src/main/spark-{3.x,4.0,4.1,4.2,4.x}/.../shims/` |
| Native (Rust) | FFI / planner / jni-bridge cleanup |
| Docs | `docs/source/user-guide/latest/scala_java_udfs.md` |
| CI | New suite names in `pr_build_*.yml` |

### Where to focus review

[...]

PR #5027, opened 2026-07-24
feat: account JVM UDF Arrow allocations in Spark task memory

### LINKED ISSUE #4174: Register CometArrowAllocator as a Spark MemoryConsumer for JVM-UDF dispatch
## Describe the problem

`CometUdfBridge.evaluate` (`common/src/main/java/org/apache/comet/udf/CometUdfBridge.java`, on the JVM-scalar-UDF prototype branch) allocates output Arrow vectors via the project-wide `CometArrowAllocator`. That allocator is a `RootAllocator` that is not registered with Spark's `TaskMemoryManager`, so off-heap memory consumed by the UDF dispatch path is invisible to Spark's task memory accounting and back-pressure machinery.

Under workloads with many concurrent JVM-UDF tasks per executor, this can drive native off-heap usage past the operator-level limits Spark would otherwise enforce.

Either:

1. Register `CometArrowAllocator` as a `MemoryConsumer` in Spark's `TaskMemoryManager` so allocations and frees update the task's accounting.
2. Allocate UDF output vectors from a child allocator that is itself registered as a per-task consumer, so leakage and accounting stay scoped to the task.

Option (2) is closer to the existing Spark-Arrow integration pattern.

## Additional context

Identified during code review of the JVM-scalar-UDF prototype. Filed as a follow-up so the prototype PR can ship without a Spark-integration redesign.

peterxcli:

Closes #4174.

## Rationale for this change

`CometScalaUDFCodegen`, currently the only `CometUDF` implementation, creates output Arrow vectors from the process-wide `CometArrowAllocator`. These JVM allocations bypass Spark’s task memory accounting.

FFI input vectors only wrap native-owned buffers that are already accounted by Comet, so they must not be charged again. Output buffers, however, are created on the JVM and may remain alive through the FFI release callback after the Java vector closes.

- Add a per-task Arrow child allocator backed by a non-spillable Spark `MemoryConsumer`.
- Pass the allocator through the generic `CometUDF` interface and codegen output path.
- Keep imported native buffers on the root allocator to avoid double-accounting.
- Charge output buffers to the Spark task while the UDF holds them; at export, buffer accounting moves to the root allocator along with native ownership, so native operators that retain the buffers (hash join builds, shuffle) take the only Spark reservation instead of charging the same physical buffers twice.
- Accounting requires off-heap Tungsten memory (`spark.memory.offHeap.enabled`); with on-heap Tungsten memory the off-heap Arrow buffers have no matching Spark pool and are tracked by the allocator but not charged.
- Interface note: `CometUDF.evaluate` gains an allocator parameter, so existing implementations need updating. The doc also now states that `evaluate` may be called concurrently from multiple Tokio workers within one task — this was already true before this PR (`CometScalaUDFCodegen` has synchronized its body since #4267 for exactly this reason); the previous "at most one thread" wording was incorrect.

Added `CometUdfBridgeTest` covering allocation accounting, export ownership transfer, task completion, straggler evaluations racing teardown, and final cleanup. The focused test passes with Spark 4.1 and Spark 3.5/Scala 2.12, and `CometCodegenSuite` passes end-to-end.

Also ran few scala script with spark shell to verify:

Check | main | branch
-- | -- | --
Small output task peak | 0 | 262,144 bytes
32 MiB output task peak | 0 | 50,397,184 bytes
Peak increase | 0 | 50,135,040 bytes
256 MiB batch under 256 MiB pool | Succeeds, bypassing Spark | Rejected by TaskMemoryManager
Same workload with batch size 32 | — | Succeeds, 268,435,456 bytes processed


peterxcli:
@sunchao would you like to take a look at this? TIA!

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
```diff
@@ -254,4 +261,152 @@ private static void evaluateInternal(
...
+    }
+
+    @Override
+    public synchronized void onPreAllocation(long size) {
+      if (completed) {
+        throw new OutOfMemoryException(
+            "Cannot allocate " + size + " JVM UDF bytes after task completion");
+      }
+      if (consumer != null) {
+        long acquired = consumer.acquireMemory(size);
```
sunchao: **[P2] Keep buffer releases runnable during blocking memory acquisition**

Could we avoid holding the `TaskState` monitor while calling `consumer.acquireMemory(size)`? Native prefetch can allocate the next UDF output concurrently with the Spark thread closing the previous output. Under memory pressure, Spark's execution pool waits here while retaining this monitor, so `onRelease` cannot return the previous buffer's Spark charge even when that release would satisfy the pending allocation.

I reproduced this with two real Spark 4.1.3 tasks and a 64 MiB off-heap pool: another task holds 62 MiB, this task retains 1 MiB of input and 1 MiB of previous output, and its producer requests another 1 MiB. The producer waits in `ExecutionMemoryPool.acquireMemory`, while closing the previous output and completing the task block in `TaskState.onRelease`. Both proceed only after the other task releases memory. The unchanged `CometTaskMemoryManager` control releases the previous buffer and completes before the other task frees anything. Please preserve accounting synchronization without blocking releases behind the pool wait.

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
```diff
@@ -254,4 +261,152 @@ private static void evaluateInternal(
...
+        completed = true;
+        instances.clear();
+        if (consumer != null && consumer.getUsed() > 0) {
+          // Spark discards all remaining task accounting immediately after completion listeners.
+          // Release it here; later FFI callbacks only drive allocator cleanup.
+          consumer.freeMemory(consumer.getUsed());
+        }
+        if ((allocator == null || allocator.getAllocatedMemory() == 0) && !closed) {
+          closed = true;
+          toClose = allocator;
```
sunchao: **[P2] Fence in-flight allocations before closing the task allocator**

Could we account for admitted/in-flight allocations before treating this allocator as empty? Arrow 18.3.0 calls `onPreAllocation` before updating `getAllocatedMemory()`. Completion can therefore observe zero and set `closed = true` while an allocation has already passed the completion check. If that allocation finishes before the subsequent `allocator.close()`, Arrow marks the allocator closed and throws because it still owns a buffer. The bridge removes the `TASKS` entry, but the child remains permanently attached to `ROOT_ALLOCATOR`, and later releases cannot repair cleanup.

This overlap is reachable when `CometCollectLimitExec.executeCollect` stops through Spark's `executeTake` while a native-scan producer is still in a synchronous JVM UDF call. `releasePlan` drops the receiver/context without joining that detached producer, so listener ordering alone does not fence the allocation.

A controlled interleaving against the exact bridge and Arrow versions reproduced permanent child-allocator and `TaskContext` retention with assertions disabled. With assertions enabled, the subsequent buffer release also fails and retains the buffer. Please include in-flight allocations in the lifecycle check before closing the child.

sunchao (commented):
Two P2 findings in task-allocator concurrency, detailed inline. Validated with targeted JVM reproductions against this exact head and Arrow 18.3.0. The native producer/cleanup paths were source-traced, not run end-to-end.

sunchao:
@peterxcli can you check the CI failures?

peterxcli:
> @peterxcli can you check the CI failures?

@sunchao I opened a fix for it 

https://github.com/apache/datafusion-comet/pull/5439

sunchao:
Ah thanks! Approved the PR and will merge soon

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
```diff
@@ -254,4 +271,225 @@ private static void evaluateInternal(
...
+      // releases that can satisfy a blocking acquire.
+      synchronized (taskMemoryManager) {
+        synchronized (this) {
+          if (completed) {
+            throw new OutOfMemoryException(
+                "Cannot allocate " + size + " JVM UDF bytes after task completion");
+          }
+        }
+
+        long acquired = consumer == null ? size : consumer.acquireMemory(size);
```
sunchao, line 510: **[P2] Avoid charging exported UDF buffers twice**

The new Spark charge remains until FFI release, but native hash joins and shuffles reserve the same buffers against Spark again. `JvmScalarUdfExpr` imports the output without copying, and Arrow's single-batch concatenation also shares the original buffers. A UDF feeding a single-batch hash-join build can therefore fail its native reservation even when the physical allocations fit the pool.

A focused component probe using the exact JVM sources with Spark 4.1.3 and Arrow 18.3.0 confirmed this: a 32 MiB buffer in a 48 MiB pool receives the full 32 MiB native reservation with the root-allocator control. With this change, the retained JVM charge leaves only about 16 MiB for that same reservation. The FFI buffer address is unchanged.

Could we coordinate accounting ownership across the FFI boundary, or exclude already-charged buffers from native reservations? The native query path was source-traced; this reproduction was not an end-to-end native query.

andygrove:

Getting JVM UDF output vectors onto Spark's task accounting is worth doing, and the distinction between imported native buffers (already accounted, keep on the root allocator) and JVM-created output buffers (charge to the task) is the right one. The retention logic that keeps the allocator alive until the last FFI release, while dropping Spark accounting at task completion, is carefully thought through.

Four things.

**The threading contract for `CometUDF` changed**

The class doc goes from:

> At any instant at most one thread is inside `evaluate()` for a given `taskAttemptId`.

to:

> Calls for a task may arrive concurrently from different Tokio workers. Implementations with mutable state are responsible for synchronizing `evaluate()`.

That is a breaking change to the contract that user-written `CometUDF` implementations were told they could rely on. Anyone who wrote a UDF holding mutable per-task state, which the old wording explicitly invited, now has a data race.

Was the old guarantee wrong all along, or does this PR change when concurrent calls can happen? Either way this deserves its own line in the description and probably a migration-guide note, because it is a much bigger deal for users than the memory accounting is. Right now it reads like an incidental doc edit.

**Lock ordering deserves to be written down**

`onPreAllocation` takes the `TaskMemoryManager` monitor, then the `TaskState` monitor, then calls back into `TaskMemoryManager.acquireExecutionMemory`. `taskCompleted` and `onRelease` take the `TaskState` monitor and then call into `MemoryConsumer.freeMemory`, which reaches the `MemoryManager` monitor. Spark's own `acquireExecutionMemory` can call `spill()` on other consumers while holding the `TaskMemoryManager` monitor, and an Arrow buffer release from such a path would re-enter `onRelease`.

I worked through it and did not find an actual inversion, but it took a while and I am not certain. Could you add a short comment naming the three locks and the order they must always be taken in? Anything that acquires a Spark memory lock from inside an Arrow allocation listener callback deserves that much.

**On-heap Tungsten mode silently does nothing**

```java
this.consumer = taskMemoryManager.getTungstenMemoryMode() == MemoryMode.OFF_HEAP
    ? new TaskMemoryConsumer(taskMemoryManager) : null;
```

With on-heap Tungsten memory the whole accounting path is disabled. That is probably correct, since Arrow's buffers are off-heap and charging them to Spark's on-heap pool would be wrong. But it is not stated anywhere, and a user running on-heap who reads the release notes will believe their UDF memory is accounted. Could the class doc say so explicitly, and ideally log once at debug level when accounting is skipped?

**The registration-order dependency**

`registerTask` has to be called before `CometExecIterator` registers its own completion listener, because Spark runs listeners in reverse order. That is documented in the javadoc, which is good, but it is enforced by nothing. If someone reorders those two calls the failure is a use-after-free of an allocator with buffers still exported, which will not reproduce reliably.

Is there a way to make that structural rather than conventional, for example having `CometExecIterator` obtain the `TaskState` and register both listeners itself? If not, is one of the tests in `CometUdfBridgeTest` specifically asserting the ordering?

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
peterxcli: Good catch — fixed in 2fafac953 by transferring accounting ownership across the FFI boundary at export: the result's buffers move to the root allocator (zero-copy, FFI addresses unchanged) and the Spark charge is released at that point, so a retaining native operator's reservation is the only charge. The UDF is still charged while it holds the memory, preserving the admission control this PR adds.

Verified per-chunk exact against Arrow 18.3.0's allocator internals: ownership transfers fire no `AllocationListener` callbacks and the eventual FFI release only notifies the chunk's owner at destruction time (the root allocator), so the explicit release cannot double-free; `getAccountedSize()` is non-zero only on the owning ledger, so pass-through input buffers and re-exported chunks contribute nothing. One wrinkle: the C schema is exported from the result's own `Field` (a transferred complex vector rebuilds child names from runtime data vectors, which Arrow hardcodes to `$data$`), while the C array — which carries no names — comes from the transferred vector.

The focused test was rewritten to assert the new contract (charge present while the UDF holds the output, gone at export while buffers are still readable, unchanged by FFI release), and `CometCodegenSuite` passes end-to-end (85/85). A side benefit: exported buffers no longer pin `TaskState` past task completion, so the deferred-release retention and the task-attempt-ID-reuse concern it covered are gone.

peterxcli:
@andygrove Thanks — evaluated all four; addressed in 1dc8fc71e.

**(1) Threading contract**: this is a doc correction, not a behavior change. The "at most one thread" wording was already contradicted on main: `CometScalaUDFCodegen` — the only `CometUDF` implementation — has run `evaluate` under `this.synchronized` since its introduction in #4267, with a comment explaining that `HashJoinExec` pipelines build/probe via `OnceAsync` (`tokio::spawn`), so multiple Tokio workers can call one task's dispatcher. Two more sources in `jni_api.rs`: the scan-free path drives each plan from a spawned prefetching Tokio task, and one Spark task can hold several such plans concurrently. Since this PR also changes the `evaluate` signature (allocator parameter), any out-of-tree UDF must be revisited regardless. I've grounded the trait doc in these concrete causes and will call it out in the PR description.

**(2) Lock ordering**: added a javadoc on `TaskState` naming the order — `TaskMemoryManager` monitor → `TaskState` monitor → `MemoryManager` monitor — verified against Spark 3.5.8 and 4.1.3 sources: `acquireExecutionMemory` holds the TMM monitor across spills (so a spill releasing Arrow buffers re-enters `onRelease` in the same TMM→TaskState order), and `releaseExecutionMemory` never takes the TMM monitor, so every release-side path is a suffix of the acquire-side order.

**(3) On-heap mode**: agreed — the behavior is correct (Arrow buffers are off-heap; there is no matching Spark pool), but it is now stated in the `TaskState` javadoc and the `CometUDF` contract, with a per-task debug log when accounting is skipped.

**(4) Registration order**: the failure mode is not use-after-free — the allocator only closes once in-flight evaluations finish and it holds no memory, and post-completion evaluation is rejected. But investigating this surfaced a real edge: a straggler `evaluate()` after state removal re-created state via `computeIfAbsent`, whose mapping function registered a completion listener Spark can invoke immediately — re-entering `TASKS.remove` from inside `computeIfAbsent`. Fixed by registering the listener outside `computeIfAbsent`, which makes the ordering structural (a graceful-shutdown nicety rather than a safety requirement), and added `stragglerEvaluationsAroundTaskCompletionAreSafe` covering evaluations racing and following task completion (Spark's deferred-listener-drain semantics verified identical in 3.5.8 and 4.1.3).

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
```diff
@@ -252,6 +302,332 @@ private static void evaluateInternal(
...
+   * Bytes of the result's buffers currently accounted against the task allocator, i.e. the portion
+   * of the Spark task charge that moves to native ownership on export. {@code getAccountedSize()}
+   * is non-zero only on the ledger that owns a chunk, so pass-through input buffers (owned by the
+   * root allocator) and chunks whose ownership already moved on a previous export contribute
+   * nothing.
+   */
+  private static long chargedOutputSize(FieldVector result, BufferAllocator outputAllocator) {
+    long charged = 0L;
+    Set<ReferenceManager> seen = Collections.newSetFromMap(new IdentityHashMap<>());
+    for (ArrowBuf buf : result.getBuffers(false)) {
```
sunchao: **[P2] Include allocated empty children when releasing export charges**

Could we enumerate each vector's physical field buffers recursively here, preserving ledger deduplication? `getBuffers(false)` omits allocated buffers for zero-length children, but `transferPair.transfer()` still moves them to the root allocator. The omitted bytes never reach `releaseExportedCharge`, and subsequent FFI release cannot return their Spark reservation.

On this head with Spark 4.1.3, Arrow 18.3.0 and JDK 17, a native `CometNativeScan -> CometProject` query using `regexp_extract_all(s, '(z)', 1)` on 8,192 nonmatching input rows left 786,432 bytes reserved after consuming all output. The matching control left zero. A diagnostic change to recursively collect physical field buffers also left zero. A separate empty-integer-array bridge probe exhausted a 1 MiB pool after 31 batches while no Arrow buffers remained live. Repeated empty/null array batches can therefore fail otherwise-fitting tasks.

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
```diff
@@ -252,6 +302,332 @@ private static void evaluateInternal(
...
+   *
+   * <p>The returned vector shares the original buffers and must be closed by the caller after
+   * export; the exported FFI array keeps the buffers alive until native execution releases them.
+   */
+  private static FieldVector transferForExport(
+      TaskState state, BufferAllocator outputAllocator, FieldVector result) {
+    long charged = chargedOutputSize(result, outputAllocator);
+    TransferPair transferPair = result.getTransferPair(result.getField(), ROOT_ALLOCATOR);
+    transferPair.transfer();
+    state.releaseExportedCharge(charged);
```
sunchao, line 331: **[P2] Account for ownership returning to retained scratch buffers**

For a custom `CometUDF` that retains a scratch vector and returns an aligned `splitAndTransfer` slice, this releases the whole chunk's Spark charge even though a task-allocator ledger still references it. When native releases the FFI result, Arrow can transfer ownership back to the retained scratch without an `onPreAllocation` callback. Closing scratch later invokes `onRelease` and frees the same Spark charge again.

A probe through the current public `CometUdfBridge.evaluate` with Spark 4.1.3 and Arrow 18.3.0 reproduced this: after FFI release, scratch still held 8,192 readable Arrow bytes with zero Spark charge. After allocating an unrelated 8,192-byte buffer, closing scratch reduced the Spark charge to zero while that unrelated buffer remained live; closing it then logged an over-release. Could we retain or coordinate accounting for shared chunks so returning ownership cannot cause duplicate release?

This is limited to the documented custom-CometUDF scratch-buffer contract. The in-tree codegen implementation creates fresh outputs; this reproduction exercised the bridge and Arrow C exports, not a native query using a custom UDF.

sunchao (commented):
Two P2 findings in output-buffer accounting, detailed inline. The empty-output issue was reproduced through a native SQL query; the shared-scratch issue was reproduced through the public bridge with a custom CometUDF.

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
peterxcli: Good catch — fixed in 9b728163b. `chargedOutputSize` now enumerates each vector's physical field buffers recursively (`getFieldBuffers()` + `getChildrenFromFields()`), preserving per-ledger dedup, instead of `getBuffers(false)`, whose `getBufferSize() == 0` short-circuit omits the allocated buffers of zero-length children while `transfer()` moves them to the root allocator all the same. Added `emptyChildAllocationsReleaseSparkChargeAtExport`: a 1024-row all-empty `ListVector` with an allocated child data vector previously stranded 32,768 bytes of Spark charge at export; it now moves the full charge and returns to baseline after the FFI release.

spark/src/main/java/org/apache/comet/udf/CometUdfBridge.java
peterxcli: Fixed in 9b728163b by retaining accounting for shared chunks, as you suggested. `chargedOutputSize` now releases a chunk's charge at export only when every live reference to its ledger comes from the result tree (`getRefCount()` equals the distinct result buffers on it). A chunk shared with retained scratch keeps its Spark charge, so when native releases the FFI result and Arrow silently returns ownership to the scratch ledger, the eventual scratch close fires `onRelease` and frees the charge exactly once; if scratch closes first while native still holds, the retained charge is dropped wholesale at task completion, the same bound as the pre-export model. Fresh outputs are unaffected: fixed-width vectors slice data+validity out of one combined chunk with two retained slices, both visible in `getFieldBuffers()`, so exclusive ownership is still detected and released at export (the existing export test guards this). Added `sharedScratchChunkChargeIsReleasedExactlyOnce` reproducing your scenario, including an unrelated allocation surviving the scratch close with no over-release.

andygrove:
Triage note: #5998 overlaps with this. It attaches an `AllocationListener` to the process-wide `CometArrowAllocator` root so that every JVM Arrow allocation — FFI export, broadcast coalescing, `CometSparkToColumnarExec`, the cached batch serializer, codegen output and the Python runner — is reserved from Spark's `TaskMemoryManager` in fixed blocks. That one is mine, and it is the first item of #5997.

They are not the same change. This PR cuts a per-task child allocator for the UDF output path specifically and charges it to a non-spillable `MemoryConsumer`, which is more precise on that path than a root-level listener, and the `CometUDF` interface change has no counterpart in #5998. But both charge the same output buffers and both touch `CometUdfBridge.java`, so landing them unchanged would double-count. Could we work out which layer owns the UDF output buffers before either goes further?

Issue #240, opened 2024-03-31
Add Spark expression coverage


viirya:
### What is the problem the feature request solves?

This is an umbrella ticket for the list of unsupported Spark expressions. This is not necessary comprehensive list of all Spark expressions because they are too many. We can start from frequently used expressions.

comphead:
The list of Spark expression can be found https://spark.apache.org/docs/latest/api/sql/index.html

viirya:
I added hash expressions. Free feel to edit the expression list in the issue description to add more expressions.

comphead:
I was thinking to add Spark `Scan OneRowRelation` scan support in addition to Parquet Scan. This will allow Comet be enabled when running queries like
```
select sqrt(2) from (select 1 union all select 2)
```

Once its done, we can just download all the queries from https://spark.apache.org/docs/latest/api/sql/index.html and run it automatically and see the coverage. How does it sound?

comphead:
Another potential solution we can do is to transform OneRowRelation  and to use  DF `PlaceholderRowExec`. I'll check if its doable

comphead:
I have an idea to do that. Planning to create a draft soon

comphead:
> > Basic idea is to grab queries from https://spark.apache.org/docs/latest/api/sql/index.html and create a separate task in Comet which will check if the Comet being triggered.
> 
> Hmmm, I think you can get expression example usage directly from its annotated class. See 'org.apache.spark.sql.expressions.ExpressionInfoSuite' for how to get examples directly.

Great idea, I think I'm able to fetch it as here https://github.com/apache/spark/blob/6fdf9c9df545ed50acbce1ec874625baf03d4d2e/sql/core/src/test/scala/org/apache/spark/sql/expressions/ExpressionInfoSuite.scala#L166

andygrove:
we have an updated list in the contrib guide now

PR #4898, opened 2026-07-12
feat: add CalendarIntervalType support

peterxcli:
## Which issue does this PR close?

Part of https://github.com/apache/datafusion-comet/issues/4540.

## Rationale for this change

Comet does not currently support Spark's `CalendarIntervalType`, causing queries carrying calendar interval columns through Comet operators to fall back to Spark.

This change adds the foundational Arrow and codegen support needed to preserve calendar intervals—months, days, and microseconds—across the Spark, Arrow, and native execution boundaries.

## What changes are included in this PR?

- Map Spark `CalendarIntervalType` to Arrow `Interval(MonthDayNano)`.
- Add `CALENDAR_INTERVAL` to the serialized data type protocol and native Arrow type conversion.
- Support reading and writing calendar intervals through Arrow vectors, including microsecond-to-nanosecond conversion.
- Add `CalendarIntervalType` support to Comet batch-kernel code generation for scalar and nested values.
- Allow calendar interval columns through Comet’s data type support checks and query plan serialization.
- Add SQL and unit-test coverage for nullable and non-null calendar interval values.

## How are these changes tested?

Added coverage for:

- Round-tripping `CalendarIntervalType` through the Arrow writer and `CometPlainVector`.
- Round-tripping calendar intervals through the batch-kernel codegen path.
- SQL queries containing positive, negative, mixed-component, and null calendar interval values with native local table scans enabled.

Issue #3099, opened 2026-01-14
[Feature] Support Spark expression: make_interval


andygrove:
## What is the problem the feature request solves?

> **Note:** This issue was generated with AI assistance. The specification details have been extracted from Spark documentation and may need verification.

Comet does not currently support the Spark `make_interval` function, causing queries using this function to fall back to Spark's JVM execution instead of running natively on DataFusion.

The `MakeInterval` expression creates a calendar interval from separate year, month, week, day, hour, minute, and second components. It supports flexible argument lists allowing omission of trailing components, which default to zero.

Supporting this expression would allow more Spark workloads to benefit from Comet's native acceleration.

### Spark Specification

**Syntax:**
```sql
make_interval(years, months, weeks, days, hours, mins, secs)
make_interval(years, months, weeks, days, hours, mins)
make_interval(years, months, weeks, days, hours)
-- ... (supports up to 0 arguments with overloaded constructors)
```

**Arguments:**
| Argument | Type | Description |
|----------|------|-------------|
| years | IntegerType | Number of years in the interval |
| months | IntegerType | Number of months in the interval |
| weeks | IntegerType | Number of weeks in the interval |
| days | IntegerType | Number of days in the interval |
| hours | IntegerType | Number of hours in the interval |
| mins | IntegerType | Number of minutes in the interval |
| secs | DecimalType(MAX_LONG_DIGITS, 6) | Number of seconds with microsecond precision |

**Return Type:** `CalendarIntervalType` - Returns a calendar interval object containing the specified time components.

**Supported Data Types:**
- **Integer types**: years, months, weeks, days, hours, minutes
- **Decimal type**: seconds (with scale 6 for microsecond precision)
- All arguments support implicit casting to their required types

**Edge Cases:**
- **Null handling**: Returns null if any input argument is null (null intolerant)
- **Overflow behavior**: 
  - When `failOnError=true` (ANSI mode): Throws `QueryExecutionErrors.arithmeticOverflowError`
  - When `failOnError=false`: Returns null on arithmetic overflow
- **Default values**: Missing trailing arguments default to `Literal(0)` except seconds which defaults to `Decimal(0, MAX_LONG_DIGITS, 6)`
- **Nullable result**: When `failOnError=false`, result is always nullable; when `true`, nullable only if any child is nullable

**Examples:**
```sql
-- Create interval with all components
SELECT make_interval(1, 2, 3, 4, 5, 6, 7.123456);
[...]

PR #5030, opened 2026-07-24
feat: support timestampadd and timestampdiff via codegen dispatch

andygrove:

There is no dedicated issue. This is part of the ongoing effort to provide native support for expressions that currently force a full fallback to Spark. Related to the expression coverage epic #240.

## Rationale for this change

`timestampadd` and `timestampdiff` have no Comet handler today, so any query using them falls the entire operator back to Spark. Both are regular expressions, not `RuntimeReplaceable`, so they reach serde directly.

Rather than native implementations, this PR routes them through the JVM codegen dispatcher. The dispatcher runs Spark's own generated code inside the native Comet pipeline, which keeps the operator native while guaranteeing bit-for-bit Spark compatibility across all supported versions. This avoids the datetime edge-case divergences (timezone and calendar handling) that a `chrono`-based native implementation would be prone to.

`make_interval` was originally part of this PR and has been split out into #5260, since its `CalendarIntervalType` output makes it independent of these two.

- `CometTimestampAdd` and `CometTimestampDiff` codegen-dispatch serdes in `datetime.scala`, registered in `QueryPlanSerde`'s `temporalExpressions` map.
- Comet SQL file tests covering all time units, negative quantities, month-end and leap-day rollover, whole-unit truncation, DST transitions, `TIMESTAMP_NTZ` inputs, ANSI overflow, and NULL inputs.
- Support status for the new functions in the Supported Spark Expressions guide, plus corrections to four existing rows.

### Grammar aliases

`timestampadd` and `timestampdiff` are not the only spellings that build these nodes. The grammar routes `TIMESTAMPADD | DATEADD | DATE_ADD` to `TimestampAdd`, and `TIMESTAMPDIFF | DATEDIFF | DATE_DIFF` (plus `TIMEDIFF` on 4.0+) to `TimestampDiff`, whenever the first argument is a `datetimeUnit` keyword. So `dateadd(DAY, 1, ts)` is a `TimestampAdd`, not a `DateAdd`, and before this PR it fell back.

That made four rows in the expression guide misleading: `date_add`, `date_diff`, `dateadd` and `datediff` all claimed `Native`, which is true only of the two-argument form. Those rows are corrected here, and `timediff`, `timestampadd` and `timestampdiff` rows are added.

The alias spellings are covered by tests, split by the Spark version that introduced each:

| Spelling | Available | Fixture |
| --- | --- | --- |
| `dateadd` / `datediff` | 3.4+ | inline in `timestampadd.sql` / `timestampdiff.sql` |
| `date_add` / `date_diff` (unit form) | 3.5+ | `date_add_unit_alias.sql` |
| `timediff` | 4.0+ | `timediff.sql` |

Spark 3.4's rule is only `TIMESTAMPADD | DATEADD` and `TIMESTAMPDIFF | DATEDIFF`; `DATE_ADD` and `DATE_DIFF` joined in 3.5, which is why the fixtures are gated with `MinSparkVersion` rather than kept in one file.

The string-literal unit form needs no fixture: it is rejected at parse time, not runtime. `visitTimestampadd` throws `invalidDatetimeUnitError` when `ctx.invalidUnit` is set, so `timestampadd('MONTH', 1, ts)` never reaches Comet.

New Comet SQL file tests run each query through both Spark and Comet and verify the results match and that Comet executes the expression natively (through the dispatcher) rather than falling back.

Coverage includes every unit accepted by `DateTimeUtils.timestampAdd` and `timestampDiffMap` (including `DAYOFYEAR`, which shares a case arm with `DAY`, and `QUARTER`, the one entry with arithmetic of its own), positive and negative quantities, month-end and leap-day boundaries, whole-unit truncation toward zero, and NULL inputs.

Because both functions do calendar arithmetic on local time rather than elapsed time, the fixtures pin `America/Los_Angeles` and assert across real DST transitions: spring-forward, fall-back with its ambiguous local hour, and a result landing on the nonexistent `2024-03-10 02:30`. These are exactly the cases a `chrono`-based native implementation would be most likely to get wrong.

`TIMESTAMP_NTZ` is covered separately for `timestampadd`, where `dataType` follows the input and an NTZ argument therefore compiles a distinct kernel with a different Arrow vector class and resolves `zoneIdForType` to UTC, and once for `timestampdiff`, where `inputTypes` casts NTZ up to `TimestampType`.

The error path is covered too: `DATETIME_OVERFLOW` from `timestampAddOverflowError`. The error fixture sits alongside valid-input queries so the case cannot pass vacuously through a fallback.

Verified green on Spark 3.5 and 4.1 locally after the rebase onto current `main`; previously verified on all four supported profiles (3.4, 3.5, 4.0, 4.1).

### DIFF
--- a/docs/source/user-guide/latest/expressions.md
+++ b/docs/source/user-guide/latest/expressions.md
@@ -254,15 +254,15 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `current_time` | 🔜 | — | Blocked on Spark 4.1 TIME type support ([#4288](https://github.com/apache/datafusion-comet/issues/4288)) |
 | `current_timestamp` | ✅ | — | Constant-folded to a literal before Comet sees the plan |
 | `current_timezone` | ✅ | — |  |
-| `date_add` | ✅ | Native |  |
-| `date_diff` | ✅ | Native |  |
+| `date_add` | ✅ | Native | The 2-argument form is native; the `date_add(UNIT, n, ts)` form parses to `timestampadd` and runs through codegen dispatch |
+| `date_diff` | ✅ | Native | The 2-argument form is native; the `date_diff(UNIT, start, end)` form parses to `timestampdiff` and runs through codegen dispatch |
 | `date_format` | ✅ | Hybrid |  |
 | `date_from_unix_date` | ✅ | Native |  |
 | `date_part` | ✅ | — |  |
 | `date_sub` | ✅ | Native |  |
 | `date_trunc` | ✅ | Hybrid |  |
-| `dateadd` | ✅ | Native |  |
-| `datediff` | ✅ | Native |  |
+| `dateadd` | ✅ | Native | The 2-argument form is native; the `dateadd(UNIT, n, ts)` form parses to `timestampadd` and runs through codegen dispatch |
+| `datediff` | ✅ | Native | The 2-argument form is native; the `datediff(UNIT, start, end)` form parses to `timestampdiff` and runs through codegen dispatch |
 | `datepart` | ✅ | — |  |
 | `day` | ✅ | Native |  |
 | `dayname` | ✅ | — | Abbreviated day name (Spark 4.0+) |
@@ -294,9 +294,12 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `session_window` | 🔜 | — | Batch session-window grouping falls back (`UpdatingSessionsExec` is not yet native); tracked by [#4785](https://github.com/apache/datafusion-comet/issues/4785) |
 | `time_diff` | 🔜 | — | Spark 4.1 TIME type; tracked by [#4288](https://github.com/apache/datafusion-comet/issues/4288) |
 | `time_trunc` | 🔜 | — | Spark 4.1 TIME type; tracked by [#4288](https://github.com/apache/datafusion-comet/issues/4288) |
+| `timediff` | ✅ | — | Spark 4.0+ grammar alias that parses to `timestampdiff`; runs through codegen dispatch |
 | `timestamp_micros` | ✅ | Codegen dispatch |  |
 | `timestamp_millis` | ✅ | Codegen dispatch |  |
 | `timestamp_seconds` | ✅ | Native |  |
+| `timestampadd` | ✅ | — | Reached through the grammar rather than the function registry; runs through codegen dispatch |
+| `timestampdiff` | ✅ | — | Reached through the grammar rather than the function registry; runs through codegen dispatch |
 | `to_date` | ✅ | — | Rewrites to `Cast` (or `Cast(GetTimestamp)` with a format) before Comet sees the plan |
 | `to_time` | 🔜 | — | Spark 4.1 TIME type; tracked by [#4288](https://github.com/apache/datafusion-comet/issues/4288) |
 | `to_timestamp` | ✅ | — | Rewrites to `Cast` (or `GetTimestamp` with a format) before Comet sees the plan |
--- a/docs/source/user-guide/latest/expressions.md
+++ b/docs/source/user-guide/latest/expressions.md
@@ -254,8 +254,8 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `current_time` | 🔜 | — | Blocked on Spark 4.1 TIME type support ([#4288](https://github.com/apache/datafusion-comet/issues/4288)) |
 | `current_timestamp` | ✅ | — | Constant-folded to a literal before Comet sees the plan |
 | `current_timezone` | ✅ | — |  |
-| `date_add` | ✅ | Native | The 2-argument form is native; the `date_add(UNIT, n, ts)` form parses to `timestampadd` and runs through codegen dispatch |
-| `date_diff` | ✅ | Native | The 2-argument form is native; the `date_diff(UNIT, start, end)` form parses to `timestampdiff` and runs through codegen dispatch |
+| `date_add` | ✅ | Native | The 2-argument form is native; the `date_add(UNIT, n, ts)` form (Spark 3.5+) parses to `timestampadd` and runs through codegen dispatch |
+| `date_diff` | ✅ | Native | The 2-argument form is native; the `date_diff(UNIT, start, end)` form (Spark 3.5+) parses to `timestampdiff` and runs through codegen dispatch |
 | `date_format` | ✅ | Hybrid |  |
 | `date_from_unix_date` | ✅ | Native |  |
 | `date_part` | ✅ | — |  |
--- a/spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
+++ b/spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
@@ -1,0 +1,58 @@
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file
+-- distributed with this work for additional information
+-- regarding copyright ownership.  The ASF licenses this file
+-- to you under the Apache License, Version 2.0 (the
+-- "License"); you may not use this file except in compliance
+-- with the License.  You may obtain a copy of the License at
+--
+--   http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+-- timestampadd runs through the codegen dispatcher so results match Spark exactly.
+-- Config: spark.sql.session.timeZone=America/Los_Angeles
+-- Config: spark.comet.exec.scalaUDF.codegen.enabled=true
+
+statement
+CREATE TABLE test_timestampadd(ts timestamp, q int) USING parquet
+
+statement
+INSERT INTO test_timestampadd VALUES
+  (timestamp'2024-01-15 10:30:45', 3),
+  (timestamp'2024-01-31 23:00:00', 1),
+  (timestamp'2024-02-29 12:00:00', 12),
+  (timestamp'2024-12-31 23:59:59', 2),
+  (timestamp'1970-01-01 00:00:00', -5),
+  (NULL, 1),
+  (timestamp'2024-06-15 00:00:00', NULL)
+
+-- column quantity across a range of units, including month-end and leap-day rollover
+query
+SELECT timestampadd(HOUR, q, ts) FROM test_timestampadd
+
+query
+SELECT timestampadd(MONTH, q, ts) FROM test_timestampadd
+
+query
+SELECT
+  timestampadd(YEAR, 1, ts),
+  timestampadd(QUARTER, 1, ts),
+  timestampadd(WEEK, 2, ts),
+  timestampadd(DAY, -10, ts),
+  timestampadd(MINUTE, 90, ts),
+  timestampadd(SECOND, 30, ts),
+  timestampadd(MICROSECOND, 500, ts)
+FROM test_timestampadd
+
+-- literal arguments (constant folding is disabled by the test suite)
+query
+SELECT
+  timestampadd(HOUR, 3, timestamp'2024-01-01 10:00:00'),
+  timestampadd(MONTH, 1, timestamp'2024-01-31 00:00:00'),
+  timestampadd(YEAR, 1, timestamp'2024-02-29 00:00:00')
--- a/spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
+++ b/spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
@@ -16,21 +16,22 @@
 -- under the License.
 
 -- timestampadd runs through the codegen dispatcher so results match Spark exactly.
+-- America/Los_Angeles is pinned so the DST cases below straddle real transitions.
 -- Config: spark.sql.session.timeZone=America/Los_Angeles
 -- Config: spark.comet.exec.scalaUDF.codegen.enabled=true
 
 statement
-CREATE TABLE test_timestampadd(ts timestamp, q int) USING parquet
+CREATE TABLE test_timestampadd(ts timestamp, ts_ntz timestamp_ntz, q int) USING parquet
 
 statement
 INSERT INTO test_timestampadd VALUES
-  (timestamp'2024-01-15 10:30:45', 3),
-  (timestamp'2024-01-31 23:00:00', 1),
-  (timestamp'2024-02-29 12:00:00', 12),
-  (timestamp'2024-12-31 23:59:59', 2),
-  (timestamp'1970-01-01 00:00:00', -5),
-  (NULL, 1),
-  (timestamp'2024-06-15 00:00:00', NULL)
+  (timestamp'2024-01-15 10:30:45', timestamp_ntz'2024-01-15 10:30:45', 3),
+  (timestamp'2024-01-31 23:00:00', timestamp_ntz'2024-01-31 23:00:00', 1),
+  (timestamp'2024-02-29 12:00:00', timestamp_ntz'2024-02-29 12:00:00', 12),
+  (timestamp'2024-12-31 23:59:59', timestamp_ntz'2024-12-31 23:59:59', 2),
+  (timestamp'1970-01-01 00:00:00', timestamp_ntz'1970-01-01 00:00:00', -5),
+  (NULL, NULL, 1),
+  (timestamp'2024-06-15 00:00:00', timestamp_ntz'2024-06-15 00:00:00', NULL)
 
 -- column quantity across a range of units, including month-end and leap-day rollover
 query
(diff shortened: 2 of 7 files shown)


peterxcli:
> make_interval was considered but left for a follow-up: its output type is CalendarIntervalType, which Comet's columnar layer does not yet support (the dispatcher reports unsupported output type CalendarIntervalType and falls back). It can be enabled once CalendarIntervalType support lands (https://github.com/apache/datafusion-comet/pull/4898).

I think you're refer to https://github.com/apache/datafusion-comet/issues/3099? I saw the native `SparkMakeInterval` implementation is ready and registered as comet udf, what we need to do is adding a new catalyst mapping `classOf[MakeInterval] -> CometMakeInterval` and `CometMakeInterval` scala implementation to convert expr to `make_interval` proto?

peterxcli (approved):
LGTM, only need to merge upstream and resolve conflict in expressions.md. 

ai review: non-blocker review:
1. add MILLISECOND to both tests and MICROSECOND to timestampdiff.
2. Add one TIMESTAMP_NTZ or DST-boundary case; current tests only exercise timestamp-LTZ away from DST transitions.

mbutrovich:
The description defers `make_interval` because "the dispatcher reports unsupported output type CalendarIntervalType and falls back". That is no longer true on main: commit 69df0f12b (#4898) added `CalendarIntervalType => true` to `CometBatchKernelCodegen.isSupportedDataType` (`spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala:92`), and it is wired end to end, not just in the predicate: `serializeDataType` handles it (`QueryPlanSerde.scala:530` and `:571`), the output side maps it to `IntervalMonthDayNanoVector` and emits a writer (`CometBatchKernelCodegenOutput.scala:178` and `:222`), and the input side emits a `getInterval` getter (`CometBatchKernelCodegenInput.scala:610` and `:717`). This PR is based on a pre-#4898 main, which is also why it conflicts. @peterxcli asked about exactly this. `MakeInterval` is a plain `SeptenaryExpression` with `nullIntolerant` and is not RuntimeReplaceable (Spark `intervalExpressions.scala:338-348`), so after the rebase `object CometMakeInterval extends CometCodegenDispatch[MakeInterval]` plus a `classOf[MakeInterval]` entry should be all it takes. Either add it here or correct the description. Note that `MakeInterval.failOnError` defaults to `SQLConf.get.ansiEnabled`, so if you do add it, the ANSI overflow path needs an `expect_error` case of its own.

docs/source/user-guide/latest/expressions.md
```diff
@@ -284,6 +284,8 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `timestamp_micros` | ✅ |  |
 | `timestamp_millis` | ✅ |  |
 | `timestamp_seconds` | ✅ |  |
+| `timestampadd` | ✅ | Runs through the JVM codegen dispatcher |
```
mbutrovich: `timestampadd` and `timestampdiff` are not the only way to build these nodes. The grammar routes `TIMESTAMPADD | DATEADD | DATE_ADD` to `TimestampAdd` and `TIMESTAMPDIFF | DATEDIFF | DATE_DIFF` (plus `TIMEDIFF` on 4.0) to `TimestampDiff` whenever the first argument is a `datetimeUnit` keyword (`SqlBaseParser.g4:939-940` on branch-3.5, `:1158-1159` on branch-4.0). So `dateadd(DAY, 1, ts)` is a `TimestampAdd`, not a `DateAdd`, and before this PR it fell back.

That makes four existing rows in `docs/source/user-guide/latest/expressions.md` misleading: `date_add` (`:257`), `date_diff` (`:258`), `dateadd` (`:264`) and `datediff` (`:265`) all claim `Native`, which is true only of the two-argument form. `timediff` has no row at all. Since this PR is what makes the unit form work, it is the right place to correct those and to add at least one query per fixture using an alias spelling, so the parse path is proven to land on the same serde.

Worth noting so nobody adds a test that cannot fail: the string-literal unit form is rejected at parse time, not runtime. `visitTimestampadd` throws `invalidDatetimeUnitError` when `ctx.invalidUnit` is set (Spark `AstBuilder.scala:6012-6021`), so `timestampadd('MONTH', 1, ts)` never reaches Comet and needs no fixture.

docs/source/user-guide/latest/expressions.md
```diff
@@ -284,6 +284,8 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
...
+| `timestampdiff` | ✅ | Runs through the JVM codegen dispatcher |
```
mbutrovich: The `expressions.md` hunk here adds three-column rows, but main gained an Implementation column in 376676c7a, which is the conflict. That column is generated by `GenerateDocs` from the serde's trait mixins, so please regenerate with `dev/generate-release-docs.sh` rather than hand-merging the rows. The value will not be uniform across profiles: `TimestampAdd` and `TimestampDiff` are absent from the `FunctionRegistry` on 3.4 and 3.5 (reachable only through the grammar above) and are `registerInternalExpression` on 4.0 (`FunctionRegistry.scala:927-928`), and unregistered names get the placeholder.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
```diff
@@ -0,0 +1,58 @@
...
+query
+SELECT
+  timestampadd(YEAR, 1, ts),
+  timestampadd(QUARTER, 1, ts),
+  timestampadd(WEEK, 2, ts),
+  timestampadd(DAY, -10, ts),
+  timestampadd(MINUTE, 90, ts),
+  timestampadd(SECOND, 30, ts),
+  timestampadd(MICROSECOND, 500, ts)
+FROM test_timestampadd
```
mbutrovich, line 56: `DateTimeUtils.timestampAdd` handles MICROSECOND, MILLISECOND, SECOND, MINUTE, HOUR, DAY, DAYOFYEAR, WEEK, MONTH, QUARTER, YEAR (Spark `DateTimeUtils.scala:670-708`). This block covers all of them except MILLISECOND and DAYOFYEAR, and both are valid grammar keywords on every supported version (`SqlBaseParser.g4:931-935` on branch-3.5, `:1150-1154` on branch-4.0), so both are testable as written. DAYOFYEAR is worth adding because it is an alias arm sharing a case with DAY (`case "DAY" | "DAYOFYEAR"`, `:687`), so nothing currently proves the alias both parses and dispatches.

spark/src/test/resources/sql-tests/expressions/datetime/timestampdiff.sql
```diff
@@ -0,0 +1,51 @@
...
+query
+SELECT
+  timestampdiff(YEAR, a, b),
+  timestampdiff(MONTH, a, b),
+  timestampdiff(WEEK, a, b),
+  timestampdiff(DAY, a, b),
+  timestampdiff(HOUR, a, b),
+  timestampdiff(MINUTE, a, b),
+  timestampdiff(SECOND, a, b)
+FROM test_timestampdiff
```
mbutrovich, line 45: `timestampDiffMap` covers MICROSECOND, MILLISECOND, SECOND, MINUTE, HOUR, DAY, WEEK, MONTH, QUARTER, YEAR (Spark `DateTimeUtils.scala:719-730`). This block omits MICROSECOND, MILLISECOND and QUARTER. QUARTER is the one entry in that map with arithmetic of its own (`ChronoUnit.MONTHS.between(...) / 3`, integer division toward zero), so it is both the most likely to diverge and the only one untested.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
```diff
@@ -0,0 +1,58 @@
...
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+-- timestampadd runs through the codegen dispatcher so results match Spark exactly.
+-- Config: spark.sql.session.timeZone=America/Los_Angeles
```
mbutrovich, line 20: Same applies to timestampdiff.sql:

Pinning `America/Los_Angeles` only pays off if a fixture straddles a transition. Both functions do calendar arithmetic on local time, not elapsed time: timestampadd goes through `timestampAddInterval`, which is `.atZone(zoneId).plusDays(...).plus(micros)` (Spark `DateTimeUtils.scala:269-274`), and timestampdiff converts both sides with `getLocalDateTime` and then calls `ChronoUnit.X.between` on the local values (`:745-747`). So `timestampdiff(HOUR, timestamp'2024-03-09 12:00:00', timestamp'2024-03-10 12:00:00')` is 24 despite 23 real hours elapsing, and `timestampadd(DAY, 1, timestamp'2024-03-09 12:00:00')` is a local plus-one-day rather than plus-24-hours. Those are precisely the results a chrono-based native implementation gets wrong, which is the stated reason for choosing the dispatcher, and neither fixture asserts them. Please add a spring-forward row, a fall-back row (2024-11-03 01:30 is ambiguous), and one landing on the nonexistent local hour, `timestampadd(HOUR, 1, timestamp'2024-03-10 01:30:00')`.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
```diff
@@ -0,0 +1,58 @@
...
+  timestampadd(SECOND, 30, ts),
+  timestampadd(MICROSECOND, 500, ts)
+FROM test_timestampadd
+
+-- literal arguments (constant folding is disabled by the test suite)
+query
+SELECT
+  timestampadd(HOUR, 3, timestamp'2024-01-01 10:00:00'),
+  timestampadd(MONTH, 1, timestamp'2024-01-31 00:00:00'),
+  timestampadd(YEAR, 1, timestamp'2024-02-29 00:00:00')
```
mbutrovich, line 108: `DateTimeUtils.timestampAdd` wraps its whole body in a try that maps ArithmeticException and DateTimeException to `timestampAddOverflowError` (Spark `DateTimeUtils.scala:709-713`), which raises error class `DATETIME_OVERFLOW` (`QueryExecutionErrors.scala:2484-2492`). Nothing here reaches it, and this is the only path in the PR where an exception has to cross out of the generated kernel and surface as the same Spark error. Add `query expect_error(DATETIME_OVERFLOW)` over something like `timestampadd(YEAR, 1000000000, timestamp'2024-01-15 10:30:45')`; the existing valid-input blocks already satisfy the sentinel requirement.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
```diff
@@ -0,0 +1,58 @@
...
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+-- timestampadd runs through the codegen dispatcher so results match Spark exactly.
+-- Config: spark.sql.session.timeZone=America/Los_Angeles
+-- Config: spark.comet.exec.scalaUDF.codegen.enabled=true
+
+statement
+CREATE TABLE test_timestampadd(ts timestamp, q int) USING parquet
```
mbutrovich: `TimestampAdd.dataType` is `timestamp.dataType` (Spark `datetimeExpressions.scala:3443`), so an NTZ input yields an NTZ output. That is a different Arrow vector class in the kernel (`TimeStampMicroVector` instead of `TimeStampMicroTZVector`, `CometBatchKernelCodegen.scala:71-72`), hence a separately compiled kernel, and `zoneIdForType` resolves to UTC instead of the session zone. This is a distinct code path rather than a variation, so please add an NTZ column here. For timestampdiff, `inputTypes` is `Seq(TimestampType, TimestampType)` (`datetimeExpressions.scala:3531`) so NTZ gets cast up; one case is enough to lock that in.

mbutrovich (commented):
First round, thanks @andygrove!

docs/source/user-guide/latest/expressions.md
andygrove: Corrected. All four rows now note that the `Native` implementation applies only to the two-argument form, and `timediff`, `timestampadd` and `timestampdiff` rows are added.

Each fixture now proves the parse path. One caveat on the grammar: 3.4 has only `TIMESTAMPADD | DATEADD` and `TIMESTAMPDIFF | DATEDIFF`; `DATE_ADD` and `DATE_DIFF` joined the rule in 3.5. Keeping them inline failed 3.4 with `UNRESOLVED_COLUMN.WITH_SUGGESTION` on `DAY`, since `date_add` fell through to the registered two-argument function. So the coverage is split by introducing version: `dateadd`/`datediff` inline (3.4+), `date_add`/`date_diff` in `date_add_unit_alias.sql` (`MinSparkVersion: 3.5`), `timediff` in `timediff.sql` (4.0+).

Noted on the string-literal unit form being a parse-time rejection; no fixture added for it.

docs/source/user-guide/latest/expressions.md
andygrove: Rebased onto main, so the rows now carry the Implementation column.

On regenerating: I ran `GenerateDocs` to check my values rather than hand-guessing, and it agrees on every row this PR touches. `make_interval` resolves to `Codegen dispatch`; `timestampadd`, `timestampdiff` and `timediff` all get the em-dash placeholder, as you predicted, since `registerInternalExpression` writes to the `internal` registry rather than `FunctionRegistry.expressions`, which is what `buildFunctionNameToKind` iterates. So the value is uniform across profiles here, just uniformly the placeholder.

I did not commit the generator output. On this checkout it also flipped about eight unrelated rows (`hour`, `minute`, `second`, `make_timestamp` to the placeholder; `<<`, `>>`, `>>>` to `Native`), which looks like main was last generated from a different profile tail. That churn does not belong in this PR, so the committed `expressions.md` diff is just the rows I edited. Happy to fold the regeneration in as a separate commit if you would rather have main resynced.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
andygrove: Added. `MILLISECOND` and `DAYOFYEAR` are in the all-units block now, with a comment noting `DAYOFYEAR` shares the `case "DAY" | "DAYOFYEAR"` arm so the alias is proven to parse and dispatch.

spark/src/test/resources/sql-tests/expressions/datetime/timestampdiff.sql
andygrove: Added `MICROSECOND`, `MILLISECOND` and `QUARTER` in a second block, with `QUARTER` also exercised on literals. Agreed it was the risky one, being the only entry with its own arithmetic.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
andygrove: Added to both fixtures. `timestampadd.sql` has a spring-forward block, a fall-back block using the ambiguous `2024-11-03 01:30`, and `timestampadd(HOUR, 1, timestamp'2024-03-10 01:30:00')` landing on the nonexistent local hour, plus the plus-one-day versus plus-24-hours pairing on both transitions. `timestampdiff.sql` has the matching cases, including `timestampdiff(HOUR, ...)` across spring-forward reporting 24 despite 23 elapsed hours.

The comments in both files spell out why: local-time calendar arithmetic via `timestampAddInterval` on one side and `getLocalDateTime` plus `ChronoUnit.X.between` on the other.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
andygrove: Added: `query expect_error(DATETIME_OVERFLOW)` over `timestampadd(YEAR, 1000000000, timestamp'2024-01-15 10:30:45')`, which overflows in `Math.multiplyExact(quantity, MONTHS_PER_YEAR)`. Confirmed it passes on 3.4, 3.5, 4.0 and 4.1, so the exception does cross out of the generated kernel with the condition name intact.

Worth recording for anyone adding a similar case: this is version-sensitive. The `make_interval` ANSI overflow I added in the same pass could not match on `ARITHMETIC_OVERFLOW`, because 3.5 renders that one as a bare `"integer overflow. If necessary set ..."` with no condition prefix. That fixture matches on `overflow` instead. `DATETIME_OVERFLOW` happens to carry its prefix on every supported version.

spark/src/test/resources/sql-tests/expressions/datetime/timestampadd.sql
andygrove: Added an NTZ column to `timestampadd.sql` with its own query block covering `HOUR`, `MONTH`, `DAY` and `MICROSECOND`, and an NTZ block in `timestampdiff.sql` to lock in the cast-up behavior.

andygrove:
Rebased onto main and addressed both rounds. Replies are inline on each thread; summarizing the cross-cutting items here.

**`make_interval` is now in this PR** rather than deferred, since @mbutrovich is right that #4898 makes the stated reason stale. It is `object CometMakeInterval extends CometCodegenDispatch[MakeInterval]` plus the `classOf[MakeInterval]` entry, exactly as you both described, and the PR title and description are updated.

@peterxcli, on the native route: I looked for `SparkMakeInterval` and found no `make_interval` implementation anywhere under `native/`, so I went with codegen dispatch to match the rest of this PR. If you were thinking of an upstream `datafusion-spark` function, a native path would be a reasonable follow-up, but it would want its own PR given the CalendarInterval ANSI semantics.

Per @mbutrovich's note, the ANSI overflow path has its own coverage in `make_interval_ansi.sql`, where `failOnError` is true, alongside `make_interval.sql` for the non-ANSI path where the same input returns NULL instead.

**A pre-existing bug surfaced while writing those tests**, filed as #5058. A null `CalendarIntervalType` literal in a projection throws `CometNativeException: Interval(MonthDayNano) is not supported in Comet` instead of falling back or evaluating. #4898 added the type to the support checks, codegen kernels and serialization, but the null-literal match in `planner.rs` has no `Interval` arm, so the plan is accepted and then fails natively. `YearMonthIntervalType` is not in the supported set and falls back cleanly, which is the behavior a null calendar interval should have at minimum.

It is easy to hit indirectly: `MakeInterval` is `NullIntolerant`, so `NullPropagation` rewrites `make_interval(NULL, 2, 3, ...)` into a null interval literal and the expression vanishes before Comet sees it. Not caused by this PR, and since literal NULL arguments never exercise the expression anyway, the fixture drives its NULL coverage from nullable columns and points at #5058.

@peterxcli, on the non-blocking suggestions: `MILLISECOND` is in both fixtures, `MICROSECOND` is in `timestampdiff`, and both DST boundaries and `TIMESTAMP_NTZ` are covered rather than just one of them.

Verified green on all four supported profiles (3.4, 3.5, 4.0, 4.1). One thing worth flagging from that sweep: the alias coverage had to be split by Spark version, because 3.4's grammar rule is only `TIMESTAMPADD | DATEADD` and `TIMESTAMPDIFF | DATEDIFF`, with `DATE_ADD` and `DATE_DIFF` joining in 3.5. Details are in the thread on `expressions.md`.

peterxcli (approved):
@andygrove thanks for the update, LGTM

> on the native route: I looked for SparkMakeInterval and found no make_interval implementation anywhere under native/, 

yes, the native implementation would be handled in https://github.com/apache/datafusion-comet/pull/5039 + https://github.com/apache/datafusion-comet/issues/5131

> so I went with codegen dispatch to match the rest of this PR. 

yes, make sense, codegen dispatch support minimize the change in this PR as first step.

mbutrovich (approved):
I think this LGTM after revision, thanks @andygrove!

andygrove:
Merged. Thanks @mbutrovich

Issue #3232, opened 2026-01-21
to_csv: Handle edge cases found during fuzz testing


andygrove:
## Summary

During review of PR #3004 (which adds basic `to_csv` support), fuzz testing revealed several edge cases that are not handled correctly. These should be addressed in follow-up work after the initial implementation is merged.

### 1. Null value not quoted when it contains special characters
When the `nullValue` option contains the delimiter or other special characters (e.g., `"N,A"`), it's written unquoted, corrupting the CSV output.

| Expected (Spark) | Actual (Comet) |
|------------------|----------------|
| `"N,A",world` | `N,A,world` |
| `hello,"N,A"` | `hello,N,A` |

**Location:** `native/spark-expr/src/csv_funcs/to_csv.rs:164-171`

**Fix:** Check if `null_value` contains special characters and quote/escape it appropriately.

### 2. Whitespace trimming applied incorrectly  
When `ignoreLeadingWhiteSpace=false` or `ignoreTrailingWhiteSpace=false`, strings containing whitespace plus special characters are incorrectly handled. The code trims whitespace before checking if quoting is needed.

| Expected (Spark) | Actual (Comet) |
|------------------|----------------|
| `  \"` (preserved whitespace with escaped quote) | `""` (empty) |

**Location:** `native/spark-expr/src/csv_funcs/to_csv.rs:176-183`

**Fix:** Review the order of operations - quoting determination should consider the original (untrimmed) value.

### 3. Decimal formatting mismatch
Spark uses scientific notation for small decimal values, while Comet uses fixed-point notation.

| Expected (Spark) | Actual (Comet) |
|------------------|----------------|
| `0E-18` | `0.000000000000000000` |

**Fix:** Align decimal-to-string casting with Spark's formatting behavior.

### 4. NPE with single-column struct (needs investigation)
`NullPointerException` occurs when processing single-column structs with certain null patterns. This may be a Spark-side issue with how Comet's output is handled, but needs investigation.

## Reproduction

Fuzz tests were added in `CometCsvExpressionSuite.scala` that reproduce these issues:
- `to_csv - edge case: delimiter in null value representation`
- `to_csv - fuzz test: comprehensive random data and options`  
- `to_csv - edge case: numeric boundary values`
- `to_csv - edge case: single column struct`

## Related

- PR #3004 - Initial `to_csv` implementation

andygrove:
## Fuzz Tests for Reproduction

[...]

PR #5032, opened 2026-07-24
feat: route translate and to_csv through codegen dispatch by default

andygrove:

Closes #.

## Rationale for this change

`translate` (StringTranslate) and `to_csv` (StructsToCsv) have native implementations that are incompatible with Spark by default, so today the whole enclosing operator falls back to Spark unless allowIncompatible is set. Routing them through the JVM codegen dispatcher instead keeps the operator on the native pipeline while producing bit-exact Spark results, and preserves the faster native path as an opt-in.

- Route `translate` (StringTranslate) and `to_csv` (StructsToCsv) through the JVM codegen dispatcher by default, using the same `NativeOptInAvailable` pattern as `to_json` and `split`. The dispatcher runs Spark's own generated code (bit-exact) while keeping the operator on the native pipeline; the native path stays available via `allowIncompatible`.
- Update the docs to note the codegen-dispatch default for `to_csv`.

- Update the `string_translate` SQL test (previously asserting fallback) to verify native codegen dispatch.
- Add a `to_csv` SQL file test.


spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -109,15 +109,38 @@ object CometOctetLength extends CometScalarFunction[OctetLength]("octet_length")
...
+    }
+
+  override def convert(
+      expr: StringTranslate,
+      inputs: Seq[Attribute],
+      binding: Boolean): Option[Expr] = {
+    if (CometConf.isExprAllowIncompat(getExprConfigName(expr))) {
+      val childExprs = expr.children.map(exprToProtoInternal(_, inputs, binding))
+      val optExpr = scalarFunctionExprToProto("translate", childExprs: _*)
+      optExprWithFallbackReason(optExpr, expr, expr.children: _*)
```
mbutrovich: First, lines 134-136 are a verbatim copy of `CometScalarFunction.convert` (`spark/src/main/scala/org/apache/comet/serde/CometScalarFunction.scala:29-33`), so the function name `"translate"` now lives in two places. Dropping the `CometScalarFunction[StringTranslate]("translate")` base to hand-inline its body is also a step away from the established idiom for this exact case, which keeps the base and overrides `convert` to choose a branch: `CometInitCap` (`strings.scala:162-174`), `CometStringReplace` (`:192-199`), `CometCaseConversionBase` (`:67-78`).

Second, and more to the point: the whole `getSupportLevel` plus `convert` branching structure is what `CodegenDispatchFallback` exists to express declaratively. `QueryPlanSerde.exprToProtoInternal` already routes a non-opted-in `Incompatible` result from such a serde through `emitJvmCodegenDispatch` and falls back to Spark only when the dispatcher refuses (`QueryPlanSerde.scala:838-870`, `dispatchIfFallback` at `:1109-1117`), and `CodegenDispatchFallback extends NativeOptInAvailable` so `GenerateDocs` renders the same compatible-by-default row (`GenerateDocs.scala:193`). So the entire diff to this object should be one mixin on the original:

```scala
object CometStringTranslate
    extends CometScalarFunction[StringTranslate]("translate")
    with CodegenDispatchFallback {
  private val incompatReason = ...
  override def getIncompatibleReasons(): Seq[String] = Seq(incompatReason)
  override def getSupportLevel(expr: StringTranslate): SupportLevel = Incompatible(Some(incompatReason))
}
```

That keeps the existing `Incompatible` declaration honest, keeps the function name in one place, and gets the `logWarning` on the opted-in path (`QueryPlanSerde.scala:841-846`) that the hand-rolled version drops.

spark/src/main/scala/org/apache/comet/serde/structs.scala
```diff
@@ -259,50 +259,50 @@ object CometJsonToStructs extends CometCodegenDispatch[JsonToStructs] with Nativ
...
+        val optionsProto = options2Proto(expr.options, expr.timeZoneId)
+        val toCsv = ExprOuterClass.ToCsv
+          .newBuilder()
+          .setChild(childProto)
+          .setOptions(optionsProto)
+          .build()
+        ExprOuterClass.Expr.newBuilder().setToCsv(toCsv).build()
+      }
+    } else {
+      // Default: route through the codegen dispatcher so Spark's own doGenCode runs inside the
```
mbutrovich: The same argument applies here, and more strongly because the pre-existing `getSupportLevel` already returned exactly the right classifications: `Unsupported` for complex field types, `Incompatible` for the #3232 types, `Incompatible()` otherwise. With `with CodegenDispatchFallback` on line 262 those three outcomes route as intended with no other change: `Unsupported` to the dispatcher plus a `logDebug` (`QueryPlanSerde.scala:827-833`), non-opted-in `Incompatible` to the dispatcher plus the `[COMET-INFO]` opt-in hint (`:854-860`), opted-in `Incompatible` to the native proto (`:847`). That turns a +30/-30 rewrite into a one-line change and removes the duplicated `nativeSupported(expr)` test that currently has to be kept in sync between `:279` and `:290`.

Deleting `getUnsupportedReasons` (was `structs.scala:270-271`) also loses real information. The native path still cannot do complex types, and under `NativeOptInAvailable` the documented surface for native-path limitations is `getIncompatibleReasons` (`spark/src/main/scala/org/apache/comet/serde/CometExpressionSerde.scala:52-63`). Keeping `CodegenDispatchFallback` preserves both accessors as-is; if you keep the rewrite instead, the complex-type limitation needs folding into `getIncompatibleReasons` at `:266-268`, otherwise a user who sets `allowIncompatible` has no way to find out why they did not get the native path.

spark/src/main/scala/org/apache/comet/serde/structs.scala
```diff
@@ -259,50 +259,50 @@ object CometJsonToStructs extends CometCodegenDispatch[JsonToStructs] with Nativ
...
-      childProto <- exprToProtoInternal(expr.child, inputs, binding)
-    } yield {
-      val optionsProto = options2Proto(expr.options, expr.timeZoneId)
-      val toCsv = ExprOuterClass.ToCsv
-        .newBuilder()
-        .setChild(childProto)
-        .setOptions(optionsProto)
-        .build()
-      ExprOuterClass.Expr.newBuilder().setToCsv(toCsv).build()
+    if (CometConf.isExprAllowIncompat(getExprConfigName(expr)) && nativeSupported(expr)) {
```
mbutrovich: When `isExprAllowIncompat` is true and `nativeSupported` is false, the user asked for the native path and silently gets the dispatcher with nothing recorded. `CometStructsToJson` has the same hole so this is not new, but it deserves a `logDebug` naming the disqualifying field type. Using `CodegenDispatchFallback` gets this for free through the `Unsupported` arm.

spark/src/test/resources/sql-tests/expressions/string/string_translate.sql
```diff
@@ -15,29 +15,29 @@
...
+-- native path diverges from Spark (DataFusion iterates over Unicode graphemes where Spark uses code
+-- points, and substitutes U+0000 instead of treating it as a deletion sentinel), so it is opt-in
+-- via spark.comet.expression.StringTranslate.allowIncompatible. See string_translate_enabled.sql
+-- for the opt-in native path.
 
 statement
 CREATE TABLE test_translate(s string, from_str string, to_str string) USING parquet
 
 statement
 INSERT INTO test_translate VALUES ('hello', 'el', 'ip'), ('hello', 'aeiou', '12345'), ('', 'a', 'b'), (NULL, 'a', 'b'), ('hello', '', ''), ('abc', 'abc', 'x')
```
mbutrovich: The four `query` blocks do assert native execution, so flipping them off `expect_fallback` is real new coverage of the routing. What is missing is the payload: the header comment at `:18-22` names two specific ways the native path diverges, and the six rows on line 28 are ASCII-only and hit neither. The file is now identical to `string_translate_enabled.sql` in CREATE TABLE, INSERT rows and queries, differing only in the config header, so nothing distinguishes the dispatcher from the native path here. Please add the two rows that are only assertable now that the default is bit-exact: a combining-mark input for the grapheme versus code-point difference, and a `to` argument containing U+0000 for the deletion-sentinel difference. To keep the fixture ASCII, build them with `decode(X'65CC81', 'UTF-8')` and `decode(X'00', 'UTF-8')` rather than embedding the characters. Those two rows are the regression test for this PR.

spark/src/main/scala/org/apache/comet/serde/structs.scala
```diff
@@ -259,50 +259,50 @@ object CometJsonToStructs extends CometCodegenDispatch[JsonToStructs] with Nativ
...
-      return Incompatible(
-        Some(
-          s"The schema ${expr.inputSchema} is not supported because " +
-            s"it includes a incompatible data types: $incompatibleDataTypes"))
-    }
-    // https://github.com/apache/datafusion-comet/issues/3232
-    Incompatible()
+    !dataTypes.exists(DataTypeSupport.isComplexType) &&
+    !dataTypes.exists(incompatibleDataTypes.contains)
   }
```
mbutrovich: `nativeSupported` is the new branch predicate driving both `getSupportLevel` (`:279`) and `convert` (`:290`), and no test takes it false. The two cases it excludes are the reason the change exists:

Date, Timestamp, TimestampNTZ and Binary fields were `Incompatible` before this PR and fell the operator back to Spark. That is #3232.

Complex field types were `Unsupported` before and also fell back. They are legal `to_csv` input: `StructsToCsv.checkInputDataTypes` accepts arrays, maps and nested structs and rejects only Variant (Spark `csvExpressions.scala:240-261` on master and branch-4.0; branch-3.5 has no gate at all). Nothing tests that the kernel can hand `converter()` a nested `InternalRow`, which is the new and untried part of this path.

The cheapest place to cover the first group is the existing suite rather than a new fixture. `CometCsvExpressionSuite`'s test named "to_csv - default options" (`spark/src/test/scala/org/apache/comet/CometCsvExpressionSuite.scala:36`) is actually run under `allowIncompatible=true` (`:50`), and it selects c0-c5, c7, c8, c9 and c12 out of the 14-column fuzz schema, skipping exactly c6 Double, c10 Timestamp, c11 TimestampNTZ and c13 Binary (`FuzzDataGenerator.scala:303-317`). Those are the columns the native path cannot match Spark on. After this PR the dispatcher matches them by construction, so that test can drop the `allowIncompatible` wrapper and add those four columns, which is what would actually justify the name it already has.

spark/src/test/resources/sql-tests/expressions/csv/to_csv.sql
```diff
@@ -0,0 +1,41 @@
...
+INSERT INTO test_to_csv VALUES
+  (1, 'x', 2.5),
+  (-3, 'hello,world', 0.0),
+  (0, 'has "quote"', -1.5),
+  (NULL, NULL, NULL),
+  (7, '', 3.0)
+
+-- column struct: values with delimiters and quotes exercise Spark's CSV quoting rules
+query
+SELECT to_csv(named_struct('a', a, 'b', b, 'c', c)) FROM test_to_csv
```
mbutrovich, line 35: `options` is the main axis of behavior difference for this expression and nothing in the new fixture varies it. `CometCsvExpressionSuite:71-161` covers delimiter, escape, quoteAll and nullValue but only under `allowIncompatible=true`, so the dispatcher path has no options coverage anywhere. `to_csv(s, map('sep', ';'))` plus a `timestampFormat` over a timestamp field would cover both this and the #3232 types in one query.

mbutrovich (commented):

spark/src/main/scala/org/apache/comet/serde/strings.scala
andygrove: You are right on both counts, and the mixin does the whole job. Done in 26994c919 — the entire change to this object is now `with CodegenDispatchFallback` on the original declaration, exactly as you wrote it. The `CometScalarFunction[StringTranslate]("translate")` base is restored, so `"translate"` lives in one place again, and the opted-in `logWarning` at `QueryPlanSerde.scala:841-846` comes back with it. The pre-existing `Incompatible(Some(incompatReason))` was already the right classification, so it stays untouched.

spark/src/main/scala/org/apache/comet/serde/structs.scala
andygrove: Agreed — reverted to the original three-way `getSupportLevel` plus the mixin in 26994c919. The +30/-30 rewrite is now one mixin and a comment.

Confirming your reading of the routing, since it is the whole basis for the change: `Unsupported` (complex field types) reaches the dispatcher plus `logDebug` at `QueryPlanSerde.scala:819-831`, non-opted-in `Incompatible` (#3232 types, and the `Incompatible()` catch-all) reaches the dispatcher plus the `[COMET-INFO]` hint at `:848-860`, and opted-in `Incompatible` reaches `handler.convert` — the native proto — at `:846`. So all three original outcomes route as intended with no other change.

`getUnsupportedReasons` is restored, and the duplicated `nativeSupported` predicate is gone along with the sync hazard between the two call sites. I also dropped `CometCodegenDispatch` as the base: it hardcodes `getSupportLevel = Compatible()` and `convert = emitJvmCodegenDispatch` for expressions with no native path at all, so overriding `convert` to sometimes go native was working against it.

spark/src/main/scala/org/apache/comet/serde/structs.scala
andygrove: Resolved by the redesign rather than by adding a `logDebug`: with `CodegenDispatchFallback` this case lands in the framework's `Unsupported` arm, which already logs at debug naming the disqualifying reason (`QueryPlanSerde.scala:819-831`). So it is covered for free, as you noted.

spark/src/test/resources/sql-tests/expressions/string/string_translate.sql
andygrove: Added in 26994c919, built with `decode()` as you suggested so the fixture stays ASCII:

```sql
(concat('caf', decode(X'65CC81', 'UTF-8')), 'e', 'E'),   -- "e" + U+0301: one grapheme, two code points
('hello', 'l', decode(X'00', 'UTF-8'))                 -- U+0000 as the `to` argument
```

I also checked that they actually discriminate rather than passing vacuously, since that is the point of them. Adding the same two rows to `string_translate_enabled.sql`, which opts into the native path, fails:

```
- sql-file: expressions/string/string_translate_enabled.sql *** FAILED ***
  Error executing SQL 'SELECT translate(s, from_str, to_str) FROM test_translate_enabled'
  Results do not match for query
```

So the rows pass under the dispatcher and fail under the native path, which makes them a real regression test for the routing. I left `string_translate_enabled.sql` as it was — its header explicitly says it covers inputs where the two implementations agree, so the divergent rows belong in the default file only.

spark/src/test/resources/sql-tests/expressions/csv/to_csv.sql
andygrove: Added in 26994c919. `to_csv.sql` now covers `map('sep', ';')` over the existing rows (which also changes which values need quoting), plus a `timestampFormat`/`dateFormat` query over a new date+timestamp table — so the same query exercises both the options axis and the #3232 types that previously fell back. The dispatcher path had no options coverage anywhere before this.

spark/src/main/scala/org/apache/comet/serde/structs.scala
andygrove: Took your suggestion for the first group and it works: in 26994c919 `CometCsvExpressionSuite`'s "to_csv - default options" drops the `allowIncompatible=true` wrapper and now selects the whole fuzz schema rather than a hand-picked subset, so it covers **c6 Double, c10 Timestamp and c11 TimestampNTZ**. It finally tests what its name says.

**c13 Binary I had to leave out, and not because of a divergence.** Spark's CSV converter renders `BinaryType` with Java's default `Object.toString()`, so a row comes out as:

```
...,3128558622871195782,[B@731af74
```

That trailing token is an identity hash of the byte-array instance, so it differs between any two evaluations. With c13 included, every row mismatches and the *only* differing token is that hash — the double, timestamp and TimestampNTZ columns all match exactly:

```
![false,-1,-28684,...,3332-12-27T13:01:58.882-08:00,3332-12-15T18:12:14.824,...,[B@731af74]
 [false,-1,-28684,...,3332-12-27T13:01:58.882-08:00,3332-12-15T18:12:14.824,...,[B@72f0225a]
```

So the value is not assertable by any engine, including Spark against itself. I noted that in the test rather than silently dropping the column.

**On the complex-type group, the news is worse and I could not write the test.** You are right that `checkInputDataTypes` accepts arrays, maps and nested structs, so they reach the converter — but Spark's output for them is not a comparable value. Probed against Spark 3.5 with Comet disabled entirely (`spark.comet.enabled=false`), so this is pure Spark:

- non-null values render as Java identity strings: `org.apache.spark.sql.vectorized.ColumnarArray@1ada50f0`, `ColumnarRow@4dd6f621`, `ColumnarMap@71e38661`
- **any null complex value throws `NullPointerException`** inside `UnsafeWriter.write(UnsafeWriter.java:110)`, with no Comet frames in the stack

So `to_csv` over complex types produces either an identity hash or a crash in Spark itself. There is no value-parity assertion to make, and I did not want to add a fixture that pins garbage. I documented the finding in `to_csv.sql` with the reasoning and the fact that it is Spark behavior.

Worth noting this is not a regression from the PR: before the change these schemas were `Unsupported` and fell the projection back to Spark, and now they reach the dispatcher, which runs the same Spark `doGenCode`. The NPE and the identity strings are identical either way — the change only decides whether the *rest* of the projection stays native. Happy to file a separate issue against Spark's `to_csv` for the complex-type behavior if you think it is worth tracking on our side.

andygrove:
@mbutrovich all six threads addressed in 26994c919, details inline. You were right about the central point, and taking it shrank the PR considerably.

**The whole serde diff is now two mixins.** `CodegenDispatchFallback` already expresses exactly this routing, and both objects' original `getSupportLevel` declarations were already the correct classifications, so the hand-rolled branching is gone:

- `CometStringTranslate` is back to `CometScalarFunction[StringTranslate]("translate")` plus the mixin — `"translate"` lives in one place again, and the opted-in `logWarning` comes back.
- `CometStructsToCsv` is back to the original three-way `getSupportLevel` plus the mixin. `getUnsupportedReasons` is restored, the duplicated `nativeSupported` predicate and its two-call-site sync hazard are gone, and the missing `logDebug` you flagged separately comes for free from the framework's `Unsupported` arm. I also dropped `CometCodegenDispatch` as the base, since it is for expressions with no native path at all.

**Tests.** `string_translate.sql` gains the two rows that actually distinguish the paths, and I verified they discriminate rather than pass vacuously: adding them to `string_translate_enabled.sql`, which opts into the native path, fails. `to_csv.sql` gains `sep` and `timestampFormat`/`dateFormat` coverage. `CometCsvExpressionSuite`'s "to_csv - default options" drops the `allowIncompatible` wrapper and now covers c6 Double, c10 Timestamp and c11 TimestampNTZ.

**Two things I could not assert, both Spark behavior — verified with `spark.comet.enabled=false`:**

- **c13 Binary:** Spark's CSV converter renders it via Java's default `toString()`, e.g. `[B@731af74`. With c13 included, every row mismatches and the only differing token is that identity hash — the double and both timestamp columns match exactly. Not assertable by any engine, Spark included.
- **Complex types:** non-null values render as `org.apache.spark.sql.vectorized.ColumnarArray@1ada50f0` / `ColumnarRow@...` / `ColumnarMap@...`, and **any null complex value throws `NullPointerException`** in `UnsafeWriter.write`, with no Comet frames. So `to_csv` over complex types is either an identity hash or a crash in Spark itself, and there is no value-parity test to write. I documented this in the fixture instead of pinning garbage.

Neither is a regression here: those schemas previously fell back to Spark and now reach the dispatcher, which runs the same `doGenCode`, so the observable result is identical — the change only decides whether the rest of the projection stays native. Say the word if you want a separate issue tracking the complex-type behavior.

**Verified on spark-3.5:** `to_csv.sql`, `string_translate.sql`, `string_translate_enabled.sql` and both `CometCsvExpressionSuite` tests pass; spotless and scalastyle clean.

One unrelated note: this branch is behind `main`, which has since added an **Implementation** column to `expressions.md`. The two Notes cells this PR changes are still accurate, but the docs table will need regenerating after a rebase.

spark/src/test/resources/sql-tests/expressions/csv/to_csv.sql
mbutrovich, line 1: Now that `to_csv` routes through the dispatcher by default for a struct built from `named_struct`, would it be worth adding a case where one of the struct's fields is itself a cast that can raise under ANSI, for example `to_csv(named_struct('a', CAST(bad_str AS INT)))`? #5219 found and fixed a bug in the shared dispatcher for exactly this shape, and it merged after this branch's last commit, so there is not yet a test here that pins the interaction for `to_csv`.

Would it also be worth adding a query that runs `to_csv` over a struct with an array, map, or nested-struct field, even without asserting the exact output value? The comment block above explains well why Spark's own output for complex types is not a stable value to assert, but that is a different question from whether Comet's dispatcher path can hand a nested `InternalRow`/`ArrayData`/`MapData` through `CometSpecializedGettersDispatch` without hitting an unhandled exception of its own. That path is not exercised by any test in this PR, and it is the same class of gap #5219 found for TIME types.

andygrove: Both were worth adding, and both went in as 7a7942c5 — I never came back to this thread to say so, sorry.

**The ANSI cast.** `to_csv_ansi.sql` covers exactly the shape you named, including the specific #5219 one:

```sql
query expect_error(CAST_INVALID_INPUT)
SELECT to_csv(named_struct('a', CAST(s AS INT))) FROM test_to_csv_ansi

query expect_error(CAST_INVALID_INPUT)
SELECT to_csv(named_struct('a', CAST(s AS INT), 'b', i)) FROM test_to_csv_ansi
```

The second is the #5219 configuration — two input ordinals where the non-cast one is NULL on the row whose cast raises. Worth noting what I found writing it: `to_csv` does not actually reach that null short-circuit today, because `CreateNamedStruct` is not `NullIntolerant` so `allNullIntolerant` already fails. So these pin a shape that is one wire-up change away from mattering rather than one that is live now. I said that in the file so the next reader does not conclude the guard is load-bearing.

There is also a `TRY_CAST` case, since the interesting inverse is that it must *not* raise and must render as the CSV nullValue.

The file needed a sentinel to be worth anything, which is the third query: an `expect_error` fixture passes vacuously if Comet silently rejects the expression and Spark raises the same error on the fallback path. That query uses `checkSparkAnswerAndOperator` on the well-formed row, so a silent dispatcher rejection fails there instead.

**Complex field types.** Also covered, and you were right that it is the same class of gap. `to_csv.sql` now runs a struct with array / map / nested-struct fields three ways — straight from the scan (so the converter's per-field reads land on `CometInternalRow`), through `named_struct` over those fields, and with complex types nested inside complex types so the converter recurses through the element getters rather than stopping at the first level.

Those assert `IS NOT NULL` rather than a value, for the reason you anticipated: on 3.4 / 3.5 `UnivocityGenerator.makeConverter` has no complex-type branch and lands on `getter.get(ordinal, dataType).toString`, an identity string whose class differs between Spark's converter input and the kernel's. That generic `get` is precisely what `CometSpecializedGettersDispatch` implements, so on those versions these queries are its coverage. `to_csv_nested.sql` asserts the real values on 4.0+, where Spark has deterministic converters.

All three fixtures pass on this head.

andygrove:
This PR is now stale and conflcits with newer PRs, so will close this one

mbutrovich (commented):
Thanks for sticking with this one, @andygrove!

mbutrovich:
Whoops I just approved it and was about to merge it. What PRs does it conflict with?

andygrove:
> Whoops I just approved it and was about to merge it. What PRs does it conflict with?

Ok, myabe I should reopen it. There is newer https://github.com/apache/datafusion-comet/pull/5829

andygrove:
reopening

mbutrovich:
> There is newer #5829

I generally favor the PR that's been through multiple reviews already rather than starting over with a new one.

mbutrovich (approved):
LGTM pending CI, thanks @andygrove!

andygrove:
Merged. Thanks @mbutrovich!

PR #5033, opened 2026-07-24
fix: make native pow implementation compatible with Spark

andygrove:

There is no dedicated tracking issue. The native path's incompatibility stems from the upstream DataFusion behavior in apache/datafusion#22598.

Closes #.

## Rationale for this change

Spark's `Pow` expression delegates to Java's `Math.pow`, which returns `Infinity` for a zero base raised to a negative exponent (`pow(0, -1)`). Comet's native path routed `pow` to DataFusion's `PowerFunc`, whose `float64_power_checked` instead **errors** on that input (apache/datafusion#22598). Because of that single edge case, `CometPow` was marked `Unsupported`, so any query using `pow`/`power` fell back to Spark for that part of the plan.

Rust's `f64::powf` follows C99 `pow` semantics, which agree with Java's `Math.pow` on almost every input, including `0^-1 = Infinity`, `(-0)^-1 = -Infinity`, the signed-zero rules, and the infinite-base rules. They diverge in exactly one place: when `|base| == 1` and the exponent is infinite or `NaN`, C99 `pow` (and thus `powf`) returns `1`, whereas `Math.pow` returns `NaN` (for example `pow(1, Infinity)` and `pow(1, NaN)`). A thin native kernel over `powf` that special-cases that one rule matches Spark exactly and lets `pow` run natively by default instead of falling back.

Asserting that behavior in a test requires comparing results exactly rather than with a tolerance, because the tolerance comparison makes no assertion at all for `NaN` and ignores the sign of `Infinity`, so it would pass vacuously on exactly the cases that matter here. Exact comparison, however, was not usable for `NaN`-producing expressions: `QueryTest.compare` compares doubles by raw bits, so it also distinguished `NaN` payloads. Those payloads are unspecified. `Math.pow` may return any `NaN`, and the payload the JVM produces differs across architectures, so Spark and a native kernel can return `NaN`s that print identically but compare unequal. Spark fixed this in `compare` itself for 4.1.2 and 4.2.0 ("in some hardware NaN can be represented with different bits, so first check for it"), leaving 3.4, 3.5, 4.0 and 4.1.1 comparing raw bits. Comet's comparison path needs the same treatment so that every supported version can assert `NaN`-ness without depending on the payload.

- Add a `spark_pow` native kernel (`native/spark-expr/src/math_funcs/pow.rs`) that computes the power element-wise with null propagation, covering all Array/Scalar argument combinations. The per-element `spark_powf` helper special-cases `|base| == 1` with a non-finite exponent to return `NaN`, matching `Math.pow`; every other input goes straight to `powf`.
- Register `"pow"` to `spark_pow` in `create_comet_physical_fun_with_eval_mode`, so it no longer falls through to DataFusion's checked `power`, plus the module/re-export wiring.
- Simplify `CometPow` to a plain `Compatible` serde (drop the `Unsupported`/"Power has correctness issues" reporting).
- Canonicalize `NaN` in `CometTestBase.normalizeForComparison`, the recursive walker that already normalizes decimals and boxed primitives through nested rows, arrays and maps before `QueryTest.compare` runs. This backports Spark's own 4.1.2 fix to the versions that lack it. Signed zero is deliberately left alone, since distinguishing `0.0` from `-0.0` is the reason the raw-bit comparison exists. Beyond this PR it closes a coverage gap for any `NaN`-producing expression: tests no longer have to choose between a tolerance that skips `NaN` entirely and an exact comparison that depends on the payload.

- Rust unit tests in `pow.rs` cover the basic cases, `pow(0, -1) == Infinity`, null propagation, the scalar argument paths, and the edge cases: `|base| == 1` with an infinite or `NaN` exponent (returns `NaN`), signed-zero bases, infinite bases and exponents, and subnormal values.
- The `expressions/math/pow.sql` SQL file test (run via `CometSqlFileTestSuite`) asserts native execution matches Spark across ordinary values, nulls, signed zero, positive and negative infinity in both base and exponent, subnormals, and the `|base| == 1` non-finite exponent case. These rows are compared exactly rather than with a tolerance, so `NaN` results and the sign of `Infinity` are asserted rather than skipped. Confirming the fix is load-bearing: with the `|base| == 1` special case disabled, the SQL test fails.
- `CometSqlFileTestSuite` passes in full (424 tests) on Spark 3.5 as well as the default 4.1 build. Spark 3.5 is one of the versions whose `QueryTest.compare` still compares raw bits, so it exercises the `NaN` canonicalization rather than Spark's own.


spark/src/test/resources/sql-tests/expressions/math/pow.sql
```diff
@@ -15,27 +15,25 @@
 -- specific language governing permissions and limitations
 -- under the License.
 
+-- pow runs natively and matches Spark exactly, including edge cases such as pow(0, -1) = Infinity.
+
 statement
 CREATE TABLE test_pow(base double, exp double) USING parquet
 
 statement
 INSERT INTO test_pow VALUES (0.0, -1), (2.0, 3.0), (0.0, 0.0), (-1.0, 2.0), (-1.0, 0.5), (2.0, -1.0), (NULL, 2.0), (2.0, NULL), (cast('NaN' as double), 2.0), (cast('Infinity' as double), 2.0), (2.0, cast('Infinity' as double))
```
kazuyukitanimura: What about negative zero cases? and subnormal numbers?

kazuyukitanimura: negative infinity can be interesting too. we probably need a bit more combinations.

andygrove: Thanks for the review @kazuyukitanimura. I added more tests as suggested and it did expose a bug, which is now fixed.

kazuyukitanimura (approved):
Thank you @andygrove

PR #5034, opened 2026-07-24
feat: native uuid() implementation compatible with Spark

andygrove:

There is no dedicated tracking issue; `uuid` was previously listed as planned (🔜) in the expression support guide.

Closes #.

## Rationale for this change

Spark's `uuid()` fell back to Spark because Comet had no native implementation. Spark generates RFC 4122 version 4 UUID strings via `org.apache.spark.sql.catalyst.util.RandomUUIDGenerator`, which seeds a Commons Math3 `MersenneTwister` with `randomSeed + partitionIndex` and, per row, draws two `nextLong()`s, masks in the version and variant bits, and formats via `java.util.UUID.toString`. Comet already ports that exact `MersenneTwister` for native `shuffle` (`SparkMersenneTwister`), so `uuid()` can run natively and produce results that match Spark bit for bit, keeping the enclosing operator on the native pipeline instead of falling back.

This change was scaffolded with the project's `implement-comet-expression` skill (and audited with `audit-comet-expression`).

- Native `UuidExpr` (`native/spark-expr/src/nondetermenistic_funcs/uuid.rs`), modeled on `ShuffleExpr`: it keeps a per-partition `SparkMersenneTwister` in a `Mutex` so state advances across batches, and produces a `Utf8` column. Added `SparkMersenneTwister::next_long()` (a port of Commons Math3 `BitsStreamGenerator.nextLong()`).
- Proto `Uuid` message and oneof field, wired through the expression registry with a `UuidBuilder` that combines the seed with the partition index (matching `rand`/`randn`/`shuffle`).
- `CometUuid` serde emitting the resolved random seed, registered in `QueryPlanSerde`.
- Support-status doc flipped from planned to supported, and a `misc_funcs` audit entry recording the cross-version findings.

The audit confirmed `RandomUUIDGenerator` is identical across Spark 3.4.3 / 3.5.8 / 4.0.1 / 4.1.1, so no shim is needed. The `uuid(seed)` SQL form exists only in Spark 4.0+, so the seeded tests are gated with `MinSparkVersion: 4.0`.

- Rust unit tests in `uuid.rs`: RFC 4122 v4 format/version/variant, determinism per seed, and per-batch state continuity.
- `expressions/misc/uuid.sql` (all versions): no-seed `uuid()` runs natively and satisfies deterministic properties (length, RFC 4122 layout, version/variant nibbles) that hold identically on Comet and Spark.
- `expressions/misc/uuid_with_seed.sql` (Spark 4.0+): seeded `uuid(seed)` asserts bit-for-bit equality with Spark across single and multiple rows, negative/zero/boundary seeds, and combined with other expressions. Verified natively on Spark 4.1.2 (default) and 3.5.8 (seeded file correctly skipped).


comphead (commented):
Thanks @andygrove the PR looks solid. 

LLM suggests couple of tests gaps

 ### Gap 1 — Multi-partition bit-for-bit (highest priority)

  **What's missing.** `test_uuid_seed` has 5 rows in one parquet file, so `planner.partition() == 0` and the `expr.seed.wrapping_add(partition)` offset is never exercised on the SQL path.

  **Why it matters.** The whole reason `RandomUUIDGenerator` is seeded with `randomSeed + partitionIndex` is to decorrelate partitions. If `planner.partition()` ever returned the wrong value (e.g. `usize`
  widening quirks, or a future refactor), the SQL suite would still pass — every partition would silently produce the same UUIDs and neither engine would flag it because both would be wrong in the same way
  relative to a single-partition baseline. Only *cross-engine* comparison across ≥2 partitions catches this.
  
  **Add.**

      -- forces multiple partitions so partitionIndex ≠ 0 is exercised
      statement
      CREATE TABLE test_uuid_parts(id int) USING parquet

      statement
      INSERT INTO test_uuid_parts SELECT id FROM range(0, 32)

      query
      SELECT uuid(42) FROM test_uuid_parts DISTRIBUTE BY id

  Runs in default `query` mode → bit-for-bit vs Spark across partitions.

  ---

  ### Gap 2 — Aliasing (stateful semantics)

  **What's missing.** No test covers two `uuid` projections on the same row.

  **Why it matters.** Spark's `Uuid` is `stateful = true`, and `freshCopyIfContainsStatefulExpression` rewrites aliased instances. For the seeded form `uuid(0), uuid(0)` both nodes carry the same explicit seed
  → identical sequences. For the unseeded form `uuid(), uuid()` the analyzer assigns fresh random seeds → different sequences. Comet must not accidentally share a single `state_holder` across the two nodes
  (would produce interleaved draws) or hoist a single instance (would produce identical values for the unseeded case).
  
  **Add.**

      -- uuid_with_seed.sql: seeded aliases must agree
      query
      SELECT uuid(0) = uuid(0) FROM test_uuid_seed

      -- uuid.sql: unseeded aliases must diverge (spark_answer_only — random)
      query spark_answer_only
      SELECT uuid() = uuid() FROM test_uuid

  ---
  
  ### Gap 3 — Empty / zero-row batches

  **What's missing.** `evaluate` with `num_rows == 0` is not tested. The RNG must not advance on an empty batch.
  
  **Why it matters.** A `LIMIT 0` above a `uuid` projection, or a fully filtered partition, produces a zero-row batch. If the code ever grew a `next_uuid()` call outside the `0..num_rows` loop (bootstrap,
  warmup, etc.), the RNG state would drift and downstream bit-for-bit tests would fail intermittently.

  **Add** (Rust unit test — cheaper than a SQL test for this):

      #[test]
      fn test_uuid_empty_batch_does_not_advance_state() {
          let expr = UuidExpr::new(7);
          let _ = collect_uuids(&expr, &empty_batch(0));
          // After an empty eval, the next non-empty eval must equal a fresh generator.
          assert_eq!(collect_uuids(&expr, &empty_batch(3)), eval_uuids(7, 3));
      }

  ---
  
  ### Gap 4 — Rust golden fixture

  **What's missing.** `test_uuid_version_and_variant_bits` only checks the version nibble (`bytes[14] == b'4'`) and variant nibble (`bytes[19] ∈ {8,9,a,b}`). No test pins the full 128 bits.

  **Why it matters.** `next_long()` (two `next(32)` composed high|low) is used **only** by uuid — the shuffle tests exercise `next_int` and would not catch a `next_long` regression (e.g. accidentally swapping
  high/low, or masking with `0xffffffff` in the wrong place). A single golden fixture makes the Rust suite a first-line defense; today it depends on cross-engine SQL tests running to catch a twister bug.

  **Add** (values captured by running `RandomUUIDGenerator(0)` in Spark once):

      #[test]
      fn test_uuid_matches_spark_random_uuid_generator() {
          // Seeds and first three UUIDs captured from
          // `new RandomUUIDGenerator(0L).getNextUUID().toString` in spark-shell.
          let expected = vec![
              "…-4…-…", // fill from Spark
              "…-4…-…",
              "…-4…-…",
          ];
          assert_eq!(eval_uuids(0, 3), expected);
      }

  Also add a case for a negative seed (e.g. `Long.MIN_VALUE`) to lock in the sign-extension path through `set_seed_long`.

  ---

  ### Gap 5 — No-scan / typeof paths

  **What's missing.** `uuid_with_seed.sql` has a `SELECT uuid(0)` line (no FROM), but no assertion around it — just the raw value. And no `typeof` check.

  **Why it matters.** `SELECT uuid(0)` with no scan goes through a `LocalTableScan`/`OneRowRelation` execution shape rather than the parquet scan path. Different codegen/serde context, and worth pinning that
  it doesn't fall back. `typeof(uuid())` guards the dataType/collation contract without depending on the random value.

  **Add.**

      -- uuid.sql
      query
      SELECT typeof(uuid())

      -- uuid_with_seed.sql — pin the no-scan path with a deterministic assertion
      query
      SELECT length(uuid(0)), typeof(uuid(0))

  ---

  Priority order: **Gap 1 first** (real correctness risk), then Gap 4 (cheap Rust guard), then 2 / 3 / 5.

andygrove:
Thanks for the thorough review, @comphead — all five gaps addressed in 7ef6420c6:

- **Gap 1 (multi-partition)** — added `SELECT uuid(42) FROM test_uuid_parts DISTRIBUTE BY id` in `uuid_with_seed.sql`, backed by a 32-row `range(0, 32)` table so uuid runs post-shuffle and exercises `partitionIndex != 0`.
- **Gap 2 (aliased projections)** — `SELECT uuid(0) = uuid(0) FROM test_uuid_seed` in `uuid_with_seed.sql`, plus `SELECT uuid() = uuid() FROM test_uuid` in `uuid.sql` under `spark_answer_only` (unseeded aliases produce distinct fresh seeds).
- **Gap 3 (empty batch)** — new Rust test `test_uuid_empty_batch_does_not_advance_state`: evaluate on a zero-row batch must not consume RNG draws; the next non-empty batch must match a fresh generator.
- **Gap 4 (golden fixture)** — new Rust test `test_uuid_matches_commons_math3_random_uuid_generator` pinning 5 × 5 UUIDs across seeds 0, -1, `Long.MinValue`, `Long.MaxValue`, and 42. Values captured directly from Commons Math3's `MersenneTwister` — the exact RNG behind `RandomUUIDGenerator` — so a `next_long` regression (high/low swap, wrong mask, sign-extension bug in `set_seed_long`) fires without a JVM roundtrip.
- **Gap 5 (no-scan path)** — added `SELECT length(uuid(0))` alongside the existing `SELECT uuid(0)` to pin the `OneRowRelation` path with a deterministic assertion. Dropped the `typeof(uuid(...))` variant: Spark 4.1 codegen fails to compile that expression with `Unknown variable or type "string"` under the SQL fixture's config; `length()` gives us the deterministic pin without touching that path.

Verified locally: all 5 Rust unit tests pass; both SQL fixtures pass on Spark 4.1 (default) and Spark 3.5 (the seeded fixture is correctly skipped).

spark/src/test/resources/sql-tests/expressions/misc/uuid_with_seed.sql
```diff
@@ -0,0 +1,79 @@
...
+
+-- fixed seed, multiple rows: the generator advances per row within the partition
+query
+SELECT uuid(42) FROM test_uuid_seed
+
+-- Forces a hash exchange so uuid runs post-shuffle across several partitions. Locks the
+-- `seed + partitionIndex` offset used by RandomUUIDGenerator; a single-partition test would
+-- silently agree with Spark even if Comet ignored partitionIndex.
+query
+SELECT uuid(42) FROM test_uuid_parts DISTRIBUTE BY id
```
mbutrovich: This query does not exercise `partitionIndex != 0`, so the multi-partition coverage it was added for is still missing.

The update says it makes "uuid run post-shuffle". The plan is the other way round. `DISTRIBUTE BY` is applied by `withQueryResultClauses`, which calls `withRepartitionByExpression(ctx, expressionList(distributeBy), query)` where `query` is the already-constructed plan including the `Project` (Spark `AstBuilder.scala:1272-1274`). So the shape is `RepartitionByExpression [id]` on top of `Project [uuid(42)]`, and uuid is evaluated on the scan's partitions, before the exchange. The exchange's partition count never reaches this expression.

That leaves the scan's partition count, and it is very likely 1. `CometTestBase` runs `local[5]` (`CometTestBase.scala:545`), so `range(0, 32)` produces 5 partitions and the insert writes 5 tiny files. Under the default `spark.sql.files.maxPartitionBytes` of 128MB and `openCostInBytes` of 4MB, 5 tiny files pack into a single `FilePartition`. So `planner.partition()` is 0 and `expr.seed.wrapping_add(partition)` in `UuidBuilder::build` is never exercised on the SQL path.

This is exactly the silent-agreement failure that motivated adding this query: with one partition, Comet and Spark agree even if the partition offset were dropped entirely, so the test passes while covering nothing. Please move the projection above the exchange so it runs on post-shuffle partitions:

```sql
SELECT uuid(42) FROM (SELECT id FROM test_uuid_parts DISTRIBUTE BY id)
```

Since the value of this test rests entirely on an assumption about partition count, make that assumption checkable rather than implicit. A companion query such as `SELECT count(DISTINCT spark_partition_id()) FROM (SELECT id FROM test_uuid_parts DISTRIBUTE BY id)` fails loudly if the plan ever collapses back to one partition, instead of quietly returning to vacuous. The local run passing is consistent with either reading, since a one-partition plan agrees with Spark too, so the partition count is the thing to check rather than the pass.

spark/src/test/resources/sql-tests/expressions/misc/uuid_with_seed.sql
```diff
@@ -0,0 +1,79 @@
...
+statement
+INSERT INTO test_uuid_parts SELECT id FROM range(0, 32)
+
+-- fixed seed, single row -- also exercises the no-scan (OneRowRelation) planning path
+query
+SELECT uuid(0)
+
+-- pin the no-scan path with a deterministic assertion on top of the raw value above
+query
+SELECT length(uuid(0))
```
mbutrovich, line 53: This line replaced the `typeof(uuid(0))` variant, and the reason recorded for dropping it is not the actual cause.

It is not a Spark 4.1 codegen quirk. It is a latent bug in Spark's `TypeOf.doGenCode`, which interpolates the catalog string unquoted into generated Java:

```scala
defineCodeGen(ctx, ev, _ => s"""UTF8String.fromString(${child.dataType.catalogString})""")
```

For `typeof(uuid(0))` that emits `UTF8String.fromString(string)`, which is precisely the Janino error observed. It needs to be a Java string literal. This code is byte-identical on 3.5 (`misc.scala:280-282`), 4.0 (`:330-332`) and master (`:370-372`), so it reproduces on every supported version and has nothing to do with uuid. It is normally invisible because `TypeOf.foldable = true` and ConstantFolding eliminates the node before codegen runs. The SQL fixture suite excludes ConstantFolding, which is what exposes it.

Two things worth capturing rather than losing. This deserves an upstream Spark report. And it is a harness-level constraint: no Comet SQL fixture can use `typeof` on a non-foldable input, which the next person will otherwise rediscover the same way. Please record the real cause in the description instead of the version-specific framing.

spark/src/test/resources/sql-tests/expressions/misc/uuid.sql
```diff
@@ -0,0 +1,43 @@
...
+-- This regex subsumes the length, alphabet, version, and variant checks.
+query
+SELECT uuid() RLIKE '^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$' FROM test_uuid
+
+-- Two unseeded uuid() nodes get distinct fresh random seeds during analysis (Uuid is stateful,
+-- freshCopyIfContainsStatefulExpression rewrites aliases). Every row must therefore compare
+-- unequal -- if Comet ever hoisted a single instance, this would flip to true.
+-- spark_answer_only: value is random, so we only cross-check that Spark and Comet agree.
+query spark_answer_only
+SELECT uuid() = uuid() FROM test_uuid
```
mbutrovich, line 45: `spark_answer_only` gives up the operator assertion here for no reason.

`SparkAnswerOnly` maps to `checkSparkAnswer` (`CometSqlFileTestSuite.scala:153`), which compares values but does not assert the expression ran in Comet. The comment above says this query guards against Comet hoisting a single shared instance, but that guard only holds while the projection is actually native, which this mode does not check.

It also does not need the relaxed mode. `uuid() = uuid()` is deterministic: the two nodes get distinct fresh seeds, so the result is `false` on every row on both engines. Dropping `spark_answer_only` and using default `query` mode keeps the value comparison and adds the nativeness assertion.

native/spark-expr/src/nondetermenistic_funcs/uuid.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// }
+    /// ```
+    ///
+    /// `next_long()` is used *only* by `uuid`, so the shuffle tests (which exercise
+    /// `next_int`) do not guard against a `next_long` regression such as swapping the
+    /// high/low words or masking with the wrong constant. This test locks all 128 bits
+    /// per row, catches sign-extension bugs in `set_seed_long` via the negative seeds,
+    /// and runs entirely in Rust so it fires without a JVM roundtrip.
+    #[test]
+    fn test_uuid_matches_commons_math3_random_uuid_generator() {
```
mbutrovich, line 228: On 3.4 and 3.5 these golden values are the only line of defense, and their provenance cannot be checked from the repo.

The doc comment above documents how to regenerate them with a Java snippet, which is the right instinct, but nothing here lets a reviewer confirm the committed values came from Commons Math3 rather than from this same Rust code. If they are ever regenerated from the Rust side the test becomes circular and silently stops testing anything.

On 4.0+ that risk is contained, because `uuid_with_seed.sql` compares against Spark bit for bit and would catch a `next_long` port bug regardless. On 3.4 and 3.5 there is no such backstop: the seeded fixture is skipped by `MinSparkVersion: 4.0`, and unseeded `uuid()` cannot be compared across engines, so a twister regression on those profiles would be caught only by these constants.

`Uuid(Some(seed))` is constructible from Scala on 3.4 and 3.5 even though the SQL form is not. A small Scala test building the expression directly and comparing Comet against Spark would give every profile a bit-for-bit assertion and remove the dependence on hand-captured constants. That closes the one place where this PR can regress without any test noticing.

mbutrovich (commented):
Thanks @andygrove! Four comments below, against the state after 7ef6420c6.

Taking the five test additions in turn: the empty-batch Rust test is correct as written and I have no comment on it. The seeded aliasing test is sound, since a shared `state_holder` would interleave draws and flip `uuid(0) = uuid(0)` to false, so it does discriminate the case it targets; my comment on `uuid.sql:43` is only about the unseeded half. The Commons Math3 golden fixture is a good addition and my comment on it concerns provenance, not the test. The remaining two I disagree with: the `DISTRIBUTE BY` query does not reach `partitionIndex != 0`, and the `typeof` failure is not a Spark 4.1 quirk. Details inline.

No issues in the RNG port itself. The masking matches `RandomUUIDGenerator.getNextUUID` mask for mask including the variant ordering, `next_long` matches Commons Math3 `BitsStreamGenerator.nextLong()`, and the arithmetic `seed >> 32` in `set_seed_long` is equivalent to Commons Math3's `>>> 32` because both are truncated by the `as i32` cast. `nullable = false` and `data_type = Utf8` are both correct, the latter including on 4.0+ where `Uuid` mixes in `DefaultStringProducingExpression`: `ApplyDefaultCollationToStringType` wraps the node in an explicit `Cast` rather than mutating its `dataType` (`ApplyDefaultCollationToStringType.scala:238-247`), so the node itself is always plain `StringType`. The `MinSparkVersion: 4.0` gating is right, since 3.5's `Uuid` has only `this()` (`misc.scala:204-207`) and 4.0 adds `this(seed: Expression)` (`misc.scala:257`).

spark/src/test/resources/sql-tests/expressions/misc/uuid_with_seed.sql
andygrove: You are right about the plan shape, and the query was covering nothing. Fixed in 2f6a5839e with your rewrite:

```sql
SELECT uuid(42) FROM (SELECT id FROM test_uuid_parts DISTRIBUTE BY id)
```

On the companion partition-count query: I tried it and it cannot work in this suite, for a reason worth recording. `count(DISTINCT spark_partition_id())` is rejected outright — `[AGGREGATE_FUNCTION_WITH_NONDETERMINISTIC_EXPRESSION]` — and materialising it first (`count(DISTINCT pid) FROM (SELECT spark_partition_id() AS pid FROM ...)`) does run, but it still asserts nothing: `CometSqlFileTestSuite` compares Comet against **Spark**, so if the plan ever collapsed to one partition, Spark would collapse identically, both sides would return the same value, and the check would pass. There is no expected-literal mode to pin it against.

So the assumption is only assertable from Scala, and that is where I put it: `CometUuidExpressionSuite` has a multi-partition test that does `assert(repartitioned.rdd.getNumPartitions > 1)` before comparing. The fixture now carries a comment explaining why the check lives there rather than inline.

I also verified the coverage is real rather than assuming it this time. Mutating `UuidBuilder::build` to drop the offset:

```rust
let seed = expr.seed;  // was expr.seed.wrapping_add(planner.partition().into())
```

fails both new Scala tests. Restored, and everything passes on spark-3.5 and spark-4.1.

spark/src/test/resources/sql-tests/expressions/misc/uuid.sql
andygrove: Agreed — dropped in 2f6a5839e, now default `query` mode. You are right that the comparison is deterministic even though the values are not: the two nodes get distinct fresh seeds, so it is `false` on every row on both engines. It passes on 3.5 and 4.1 in default mode, so the nativeness assertion the guard depends on is now actually being made.

spark/src/test/resources/sql-tests/expressions/misc/uuid_with_seed.sql
andygrove: You are right, and I checked the mechanism rather than just taking the correction: `TypeOf.foldable` is true, so ConstantFolding normally removes the node before codegen ever sees it, and `CometSqlFileTestSuite` sets `spark.sql.optimizer.excludedRules` to `ConstantFolding` (lines 91-92), which is exactly what exposes the latent bug. So the version-specific framing in my earlier comment was wrong on both counts — it is not 4.1 and it is not uuid.

Recorded in 2f6a5839e as a comment on the query, framed as the harness-level constraint you asked for: no SQL fixture in this suite can use `typeof` on a non-foldable input, with the unquoted-`catalogString` cause spelled out so the next person does not rediscover it through the Janino error. I will correct the PR description too.

On the upstream report: I have not filed it, since that is an outward-facing action on someone else's tracker and I would rather confirm before opening a Spark ticket. Happy to file it if you want, or leave it to you — the repro is `SELECT typeof(<non-foldable>)` with ConstantFolding excluded, on any of 3.5, 4.0 or master.

native/spark-expr/src/nondetermenistic_funcs/uuid.rs
andygrove: This was the most useful comment on the PR — the circularity risk is real and I had not protected against it. Fixed in 2f6a5839e.

Added `CometUuidExpressionSuite`, which builds `Uuid(Some(seed))` directly via the existing version-shimmed `getColumnFromExpression` in `ShimCometTestBase` (`new Column(expr)` on 3.x, `ExpressionColumnNode` on 4.x), so it compiles and runs on every profile. Two tests: one over seeds 0, -1, 42, `Long.MinValue` and `Long.MaxValue`, and one that repartitions and asserts `getNumPartitions > 1` before comparing. Both use `checkSparkAnswerAndOperator`, so they assert bit-for-bit equality with Spark *and* that the projection stayed native.

That gives 3.4 and 3.5 a real cross-engine assertion for the first time, so the golden constants are no longer the only line of defense on those profiles.

Confirmed non-vacuous by mutation: dropping the partition offset in `UuidBuilder::build` fails **both** new tests (the single-partition one too — `withParquetTable` over 20 rows under `local[5]` already gives more than one partition). Restored, and all pass on spark-3.5 and spark-4.1.

I also left a note on the Rust golden test pointing at the new suite as its backstop and stating explicitly that regenerating these constants from the Rust side would make the test circular — so the trap is documented where someone would fall into it.

andygrove:
@mbutrovich all four addressed in 2f6a5839e, details inline. Two of them found real holes:

**The multi-partition query was covering nothing.** Your reading of the plan shape is correct — `DISTRIBUTE BY` wraps the already-built Project, so uuid ran pre-shuffle on a single `FilePartition`. Moved the projection above the exchange.

**The golden constants were the only guard on 3.4/3.5, and could have gone circular.** Added `CometUuidExpressionSuite`, which builds `Uuid(Some(seed))` directly through the existing version-shimmed `getColumnFromExpression`, so every profile now gets a bit-for-bit cross-engine assertion. Five seeds including negative/MIN/MAX, plus a multi-partition test that asserts `getNumPartitions > 1` before comparing.

**I verified both are non-vacuous rather than asserting it.** Mutating `UuidBuilder::build` to drop `wrapping_add(planner.partition())` fails both new Scala tests; restored, everything passes on spark-3.5 and spark-4.1.

**On the companion partition-count query — it cannot work here, and this is worth knowing for future fixtures.** `count(DISTINCT spark_partition_id())` is rejected outright with `[AGGREGATE_FUNCTION_WITH_NONDETERMINISTIC_EXPRESSION]`, and the materialised form runs but asserts nothing: this suite compares Comet against *Spark*, so a collapse to one partition would collapse both sides identically and still pass. There is no expected-literal mode. So the check is only meaningful from Scala, which is where it now lives; the fixture documents why.

**`spark_answer_only`** dropped — you were right that the comparison is deterministic even though the values are not.

**`typeof`:** I checked the mechanism, and your diagnosis holds — `TypeOf.foldable` is true so ConstantFolding normally removes the node, and this suite excludes ConstantFolding (`CometSqlFileTestSuite:91-92`), which is what exposes the unquoted-`catalogString` bug. My earlier "Spark 4.1 codegen" framing was wrong on both counts. Recorded as a harness constraint on the query, and I will fix the PR description. I have **not** filed the upstream Spark ticket — that is an outward-facing action on another project's tracker, so I would rather you confirm or take it; the repro is `SELECT typeof(<non-foldable>)` with ConstantFolding excluded, on 3.5, 4.0 or master.

Verified: 5 Rust uuid unit tests pass; `uuid.sql`, `uuid_with_seed.sql` and both new Scala tests pass on spark-3.5 and spark-4.1.

mbutrovich (approved):

PR #5036, opened 2026-07-25
fix: prevent libhdfs thread destructor use-after-free

### LINKED ISSUE #5023: Sporadic SIGSEGV in macOS [scans] workflow in CI
I've seen this a few different times now in the macos [scans] workflow in CI. Also reported here: https://github.com/apache/datafusion-comet/pull/5014#issuecomment-5066355705.

Looks like:
```
ParquetReadFromFakeHadoopFsSuite:
- native scan on fake fs (530 milliseconds)
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000000000000, pid=22862, tid=59915
#
# JRE version: OpenJDK Runtime Environment Zulu17.68+17-CA (17.0.20+8) (build 17.0.20+8-LTS)
# Java VM: OpenJDK 64-Bit Server VM Zulu17.68+17-CA (17.0.20+8-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# C  0x0000000000000000
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/runner/work/datafusion-comet/datafusion-comet/spark/hs_err_pid22862.log
#
# If you would like to submit a bug report, please visit:
#   http://www.azul.com/support/
#
ParquetTimestampLtzAsNtzSuite:
```
Recent failure: https://github.com/apache/datafusion-comet/actions/runs/30050199441/job/89355580948?pr=5014

peterxcli:

Closes #5023.

## Rationale for this change

Vendored libhdfs detaches every thread with a cached `JNIEnv`, including threads that were already attached by the JVM. During Spark worker teardown, that `JNIEnv` can already be freed when `hdfsThreadDestructor` calls through it, causing a jump to address zero. [HDFS-16021](https://issues.apache.org/jira/browse/HDFS-16021) describes the same ownership bug, but no released `hdfs-sys` version contains the fix.

PR for hdfs-sys: https://github.com/Xuanwo/hdfs-sys/pull/47, this patch is basically copy from [HDFS-16021 patch](https://issues.apache.org/jira/secure/attachment/13025245/fix-hdfsThreadDestructor.patch)

Patch crates.io resolution at the Cargo workspace level to pin `hdfs-sys` to an immutable commit carrying the ownership fix. The workspace patch makes both Comet's direct dependency and the OpenDAL/HDFS path through `hdrs` use the same fixed crate.

- `cargo build --locked --profile ci` on macOS ARM64 with JDK 17
- `cargo tree --locked -p datafusion-comet -i hdfs-sys -e features` confirms one Git-sourced `hdfs-sys` instance with `hdfs_3_3` and `vendored`
- 100 fresh-JVM runs of `ParquetReadFromFakeHadoopFsSuite` against the patch, followed by 10 fresh-JVM runs against the exact GitHub-pinned artifact
- Disassembly confirms `hdfsThreadDestructor` checks libhdfs ownership before dereferencing `JNIEnv`


sunchao:
@peterxcli are you still working on this? This should help to make Comet CI more stable - today there are some flaky tests on the `libhdfs` path

peterxcli:
@sunchao I've raised a patch to the hdfs-sys upstream. should we push it to proceed? Or just use my fork(this pr is now going this way)?

https://github.com/Xuanwo/hdfs-sys/pull/47

sunchao:
@peterxcli Thanks, I think we can switch to your fork for the time being before the issue is resolved in the other repo.

cc @andygrove too

peterxcli:
@sunchao Feel free to merge this anytime, thanks!

andygrove:

The diagnosis is convincing and the verification is thorough. Detaching a thread the JVM attached is exactly the ownership bug HDFS-16021 describes, and 100 fresh-JVM runs plus the disassembly check is more evidence than most crash fixes get.

My concern is entirely about where the fix comes from.

**Pinning to a personal fork**

`[patch.crates-io] hdfs-sys = { git = "https://github.com/peterxcli/hdfs-sys", rev = "..." }` makes every Comet native build, including release builds, fetch source from a contributor's personal GitHub account. For an ASF project that is a problem beyond the usual supply-chain discomfort. The release artifacts would carry a dependency that is not in any registry, is not covered by any release vote, and disappears if the account or repo goes away.

What is the upstream path here? Some questions that would help:

- Has a PR been opened against `Xuanwo/hdfs-sys`, and is there any signal on whether it will be merged and released? If there is, linking it here would let us track it.
- If upstream is unresponsive, is vendoring the patched C source into Comet's tree an option? That keeps the fix under ASF's own release process and is auditable in the repo.
- Failing both, could the fork at least live under an org rather than a personal account, with a comment in `Cargo.toml` pointing at the upstream issue and stating explicitly that this is temporary?

I would not want to merge this as-is on a release branch, but I also do not want the crash to stay unfixed. If there is a way to land the fix under Comet's own control, or to get a release out of upstream, that seems worth a bit more effort before merging.

**One smaller note**

`[patch.crates-io]` only applies to builds of this workspace. That is fine today since Comet's native crates are not published, but if `datafusion-comet-spark-expr` is ever published to crates.io the patch will be silently ignored for downstream consumers and they will get the broken `hdfs-sys`. Worth a comment in `Cargo.toml` so that does not surprise someone later.

peterxcli:
@andygrove sorry for the late response

> Has a PR been opened against Xuanwo/hdfs-sys, and is there any signal on whether it will be merged and released? If there is, linking it here would let us track it.

Updated in PR description 

> If upstream is unresponsive, is vendoring the patched C source into Comet's tree an option? That keeps the fix under ASF's own release process and is auditable in the repo.

Seems like we need to copy the whole source code of patched hdfs-sys. if we want to go this way.

> Failing both, could the fork at least live under an org rather than a personal account, with a comment in Cargo.toml pointing at the upstream issue and stating explicitly that this is temporary?

I can create a org, any prefer name?

andygrove:
> I can create a org, any prefer name?

Sorry, disregard the suggestion to create an org - that was Claude's idea

andygrove:
> @sunchao I've raised a patch to the hdfs-sys upstream. should we push it to proceed? Or just use my fork(this pr is now going this way)?
> 
> [Xuanwo/hdfs-sys#47](https://github.com/Xuanwo/hdfs-sys/pull/47)

@xuanwo Could you please review [Xuanwo/hdfs-sys#47](https://github.com/Xuanwo/hdfs-sys/pull/47) to unblock us?

peterxcli:
@sunchao looks like #5453 accidentally closed this, could you help me to reopen? Thanks!

andygrove:
I'm going to create an alternative PR to this one, to vendor the necessary hdfs code within the Comet repo

andygrove:
> I'm going to create an alternative PR to this one, to vendor the necessary hdfs code within the Comet repo

https://github.com/apache/datafusion-comet/pull/5890

Issue #4465, opened 2026-05-27
[Bug] decode ignores Spark 4.0 legacyCharsets and legacyErrorAction flags


andygrove:
## Describe the bug

Spark 4.0 refactored `StringDecode` from a `BinaryExpression` to a `RuntimeReplaceable` whose `replacement` is `StaticInvoke(StringDecode.decode, bin, charset, legacyCharsets, legacyErrorAction)`. The two new boolean arguments control malformed-character handling: with `legacyErrorAction = true`, Spark substitutes replacement characters for invalid UTF-8 sequences (matching the Spark 3.x behavior); with `legacyErrorAction = false` (the default), Spark raises `QueryExecutionErrors.malformedCharacterCoding(...)`.

Comet's Spark 4.0 shim (`spark/src/main/spark-4.0/org/apache/comet/shims/CometExprShim.scala`) destructures the `StaticInvoke` arguments and discards both flags, then routes through `CommonStringExprs.stringDecode` which always lowers to `Cast(bin, StringType, TRY)`. The Cast TRY path produces NULL on invalid UTF-8 in all cases. That means:

- Under Spark 4.0 default mode (`legacyErrorAction = false`): Spark raises, Comet returns NULL.
- Under Spark 4.0 legacy mode (`legacyErrorAction = true`): Spark substitutes replacement characters, Comet returns NULL.
- Under Spark 3.x: Spark substitutes replacement characters, Comet returns NULL.

Surfaced by the string-expressions audit in apache/datafusion-comet#4461.

## Steps to reproduce

```sql
SET spark.sql.legacy.javaCharsets = true;
SELECT decode(X'FF', 'UTF-8');
```

Spark 3.x: returns `?` (Unicode replacement).
Spark 4.0 (legacy mode): same as 3.x.
Spark 4.0 (default mode): raises `MALFORMED_CHARACTER_CODING`.
Comet: returns NULL in all three cases.

## Expected behavior

Honor `legacyCharsets` / `legacyErrorAction` when running under Spark 4.0+. At minimum, the flags should be propagated through the proto so the native impl can choose between the substitute/throw/null modes.

## Additional context

- Shim location: `spark/src/main/spark-4.0/org/apache/comet/shims/CometExprShim.scala` (and `spark-4.1`, `spark-4.2`)
- Helper: `CommonStringExprs.stringDecode` in `spark/src/main/scala/org/apache/comet/serde/strings.scala`
- Related: #4465 (decode not surfaced in compatibility docs)

PR #5037, opened 2026-07-25
feat: support Spark encode expression via codegen dispatch

### LINKED ISSUE #3183: [Feature] Support Spark expression: encode

> **Note:** This issue was generated with AI assistance. The specification details have been extracted from Spark documentation and may need verification.

Comet does not currently support the Spark `encode` function, causing queries using this function to fall back to Spark's JVM execution instead of running natively on DataFusion.

The Encode expression converts a string to binary data using a specified character encoding. It is a runtime-replaceable expression that delegates to a static method for actual encoding operations, supporting legacy charset and error handling configurations.

Supporting this expression would allow more Spark workloads to benefit from Comet's native acceleration.

### Spark Specification

**Syntax:**
```sql
ENCODE(string_expr, charset_expr)
```

**Arguments:**
| Argument | Type | Description |
|----------|------|-------------|
| str | String | The input string expression to be encoded |
| charset | String | The character set/encoding name to use for conversion |
| legacyCharsets | Boolean | Internal flag for legacy Java charset handling behavior |
| legacyErrorAction | Boolean | Internal flag for legacy coding error action behavior |

**Return Type:** BinaryType - Returns binary data representing the encoded string.

**Supported Data Types:**
- Input string: StringTypeWithCollation (supports trim collation)
- Input charset: StringTypeWithCollation (supports trim collation)
- Both string inputs support collation-aware string types

**Edge Cases:**
- Null input string or charset results in null output
- Invalid charset names may throw runtime exceptions
- Legacy flags affect error handling behavior for malformed input
- Empty string input produces empty binary output
- Behavior varies based on SQLConf legacy settings for charset and error handling

**Examples:**
```sql
-- Encode string using UTF-8
SELECT ENCODE('hello world', 'UTF-8');

-- Encode with different charset
SELECT ENCODE('café', 'ISO-8859-1');
```

```scala
// DataFrame API usage
df.select(expr("ENCODE(name, 'UTF-8')").as("encoded_name"))

// Using column expressions
df.select(encode(col("text_column"), lit("UTF-16")))
```

### Implementation Approach

See the [Comet guide on adding new expressions](https://datafusion.apache.org/comet/contributor-guide/adding_a_new_expression.html) for detailed instructions.

1. **Scala Serde**: Add expression handler in `spark/src/main/scala/org/apache/comet/serde/`
2. **Register**: Add to appropriate map in `QueryPlanSerde.scala`
3. **Protobuf**: Add message type in `native/proto/src/proto/expr.proto` if needed
4. **Rust**: Implement in `native/spark-expr/src/` (check if DataFusion has built-in support first)

## Additional context

**Difficulty:** Medium
**Spark Expression Class:** `org.apache.spark.sql.catalyst.expressions.Encode`

**Related:**
- Decode - inverse operation to convert binary back to string
- Cast expressions for type conversions
- String manipulation functions in string_funcs group

---
*This issue was auto-generated from Spark reference documentation.*

andygrove:

Closes #3183.

## Rationale for this change

The Spark `encode(str, charset)` expression (StringType, StringType) -> BinaryType falls back to Spark on 3.4 and 3.5: there is no native charset encoder in Comet, so any query using `encode` breaks the native pipeline on those versions. On 4.0+ `encode` is `RuntimeReplaceable` and lowers to `StaticInvoke(Encode.encode, ...)`, which the `StaticInvoke` catch-all from #5692 already routes through the codegen dispatcher, so 4.x works on main today and only 3.x is left.

Rather than reimplement charset encoding natively, this routes the 3.x `Encode` node through the same JVM codegen dispatcher, which compiles Spark's own `Encode` code into an Arrow-in/Arrow-out kernel that runs inside the Comet pipeline. Behavior therefore matches Spark exactly, including Spark 4.0+'s `legacyJavaCharsets` / `legacyCodingErrorAction` conf-driven behavior, the UTF-8 fast path, and the strict-mode error on malformed input. This is the same approach already used for the dual `decode` expression (#4465), and reimplementing the charset logic natively would risk drifting from Spark's per-version semantics.

- **Spark 3.4 / 3.5**, where `Encode` is a `BinaryExpression`: a new one-line serde `object CometEncode extends CometCodegenDispatch[Encode]` in the shared `spark-3.x` source set, mirroring `CometStringDecode`, registered in the `sparkVersionSpecificStringExpressions` map of both the 3.4 and 3.5 `CometExprShim`. This is the only production change in the PR.
- **Spark 4.0 / 4.1 / 4.2**: nothing. An earlier revision added an `("encode", classOf[Encode]) -> CometStaticInvokeCodegenDispatch` entry to the `staticInvokeExpressions` map in `statics.scala`, alongside the existing `decode` entry. #5692 landed the `StaticInvoke` catch-all in `CometStaticInvoke.convert` in the meantime, which routes exactly this lowering through the dispatcher, so the explicit entry was redundant and has been dropped. The 4.x fixtures below still cover the behavior.
- The stale note on the `encode` row of the expressions user-guide page is removed. The row was already ✅ because of that catch-all; the note said "on Spark 3.x it falls back", which this PR makes untrue. The Implementation cell stays `—`, which is what `GenerateDocs` emits on 4.x (see below), matching `decode`.
- The `encode(utf-8)` and `to_binary(utf-8)` cases in `CometCodegenDispatchBenchmark` were gated on `isSpark40Plus`, which skipped exactly the versions this PR changes. Both now run unconditionally, with a comment recording that 4.0+ reaches the dispatcher via `StaticInvoke` and 3.x via `CometEncode`. The `to_time(fmt)` case is still gated on 4.1, correctly.
- The `MaxSparkVersion` fixture directive, which already existed in `SqlFileTestParser` but was undocumented, is documented in the SQL file tests contributor guide.

No native Rust is added. `encode` is eligible for the dispatcher because its output (`BinaryType`) and inputs (`StringType`, plus the Spark 4.0 boolean flag literals) are all in `CometBatchKernelCodegen.canHandle`'s supported type surface.

Spark's behavior for `encode` differs across versions in two ways, and since Spark's own code runs under the dispatcher, Comet matches each version by construction:

- **Unmappable input.** On 3.4/3.5, `Encode` is `input.toString.getBytes(charset)`, and `String.getBytes` uses `CodingErrorAction.REPLACE`, so unmappable characters become `0x3F` (`?`). On 4.0+, `Encode.encode` builds the encoder with `CodingErrorAction.REPORT` and converts the resulting `CharacterCodingException` into `MALFORMED_CHARACTER_CODING`.
- **Charset validation.** On 3.4/3.5, any JVM charset is accepted and an unknown name surfaces as `UnsupportedEncodingException`. On 4.0+, `CharsetProvider.forName` permits only the seven names in `VALID_CHARSETS` unless `spark.sql.legacy.javaCharsets` is set, and otherwise raises `INVALID_PARAMETER_VALUE.CHARSET`.

Six new SQL file tests under `sql-tests/expressions/string/`. Every `query` block runs `checkSparkAnswerAndOperator`, which fails if the expression fell back to Spark instead of executing natively, so the coverage is non-vacuous. `CometSqlFileTestSuite` also disables constant folding and requires any fixture using `expect_error` to include a plain `query` sentinel over valid input, so the error cases cannot pass through an operator-level fallback.

- `encode.sql` (all versions): all seven `VALID_CHARSETS` entries (UTF-8 in both cases, UTF-16, UTF-16BE, UTF-16LE, UTF-32, US-ASCII, ISO-8859-1), null and empty-string input, a non-literal charset column, and a `decode(encode(...))` round-trip. UTF-16LE is covered separately from UTF-16 (BOM) and UTF-16BE (byte order) because the output bytes genuinely differ. The US-ASCII block uses ASCII-only input so this file stays version-agnostic.
- `encode_unmappable.sql` (`MaxSparkVersion: 3.5`): substitution behavior over column and literal input, pinning the substituted bytes rather than only asserting Spark/Comet agreement.
- `encode_unmappable_strict.sql` (`MinSparkVersion: 4.0`): `expect_error(MALFORMED_CHARACTER_CODING)` for the same inputs, plus a sentinel query.
- `encode_invalid_charset.sql` (`MaxSparkVersion: 3.5`): `windows-1252` accepted, unknown name gives `expect_error(UnsupportedEncodingException)`.
- `encode_invalid_charset_strict.sql` (`MinSparkVersion: 4.0`): `expect_error(INVALID_PARAMETER_VALUE)` for `windows-1252` as a literal and as a column value, and for an unknown charset name.
- `encode_legacy_charsets.sql` (`MinSparkVersion: 4.0`, `spark.sql.legacy.javaCharsets=true`): `windows-1252` and `Shift_JIS` succeed with the flag on. Rows are representable in `windows-1252` so the still-strict coding-error action does not fire and the two flags stay independently tested.

The suite passes with all blocks native under the Spark 3.4, 3.5, 4.0, and 4.1 profiles.

The `encode` row in the expressions user-guide page was checked by running `GenerateDocs` against a scratch copy of the user guide under the 3.5, 4.0 and 4.1 profiles: the generated Implementation cell is `—` on 4.x (matching what is committed, and matching `decode`, since on 4.x no serde is keyed on `classOf[Encode]`) and `Codegen dispatch` on 3.5.

`CometCodegenDispatchBenchmark` was run on Spark 3.5.9 / JDK 17 / `local[1]` now that the two cases are ungated there. At 1,048,576 rows, `encode(utf-8)` is 61 ms with dispatch against 62 ms with dispatch off and 78 ms with Comet disabled; the repeated baseline came in at 59 ms, so the noise floor is wider than the dispatch/no-dispatch gap and the honest reading is that dispatch is performance-neutral versus falling the projection back, not that it is faster. `to_binary(utf-8)` has a 0 ms spread between its two baselines and a 5 ms (~8%) gap, which is a real if modest win. Both Comet arms beat Comet-disabled Spark by ~1.25x. That is the expected shape: both arms run Spark's own `Encode` implementation, so what is being traded is bridge overhead against losing the operator. The run emitted no `WARNING` lines, which matters more than the timings — `checkPlans` verifies per case that the dispatch arm is fully native and that the dispatch-off arm is not, so on 3.5 the two arms really are different plans and `CometEncode` is doing the work.

---

This work was scaffolded with the superpowers brainstorming, writing-plans, and subagent-driven-development skills.


spark/src/test/resources/sql-tests/expressions/string/encode.sql
```diff
@@ -0,0 +1,84 @@
...
+SELECT encode(s, 'UTF-16') FROM test_encode
+
+query
+SELECT encode(s, 'UTF-16BE') FROM test_encode
+
+query
+SELECT encode(s, 'ISO-8859-1') FROM test_encode
+
+-- US-ASCII: use ASCII-only input. Non-ASCII input under US-ASCII throws on Spark 4.0+ strict
+-- mode (matching Spark), so it is not exercised here.
```
mbutrovich: The comment works out the per-version split correctly and then declines to test it. The split is real: on 3.4 and 3.5, `nullSafeEval` is `input1.toString.getBytes(toCharset)` (Spark `stringExpressions.scala:2708`), which substitutes 0x3F for unmappable characters, while on 4.0 and later `Encode.encode` builds an encoder via `CharsetProvider.newEncoder` with `CodingErrorAction.REPORT` and converts `CharacterCodingException` into `QueryExecutionErrors.malformedCharacterCoding("encode", charset)`, error class `MALFORMED_CHARACTER_CODING` (`stringExpressions.scala:3192-3200`, `CharsetProvider.scala:52-67`, `QueryExecutionErrors.scala:2844-2850`).

The `decode` work this PR mirrors does cover its equivalent, with two version-split fixtures: `decode_invalid_utf8.sql` and `decode_invalid_utf8_strict.sql` (MinSparkVersion 4.0, `expect_error(MALFORMED_CHARACTER_CODING)`, plus a sentinel query so the assertion cannot be satisfied by an operator-level fallback). Please add the matching pair, an `encode_unmappable.sql` with MaxSparkVersion 3.5 asserting substitution and an `encode_unmappable_strict.sql` with MinSparkVersion 4.0 asserting the error. Right now the substituting branch and the throwing branch are both untested on every profile.

spark/src/test/resources/sql-tests/expressions/string/encode.sql
```diff
@@ -0,0 +1,84 @@
...
+-- Tests for the SQL `encode(str, charset)` function (StringType, StringType) -> BinaryType.
+--
+-- `encode` runs through the codegen dispatcher (Spark's own doGenCode inside the Comet
+-- pipeline) so behavior matches Spark exactly across all supported charsets and across the
+-- Spark 4.0 `legacyCharsets` / `legacyErrorAction` modes. This is the dual of the `decode`
+-- codegen-dispatch path (#4465).
+--
+-- Each `query` block runs checkSparkAnswerAndOperator, which fails if the expression fell back
+-- to Spark instead of executing natively, so these are non-vacuous.
+-- Config: spark.comet.exec.scalaUDF.codegen.enabled=true
```
mbutrovich, line 27: `CharsetProvider.forName` permits only us-ascii, iso-8859-1, utf-8, utf-16be, utf-16le, utf-16 and utf-32 unless `spark.sql.legacy.javaCharsets` is set, and otherwise raises `invalidCharsetError`, error class `INVALID_PARAMETER_VALUE.CHARSET` (`CharsetProvider.scala:29-49`, `QueryExecutionErrors.scala:2834-2842`); on 3.x an unknown charset surfaces as UnsupportedEncodingException through `Platform.throwException` (`stringExpressions.scala:2716-2717`). So `encode(s, 'windows-1252')` has three distinct behaviors across version and conf. The description asserts the flag is honored by construction, which is true, but nothing demonstrates it. Add an invalid-charset error case and one query under `spark.sql.legacy.javaCharsets=true`.

spark/src/test/resources/sql-tests/expressions/string/encode.sql
```diff
@@ -0,0 +1,84 @@
...
+SELECT encode(s, 'UTF-8') FROM test_encode
+
+query
+SELECT encode(s, 'UTF-16') FROM test_encode
+
+query
+SELECT encode(s, 'UTF-16BE') FROM test_encode
+
+query
+SELECT encode(s, 'ISO-8859-1') FROM test_encode
```
mbutrovich, line 59: The description says six charsets including UTF-16LE. This block plus line 62 covers utf-8, UTF-8, UTF-16, UTF-16BE, ISO-8859-1 and US-ASCII. UTF-16LE is absent and worth adding for real: UTF-16 emits a BOM and UTF-16LE does not, so the byte output genuinely differs. UTF-32 is the remaining entry in VALID_CHARSETS with no coverage.

docs/source/user-guide/latest/expressions.md
```diff
@@ -562,7 +562,7 @@ expression-level). The `outer` variants are wired but marked `Incompatible`; the
 | `contains` | ✅ | — |  |
 | `decode` | ✅ | — |  |
 | `elt` | ✅ | Codegen dispatch |  |
-| `encode` | 🔜 | — | Lowers to `StaticInvoke(encode)` (not allowlisted); falls back |
+| `encode` | ✅ | — |  |
```
mbutrovich, line 565: The Implementation cell stays `—`, which is the correct generated value on 4.x: `encode` is registered in the `FunctionRegistry` on 3.4, 3.5 and 4.0 (`FunctionRegistry.scala:523`, `:527`, `:546` respectively), but on 4.x it is RuntimeReplaceable so no serde is keyed on `classOf[Encode]` in that profile and `GenerateDocs` emits the placeholder. On 3.4 and 3.5 the same cell becomes `Codegen dispatch` because `CometEncode` is a `CometCodegenDispatch` (`GenerateDocs.scala:194`). `decode` already has this shape, so the committed value is consistent. Just regenerate with `dev/generate-release-docs.sh` rather than hand-editing the row.

mbutrovich (commented):

spark/src/test/resources/sql-tests/expressions/string/encode.sql
andygrove: Added the pair in f4cdc33e9.

`encode_unmappable.sql` carries `MaxSparkVersion: 3.5` and asserts substitution over both column and literal input, pinning the bytes rather than only asserting Spark/Comet agreement: `encode('café', 'US-ASCII') = CAST('caf?' AS BINARY)`, plus CJK input under both US-ASCII and ISO-8859-1 (and the ISO-8859-1 case where `é` maps to 0xE9 but the CJK characters do not).

`encode_unmappable_strict.sql` carries `MinSparkVersion: 4.0` and uses `expect_error(MALFORMED_CHARACTER_CODING)` for literal US-ASCII, literal ISO-8859-1 and column input, with a sentinel `query` over ASCII-only rows first so the assertions cannot be satisfied by an operator-level fallback.

spark/src/test/resources/sql-tests/expressions/string/encode.sql
andygrove: Added all three branches in f4cdc33e9.

- `encode_invalid_charset_strict.sql` (`MinSparkVersion: 4.0`): `expect_error(INVALID_PARAMETER_VALUE)` for `windows-1252` as a literal, `windows-1252` as a column value, and a name the JVM does not know at all, since `CharsetProvider.forName` rejects all three the same way.
- `encode_invalid_charset.sql` (`MaxSparkVersion: 3.5`): `windows-1252` succeeds because 3.x calls `String.getBytes` directly, and an unknown name gives `expect_error(UnsupportedEncodingException)`.
- `encode_legacy_charsets.sql` (`MinSparkVersion: 4.0`, `spark.sql.legacy.javaCharsets=true`): `windows-1252` and `Shift_JIS` both succeed, demonstrating the flag rather than only asserting it is carried. Rows are chosen to be representable in windows-1252 so the still-strict coding-error action does not fire and the two flags stay independently tested.

Each of the error fixtures leads with a sentinel `query` over valid input. `CometSqlFileTestSuite` enforces that convention and also disables constant folding, so the literal-input error cases really do run through the dispatcher.

spark/src/test/resources/sql-tests/expressions/string/encode.sql
andygrove: Restored UTF-16LE and added UTF-32 in f4cdc33e9, with a comment recording why UTF-16LE is not redundant with UTF-16 (BOM) or UTF-16BE (byte order). All seven `VALID_CHARSETS` entries now have coverage. The stale "six charsets" wording in the description is fixed.

docs/source/user-guide/latest/expressions.md
andygrove: Verified by regeneration rather than by hand. I ran `GenerateDocs` (the same invocation `dev/generate-release-docs.sh` uses) against a scratch copy of `docs/source/user-guide/latest` under three profiles:

| profile | generated `encode` cell |
| --- | --- |
| spark-3.5 | `Codegen dispatch` |
| spark-4.0 | `—` |
| spark-4.1 | `—` |

The committed cell matches the 4.x generation exactly, so the file is unchanged. The 3.5 run is also a useful independent check that `CometEncode` really is keyed on `classOf[Encode]` in that profile.

One thing the run surfaced that is out of scope here: the tracked file has pre-existing drift on unrelated rows. Under 4.0 and 4.1 the generator emits `Native` for `<<`, `>>`, `>>>` (committed `—`) and `—` for `to_char`, `to_varchar` (committed `Codegen dispatch`); 4.1 additionally differs on `hour`, `minute`, `second`, `make_timestamp`. Since the published site regenerates into a temp tree at build time, this only affects the in-repo copy. I left it alone rather than adding unrelated churn to this PR, but it may be worth a separate cleanup, or a CI check that the committed column matches generation.

spark/src/main/spark-3.x/org/apache/comet/serde/CometEncode.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package org.apache.comet.serde
+
+import org.apache.spark.sql.catalyst.expressions.Encode
+
+/**
+ * Spark 3.x `encode(str, charset)` runs through the codegen dispatcher so Spark's own encoder
+ * handles charset selection and malformed-input behavior. Dual of `CometStringDecode`.
+ */
+object CometEncode extends CometCodegenDispatch[Encode]
```
rich7420, line 28: Could you enable the existing `encode` benchmark on Spark 3.x and share dispatch-on/off results? Its `isSpark40Plus` gate currently skips the versions this PR adds support for.

andygrove: Ungated in 17c2d653f, and you were right that the gate was pointed at exactly the wrong versions — `encode(utf-8)` and `to_binary(utf-8)` were skipped on 3.x, which is the only place this PR changes anything.

The gate was written when both cases only reached the dispatcher via `StaticInvoke`. That is still how 4.0+ gets there (`Encode` is `RuntimeReplaceable`), but on 3.x it stays an ordinary `Encode` expression and `CometEncode` routes it, so there is something to measure either way. Both cases now run unconditionally and the comment says which route each version takes.

### Results, Spark 3.5.9 / JDK 17 / `local[1]`, best of the timed iterations

**1,048,576 rows**

| case | arm | best | per row |
| --- | --- | ---: | ---: |
| `encode(utf-8)` | dispatch off (Spark fallback) | 62 ms | 58.9 ns |
| | **codegen dispatch** | **61 ms** | **58.0 ns** |
| | Spark (Comet disabled) | 78 ms | 74.4 ns |
| | dispatch off (repeat) | 59 ms | 56.0 ns |
| `to_binary(utf-8)` | dispatch off (Spark fallback) | 64 ms | 60.8 ns |
| | **codegen dispatch** | **59 ms** | **56.3 ns** |
| | Spark (Comet disabled) | 77 ms | 73.2 ns |
| | dispatch off (repeat) | 64 ms | 61.2 ns |

**1,024 rows** (sub-batch): `encode` 5 ms dispatch vs 5 ms fallback vs 9 ms Spark; `to_binary` 5 / 4 / 9.

### Reading them honestly

For `encode` the repeated baseline is 59 ms against the first baseline's 62 ms, so this machine's noise floor for that table is ~3 ms — larger than the 1 ms between dispatch-on and dispatch-off. **The right conclusion is that dispatch is performance-neutral there, not that it is 1 ms faster.** `to_binary` has a 0 ms spread between its two baselines and a 5 ms gap (~8%), which is outside the noise, so that one is a real if modest win.

What both show is that routing through the dispatcher costs nothing relative to falling the whole projection back, and that either Comet arm beats Comet-disabled Spark by ~1.25x. That is the expected shape: both arms run Spark's own `Encode` implementation, so what is being traded is bridge overhead against losing the operator.

The run produced no `WARNING` lines, which matters more than the timings — `checkPlans` verifies per case that the dispatch arm is fully Comet native, that the dispatcher actually compiled or cache-hit a kernel, and that the dispatch-off arm is *not* fully native. So on 3.5 the two arms really are different plans and `CometEncode` is doing the work.

The `to_time(fmt)` case is still gated on 4.1, correctly, and the skip note now lists only that.

mbutrovich (approved):
LGTM, do you mind updating the PR description to reflect the current state? I don't see any changes to `statics.scala` anymore. Thanks @andygrove!

andygrove:
> LGTM, do you mind updating the PR description to reflect the current state? I don't see any changes to `statics.scala` anymore. Thanks @andygrove!

Updated. Thanks for the review @mbutrovich!

PR #4888, opened 2026-07-10
chore: upgrade to DataFusion main + arrow/parquet 59.1.0 [WIP]

andygrove:
## Summary

WIP draft to track the upgrade to DataFusion main (currently pinned to `179b32c9b60103d9c4e6a4364f10f6286c963904`, dated 2026-08-04). Bumps arrow/parquet from 58.4.0 to 59.1.0 to match what DataFusion main uses. `object_store` stays at 0.13.2 (DF main is still on that).

Not ready for merge, opened as a draft so CI runs against latest DF main and breaking changes are surfaced continuously. The pin is bumped periodically to keep pace with main.

## Changes

- `native/Cargo.toml`: swap `datafusion`, `datafusion-datasource`, `datafusion-physical-expr-adapter`, `datafusion-spark` from crates.io `54.1.0` to `git = "https://github.com/apache/datafusion", rev = "..."`
- `native/core/Cargo.toml`: same treatment for the dev-dependency `datafusion-functions-nested`
- Bump `arrow` and `parquet` workspace pins to `59.1.0`
- Adapt native code to the breaking changes surfaced by the upgrade
- Put the native Iceberg scan behind an off-by-default `iceberg-scan` feature (see below)

## Adaptations made

Compile breakage from pinning to DataFusion main + arrow 59 has been worked through:

- **`GroupsAccumulator::convert_to_state` is now a required trait method** and `supports_convert_to_state()` has been removed. That gate is what previously kept skip-partial-aggregation away from Comet's accumulators, so stubbing `convert_to_state` with `not_impl_err` would now fail at runtime once the probe fires — a behavior regression, not just a compile fix. It is instead implemented for all 13 accumulators (avg, avg_decimal, correlation, covariance, hll_plus_plus, percentile, stddev, sum_decimal, sum_int ×3, variance, merge_as_partial), via a shared `convert_to_state_per_row` helper that accumulates each row into its own group on a freshly created accumulator, so the emitted state columns match `state()` by construction. `avg` builds its state arrays directly because it holds a non-cloneable `avg_fn`. `MergeAsPartialGroupsAccumulator` returns its input verbatim, since that input is already the inner accumulator's intermediate state.
- Drop `opt_filter` from `GroupsAccumulator::merge_batch` impls and their delegating call sites to match the upstream signature change (avg, avg_decimal, correlation, covariance, hll_plus_plus, percentile, stddev, sum_decimal, sum_int, variance, merge_as_partial).
[...]

PR #5038, opened 2026-07-25
perf: reuse Arrow IPC compression context across shuffle blocks

### LINKED ISSUE #5008: perf: reuse Arrow IPC CompressionContext across shuffle blocks instead of per-batch allocation
## Background

Follow-on from review of #5006, which encodes the shuffle IPC schema once per writer instead of per block.

In `ShuffleBlockWriter::encode_ipc_stream` (`native/shuffle/src/writers/shuffle_block_writer.rs`), the fast path constructs a fresh `CompressionContext` (an alias of arrow-ipc's `IpcWriteContext`) for every batch:

```rust
let mut compression_context = CompressionContext::default();
let (encoded_dictionaries, encoded_batch) = data_gen.encode(
    batch,
    &mut dictionary_tracker,
    &options,
    &mut compression_context,
)?;
```

Unlike the other per-batch objects on this path (`IpcDataGenerator` is zero-sized; `DictionaryTracker` does no heap allocation until used and must be fresh per standalone stream), `CompressionContext` holds reusable buffers: a `FlatBufferBuilder` and, for zstd, a `Compressor`. arrow-ipc's own documentation describes it as intended to be reused across writes to avoid re-initializing the context (and re-allocating the flatbuffer builder / zstd compressor) for every block.

## Proposal

Reuse a single `CompressionContext` across the blocks written by a given partition writer.

The wiring is not trivial: `CompressionContext` is not `Clone` (the zstd `Compressor` isn't), while `ShuffleBlockWriter` is `Clone` and shared across partitions via `Borrow`, and `write_batch` takes `&self`. So the context can't simply become a field on `ShuffleBlockWriter`. The natural owner is the per-partition `BufBatchWriter` (the actually-mutable, per-partition writer); reuse means threading a `&mut CompressionContext` down through `write_batch` and its call sites.

## Related

Part of the shuffle-writer optimization umbrella #5002.

peterxcli:

Closes #5008.

## Rationale for this change

Avoid recreating Arrow IPC compression state for every shuffle block.

Reuse one `CompressionContext` per buffered partition writer and across batches in the JVM row conversion path. Update benchmarks and callers accordingly.

Existing shuffle IPC round-trip, buffered writer, and shuffle scan tests pass. The shuffle writer benchmark also compiles successfully.


andygrove:
Thanks @peterxcli. Can you share any benchmark results?

peterxcli:
@andygrove this would take effect only after https://github.com/apache/datafusion-comet/pull/4888 is merged.
and the reason is:

---
with arrow 58.4:

Production path | Baseline / patch blocks | FlatBuffer allocations per block | FlatBuffer bytes per block | Context allocations
-- | -- | -- | -- | --
JVM sort → row.rs | 2,688 / 2,441 | 11 / 11 | 4,136 / 4,136 | 0 / 0
Native → BufBatchWriter | 5,857 / 4,298 | 11 / 11 | 4,136 / 4,136 | 0 / 0

with arrow 59.1:

1,000 blocks | Fresh context | Reused context
-- | -- | --
Allocation calls | 212,000 | 201,011
Requested bytes | 8,508,496,000 | 8,504,364,136

- In Arrow 58.4, `CompressionContext` contains only an optional zstd compressor—no `FlatBufferBuilder`: [compression.rs (lines 25–34)](https://github.com/apache/arrow-rs/blob/0ff81c1215cc026a1de93ce3d2078df1ecba6f09/arrow-ipc/src/compression.rs#L25-L34)
- Comet does not enable Arrow IPC compression, so that compressor remains uninitialized: https://github.com/apache/datafusion-comet/blob/8c504872a1d215a6f454eba841969e5a3a85cd94/native/shuffle/src/writers/shuffle_block_writer.rs#L107-L110
- Arrow 58.4 independently constructs a new FlatBuffer builder for every batch: [writer.rs (lines 520–526)](https://github.com/apache/arrow-rs/blob/0ff81c1215cc026a1de93ce3d2078df1ecba6f09/arrow-ipc/src/writer.rs#L520-L526).
- The review description in https://github.com/apache/datafusion-comet/pull/5006#discussion_r3640605171 matches Arrow 59.1’s newer `IpcWriteContext`, which owns the builder: [compression.rs (lines 27–38)](https://github.com/apache/arrow-rs/blob/b1de629ae82ee417e24a413f8f6815c884ec971b/arrow-ipc/src/compression.rs#L27-L38).

peterxcli:
@andygrove I've publish the benchmark, would appreciate it if you could take a look at this and see if we can get this merged.

sunchao (approved):
LGTM. The mutable context stays with the buffered writer or row-conversion loop, while dictionary tracking and outer compression frames remain independent for each block.

I also compared the base and head writers using a focused Arrow 58.4 probe covering all four codecs, dictionary and schema changes, repeated flushes, and output failures. All seven tests passed. The Arrow-version dependency of the performance benefit is already covered in the benchmark discussion.

peterxcli:
@sunchao thanks for the review! Do you think this is ready to merge now?

sunchao:
Merged, thanks @peterxcli ! 

I think later on we also need to bump Arrow to 59.1+ in order to get the allocation savings.

In Arrow 58.4.0, its `CompressionContext` contains only an optional zstd compressor, which Comet never initializes because Arrow IPC buffer compression is disabled. The `FlatBuffer` builder is allocated separately on every batch. Moving this unused context outside the loop therefore saves no allocations today. Comet’s outer compression encoder is still recreated per block.

Starting with Arrow 59.1, the context owns the reusable `FlatBuffer` builder. At that point, this change genuinely reduces repeated metadata-allocation work.

PR #5040, opened 2026-07-26
chore: Use native impl of elt function

kazantsev-maksim:

- N/A

## Rationale for this change

This PR migrates the elt string function from Comet's codegen-based dispatch path to the shared datafusion-spark implementation, registering it as a native scalar UDF.

- Register the SparkElt scalar UDF in register_datafusion_spark_function so it is available to the DataFusion session context alongside the other Spark-compatible UDFs.
- Change `CometElt` from `CometCodegenDispatch[Elt]` to `CometScalarFunction[Elt]("elt")`

Significantly expand the test coverage with many additional inputs to exercise edge cases


comphead (approved):
Thanks @kazantsev-maksim 

CI pending

kazantsev-maksim:
Thanks @comphead!

andygrove (changes requested):
I don't think the upstream `elt` implementation is Spark-compatible? It doesn't implement ANSI mode, for example.

From DF 54.1.0:

```rust
        // TODO: if spark.sql.ansi.enabled is true,
        //  throw ArrayIndexOutOfBoundsException for invalid indices;
        //  if false, return NULL instead (current behavior).
        if index < 1 || (index as usize) > num_values {
            builder.append_null();
            continue;
        }
```

andygrove:
> I don't think the upstream `elt` implementation is Spark-compatible? It doesn't implement ANSI mode, for example.
> 
> From DF 54.1.0:
> 
> ```rust
>         // TODO: if spark.sql.ansi.enabled is true,
>         //  throw ArrayIndexOutOfBoundsException for invalid indices;
>         //  if false, return NULL instead (current behavior).
>         if index < 1 || (index as usize) > num_values {
>             builder.append_null();
>             continue;
>         }
> ```

One option would be to mix in `CodegenDispatchFallback` and return `Incompatible` when `expr.failOnError` is true. It would be good to file an issue upstream for implementing ANSI support too.

comphead:
> > I don't think the upstream `elt` implementation is Spark-compatible? It doesn't implement ANSI mode, for example.
> > From DF 54.1.0:
> > ```rust
> >         // TODO: if spark.sql.ansi.enabled is true,
> >         //  throw ArrayIndexOutOfBoundsException for invalid indices;
> >         //  if false, return NULL instead (current behavior).
> >         if index < 1 || (index as usize) > num_values {
> >             builder.append_null();
> >             continue;
> >         }
> > ```
> 
> One option would be to mix in `CodegenDispatchFallback` and return `Incompatible` when `expr.failOnError` is true. It would be good to file an issue upstream for implementing ANSI support too.

Thanks @andygrove for catching this. for ANSI we should start an epic there. The ANSI flag is already there, and we can try it on `elt`

comphead:
Created https://github.com/apache/datafusion/pull/23928

andygrove:

There is already `CHANGES_REQUESTED` feedback on this, so I will keep to points I do not think have been covered.

**No `BinaryType` coverage**

Spark's `Elt` accepts `StringType` or `BinaryType` inputs and returns the matching type. Both `elt.sql` and `elt_ansi.sql` appear to test only string arguments. Does `datafusion-spark`'s `SparkElt` handle binary, and does the serde produce the right return type for it?

This matters more than usual because `CometScalarFunction[Elt]("elt")` does not set an explicit return type, so the native planner resolves it from the registry. If binary is unsupported or mistyped, the failure is a native execution error rather than a fallback. A couple of `elt(i, X'01', X'02')` rows would settle it either way.

**No performance evidence**

The whole point of moving from the dispatcher to a native UDF is speed, and the description has no numbers. What does `elt` cost through the dispatcher versus natively for a realistic batch? Without that, this is a change that gives up exact-Spark behavior under ANSI for an unquantified benefit. Even a rough microbenchmark would make the trade explicit.

**Wildcard import**

```scala
-import org.apache.spark.sql.types.{BinaryType, DataTypes, IntegerType, LongType, StringType}
+import org.apache.spark.sql.types._
```

`strings.scala` uses explicit imports throughout. Swapping to a wildcard makes it harder to see what the file depends on and can silently shadow names later. Whatever new type is needed can just be added to the list.

**68 changed lines in `expressions.md`**

That is a lot of churn for one expression's status. If most of it is table realignment rather than content, could it be reverted so the diff shows only the `elt` row? It makes the change much easier to review and avoids conflicts with the other PRs currently touching that file.

PR #5042, opened 2026-07-26
perf: optimize levenshtein function

kazantsev-maksim:

- N/A

## Rationale for this change

Optimize the existing native `spark_levenshtein` implementation. The previous implementation allocated fresh `Vec<i32>` scratch rows for every row pair and always walked the string as `chars()`, even for pure-ASCII input. Two changes remove both costs:

- A thread-local scratch-buffer pool reused across row evaluations, so the distance matrices are no longer heap-allocated inside the processing loop.
- An ASCII fast path that operates directly on `&[u8]` when both operands are ASCII, falling back to the `chars()` path only for non-ASCII input.

- **Thread-local scratch buffers**, accessed through `with_scratch_buffers`. The buffers live in `thread_local!(LEVENSHTEIN_SCRATCH)` and are reused across rows, so the DP rows are not reallocated per pair.
- **Bounded TLS retention.** The buffers are capped at `MAX_RETAINED_CAPACITY = 1024` elements (4 KB each). Rows requiring a larger buffer allocate call-local vectors instead, so the TLS buffers can never grow past the cap and cannot retain a large allocation after a pathological row. This addresses the earlier P2 finding on unbounded thread-local memory.
- **Full re-initialization on reuse.** The scratch path does `clear()` followed by `resize(len, default_val)`, so no cell can carry a value from a previous row. The old code relied on a fresh `vec![out_of_band; m + 1]`; with reuse, stale values would otherwise leak into cells outside the threshold band.
- **ASCII fast path** (`s.is_ascii() && t.is_ascii()`) in both the unthresholded and thresholded distance functions, operating on raw UTF-8 bytes.
- **Threshold-band `curr[0]` fix.** When `j > threshold`, `curr[0]` is now explicitly set to `out_of_band` in both the ASCII and Unicode paths. Previously a stale `curr[0]` admitted an invalid insertion path, e.g. `spark_levenshtein("a", "bb", 1)` returned `1` instead of `-1`.
- **Criterion benchmark** extended with `non-ascii` and `mixed` datasets in addition to the existing `ascii` dataset.
- Regenerated nothing else; the public signature and Spark semantics are unchanged.

Existing tests.

Benchmark (criterion):

Baseline is `main`; each cell is the criterion median over 100 samples with the percentage change against the baseline. All deltas are statistically significant (p = 0.00 < 0.05).

### ASCII (kitten / sitting)

| Batch | Profile | Median time | Change vs main |
| --- | --- | --- | --- |
| 8192 | no_nulls | 1.245 ms | -45.0% |
| 8192 | sparse | 1.147 ms | -65.1% |
| 8192 | all_null | 20.93 us | -30.7% |
| 65536 | no_nulls | 10.46 ms | -39.3% |
| 65536 | sparse | 9.13 ms | -41.9% |
| 65536 | all_null | 172.2 us | -30.2% |
| 524288 | no_nulls | 79.05 ms | -41.8% |
| 524288 | sparse | 73.53 ms | -40.6% |
| 524288 | all_null | 1.330 ms | -32.0% |

### Non-ASCII (naive / smorgas)

| Batch | Profile | Median time | Change vs main |
| --- | --- | --- | --- |
| 8192 | no_nulls | 1.858 ms | -2.8% |
| 8192 | sparse | 1.642 ms | -5.7% |
| 8192 | all_null | 20.63 us | -31.3% |
| 65536 | no_nulls | 14.65 ms | -3.8% |
| 65536 | sparse | 13.21 ms | -5.8% |
| 65536 | all_null | 168.0 us | -31.1% |
| 524288 | no_nulls | 115.1 ms | -6.3% |
| 524288 | sparse | 105.1 ms | -7.3% |
| 524288 | all_null | 1.316 ms | -32.1% |

### Mixed (ascii left / non-ascii right)

| Batch | Profile | Median time | Change vs main |
| --- | --- | --- | --- |
| 8192 | no_nulls | 710.4 us | -23.7% |
| 8192 | sparse | 652.4 us | -24.4% |
| 8192 | all_null | 20.51 us | -31.4% |
| 65536 | no_nulls | 5.661 ms | -23.9% |
| 65536 | sparse | 5.262 ms | -24.0% |
| 65536 | all_null | 165.7 us | -31.1% |
| 524288 | no_nulls | 45.19 ms | -24.1% |
| 524288 | sparse | 42.04 ms | -24.1% |
| 524288 | all_null | 1.312 ms | -31.9% |

#### Notes on the results

- The `non-ascii` row is the one that matters for the fast path: `s.is_ascii()` adds one full scan before falling into `chars()`, but this is more than offset by scratch-buffer reuse, so non-ASCII inputs improve (-2.8% .. -7.3%) rather
  than regress.
- Absolute times are not comparable *across* datasets, because string lengths differ between them; only the change *within* a dataset is meaningful.
- `all_null` is essentially flat across the three datasets (~ -31%). A fully-null batch runs no DP, no scratch reuse and no fast path, so this delta is consistent with baseline/environment movement rather than the code change and should be read with that in mind.
- Some `non-ascii` and large-batch `ascii` cases need a longer target time or a lower sample count than criterion's 5 s default.


native/spark-expr/src/string_funcs/levenshtein.rs
```diff
@@ -26,67 +26,188 @@ use datafusion::common::{cast::as_generic_string_array, DataFusionError, Result}
 use datafusion::physical_plan::ColumnarValue;
 use std::sync::Arc;
 
+// Thread-local scratch buffers to avoid heap allocations in the row processing loop
+thread_local! {
+    static LEVENSHTEIN_SCRATCH: std::cell::RefCell<(Vec<i32>, Vec<i32>)> =
+        std::cell::RefCell::new((Vec::with_capacity(64), Vec::with_capacity(64)));
```
sunchao, line 36: [P2] Bound the retained capacity of the scratch vectors

Could you cap the cached capacity or keep these buffers at batch/task scope? `Vec::resize` reduces the length but never releases capacity, so moving the vectors into TLS retains the largest allocation until the worker thread exits, including after the query finishes. I ran the exact-head kernels with two equal 16,000,000-character ASCII strings and threshold 0, then dropped the input and evaluated a one-character pair. The two vectors still retained 128,000,008 bytes after that small call. These are valid inputs and the thresholded case completes in linear time. Comet's long-lived worker threads can therefore retain this allocation independently, outside the query memory pool. The previous call-local vectors released it on return.

sunchao (dismissed):
The existing [P2] [retained thread-local scratch memory finding](https://github.com/apache/datafusion-comet/pull/5042#discussion_r3837586414) still applies to this head and needs to be addressed before merge. This approval does not mark that finding resolved.

The exact-head kernel passed 206,090 comparisons against an independent reference. The retained-capacity reproduction remains valid. GitHub CI is still marked `action_required`, so there is no successful full CI result for this head.

andygrove:

Thanks for working on this. The ASCII fast path is a good idea and the benchmark numbers look worthwhile. I think there is a correctness problem in the scratch-buffer reuse, though, and I would like to work through it before this goes further.

**`Vec::resize` does not reset existing elements**

`resize` only fills the *new* slots when growing, and simply truncates when shrinking. Existing elements keep whatever the previous call left in them. That matters in `levenshtein_distance_with_threshold`, where the old code allocated `vec![out_of_band; m + 1]` and therefore started from a fully initialized row, but the new code only writes some of the cells.

Concretely, in the threshold path:

```rust
prev.resize(m + 1, out_of_band);
for (i, value) in prev.iter_mut().enumerate().take(m.min(threshold) + 1) {
    *value = i as i32;
}
```

If the previous row used a longer buffer, `resize` truncates and leaves `prev[0..=m]` holding stale distances. The loop then initializes only `prev[0..=min(m, threshold)]`. With `threshold < m`, the cells from `threshold + 1` through `m` should be `out_of_band` but instead hold values from an earlier, unrelated pair of strings.

The same applies to `curr`. Inside the `j` loop only `curr[0]`, `curr[start - 1]`, `curr[start..=end]`, and `curr[end + 1]` are written. Everything outside that band used to be `out_of_band` from the fresh allocation and is now stale.

So `levenshtein(s, t, threshold)` can return a different answer depending on what the *previous* row on the same thread computed. That is the worst kind of bug to find in production: nondeterministic, data-dependent, and invisible in a test that evaluates one pair at a time.

The fix is to explicitly fill rather than rely on `resize`:

```rust
prev.clear();
prev.resize(m + 1, out_of_band);
```

or `prev.resize(m + 1, out_of_band); prev.fill(out_of_band);` before the partial initialization. Same for `curr`. The plain `levenshtein_distance` path happens to be safe because both rows are fully overwritten, but it would be worth making the initialization uniform in both so the invariant is not something a future reader has to rediscover.

**"Existing tests" is not enough for this change**

Because the bug only appears when one call leaves state behind for the next, no single-pair test can catch it. Could you add a test that evaluates a whole column in one `invoke`, mixing long and short strings and ASCII with non-ASCII, and compares against a straightforward reference implementation? Something like: compute a batch in one order, compute the same pairs in reverse order, and assert the results match. That property directly targets the class of bug above and would have caught it.

It would also be worth adding cases with a small threshold and a long shorter-string, since that is the shape that leaves the most cells uninitialized.

**Two smaller points**

`s.is_ascii()` is a full scan of the string, so non-ASCII inputs now pay an extra pass before falling into the `chars()` path. For a mostly-non-ASCII column that is a small regression. Probably fine, but worth confirming the benchmark covers a non-ASCII case so we know the fallback did not get slower.

The `thread_local!` buffers grow to the longest string ever seen on that thread and never shrink. On a pool of executor threads that could retain a fair amount of memory after one pathological row. Is it worth calling `shrink_to` when the buffer is far larger than the current need?

native/spark-expr/src/string_funcs/levenshtein.rs
kazantsev-maksim: Thanks for pointing this out. I have addressed the retained thread-local scratch memory finding (P2):

**Retained scratch buffer memory cap:** 
   Added an upper bound check (`MAX_RETAINED_CAPACITY = 1024`, ~4 KB). If the buffer capacity exceeds this limit due to an occasional pathologically long string, it gets reset/shrunk back to the bounded capacity on subsequent evaluations via `prepare_scratch`:
   ```rust
   const MAX_RETAINED_CAPACITY: usize = 1024;

   #[inline]
   fn prepare_scratch(buf: &mut Vec<i32>, len: usize, default_val: i32) {
       if buf.capacity() > MAX_RETAINED_CAPACITY && len <= MAX_RETAINED_CAPACITY {
           *buf = Vec::with_capacity(MAX_RETAINED_CAPACITY);
       }
       buf.clear();
       buf.resize(len, default_val);
   }

native/spark-expr/src/string_funcs/levenshtein.rs
```diff
@@ -26,67 +26,195 @@ use datafusion::common::{cast::as_generic_string_array, DataFusionError, Result}
...
+                    return -1;
+                }
+
+                if start > 1 {
+                    curr[start - 1] = out_of_band;
+                }
+
+                if j <= threshold {
+                    curr[0] = j as i32;
+                }
```
sunchao, line 206: [P1] Reset `curr[0]` when it leaves the threshold band

When `j == threshold + 1`, `start == 1`, so neither branch resets `curr[0]`; its old value admits an invalid insertion path. On `8456bc5`, `spark_levenshtein` returns `1` for (`"a"`, `"bb"`, `1`) instead of `-1` (distance `2`), causing false matches. This reproduces through the public Rust API for scalar, array, and mixed Utf8/LargeUtf8 inputs; the base returns `-1`. Please set `curr[0]` to `out_of_band` when `j > threshold` and add this regression case. Validation is Rust component/API execution, not an end-to-end Spark run.

native/spark-expr/src/string_funcs/levenshtein.rs
sunchao: I rechecked `8456bc5`; this is only partly addressed. `prepare_scratch` shrinks buffers only on a later small, nonempty ASCII calculation. Both Unicode paths still use `resize`, and empty inputs return before cleanup. Processing equal 16,000,000-character ASCII strings with threshold `0`, dropping the inputs, then evaluating a one-character Unicode pair and an empty ASCII pair still leaves 128,000,008 scratch bytes allocated. An oversized final row also remains attached to the thread after evaluation. Could we use temporary buffers above the cap or release oversized buffers after evaluation on both paths? This was verified with Rust kernel/capacity probes, not an end-to-end Spark run.

native/spark-expr/src/string_funcs/levenshtein.rs
kazantsev-maksim: Thank you for the thorough review and catching the `curr[0]` edge case! I have addressed both findings:

1. **[P1 Fix] Explicitly reset `curr[0]` outside threshold band:**
   Replaced the condition with an explicit branch to set `curr[0] = out_of_band` when `j > threshold`:
   ```rust
   if j <= threshold {
       curr[0] = j as i32;
   } else {
       curr[0] = out_of_band;
   }

kazantsev-maksim: ### Criterion Benchmark Results (vs `main` Baseline)

| Batch Size | Data Profile | Time (mean) | Change vs `main` | Status |
| :--- | :--- | :--- | :--- | :--- |
| **8,192** | `no_nulls` | 1.28 ms | **-35.60%** | Improved |
| | `sparse` | 1.20 ms | **-36.68%** | Improved |
| | `all_null` | 24.04 µs | **-21.52%** | Improved |
| **65,536** | `no_nulls` | 10.11 ms | **-39.54%** | Improved |
| | `sparse` | 9.11 ms | **-40.06%** | Improved |
| | `all_null` | 182.18 µs | **-26.20%** | Improved |
| **524,288** | `no_nulls` | 78.79 ms | **-42.06%** | Improved |
| | `sparse` | 72.63 ms | **-42.06%** | Improved |
| | `all_null` | 1.42 ms | **-26.96%** | Improved |

kazantsev-maksim:
Thanks @andygrove  — agreed on both counts, both are addressed in the current head.

1. Residual state in the scratch buffers: the scratch path now fully re-initializes on every call (`clear()` + `resize(len, default_val)`), so no cell can carry a value from a previous row. There is a regression test that runs a whole column in one `invoke` and asserts forward-order == reversed-order results, plus the small-threshold / long-string shape (`"a"`, `"bb"`, `1`).

2. Non-ASCII benchmark coverage: added `non-ascii` and `mixed` datasets to the bench alongside the existing `ascii` one. On the current head, non-ASCII is *not* slower — it improves by 2.8-7.3% (p < 0.05), i.e. the extra `is_ascii()` scan is paid back by scratch reuse. Full numbers are in the PR description.

One caveat I want to be explicit about: the `all_null` row moves ~-31% on all three datasets, but that case does no DP and no scratch reuse at all, so I don't attribute that delta to this change.

Also worth noting for context: the thread-local buffers are now capped (`MAX_RETAINED_CAPACITY`), and rows larger than the cap allocate call-local vectors instead, which is what bounds the retained TLS memory from the earlier
P2 finding.

native/spark-expr/src/string_funcs/levenshtein.rs
kazantsev-maksim: Thanks. That finding applied to the `prepare_scratch` revision (8456bc5). On the current head the scratch handling was replaced by `with_scratch_buffers`, which addresses all four points directly:

- Rows larger than `MAX_RETAINED_CAPACITY` (1024) do not use the thread-local buffers at all; they allocate call-local `vec![..; len]` that are dropped when the call returns. So an oversized row never attaches to the thread.
- Both the ASCII and the Unicode paths, thresholded and not, go through the same helper, so Unicode no longer diverges.
- Empty inputs return before touching the buffers, and because oversized cells never enter TLS, there is nothing left to clean up on that path.

The maximum retained TLS is therefore bounded at `2 * 1024 * 4 B = 8 KB` per worker thread. I added a capacity-probe test that runs a 16,000,000-character row and asserts the cached capacities stay within the cap, which captures your replacement scenario.

If you were reviewing 8456bc5 specifically, could you're-check the current head?

PR #5044, opened 2026-07-26
perf: track decimal overflow without rescanning results

### LINKED ISSUE #4943: Optimize `WideDecimalBinaryExpr`: skip the null-masking pass in non-ANSI mode when nothing overflows
Part of #4936.

## Problem

`WideDecimalBinaryExpr` implements wide (128-bit intermediate) decimal `+ - *`, which runs on the hot path of most TPC-DS queries. It computes the result in a single fused `try_binary` pass (`native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs`), writing `i128::MAX` as an overflow sentinel in non-ANSI mode. It then unconditionally runs `null_if_overflow_precision` to convert sentinels into nulls:

```rust
let result = if eval_mode != EvalMode::Ansi {
    result.null_if_overflow_precision(p_out)
} else {
    result
};
```

`null_if_overflow_precision` is a second full pass that allocates a new array, and it runs even when no value overflowed, which is the common case.

## Proposed change

Run the masking pass only when a sentinel is actually present, mirroring what was done for `DecimalRescaleCheckOverflow` in #4938:

```rust
let result = if eval_mode != EvalMode::Ansi && result.values().contains(&i128::MAX) {
    result.null_if_overflow_precision(p_out)
} else {
    result
};
```

`contains` short-circuits at the first sentinel, so the overflow path is unaffected while the common no-overflow path skips the extra allocation. ANSI mode never produces a sentinel (it errors during the `try_binary` pass) and is unchanged.

Output stays bit-identical: when no sentinel is present, `null_if_overflow_precision` would have nulled nothing, so skipping it yields the same array.

## Notes

This is the non-ANSI counterpart to the `CheckOverflow` ANSI fast path in #4937. A review of the other decimal expressions found no remaining ANSI-detection asymmetry: `DecimalRescaleCheckOverflow` (#4938), `WideDecimalBinaryExpr`, and the decimal casts already do their ANSI overflow check inside a single cheap pass.

A criterion benchmark (no-overflow, no-overflow-with-nulls, sparse-overflow, dense-overflow, and an ANSI shape) should accompany the change to confirm the win and that overflow shapes do not regress.

peterxcli:

Closes #4943.

## Rationale for this change

`WideDecimalBinaryExpr` previously allocated a null-masked result for every non-ANSI batch, even when nothing overflowed. The expression already knows when overflow occurs while evaluating each value. Recording that state during evaluation avoids rescanning the result and skips null masking entirely for no-overflow batches.

- Track overflow during `WideDecimalBinaryExpr` binary arithmetic using a per-evaluation `Cell<bool>`.
- Run the allocating null-masking pass only when overflow actually occurs.
- Add Scalar × Scalar legacy-overflow regression coverage.
- Add Criterion cases covering no overflow, nulls, overflow density, overflow at the end of a batch, and ANSI mode.

An earlier revision of this PR applied the same approach to `DecimalRescaleCheckOverflow`; that expression was since rewritten to use Arrow's decimal cast in #5440 (closing #5094), which removes the sentinel scan and masking pass entirely, so this PR now only changes `WideDecimalBinaryExpr`.

- `cargo test --manifest-path native/Cargo.toml -p datafusion-comet-spark-expr wide_decimal_binary_expr`
- `cargo clippy --manifest-path native/Cargo.toml -p datafusion-comet-spark-expr --all-targets -- -D warnings`
- `cargo check --manifest-path native/Cargo.toml -p datafusion-comet-spark-expr --bench wide_decimal`
- `git diff --check`

Paired Criterion runs used two independent comparisons on Apple Silicon.

### `WideDecimalBinaryExpr`

Compared with `1104c78e^`:

| Shape | Before | After | Change |
| --- | ---: | ---: | ---: |
| No overflow | 60.63 µs | 54.52 µs | 10.1% faster |
| No overflow + nulls | 67.70 µs | 58.69 µs | 13.3% faster |
| Sparse overflow | 58.18 µs | 59.11 µs | 1.6% slower |
| Dense overflow | 58.47 µs | 61.39 µs | 5.0% slower |
| Overflow at end of batch | 58.78 µs | 60.95 µs | 3.7% slower |
| ANSI, no overflow | 54.32 µs | 54.53 µs | Unchanged |

### `DecimalRescaleCheckOverflow` (historical)

These numbers measured an earlier revision of this PR against the pre-#5440 custom kernel, which no longer exists; they are kept for the review record only.

Compared with `07944c8a^`:

| Shape | Before | After | Change |
| --- | ---: | ---: | ---: |
| Scale up, no overflow | 10.57 µs | 7.52 µs | 28.8% faster |
| Scale up, no overflow + nulls | 16.33 µs | 13.61 µs | 16.7% faster |
| Scale down, no overflow | 34.09 µs | 30.99 µs | 9.1% faster |
| Sparse overflow | 13.99 µs | 14.12 µs | 0.9% slower |
| Dense overflow | 16.27 µs | 17.15 µs | 5.4% slower |
| ANSI scale up, no overflow | 7.33 µs | 7.33 µs | Unchanged |

The common no-overflow path improves by 10.1% to 13.3%. Overflow-bearing batches regress by up to 5.0%; this is an accepted trade-off for the common-path improvement, and the cost comes from the `Cell` write being an observable side effect inside the kernel closure rather than from the number of stores. Follow-up #5309 tracks writing nulls directly in `WideDecimalBinaryExpr` to eliminate its sentinel, `Cell<bool>`, and masking pass. The corresponding `DecimalRescaleCheckOverflow` cleanup was completed in #5440 (closing #5094).


andygrove (commented):
Nice targeted perf fix, and the new bench shapes make the win legible. One small ask: the sibling optimization in `decimal_rescale_check.rs:200` carries a comment explaining that `i128::MAX` is the overflow sentinel and that `contains` short-circuits so no-overflow batches skip the extra allocation. Could we add the same comment on the new guard in `wide_decimal_binary_expr.rs:284` so both call sites read the same way? Without it, a reader landing on this line has to jump to the doc comment on `check_overflow_and_convert` to figure out why `i128::MAX` is the value being probed.

peterxcli:
@andygrove genteelly ping, I've added the comment as your review, do you think this is ok to merge now? Thanks!

andygrove (commented):
Thanks for the update, the comment reads much better alongside the sibling in `decimal_rescale_check.rs` now.

I went through the equivalence argument carefully and I agree the guard is sound. `try_binary` zero-fills null slots rather than leaving stale values, so the scan cannot false-positive on a null. A non-overflowing result is clamped to `±(10^p_out - 1)`, and `10^38 - 1` is about `9.99e37` against `i128::MAX` at roughly `1.70e38`, so a legitimate value can never collide with the sentinel. And `null_if_overflow_precision` nulls exactly the values outside `±(10^p - 1)`, which is exactly the sentinel set. So when no sentinel is present the pass really is a no-op and the output stays bit-identical.

CI is fully green.

I have a few things I would like to resolve before this merges. The main one is that neither new overflow benchmark actually measures the shape that could regress, so we do not yet have evidence that the overflow path is unaffected. Details inline.

native/spark-expr/benches/wide_decimal.rs
```diff
@@ -49,6 +50,30 @@ fn make_decimal_batch(p1: u8, s1: i8, p2: u8, s2: i8) -> RecordBatch {
     RecordBatch::try_new(Arc::new(schema), vec![Arc::new(left), Arc::new(right)]).unwrap()
 }
 
+fn make_overflow_batch(null_every: usize, overflow_every: usize) -> RecordBatch {
```
andygrove, line 53: The two overflow shapes both end up with an overflow at index 0, since `0 % 17 == 0` and `0 % 2 == 0`. That means `contains` returns on the very first element in both cases, so the sparse and dense benches never exercise a scan longer than one element.

Could we add a shape where the only overflow sits at the last index? That is the case where the guard costs the most, because you pay the full 8192-element scan and then still pay the masking pass on top. It would be good to see a number for it. Right now the "No significant change" results for the sparse and dense shapes are really measuring the index-0 case, so they do not tell us much about the overflow path.

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
```diff
@@ -281,7 +281,13 @@ impl PhysicalExpr for WideDecimalBinaryExpr {
             }
         };
 
-        let result = if eval_mode != EvalMode::Ansi {
+        let result = if eval_mode != EvalMode::Ansi && result.values().contains(&i128::MAX) {
```
andygrove: Did you consider tracking the overflow during the arithmetic pass instead of scanning for the sentinel afterwards? `check_overflow_and_convert` already knows the answer at the moment it writes the sentinel, so `contains` is re-deriving a fact we had for free. `try_binary` takes an `Fn`, so a captured `Cell<bool>` works and only gets written on the rare overflow branch.

That would drop the extra 128 KB read from both paths, which should make the no-overflow win larger than the 8-10% you measured, and it removes the regression risk on the overflow shapes entirely.

I realize the tradeoff is losing the symmetry with `decimal_rescale_check.rs`, and the issue text did prescribe `contains`, so I am happy to hear the case for keeping it as is. But the scan is essentially all that is left of the cost here, so it seems worth weighing.

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
```diff
@@ -281,7 +281,13 @@ impl PhysicalExpr for WideDecimalBinaryExpr {
...
+            // The arithmetic pass writes i128::MAX as an overflow sentinel for values that do
+            // not fit the output precision. Only when a sentinel is present do we need the
+            // extra null-masking pass (which allocates a new array); `contains` short-circuits
+            // at the first sentinel, so the common no-overflow case skips that allocation
+            // entirely. ANSI mode raises on overflow and never produces a sentinel, so it also
+            // skips this pass.
```
andygrove: One thing about the wording. "`contains` short-circuits at the first sentinel, so the common no-overflow case skips that allocation entirely" joins two facts that are not actually connected. In the no-overflow case `contains` does not short-circuit at all, it reads the entire buffer. The short-circuit only helps when an overflow is present and found early.

Could we say instead that the no-overflow path trades the allocating masking pass for a single read-only scan? Since this comment is the place a reader learns the cost model, it would help for it to be explicit that the common path is not free. The same wording exists in `decimal_rescale_check.rs`, so it is probably worth fixing in both places.

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
```diff
@@ -506,6 +512,23 @@ mod tests {
         assert!(arr.is_null(0));
     }
 
+    #[test]
+    fn test_overflow_with_nulls_legacy_mode() {
```
andygrove, line 516: Good test, and I like that `test_null_propagation` already covers the complementary shape where the fast path runs and nulls have to survive.

Since the guard now also governs the Scalar x Scalar path, it might be worth adding a companion to `test_scalar_scalar_returns_scalar` that overflows in legacy mode and asserts the result comes back as `ScalarValue::Decimal128(None, p, s)` rather than an array or a raw sentinel. That path routes through `ScalarValue::try_from_array` after the masking pass, so it seems worth pinning down.

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
peterxcli: yes, capture the overflow in `check_overflow_and_convert` makes more sense

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
peterxcli: added. thanks!

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
peterxcli: I change this to follow your direction: https://github.com/apache/datafusion-comet/pull/5044#discussion_r3691138122, so this no longer a problem.

native/spark-expr/benches/wide_decimal.rs
peterxcli: added. thanks!

native/spark-expr/src/math_funcs/internal/decimal_rescale_check.rs
```diff
@@ -199,11 +199,10 @@ impl PhysicalExpr for DecimalRescaleCheckOverflow {
 
                 let result = if !fail_on_error && result.values().contains(&i128::MAX) {
```
mbutrovich: `decimal_rescale_check.rs:200` still runs `result.values().contains(&i128::MAX)` on every batch. In the common no-overflow case that is a full read of the output buffer, the same cost this PR removes from `WideDecimalBinaryExpr` by tracking overflow with a `Cell<bool>` inside the `try_binary` closure instead of scanning for the sentinel afterward.

`DecimalRescaleCheckOverflow::evaluate` has the same sentinel-then-mask shape, just built on `try_unary` instead of `try_binary`. The same closure-capture technique applies there without changes to the surrounding logic. Right now this PR updates the comment on `decimal_rescale_check.rs:201-205` to admit the scan still happens, but leaves the scan itself in place, so one of the two call sites gets the actual fix and the other gets a note explaining why it still pays the cost.

Could you either apply the same `Cell<bool>` guard to `decimal_rescale_check.rs`, or open a follow-up issue for it so the asymmetry is tracked rather than left implicit in a comment? If it is deliberately out of scope for this PR, a one-line note here (or in the PR description) saying so would help.

peterxcli: ok, I've addressed it in this along with this PR, updated the title and description, too.

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
```diff
@@ -276,12 +278,12 @@ impl PhysicalExpr for WideDecimalBinaryExpr {
...
                         raw
                     };
-                    check_overflow_and_convert(result, bound, neg_bound, eval_mode)
+                    check_overflow_and_convert(result, bound, neg_bound, eval_mode, &overflowed)
                 })?
             }
         };
 
-        let result = if eval_mode != EvalMode::Ansi {
+        let result = if overflowed.get() {
```
andygrove, line 286: This dropped the only comment in the codebase explaining why skipping the masking pass is safe. The six lines removed from `decimal_rescale_check.rs` are gone, and neither guard has a replacement, so both call sites are now a bare `if overflowed.get()`.

The invariant is not obvious from reading the guard. It rests on the fact that every non-sentinel value the closure returns is already clamped inside `±(10^p_out - 1)`, which is exactly the set `null_if_overflow_precision` leaves untouched, so the pass can only ever null sentinels. It is also worth saying that ANSI mode returns `Err` before setting the flag, so ANSI still skips the pass for the same reason it did before.

Could we add that back at both sites, here and at `decimal_rescale_check.rs:212`? We worked the argument out in this thread, and it would be good for it to live in the code rather than only in the PR conversation.

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
```diff
@@ -329,20 +331,22 @@ impl PhysicalExpr for WideDecimalBinaryExpr {
...
     neg_bound: i256,
     eval_mode: EvalMode,
+    overflowed: &Cell<bool>,
 ) -> Result<i128, ArrowError> {
     if result > bound || result < neg_bound {
         if eval_mode == EvalMode::Ansi {
             return Err(ArrowError::ComputeError("Arithmetic overflow".to_string()));
         }
         // Sentinel value — will be nullified by null_if_overflow_precision
+        overflowed.set(true);
```
andygrove, line 351: Now that the flag records the overflow, the `i128::MAX` sentinel and the masking pass it feeds are both redundant. Would you be up for filing a follow-up issue to drop them and write the null bit directly at the point the overflow is detected? That would remove the extra pass and the `Cell` side effect, which is what is costing the 3% to 5% on the overflow shapes.

Clearly out of scope here. I would just like it tracked rather than left implicit, and a link to the issue in this thread would be enough.

andygrove (commented):
I measured both benches before and after on this branch to fill in the shapes that are not in the description. Two independent comparisons each, on Apple Silicon.

`benches/wide_decimal.rs`, against `1104c78e^`:

| Shape | Before | After | Change |
| --- | --- | --- | --- |
| no overflow | 60.63 µs | 54.52 µs | 10.1% faster |
| no overflow, nulls | 67.70 µs | 58.69 µs | 13.3% faster |
| sparse overflow | 58.18 µs | 59.11 µs | 1.6% slower |
| dense overflow | 58.47 µs | 61.39 µs | 5.0% slower |
| overflow at end of batch | 58.78 µs | 60.95 µs | 3.7% slower |
| ansi no overflow | 54.32 µs | 54.53 µs | unchanged |

`benches/decimal_rescale.rs`, against `07944c8a^`:

| Shape | Before | After | Change |
| --- | --- | --- | --- |
| scale up, no overflow | 10.57 µs | 7.52 µs | 28.8% faster |
| scale up, no overflow, nulls | 16.33 µs | 13.61 µs | 16.7% faster |
| scale down, no overflow | 34.09 µs | 30.99 µs | 9.1% faster |
| sparse overflow | 13.99 µs | 14.12 µs | 0.9% slower |
| dense overflow | 16.27 µs | 17.15 µs | 5.4% slower |
| ansi scale up, no overflow | 7.33 µs | 7.33 µs | unchanged |

Two things come out of this.

The `decimal_rescale_check.rs` half is the bigger win of the two, at 28.8% on `scale up, no overflow`. `benches/decimal_rescale.rs` already existed with matching shapes, so those numbers come for free. Could we get them into the description? The table as it stands understates what this PR does.

The other one I would like to correct before this merges. I could not reproduce "No stable change across reruns" for the overflow shapes. `dense overflow` and `overflow at end of batch` are consistently 3% to 5% slower at `p = 0.00`, in both expressions. Note that `overflow at end of batch` has exactly one overflowing row, so a single `Cell` store cannot account for 3.7%. The cost looks like passing `&Cell<bool>` into the closure making the write an observable side effect, which inhibits optimization of the whole kernel loop. That would also explain why the no-overflow win is 10% rather than the roughly 14% the skipped allocation alone should buy.

I think the trade is clearly the right one. Overflow-bearing decimal batches are the rare case, and the common path gets 10% to 29% faster. I would just rather the description say that plainly than assert the overflow path is unchanged. Does that match what you see locally?

For the record, I also went through the equivalence argument for the new `decimal_rescale_check.rs` guard and it holds. Both overflow branches in `rescale_and_check` return `Err` before `overflowed.set(true)`, so dropping the `!fail_on_error` condition is safe, and `try_unary` only invokes the closure on valid indices so nulls cannot set the flag.

CI is green.

native/spark-expr/src/math_funcs/internal/decimal_rescale_check.rs
```diff
@@ -194,16 +199,17 @@ impl PhysicalExpr for DecimalRescaleCheckOverflow {
...
+                        )
                     })?;
 
-                let result = if !fail_on_error && result.values().contains(&i128::MAX) {
-                    // The rescale pass writes i128::MAX as an overflow sentinel for values that
-                    // do not fit the output precision. Only when a sentinel is present do we need
-                    // the extra null-masking pass (which allocates a new array); `contains`
-                    // short-circuits at the first sentinel, so the common no-overflow case skips
-                    // that allocation entirely. ANSI mode raises on overflow and never produces a
-                    // sentinel, so it also skips this pass.
```
peterxcli, line 206: https://github.com/apache/datafusion-comet/pull/5044#discussion_r3728643261

> This dropped the only comment in the codebase explaining why skipping the masking pass is safe. The six lines removed from `decimal_rescale_check.rs` are gone, and neither guard has a replacement, so both call sites are now a bare `if overflowed.get()`.
> The invariant is not obvious from reading the guard. It rests on the fact that every non-sentinel value the closure returns is already clamped inside `±(10^p_out - 1)`, which is exactly the set `null_if_overflow_precision` leaves untouched, so the pass can only ever null sentinels. It is also worth saying that ANSI mode returns `Err` before setting the flag, so ANSI still skips the pass for the same reason it did before.
> Could we add that back at both sites, here and at `decimal_rescale_check.rs:212`? We worked the argument out in this thread, and it would be good for it to live in the code rather than only in the PR conversation.

I guess this comment should be here.

native/spark-expr/src/math_funcs/wide_decimal_binary_expr.rs
peterxcli: filed https://github.com/apache/datafusion-comet/issues/5309. also update PR description to include it

peterxcli (commented):
@andygrove thanks for another round of review, refine the comments you pointed out and file #5309 as followup per your inline comment.

peterxcli:
@mbutrovich mind taking another look when you have a moment? Thanks!

andygrove:

The idea is right: the expression already knows whether anything overflowed, so scanning the output afterwards is redundant work. Recording it in a `Cell` during evaluation and skipping `null_if_overflow_precision` entirely on clean batches is a clean way to get there, and I convinced myself it is safe because `check_overflow_and_convert` bounds by exactly the same `max_for_precision(p_out)` that the masking pass uses, so an unmasked value can never need masking. The new test with a null in the middle of an overflowing batch is a good one to have.

Three things.

**The description does not match the diff**

"Apply the same approach to `DecimalRescaleCheckOverflow` unary evaluation" is listed under what changes are included, but `native/spark-expr/src/math_funcs/internal/decimal_rescale_check.rs` is not in the PR. The only two files changed are the benchmark and `wide_decimal_binary_expr.rs`. Did that part get lost in a rebase, or was it intentionally dropped? Either is fine, but the description should say what actually landed, and if `DecimalRescaleCheckOverflow` still rescans the output buffer then the same win is still on the table.

**The overflow-path regression**

Sparse overflow is 1.6% slower and dense overflow 5.0% slower. That is a fine trade against 10 to 13% on the clean path, and I would take it. But it is worth saying explicitly in the description that the regression is accepted and why, rather than leaving the table to speak for itself. Is #5309, writing the null bit directly, expected to recover both, or just the masking pass?

**Where the extra cost comes from**

The `overflowed.set(true)` store happens on every overflowing row, not just the first. For the dense case that is 4096 redundant stores. I would not add a `if !overflowed.get()` guard, since the branch is probably worse than the store. But is the 5% actually coming from the stores, or from the `Cell` inhibiting some optimization inside the `try_binary` closure? If you happen to know, a one-line comment would save the next person from guessing. If #5309 removes the sentinel entirely then it does not matter much.

**One benchmark question**

`make_overflow_batch` builds `right` as all zeros, so `Add` reduces to passing `left` through and the overflow is determined entirely by whether `left` exceeds the output precision. That is fine for measuring the masking pass, but it means the arithmetic itself is the cheapest possible case. Would a benchmark with two non-trivial operands, where the addition itself is doing real i256 work, change the ratio? If the arithmetic dominates, the percentages above would all shrink, which would be worth knowing.

peterxcli:
@andygrove Thanks for the pass — responses point by point.

**Description vs. diff:** you're right, and the history is: the `DecimalRescaleCheckOverflow` half of this PR was superseded on main. #5440 (closing #5094) replaced that expression's custom kernel with a plain Arrow `cast_to`, so the sentinel, the scan, and the masking pass no longer exist there — the win isn't still on the table, it was collected a different way. When I merged upstream/main, this PR's rescale changes correctly dropped out of the diff. I've updated the description to reflect what actually remains and moved the old rescale benchmark table under a note explaining it measured code that #5440 has since removed.

**Regression wording:** the description already states the 3–5% overflow-path regression is an accepted trade-off (added after your Aug 6 review). On your question: #5309 is expected to recover both costs — writing validity bits directly during evaluation removes the masking pass *and* the `Cell`, so no observable side effect remains in the closure.

**Cost attribution:** your Aug 6 measurement already answered this — `overflow at end of batch` has exactly one overflowing row yet regresses 3.7%, so it can't be the stores; it's the `Cell` write being an observable side effect that inhibits optimization of the kernel loop. I've added a comment saying exactly that at the `Cell` creation site, with a pointer to #5309.

**Zero right operand:** the arithmetic isn't actually cheaper for zeros — `i256::wrapping_mul`/`wrapping_add` do identical work regardless of operand values, and the scale factors are runtime values the compiler can't fold, so every row pays the full add-shape cost. What is true is that Add-with-equal-scales is the lightest op (no rescale division), so these shapes deliberately maximize the guard's relative visibility — the reported 3–5% is an upper bound, and heavier arithmetic (multiply + scale-down) would only shrink it. Happy to add a multiply-based overflow shape if you'd like that quantified, though #5309 should make the question moot.

I also restored the safety-invariant comment from your Aug 6 review at the surviving guard site — it had only been added to `decimal_rescale_check.rs`, whose version was deleted by the #5440 merge.

PR #5046, opened 2026-07-27
perf: improve ArrowWriter performance for fixed-length vectors

### LINKED ISSUE #888: Improve ArrowWriter performance for fixed-length vectors

Writing to Arrow vectors through Comet's `ArrowWriter` class only uses `setSafe` on the underlying vectors. This incurs a capacity and resize check for every write. For `CometSparkToColumnarExec` we know the sizes ahead of time, so this is unnecessary overhead when performing a lot of writes.

For `CometSparkToColumnarExec` we should:
1. set the capacities for any column vectors. Note that `setInitialCapacity` on the underlying `BaseFixedWidthVector` doesn't actually allocate anything, so we may need to call `allocateNew` as well.
2. Extend `ArrowWriter` to have unsafe methods that call `set` on the underlying vectors, probably with assertions that we can elide in release builds.

### Additional context

_No response_

peterxcli:

Closes #888.

## Rationale for this change

`ArrowWriter` currently uses Arrow's resizing-safe setters for every value. This performs a capacity check on every write and can reallocate fixed-width vectors while converting Spark columnar batches.

`CometSparkToColumnarExec` knows the output batch size before conversion, so fixed-width vectors can be allocated to the required capacity upfront and written without repeated resize checks.

**Note**: `setInitialCapacity` on the underlying `BaseFixedWidthVector` doesn't actually allocate anything, so we may need to call `allocateNew` as well.

- Add an `ArrowWriter.create` overload that preallocates fixed-width vectors using the known batch row count.
- Add capacity-asserted column write paths that use non-resizing setters for fixed-width values.
- Clear fixed-width validity bits directly after checking capacity.
- Keep resizing-safe setters for variable-width and nested vectors.
- Update `SparkColumnarArrowReader` to allocate for `rowsToProduce` and use the new write paths.
- Add a regression test covering nullable and non-null fixed-width columns beyond Arrow's default initial capacity and verifying that no reallocation occurs.

- Built the native core with `make core`.
- Ran `CometArrowStreamSuite` with Spark 4.1 / Scala 2.13: 6/6 tests passed.
- Ran `CometArrowStreamSuite` with Spark 3.4 / Scala 2.12: 6/6 tests passed.
- Spotless checks passed.

A targeted paired A/B benchmark compared the previous `allocateNew` plus resizing-safe write path with the new exact-capacity, non-resizing path.

Benchmarked current code against fresh upstream/main.
Environment: Apple M4, JDK 17, Spark 4.1, 4,096 rows, two Long columns, identical benchmark source, five alternating JVM forks.

Path | Main median | Current median | Result
-- | -- | -- | --
Columnar, no nulls | 8.9 ns/row | 6.5 ns/row | 27.0% faster, 1.37×
Columnar, 50% nulls | 7.0 ns/row | 6.9 ns/row | No stable change
Rows, no nulls | 10.2 ns/row | 10.2 ns/row | No stable change
Rows, 50% nulls | 14.9 ns/row | 10.9 ns/row | 26.8% faster, 1.37×

</body></html>
All three JVM forks showed an improvement.


andygrove (commented):
Thanks for this. The core idea is right, and the measurement is convincing. I checked all 14 `setValue` / `setValueUnsafe` pairs and every one is a faithful `setSafe` to `set` swap with the same getter. `DecimalWriter` correctly routes its `changePrecision` failure branch to `setNullUnsafe`, and `BitVector` is handled properly by Arrow's `allocateNew(int)` thanks to the `typeWidth == 0` special case in `computeCombinedBufferSize`, so the boolean writer is not a landmine. Nested writers and `NullWriter` correctly stay on the safe path.

I have a few things I would like to work through before merge.

## The unsafe API is decoupled from the thing that makes it safe

`writeColUnsafe` is only in-bounds when the writer came from `create(root, capacity)`, and nothing connects those two. `RowArrowReader`, `CometArrowConverters`, and `ConstantColumnVectors.materialize` all build writers with the plain `create(root)` and its 3970 default, so pointing any of them at the new methods later would compile fine and write out of bounds.

Have you considered folding the capacity guarantee into `writeCol` instead? The base `writeCol` already calls `valueVector.setInitialCapacity(inputNumElements)`, and as far as I can tell that call does nothing at that point, since `setInitialCapacity` only affects the next `allocateNew()` and we are already past it. If a `FixedWidthArrowFieldWriter` override replaced it with something like:

```scala
while (valueVector.getValueCapacity < inputNumElements) valueVector.reAlloc()
```

then the non-resizing setters would always be in-bounds by construction. That would let `writeColUnsafe` and `writeColNoNullUnsafe` go away entirely, and `ConstantColumnVectors.materialize` would pick up the same win, which matters because it runs per batch for partition and `_metadata` constants.

## `assert` is the wrong guard given that we disable Arrow bounds checking

Could the capacity check in `writeColUnsafe` be `require` rather than `assert`? `NativeBase.setArrowProperties()` sets `arrow.enable_unsafe_memory_access=true` whenever debug mode is off, which is the default, and that disables Arrow's own `ArrowBuf` bounds checks. So that line is the only guard against an out-of-bounds native write if a writer ever reaches there under-allocated. It runs once per column per batch and `getValueCapacity` is just a cached field read, so making it unconditional should not show up in the numbers.

## The row-based path was left out

Since `create(root, capacity)` exists now, is there a reason to leave `RowArrowReader` on the plain `create(root)`? It knows `maxRecordsPerBatch`, so today it allocates 3970 and reallocs to 8192 on every batch. That path is the fallback for any child that does not support columnar output, so it may well be hit more often than the columnar one.

Also, the class doc on `SparkColumnarArrowReader` still says the rows go in "via `ArrowWriter.writeCol`". Worth updating to match.

## Benchmark

The A/B numbers are exactly what I wanted to see, and thanks for running three forks. Could the harness be committed? There are around 40 benchmarks in `spark/src/test/scala/org/apache/spark/sql/benchmark/`, and `CometC2RIsolatedBench` covers the reverse direction, but there is nothing for Spark to Arrow conversion. Having it in tree would make the 23% reproducible and give us something to catch a future regression in this path.

## Tests

A couple of thoughts on the new test in `CometArrowStreamSuite`.

`nullableArrow.getValueCapacity should be >= numRows` would pass on the old resizing path too, since `reAlloc` grows the vector past `numRows` either way. The assertion that actually pins the new behavior is `allocatedBytes shouldBe allocator.getAllocatedMemory`, and its intent is not obvious from reading it. A short comment saying that a realloc would make cumulative allocations exceed live allocated memory would help whoever touches this next.

It might also be worth extending this beyond `IntVector`. `BooleanWriter` and `DecimalWriter` have the only non-trivial `setValueUnsafe` bodies, `BitVector` writes bits into the value buffer rather than fixed-width slots, and `DecimalWriter`'s `changePrecision`-returns-false branch calling `setNullUnsafe` is not exercised anywhere I can find. Comparing all `numRows` values against the input instead of three spot checks would also catch an off-by-one in the new loops.

Lastly, given the point about `arrow.enable_unsafe_memory_access` above, a test that a mis-sized writer fails loudly would be worth having. Build one with `create(root)`, then write more than 3970 rows.

peterxcli:
@andygrove pretty thanks for another round of review!

> `writeColUnsafe` is only in-bounds when the writer came from `create(root, capacity)`, and nothing connects those two. `RowArrowReader`, `CometArrowConverters`, and `ConstantColumnVectors.materialize` all build writers with the plain `create(root)` and its 3970 default, so pointing any of them at the new methods later would compile fine and write out of bounds.
> Have you considered folding the capacity guarantee into `writeCol` instead? The base `writeCol` already calls `valueVector.setInitialCapacity(inputNumElements)`, and as far as I can tell that call does nothing at that point, since `setInitialCapacity` only affects the next `allocateNew()` and we are already past it. If a `FixedWidthArrowFieldWriter` override replaced it with something like:
> ```scala
> while (valueVector.getValueCapacity < inputNumElements) valueVector.reAlloc()
> ```
> then the non-resizing setters would always be in-bounds by construction. That would let `writeColUnsafe` and `writeColNoNullUnsafe` go away entirely, and `ConstantColumnVectors.materialize` would pick up the same win, which matters because it runs per batch for partition and `_metadata` constants.

Addressed, this is brilliant. great cleanup.

> Could the capacity check in `writeColUnsafe` be `require` rather than `assert`? `NativeBase.setArrowProperties()` sets `arrow.enable_unsafe_memory_access=true` whenever debug mode is off, which is the default, and that disables Arrow's own `ArrowBuf` bounds checks. So that line is the only guard against an out-of-bounds native write if a writer ever reaches there under-allocated. It runs once per column per batch and `getValueCapacity` is just a cached field read, so making it unconditional should not show up in the numbers.

make sense, changed assert to `while (valueVector.getValueCapacity < inputNumElements) { valueVector.reAlloc() }`

> Since `create(root, capacity)` exists now, is there a reason to leave `RowArrowReader` on the plain `create(root)`? It knows `maxRecordsPerBatch`, so today it allocates 3970 and reallocs to 8192 on every batch. That path is the fallback for any child that does not support columnar output, so it may well be hit more often than the columnar one.

you're right, remaining usage are only existed in test code. I've completely removed the plain `create(root)` and change its caller code.

> Also, the class doc on `SparkColumnarArrowReader` still says the rows go in "via `ArrowWriter.writeCol`". Worth updating to match.

class doc on `SparkColumnarArrowReader` is updated.

> benchmark: the A/B numbers are exactly what I wanted ... . ... there is nothing for Spark to Arrow conversion.

Added as `CometArrowWriterBenchmark.scala`.

> `nullableArrow.getValueCapacity should be >= numRows` would pass on the old resizing path too, since `reAlloc` grows the vector past `numRows` either way. The assertion that actually pins the new behavior is `allocatedBytes shouldBe allocator.getAllocatedMemory`, and its intent is not obvious from reading it. A short comment saying that a realloc would make cumulative allocations exceed live allocated memory would help whoever touches this next.

Added into existing "pre-sized Arrow writer avoids fixed-width reallocations" test case in `CometArrowStreamSuite.scala`.

> It might also be worth extending this beyond `IntVector`. `BooleanWriter` and `DecimalWriter` have the only non-trivial `setValueUnsafe` bodies, `BitVector` writes bits into the value buffer rather than fixed-width slots, and 

Added into existing "pre-sized Arrow writer avoids fixed-width reallocations" test case in `CometArrowStreamSuite.scala`.

> `DecimalWriter`'s `changePrecision`-returns-false branch calling `setNullUnsafe` is not exercised anywhere I can find. Comparing all `numRows` values against the input instead of three spot checks would also catch an off-by-one in the new loops.

Added in `CometArrowStreamSuite.scala` as a new "fixed-width decimal write clears validity on precision overflow" test case.

> Lastly, given the point about `arrow.enable_unsafe_memory_access` above, a test that a mis-sized writer fails loudly would be worth having. Build one with `create(root)`, then write more than 3970 rows.

Added in `CometArrowStreamSuite.scala` as a new "fixed-width column write grows an undersized vector" test case.

peterxcli:
@mbutrovich you might also want to take a look at this as you created #888. TIA!

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowWriters.scala
mbutrovich, line 190: `RowArrowReader` (RowArrowReader.scala:62) drives this method through `writer.write(row)` for every row of the fallback, non-columnar path. For a `FixedWidthArrowFieldWriter`, `setValue` here is still the original `setSafe` call, since `setValueUnsafe` is only reached from `writeCol`/`writeColNoNull`. So every field of every row on this path still runs Arrow's `handleSafe` capacity check, the exact per-write cost the PR description says this change removes.

`RowArrowReader` now builds its writer with `create(root, maxRecordsPerBatch.toInt)`, so `count < capacity` holds for the whole batch by construction, the same invariant `writeCol`/`writeColNoNull` rely on to call `setValueUnsafe` directly. Routing `write` through the unsafe setter for fixed-width fields looks safe under that same invariant.

This path backs `CometSparkRowToColumnar`, used whenever the child doesn't support columnar output. That's likely to be a common case, and it's worth having a benchmark number for it the same way `CometArrowWriterBenchmark` covers `SparkColumnarArrowReader`, since right now there is no measurement for how much of the intended win this path is actually getting.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowWriters.scala
```diff
@@ -224,89 +228,133 @@ private[arrow] abstract class ArrowFieldWriter {
...
+      } else {
+        setValueUnsafe(input, count)
+      }
+      count += 1
+    }
+  }
 
-  override def setNull(): Unit = {
-    valueVector.setNull(count)
+  override def writeColNoNull(input: ColumnarArray): Unit = {
```
mbutrovich, line 273: For the no-null case, is a bulk copy of the source column into the Arrow data buffer worth exploring instead of one `setValueUnsafe` call per row? `BaseFixedWidthVector.getDataBuffer()` / `getDataBufferAddress()` plus a raw memory copy is already the pattern `copyFrom` uses internally in arrow-java when the source is another Arrow vector. Spark's on-heap and off-heap `ColumnVector` implementations back fixed-width columns with contiguous primitive arrays, so the same kind of bulk transfer should be reachable here for the no-null case, cutting the per-row virtual dispatch entirely rather than just skipping the capacity check on it.

This is a bigger change than this PR's scope, so if it's not something you want to take on here, could you file a follow-up issue for it? I'd rather have that tracked than have it drop once this merges.

mbutrovich (commented):
Nice work addressing the previous round, the capacity-guarantee-in-writeCol change and the new tests cover what I'd have asked for on the columnar path. The two things above are about the row-based fallback path (`RowArrowReader`) not picking up the same optimization, and a possible further win on the no-null bulk case using arrow-java's existing buffer-level copy APIs.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowWriters.scala
peterxcli: filed. https://github.com/apache/datafusion-comet/issues/5299

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowWriters.scala
peterxcli: 1. added `writeUnsafe` for `writer.write(row)` calls. (this shows a little improvement, about 0.1 ns/row)
2. added spark row to arrow benchmark. also updated the benchmark result in pr description.

peterxcli:
@mbutrovich Thanks for the review, filed a followup for exploring "no-null bulk case using arrow-java's existing buffer-level copy APIs", added unsafe write function for `writer.write(row)` to bypass the capacity check in the original safe write api and benchmark that clone the behaviour from `SparkColumnarArrowReader`, shows about `1.37×` faster improvement(updated in PR description).

peterxcli:
@mbutrovich I've addressed your review, would appreciate it if you could take a look at the update and see if we can get this merged.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/RowArrowReader.scala
```diff
@@ -55,7 +55,7 @@ private[comet] class RowArrowReader(
     }
 
     val startNs = System.nanoTime()
-    val writer = ArrowWriter.create(getVectorSchemaRoot)
+    val writer = ArrowWriter.create(getVectorSchemaRoot, maxRecordsPerBatch.toInt)
     var rowCount = 0L
     while (rowIter.hasNext &&
       (maxRecordsPerBatch <= 0 || rowCount < maxRecordsPerBatch)) {
```
mbutrovich: The `maxRecordsPerBatch <= 0` branch here looks like it's no longer safe now that `write(row)` routes fixed-width fields through `writeUnsafe`.

`create(root, maxRecordsPerBatch.toInt)` on line 58 pre-allocates each fixed-width vector to exactly `maxRecordsPerBatch` capacity, and `FixedWidthArrowFieldWriter.writeUnsafe` calls Arrow's non-resizing `set()` with no bounds check. If `maxRecordsPerBatch` is ever `0`, `ArrowWriter.create` allocates fixed-width vectors at capacity 0 (its `require(fixedWidthCapacity >= 0, ...)` on `ArrowWriters.scala:43` permits 0), but this loop treats `0` as "unlimited" and keeps writing through the unsafe setter anyway. Under `arrow.enable_unsafe_memory_access=true` (the default in non-debug builds), that's an out-of-bounds native write, not an exception.

This isn't reachable today because `CometConf.COMET_BATCH_SIZE` has `checkValue(v => v > 0, ...)`, so both production callers (`CometSparkToColumnarExec`, `CometLocalTableScanExec`) always pass a positive value. But `CometArrowConverters.rowToArrowBatchIter` hit this same class of problem and picked up `require(maxRecordsPerBatch > 0, ...)` at `CometArrowConverters.scala:58`. `RowArrowReader` shares the same invariant (a writer built once per batch, sized for the whole batch, then written unsafely) and doesn't have the matching guard. `SparkColumnarArrowReader` is fine here since `writeCol`/`writeColNoNull` self-heal capacity via their own `reAlloc` loop regardless of what `create` allocated, but `RowArrowReader.write(row)` has no such self-healing.

Could this get the same `require` (or have the dead `<= 0` branch removed, since `maxRecordsPerBatch` is always positive in practice)? Right now it's unreachable only because of a config check upstream in `CometConf`, and there's no compiler-enforced link between the two.

peterxcli:
@mbutrovich thanks for the review, added `require` to assert the `maxRecordsPerBatch` for `RowArrowReader` is always greater than zero. and because when it calls `ArrowWriter.create(root, capacity = maxRecordsPerBatch)` which require `capacity` to be `Int`, so I change `maxRecordsPerBatch` to `Int`, too.

mbutrovich (approved):
Approved pending CI, thanks @peterxcli!

peterxcli:
Thanks!

PR #4852, opened 2026-07-07
feat: add comet_version() SQL function

andygrove:
## Rationale for this change

There is currently no built-in way for a Spark user to programmatically query which Comet version is loaded. The JVM `COMET_VERSION` constant exists in `org.apache.comet.package` but is only accessible via Scala imports; the native library also logs the version on init, but that requires driver log access. This PR exposes the version as a plain SQL function so it is reachable from PySpark, spark-sql, notebooks, and any other client that can issue SQL — no JVM/Scala imports and no log spelunking required.

## What changes are included in this PR?

- Register a nullary SQL function `comet_version()` via `SparkSessionExtensions.injectFunction` in `CometSparkSessionExtensions`. The function returns the `COMET_VERSION` string (already populated from `comet-git-info.properties` at build time) as a Spark `StringType` literal.
- Reject any arguments with an `IllegalArgumentException` mentioning the function name.
- Add two unit tests in `CometSparkSessionExtensionsSuite`: one asserting `SELECT comet_version()` returns the expected build version, and one asserting `comet_version(1)` fails.

Compiles cleanly against `spark-3.4`, `spark-3.5`, `spark-4.0`, and `spark-4.1` profiles.

Example usage:

```
scala> spark.sql("SELECT comet_version()").show(false)
+----------------+
|comet_version() |
+----------------+
|1.0.0-SNAPSHOT  |
+----------------+
```

## How are these changes tested?

- New unit tests in `CometSparkSessionExtensionsSuite` (`comet_version() returns the Comet build version` and `comet_version() rejects arguments`).
- Full suite runs green locally with the `spark-3.5` profile.
- Verified compilation with `-Pspark-3.4`, `-Pspark-3.5`, and `-Pspark-4.1 -Pscala-2.13`.

PR #5049, opened 2026-07-27
feat: expose Comet version as spark.comet.version runtime config

andygrove:

Supersedes #4852.

## Rationale for this change

There is currently no convenient way for a Spark user to query which Comet version is loaded. The JVM `org.apache.comet.COMET_VERSION` constant exists but is only reachable via Scala imports, and the native library logs the version on init but that requires driver log access.

The previous attempt (#4852) exposed the version as a SQL function via `SparkSessionExtensions.injectFunction`. Review feedback on that PR questioned whether a function was the right mechanism, and registering a function into the session registry also broke the upstream Spark `ShowFunctionsSuite` tests, since those suites assert on the exact set of registered functions. Exposing the version as a Spark config avoids polluting the function registry entirely and matches how Spark surfaces similar build metadata.

- The Comet driver plugin sets `spark.comet.version` on the SparkConf during `init`, so it is queryable at runtime via `spark.conf.get("spark.comet.version")` or `SET spark.comet.version` in SQL. It is set before the off-heap check so the version is reported even when Comet is otherwise disabled.
- Documents both the new `spark.comet.version` runtime config and the existing `org.apache.comet.COMET_VERSION` (plus `COMET_BRANCH` / `COMET_REVISION`) programmatic accessors in the installation guide.

- New unit test in `CometPluginsSuite` (`Comet version is exposed as a Spark config`) asserting `spark.comet.version` is visible on both the SparkContext conf and the session runtime config, using the same plugin-loading harness that covers the existing memory-overhead behavior.

### DIFF
--- a/docs/source/user-guide/latest/installation.md
+++ b/docs/source/user-guide/latest/installation.md
@@ -166,6 +166,33 @@ CometNativeColumnarToRow
    +- CometNativeScan parquet [a#6] Batched: true, DataFilters: [isnotnull(a#6), (a#6 > 5)], Format: CometParquet, Location: InMemoryFileIndex(1 paths)[file:/tmp/test], PartitionFilters: [], PushedFilters: [IsNotNull(a), GreaterThan(a,5)], ReadSchema: struct<a:int>
 ```
 
+## Checking the Comet Version
+
+When the Comet plugin is loaded, it exposes its build version as the Spark config
+`spark.comet.version`. This can be queried at runtime from any supported language, for example:
+
+```scala
+scala> spark.conf.get("spark.comet.version")
+```
+
+```sql
+SET spark.comet.version;
+```
+
+The same value is available programmatically on the JVM classpath, along with additional build
+metadata that is useful when reporting issues:
+
+```scala
+scala> import org.apache.comet.{COMET_VERSION, COMET_BRANCH, COMET_REVISION}
+scala> println(COMET_VERSION)
+```
+
+Comet also logs its version when the native library is initialized:
+
+```shell
+INFO core/src/lib.rs: Comet native library version <version> initialized
+```
+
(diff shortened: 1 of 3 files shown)


docs/source/user-guide/latest/installation.md
```diff
@@ -166,6 +166,33 @@ CometNativeColumnarToRow
    +- CometNativeScan parquet [a#6] Batched: true, DataFilters: [isnotnull(a#6), (a#6 > 5)], Format: CometParquet, Location: InMemoryFileIndex(1 paths)[file:/tmp/test], PartitionFilters: [], PushedFilters: [IsNotNull(a), GreaterThan(a,5)], ReadSchema: struct<a:int>
 ```
 
+## Checking the Comet Version
+
+When the Comet plugin is loaded, it exposes its build version as the Spark config
+`spark.comet.version`. This can be queried at runtime from any supported language, for example:
+
+```scala
+scala> spark.conf.get("spark.comet.version")
```
comphead, line 175: would be nice to get an example output?

docs/source/user-guide/latest/installation.md
```diff
@@ -166,6 +166,33 @@ CometNativeColumnarToRow
...
+```
+
+```sql
+SET spark.comet.version;
```
comphead, line 179: I hardly can imagine scenarios for user to change comet version

andygrove: `SET` shows the current value.

```
spark-sql (default)> SET spark.sql.adaptive.enabled;
spark.sql.adaptive.enabled	true
Time taken: 0.276 seconds, Fetched 1 row(s)
```

mbutrovich (approved):
Thanks @andygrove!

PR #5053, opened 2026-07-27
fix: include ABFS container in object store cache key

### LINKED ISSUE #4993: Object store cache key drops the ABFS container, so two containers in one storage account share a store instance

peterxcli:

Closes #4993.

## Rationale for this change

ABFS URLs encode the container in URL userinfo, for example
`abfss://container@account.dfs.core.windows.net/path`. The process-wide object store cache
previously keyed stores by scheme, host, and port only, so different containers in the same
storage account shared one cache entry. Since each Azure object store is bound to the container
from its URL, a later read could reuse the first container's store and return data from the wrong
container.

This also prevents containers sharing one Hadoop configuration map from resolving different `fs.azure.sas.<container>.<account>` tokens into the same cached store.

- Preserve URL userinfo when building object store keys for `abfs` and `abfss`, while retaining
  the existing host-only behavior for other schemes.
- Update the cache documentation to describe container-scoped keys.
- Add an offline regression test that prepares two containers in the same account with identical
  configuration and verifies that they resolve to distinct object store instances.

- `cargo test --manifest-path native/Cargo.toml -p datafusion-comet --lib`
  (`129 passed; 4 ignored`)
- `cargo clippy --manifest-path native/Cargo.toml -p datafusion-comet --lib --tests -- -D warnings`

The regression constructs the Azure stores locally and does not issue network requests.


andygrove:
Nice targeted fix. I confirmed the analysis by reading `native/core/src/parquet/objectstore/azure.rs` — `MicrosoftAzureBuilder::parse_url` binds the container into the store instance and the resource `Path` is container-relative, so a store genuinely cannot be reused across containers. `Position::BeforeUsername..AfterPort` correctly includes the userinfo when present and collapses to `BeforeHost` when it isn't, so non-Azure schemes are untouched.

A few small things worth considering:

1. **DataFusion's own registry has the same blind spot.** `RuntimeEnv::register_object_store` uses `get_url_key`, which also strips userinfo (`datafusion-execution/src/object_store.rs`). So within a single `RuntimeEnv`, only one abfss store per account can be registered — a second registration would overwrite the first at the DF layer. Today this is fine because Comet constructs a fresh `RuntimeEnv` per Parquet file read, as the module doc calls out. But that invariant is now load-bearing for correctness on Azure. A sentence tying the two together in the module doc block would help a future reader who tries to share a `RuntimeEnv` across container reads.

2. **A one-line comment on the `is_azure_scheme(scheme)` branch** explaining that ABFS URLs encode the container in the userinfo would save future readers a trip to #4993.

3. **Two small extensions to the new test**, if you want them:
   - Assert that a second call with the same container returns the same `Arc`, so a future change that accidentally invalidates the same-URL cache path is also caught.
   - Add an `s3://bucket@…`-style case to pin down that the non-Azure branch still keys by host only.

None of these are blockers.

peterxcli:
@andygrove thanks for the review, added comment on the abfs branch and new test coverage per your review comment

peterxcli:
df PR: https://github.com/apache/datafusion/pull/23935

andygrove:
Fix looks good and I confirmed it works. Two containers in the same account now resolve to their own stores across tasks, where on `main` the second one gets the first's data.

One question before I approve. Within a single plan, `prepare_object_store_with_configs` registers against the shared `session_ctx.runtime_env()` (`planner.rs:1528`), and DataFusion's `get_url_key` strips userinfo, so a second container in the same account would overwrite the first in the registry and both scans would resolve to the same store. Your apache/datafusion#23935 fixes that upstream. Do you know whether a single plan can actually reference two containers in one account today? If it can, could we get a follow-up issue tracking the DF bump?

peterxcli:
> `prepare_object_store_with_configs` registers against the shared `session_ctx.runtime_env()` (`planner.rs:1528`), and DataFusion's `get_url_key` strips userinfo, so a second container in the same account would overwrite the first in the registry and both scans would resolve to the same store. Your [apache/datafusion#23935](https://github.com/apache/datafusion/pull/23935) fixes that upstream. Do you know whether a single plan can actually reference two containers in one account today? If it can, could we get a follow-up issue tracking the DF bump?

We didn't use the datafusion's `get_url_key`.

andygrove:
> We didn't use the datafusion's `get_url_key`.

I think we may be talking past each other here. Comet doesn't call `get_url_key` directly, but `runtime_env.register_object_store()` and `runtime_env.object_store()` both go through `DefaultObjectStoreRegistry`, which does call it. In `datafusion-execution-54.1.0/src/object_store.rs:268` it slices `Position::BeforeHost..Position::AfterPort`, and the doc comment on it reads "The credential info will be removed."

I put together a quick probe on your branch that registers two containers against one shared `RuntimeEnv`, the way `planner.rs:1587` does:

```rust
#[test]
fn test_shared_runtime_env_collides_across_containers() {
    let configs = HashMap::from([("fs.azure.account.key".into(), "c2VjcmV0".into())]);
    // One plan => one shared RuntimeEnv, as in planner.rs
    let runtime_env = Arc::new(RuntimeEnv::default());
    let register = |container: &str| {
        super::prepare_object_store_with_configs(
            Arc::clone(&runtime_env),
            format!("abfss://{container}@shared-acct.dfs.core.windows.net/path/file.parquet"),
            &configs,
        )
        .unwrap()
        .0
    };
    let url_a = register("container-a");
    let url_b = register("container-b");
    let store_a = runtime_env.object_store(&url_a).unwrap();
    let store_b = runtime_env.object_store(&url_b).unwrap();
    println!("url_a={url_a} url_b={url_b}");
    println!("same store: {}", Arc::ptr_eq(&store_a, &store_b));
    println!("store_a={store_a:?}");
}
```

Output:

```
url_a=abfss://container-a@shared-acct.dfs.core.windows.net/
url_b=abfss://container-b@shared-acct.dfs.core.windows.net/
same store: true
store_a=MicrosoftAzure { ... container: "container-b" ... }
```

Both URLs resolve to the same store, and that store is bound to `container-b`. So a scan of `container-a` would silently read `container-b`'s data. `FileScanConfig` resolves the store at execute time (`datafusion-datasource/src/file_scan_config/mod.rs:640`), after every scan in the plan has already registered, so the last registration wins. A join or union across two containers in one storage account would hit this. The `native_iceberg_compat` path in `parquet/mod.rs:163` is safe because it builds a fresh `SessionContext` per file, but `native_datafusion` shares the task-scoped one.

Would you be up for closing this in the same PR? Comet already builds its `RuntimeEnv` through `RuntimeEnvBuilder` at `jni_api.rs:559`, and DF 54.1 exposes `RuntimeEnvBuilder::with_object_store_registry`, so a small registry impl that keys on the full authority would fix it now without waiting on apache/datafusion#23935 and the version bump. If you'd rather keep this PR scoped to the cache key, that's reasonable, but could we get an issue filed for the registry side so it doesn't get lost?

Two smaller things while I'm here.

The two new tests build a fresh `RuntimeEnv` per call, which is what makes the assertions hold. That's the right shape for testing the Comet cache in isolation. It might be worth adding one more that shares a single `RuntimeEnv` across both registrations, since that's what `planner.rs` actually does and it's the case that still resolves to one store today.

On the module doc at `parquet_support.rs:501`, since the container is now part of the Comet cache key but not part of the DataFusion registry key, could we add a sentence noting that per-container isolation depends on the registry side too? As written a reader could come away thinking the key change alone is sufficient.

One thing I noticed that works in your favour and isn't mentioned in the description: `translate_hadoop_configs` resolves `fs.azure.sas.<container>.<account>`, so before this change two containers sharing one config map would resolve different SAS tokens into the same cached store. This fixes that as well.

peterxcli:
@andygrove thanks for further explanation! you're really right on this. I've addressed all of your review/suggestion.

> `runtime_env.register_object_store()` and `runtime_env.object_store()` both go through `DefaultObjectStoreRegistry`, which does call it. In `datafusion-execution-54.1.0/src/object_store.rs:268` it slices `Position::BeforeHost..Position::AfterPort`, and the doc comment on it reads "The credential info will be removed."
> 
> I put together a quick probe on your branch that registers two containers against one shared `RuntimeEnv`, the way `planner.rs:1587` does:
> ....
> Both URLs resolve to the same store, and that store is bound to `container-b`. So a scan of `container-a` would silently read `container-b`'s data. `FileScanConfig` resolves the store at execute time (`datafusion-datasource/src/file_scan_config/mod.rs:640`), after every scan in the plan has already registered, so the last registration wins. A join or union across two containers in one storage account would hit this. The `native_iceberg_compat` path in `parquet/mod.rs:163` is safe because it builds a fresh `SessionContext` per file, but `native_datafusion` shares the task-scoped one.

Thanks for catching this! sorry my previous response just refuse to address this because that's my blindspot.

> The two new tests build a fresh `RuntimeEnv` per call, which is what makes the assertions hold. That's the right shape for testing the Comet cache in isolation. It might be worth adding one more that shares a single `RuntimeEnv` across both registrations, since that's what `planner.rs` actually does and it's the case that still resolves to one store today.

added as `test_shared_runtime_env_uses_distinct_azure_container_stores` in `native/core/src/parquet/parquet_support.rs`

> On the module doc at `parquet_support.rs:501`, since the container is now part of the Comet cache key but not part of the DataFusion registry key, could we add a sentence noting that per-container isolation depends on the registry side too? As written a reader could come away thinking the key change alone is sufficient.

added.

> One thing I noticed that works in your favour and isn't mentioned in the description: `translate_hadoop_configs` resolves `fs.azure.sas.<container>.<account>`, so before this change two containers sharing one config map would resolve different SAS tokens into the same cached store. This fixes that as well.

Nice!

andygrove:

This is a real correctness bug and the diagnosis is clear: two containers in the same storage account are two different stores, and keying on `scheme://host:port` collapses them. Returning data from the wrong container is about as bad as it gets, so this is worth fixing.

A few questions before merge.

**Is the custom registry installed everywhere a `RuntimeEnv` is built?**

`CometObjectStoreRegistry` is wired in at `prepare_datafusion_session_context`. Is that the only place Comet constructs a `RuntimeEnv`? The native Parquet reader entry point in `parquet/mod.rs` also does `session_ctx.runtime_env()`, and if any path builds its own runtime with the default registry, ABFS isolation quietly stops working there while the process-wide cache in `prepare_object_store_with_configs` still does the right thing. A grep for `RuntimeEnvBuilder::new` in the description, or an assertion somewhere, would settle it. The two-layer design here means both layers have to agree, and only one of them is hard to forget.

**Should this be fixed upstream instead?**

`DefaultObjectStoreRegistry::get_store` builds its key from `Position::BeforeHost`, dropping userinfo. That is arguably a DataFusion bug rather than something specific to Comet, since any object store that encodes identity in userinfo hits it. Has an issue been filed upstream? If it has, linking it here and marking `CometObjectStoreRegistry` as a temporary workaround would make it easier to remove later. If DataFusion would not accept the change, that is worth writing down too.

**The S3 test asserts a URL shape that does not occur**

`test_s3_store_cache_keys_by_host` constructs `s3://{bucket}@shared-host/path/file.parquet`, putting the bucket in userinfo and a separate host after it. Real S3 URLs put the bucket in the host position. So the test proves that userinfo is ignored for non-Azure schemes, which is what you intended, but the name and the variable called `bucket` suggest it is testing something about buckets that it is not. Could it be renamed to something like `test_non_azure_schemes_ignore_userinfo`, with a comment saying the URL shape is synthetic?

**Two smaller notes**

`azure_stores` uses `parking_lot::RwLock` while the neighbouring `ObjectStoreCache` uses `std::sync::RwLock`. Not a problem, but consistency within one file is worth having unless there is a reason for the difference.

`register_store` for Azure inserts into `azure_stores` and never touches `self.default`. So `DefaultObjectStoreRegistry` never learns about ABFS stores at all. Is there anything in DataFusion that enumerates the registry, or looks up a store through a path other than `get_store`, that would now miss them?

peterxcli:
Thanks for the careful pass — all five points were worth acting on; addressed in ff694ec5a.

**Registry coverage:** Good catch. Auditing all `RuntimeEnv` construction sites turned up one production path without the custom registry: `initRecordBatchReader` (`native_iceberg_compat`) builds a private `SessionContext` per reader. It wasn't observably wrong — a per-reader registry only ever sees one store — but isolation there depended on that invariant staying true, so I've installed `CometObjectStoreRegistry` on that path as well. The only other construction sites are test-only (`PhysicalPlanner::default()` and friends).

**Upstream:** Yes — apache/datafusion#23935 fixes `get_url_key` upstream. I've added a doc comment on `CometObjectStoreRegistry` marking it a temporary workaround with a link, so it can be removed at the version bump.

**S3 test:** Agreed the name oversold it. Renamed to `test_non_azure_schemes_ignore_userinfo` with a comment that the URL shape is synthetic (real S3 URLs carry the bucket in the host position).

**Lock consistency:** Switched to `std::sync::RwLock` to match the rest of the file, recovering from poisoning via `PoisonError::into_inner`.

**Registry bypass:** Nothing in DataFusion 54.1 enumerates the registry, and `RuntimeEnv::object_store` is the only lookup path — but your question surfaced that the trait's provided `deregister_store` defaults to a not-implemented error, and `DefaultObjectStoreRegistry`'s own implementation wouldn't see Azure stores either. Nothing calls it today, but I've implemented it on `CometObjectStoreRegistry` so the registry is complete.

sunchao (commented):
Reviewed `dc4e864de5059e304f214ade2e53bd665d1b816b` across the native cache/registry, Spark/JNI integration, ownership and cleanup, URL/backend compatibility, encryption, and tests. I found one regression-test coverage issue, detailed inline. The current custom registry passed the stronger late-lookup and cross-container read checks; I did not find a production correctness regression.

Validation:
- Local: 25 existing native tests and 3 temporary review probes passed with locked dependencies, offline, using `--no-default-features`. The probes covered ABFS/ABFSS lookup after all registrations, reads from separate in-memory stores in both registration orders, registry replacement/removal, and non-Azure delegation.
- [CI](https://github.com/apache/datafusion-comet/actions/runs/34621244261): 70 successful checks and 9 skipped checks, including successful Required Checks. The native job reports 1,363 passed and 5 skipped. Its tested merge commit `cdd689d0810ba3114092bcd1c28d71c72ee589ed` has the same source tree as the reviewed head.
- Limits: no live Azure requests, libhdfs I/O, or local Spark execution. The Hadoop-routing probe used cached in-memory stores.

native/core/src/parquet/parquet_support.rs
```diff
@@ -1153,6 +1232,34 @@ mod tests {
...
+            (url.clone(), runtime.object_store(&url).unwrap())
+        };
+
+        let (url_a, store_a) = register("container-a");
+        let (same_url_a, same_store_a) = register("container-a");
+        let (url_b, store_b) = register("container-b");
+        assert_eq!(url_a, same_url_a);
+        assert!(Arc::ptr_eq(&store_a, &same_store_a));
+        assert_ne!(url_a, url_b);
+        assert!(!Arc::ptr_eq(&store_a, &store_b));
```
sunchao, line 1260: **[P2] Resolve both URLs after registering both containers**

Could we look up `url_a` and `url_b` again after registering container B and assert that each still resolves to its expected store? `register` captures A's `Arc` before B is inserted, so this pointer comparison only compares handles captured at different times. I ran all four existing assertions against `DefaultObjectStoreRegistry`: they still passed for both `abfs` and `abfss`, while subsequent lookups of A and B both returned B's store. The current `CometObjectStoreRegistry` handles this correctly, but this regression test would not detect replacing it with the default registry. Re-fetching both stores after both registrations would cover the shared-runtime overwrite.

sunchao: Verified fixed in `9a50d58114f646a85dc3b70e54f5580d1409dfac`. The new late lookups preserve both expected stores. I also temporarily substituted `DefaultObjectStoreRegistry`: the test now fails at the new container-A assertion, then passes again after restoring `CometObjectStoreRegistry`. This covers the shared-runtime overwrite raised here.

sunchao (approved):
Re-reviewed `9a50d58114f646a85dc3b70e54f5580d1409dfac`. No remaining findings.

The previous P2 is addressed: the regression test now resolves both URLs after both containers are registered. I verified that the updated test passes, fails at the new container-A assertion when temporarily changed to use `DefaultObjectStoreRegistry`, and passes again after restoring `CometObjectStoreRegistry`. The fresh review of the full change found no new correctness, compatibility, or cleanup issues.

Validation:

- 25 focused native tests passed locally on the reviewed head with locked dependencies, offline, using `--no-default-features`.
- [CI](https://github.com/apache/datafusion-comet/actions/runs/35452669281): 23 successful checks and 14 skipped, with Required Checks passing. The native suite reports 1,578 passed and 5 skipped. CI tested merge `bbbe65066cb87fcb2fa61d47df2ae24a1e59c971`, which includes newer `main`; this is separate from the exact-head local validation.
- No live Azure requests, libhdfs I/O, or local Spark execution. Temporary test changes were removed, and the checkout is clean.

Thanks for strengthening the regression test.

Issue #4724, opened 2026-06-24
Support fully-native multi-stage (distinct-combined) collect_list / collect_set


andygrove:
## Describe the bug

`collect_list` / `collect_set` declare their aggregate buffer as `BinaryType` in Spark (serialized `TypedImperativeAggregate` state) but produce a native `ArrayType` (Arrow `List`) state in Comet. Comet bridges this only for the simple two-stage shape: `CometObjectHashAggregateExec.adjustOutputForNativeState` rewrites the buffer column of a **pure-`Partial`** aggregate to `ArrayType`, so `Partial -> Final` runs natively and correctly.

It does **not** handle multi-stage aggregates that contain a `PartialMerge` stage, which Spark introduces for the distinct-aggregate rewrite, e.g.:

```sql
SELECT x, count(DISTINCT y), collect_list(z) FROM t GROUP BY x
```

This plans as `Partial(x,y) -> PartialMerge(x,y) -> [PartialMerge, Partial](x) -> Final(x)`. The intermediate `PartialMerge` outputs are still declared `BinaryType`, so a fully-native pipeline crashes at runtime:

```
CometNativeException: Cast error: Cannot cast LIST to non-list data type Binary
```

(and a related nullability drift `List(non-null T)` vs `List(nullable T)` for nested element types). This is a facet of the broader Arrow-type-drift tracked in #4515.

This affects both `collect_list` (added in #4720) and the already-shipped `collect_set` — `collect_set` has the same latent crash on this shape.

## Current behavior (workaround in #4720)

To avoid the crash, multi-stage `collect_list`/`collect_set` aggregates now fall back to Spark consistently:
- `CometExecRule.tagUnsafePartialAggregates` tags the feeding pure-`Partial` when a `PartialMerge` stage of a `CollectList`/`CollectSet` aggregate is present.
- `CometBaseAggregate.doConvert` falls back a `PartialMerge`/`Final` stage of these functions when no Comet partial produced the buffer (the cross-engine `LocalTableScan` case).

The simple two-stage `collect_list`/`collect_set` cases continue to run natively.

## Expected behavior

Enable fully-native multi-stage execution by correcting the intermediate buffer schema for `PartialMerge` (and multi-mode) stages of `CollectList`/`CollectSet` (extend `adjustOutputForNativeState` beyond pure-`Partial`, and fix the element-nullability drift), so the `ArrayType` buffer round-trips across all stages. Then remove the fallback guards added in #4720.

## Additional context

[...]

PR #4720, opened 2026-06-24
feat: native collect_list / array_agg aggregate

andygrove:
## Which issue does this PR close?

Closes #2524.

## Rationale for this change

`collect_list` and its alias `array_agg` are common aggregate functions that previously fell back to Spark, breaking native execution for many real workloads (notably plans that group rows into arrays before further processing). Adding native support keeps these queries on the Comet path.

## What changes are included in this PR?

This change was scaffolded with the `implement-comet-expression` skill.

- Add `CollectList` message to `expr.proto` and a new `collectList = 18` arm in the `AggExpr` oneof.
- Add `CometCollectList` serde in `aggregates.scala` and register `classOf[CollectList] -> CometCollectList` in `QueryPlanSerde`.
- Wire the native side to `datafusion_spark::function::aggregate::collect::SparkCollectList`, the upstream Spark-compatible accumulator (Spark 3.4 through 4.1 use `ignore_nulls = true` semantics that match it). No Comet-local Rust function is added.
- Extend the Partial-mode `adjustOutputForNativeState` fix in `operators.scala` to cover `CollectList` (same pattern already used for `CollectSet`: native produces `ArrayType(elementType)` while Spark declares the buffer as `BinaryType`).
- Mark `collect_list` and `array_agg` as supported in the user-guide expression page.
- Add an audit entry under `docs/source/contributor-guide/expression-audits/agg_funcs.md` covering Spark 3.4.3, 3.5.8, 4.0.1, 4.1.1.

## How are these changes tested?

- New SQL fixture `spark/src/test/resources/sql-tests/expressions/aggregate/collect_list.sql` exercises ~30 queries across types (boolean, byte/short/int/bigint, float/double including NaN/Inf/-0, string, binary, decimal up to (38,0), date, timestamp, struct, nested array), GROUP BY, NULLs, all-NULL groups, empty tables, single-row, mixed aggregates, multiple `collect_list` columns, DISTINCT, HAVING, the `array_agg` alias, INT/BIGINT boundary values, and an SPARK-17641 null-filter regression. The fixture runs under `ConfigMatrix: parquet.enable.dictionary=false,true`, so each query executes twice.
- All 21 aggregate `CometSqlFileTestSuite` tests pass (`./mvnw test -Dsuites="org.apache.comet.CometSqlFileTestSuite expressions/aggregate" -Dtest=none -Pspark-3.5`), confirming no regression to `collect_set` or other aggregates.
- `cargo clippy --all-targets --workspace -- -D warnings` is clean.

PR #5055, opened 2026-07-27
test: docs and test-coverage hardening for native collect_list / collect_set

### LINKED ISSUE #5024: Follow-up: docs and test-coverage hardening for native collect_list / array_agg (#4720)
Follow-up to address review feedback from [PR #4720](https://github.com/apache/datafusion-comet/pull/4720) (native `collect_list` / `array_agg`): https://github.com/apache/datafusion-comet/pull/4720#issuecomment-5071730068

The PR merges the core functionality; the items below are docs and test-coverage hardening deferred to a follow-on so they don't block the feature. All were flagged Medium severity.

### 1. Add a `collect_set` audit entry

`docs/source/contributor-guide/expression-audits/agg_funcs.md` documents `## collect_list` but has no `collect_set` entry, even though PR #4720 modifies `CometCollectSet` semantically (new RESPECT NULLS branch via `CometCollectShim.ignoreNulls(CollectSet)`, name-matched in `hasNativeArrayBufferAgg`). Add a `## collect_set` section alongside `## collect_list` (covering Spark 3.4.3 / 3.5.8 / 4.0.1 / 4.1.1, the `SparkCollectSet` delegation, the buffer-shape mismatch, and the `NaN`-dedup `Incompatible` reason on float/double), and replace the "same pattern already in use for collect_set" prose in the `collect_list` entry with a link to the new section.

### 2. Regression-cover the split-execution cascade

`spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala` — neither new `collect_list` test toggles `COMET_ENABLE_FINAL_HASH_AGGREGATE` / `COMET_ENABLE_PARTIAL_HASH_AGGREGATE`, so the `hasNativeArrayBufferAgg` + `tagUnsafePartialAggregates` path is untested. Mirror the existing `"mixed engine sum/avg"` tests (lines ~211/224) with:
- `mixed engine collect_list: Comet partial + Spark final matches Spark`
- `mixed engine collect_list: Spark partial + Comet final matches Spark` (the fallback case where Comet cannot read Spark's `BinaryType` buffer — `adjustOutputForNativeState` would misinterpret Binary state as an Array if it ran natively)

### 3. Assert the fallback reason in the distinct-combined test

In the new `"collect_list/collect_set combined with distinct aggregate falls back safely"` test, `checkSparkAnswer` doesn't verify anything ran/fell-back as expected. Use `checkSparkAnswerAndFallbackReason` (`CometTestBase.scala:296`) to assert the exact reason so the #4724 guard is actually protected:

```
Partial aggregate disabled: part of a multi-stage CollectList/CollectSet aggregate whose intermediate buffer cannot round-trip in Comet (issue #4724)
```

Apply to both the `collect_list` and `collect_set` variants, and to both the LocalTableScan and Parquet branches.

### 4. Extend SQL fixture datatype coverage

`spark/src/test/resources/sql-tests/expressions/aggregate/collect_list.sql` exercises `timestamp` but not `timestamp_ntz` (which is in `QueryPlanSerde.supportedDataType`, so it flows through the native `SparkCollectList` accumulator — an untested native path), and no year-month / day-time interval types (required by the "every datatype Spark accepts, NULLs in every column" rule; these exercise the Spark fallback). Append `timestamp_ntz` and ANSI interval blocks (both Spark 3.4+, already the file's floor). Gate the interval block with `-- MinSparkVersion: 3.4` if it can't persist to Parquet on some target versions.

andygrove:

Closes #5024.

## Rationale for this change

PR #4720 landed the native `collect_list` / `array_agg` implementation and deferred four docs and test-coverage items to a follow-on so they would not block the feature. This PR addresses those items.

Two of them close real gaps in the regression net: neither `collect_list` test toggled the partial/final hash aggregate configs, so the `hasNativeArrayBufferAgg` + `tagUnsafePartialAggregates` split-execution path was untested; and the distinct-combined test only compared answers, so it would have passed even if the #4724 guard stopped firing and the query silently ran natively.

1. **`collect_set` audit entry.** `docs/source/contributor-guide/expression-audits/agg_funcs.md` documented `collect_list` but not `collect_set`, even though #4720 changed `CometCollectSet` semantically. Adds a `## collect_set` section covering Spark 3.4.3 / 3.5.8 / 4.0.1 / 4.1.1, the `SparkCollectSet` delegation, the `BinaryType`-vs-list buffer-shape mismatch, and the `NaN`-dedup `Incompatible` reason on float/double. The `collect_list` entry now links to it instead of referring to it in prose.

2. **Split-execution cascade regression tests.** Two new tests in `CometAggregateSuite`, mirroring the existing `mixed engine sum/avg` pair:
   - `mixed engine collect_list: Comet partial + Spark final matches Spark` (`COMET_ENABLE_FINAL_HASH_AGGREGATE=false`)
   - `mixed engine collect_list: Spark partial + Comet final matches Spark` (`COMET_ENABLE_PARTIAL_HASH_AGGREGATE=false`), the case where `adjustOutputForNativeState` would misinterpret Spark's `BinaryType` buffer as a list if the final ran natively.

3. **Fallback-reason assertions in the distinct-combined test.** `checkSparkAnswer` is replaced with `checkSparkAnswerAndFallbackReason(s)`. The two branches are protected by different guards, so each asserts its own:
   - LocalTableScan branch: the Partial was never convertible, so `tagUnsafePartialAggregates` does not fire and the dependent stages fall back via the buffer-source check (`Comet aggregate that merges intermediate buffers requires a Comet child aggregate ... Incompatible aggregate function(s): collect_list`).
   - Parquet branch: the Partial would otherwise convert, so the #4724 tagging reason (`Partial aggregate disabled: part of a multi-stage CollectList/CollectSet aggregate whose intermediate buffer cannot round-trip in Comet (issue #4724)`) is asserted alongside the buffer-source reason.

4. **SQL fixture datatype coverage.** `collect_list.sql` gains a `timestamp_ntz` block (in `QueryPlanSerde.supportedDataType`, so it runs through the native `SparkCollectList` accumulator, previously untested) and a year-month / day-time ANSI interval block (not supported natively, so `spark_answer_only` covers the Spark fallback). No `MinSparkVersion` gate is needed: both work on 3.4, which is already the file's floor.

`CometAggregateSuite` and `CometSqlFileTestSuite` were run against both ends of the supported range, Spark 3.4 and the default Spark 4.1, and pass on both:

```
./mvnw test -Dsuites="org.apache.comet.exec.CometAggregateSuite collect_list" -Dtest=none
./mvnw test -Dsuites="org.apache.comet.CometSqlFileTestSuite collect_list" -Dtest=none
```

The `NaN` / `-0.0` deduplication behavior asserted in the new audit entry was verified against Spark directly rather than inferred: `collect_set` over `(0.0, -0.0, NaN, NaN, 1.0)` returns `[0.0, 1.0, NaN, NaN]`, confirming Spark keeps each `NaN` while collapsing `+0.0` and `-0.0`.


sunchao (approved):
Reviewed the collect_list / collect_set test hardening against the aggregate planning and fallback paths, including partial/final execution and timestamp/interval cases. I found no P1/P2 issues.

Exact-head CI passed. I did not rerun the full Scala aggregate suite locally.

PR #5056, opened 2026-07-27
docs: define post-1.0 versioning policy and legacy config process

andygrove:

Part of #4082.

## Rationale for this change

The existing versioning policy documents the `0.x` series and explicitly defers the post-1.0 rules: its "Road to 1.0" section says the stability commitments for configuration keys and the protobuf plan format "will be re-evaluated and documented as part of the `1.0.0` release." `config_conventions.md` already links to the versioning policy for alias-removal timing, so that cross-reference currently points at a rule that does not exist yet.

Semantic versioning is usually described in terms of API signatures, which does not fit Comet well. Comet is a plugin, not a library: the overwhelming majority of users never touch a Comet type, they install it through Spark's plugin system and configure it with `spark.comet.*` properties. Configuration is the primary API surface, and the policy needs to say so before 1.0 ships rather than after.

There is a genuine public Java and Scala API, but it is small, and today it is nowhere written down. Leaving it implied invites both failure modes: users treating an internal class as stable because it happens to be `public`, and contributors breaking the S3 SPI without realizing a vendor jar compiled against an earlier release depends on it. So this PR enumerates it in full.

Three positions in particular need to be settled up front, because they determine what a major version bump means:

- **Behavior changes.** Following Apache Spark, a behavior change may ship in a minor release provided users have a documented way to opt back out. That escape hatch is a `spark.comet.legacy.*` key, mirroring `spark.sql.legacy.*`.
- **Correctness fixes.** Comet's goal is to return what Spark returns. When a `Compatible` expression or operator does not, that is a bug, not a behavior users are entitled to rely on. Treating those fixes as breaking changes would make the major version number meaningless, since it would tick over for every fix.
- **Spark version support.** Comet's supported Spark matrix tracks the upstream project's maintenance windows. Tying it to semantic versioning would force Comet major releases on a schedule set by another project, and would make the major version say something about Spark rather than about Comet's own compatibility.

Documentation, plus a `@Public` annotation marking the API the documentation enumerates.

`docs/source/about/versioning_policy.md` is rewritten to be 1.0-forward. The "Pre-1.0 Releases" body collapses to a note recording that the `0.x` series remains unstable until `1.0.0` ships, and "Road to 1.0" is removed because this document answers it. New content:

- **Comet's Compatibility Surface.** What the policy covers (`spark.comet.*` keys, the plugin entry point, and query results for `Compatible` items) and what it does not (the protobuf plan format, all Scala/Java/Rust types, `EXPLAIN` output and plan shapes, metric names, log output, and performance characteristics including which operators fall back to Spark).
- **What Each Version Component Means.** What justifies a major, minor, and patch increment.
- **Behavior Changes and Legacy Configurations.** Defines a behavior change as the same query over the same data with the same explicit configuration producing a different result or error. Permits it in a minor release only when a `spark.comet.legacy.*` boolean restores the previous behavior, the upgrade guide gains an entry, and the release notes call it out. Legacy keys are deprecated on arrival and removable only in a major release.
- **Correctness Fixes Are Not Breaking Changes.** With two riders: maintainers may still add a legacy key for a fix with unusually wide blast radius, and following an upstream Spark change is likewise not a Comet breaking change.
- **Public Scala and Java API,** rewritten to enumerate the API rather than gesture at it. The previous section promised a deprecation cycle for "public classes and methods" without ever saying which ones those were. It is now an explicit list in two parts, with everything outside it declared internal whatever its access modifier says:
  - **Class names referenced from configuration**, in a table: `org.apache.spark.CometPlugin` (`spark.plugins`) and `org.apache.comet.ExtendedExplainInfo` (`spark.sql.extendedExplainProviders`). Users write these as config strings rather than compiling against them, so they get the same treatment as a config key rename. Their internal structure carries no guarantee.
  - **The S3 credential provider SPI**: `CometS3CredentialProvider`, `CometS3Credentials`, `CometS3CredentialContext`, and `CometS3AccessMode` in `org.apache.comet.cloud.s3`. Vendors compile against these with `provided` scope and ship a separate jar, so source *and* binary compatibility both matter. Additive changes such as a new accessor on `CometS3CredentialContext` are allowed in a minor release, matching the promise `s3-credential-providers.md` already makes to vendors; anything that would break an existing vendor jar, including adding an abstract method without a default, requires a major. `CometS3CredentialDispatcher` is called out as internal despite sharing the package, since it is the JNI entry point.
- **Spark version support placed outside semantic versioning,** folded into the existing Apache Spark Version Support section. Adding or removing a Spark minor, or a Scala binary version, is always a Comet minor release.

`docs/source/user-guide/latest/migration-guide.md` is new, and is added to the user guide toctree as "Upgrading Comet". It mirrors Spark's SQL migration guide: one section per Comet version listing behavior changes alongside the key that reverts each. It is seeded with the page structure and a `1.0.0` section; `0.x` history is not backfilled.

`docs/source/contributor-guide/config_conventions.md` gains "Changing the Behavior of an Existing Config" beside the existing rename section, covering the `spark.comet.legacy.*` recipe, the `CATEGORY_LEGACY` category, a checklist, and the two cases that do not need a legacy key.

### Marking the public API in source

A list in a document is easy to contradict without noticing, so the enumeration is also expressed in code:

- **`org.apache.comet.annotation.Public`** (new) is applied to all six members: the four `org.apache.comet.cloud.s3` SPI types, `CometPlugin`, and `ExtendedExplainInfo`. Its javadoc states the guarantee and that anything unannotated is internal.
- **`CometPublicApiSuite`** (new) scans the module's compiled classes and asserts the annotated set is *exactly* the six. Adding or removing `@Public` fails the build until the policy document and the suite are updated in the same PR, which is the intent: growing the public API commits the project to supporting the addition indefinitely, and that is a decision worth forcing into the open rather than a routine code change. It is registered in `pr_build_linux.yml` and `pr_build_macos.yml` next to `CometConfSuite`, the other suite that needs no Spark session.

Note the annotation is `@Public`, not the `@Unstable` the old policy anticipated. `@Unstable` says "may change at any time, no deprecation cycle," which is the opposite of what these six members now promise, and it would have had zero usages. The gap worth closing was that the public API itself was unmarked. If a member later needs an explicit exemption, `@Unstable` can be added then, against a real case.

A couple of things deliberately left out:

- `CATEGORY_LEGACY` is documented as the convention but not declared in `CometConf.scala`; that belongs to whichever PR first adds a real legacy config.
- The `@Unstable` annotation discussion is deleted from the policy rather than reworded, for the reason above.

Open question for reviewers: `understanding-comet-plans.md` shows `import org.apache.comet.ExtendedExplainInfo` in a Scala sample, and the class is now `@Public`, so that sample is on firmer ground than it was. What is *not* promised is the rest of that file, in particular `CometCoverageStats`, which is returned by the public class's neighbours but is itself internal. Worth a look from someone who knows whether users lean on it.

`CometPublicApiSuite` is new and covers the annotation work: it fails if a type gains or loses `@Public` without the policy list being updated to match. It was verified in both directions rather than only observed to pass, on a clean `spark/target/classes` rebuild:

- Passes against the six enumerated types.
- Removing `@Public` from `CometS3AccessMode` fails with `missing @Public: org.apache.comet.cloud.s3.CometS3AccessMode`.
- Adding a stray `@Public` to an internal Scala class fails with `unexpectedly annotated: org.apache.comet.CometCoverageStats`, confirming the scan sees the whole module and not just the expected set.

`dev/ci/check-suites.py` also passes, which it did not on the first push: the preflight job requires every `*Suite.scala` to be named in both PR build workflows, and the new suite was not yet registered.

The policy and guide changes are documentation, so there is no test for them. Verified locally:

- `sphinx -b html docs/source` builds successfully with no warnings attributable to the added or modified pages, and the new upgrade guide is reachable from the user guide navigation.
- All relative links and internal anchors in the touched files resolve, in both directions across the versioning policy, the contributor guide, and the new upgrade guide.
- `prettier --check` passes on all touched markdown, and `spotless` passes on the Java and Scala changes.


spark/src/main/java/org/apache/comet/cloud/s3/CometS3CredentialContext.java
```diff
@@ -21,12 +21,15 @@
...
+import org.apache.comet.annotation.Public;
+
 /**
  * Per-request context passed to {@link
  * CometS3CredentialProvider#getCredentialsForPath(CometS3CredentialContext)}. New fields can be
  * added here without changing the SPI method signature, so vendors compiled against earlier
  * versions stay binary-compatible.
  */
+@Public
 public final class CometS3CredentialContext {
```
andygrove, line 33: @mbutrovich How confident are we that we can maintain these APIs without breaking changes? Should we add another annotation like `@Unstable` ?

docs/source/user-guide/latest/migration-guide.md
```diff
@@ -0,0 +1,63 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Comet Upgrade Guide
```
andygrove, line 20: @comphead I added the migration guide as part of this PR based on your feedback on my config PRs

kazuyukitanimura (approved):

spark/src/main/java/org/apache/comet/cloud/s3/CometS3CredentialContext.java
mbutrovich: I have not heard of a use case that we're missing from the current API, and a couple of vendors have used it successfully AFAIK.

Issue #5061, opened 2026-07-28
[EPIC] Complete interval type support: scan, operator gates, and remaining interval expressions


andygrove:
### What is the problem the feature request solves?

#4540 tracked the foundational interval work: map the three Spark interval types to Arrow and get them through the FFI, serde and codegen-dispatch paths. That part has now largely landed via #4898 (CalendarIntervalType) and #4976 (nested values and native shuffle), so #4540's premise ("Comet has no support for Spark's interval data types") no longer describes the state of the tree. This issue supersedes it and tracks what the audit of `1d3044f35` found still missing.

The headline result: the native engine handles `Interval` / `Duration` arrays better than the Scala side admits. Almost every remaining gap is a type gate in Scala, not missing Rust.

#### Verified working today

All of the following ran fully native and matched Spark on `1d3044f35` (Spark 4.1 profile), against Parquet-backed data so the expressions were not constant-folded:

- `make_dt_interval`, `make_ym_interval`, `multiply_dt_interval`
- an interval column through an equi-join, a multi-key sort, `first` / `last`, `coalesce` / `if`, and `=`
- native shuffle carrying an ANSI interval data column

#### Verified broken or falling back

| Area | Behavior on `1d3044f35` |
|---|---|
| `hash` / `xxhash64` of a CalendarInterval | **Query fails**: `Unsupported data type in hasher: Interval(MonthDayNano)` |
| Null CalendarInterval literal | **Query fails** (#5058) |
| Parquet scan of an ANSI interval column | Entire scan falls back |
| Single-column sort on an interval | `Sort` falls back (multi-column sort is not checked and works) |
| `min` / `max` / `sum` / `avg` over intervals | `HashAggregate` falls back |
| `GROUP BY <interval>` | `Exchange` falls back |
| Window with an interval in `PARTITION BY` / `ORDER BY` / the aggregate | `Window` falls back |
| `hash` / `xxhash64` of an ANSI interval | `Project` falls back |
| `-interval`, `abs(interval)` | `Project` falls back, despite native support existing |
| `cast(interval as string)`, `cast(string as interval)` | `Project` falls back |
| `extract` / `date_part` of an interval field | `Project` falls back |
| `date + interval`, `ts + interval`, `ts - ts`, `date - date` | `Project` falls back |
| `interval / n`, `ym_interval * n` | `Project` falls back |

#### 1. Correctness bugs (do these first)

[...]

Issue #4540, opened 2026-05-30
[EPIC] Support Spark interval types (CalendarInterval / YearMonthInterval / DayTimeInterval) and interval expressions


andygrove:
## What is the problem the feature request solves?

Comet has no support for Spark's interval data types:

- `CalendarIntervalType` (months + days + microseconds)
- `YearMonthIntervalType` (ANSI `INTERVAL YEAR TO MONTH`)
- `DayTimeIntervalType` (ANSI `INTERVAL DAY TO SECOND`)

Because the types are unsupported, every expression that produces or consumes an interval falls back to Spark, and any query carrying an interval column through a Comet operator falls back as well. `CometBatchKernelCodegen.isSupportedDataType` also rejects these types, so they cannot even be routed through the JVM codegen dispatcher (see #4506 / #4538): the interval expressions are a genuine arrow-native gap with no stopgap.

This issue tracks the foundational type support plus the dependent expression family. It is the prerequisite for the already-filed per-expression requests below.

### Type support (prerequisite)

- Map the three Spark interval types to Arrow:
  - `YearMonthIntervalType` -> Arrow `Interval(YearMonth)`
  - `DayTimeIntervalType` -> Arrow `Interval(MonthDayNano)` / `Duration` (decide representation that round-trips with Spark's microsecond storage)
  - `CalendarIntervalType` -> Arrow `Interval(MonthDayNano)` (Spark stores months/days/micros)
- Wire the types through the `CometVector` hierarchy, FFI import/export (`NativeUtil` / `scan.rs`), and `serializeDataType` in `QueryPlanSerde`.
- Allow these types in `CometBatchKernelCodegen.isSupportedDataType` once the FFI path is correct, so codegen dispatch can also cover interval expressions.

### Expressions (depend on the type work)

Constructors and arithmetic already tracked individually:

- `make_interval` (#3099), `make_dt_interval` (#3098), `make_ym_interval` (#3100), `try_make_interval` (#3103)
- `multiply_ym_interval` (#3102), `multiply_dt_interval` (#3101), `divide_dt_interval` (#3096)
- `date_add_interval` (#3086), `timestamp_add_interval` (#3114), `timestamp_add_ym_interval` (#3115), `time_add_interval` (#3121)
- `subtract_timestamps` (#3112), `subtract_dates` (#3094), `subtract_times` (#3139), `datetime_sub` (#3134), `datetime_add`
- `extract` / `date_part` of interval fields

[...]

Issue #4463, opened 2026-05-27
[Bug] translate uses graphemes vs Spark code points and ignores U+0000 deletion


andygrove:
## Describe the bug

`translate` is wired as `CometScalarFunction("translate")` and currently reports `Compatible`, but DataFusion's `translate` diverges from Spark in two ways:

1. **Grapheme vs code-point semantics.** DataFusion iterates over Unicode graphemes; Spark uses code points (via `Character.charCount`). For supplementary BMP code points these match, but for multi-code-point graphemes (combining marks, ZWJ sequences such as flag emoji) the two implementations disagree.
2. **NUL byte in the `to` argument.** Spark's `StringTranslate.buildDict` treats any character mapped to U+0000 in `to` as a deletion. DataFusion substitutes U+0000 instead.

Surfaced by the string-expressions audit in apache/datafusion-comet#4461.

## Steps to reproduce

```sql
-- (1) grapheme vs code point: combining mark
SELECT translate(concat('e', char(0x0301)), 'e', 'a');

-- (2) U+0000 deletion: expected to delete 'b' under Spark
SELECT translate('abc', 'b', char(0));
```

Spark deletes the matched character in the second query; Comet substitutes a NUL character. Spark's per-code-point translation and Comet's grapheme-based translation diverge for combining-mark inputs.

## Expected behavior

Match Spark behavior, or downgrade `translate` to `Incompatible(Some(...))` so the non-ASCII path falls back unless explicitly enabled.

## Additional context

- Comet wiring: `QueryPlanSerde.scala` -> `classOf[StringTranslate] -> CometScalarFunction("translate")`
- Spark reference: `UTF8String.translate(dict)` with `StringTranslate.buildDict`
- DataFusion impl: `datafusion-functions::unicode::translate`

Issue #4219, opened 2026-05-05
native_datafusion more permissive than Spark 3.x when reading Parquet TimestampNTZ columns


andygrove:
### Describe the bug

Spark 3.x does not allow reading Parquet timestamp columns as TimestampNTZ due to the ambiguity. Spark 4.0 relaxed this restriction.

See https://issues.apache.org/jira/browse/SPARK-47447 for details.

The behavior of `native_datafusion` scan is correct when running with Spark 4.0+ but incorrect when running with Spark 3.x because it returns a value rather than throwing an exception.

The value returned is correct per Spark 4 definition, so this is not a data correctness issue per se, but Comet is more permissive with Spark 3.x

PR #5062, opened 2026-07-28
docs: refresh stale issue references and normalize issue link format

andygrove:

Closes #.

## Rationale for this change

Several pages in the user guide and contributor guide link to issues that have since been closed, and in most of those cases the surrounding prose still describes the limitation the issue was filed for. A reader following the link lands on a closed issue and has no way to tell whether the documented behavior is still current.

A separate problem is that issue and PR references are written four different ways across these pages: inline `[#NNNN](url)`, bare URLs, bare `#NNNN` with no link at all, and reference-style `[#NNNN]:` definitions. The bare forms are not clickable in the rendered docs.

Every issue and PR reference in `docs/source/user-guide` and `docs/source/contributor-guide` was checked against its current state on GitHub. Where an issue was closed, the claim in the docs was re-verified against the serde and native code rather than taken from the issue state alone.

Thirteen places described a limitation that no longer exists:

| Page | Previous claim | Current behavior |
| --- | --- | --- |
| `collection_funcs.md` | `size()` over `MapType` falls back | `CometSize` returns `Compatible()` for `ArrayType` and `MapType` |
| `math_funcs.md` | `CometRemainder` rejects `EvalMode.TRY`, so `try_mod` falls back | gates only on the left input type; all three eval modes serialize |
| `conversion_funcs.md` | `CAST(map AS map)` falls back | runs natively via `cast_map_to_map`; support recurses into the key and value casts |
| `conversion_funcs.md` | `spark.sql.legacy.castComplexTypesToString` is not honoured | detected in `CometCast` and reported `Unsupported` |
| `string_funcs.md` | `replace` marks `Incompatible` on a literal empty search string | `Compatible` with a `NativeOptIn`; codegen dispatch by default |
| `string_funcs.md` | `initcap` is unconditionally `Incompatible` | `Compatible` with a `NativeOptIn` |
| `map_funcs.md` | `str_to_map` ignores `spark.sql.legacy.truncateForEmptyRegexSplit` | reads the config and reports `Incompatible`, routing to codegen dispatch |
| `window_funcs.md` | `AVG(decimal)` over a window is dead on Spark 4.x | `CometWindowExec.convert` unwraps the `Cast(Divide(...))` shape |
| `agg_funcs.md` | `Percentile` reports `Incompatible`, native path opt-in | plain `Compatible()`, native by default |
| `pyarrow-udfs.md` | the row to Arrow round-trip is unchanged | `CometArrowPythonRunner` consumes `ColumnarBatch` directly |
| `roadmap.md` | datafusion-spark migration tracked under #2084 | closed, superseded by #4150 |
| `roadmap.md` | awslabs TPC-DS results tracked under #3799 | closed; the published results improved, so the section was trimmed |
| `datatypes.md` | "Interval types fall back to Spark today" | all three types moved from planned to supported-with-caveats, with the per-operator gaps listed and tracked by #5061 |

Links that pointed at a closed or unrelated tracker were dropped or repointed: the ten `tracking #4098` rows in `expressions.md` (that issue tracked Spark 4.1 CI failures and never covered `kurtosis`, `skewness`, or the Variant functions), `translate` (#4463, where the incompatibility is real but the issue is closed, so the reason is now stated inline), the `TimestampNTZ` scan note (#4219), the `from_utc_timestamp` timezone-parser note (#2013, whose linked issue was "Add from_utc_timestamp support"), and the `make_interval` / `try_make_interval` rows (#4540 to #5061).

Seven closed-issue references were kept deliberately because they are historical or illustrative: the 0.9.1 jar-size incident in `release_process.md`, the hash-aggregate epic in `bug_triage.md` (now explicitly labelled as closed), the `to_json` code sample in `adding_a_new_expression.md` that mirrors live code in `structs.scala`, the two `query ignore(...)` syntax examples in `sql-file-tests.md`, and the two audit notes whose workarounds are still present in the code.

For formatting, 52 bare URLs and 27 bare `#NNNN` mentions were converted to inline `[#NNNN](url)`. Reference-style definitions were left in place in `roadmap.md` and `iceberg-spark-tests.md`, which already used that style consistently. Cross-repo references use the `[apache/arrow-rs#10028](url)` form. 22 URLs were repointed from `/issues/` to `/pull/` where the number is a pull request. URLs inside fenced code blocks were left untouched, including the `issues/NNNN` placeholder in `adding_a_new_spark_version.md`.

Documentation only, so there is no runtime behavior to test. Every claim that changed was checked against the code it describes (`CometSize`, `CometRemainder`, `CometCast`, `CometStringReplace`, `CometInitCap`, `CometStrToMap`, `CometWindowExec`, `CometPercentile`, `CometArrowPythonRunner`). Link formatting was verified with a script that finds any GitHub issue or PR reference outside a fenced code block that is not an inline link or a reference definition; it reports zero remaining. `prettier --check` passes on both guides.


mbutrovich (approved):
LGTM, thanks @andygrove!

PR #5079, opened 2026-07-28
feat: add codegen dispatch fallback for CometCast incompatible/unsupported cases (including legacy config paths)

comphead:
Split from #4799 

## Summary

  - Mix `CodegenDispatchFallback` into `CometCast` so cast expressions that Comet declines to run natively (e.g. `spark.sql.legacy.castComplexTypesToString.enabled=true`) can execute via the JVM codegen
  dispatcher instead of falling all the way back to Spark.
  - Guard `VariantType` (Spark 4) in `CometTypeShim` and reject Variant casts in `CometCast.convert`/`isSupported`, since neither the native serde nor the codegen dispatcher can carry `VariantType` in
  data args or return type.
  - Deduplicate the complex-type check with `DataTypeSupport.isComplexType` and trim the `legacyCastComplexTypesToString` comment to reflect the new dispatch path.
  - Refresh SQL golden tests: drop `expect_fallback` markers from non-legacy `cast_complex_types_to_string`, keep them on the legacy variant, and run `cast_string_to_date_time_parser_policy_legacy` in
  plain query mode.
  - Update `CometCastSuite` / `CometExpressionSuite` expectations and the `4.0.2` / `4.1.2` spark diffs so `VariantEndToEndSuite` accepts a Comet plan under codegen dispatch.


spark/src/main/scala/org/apache/comet/expressions/CometCast.scala
```diff
@@ -25,30 +25,41 @@ import org.apache.spark.sql.types.{ArrayType, DataType, DataTypes, DecimalType,
...
     SQLConf.get
       .getConfString("spark.sql.legacy.castComplexTypesToString.enabled", "false")
       .toBoolean
 
+  // Non-CORRECTED policies (LEGACY, EXCEPTION) change string-to-date/timestamp parsing behavior in
+  // ways the native cast kernel does not replicate.
+  private def isNonDefaultTimeParserPolicy: Boolean =
+    !SQLConf.get
+      .getConfString("spark.sql.legacy.timeParserPolicy", "CORRECTED")
+      .equalsIgnoreCase("CORRECTED")
```
andygrove: This does not seem correct. The default `timeParserPolicy` in Spark 3.5.8 is `EXCEPTION`.

andygrove: To be more specific, I am really saying that the variable name `isNonDefaultTimeParserPolicy` is incorrect.

comphead: thanks @andygrove , 
```
- Since Spark 4.0, The default value for `spark.sql.legacy.timeParserPolicy` has been changed from `EXCEPTION` to `CORRECTED`. Instead of raising an `INCONSISTENT_BEHAVIOR_CROSS_VERSION` error, `CANNOT_PARSE_TIMESTAMP` will be raised if ANSI mode is enable. `NULL` will be returned if ANSI mode is disabled. See [Datetime Patterns for Formatting and Parsing](sql-ref-datetime-pattern.html).
```

checking it

andygrove: Claude tells me that `timeParserPolicy` has no impact on cast operations. Could you confirm this is really an issue?

spark/src/main/scala/org/apache/comet/expressions/CometCast.scala
```diff
@@ -159,14 +178,22 @@ object CometCast extends CometExpressionSerde[Cast] with CometExprShim {
...
       return Compatible()
     }
 
-    if (toType == DataTypes.StringType && legacyCastComplexTypesToString && (fromType
-        .isInstanceOf[ArrayType] || fromType.isInstanceOf[StructType] ||
-        fromType.isInstanceOf[MapType])) {
-      return Unsupported(Some(legacyCastComplexTypesToStringReason))
+    if (toType == DataTypes.StringType && legacyCastComplexTypesToString && isComplexType(
+        fromType)) {
+      return Incompatible(Some(legacyCastComplexTypesToStringReason))
```
andygrove: Since `CodegenDispatchFallback` already routes `Unsupported` through the dispatcher, keeping this as `Unsupported` would give the same in-pipeline behavior. Switching to `Incompatible` additionally allows the native cast to run when `spark.comet.expr.allowIncompatible=true`, and the native cast produces the default formatting rather than the legacy formatting the user explicitly asked for with the legacy flag. That would be a silent results change for those users, where today they always get correct results via fallback. Was that intentional? If the goal is just to stay in the Comet pipeline, `Unsupported` plus the mixin achieves it without that risk.

spark/src/main/scala/org/apache/comet/expressions/CometCast.scala
```diff
@@ -159,14 +178,22 @@ object CometCast extends CometExpressionSerde[Cast] with CometExprShim {
       timeZoneId: Option[String],
       evalMode: CometEvalMode.Value): SupportLevel = {
 
+    // Spark 4's `VariantType` (SPARK-45827) has no native counterpart in Comet: serializing it
+    // into the DataFusion plan would fail in `serializeDataType`, and the codegen dispatcher
+    // cannot compile Variant read/write kernels either. Detect it via the version-shimmed
+    // `isVariantType` (which returns false on Spark 3.x where the class does not exist) and
+    // report `Unsupported` so the enclosing operator falls back to Spark.
```
andygrove: This comment (and the one in `getSupportLevel` above) says the guard forces Spark fallback, but with the `CodegenDispatchFallback` mixin an `Unsupported` result first attempts the codegen dispatcher. Variant casts fall back only because the dispatcher fails to serialize `VariantType` in the data args or return type. The end behavior is the same, but could the comments describe the actual routing?

It might also be worth adding a Spark 4.x test that casts to and from variant, for example `CAST(parse_json('{"a":1}') AS STRING)` on a column, to pin down that this path cleanly falls back and matches Spark. I could not find any Variant cast coverage in the repo today.

spark/src/test/scala/org/apache/comet/CometCastSuite.scala
```diff
@@ -1814,9 +1801,7 @@ class CometCastSuite extends CometTestBase with AdaptiveSparkPlanHelper {
           assert(
             CometCast.isSupported(fromType, toType, None, CometEvalMode.LEGACY) ==
               Unsupported(Some(expectedMessage)))
-          checkSparkAnswerAndFallbackReason(
-            data.select(col("a").cast(toType).as("converted")),
-            expectedMessage)
+          checkSparkAnswerAndOperator(data.select(col("a").cast(toType).as("converted")))
```
andygrove, line 1817: The test is still named "cast ArrayType(DateType) to unsupported ArrayType falls back" but it now asserts the Comet operator is present. Could you rename it to reflect the dispatcher routing?

andygrove (commented):
Thanks @comphead. I like the direction of routing these casts through the codegen dispatcher instead of falling the whole projection back to Spark. I left a few comments. The main one is in the existing thread about `timeParserPolicy`, where I confirmed against Spark source that the policy has no effect on cast.

Also, could you fill in the PR template sections? "Closes #." is empty and the rationale and testing sections are still the template comments, even though the free-form description at the top is good.

spark/src/main/scala/org/apache/comet/expressions/CometCast.scala
andygrove: I checked the Spark source to close this out. `Cast.castToDate` and `castToTimestamp` for string inputs call `DateTimeUtils.stringToDate` and `stringToTimestamp` directly, and there are no references to `timeParserPolicy` anywhere in `Cast.scala` on either master or v3.5.8. The policy is only consulted by `DateFormatter` and `TimestampFormatter`, which back `to_date`, `to_timestamp`, `date_format`, and the CSV/JSON/XML parsers. Cast is unaffected.

Could we remove this branch, the reason string, and `cast_string_to_date_time_parser_policy_legacy.sql`? The new test passes with or without this change because the results are identical under any policy value, so it does not actually verify anything.

If we ever did need to read this conf, note that `getConfString` with a hard-coded `"CORRECTED"` default ignores the registered default, which is `EXCEPTION` on Spark 3.x, so the typed `SQLConf.get.legacyTimeParserPolicy` getter would be the safer read.

comphead:
Thanks @andygrove for the review, I'm going through it

spark/src/main/scala/org/apache/comet/expressions/CometCast.scala
```diff
@@ -25,24 +25,24 @@ import org.apache.spark.sql.types.{ArrayType, DataType, DataTypes, DecimalType,
...
 import org.apache.comet.serde.QueryPlanSerde.{evalModeToProto, exprToProtoInternal, serializeDataType}
-import org.apache.comet.shims.CometExprShim
+import org.apache.comet.shims.{CometExprShim, CometTypeShim}
 
-object CometCast extends CometExpressionSerde[Cast] with CometExprShim {
+object CometCast
+    extends CometExpressionSerde[Cast]
+    with CometExprShim
+    with CometTypeShim
+    with CodegenDispatchFallback {
```
andygrove, line 38: Adding a codegen dispatch fallback for Cast is a good idea and I think this should be split out into a standalone PR separate from legacy config handling

andygrove: Ignore this comment. I see that you removed the legacy config handling now

spark/src/main/scala/org/apache/comet/expressions/CometCast.scala
```diff
@@ -25,24 +25,24 @@ import org.apache.spark.sql.types.{ArrayType, DataType, DataTypes, DecimalType,
...
 
   // Shared with CometCastSuite so the asserted reason cannot drift from production.
   private[comet] val negativeScaleDecimalToStringReason: String =
     "Negative-scale decimal requires spark.sql.legacy.allowNegativeScaleOfDecimal=true"
 
-  // When `spark.sql.legacy.castComplexTypesToString.enabled` is true, Spark wraps maps and
-  // structs with `[]` (instead of `{}`) when casting to string, and omits NULL elements of
-  // structs/maps/arrays (instead of rendering them as the literal "null"). Comet only
-  // implements the default formatting, so fall back to Spark for any array/map/struct to-string
-  // cast when the flag is enabled. The flag is internal in Spark 4.0 and defaults to false.
```
andygrove, line 43: Why is this comment removed? It seems useful to me

spark/src/main/scala/org/apache/comet/expressions/CometCast.scala
```diff
@@ -25,24 +25,24 @@ import org.apache.spark.sql.types.{ArrayType, DataType, DataTypes, DecimalType,
...
   private[comet] val legacyCastComplexTypesToStringReason: String =
-    "spark.sql.legacy.castComplexTypesToString.enabled=true is not supported"
+    "spark.sql.legacy.castComplexTypesToString.enabled=true is not supported natively"
```
andygrove: It is not necessary to specify "natively" in the message. The documentation for `getIncompatibleReasons` explains its purpose as `Get documentation for ways the native implementation of this expression may differ from Spark`. The incompatibility reasons are always specific to native code. The documentation generator already generates prose explaining the context.

andygrove: The same applies to `getUnsupportedReason`

spark/src/test/scala/org/apache/comet/CometCastSuite.scala
```diff
@@ -735,19 +735,6 @@ class CometCastSuite extends CometTestBase with AdaptiveSparkPlanHelper {
...
-    // (e.g. 12300 stored as Decimal(7,-2) with unscaled=123 → "1.23E+4").
-    // CometCast.canCastToString checks the
-    // config and returns Incompatible when it is false.
-    //
-    // Parquet does not support negative-scale decimals so we use checkSparkAnswer directly
-    // (no parquet round-trip) to avoid schema coercion.
-
-    // With config enabled, enable localTableScan so Comet can take over the full plan
-    // and execute the cast natively. Parquet does not support negative-scale decimals so
-    // the data is kept in-memory; localTableScan.enabled bridges that gap.
```
andygrove, line 750: Why is the comment removed?

andygrove:
Thanks for iterating on this @comphead. I think this is close now. I left feedback. Could you also update the PR title and description as they are now outdated.

andygrove (approved):
LGTM pending CI. PR title and description still need updating though. Thanks @comphead

andygrove:
Once this merges, we can stop enabling incompatible casts in the stability suite

comphead:
Thanks @andygrove for the cycles spent on review :)

Issue #5078, opened 2026-07-28
[EPIC] ANSI mode audit follow-ups (vs Spark 4.1.1)


andygrove:
### What is the problem the feature request solves?

This epic tracks the follow-ups from a full audit of ANSI mode support across all natively-supported Comet expressions, compared against Spark 4.1.1 source. The audit covered every expression whose runtime behavior depends on `failOnError` / `evalMode` / `SQLConf.get.ansiEnabled` in Spark 4.1 (arithmetic, aggregates, decimal internals, Cast/TryEval, collections, datetime, math/string/url), including functions wired in from the datafusion-spark crate.

Since Comet no longer has a global ANSI fallback config, ANSI correctness is entirely per-expression, and ANSI is the default in Spark 4.x, which raises the stakes for every gap below.

### Findings: correctness bugs (ANSI silently ignored or wrong result)

- [ ] #5065 IntegralDivide: `Long.MinValue div -1` returns wrong value instead of throwing under ANSI
- [ ] #5066 MakeDecimal: `fail_on_error` serialized but discarded natively, NULL where Spark throws
- [ ] #5067 Remainder: Float/Double `% 0` returns NaN under ANSI instead of throwing
- [ ] #5068 Cast boolean to decimal ignores eval mode: throws where Spark returns NULL in legacy/try mode
- [ ] #5070 round on LongType with scale <= -19 returns 0 instead of overflowing (both modes)

### Findings: error fidelity (throw/NULL decision correct, wrong exception type or message)

- [ ] #5071 Arithmetic overflow errors: wrong error class for Byte/Short, wrong type names, missing `try_` suggestions
- [ ] #5072 Native ANSI errors raised as Arrow errors bypass SparkError conversion (wide decimal, decimal divide, decimal-to-decimal cast)
- [ ] #5073 next_day and make_date ANSI errors surface as CometNativeException instead of Spark exception classes

### Findings: enhancements and hardening

- [ ] #5069 Cast float/double to decimal: return NULL for NaN/Infinity under ANSI (pair is already marked incompatible)
- [ ] #5074 Guard against silent `fail_on_error` loss in scalar function wiring
- [ ] #5075 DecimalPrecision.promote uses session ANSI flag instead of per-expression eval mode (latent)
- [ ] #5076 Test coverage: re-enable stale ignored ANSI tests, add missing ANSI cases, strengthen error-class assertions
- [ ] #5077 Clean up dead ANSI-related plumbing

[...]

PR #5080, opened 2026-07-28
fix: honor fail_on_error in native make_decimal

### LINKED ISSUE #5066: MakeDecimal ignores fail_on_error: returns NULL where Spark throws under ANSI mode
### Describe the bug

`MakeDecimal` returns NULL on overflow where Spark throws under ANSI mode. The Scala serde serializes the flag correctly, but the native registration discards it.

Spark 4.1 (`decimalExpressions.scala`): `MakeDecimal(child, precision, scale, nullOnOverflow)` with `nullOnOverflow = !SQLConf.get.ansiEnabled`. With `nullOnOverflow = false` it calls `Decimal.set(longInput, precision, scale)`, which throws `NUMERIC_VALUE_OUT_OF_RANGE` on overflow.

Comet serde (`spark/src/main/scala/org/apache/comet/serde/decimalExpressions.scala:56-60`) passes `failOnError = !expr.nullOnOverflow` via `scalarFunctionExprToProtoWithReturnType`, and the planner forwards it (`native/core/src/execution/planner.rs`, `create_comet_physical_fun(..., Some(expr.fail_on_error))`). But the native registration uses the macro arm that does not thread the flag through:

```rust
// native/spark-expr/src/comet_scalar_funcs.rs:142-144
"make_decimal" => {
    make_comet_scalar_udf!("make_decimal", spark_make_decimal, data_type)
}
```

and `spark_make_decimal` (`native/spark-expr/src/math_funcs/internal/make_decimal.rs`) has no `fail_on_error` parameter at all; `long_to_decimal` unconditionally returns `None` on overflow.

### Steps to reproduce

`MakeDecimal` is generated by the `DecimalAggregates` optimizer rule rather than user SQL, so it surfaces on decimal aggregations over low-precision decimals (sum/avg of `DECIMAL(p, s)` with small `p`) where the unscaled long overflows the target precision, with `spark.sql.ansi.enabled=true`. Spark throws `NUMERIC_VALUE_OUT_OF_RANGE`; Comet returns NULL.

### Expected behavior

The native `make_decimal` should receive `fail_on_error` (compare the `"spark_modulo"` registration in `comet_scalar_funcs.rs`, which uses the `without $data_type, $fail_on_error` arm) and raise `SparkError::NumericValueOutOfRange` on overflow when set.

### Additional context

Found by an ANSI-mode audit of all native expressions against Spark 4.1.1. This is an instance of a general hazard: the scalar-function wiring makes it easy to serialize `fail_on_error` and silently lose it natively.

andygrove:

Closes #5066.

## Rationale for this change

`MakeDecimal` with `nullOnOverflow = false` (ANSI mode) must fail on overflow, matching Spark's `Decimal.set(unscaled, precision, scale)`, which throws `NUMERIC_VALUE_OUT_OF_RANGE`. The Scala serde already serializes `failOnError = !nullOnOverflow`, but the native registration of `make_decimal` discarded the flag, so Comet returned NULL where Spark throws. Since `MakeDecimal` is generated by the `DecimalAggregates` optimizer rule, the divergence surfaces on decimal aggregations rather than user-written SQL.

- `spark_make_decimal` now takes `fail_on_error` and raises `SparkError::NumericValueOutOfRange` (via the existing `decimal_overflow_error` helper, as `CheckOverflow` already does) when the unscaled long does not fit the target precision. Non-ANSI behavior is unchanged: overflow becomes NULL.
- The `make_decimal` registration in `comet_scalar_funcs.rs` threads `fail_on_error` through instead of dropping it.
- Test-only: `createMakeDecimalColumn` in `ShimCometTestBase` (all Spark version variants) gained an overload that accepts `nullOnOverflow`, since the existing helper hardcoded the null-on-overflow behavior.

- New Rust unit tests in `make_decimal.rs` covering array and scalar overflow with `fail_on_error` set (error) and unset (NULL), plus NULL input under `fail_on_error`.
- New JVM test `make decimal using DataFrame API - overflow throws when nullOnOverflow=false` in `CometExpressionSuite`, asserting Spark and Comet both fail with the `cannot be represented as Decimal(3, 0)` message. The test was verified to fail before the native fix and pass after it.

This issue was found by an audit run with the `audit-comet-expression` project skill (tracked under epic #5078).

### DIFF
--- a/native/spark-expr/src/math_funcs/internal/make_decimal.rs
+++ b/native/spark-expr/src/math_funcs/internal/make_decimal.rs
@@ -15,13 +15,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::error::decimal_overflow_error;
 use crate::math_funcs::utils::get_precision_scale;
 use arrow::datatypes::DataType;
 use arrow::{
     array::{AsArray, Decimal128Builder},
     datatypes::{validate_decimal_precision, Int64Type},
 };
-use datafusion::common::{internal_err, Result as DataFusionResult, ScalarValue};
+use datafusion::common::{internal_err, DataFusionError, Result as DataFusionResult, ScalarValue};
 use datafusion::physical_plan::ColumnarValue;
 use std::sync::Arc;
 
@@ -29,12 +30,13 @@ use std::sync::Arc;
 pub fn spark_make_decimal(
     args: &[ColumnarValue],
     data_type: &DataType,
+    fail_on_error: bool,
 ) -> DataFusionResult<ColumnarValue> {
     let (precision, scale) = get_precision_scale(data_type);
     match &args[0] {
         ColumnarValue::Scalar(v) => match v {
             ScalarValue::Int64(n) => Ok(ColumnarValue::Scalar(ScalarValue::Decimal128(
-                long_to_decimal(n, precision, scale),
+                long_to_decimal(n, precision, scale, fail_on_error)?,
                 precision,
                 scale,
             ))),
@@ -45,7 +47,7 @@ pub fn spark_make_decimal(
                 let arr = a.as_primitive::<Int64Type>();
                 let mut result = Decimal128Builder::new();
                 for v in arr.into_iter() {
-                    result.append_option(long_to_decimal(&v, precision, scale))
+                    result.append_option(long_to_decimal(&v, precision, scale, fail_on_error)?)
                 }
                 let result_type = DataType::Decimal128(precision, scale);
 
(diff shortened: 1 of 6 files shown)


native/spark-expr/src/math_funcs/internal/make_decimal.rs
```diff
@@ -58,14 +60,83 @@ pub fn spark_make_decimal(
...
+/// otherwise.
 #[inline]
-fn long_to_decimal(v: &Option<i64>, precision: u8, scale: i8) -> Option<i128> {
+fn long_to_decimal(
+    v: &Option<i64>,
+    precision: u8,
+    scale: i8,
+    fail_on_error: bool,
+) -> DataFusionResult<Option<i128>> {
     match v {
```
comphead: the block can be rewritten more concise

```
v.map_or(Ok(None), |v| {
    let v = v as i128;
    match validate_decimal_precision(v, precision, scale) {
        Ok(()) => Ok(Some(v)),
        Err(_) if fail_on_error => Err(DataFusionError::External(Box::new(
            decimal_overflow_error(v, precision, scale),
        ))),
        Err(_) => Ok(None),
    }
})
```

comphead (approved):
Thanks @andygrove

PR #5081, opened 2026-07-28
fix: raise REMAINDER_BY_ZERO for Float/Double under ANSI mode

### LINKED ISSUE #5067: Remainder on Float/Double returns NaN for zero divisor under ANSI mode instead of throwing
### Describe the bug

Under ANSI mode, `Remainder` on Float/Double operands returns `NaN` for a zero divisor instead of throwing. Spark 4.1 throws `REMAINDER_BY_ZERO` (`DIVIDE_BY_ZERO` on earlier versions) for all numeric types including floating point.

Comet's native implementation (`native/spark-expr/src/math_funcs/modulo_expr.rs`) relies on Arrow's `rem` kernel to signal the error: it catches `ArrowError::DivideByZero` and re-raises `remainder_by_zero_error()`. But arrow-arith only produces `DivideByZero` for integer and decimal types. For floats, `Op::Rem` uses `mod_wrapping`, which returns `NaN` for `x % 0.0` and never errors, so the error branch is unreachable for Float32/Float64.

The non-ANSI path is correct: `null_if_zero_primitive` nulls zero divisors for floats before the kernel runs. Integral and decimal ANSI remainder are also correct.

### Steps to reproduce

```sql
SELECT CAST(1.0 AS DOUBLE) % CAST(0.0 AS DOUBLE);
```

Spark: throws `SparkArithmeticException` with error class `REMAINDER_BY_ZERO`.
Comet: returns `NaN`.

### Expected behavior

Under ANSI mode, Comet should detect zero divisors for Float32/Float64 explicitly (the Arrow kernel will not do it) and raise the remainder-by-zero error.

### Additional context

Found by an ANSI-mode audit of all native expressions against Spark 4.1.1. Related: #2855 tracks using the `Remainder by zero` wording (vs `Divide by zero`) for Spark 4.1+; this issue is about the float path raising no error at all.

andygrove:

Closes #5067.

## Rationale for this change

Under ANSI mode, `Remainder` on Float/Double operands returned `NaN` for a zero divisor instead of throwing, because Comet's native modulo implementation relies on Arrow's `rem` kernel to signal the error. `arrow-arith` only produces `DivideByZero` for integer and decimal types; for floats `Op::Rem` uses `mod_wrapping`, which returns `NaN` for `x % 0.0` and never errors, so the existing error branch was unreachable for Float32/Float64.

Spark 4.1 raises `REMAINDER_BY_ZERO` (and earlier Spark versions raise `DIVIDE_BY_ZERO`) for all numeric types including floating point, so this was a native correctness gap.

Under ANSI mode, `spark_modulo` no longer routes Float32/Float64 through the `rem` kernel. It computes the remainder with `arrow::compute::kernels::arity::try_binary` and `ArrowNativeTypeOp::mod_checked` instead, which is the same route `checked_arithmetic.rs` already takes for the equivalent float divide-by-zero problem. This means the Spark-compliant semantics fall out of arrow-rs rather than being reimplemented here:

- `mod_checked` returns `DivideByZero` when the divisor `is_zero()`, so a zero divisor raises instead of yielding `NaN`.
- `-0.0` raises too, since `is_zero()` for floats is `self == 0.0` — matching Spark, which also treats `-0.0` as a zero divisor.
- `try_binary` only invokes the closure on rows valid in both operands, so rows where either operand is null produce a null result without raising, matching Spark's row-wise semantics.

Details worth noting for review:

- Only `ArrowError::DivideByZero` maps to `remainder_by_zero_error()`; other errors pass through, so a length mismatch cannot surface as `REMAINDER_BY_ZERO`.
- Scalar/scalar operands still evaluate to a `ColumnarValue::Scalar`, preserving the shape DataFusion's `apply` produced.
- The branch is guarded on both operand types being equal. Spark casts `Remainder` operands to a common type so this always holds in practice, but it prevents a panic in `as_primitive` if it ever does not.

Non-ANSI mode is unchanged: `create_modulo_expr` still rewrites the divisor so zeros become `NULL`, and `rem`/`mod_wrapping` handles the arithmetic.

Rust unit tests in `modulo_expr` (19 total, 15 covering floats) exercise Float32/Float64 in ANSI and non-ANSI mode across all four operand shapes — `Array/Array`, `Array/Scalar`, `Scalar/Array`, and `Scalar/Scalar` — including zero and `-0.0` divisors, all-null dividends, mixed null-and-zero-divisor rows, and non-zero divisors where results (not just errors) are checked.

SQL-file coverage in `arithmetic_ansi.sql` adds four float/double `% 0` cases that keep a column operand so Spark's `ConstantFolding` rule cannot evaluate them during optimization, plus a non-zero-divisor case asserting results. I verified these actually reach the native path rather than assuming it: with the new float branch disabled *and* the pre-existing column-vs-column cases removed, the file still fails with `Expected Comet to throw an error matching 'BY_ZERO' but query succeeded`. The three fully-literal cases are retained but now carry a comment noting they are folded before reaching Comet, so they assert Spark's own behavior rather than this PR's code.

Also green locally: `cargo test -p datafusion-comet-spark-expr` (553 tests), `cargo clippy -p datafusion-comet-spark-expr --all-targets -- -D warnings`, `cargo fmt --check`, and the existing `remainder function` test in `CometExpressionSuite`.

### DIFF
--- a/native/spark-expr/src/math_funcs/modulo_expr.rs
+++ b/native/spark-expr/src/math_funcs/modulo_expr.rs
@@ -344,11 +344,8 @@ mod tests {
                         if !actual_arr.is_null(i) {
                             let actual_value = actual_arr.value(i);
                             let expected_value = expected_arr.value(i);
-                            // `is_eq` uses arrow's total ordering, under which `NaN` equals
-                            // `NaN`. Plain `==` would fail for the NaN results that a
-                            // non-zero divisor produces from NaN or infinite dividends.
-                            assert!(
-                                actual_value.is_eq(expected_value),
+                            assert_eq!(
+                                actual_value, expected_value,
                                 "Mismatch at index {i}, actual {actual_value:?}, expected {expected_value:?}"
                             );
                         }
@@ -843,12 +840,41 @@ mod tests {
         }
     }
 
+    /// Evaluates `a % b` over two Float64 columns and returns the result array.
+    ///
+    /// Used by tests whose expected result is NaN. `verify_result` compares with `==`, under
+    /// which `NaN != NaN`, and arrow's `is_eq`/`total_cmp` is no better here because the NaN
+    /// produced by an invalid `fmod` has a platform-dependent sign — x86 and aarch64 disagree
+    /// on whether it is `+NaN` or `-NaN`. These tests assert NaN-ness instead.
+    fn eval_float64_modulo(lhs: ArrayRef, rhs: ArrayRef, fail_on_error: bool) -> Float64Array {
+        let schema = Arc::new(Schema::new(vec![
+            Field::new("a", DataType::Float64, true),
+            Field::new("b", DataType::Float64, true),
+        ]));
+        let batch = RecordBatch::try_new(Arc::clone(&schema), vec![lhs, rhs]).unwrap();
+
+        let session_ctx = SessionContext::new();
+        let modulo_expr = create_modulo_expr(
+            Arc::new(Column::new("a", 0)),
+            Arc::new(Column::new("b", 1)),
+            DataType::Float64,
+            schema,
+            fail_on_error,
+            &session_ctx.state(),
+        )
+        .unwrap();
+
+        let ColumnarValue::Array(result) = modulo_expr.evaluate(&batch).unwrap() else {
+            panic!("Expected an array result");
+        };
+        result.as_primitive::<Float64Type>().clone()
+    }
+
     #[test]
     fn test_modulo_special_dividends_non_zero_divisor_float64_ansi() {
-        // The same dividends must not raise when the divisor is non-zero. `x % 2.0` is
-        // NaN for NaN and for both infinities, and preserves the sign of a zero dividend.
-        run_float_modulo::<Float64Type>(
-            DataType::Float64,
+        // The same dividends must not raise when the divisor is non-zero: `x % 2.0` is NaN
+        // for NaN and for both infinities, 0.0 for a zero dividend, and 1.0 for 5.0.
+        let result = eval_float64_modulo(
             Arc::new(Float64Array::from(vec![
                 Some(f64::NAN),
                 Some(f64::INFINITY),
@@ -858,15 +884,18 @@ mod tests {
             ])),
             Arc::new(Float64Array::from(vec![Some(2.0); 5])),
             true,
-            false,
-            Some(Arc::new(Float64Array::from(vec![
-                Some(f64::NAN),
-                Some(f64::NAN),
-                Some(f64::NAN),
-                Some(0.0),
-                Some(1.0),
-            ]))),
         );
+
+        assert_eq!(result.null_count(), 0);
+        for i in 0..3 {
+            assert!(
+                result.value(i).is_nan(),
+                "Expected NaN at index {i}, got {}",
+                result.value(i)
+            );
+        }
+        assert_eq!(result.value(3), 0.0);
+        assert_eq!(result.value(4), 1.0);
     }
 
     #[test]
@@ -941,13 +970,16 @@ mod tests {
     #[test]
     fn test_modulo_nan_divisor_float64_ansi() {
         // A NaN divisor is not zero either, so `x % NaN` is NaN rather than an error.
-        run_float_modulo::<Float64Type>(
-            DataType::Float64,
+        let result = eval_float64_modulo(
             Arc::new(Float64Array::from(vec![Some(1.0)])),
             Arc::new(Float64Array::from(vec![Some(f64::NAN)])),
             true,
-            false,
-            Some(Arc::new(Float64Array::from(vec![Some(f64::NAN)]))),
+        );
+        assert_eq!(result.null_count(), 0);
+        assert!(
+            result.value(0).is_nan(),
+            "Expected NaN, got {}",
+            result.value(0)
         );
     }
 }
--- a/native/spark-expr/src/math_funcs/modulo_expr.rs
+++ b/native/spark-expr/src/math_funcs/modulo_expr.rs
@@ -17,9 +17,11 @@
 
 use crate::{create_comet_physical_fun, IfExpr};
 use crate::{remainder_by_zero_error, Cast, EvalMode, SparkCastOptions};
-use arrow::array::{Array, ArrayRef, AsArray};
+use arrow::array::{ArrayRef, ArrowNativeTypeOp, AsArray, PrimitiveArray};
+use arrow::compute::kernels::arity::try_binary;
 use arrow::compute::kernels::numeric::rem;
 use arrow::datatypes::*;
+use arrow::error::ArrowError;
 use datafusion::common::{exec_err, internal_err, DataFusionError, Result, ScalarValue};
 use datafusion::config::ConfigOptions;
 use datafusion::execution::FunctionRegistry;
@@ -54,11 +56,15 @@ pub fn spark_modulo(args: &[ColumnarValue], fail_on_error: bool) -> Result<Colum
         return apply_cmp_for_nested(Operator::Modulo, lhs, rhs);
     }
 
-    // Arrow's `rem` kernel only signals `DivideByZero` for integer and decimal types.
-    // For floating point, `x % 0.0` yields NaN silently, so under ANSI mode we must
-    // detect zero divisors explicitly to raise the Spark-compliant error.
-    if fail_on_error && matches!(right_data_type, DataType::Float32 | DataType::Float64) {
-        check_float_remainder_by_zero(lhs, rhs)?;
+    // Arrow's `rem` kernel only signals `DivideByZero` for integer and decimal types. For
+    // floating point it uses `mod_wrapping`, so `x % 0.0` silently yields NaN. Under ANSI
+    // mode we compute the float remainder with `mod_checked` instead, which reports
+    // `DivideByZero` for a zero divisor.
+    if fail_on_error
+        && left_data_type == right_data_type
+        && matches!(right_data_type, DataType::Float32 | DataType::Float64)
+    {
+        return checked_float_modulo(lhs, rhs, &right_data_type);
     }
 
     match apply(lhs, rhs, rem) {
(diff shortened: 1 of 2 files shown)


native/spark-expr/src/math_funcs/modulo_expr.rs
```diff
@@ -63,6 +71,77 @@ pub fn spark_modulo(args: &[ColumnarValue], fail_on_error: bool) -> Result<Colum
...
+            if mask.is_null(i) {
+                continue;
+            }
+        }
+        if arr.value(i) == zero {
+            return true;
+        }
+    }
+    false
+}
```
mbutrovich: Thanks for tracking down the root cause here. The `mod_wrapping` vs `mod_checked` split in arrow-arith explains exactly why the existing error branch was unreachable for floats.

`check_float_remainder_by_zero` and its helpers (`float_array_has_zero`, `scan_float_array`) reimplement something arrow-rs already provides. `ArrowNativeTypeOp::mod_checked` for `f32`/`f64` already returns `DivideByZero` when the divisor `is_zero()`, and computes `self % rhs` otherwise, so `-0.0` is handled the same way there as it is in this hand-written check.

`checked_arithmetic.rs` already takes this route for the equivalent float-divide-by-zero problem: it calls `arrow::compute::kernels::arity::try_binary` with `div_checked` instead of relying on `arrow::compute::kernels::numeric::div`, because that kernel has the same wrapping-only behavior for floats that `rem()` has here. Would `try_binary` with `mod_checked` work for this fix too? Since `mod_checked` for floats can only fail with `DivideByZero` (there is no overflow case for remainder), any `Err` from that call maps directly to `remainder_by_zero_error()`. `try_binary` also only invokes the closure on rows where both operands are valid, so the null-dividend skip logic in `float_array_has_zero` and `scan_float_array` would not be needed. This would also remove the extra full pass over the divisor array that the current code does before `rem()` runs its own pass, on every ANSI-mode float remainder call, including the common case with no zero divisor.

Rough sketch of what this could look like in place of `check_float_remainder_by_zero`, `float_array_has_zero`, and `scan_float_array` (not compiled, just to make the suggestion concrete):

```rust
if fail_on_error && matches!(right_data_type, DataType::Float32 | DataType::Float64) {
    return checked_float_modulo(lhs, rhs, &right_data_type);
}

match apply(lhs, rhs, rem) {
    Ok(result) => Ok(result),
    Err(e) if e.to_string().contains("Divide by zero") && fail_on_error => {
        Err(remainder_by_zero_error().into())
    }
    Err(e) => Err(e),
}
```

```rust
fn checked_float_modulo(
    lhs: &ColumnarValue,
    rhs: &ColumnarValue,
    data_type: &DataType,
) -> Result<ColumnarValue> {
    let len = match (lhs, rhs) {
        (ColumnarValue::Array(a), _) | (_, ColumnarValue::Array(a)) => a.len(),
        _ => 1,
    };
    let l_arr = match lhs {
        ColumnarValue::Array(a) => Arc::clone(a),
        ColumnarValue::Scalar(s) => s.to_array_of_size(len)?,
    };
    let r_arr = match rhs {
        ColumnarValue::Array(a) => Arc::clone(a),
        ColumnarValue::Scalar(s) => s.to_array_of_size(len)?,
    };

    let result: ArrayRef = match data_type {
        DataType::Float32 => Arc::new(
            try_binary::<_, _, _, Float32Type>(
                l_arr.as_primitive::<Float32Type>(),
                r_arr.as_primitive::<Float32Type>(),
                |l, r| l.mod_checked(r),
            )
            .map_err(|_| remainder_by_zero_error())?,
        ),
        DataType::Float64 => Arc::new(
            try_binary::<_, _, _, Float64Type>(
                l_arr.as_primitive::<Float64Type>(),
                r_arr.as_primitive::<Float64Type>(),
                |l, r| l.mod_checked(r),
            )
            .map_err(|_| remainder_by_zero_error())?,
        ),
        _ => unreachable!(),
    };
    Ok(ColumnarValue::Array(result))
}
```

Needs `use arrow::array::{ArrowNativeTypeOp, AsArray}` and `use arrow::compute::kernels::arity::try_binary`, both already imported in `checked_arithmetic.rs` for the same purpose.

native/spark-expr/src/math_funcs/modulo_expr.rs
```diff
@@ -441,4 +520,126 @@ mod tests {
             verify_result(modulo_expr, batch, fail_on_error, Some(expected_result));
         })
     }
+
+    fn run_float_modulo<T: ArrowPrimitiveType>(
```
mbutrovich, line 516: All six new float tests go through `run_float_modulo`, which builds both operands as `Column` expressions, so `spark_modulo` always sees `(ColumnarValue::Array, ColumnarValue::Array)`. The `Scalar/Scalar`, `Array/Scalar`, and `Scalar/Array` match arms in `check_float_remainder_by_zero` are not exercised by any of these tests.

Could you add a case that passes at least one operand as a `Literal`, the way `test_division_by_zero_with_complex_int_expr` does with `Literal::new(ScalarValue::Int32(Some(0)))`?

spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
```diff
@@ -147,6 +147,37 @@ SELECT c div d FROM ansi_div_zero
...
+query expect_error(BY_ZERO)
+SELECT a % b FROM ansi_float_div_zero WHERE b = 0.0
+
+-- double column % 0 should throw
+query expect_error(BY_ZERO)
+SELECT c % d FROM ansi_float_div_zero WHERE d = 0.0
+
+-- literal double % 0.0 should throw
+query expect_error(BY_ZERO)
+SELECT CAST(1.0 AS DOUBLE) % CAST(0.0 AS DOUBLE)
```
mbutrovich, line 197: This expression is fully literal, so Spark's own `ConstantFolding` rule evaluates and throws on it during optimization, before the query reaches Comet's native `spark_modulo`. The same is true for the `-0.0` and float cases on lines 175 and 179. This mirrors an existing pattern already in this file (`SELECT 1 % 0` above), so it is not a new issue, but it means these three cases do not exercise this PR's new code.

Could you add a case like `SELECT a % 0.0 FROM ansi_float_div_zero` (column dividend, literal zero divisor)? That query is not foldable, since `a` is a column, so it would actually reach Comet's native path at execution time and cover the `Array/Scalar` branch that the Rust unit tests also miss.

mbutrovich (commented):

native/spark-expr/src/math_funcs/modulo_expr.rs
andygrove: Good call — this is much better. Done in 6c815fea.

I confirmed both halves of the suggestion against the pinned arrow version before switching:

- arrow-array 59.1.0 `src/arithmetic.rs:353` — float `mod_checked` is `if rhs.is_zero() { Err(DivideByZero) } else { Ok(self % rhs) }`, and `is_zero` for floats is `self == $zero`, so `-0.0` errors just like the hand-written check did.
- arrow-arith 59.1.0 `src/arity.rs:283` — when either input has nulls, `try_binary` runs the closure under `nulls.try_for_each_valid_idx`, so it only touches rows valid in both operands. That makes the null-dividend skip logic unnecessary, as you said.

So `check_float_remainder_by_zero`, `float_array_has_zero`, and `scan_float_array` are all gone (~70 lines), along with the extra pass over the divisor array that ran ahead of `rem` on every ANSI-mode float remainder.

Two deviations from your sketch:

1. The sketch always returns `ColumnarValue::Array`. DataFusion's `apply` returns a `Scalar` for scalar/scalar operands (it calls `ScalarValue::try_from_array(array, 0)` in that arm), and returning a length-1 array when the batch has N rows would be wrong, so I convert back to a scalar in that case. `test_modulo_both_literals_non_zero_divisor_float64_ansi` pins that down.

2. I added a `left_data_type == right_data_type` guard on the branch. Spark casts `Remainder` operands to a common type so this should always hold, but without it a mismatched pair would panic in `as_primitive::<Float64Type>()`; with it, the pair falls through to `rem`, which returns an error rather than silently producing NaN.

I also matched specifically on `ArrowError::DivideByZero` rather than mapping every error to `remainder_by_zero_error()`, mirroring `checked_binary` in `checked_arithmetic.rs`, so a length mismatch can't surface as `REMAINDER_BY_ZERO`.

native/spark-expr/src/math_funcs/modulo_expr.rs
andygrove: You're right — every one of those went through `Column` expressions, so only the `Array/Array` arm was exercised. Added coverage for the rest in 6c815fea using `Literal`, as suggested:

`Array/Scalar` (column dividend, literal divisor):
- `test_modulo_literal_zero_divisor_float64_ansi`
- `test_modulo_literal_negative_zero_divisor_float64_ansi`
- `test_modulo_all_null_lhs_literal_zero_divisor_float64_ansi` — all dividends null, so no row pairs a non-null dividend with the zero divisor and nothing should raise
- `test_modulo_literal_non_zero_divisor_float64` — runs in both ANSI and non-ANSI mode and checks results, not just the error

`Scalar/Array` (literal dividend, column divisor):
- `test_modulo_literal_dividend_zero_divisor_float64_ansi`
- `test_modulo_null_literal_dividend_zero_divisor_float64_ansi`

`Scalar/Scalar`:
- `test_modulo_both_literals_zero_divisor_float64_ansi`
- `test_modulo_both_literals_non_zero_divisor_float64_ansi` — asserts the result is a `Scalar`, not a one-row array, which is the shape `apply` produced before this change

19 tests in `modulo_expr` now.

spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
andygrove: Agreed. Added non-foldable cases in 6c815fea, keeping a column operand so `ConstantFolding` can't evaluate them:

```sql
SELECT a % CAST(0.0 AS FLOAT) FROM ansi_float_div_zero
SELECT c % CAST(0.0 AS DOUBLE) FROM ansi_float_div_zero
SELECT c % CAST(-0.0 AS DOUBLE) FROM ansi_float_div_zero
SELECT CAST(1.0 AS DOUBLE) % d FROM ansi_float_div_zero
SELECT c % CAST(2.0 AS DOUBLE) FROM ansi_float_div_zero ORDER BY 1
```

The last one is a plain `query` rather than `expect_error`, so a non-zero literal divisor still has to produce correct results and not throw. I kept the three fully-literal cases, since they still assert Spark's own behavior, but added a comment above them noting they're folded before reaching Comet so nobody reads them as native coverage.

Rather than assume these reach the native path, I checked: I temporarily disabled the new float branch in `spark_modulo` *and* removed the two pre-existing `a % b` / `c % d` column-vs-column cases, then reran the file. It still failed with `cometError.isDefined was false Expected Comet to throw an error matching 'BY_ZERO' but query succeeded`, so the new cases are doing the work on their own. Restored both and the file passes.

One note for anyone reproducing locally: my Maven run resolved to the Spark 4.1 / Scala 2.13 profile, which needs JDK 17 — JDK 11 fails to compile `ShimCometSparkSessionExtensions.scala` with `Class java.lang.Record not found`. That profile is the one where `REMAINDER_BY_ZERO` actually applies, so it's the useful one to run here.

comphead (commented):
Thanks @andygrove some medium severity issues reported

```
1. [MEDIUM][performance] Pre-scan uses branchy scalar loop that blocks autovectorization
  
  Location: native/spark-expr/src/math_funcs/modulo_expr.rs (proposed scan_float_array<T>)
  Description: The ANSI float/double pre-scan uses a scalar for i in 0..arr.len() with two NullBuffer::is_null(i) bitmap lookups and an early-exit return true per row. The data-dependent control flow blocks
  LLVM autovectorization, doing serial O(N) work comparable to the SIMD rem kernel it guards, roughly doubling ANSI float-modulo cost even when no zero divisor is present.
  Reason: Arrow float Op::Rem is infallible via mod_wrapping (arrow-arith/src/numeric.rs:361-368), so a pre-scan is unavoidable, but a branch-free path using arrow::compute::kernels::cmp::eq(divisor, zero)
  intersected with the dividend validity bitmap plus BooleanBuffer::count_set_bits > 0 would preserve SIMD throughput. Estimated cost is unmeasured.
  Failure scenario: SELECT a % b FROM t on Double columns with no zero divisors pays a per-batch scan cost proportional to a full column pass on every batch.

  2. [MEDIUM][test-gap] Missing ANSI tests for NaN/Inf/0.0 dividend paired with 0.0 divisor

  Location: spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
  Description: Added SQL cases cover literal x % 0.0, x % -0.0, and column % column WHERE b=0.0. NaN % 0.0, +Inf % 0.0, -Inf % 0.0, and 0.0 % 0.0 are not exercised, and Spark v4.1.3 DivModLike.eval throws
  regardless of dividend value when the divisor is zero.
  Reason: A future regression that gates zero-divisor errors on dividend finiteness (e.g. dividend.is_finite()) would slip through undetected. The project's SQL coverage rule explicitly requires FP NaN/±0/±Inf
  boundary coverage.
  Failure scenario: Comet returns NaN for NaN % 0.0 in ANSI mode instead of throwing REMAINDER_BY_ZERO. Undetected by current tests.
  Missing test example:
  query expect_error(BY_ZERO)
  SELECT double('NaN') % 0.0
  query expect_error(BY_ZERO)
  SELECT double('Infinity') % 0.0
  query expect_error(BY_ZERO) 
  SELECT double('-Infinity') % 0.0
  query expect_error(BY_ZERO)
  SELECT 0.0d % 0.0d
  
  3. [MEDIUM][test-gap] SQL column-column mixed-row batch scenario missing

  Location: spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
  Description: The added column % column case is filtered by WHERE b=0.0, so it never exercises a genuinely mixed-row batch. The (Array, Array) branch of check_float_remainder_by_zero needs a batch where some
  Location: spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
  Description: The added column % column case is filtered by WHERE b=0.0, so it never exercises a genuinely mixed-row batch. The (Array, Array) branch of check_float_remainder_by_zero needs a batch where some
  rows have zero divisors, some do not, and some have null dividends.
  Reason: Comet emits one exception per batch, so only a mixed-row test verifies the alignment logic between dividend nullness and divisor zeroness.
  Failure scenario: A batch-level check that raises whenever any(b == 0.0) without checking dividend nullability would fail an all-null-dividend batch.
  Missing test example:
  statement
  INSERT INTO ansi_float_mixed VALUES (1.0, 2.0), (3.0, 0.0), (NULL, 0.0), (5.0, 1.5)
  query expect_error(BY_ZERO)
  SELECT a % b FROM ansi_float_mixed
  query
  SELECT a % b FROM ansi_float_mixed WHERE a IS NULL
```

native/spark-expr/src/math_funcs/modulo_expr.rs
```diff
@@ -63,6 +77,61 @@ pub fn spark_modulo(args: &[ColumnarValue], fail_on_error: bool) -> Result<Colum
...
+        (ColumnarValue::Array(l), ColumnarValue::Scalar(r)) => {
+            (Arc::clone(l), r.to_array_of_size(l.len())?)
+        }
+        (ColumnarValue::Scalar(l), ColumnarValue::Scalar(r)) => (l.to_array()?, r.to_array()?),
+    };
+
+    let result: ArrayRef = match data_type {
+        DataType::Float32 => Arc::new(checked_modulo_kernel::<Float32Type>(&left, &right)?),
+        DataType::Float64 => Arc::new(checked_modulo_kernel::<Float64Type>(&left, &right)?),
+        _ => return internal_err!("checked_float_modulo called with {data_type}"),
```
comphead:
```suggestion
        _ => return not_impl_err!("checked_float_modulo doesn't support {data_type}"),
```

andygrove: Applied in 63f174b7.

andygrove:
Thanks @comphead. Items 2 and 3 were real gaps and are fixed in 63f174b7. Item 1 was written against the previous revision — details below.

### 1. Branchy scalar pre-scan — no longer present, but there is a related finding

`scan_float_array` and the pre-scan it belonged to were removed in 6c815fea (addressing @mbutrovich's earlier comment), before this review landed. The current code has no pre-scan at all: it computes the remainder with `try_binary` + `ArrowNativeTypeOp::mod_checked`, so the zero check rides along inside the single pass that produces the result.

Your underlying concern is still worth measuring though, so I benchmarked all three strategies on 8192 `f64` rows with **no** zero divisors (the common case you called out), against the infallible `rem` kernel as baseline:

| strategy | no nulls | 1-in-17 nulls |
|---|---|---|
| `rem` (non-ANSI baseline) | 63.1 µs | 57.8 µs |
| `try_binary` + `mod_checked` (current) | 65.1 µs (+3%) | 72.1 µs (**+25%**) |
| `eq` pre-scan + `rem` (your suggestion) | 62.6 µs (~0%) | 60.2 µs (+4%) |

Two things fall out of this:

- The absolute cost is ~7.7 ns/row, i.e. ~25 cycles. Float remainder is `fmod`-bound, not throughput-bound — `mod_wrapping` compiles to a libm call per element, so it never vectorized in the first place. That's why the branch in `mod_checked` costs only 3% when there are no nulls, rather than the ~2x a blocked-autovectorization argument would predict.
- **You're right about the null case.** When either operand has nulls, `try_binary` takes its `try_for_each_valid_idx` path (arrow-arith `arity.rs:283`), which walks valid indices through the bitmap one at a time instead of computing straight through and applying the null buffer afterwards. That's the +25%.

So there's a genuine ~25% regression on ANSI float modulo over nullable columns, and your `eq` + `count_set_bits` suggestion does avoid it. I have not switched to it, because it directly reverses the change @mbutrovich asked for and re-adds the dividend-validity intersection logic that removal simplified away — I would rather you two agree on the tradeoff than have me flip-flop the implementation between reviews. Happy to make the switch immediately if you want the null-case regression closed; the measurement above is the whole argument for it.

(I did not commit the benchmark — it was a throwaway. Glad to add a permanent `modulo` bench if that would be useful, since there isn't one today.)

### 2. NaN / Inf / 0.0 dividend with a zero divisor — fixed

Good catch, and the reasoning checks out in Spark 4.1's `DivModLike.eval` (`arithmetic.scala:636`):

```scala
val input2 = right.eval(input)
if (input2 == null || (!failOnError && isZero(input2))) null
else {
  val input1 = left.eval(input)
  if (input1 == null) null
  else { if (isZero(input2)) throw QueryExecutionErrors.divideByZeroError(...) ... }
}
```

`isZero` only ever looks at the divisor, so any non-null dividend — NaN, ±Infinity, ±0.0 — throws. Added, with one adjustment to your examples: `SELECT double('NaN') % 0.0` is fully literal and gets constant-folded before it reaches Comet, so I kept the dividend as a column:

```sql
SELECT a % CAST(0.0 AS DOUBLE) FROM ansi_float_special WHERE isnan(a)
SELECT a % CAST(0.0 AS DOUBLE) FROM ansi_float_special WHERE a = double('Infinity')
SELECT a % CAST(0.0 AS DOUBLE) FROM ansi_float_special WHERE a = double('-Infinity')
SELECT a % CAST(0.0 AS DOUBLE) FROM ansi_float_special WHERE a = 0.0
SELECT a % CAST(2.0 AS DOUBLE) FROM ansi_float_special ORDER BY 1   -- must not throw
```

Matching Rust tests cover the same dividends for both `Array/Array` and `Array/Scalar` operands, and I added null-divisor and NaN-divisor cases confirming neither is mistaken for zero.

### 3. Mixed-row batch — fixed

Agreed, and I used your suggested shape almost verbatim:

```sql
INSERT INTO ansi_float_mixed VALUES (1.0, 2.0), (3.0, 0.0), (NULL, 0.0), (5.0, 1.5), (NULL, 4.0)

SELECT a % b FROM ansi_float_mixed                          -- throws, because of (3.0, 0.0)
SELECT a % b FROM ansi_float_mixed WHERE a IS NULL ORDER BY 1 -- null, must NOT throw
SELECT a % b FROM ansi_float_mixed WHERE b <> 0.0 ORDER BY 1  -- no zero divisors at all
```

The middle one is the important negative case: per the `eval` snippet above, a null dividend returns null even under ANSI mode with a zero divisor, so a batch-wide `any(b == 0.0)` check would incorrectly raise there.

### Verifying these actually reach the native path

Rather than assume, I isolated each new SQL error case into its own file and reran with the float branch disabled. All four error cases failed independently (so each one genuinely exercises the fix), and the `WHERE a IS NULL` case passed with the branch disabled, which is correct — it's the negative control and null propagation handles it either way. Re-enabled, all five pass.

Also green: 25 tests in `modulo_expr`, `cargo test -p datafusion-comet-spark-expr` (559), clippy `-D warnings`, `cargo fmt --check`.

comphead (approved):
Thanks @andygrove it LGTM

PR #5082, opened 2026-07-28
fix: round on Int64 with scale <= -19 now overflows correctly

### LINKED ISSUE #5070: round on LongType with scale <= -19 returns 0 instead of overflowing
### Describe the bug

`round` on a `LongType` column with scale <= -19 returns 0 for every row, in both ANSI and legacy mode.

Comet's `CometRound` serde correctly serializes `ansiEnabled` (`spark/src/main/scala/org/apache/comet/serde/arithmetic.scala:415-423`) and the native kernel honors it for the normal path (`integer_round!` uses checked vs wrapping arithmetic). But when `10^(-scale)` overflows the native integer type, `native/spark-expr/src/math_funcs/round.rs:66-72` and `:79-96` return `Ok(0)` for the whole batch, ignoring `fail_on_error`.

For `Int64` with `scale <= -19`, `10^19 > i64::MAX`, so this path is reachable. Spark computes `BigDecimal(x).setScale(-19, HALF_UP)` and then `toLongExact`: for `|x| >= 5e18` the rounded value is `1e19`, which does not fit a long, so ANSI throws `ARITHMETIC_OVERFLOW` and legacy mode wraps. Comet silently returns 0 in both modes.

Int8/Int16/Int32 are unaffected: their maximum values always round to 0 at the overflowing scale.

### Steps to reproduce

```sql
SELECT round(CAST(5000000000000000000 AS LONG), -19);
```

Spark: throws `ARITHMETIC_OVERFLOW` (legacy mode: wrapped value).
Comet: returns 0.

### Expected behavior

The overflow-of-`10^(-scale)` path should replicate Spark: throw under ANSI, wrap under legacy, instead of returning 0.

### Additional context

Found by an ANSI-mode audit of all native expressions against Spark 4.1.1. There is no ANSI SQL file test for `round` (`spark/src/test/resources/sql-tests/expressions/math/round.sql` has no ANSI or negative-scale overflow case).

andygrove:

Closes #5070.

## Rationale for this change

`spark_round`'s integer path returned `Ok(0)` for every row whenever `10^(-scale)` did not fit in the native integer type (`native/spark-expr/src/math_funcs/round.rs` lines 66-72 and 79-96 in the pre-fix code). For `Int64` with `scale == -19` this masked a real overflow: values with `|x| >= 5e18` round to `sign(x) * 10^19`, which does not fit in a `long`. Spark throws `ARITHMETIC_OVERFLOW` under ANSI mode and wraps (low-order 64 bits) under legacy mode; Comet silently returned `0` in both modes.

`Int8`/`Int16`/`Int32` were not affected because their maxima are strictly less than half the overflowing divisor, so `0` is the correct answer at every overflowing scale.

- In `round_integer_array!` and `round_integer_scalar!`, when `10^(-scale)` overflows the native integer type but still fits in `i128`, widen the intermediate rounding computation to `i128` and check the result against the native type's range: fail under ANSI, truncate under legacy (matches `BigDecimal.longValue`'s low-order-64-bit semantics).
- When `10^(-scale)` also overflows `i128` (`-scale > 38`), retain the fast path that returns `0` — every bounded native integer rounds to `0`.
- Add Rust unit tests covering:
  - `Int64` array + scale `-19`, ANSI mode → throws `ARITHMETIC_OVERFLOW`.
  - `Int64` array + scale `-19`, legacy mode → wraps to the two's-complement low 64 bits (`±10^19 as i64`), including `i64::MAX`, `i64::MIN`, `±5e18`, and values below the rounding threshold.
  - `Int64` array + scale `-20`, both modes → always `0`.
  - `Int64` scalar equivalents for ANSI throw and legacy wrap.

## Are these changes tested?

Yes: six new tests in `native/spark-expr/src/math_funcs/round.rs`, run via `cargo test -p datafusion-comet-spark-expr --lib math_funcs::round::`.

Yes. `round(cast(x as long), s)` where `s <= -19` and `|x| >= 5 * 10^(-s-1)`:
- ANSI mode: now throws `ARITHMETIC_OVERFLOW` (previously returned `0` silently).
- Legacy mode: now returns the wrapped low-64-bit value (previously returned `0`).
Both match Apache Spark 3.4–4.1.

### DIFF
--- a/native/spark-expr/src/math_funcs/round.rs
+++ b/native/spark-expr/src/math_funcs/round.rs
@@ -60,29 +60,24 @@ macro_rules! integer_round {
 }
 
 // Round a single native integer when `10^(-point)` does not fit in the native
-// integer type but still fits in i128. `|x|` is strictly less than `div` here
-// (otherwise the caller would have taken the in-native fast path), so the
-// only possible results are `0`, `+div`, or `-div`. The latter two overflow
-// the native type: under ANSI we throw, under legacy we wrap by truncation
-// (equivalent to `BigDecimal.longValue`'s low-64-bit semantics).
+// integer type but still fits in i128. The caller has already excluded the
+// case where `div` fits the native type, so `|x| < div` and the result is
+// either `0` (when `|x| < half`) or `sign(x) * div` — the latter always
+// overflows the native type. Under ANSI we throw, under legacy we wrap by
+// truncation (matches `BigDecimal.longValue`'s low-64-bit semantics).
 macro_rules! integer_round_widened {
     ($X:expr, $DIV:expr, $HALF:expr, $NATIVE:ty, $FAIL_ON_ERROR:expr) => {{
         let x128 = $X as i128;
-        let rounded: i128 = if x128 <= -$HALF {
-            -$DIV
-        } else if x128 >= $HALF {
-            $DIV
-        } else {
-            0
-        };
-        if rounded >= <$NATIVE>::MIN as i128 && rounded <= <$NATIVE>::MAX as i128 {
-            Ok(rounded as $NATIVE)
+        if x128 > -$HALF && x128 < $HALF {
+            Ok(0 as $NATIVE)
         } else if $FAIL_ON_ERROR {
             Err(ArrowError::ComputeError(
                 arithmetic_overflow_error("integer").to_string(),
             ))
+        } else if x128 >= $HALF {
+            Ok($DIV as $NATIVE)
         } else {
-            Ok(rounded as $NATIVE)
+            Ok((-$DIV) as $NATIVE)
         }
     }};
 }
@@ -336,15 +331,12 @@ mod test {
     // values with |x| >= 5e18 round to sign(x)*1e19, which does not fit in a long:
     // Spark throws under ANSI and wraps (low-order 64 bits) under legacy.
 
-    #[test]
-    fn test_round_int64_negative_scale_overflow_ansi() {
-        // 5e18 rounds up to 1e19, which overflows i64.
-        let args = vec![
-            ColumnarValue::Array(Arc::new(Int64Array::from(vec![
-                5_000_000_000_000_000_000i64,
-            ]))),
-            ColumnarValue::Scalar(ScalarValue::Int64(Some(-19))),
-        ];
+    // 1e19 truncated to the low 64 bits, matching `BigDecimal.longValue`.
+    const WRAPPED_POS_1E19: i64 = 10_000_000_000_000_000_000u64 as i64;
+    const WRAPPED_NEG_1E19: i64 = -WRAPPED_POS_1E19;
+
+    fn assert_round_int64_ansi_overflows(value: ColumnarValue) {
+        let args = vec![value, ColumnarValue::Scalar(ScalarValue::Int64(Some(-19)))];
         let err = spark_round(&args, &DataType::Int64, true).unwrap_err();
         assert!(
             err.to_string().to_ascii_lowercase().contains("overflow"),
@@ -353,8 +345,23 @@ mod test {
     }
 
     #[test]
+    fn test_round_int64_negative_scale_overflow_ansi() {
+        // 5e18 rounds up to 1e19, which overflows i64.
+        assert_round_int64_ansi_overflows(ColumnarValue::Array(Arc::new(Int64Array::from(vec![
+            5_000_000_000_000_000_000i64,
+        ]))));
+    }
+
+    #[test]
+    fn test_round_int64_negative_scale_overflow_ansi_scalar() {
+        assert_round_int64_ansi_overflows(ColumnarValue::Scalar(ScalarValue::Int64(Some(
+            5_000_000_000_000_000_000i64,
+        ))));
+    }
+
+    #[test]
     fn test_round_int64_negative_scale_overflow_legacy() -> Result<()> {
-        // Under legacy mode, 1e19 wraps to its low-order 64 bits.
+        // Under legacy mode, ±1e19 wraps to its low-order 64 bits.
         let args = vec![
             ColumnarValue::Array(Arc::new(Int64Array::from(vec![
                 5_000_000_000_000_000_000i64,
@@ -370,14 +377,13 @@ mod test {
             unreachable!()
         };
         let longs = as_int64_array(&result)?;
-        // 1e19 as i64 wraps to -8446744073709551616; -1e19 wraps to +8446744073709551616.
         let expected = Int64Array::from(vec![
-            10_000_000_000_000_000_000u64 as i64,
-            -10_000_000_000_000_000_000i128 as i64,
+            WRAPPED_POS_1E19,
+            WRAPPED_NEG_1E19,
             0i64,
             0i64,
-            10_000_000_000_000_000_000u64 as i64,
-            -10_000_000_000_000_000_000i128 as i64,
+            WRAPPED_POS_1E19,
+            WRAPPED_NEG_1E19,
         ]);
         assert_eq!(longs, &expected);
         Ok(())
@@ -404,19 +410,6 @@ mod test {
     }
 
     #[test]
-    fn test_round_int64_negative_scale_overflow_ansi_scalar() {
-        let args = vec![
-            ColumnarValue::Scalar(ScalarValue::Int64(Some(5_000_000_000_000_000_000i64))),
-            ColumnarValue::Scalar(ScalarValue::Int64(Some(-19))),
-        ];
-        let err = spark_round(&args, &DataType::Int64, true).unwrap_err();
-        assert!(
-            err.to_string().to_ascii_lowercase().contains("overflow"),
-            "expected arithmetic overflow error, got: {err}"
-        );
-    }
-
-    #[test]
     fn test_round_int64_negative_scale_legacy_scalar() -> Result<()> {
         let args = vec![
             ColumnarValue::Scalar(ScalarValue::Int64(Some(5_000_000_000_000_000_000i64))),
@@ -427,7 +420,7 @@ mod test {
         else {
             unreachable!()
         };
-        assert_eq!(result, 10_000_000_000_000_000_000u64 as i64);
+        assert_eq!(result, WRAPPED_POS_1E19);
         Ok(())
     }
 }
--- a/native/spark-expr/src/math_funcs/round.rs
+++ b/native/spark-expr/src/math_funcs/round.rs
@@ -59,6 +59,34 @@ macro_rules! integer_round {
     }};
 }
 
+// Round a single native integer when `10^(-point)` does not fit in the native
+// integer type but still fits in i128. `|x|` is strictly less than `div` here
+// (otherwise the caller would have taken the in-native fast path), so the
+// only possible results are `0`, `+div`, or `-div`. The latter two overflow
+// the native type: under ANSI we throw, under legacy we wrap by truncation
+// (equivalent to `BigDecimal.longValue`'s low-64-bit semantics).
+macro_rules! integer_round_widened {
+    ($X:expr, $DIV:expr, $HALF:expr, $NATIVE:ty, $FAIL_ON_ERROR:expr) => {{
+        let x128 = $X as i128;
+        let rounded: i128 = if x128 <= -$HALF {
+            -$DIV
+        } else if x128 >= $HALF {
+            $DIV
+        } else {
+            0
+        };
+        if rounded >= <$NATIVE>::MIN as i128 && rounded <= <$NATIVE>::MAX as i128 {
+            Ok(rounded as $NATIVE)
+        } else if $FAIL_ON_ERROR {
+            Err(ArrowError::ComputeError(
+                arithmetic_overflow_error("integer").to_string(),
+            ))
+        } else {
+            Ok(rounded as $NATIVE)
+        }
+    }};
+}
+
 macro_rules! round_integer_array {
     ($ARRAY:expr, $POINT:expr, $TYPE:ty, $NATIVE:ty, $FAIL_ON_ERROR:expr) => {{
         let array = $ARRAY.as_any().downcast_ref::<$TYPE>().unwrap();
@@ -68,7 +96,14 @@ macro_rules! round_integer_array {
             arrow::compute::kernels::arity::try_unary(array, |x| {
                 integer_round!(x, div, half, $FAIL_ON_ERROR)
             })?
+        } else if let Some(div) = 10_i128.checked_pow((-(*$POINT)) as u32) {
+            let half = div / 2;
+            arrow::compute::kernels::arity::try_unary(array, |x| {
+                integer_round_widened!(x, div, half, $NATIVE, $FAIL_ON_ERROR)
+            })?
         } else {
+            // Even i128 cannot hold 10^(-point); every bounded native
+            // integer rounds to 0.
             arrow::compute::kernels::arity::try_unary(array, |_| Ok(0))?
         };
         Ok(ColumnarValue::Array(Arc::new(result)))
@@ -82,6 +117,21 @@ macro_rules! round_integer_scalar {
             let half = div / 2;
             let scalar_opt = match $SCALAR {
                 Some(x) => match integer_round!(x, div, half, $FAIL_ON_ERROR) {
+                    Ok(v) => Some(v),
+                    Err(e) => {
+                        return Err(DataFusionError::ArrowError(
+                            Box::from(e),
+                            Some(DataFusionError::get_back_trace()),
+                        ))
+                    }
+                },
+                None => None,
+            };
+            Ok(ColumnarValue::Scalar($TYPE(scalar_opt)))
+        } else if let Some(div) = 10_i128.checked_pow((-(*$POINT)) as u32) {
+            let half = div / 2;
+            let scalar_opt = match $SCALAR {
+                Some(x) => match integer_round_widened!(*x, div, half, $NATIVE, $FAIL_ON_ERROR) {
                     Ok(v) => Some(v),
                     Err(e) => {
                         return Err(DataFusionError::ArrowError(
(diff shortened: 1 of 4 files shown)


native/spark-expr/src/math_funcs/round.rs
```diff
@@ -93,6 +123,21 @@ macro_rules! round_integer_scalar {
...
+                            Some(DataFusionError::get_back_trace()),
+                        ))
+                    }
+                },
+                None => None,
+            };
+            Ok(ColumnarValue::Scalar($TYPE(scalar_opt)))
         } else {
             Ok(ColumnarValue::Scalar($TYPE(Some(0))))
         }
```
mbutrovich: This is the pre-existing catch-all branch of `round_integer_scalar!` for when even `i128` cannot hold `10^(-scale)` (roughly `scale <= -39`). It returns `Ok(ColumnarValue::Scalar($TYPE(Some(0))))` unconditionally, without checking whether `$SCALAR` was `None`. A null `Int64` scalar input at `scale <= -39` comes out as `Some(0)` instead of `None`. This predates the PR, but the PR inserts a new branch directly above it (lines 126-140) that gets this exact case right (`match $SCALAR { Some(x) => ..., None => None }`), so the inconsistency is now more visible, and it's a one-line fix while this exact function is already being reworked. Fix: match on `$SCALAR` here too, e.g. `let scalar_opt = match $SCALAR { Some(_) => Some(0), None => None }; Ok(ColumnarValue::Scalar($TYPE(scalar_opt)))`.

native/spark-expr/src/math_funcs/round.rs
```diff
@@ -59,6 +59,29 @@ macro_rules! integer_round {
...
+            Err(ArrowError::ComputeError(
+                arithmetic_overflow_error("integer").to_string(),
+            ))
+        } else if x128 >= $HALF {
+            Ok($DIV as $NATIVE)
+        } else {
+            Ok((-$DIV) as $NATIVE)
+        }
+    }};
+}
```
mbutrovich, line 86: The final two arms of `integer_round_widened!` (`x128 >= $HALF -> Ok($DIV as $NATIVE)`, else `Ok((-$DIV) as $NATIVE)`) differ only by sign, and both apply only once `$FAIL_ON_ERROR` has already been ruled out above. Fix: collapse to one arm: `Ok((if x128 >= $HALF { $DIV } else { -$DIV }) as $NATIVE)`. Same behavior, one fewer branch.

native/spark-expr/src/math_funcs/round.rs
mbutrovich: round.rs:89, 94, 111, 126

`(-(*$POINT)) as u32` is written out twice in each macro: once for the native `checked_pow` and again for the `i128` `checked_pow`. Fix: compute it once, e.g. `let point_abs = (-(*$POINT)) as u32;` right after `let ten: $NATIVE = 10;`, and reuse it in both `checked_pow` calls in `round_integer_array!` and `round_integer_scalar!`.

native/spark-expr/src/math_funcs/round.rs
```diff
@@ -93,6 +123,21 @@ macro_rules! round_integer_scalar {
                 None => None,
             };
```
mbutrovich: round.rs:113-124, 128-139

The native-fits and widened branches of `round_integer_scalar!` each repeat the identical `match ... { Ok(v) => Some(v), Err(e) => return Err(DataFusionError::ArrowError(Box::from(e), Some(DataFusionError::get_back_trace()))) }` boilerplate, differing only in which inner macro is invoked. Fix: factor this into a small helper macro, e.g. `unwrap_round_result!($RESULT)` expanding to that same match, and call `Some(unwrap_round_result!(integer_round!(...)))` / `Some(unwrap_round_result!(integer_round_widened!(...)))` in each branch.

native/spark-expr/src/math_funcs/round.rs
mbutrovich, line 86: `integer_round_widened!` is a second, separate rounding macro alongside `integer_round!`, correct only because of an unstated invariant: `div` overflowed the native type, so `|x| <= NATIVE::MAX < div`, hence `x % div == x`. That's a narrow algebraic shortcut for this specific overflow band, not the same mechanism `decimal_round_f` uses one function below it (lines 232-247), which always computes in `i128` regardless of how large `div` is and needs no special-cased middle branch. Fix: generalize `integer_round!`'s rem/half formula to operate in `i128` uniformly, with a single checked-or-wrapping downcast to `$NATIVE` at the end, replacing the current three-way dispatch (native-fits / i128-widened / i128-overflows) with one code path. This removes `integer_round_widened!` entirely and matches the pattern already established in this same file for decimals.

native/spark-expr/src/math_funcs/round.rs
```diff
@@ -280,4 +325,102 @@ mod test {
...
     }
+
+    // Regression tests for https://github.com/apache/datafusion-comet/issues/5070:
+    // round(Int64, scale) where `10^(-scale)` does not fit in i64. For scale=-19,
+    // values with |x| >= 5e18 round to sign(x)*1e19, which does not fit in a long:
+    // Spark throws under ANSI and wraps (low-order 64 bits) under legacy.
+
+    // 1e19 truncated to the low 64 bits, matching `BigDecimal.longValue`.
+    const WRAPPED_POS_1E19: i64 = 10_000_000_000_000_000_000u64 as i64;
+    const WRAPPED_NEG_1E19: i64 = -WRAPPED_POS_1E19;
```
mbutrovich: `const WRAPPED_POS_1E19: i64 = 10_000_000_000_000_000_000u64 as i64;` is named as if it holds a positive value, but the truncating cast actually produces a negative number: `1e19 mod 2^64` is less than `2^64` but greater than `i64::MAX`, so reinterpreting those bits as `i64` yields `-8446744073709551616`. The name implies the wrapped result of rounding `+5e18` up to `+1e19` stays positive, which is false and will mislead anyone reading the test without independently working out the two's-complement truncation. Fix: rename to something that does not assert a sign, e.g. `WRAPPED_1E19`, and add a comment stating the wrapped value is negative (`-8446744073709551616`) because `1e19` exceeds `i64::MAX`.

native/spark-expr/src/math_funcs/round.rs
```diff
@@ -280,4 +325,102 @@ mod test {
...
+            5_000_000_000_000_000_000i64,
+        ]))));
+    }
+
+    #[test]
+    fn test_round_int64_negative_scale_overflow_ansi_scalar() {
+        assert_round_int64_ansi_overflows(ColumnarValue::Scalar(ScalarValue::Int64(Some(
+            5_000_000_000_000_000_000i64,
+        ))));
+    }
```
mbutrovich, line 378: Both new ANSI tests (`test_round_int64_negative_scale_overflow_ansi`, `test_round_int64_negative_scale_overflow_ansi_scalar`) only exercise the positive-overflow case (`+5e18`). There is no ANSI-mode test for the negative side (`-5e18`), even though the legacy test at lines 362-390 covers both signs. Fix: extend `assert_round_int64_ansi_overflows` calls (or add a third test) to also assert ANSI overflow for `-5_000_000_000_000_000_000i64`.

native/spark-expr/src/math_funcs/round.rs
```diff
@@ -280,4 +325,102 @@ mod test {
...
+            ColumnarValue::Scalar(ScalarValue::Int64(Some(-19))),
+        ];
+        let ColumnarValue::Scalar(ScalarValue::Int64(Some(result))) =
+            spark_round(&args, &DataType::Int64, false)?
+        else {
+            unreachable!()
+        };
+        assert_eq!(result, WRAPPED_POS_1E19);
+        Ok(())
+    }
```
mbutrovich, line 495: None of the six new tests pass a null `Int64` scalar through the widened path (`scale <= -19`) or the i128-overflow fallback (`scale <= -38`). A test with `ColumnarValue::Scalar(ScalarValue::Int64(None))` at `scale = -19` would confirm the widened-scalar branch preserves null (it does, by inspection of lines 126-140), and the same test at `scale = -40` would catch the null-handling bug at lines 141-143. Fix: add `test_round_int64_negative_scale_null_scalar` covering both scale bands.

native/spark-expr/src/math_funcs/round.rs
```diff
@@ -68,7 +91,14 @@ macro_rules! round_integer_array {
...
             })?
+        } else if let Some(div) = 10_i128.checked_pow((-(*$POINT)) as u32) {
+            let half = div / 2;
+            arrow::compute::kernels::arity::try_unary(array, |x| {
+                integer_round_widened!(x, div, half, $NATIVE, $FAIL_ON_ERROR)
+            })?
         } else {
+            // Even i128 cannot hold 10^(-point); every bounded native
+            // integer rounds to 0.
             arrow::compute::kernels::arity::try_unary(array, |_| Ok(0))?
```
mbutrovich, line 122: The i128-overflow fallback (`scale <= -38` for `Int64`) is untouched by this PR but is now reached through a three-way branch instead of two, and the PR's own reasoning documents the boundary between "fits native" and "fits i128" carefully while leaving this third boundary undocumented by a test. The existing `test_round_int64_negative_scale_below_threshold` test (scale=-20) exercises the "fits i128 but rounds to 0" case, not the "doesn't even fit i128" case. Fix: add a test at `scale = -40` (or any scale where `10_i128.checked_pow` returns `None`) asserting the result is `0` under both modes, to pin down this boundary now that it sits behind a new middle branch.

mbutrovich (changes requested):
First pass, thanks @andygrove! Also one comment that is in a different file so not easy to post inline with the other comments:

CometExpressionSuite.scala:3044-3077
The existing `"ANSI support for round function"` test iterates `scale` over `{-1000, -100, -10, -1, 0, 1, 10, 100, 1000}`. None of these values falls in the actual overflow band this PR fixes: for `Long.MaxValue`/`Long.MinValue`, only `scale` between `-19` and `-38` produces a nonzero-but-overflowing rounded value; every scale already in the list either fits comfortably (`-10`) or rounds to `0` well outside the overflow band (`-100`, `-1000`). This is the exact gap issue #5070 called out: no ANSI SQL-level test for round with a negative-scale overflow case. The new Rust unit tests don't close that gap at the Spark-integration level. Fix: add `-19` (and ideally `-20` through `-38`) to the `scale` sequence in this test, or add a new test alongside it that runs `SELECT round(CAST(5000000000000000000 AS LONG), -19)` under both ANSI settings against Spark.

native/spark-expr/src/math_funcs/round.rs
andygrove: Good catch — this was a real null-handling bug. Fixed in 285a211bd, though not quite as suggested: rather than repeat the `Some(0)`/`None` match in this third branch, `round_integer_scalar!` now checks for a null input *once*, up front, and only dispatches on the scale band for non-null values. That way all three bands preserve null by construction instead of each having to remember to. Covered by the new `test_round_int64_negative_scale_null_scalar`, which asserts null-in/null-out at scales -9, -19 and -40 under both eval modes.

native/spark-expr/src/math_funcs/round.rs
andygrove: Done in 285a211bd: `Ok((if x128 >= $HALF { $DIV } else { -$DIV }) as $NATIVE)`.

native/spark-expr/src/math_funcs/round.rs
andygrove: Done in 285a211bd — hoisted to `let point_abs = (-(*$POINT)) as u32;` in both `round_integer_array!` and `round_integer_scalar!`, and used for both `checked_pow` calls.

native/spark-expr/src/math_funcs/round.rs
andygrove: Done in 285a211bd, via `round_scalar_result!`. The duplication actually shrank further than suggested: because the null check moved out in front of the band dispatch (see your comment on line 143), each band is now a single `round_scalar_result!(...)` expression rather than a full `match`.

native/spark-expr/src/math_funcs/round.rs
andygrove: I tried this and I would like to push back, with numbers.

I implemented it exactly as described — one `integer_round!` that always computes `rem`/`half` in i128 and does a single checked-or-wrapping downcast at the end, deleting `integer_round_widened!` and reducing the dispatch to two branches (still need the `10^(-point)` doesn't-fit-i128 fallback, since `checked_pow` returns `None` there and we have no `div` to divide by). It is genuinely nicer to read, and it dissolved your other three comments on this macro for free. But it costs 2.3-3.3x on these kernels:

| Benchmark | native + widened | unified i128 | Change |
| --- | --- | --- | --- |
| int64 scale=-1 legacy | 16.0 µs | 42.9 µs | **+168%** |
| int64 scale=-1 ansi | 15.8 µs | 42.8 µs | **+171%** |
| int64 scale=-9 legacy | 18.6 µs | 43.3 µs | **+132%** |
| int64 scale=-19 legacy | 14.7 µs | 48.0 µs | **+226%** |
| int32 scale=-1 legacy | 15.2 µs | 40.5 µs | **+166%** |
| int32 scale=-1 ansi | 15.2 µs | 43.3 µs | **+185%** |

(criterion, 8192 rows, same machine, baseline re-measured back-to-back and reproducing within 3%, and the regression reproduced on a second run.)

Two causes. On the native band, `x % div` becomes a 128-bit division, which on arm64 and x86-64 is an `__umodti3`/`__udivti3` libcall rather than a single instruction. On the widened band it is worse: today that band does **no division at all** — the `|x| <= NATIVE::MAX < div` invariant makes `x % div == x`, so it is just two comparisons — whereas the unified version pays a 128-bit modulo by a divisor near 1e19, the most expensive shape there is.

So I have kept the two macros, but taken the substance of your comment, which I think is the real point: the invariant was load-bearing and only stated in prose. It is now (a) spelled out as `|x| <= NATIVE::MAX < div`, with the `x % div == x` consequence and the fact that no division is needed made explicit, and (b) enforced by a `debug_assert!` inside `integer_round_widened!`, so any future caller that violates it fails loudly in tests rather than silently computing the wrong thing. Tests run in debug, so this is live coverage.

The comparison to `decimal_round_f` I think cuts the other way, incidentally: it computes in i128 because its input *is* an i128 unscaled value, so there is no narrower type to fall back to. Here the input is i8/i16/i32/i64 and the narrow path is the common case, so paying i128 unconditionally is a real cost rather than a free simplification.

I have added `native/spark-expr/benches/round.rs` (alongside the existing `floor`/`ceil` benches) in 285a211bd so this is measurable rather than asserted — happy to be overruled if you still prefer the single path, but I did not want to take a 2-3x hit on the kernel silently.

native/spark-expr/src/math_funcs/round.rs
andygrove: Agreed, the name asserted a sign that does not hold. Renamed to `WRAPPED_1E19` / `WRAPPED_MINUS_1E19` in 285a211bd, with a comment spelling out that the value is negative (-8446744073709551616) because 1e19 exceeds `i64::MAX`, and that rounding -5e18 therefore wraps to a *positive* value.

native/spark-expr/src/math_funcs/round.rs
andygrove: Done in 285a211bd — both ANSI tests (array and scalar) now loop over `[5e18, -5e18]`, so the negative side is covered in ANSI mode as well as legacy.

native/spark-expr/src/math_funcs/round.rs
andygrove: Done in 285a211bd, as `test_round_int64_negative_scale_null_scalar`. It goes a bit wider than suggested: it asserts null-in/null-out at scale -9 (`10^9` fits in i64), -19 (fits only in i128) and -40 (fits in neither), each under both eval modes, so all three bands are pinned. The -40 case is the one that was actually broken — see the fix on your line 143 comment.

native/spark-expr/src/math_funcs/round.rs
andygrove: Done in 285a211bd, as `test_round_int64_scale_below_i128_range` at scale -40, asserting `0` for `i64::MAX`, `i64::MIN`, `0` and `5e18` under both eval modes, for both the array and scalar paths. Together with the existing scale=-20 test and the new scale=-19 tests, all three bands and both boundaries between them now have coverage.

andygrove:
@mbutrovich thanks for the review — all nine inline comments plus the `CometExpressionSuite` one are addressed in 285a211bd, with replies inline on each thread. Summary:

- **Null scalar at `scale <= -39`** returned `Some(0)` instead of `None`. Real (pre-existing) bug, fixed — but by hoisting the null check out in front of the band dispatch rather than patching the third branch, so all three bands preserve null by construction. This also collapsed the duplicated unwrap boilerplate you flagged separately.
- **`integer_round_widened!` sign arms** collapsed to one; **`(-point) as u32`** hoisted into `point_abs`.
- **Unifying `integer_round!` on i128** — I implemented it and am pushing back with benchmarks: it costs **2.3-3.3x** on these kernels, because the native band gains a 128-bit division libcall and the widened band, which today does no division at all, gains a 128-bit modulo by ~1e19. Details and the table are on that thread. I did take the substance of the comment: the load-bearing invariant is now stated precisely *and* enforced with a `debug_assert!`, and I added `native/spark-expr/benches/round.rs` so the tradeoff is measurable rather than asserted. Happy to be overruled if you still prefer the single path.
- **`WRAPPED_POS_1E19`** renamed to `WRAPPED_1E19`, with a comment noting the wrapped value is negative.
- **Test gaps** closed: ANSI overflow for -5e18, a `scale = -40` test for the doesn't-fit-i128 band (array + scalar, both modes), and a null-scalar test across all three bands.

On the `CometExpressionSuite` comment — you were right, and the root cause is worse than the scale list. That test builds an 8-column table but only ever queries `_1`, which is `Int.MaxValue`, so it never rounds a **long** at any scale; the long columns `_3`/`_4` are dead. No scale added to that list could have caught #5070. So I added a dedicated test that rounds longs (`Long.MaxValue`, `Long.MinValue`, ±5e18, 4999999999999999999, 0) at scales -39, -38, -20, -19 and -18 under both ANSI settings, which brackets both band boundaries — note -18 overflows too, since `Long.MaxValue` rounds up to 1e19 there, so this covers the native path as well as the widened one. I also added -19 and -20 to the existing test's scale sequence as you suggested.

Verified: 11 Rust unit tests pass; both Scala tests pass against real Spark on **Spark 3.5** and **Spark 4.1**. `cargo clippy --all-targets` clean, `cargo fmt` and `spotless:apply` applied.

mbutrovich (approved):
Thanks for the pushback! Approved pending CI, thanks @andygrove!

PR #518, opened 2024-06-05
chore: Add UnboundColumn to carry datatype for unbound reference

viirya:
## Which issue does this PR close?

Closes #511.

PR #5083, opened 2026-07-28
fix: honor eval mode when casting boolean to decimal

### LINKED ISSUE #5068: Cast from boolean to decimal ignores eval mode: throws where Spark returns NULL in legacy/try mode
### Describe the bug

The native boolean-to-decimal cast ignores `eval_mode` entirely and throws where Spark returns NULL in legacy and try mode.

`CometCast` marks boolean-to-decimal as `Compatible()` in all eval modes (`spark/src/main/scala/org/apache/comet/expressions/CometCast.scala:329-333`). The native implementation `cast_boolean_to_decimal` (`native/spark-expr/src/conversion_funcs/boolean.rs:31-58`) takes no `eval_mode` argument: it builds `10^scale` and unconditionally raises `SparkError::NumericValueOutOfRange` when the value does not fit the target precision.

Spark 4.1 (`Cast.scala`): `toPrecision(Decimal.ONE, target, ROUND_HALF_UP, nullOnOverflow = !ansiEnabled)`, so legacy and try mode return NULL, only ANSI throws.

Since ANSI is the default in Spark 4.x, Comet accidentally matches the default, but this is a wrong-exception bug on Spark 3.x and whenever ANSI is off, and `try_cast` returns an error instead of NULL.

### Steps to reproduce

```sql
SET spark.sql.ansi.enabled=false;
SELECT CAST(col AS DECIMAL(1,1)) FROM (SELECT true AS col);
```

Any target where `10^scale` exceeds the precision reproduces it, e.g. `DECIMAL(1,1)`, `DECIMAL(2,2)`, `DECIMAL(3,3)`.

Spark (legacy or try mode): `NULL`.
Comet: throws `NUMERIC_VALUE_OUT_OF_RANGE`.

### Expected behavior

`cast_boolean_to_decimal` should take the eval mode and return NULL for legacy/try, error only under ANSI.

### Additional context

Found by an ANSI-mode audit of all native expressions against Spark 4.1.1. Existing `CometCastSuite` coverage only uses non-overflowing targets (`decimal(10,2)`, `(14,4)`, `(30,0)`), which is why this was never caught.

andygrove:
## Summary

- Native `cast_boolean_to_decimal` ignored `EvalMode` and unconditionally raised `NUMERIC_VALUE_OUT_OF_RANGE` when `10^scale` did not fit the target precision. Spark's `Cast` delegates to `Decimal.toPrecision` with `nullOnOverflow = !ansiEnabled`, so legacy and try modes should return NULL and only ANSI should throw.
- Thread `EvalMode` through the cast: legacy/try produce NULL for `true` when the target precision is too small; ANSI still errors. `false` continues to map to 0 and always fits.
- Adds Rust unit tests covering all three eval modes and a `CometCastSuite` regression test for DECIMAL(1,1) / (2,2) / (3,3) targets (existing suite only used non-overflowing targets like DECIMAL(10,2)).

Fixes #5068

## Test plan

- [x] `cargo test -p datafusion-comet-spark-expr --lib conversion_funcs::boolean`
- [x] `cargo clippy -p datafusion-comet-spark-expr --lib --tests`
- [ ] `./mvnw test -Dsuites="org.apache.comet.CometCastSuite" -Dtest=none`

### DIFF
--- a/native/spark-expr/src/conversion_funcs/boolean.rs
+++ b/native/spark-expr/src/conversion_funcs/boolean.rs
@@ -15,9 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::{SparkError, SparkResult};
-use arrow::array::{Array, ArrayRef, AsArray, Decimal128Array, TimestampMicrosecondBuilder};
-use arrow::datatypes::DataType;
+use crate::{EvalMode, SparkError, SparkResult};
+use arrow::array::{
+    Array, ArrayRef, AsArray, Decimal128Array, Decimal128Builder, TimestampMicrosecondBuilder,
+};
+use arrow::datatypes::{is_validate_decimal_precision, DataType};
 use std::sync::Arc;
 
 pub fn is_df_cast_from_bool_spark_compatible(to_type: &DataType) -> bool {
@@ -32,28 +34,45 @@ pub fn cast_boolean_to_decimal(
     array: &ArrayRef,
     precision: u8,
     scale: i8,
+    eval_mode: EvalMode,
 ) -> SparkResult<ArrayRef> {
     let bool_array = array.as_boolean();
     let scaled_val = 10_i128.pow(scale as u32);
+
+    // Spark's Cast for boolean-to-decimal delegates to Decimal.toPrecision with
+    // nullOnOverflow = !ansiEnabled: legacy and try modes return NULL on overflow,
+    // only ANSI raises. `false` maps to 0 which always fits, so overflow only
+    // happens for `true` when 10^scale does not fit the target precision.
+    if !is_validate_decimal_precision(scaled_val, precision) {
+        match eval_mode {
+            EvalMode::Ansi => {
+                return Err(crate::error::decimal_overflow_error(
+                    scaled_val, precision, scale,
+                ));
+            }
+            EvalMode::Legacy | EvalMode::Try => {
+                let mut builder = Decimal128Builder::with_capacity(bool_array.len());
+                for i in 0..bool_array.len() {
+                    if bool_array.is_null(i) || bool_array.value(i) {
+                        builder.append_null();
+                    } else {
+                        builder.append_value(0);
+                    }
+                }
+                return Ok(Arc::new(
+                    builder.with_precision_and_scale(precision, scale)?.finish(),
+                ));
+            }
+        }
+    }
+
     let result: Decimal128Array = bool_array
         .iter()
         .map(|v| v.map(|b| if b { scaled_val } else { 0 }))
         .collect();
-
-    // Convert Arrow decimal overflow errors to SparkError
     let decimal_array = result
         .with_precision_and_scale(precision, scale)
-        .map_err(|e| {
-            if matches!(e, arrow::error::ArrowError::InvalidArgumentError(_))
-                && e.to_string().contains("too large to store in a Decimal128")
-            {
-                // Use the scaled value as it's the only non-zero value that could overflow
-                crate::error::decimal_overflow_error(scaled_val, precision, scale)
-            } else {
-                SparkError::Arrow(Arc::new(e))
-            }
-        })?;
-
+        .map_err(|e| SparkError::Arrow(Arc::new(e)))?;
     Ok(Arc::new(decimal_array))
 }
 
@@ -226,6 +245,49 @@ mod tests {
         assert_eq!(arr.value(0), expected_arr.value(0));
         assert_eq!(arr.value(1), expected_arr.value(1));
         assert!(arr.is_null(2));
+    }
+
+    #[test]
+    fn test_bool_to_decimal_overflow_legacy_returns_null() {
+        // 10^1 = 10 does not fit in DECIMAL(1,1); Spark legacy returns NULL for `true`.
+        let result = cast_array(
+            test_input_bool_array(),
+            &Decimal128(1, 1),
+            &SparkCastOptions::new(EvalMode::Legacy, "UTC", false),
+        )
+        .unwrap();
+        let arr = result.as_any().downcast_ref::<Decimal128Array>().unwrap();
+        assert!(arr.is_null(0));
+        assert_eq!(arr.value(1), 0);
+        assert!(arr.is_null(2));
+    }
+
+    #[test]
+    fn test_bool_to_decimal_overflow_try_returns_null() {
+        let result = cast_array(
+            test_input_bool_array(),
+            &Decimal128(1, 1),
+            &SparkCastOptions::new(EvalMode::Try, "UTC", false),
+        )
+        .unwrap();
+        let arr = result.as_any().downcast_ref::<Decimal128Array>().unwrap();
+        assert!(arr.is_null(0));
+        assert_eq!(arr.value(1), 0);
+        assert!(arr.is_null(2));
+    }
+
+    #[test]
+    fn test_bool_to_decimal_overflow_ansi_errors() {
+        let result = cast_array(
+            test_input_bool_array(),
+            &Decimal128(1, 1),
+            &SparkCastOptions::new(EvalMode::Ansi, "UTC", false),
+        );
+        let err = result.expect_err("expected ANSI overflow error");
+        assert!(
+            err.to_string().contains("cannot be represented"),
+            "unexpected error message: {err}"
+        );
     }
 
     #[test]
--- a/native/spark-expr/src/conversion_funcs/boolean.rs
+++ b/native/spark-expr/src/conversion_funcs/boolean.rs
@@ -15,6 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::conversion_funcs::utils::decimal_overflow_or_null;
 use crate::{EvalMode, SparkError, SparkResult};
 use arrow::array::{Array, ArrayRef, AsArray, Decimal128Array, TimestampMicrosecondBuilder};
 use arrow::datatypes::{is_validate_decimal_precision, DataType};
@@ -39,12 +40,13 @@ pub fn cast_boolean_to_decimal(
 
     // Spark's Cast uses `nullOnOverflow = !ansiEnabled`: legacy/try return NULL
     // on overflow, only ANSI raises. `false` maps to 0 which always fits, so
-    // overflow only happens for `true` when 10^scale exceeds `precision`.
+    // overflow only happens for `true` when 10^scale exceeds `precision`. The
+    // domain is two-valued, so one check outside the loop covers every row.
     let overflows = !is_validate_decimal_precision(scaled_val, precision);
-    if overflows && eval_mode == EvalMode::Ansi {
-        return Err(crate::error::decimal_overflow_error(
-            scaled_val, precision, scale,
-        ));
+    if overflows {
+        // Spark reports the logical value being cast, which for `true` is 1 --
+        // not the `10^scale` it would have been stored as.
+        decimal_overflow_or_null(1, precision, scale, eval_mode)?;
     }
 
     let result: Decimal128Array = bool_array
--- a/native/spark-expr/src/conversion_funcs/boolean.rs
+++ b/native/spark-expr/src/conversion_funcs/boolean.rs
@@ -16,9 +16,7 @@
 // under the License.
 
 use crate::{EvalMode, SparkError, SparkResult};
-use arrow::array::{
-    Array, ArrayRef, AsArray, Decimal128Array, Decimal128Builder, TimestampMicrosecondBuilder,
-};
+use arrow::array::{Array, ArrayRef, AsArray, Decimal128Array, TimestampMicrosecondBuilder};
 use arrow::datatypes::{is_validate_decimal_precision, DataType};
 use std::sync::Arc;
 
@@ -39,36 +37,23 @@ pub fn cast_boolean_to_decimal(
     let bool_array = array.as_boolean();
     let scaled_val = 10_i128.pow(scale as u32);
 
-    // Spark's Cast for boolean-to-decimal delegates to Decimal.toPrecision with
-    // nullOnOverflow = !ansiEnabled: legacy and try modes return NULL on overflow,
-    // only ANSI raises. `false` maps to 0 which always fits, so overflow only
-    // happens for `true` when 10^scale does not fit the target precision.
-    if !is_validate_decimal_precision(scaled_val, precision) {
-        match eval_mode {
-            EvalMode::Ansi => {
-                return Err(crate::error::decimal_overflow_error(
-                    scaled_val, precision, scale,
-                ));
-            }
-            EvalMode::Legacy | EvalMode::Try => {
-                let mut builder = Decimal128Builder::with_capacity(bool_array.len());
-                for i in 0..bool_array.len() {
-                    if bool_array.is_null(i) || bool_array.value(i) {
-                        builder.append_null();
-                    } else {
-                        builder.append_value(0);
-                    }
-                }
-                return Ok(Arc::new(
-                    builder.with_precision_and_scale(precision, scale)?.finish(),
-                ));
-            }
-        }
+    // Spark's Cast uses `nullOnOverflow = !ansiEnabled`: legacy/try return NULL
+    // on overflow, only ANSI raises. `false` maps to 0 which always fits, so
+    // overflow only happens for `true` when 10^scale exceeds `precision`.
+    let overflows = !is_validate_decimal_precision(scaled_val, precision);
+    if overflows && eval_mode == EvalMode::Ansi {
+        return Err(crate::error::decimal_overflow_error(
+            scaled_val, precision, scale,
+        ));
     }
 
     let result: Decimal128Array = bool_array
         .iter()
-        .map(|v| v.map(|b| if b { scaled_val } else { 0 }))
+        .map(|v| match v {
+            Some(false) => Some(0),
+            Some(true) if !overflows => Some(scaled_val),
+            _ => None,
+        })
         .collect();
     let decimal_array = result
         .with_precision_and_scale(precision, scale)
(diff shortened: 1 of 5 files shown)


native/spark-expr/src/conversion_funcs/boolean.rs
```diff
@@ -32,28 +32,32 @@ pub fn cast_boolean_to_decimal(
...
     let scaled_val = 10_i128.pow(scale as u32);
+
+    // Spark's Cast uses `nullOnOverflow = !ansiEnabled`: legacy/try return NULL
+    // on overflow, only ANSI raises. `false` maps to 0 which always fits, so
+    // overflow only happens for `true` when 10^scale exceeds `precision`.
+    let overflows = !is_validate_decimal_precision(scaled_val, precision);
+    if overflows && eval_mode == EvalMode::Ansi {
+        return Err(crate::error::decimal_overflow_error(
+            scaled_val, precision, scale,
+        ));
```
mbutrovich: The ANSI error reports the wrong value. `decimal_overflow_error(scaled_val, precision, scale)` passes the already-scaled internal representation (`10_i128.pow(scale)`, e.g. 10 for `Decimal(1,1)`). Spark's actual path for boolean-to-decimal ANSI overflow is `Cast.scala` calling `toPrecision(if (b) Decimal.ONE else Decimal.ZERO, target, ...)`, and `Decimal.toPrecision` on failure calls `DataTypeErrors.cannotChangeDecimalPrecisionError(this, precision, scale, context)` where `this` is `Decimal.ONE`, i.e. the unscaled logical value 1. `Decimal.ONE.toPlainString` is `"1"`, so Spark's message is `1 cannot be represented as Decimal(1, 1)....`. On the JVM side, `ShimSparkErrorConverter` (`spark/src/main/spark-3.4/org/apache/spark/sql/comet/shims/ShimSparkErrorConverter.scala:107-113`, same in spark-3.5) reconstructs `Decimal(params("value").toString)` and calls the same `cannotChangeDecimalPrecisionError`, so whatever string Comet puts in `value` is parsed back as a `Decimal` verbatim. Since Comet passes `"10"`, the reconstructed message is `10 cannot be represented as Decimal(1, 1)...`, which does not match Spark's `1 cannot be represented as Decimal(1, 1)...`.
This mismatch predates this PR (the old `map_err` branch also passed `scaled_val` into `decimal_overflow_error`), but it was never exercised by any test until this PR added the `CometCastSuite` regression test that now runs `true` values through `DECIMAL(1,1)/(2,2)/(3,3)` under ANSI. On the `spark-3.4` and `spark-3.5` profiles, `castTest`'s ANSI branch asserts `cometMessage == sparkMessage` exactly (`spark/src/test/scala/org/apache/comet/CometCastSuite.scala:2431`), so the new test will fail the exact-message comparison on those profiles. On `spark-4.0`/`spark-4.1` it will not be caught, because the `isSpark40Plus` branch only compares `cometMessage.substring(0, min(40, cometMessage.length))` as a prefix (`CometCastSuite.scala:2426-2428`), and 40 characters only covers the `[NUMERIC_VALUE_OUT_OF_RANGE.WITH_SUGGESTION] ` tag, not the value.
Fix: pass the pre-scale logical value, matching the convention used by `cast_int_to_decimal128_internal` (`native/spark-expr/src/conversion_funcs/numeric.rs:750-754`), which reports the original face-value input, not the post-multiplier value. For boolean, that is `1` for `true`: replace `decimal_overflow_error(scaled_val, precision, scale)` with `decimal_overflow_error(1, precision, scale)`.

native/spark-expr/src/conversion_funcs/boolean.rs
```diff
@@ -32,28 +32,32 @@ pub fn cast_boolean_to_decimal(
...
+            _ => None,
+        })
         .collect();
-
-    // Convert Arrow decimal overflow errors to SparkError
     let decimal_array = result
         .with_precision_and_scale(precision, scale)
-        .map_err(|e| {
-            if matches!(e, arrow::error::ArrowError::InvalidArgumentError(_))
-                && e.to_string().contains("too large to store in a Decimal128")
```
mbutrovich, line 48: This is the third hand-written copy of the same decision shape: check `is_validate_decimal_precision`, then match `eval_mode` to decide null vs `decimal_overflow_error`. The other two copies are `native/spark-expr/src/conversion_funcs/numeric.rs:744-772` (`cast_int_to_decimal128_internal`, both the `checked_mul` overflow branch and the post-multiply precision-check branch) and `numeric.rs:915-937` (`cast_floating_point_to_decimal128`'s ANSI rescan). Each function's scan strategy should stay different: boolean's domain is 2-valued so a single precheck outside the loop is correct and should not be forced into a per-row loop like the int/float paths. But the branch body itself, "precision check fails, Ansi errors with `decimal_overflow_error(value, precision, scale)`, else null", is identical in shape across all three sites and has already drifted once: boolean passes the scaled value while int/float pass the face value (see the finding above). Extract a small helper, e.g. `fn decimal_overflow_result(value: i128, precision: u8, scale: i8, eval_mode: EvalMode) -> SparkResult<Option<i128>>` in `numeric.rs` or a shared `decimal.rs`, and have all three call sites call it with the correct pre-scale `value` argument. This forces the value convention to be decided once instead of letting it diverge a third time.

native/spark-expr/src/conversion_funcs/boolean.rs
```diff
@@ -32,28 +32,32 @@ pub fn cast_boolean_to_decimal(
...
-                // Use the scaled value as it's the only non-zero value that could overflow
-                crate::error::decimal_overflow_error(scaled_val, precision, scale)
-            } else {
-                SparkError::Arrow(Arc::new(e))
-            }
-        })?;
-
+        .map_err(|e| SparkError::Arrow(Arc::new(e)))?;
     Ok(Arc::new(decimal_array))
 }
```
mbutrovich, line 64: `is_validate_decimal_precision` from arrow-rs (`arrow-data/src/decimal.rs`, re-exported via `arrow::datatypes`) is the right existing kernel for the precision check and is used consistently with the same import already in `numeric.rs`. There is no arrow-rs or DataFusion cast kernel that already implements Spark's `nullOnOverflow = !ansiEnabled` semantics for decimal casts, since that is Spark-specific behavior rather than a general Arrow cast concept, so a hand-written check here is expected and this part does not need to change. Dropping the old string-matching error detection (`e.to_string().contains("too large to store in a Decimal128")`) in favor of the `eval_mode` precheck removes a fragile pattern and is worth keeping as is.

spark/src/test/scala/org/apache/comet/CometCastSuite.scala
```diff
@@ -189,6 +189,18 @@ class CometCastSuite extends CometTestBase with AdaptiveSparkPlanHelper {
...
+    // Regression for https://github.com/apache/datafusion-comet/issues/5068.
+    // `true` scales to 10^scale, which does not fit in DECIMAL(1,1)/(2,2)/(3,3).
+    // Spark returns NULL in legacy/try mode and only throws under ANSI.
+    Seq(
+      DataTypes.createDecimalType(1, 1),
+      DataTypes.createDecimalType(2, 2),
+      DataTypes.createDecimalType(3, 3)).foreach { dt =>
+      castTest(generateBools(), dt)
+    }
+  }
```
mbutrovich, line 208: The new test relies on `castTest`'s default `testAnsi = true` to exercise the ANSI overflow path, but nothing in the test documents that this is intentional or names the ANSI behavior being checked, unlike the comment block above it which only describes the legacy/try semantics. Add a line noting that `castTest`'s ANSI branch, enabled by default, is what exercises the `EvalMode::Ansi` throw path added in `boolean.rs`, since a reader scanning this test in isolation cannot otherwise tell that ANSI is covered here at all.

mbutrovich (changes requested):

native/spark-expr/src/conversion_funcs/boolean.rs
andygrove: Confirmed and fixed in 815cc8a6f. I ran the new test on `spark-3.5` before changing anything, and it fails exactly as you predicted:

```
- cast BooleanType to DecimalType where 10^scale overflows precision *** FAILED ***
  "...ALUE_OUT_OF_RANGE] 1[0] cannot be represent..." did not equal
  "...ALUE_OUT_OF_RANGE] 1[] cannot be represent..." (CometCastSuite.scala:2431)
```

So the PR as it stood would have failed CI on `spark-3.4`/`spark-3.5`, and your reading of why `spark-4.x` misses it is right too — the 40-character prefix stops inside `[NUMERIC_VALUE_OUT_OF_RANGE.WITH_SUGGESTION] ` and never reaches the value. Worth noting the decimal-to-decimal escape hatch just above (`assert(cometMessage.contains("too large to store"))`) does not apply here either, since the *input* is boolean, so the exact comparison is what runs.

Fixed by passing the pre-scale logical value, matching `cast_int_to_decimal128_internal`. The four `cast BooleanType to DecimalType*` tests now pass on `spark-3.5`.

native/spark-expr/src/conversion_funcs/boolean.rs
andygrove: Thanks for checking this explicitly — agreed, and no change made here. Leaving `is_validate_decimal_precision` and the `eval_mode` precheck as they are.

spark/src/test/scala/org/apache/comet/CometCastSuite.scala
andygrove: Added in 815cc8a6f. The comment now names ANSI explicitly: that `castTest`'s default `testAnsi = true` is what drives the `EvalMode::Ansi` throw path in `boolean.rs`, and that on Spark 3.4/3.5 the message comparison is exact, so this test also pins the reported value to Spark's logical `1` rather than the scaled `10^scale`. That second part is worth stating given it is exactly what the first thread caught.

native/spark-expr/src/conversion_funcs/boolean.rs
andygrove: Done in 815cc8a6f, with one change to the signature you proposed.

`SparkResult<Option<i128>>` with an `i128` value does not fit the third site: `cast_floating_point_to_decimal128` reports an **f64** (`input_value.to_string()`), so forcing it through an `i128` parameter would change the string it reports (`1e30` vs a 31-digit integer) and break message parity for the float path. And the `Option<i128>` return would always be `None` in practice, since the helper never produces a value — only an error or a null.

So the helper is:

```rust
pub(crate) fn decimal_overflow_or_null<T: Display>(
    value: T,
    precision: u8,
    scale: i8,
    eval_mode: EvalMode,
) -> SparkResult<()>
```

`Ok(())` means "caller should emit NULL", `Err` is the ANSI error. Generic over `Display` so each site reports its own natural type, and the doc comment states the pre-scale convention and why (Spark's `Decimal.toPrecision` reports the logical value, and `ShimSparkErrorConverter` parses the string back into a `Decimal` verbatim). It lives in `conversion_funcs/utils.rs`, which both `boolean.rs` and `numeric.rs` already share.

All three sites route through it, and as you asked, each keeps its own scan strategy — boolean's single precheck stays outside the loop rather than becoming per-row.

One extra simplification it enabled: in `cast_int_to_decimal128_internal` the `checked_mul` overflow branch and the precision-check branch had identical bodies, both reporting `v`, so they collapse into one arm:

```rust
match scaled {
    Some(scaled) if is_validate_decimal_precision(scaled, precision) => {
        builder.append_value(scaled)
    }
    // Either the multiply overflowed i128 or the product exceeds `precision`.
    _ => {
        decimal_overflow_or_null(v, precision, scale, eval_mode)?;
        builder.append_null();
    }
}
```

62 `conversion_funcs` unit tests pass and clippy `--all-targets` is clean.

andygrove:
@mbutrovich all four threads addressed in 815cc8a6f, details inline. Summary:

**The wrong-value finding was real and would have failed CI.** I ran the new test on `spark-3.5` before touching anything:

```
- cast BooleanType to DecimalType where 10^scale overflows precision *** FAILED ***
  "...ALUE_OUT_OF_RANGE] 1[0] cannot be represent..." did not equal
  "...ALUE_OUT_OF_RANGE] 1[] cannot be represent..."
```

Exactly your diagnosis. Fixed by reporting the pre-scale logical value, matching `cast_int_to_decimal128_internal`.

**Helper extracted, with one signature change.** `SparkResult<Option<i128>>` taking an `i128` does not fit the float site, which reports an `f64` — coercing it would alter the reported string and break message parity there — and the `Option` would always be `None` since the helper only ever errors or nulls. So it is `decimal_overflow_or_null<T: Display>(value, precision, scale, eval_mode) -> SparkResult<()>` in `conversion_funcs/utils.rs`, generic over the displayed value, with the pre-scale convention documented in one place. All three sites call it; each keeps its own scan strategy, and boolean's two-valued precheck stays outside the loop as you asked. It also let the two identical branches in `cast_int_to_decimal128_internal` collapse into one.

**Test comment** now names the ANSI coverage and notes that the 3.4/3.5 exact comparison is what pins the reported value.

**Verified on spark-3.5:** all four `cast BooleanType to DecimalType*` tests pass; 62 `conversion_funcs` unit tests pass; clippy `--all-targets` clean.

One unrelated thing I hit while regression-testing the refactor: `cast StringType to DecimalType(38,10) high precision` fails when the `to DecimalType` tests run as a batch, but passes in isolation, and it fails identically on the base commit (base: 2 failures in that batch, with this change: 1, the difference being the boolean test this PR fixes). The exception comes from Spark's own `Decimal.changePrecision` → `BigInteger.pow`, so it looks like a pre-existing order dependency, most likely a leaked `allowNegativeScaleOfDecimal` conf from a neighbouring test. Untouched here — worth a separate issue if it is not already known.

mbutrovich (approved):

comphead:
checking

comphead (commented):
Thanks @andygrove there is some points we may want to address


## 1. [HIGH · spark-compat] ANSI raises eagerly on target type, before any row is inspected
  **File:** `native/spark-expr/src/conversion_funcs/boolean.rs` (`cast_boolean_to_decimal`).

  **Description.** After the change, `overflows = !is_validate_decimal_precision(10^scale, precision)` is computed once from the target type, and `decimal_overflow_or_null(1, precision, scale, eval_mode)?` is
  invoked *before* the per-row loop. Under ANSI this `?` propagates `Err(NumericValueOutOfRange)` regardless of what the input array actually contains. Any `false`-only, `null`-only, or empty boolean batch
  cast to `DECIMAL(p,s)` with `s >= p` now fails, where Spark returns `0.0` / `NULL` / empty.
  
  **Reason.** Spark's `Cast.scala` (branches 3.4, 3.5, 4.0, 4.1, master, all at `sql/catalyst/.../Cast.scala:989-991`) implements the cast as a per-row closure `buildCast[Boolean](_, b => toPrecision(if (b)
  Decimal.ONE else Decimal.ZERO, target, ctx))`. `Decimal.ZERO.toPrecision(p, s)` succeeds for every `DecimalType`, and `null` rows never invoke the closure. `cannotChangeDecimalPrecisionError` therefore fires
  only when a `true` row is actually reached. The PR conflates a schema-level property with a per-row evaluation.
  
  **Concrete divergence.** With `spark.sql.ansi.enabled=true`:
  - `SELECT CAST(CAST(false AS BOOLEAN) AS DECIMAL(1,1))` — Spark returns `0.0`, Comet throws `[NUMERIC_VALUE_OUT_OF_RANGE.WITH_SUGGESTION] 1 cannot be represented as Decimal(1, 1)`.
  - Boolean column of only `false`/`NULL` cast to `DECIMAL(1,1)` / `(2,2)` / `(3,3)` / `(38,38)` — same divergence.

  ---

  ## 2. [MEDIUM · test-gap] No test exercises a `false`/`null`-only column under ANSI with an overflowing target
  **File:** `spark/src/test/scala/org/apache/comet/CometCastSuite.scala`.

  **Description.** The new suite entry `cast BooleanType to DecimalType where 10^scale overflows precision` uses `castTest(generateBools(), Decimal(1,1)/(2,2)/(3,3))`. `generateBools()` is `withNulls(Seq(true,
  false)).toDF("a")` (line 1943), so every batch contains a `true` row. Both Spark and Comet throw under ANSI on that row, so the test passes even with the eager-raise regression in place.

  **Reason.** The regression only manifests when Comet's eager raise fires but Spark's per-row check would not — i.e. a batch that omits `true`. Coverage is missing exactly the case the new invariant needs to
  guard.

  **Example test.**
  ```scala
  test("cast BooleanType to overflowing DecimalType with false/null-only column matches Spark") {
    // Spark returns 0.0 / null; PR path currently raises NUMERIC_VALUE_OUT_OF_RANGE under ANSI.
    val df = withNulls(Seq(false)).toDF("a")
    castTest(df, DataTypes.createDecimalType(1, 1))
    castTest(df, DataTypes.createDecimalType(2, 2))
    castTest(df, DataTypes.createDecimalType(38, 38))
  }

  test("cast empty BooleanType column to overflowing DecimalType matches Spark") {
    val df = spark.createDataFrame(
      java.util.Collections.emptyList[Row](),
      StructType(Seq(StructField("a", BooleanType, nullable = true)))
    )
    castTest(df, DataTypes.createDecimalType(1, 1))
  }
  ```
  
  A SQL-file test also covers it (project convention prefers these for expressions):
  ```sql
  -- spark/src/test/resources/sql-tests/expressions/cast_boolean_to_decimal_ansi.sql
  -- ConfigMatrix: spark.sql.ansi.enabled=true

  statement
  CREATE TABLE t_bool_dec(b boolean) USING parquet

  statement
  INSERT INTO t_bool_dec VALUES (false), (NULL)

  query
  SELECT CAST(b AS DECIMAL(1,1)) FROM t_bool_dec
  ```

  ---

  ## 3. [LOW · test-gap] Precision/scale boundaries not covered
  **File:** `spark/src/test/scala/org/apache/comet/CometCastSuite.scala`.

  **Description.** Overflow depends on `scale >= precision`. The PR pins three interior points `(1,1)/(2,2)/(3,3)` and never exercises the boundaries.

  **Reason.** Auto-memory `feedback_sql_test_coverage.md`: integer/decimal max/min boundaries. The just-fits boundary and the `Decimal(38, *)` extremes are load-bearing for the `10_i128.pow(scale as u32)` call
  and the `is_validate_decimal_precision` check.

  **Example test.**
  ```scala
  test("cast BooleanType to DecimalType precision/scale boundaries") {
    val bools = generateBools()
    castTest(bools, DataTypes.createDecimalType(1, 0))    // fits: 10^0 = 1 <= 9
    castTest(bools, DataTypes.createDecimalType(38, 37))  // just fits: 10^37 has 38 digits
    castTest(bools, DataTypes.createDecimalType(38, 38))  // maximum overflow at the DecimalType ceiling
    castTest(bools, DataTypes.createDecimalType(38, 0))   // scale = 0, no overflow possible
    castTest(bools, DataTypes.createDecimalType(5, 5))    // another scale = precision point
  }
  ```

  ---

  ## 4. [LOW · test-gap] ANSI error message value not asserted (logical `1` vs scaled `10^scale`)
  **Files:** `native/spark-expr/src/conversion_funcs/boolean.rs` unit tests.

  **Description.** The PR's docstring in `utils.rs` and the boolean.rs comment both emphasise that Spark reports the pre-scale logical value (`1`), not `10^scale`. The added Rust test only checks that the
  message contains `"cannot be represented"`; no assertion pins the reported value string. `castTest` in Scala also does not exact-match on the value token for Spark >= 4.

  **Reason.** The whole purpose of the `value: T: Display` parameter in `decimal_overflow_or_null` is preserving Spark's user-visible error text. That property has no guard.

  **Example test.**
  ```rust
  #[test]
  fn test_bool_to_decimal_ansi_error_reports_logical_one_not_scaled() {
      let arr: ArrayRef = Arc::new(BooleanArray::from(vec![Some(true)]));
      let err = cast_boolean_to_decimal(&arr, 3, 3, EvalMode::Ansi)
          .unwrap_err()
          .to_string();
      // Spark reports the pre-scale literal `1`, not the scaled 10^3 = 1000.
      assert!(err.contains(" 1 cannot be represented"), "was: {err}");
      assert!(!err.contains("1000"), "must not report the scaled value: {err}");
  }
  ```

  ---

  ## 5. [LOW · code-quality] `decimal_overflow_or_null` duplicates the `NumericValueOutOfRange` constructor already in `decimal_overflow_error`
  **Files:** `native/spark-expr/src/conversion_funcs/utils.rs`, `native/common/src/error.rs:743`.

  **Description.** `decimal_overflow_or_null` builds `SparkError::NumericValueOutOfRange { value, precision, scale }` directly. The pre-existing `decimal_overflow_error(value, precision, scale)` in 
  `native/common/src/error.rs:743` still exists and is used by `math_funcs/internal/checkoverflow.rs:147,200` and `math_funcs/internal/make_decimal.rs:78`. There are now two constructors for the same variant.

  **Reason.** Two constructors invite drift. If the variant grows a field, both callers need updating. The eval-mode branch is orthogonal to error construction.

andygrove:
Thanks for the reviews @mbutrovich @comphead. This is turning into far too much work for an expression that nobody should be using in the first place, so I'm going to close this and replace with https://github.com/apache/datafusion-comet/pull/518 to remove the native implementation and just use codegen dispatch instead.

PR #5084, opened 2026-07-28
fix: throw ARITHMETIC_OVERFLOW for Long.MinValue div -1 under ANSI mode

### LINKED ISSUE #5065: IntegralDivide returns wrong result for Long.MinValue div -1 under ANSI mode
### Describe the bug

Under ANSI mode, `IntegralDivide` (the `div` operator) silently returns a wrong value for `Long.MinValue div -1` instead of throwing `ARITHMETIC_OVERFLOW`.

Spark 4.1 (`arithmetic.scala`) sets `checkDivideOverflow = true` for `LongType` when `failOnError` is set, and throws `QueryExecutionErrors.overflowInIntegralDivideError` ("Overflow in integral divide. Use `try_divide` to tolerate divisor being 0 and return NULL instead.") when `input1 == Long.MinValue && input2 == -1`.

Comet's serde (`spark/src/main/scala/org/apache/comet/serde/arithmetic.scala`, `CometIntegralDivide`) casts both operands to `DECIMAL(19,0)` and performs a decimal divide. The result `9223372036854775808` fits in precision 19, so the `CheckOverflow` wrapper passes. The final cast back to `LongType` is hard-coded to legacy eval mode:

```scala
// arithmetic.scala:348
CometCast.castToProto(expr, None, LongType, childExpr.get, CometEvalMode.LEGACY)
```

The legacy decimal-to-int64 path truncates (`native/spark-expr/src/conversion_funcs/numeric.rs`, `cast_decimal_to_int32_up!` non-ANSI arm), producing `-9223372036854775808` with no error.

`SparkError::IntegralDivideOverflow` already exists in `native/common/src/error.rs` and is wired through `ShimSparkErrorConverter`, but no native kernel ever constructs it.

### Steps to reproduce

```sql
SELECT CAST(-9223372036854775808 AS LONG) div CAST(-1 AS LONG);
```

Spark: throws `SparkArithmeticException` with error class `ARITHMETIC_OVERFLOW` ("Overflow in integral divide").
Comet: returns `-9223372036854775808`.

### Expected behavior

Comet should throw the same `ARITHMETIC_OVERFLOW` error as Spark under ANSI mode (and return the wrapped value under legacy mode, which it already does).

### Additional context

Found by an ANSI-mode audit of all native expressions against Spark 4.1.1. Divide-by-zero handling in `IntegralDivide` is correct; only the `Long.MinValue div -1` overflow case is affected. There is no existing test for this input.

andygrove:

Closes #5065.

## Rationale for this change

Under ANSI mode, Spark throws `ARITHMETIC_OVERFLOW` ("Overflow in integral divide") for `Long.MinValue div -1`, since the quotient `9223372036854775808` cannot be represented as a `LONG`. Comet performs integral division by casting both operands to `DECIMAL(19,0)`, dividing, and casting the result back to `LONG` with legacy (wrapping) semantics, so it silently returned `-9223372036854775808` instead of throwing.

Spark only performs this check when the operands are `LONG` (`checkDivideOverflow` in `arithmetic.scala`): `DECIMAL` operands wrap around on the implicit cast to `LONG` even in ANSI mode, and smaller integer types cannot overflow because the result type is `LONG`. The fix therefore has to know the original operand type, which is no longer visible on the native side after the serde casts both operands to decimal.

- Add a `check_divide_overflow` field to the `MathExpr` protobuf message, set by `CometIntegralDivide` only when the left operand is `LONG` and eval mode is ANSI (mirroring Spark's `checkDivideOverflow`)
- Plumb the flag through `IntegralDivideBuilder` / `BinaryExprOptions` into the existing `fail_on_error` parameter of `create_comet_physical_fun_with_eval_mode`, with a new `make_comet_scalar_udf!` arm that passes both `eval_mode` and `fail_on_error`
- In `spark_decimal_div_internal`, when the flag is set and the quotient does not fit in an `i64`, return the previously never-constructed `SparkError::IntegralDivideOverflow`, which already maps to `SparkArithmeticException` with error class `ARITHMETIC_OVERFLOW`
- Align the `IntegralDivideOverflow` message with Spark's wording (`Use 'try_divide' ...` with single quotes rather than backticks, matching Spark 3.4 through 4.1)

New cases in `expressions/math/arithmetic_ansi.sql` (verified to fail before the fix):

- `a div b` and literal `div` for `Long.MinValue div -1` expect an error matching `Overflow in integral divide` from both Spark and Comet
- Pinning queries for the cases that must NOT throw: `INT_MIN div -1`, `LONG_MIN div 1`, and `DECIMAL(19,0)` operands (which wrap around even in ANSI mode, matching Spark)

Also ran the full `CometSqlFileTestSuite` (427 tests) and the `div`-related tests in `CometExpressionSuite`, plus `cargo clippy --all-targets --workspace -- -D warnings`.

### DIFF
--- a/native/common/src/error.rs
+++ b/native/common/src/error.rs
@@ -69,7 +69,7 @@ pub enum SparkError {
     #[error("[ARITHMETIC_OVERFLOW] {from_type} overflow. If necessary set \"spark.sql.ansi.enabled\" to \"false\" to bypass this error.")]
     ArithmeticOverflow { from_type: String },
 
-    #[error("[ARITHMETIC_OVERFLOW] Overflow in integral divide. Use `try_divide` to tolerate overflow and return NULL instead. If necessary set \"spark.sql.ansi.enabled\" to \"false\" to bypass this error.")]
+    #[error("[ARITHMETIC_OVERFLOW] Overflow in integral divide. Use 'try_divide' to tolerate overflow and return NULL instead. If necessary set \"spark.sql.ansi.enabled\" to \"false\" to bypass this error.")]
     IntegralDivideOverflow,
 
     #[error("[ARITHMETIC_OVERFLOW] Overflow in sum of decimals. Use `try_{function_name}` to tolerate overflow and return NULL instead. If necessary set \"spark.sql.ansi.enabled\" to \"false\" to bypass this error.")]
--- a/spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
+++ b/spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
@@ -107,13 +107,11 @@ SELECT a * 2 FROM ansi_long_overflow WHERE a = 9223372036854775807
 -- ============================================================================
 
 -- LONG_MIN div -1 should overflow
--- (match on the message body rather than the ARITHMETIC_OVERFLOW condition name,
--- which is not rendered on all Spark versions)
-query expect_error(Overflow in integral divide)
+query expect_error(ARITHMETIC_OVERFLOW)
 SELECT a div b FROM ansi_long_overflow WHERE a = -9223372036854775808 AND b = -1
 
 -- literal LONG_MIN div -1 should overflow
-query expect_error(Overflow in integral divide)
+query expect_error(ARITHMETIC_OVERFLOW)
 SELECT -9223372036854775808L div -1L
 
 -- INT_MIN div -1 does not overflow because the result type is LONG
--- a/spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
+++ b/spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
@@ -103,6 +103,33 @@ query expect_error(ARITHMETIC_OVERFLOW)
 SELECT a * 2 FROM ansi_long_overflow WHERE a = 9223372036854775807
 
 -- ============================================================================
+-- Integral divide overflow
+-- ============================================================================
+
+-- LONG_MIN div -1 should overflow
+-- (match on the message body rather than the ARITHMETIC_OVERFLOW condition name,
+-- which is not rendered on all Spark versions)
+query expect_error(Overflow in integral divide)
+SELECT a div b FROM ansi_long_overflow WHERE a = -9223372036854775808 AND b = -1
+
+-- literal LONG_MIN div -1 should overflow
+query expect_error(Overflow in integral divide)
+SELECT -9223372036854775808L div -1L
+
+-- INT_MIN div -1 does not overflow because the result type is LONG
+query
+SELECT a div b FROM ansi_int_overflow WHERE a = -2147483648 AND b = -1
+
+-- LONG_MIN div 1 does not overflow
+query
+SELECT a div b FROM ansi_long_overflow WHERE b = 1
+
+-- Spark only checks integral divide overflow for LONG operands; DECIMAL operands
+-- wrap around on the implicit cast to LONG even in ANSI mode
+query
+SELECT CAST(a AS DECIMAL(19,0)) div CAST(b AS DECIMAL(19,0)) FROM ansi_long_overflow WHERE a = -9223372036854775808 AND b = -1
+
+-- ============================================================================
 -- Integer division by zero
 -- ============================================================================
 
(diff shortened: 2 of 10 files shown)


spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
```diff
@@ -102,6 +102,33 @@ SELECT a - b FROM ansi_long_overflow WHERE a = -9223372036854775808
...
 SELECT a * 2 FROM ansi_long_overflow WHERE a = 9223372036854775807
 
+-- ============================================================================
+-- Integral divide overflow
+-- ============================================================================
+
+-- LONG_MIN div -1 should overflow
+-- (match on the message body rather than the ARITHMETIC_OVERFLOW condition name,
+-- which is not rendered on all Spark versions)
+query expect_error(Overflow in integral divide)
```
comphead: can we make the error to be consistent with other tests in this file ?
```
query expect_error(ARITHMETIC_OVERFLOW)
```

native/common/src/error.rs
```diff
@@ -69,7 +69,7 @@ pub enum SparkError {
     #[error("[ARITHMETIC_OVERFLOW] {from_type} overflow. If necessary set \"spark.sql.ansi.enabled\" to \"false\" to bypass this error.")]
     ArithmeticOverflow { from_type: String },
 
-    #[error("[ARITHMETIC_OVERFLOW] Overflow in integral divide. Use `try_divide` to tolerate overflow and return NULL instead. If necessary set \"spark.sql.ansi.enabled\" to \"false\" to bypass this error.")]
+    #[error("[ARITHMETIC_OVERFLOW] Overflow in integral divide. Use 'try_divide' to tolerate overflow and return NULL instead. If necessary set \"spark.sql.ansi.enabled\" to \"false\" to bypass this error.")]
```
kazuyukitanimura, line 72: Not sure whether this is a proper back quote "`"

kazuyukitanimura (approved):
one comment

spark/src/test/resources/sql-tests/expressions/math/arithmetic_ansi.sql
andygrove: Done in f186668d7 — both now use `query expect_error(ARITHMETIC_OVERFLOW)`.

My original caveat about the condition name not being rendered on all versions was wrong, so I dropped that comment too. Verified the file passes on all five profiles: `spark-3.4`, `spark-3.5`, `spark-4.0`, `spark-4.1`, `spark-4.2`.

native/common/src/error.rs
andygrove: Good catch to question it — it does look inconsistent with the `DIVIDE_BY_ZERO` variant a few lines below, but the single quote (plain ASCII `U+0027`) is intentional here. The inconsistency is Spark's, not ours.

The two messages are built by different mechanisms:

**`DIVIDE_BY_ZERO`** has the hint hardcoded in the message template in `error-classes.json` / `error-conditions.json`, with backticks:

```
"DIVIDE_BY_ZERO" : {
  "message" : [
    "Division by zero. Use `try_divide` to tolerate divisor being 0 and return NULL instead. If necessary set <config> to \"false\" to bypass this error."
  ],
```

**`ARITHMETIC_OVERFLOW`** has no hardcoded hint — the template just interpolates an `<alternative>` parameter:

```
"ARITHMETIC_OVERFLOW" : {
  "message" : [
    "<message>.<alternative> If necessary set <config> to \"false\" to bypass this error."
  ],
```

and `<alternative>` is filled in from Scala, which uses a single quote:

```scala
val alternative = if (hint.nonEmpty) {
  s" Use '$hint' to tolerate overflow and return NULL instead."
} else ""
```

(`QueryExecutionErrors.arithmeticOverflowError` in Spark 3.4/3.5, moved to `ExecutionErrors` in `sql/api` for Spark 4.x.)

I confirmed this by dumping the string constants out of the compiled Spark jars for 3.4.3, 3.5.8, 4.0.1, 4.1.1 and 4.2.0 — all five yield `` Use ' `` / `` ' to tolerate overflow and return NULL instead. ``, i.e. `U+0027` on both sides, with no version drift. So the backticks were the bug and this line now matches Spark exactly.

comphead (approved):

PR #5105, opened 2026-07-28
chore: [branch-1.0] change version from 1.0.0-SNAPSHOT to 1.0.0

andygrove:

N/A - this is a release process step for the 1.0.0 release.

## Rationale for this change

The [release process](https://github.com/apache/datafusion-comet/blob/main/docs/source/contributor-guide/release_process.md#update-maven-version) requires that the Maven version on the release branch is changed from `1.0.0-SNAPSHOT` to `1.0.0` before the release candidate is tagged, so that the staged Maven artifacts and the source tarball carry the final release version.

- Change the Maven version from `1.0.0-SNAPSHOT` to `1.0.0` in `pom.xml`, `common/pom.xml`, `spark/pom.xml`, and `spark-integration/pom.xml`.
- Update `comet.version` in the Spark test diffs (`dev/diffs/3.4.3.diff`, `3.5.8.diff`, `4.0.2.diff`, `4.1.2.diff`).
- Update the `comet` version in the Iceberg diffs (`dev/diffs/iceberg/1.8.1.diff`, `1.9.1.diff`, `1.10.0.diff`, `1.11.0.diff`).

The Rust crate versions are already `1.0.0` and do not need to change.

Existing CI. The Spark SQL and Iceberg test workflows apply the updated diffs and build against the new version, so they cover the version references in both the poms and the diff files.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -26,7 +26,7 @@ under the License.
   <parent>
     <groupId>org.apache.datafusion</groupId>
     <artifactId>comet-parent-spark${spark.version.short}_${scala.binary.version}</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>1.0.0</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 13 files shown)


mbutrovich (approved):
Thanks @andygrove!

PR #5107, opened 2026-07-28
fix: seed native Parquet scan reader options from session config

### LINKED ISSUE #4990: Native Parquet scan ignores session-level datafusion.execution.parquet.* options
### Describe the bug

Comet's native Parquet scan ignores all session-level `datafusion.execution.parquet.*` options. `get_options()` in `native/core/src/parquet/parquet_exec.rs` constructs a fresh default:

```rust
let mut table_parquet_options = TableParquetOptions::new();
table_parquet_options.global.coerce_int96 = Some("us".to_string());
table_parquet_options.global.coerce_int96_tz = Some("UTC".to_string());
```

`ParquetSource` reads reader settings from `table_parquet_options.global` (`datafusion-datasource-parquet-54.0.0/src/source.rs:465` for `max_predicate_cache_size`, and similarly for the other reader options), so anything set on the session config is discarded for the scan.

`pushdown_filters` and `reorder_filters` are the exceptions, and only by accident: `try_pushdown_filters` consults `state.config_options()` directly and ORs the two sources together (`source.rs:739-740`):

```rust
let config_pushdown_enabled = config.execution.parquet.pushdown_filters;
let table_pushdown_enabled = self.pushdown_filters();
let pushdown_filters = table_pushdown_enabled || config_pushdown_enabled;
```

Every other reader option, including `max_predicate_cache_size`, `bloom_filter_on_read`, `binary_as_string`, `schema_force_view_types`, and `force_filter_selections`, is silently inert.

This interacts with the `spark.comet.datafusion.*` passthrough in `jni_api.rs`, which translates such keys into DataFusion session options. That passthrough is itself gated behind `spark.comet.exec.respectDataFusionConfigs` (default `false`, `CometConf.scala:839`), which is intentional and documented as a testing option. The bug is that even with that flag enabled, parquet reader options still do not take effect, so the escape hatch appears to work while doing nothing.

### Steps to reproduce

With `spark.comet.exec.respectDataFusionConfigs=true`, set `spark.comet.datafusion.execution.parquet.max_predicate_cache_size` to `0`, `104857600`, and `1073741824` in turn, with `spark.comet.parquet.rowFilterPushdown.enabled=true`, and run any query with a pushed filter.

The `Predicate cache: records read from the cache` and `Predicate cache: rows physically read and decoded` scan metrics are byte-for-byte identical across all three, confirming the setting never reaches the reader. On TPC-H SF100 Q6 all three reported 48,791,761,520 and 18,001,134,400 respectively.

A more direct check: set only `spark.comet.datafusion.execution.parquet.pushdown_filters=true` (leaving `spark.comet.parquet.rowFilterPushdown.enabled` unset). The scan still emits every row (600,037,902 for SF100 `lineitem`), showing the passthrough had no effect.

### Expected behavior

Session-level parquet reader options should reach the scan, so that `spark.comet.datafusion.execution.parquet.*` behaves as its implementation intends when `respectDataFusionConfigs` is enabled.

### Additional context

Found while investigating #4988.

**A naive fix is not correct.** Seeding the global options from the session config:

```rust
table_parquet_options.global = session_ctx
    .state()
    .config_options()
    .execution
    .parquet
    .clone();
```

makes the options reachable (verified: `max_predicate_cache_size=0` then genuinely disables the cache and the metrics disappear), but it also **regressed TPC-H SF100 Q6 with row-level pushdown enabled by ~51%, from 1.517 s to ~2.30 s** (median of last 5 of 10 iterations). Reverting and re-measuring returned it to 1.517 s, so the attribution is solid.

The cause is not understood. It is not `pushdown_filters`, which was already effectively enabled through the OR shown above, and it is not `reorder_filters`, which measures as a no-op on this query (2.302 s vs 2.303 s). Something else among the seeded defaults is responsible, and it should be identified before any fix lands, since whatever it is may also be costing performance in configurations where a user has legitimately set these options.

Measured on DataFusion 54.0.0, arrow/parquet 58.3.0, Spark 3.5.8, TPC-H SF100 local Parquet.

mbutrovich:

Closes #4990.

## Rationale for this change

Comet's native Parquet scan built a fresh `TableParquetOptions::new()` per scan and never looked at the session's `datafusion.execution.parquet.*` config. `pushdown_filters` appeared to work only because `ParquetSource::try_pushdown_filters` happens to OR the session value in independently; every other reader option (`reorder_filters`, `force_filter_selections`, `bloom_filter_on_read`, `max_predicate_cache_size`, `enable_page_index`, `pruning`) was silently inert, including when driven by `spark.comet.parquet.rowFilterPushdown.enabled`.

Fixing this also exposed a precedence bug: the `rowFilterPushdown.enabled` flag unconditionally overwrote the raw `spark.comet.datafusion.execution.parquet.{pushdown_filters,reorder_filters}` escape hatch (gated by `spark.comet.exec.respectDataFusionConfigs`), even when a developer set those keys explicitly for testing.

- `native/core/src/parquet/parquet_exec.rs`: `get_options()` now takes the session's `ParquetOptions` and seeds `table_parquet_options.global` for the fields confirmed to reach `ParquetSource`'s actual execution path (`pushdown_filters`, `reorder_filters`, `force_filter_selections`, `bloom_filter_on_read`, `max_predicate_cache_size`, `enable_page_index`, `pruning`). `coerce_int96`/`coerce_int96_tz` stay hardcoded afterward, non-overridable, for Spark INT96 compatibility. `session_ctx.state()` is now computed once in `init_datasource_exec` and reused, instead of being cloned twice.
- Deliberately not wired up: `schema_force_view_types`, `binary_as_string`, `skip_metadata`, and `global.metadata_size_hint`. All four are only consumed by `ParquetFormat::infer_schema`, which Comet's scan construction never calls, so seeding them would be a no-op.
- `native/core/src/execution/jni_api.rs`: `rowFilterPushdown.enabled` now only forces `pushdown_filters`/`reorder_filters` to `true` when the corresponding `spark.comet.datafusion.execution.parquet.*` key wasn't already set explicitly, so the escape hatch's explicit override wins.
- `spark/src/main/scala/org/apache/comet/CometConf.scala`: documented that `rowFilterPushdown.enabled`'s derived settings can be overridden via the escape hatch.

- New Rust unit test in `parquet_exec.rs` asserting `get_options()` carries the allowlisted fields through from the session while keeping the Spark-compat fields hardcoded.
- Two new Scala tests in `CometNativeReaderSuite`: one confirming row-level pushdown only fires at the native scan once `rowFilterPushdown.enabled` is set (the core #4990 regression), and one confirming an explicit `spark.comet.datafusion.execution.parquet.pushdown_filters=false` override wins over the flag.

### DIFF
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -576,6 +576,20 @@ fn prepare_datafusion_session_context(
             &ScalarValue::Float64(Some(1.1)),
         );
 
+    // Translate the Comet-namespaced row-level pushdown flag into the equivalent
+    // DataFusion session options. `pushdown_filters` enables the parquet reader's
+    // RowFilter evaluation during decode (late materialization); `reorder_filters`
+    // is only meaningful when pushdown_filters is on, so they move together. Set
+    // before the `spark.comet.datafusion.*` testing escape hatch pass-through below,
+    // so an explicit override of either key wins instead of being silently forced
+    // back to `true`.
+    if spark_config.get_bool(COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED) {
+        session_config =
+            session_config.set_str("datafusion.execution.parquet.pushdown_filters", "true");
+        session_config =
+            session_config.set_str("datafusion.execution.parquet.reorder_filters", "true");
+    }
+
     // Pass through DataFusion configs from Spark.
     // e.g: spark-shell --conf spark.comet.datafusion.sql_parser.parse_float_as_decimal=true
     // becomes datafusion.sql_parser.parse_float_as_decimal=true
@@ -584,27 +598,6 @@ fn prepare_datafusion_session_context(
         if let Some(df_key) = key.strip_prefix(SPARK_COMET_DF_PREFIX) {
             let df_key = format!("datafusion.{df_key}");
             session_config = session_config.set_str(&df_key, value);
-        }
-    }
-
-    // Translate the Comet-namespaced row-level pushdown flag into the equivalent
-    // DataFusion session options. `pushdown_filters` enables the parquet reader's
-    // RowFilter evaluation during decode (late materialization); `reorder_filters`
-    // is only meaningful when pushdown_filters is on, so they move together. Skip a
-    // key the `spark.comet.datafusion.*` testing escape hatch above already set
-    // explicitly, so an explicit override isn't silently forced back to `true`.
-    if spark_config.get_bool(COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED) {
-        const PUSHDOWN_FILTERS_KEY: &str =
-            "spark.comet.datafusion.execution.parquet.pushdown_filters";
-        const REORDER_FILTERS_KEY: &str =
-            "spark.comet.datafusion.execution.parquet.reorder_filters";
-        if !spark_config.contains_key(PUSHDOWN_FILTERS_KEY) {
-            session_config =
-                session_config.set_str("datafusion.execution.parquet.pushdown_filters", "true");
-        }
-        if !spark_config.contains_key(REORDER_FILTERS_KEY) {
-            session_config =
-                session_config.set_str("datafusion.execution.parquet.reorder_filters", "true");
         }
     }
 
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -589,11 +589,22 @@ fn prepare_datafusion_session_context(
     // Translate the Comet-namespaced row-level pushdown flag into the equivalent
     // DataFusion session options. `pushdown_filters` enables the parquet reader's
     // RowFilter evaluation during decode (late materialization); `reorder_filters`
-    // is only meaningful when pushdown_filters is on, so they move together.
+    // is only meaningful when pushdown_filters is on, so they move together. Skip a
+    // key the `spark.comet.datafusion.*` testing escape hatch above already set
+    // explicitly, so an explicit override isn't silently forced back to `true`.
     if spark_config.get_bool(COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED) {
-        session_config = session_config
-            .set_str("datafusion.execution.parquet.pushdown_filters", "true")
-            .set_str("datafusion.execution.parquet.reorder_filters", "true");
+        const PUSHDOWN_FILTERS_KEY: &str =
+            "spark.comet.datafusion.execution.parquet.pushdown_filters";
+        const REORDER_FILTERS_KEY: &str =
+            "spark.comet.datafusion.execution.parquet.reorder_filters";
+        if !spark_config.contains_key(PUSHDOWN_FILTERS_KEY) {
+            session_config =
+                session_config.set_str("datafusion.execution.parquet.pushdown_filters", "true");
+        }
+        if !spark_config.contains_key(REORDER_FILTERS_KEY) {
+            session_config =
+                session_config.set_str("datafusion.execution.parquet.reorder_filters", "true");
+        }
     }
 
     let runtime = rt_config.build()?;
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -270,6 +270,7 @@ fn op_name(op: &OpStruct) -> &'static str {
         OpStruct::CsvScan(_) => "CsvScan",
         OpStruct::ShuffleScan(_) => "ShuffleScan",
         OpStruct::BroadcastNestedLoopJoin(_) => "BroadcastNestedLoopJoin",
+        OpStruct::Sample(_) => "Sample",
     }
 }
 
--- a/native/core/src/execution/memory_pools/unified_pool.rs
+++ b/native/core/src/execution/memory_pools/unified_pool.rs
@@ -121,7 +121,7 @@ impl MemoryPool for CometUnifiedMemoryPool {
         }
         if let Err(prev) = self
             .used
-            .try_update(Relaxed, Relaxed, |old| old.checked_sub(size))
+            .fetch_update(Relaxed, Relaxed, |old| old.checked_sub(size))
         {
             panic!(
                 "Task {} overflow when releasing {size} of {prev} bytes",
@@ -149,7 +149,7 @@ impl MemoryPool for CometUnifiedMemoryPool {
             }
             if let Err(prev) = self
                 .used
-                .try_update(Relaxed, Relaxed, |old| old.checked_add(acquired as usize))
+                .fetch_update(Relaxed, Relaxed, |old| old.checked_add(acquired as usize))
             {
                 return Err(resources_datafusion_err!(
                     "Task {} failed to acquire {} bytes due to overflow. Reserved: {}",
--- a/native/core/src/execution/memory_pools/unified_pool.rs
+++ b/native/core/src/execution/memory_pools/unified_pool.rs
@@ -121,7 +121,7 @@ impl MemoryPool for CometUnifiedMemoryPool {
         }
         if let Err(prev) = self
             .used
-            .fetch_update(Relaxed, Relaxed, |old| old.checked_sub(size))
+            .try_update(Relaxed, Relaxed, |old| old.checked_sub(size))
         {
             panic!(
                 "Task {} overflow when releasing {size} of {prev} bytes",
@@ -149,7 +149,7 @@ impl MemoryPool for CometUnifiedMemoryPool {
             }
             if let Err(prev) = self
                 .used
-                .fetch_update(Relaxed, Relaxed, |old| old.checked_add(acquired as usize))
+                .try_update(Relaxed, Relaxed, |old| old.checked_add(acquired as usize))
             {
                 return Err(resources_datafusion_err!(
                     "Task {} failed to acquire {} bytes due to overflow. Reserved: {}",
--- a/native/core/src/parquet/parquet_exec.rs
+++ b/native/core/src/parquet/parquet_exec.rs
@@ -76,10 +76,11 @@ pub(crate) fn init_datasource_exec(
     use_field_id: bool,
     ignore_missing_field_id: bool,
 ) -> Result<Arc<DataSourceExec>, ExecutionError> {
-    // Computed once and reused below for `try_pushdown_filters`, since `SessionContext::state()`
-    // clones the whole `SessionState` (function registries, optimizer rules, etc.), not just the
-    // config.
-    let state = session_ctx.state();
+    // Computed once and reused below for `try_pushdown_filters`. `copied_config()` clones only
+    // `SessionConfig` (an `Arc<ConfigOptions>` plus a small extensions map); `SessionContext::
+    // state()` would clone the whole `SessionState`, including the function registries (~250+
+    // UDFs) and optimizer rules, on every scan init.
+    let session_config = session_ctx.copied_config();
     let (table_parquet_options, mut spark_parquet_options) = get_options(
         session_timezone,
         case_sensitive,
@@ -88,7 +89,7 @@ pub(crate) fn init_datasource_exec(
         allow_timestamp_ltz_to_ntz,
         &object_store_url,
         encryption_enabled,
-        &state.config_options().execution.parquet,
+        &session_config.options().execution.parquet,
     );
     spark_parquet_options.use_field_id = use_field_id;
     spark_parquet_options.ignore_missing_field_id = ignore_missing_field_id;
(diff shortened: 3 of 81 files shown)


native/core/src/execution/memory_pools/unified_pool.rs
```diff
@@ -121,7 +121,7 @@ impl MemoryPool for CometUnifiedMemoryPool {
         }
         if let Err(prev) = self
             .used
-            .fetch_update(Relaxed, Relaxed, |old| old.checked_sub(size))
+            .try_update(Relaxed, Relaxed, |old| old.checked_sub(size))
```
andygrove: `AtomicUsize::try_update` was apparently stabilized in Rust 1.95.0, and Comet MSRV is 1.88, so this may be an issue.

```
#[stable(feature = "atomic_try_update", since = "1.95.0")] 
```

andygrove: Also seems unrelated to the PR scope?

mbutrovich: I couldn't get clippy to pass locally, so I figured it was a toolchain upgrade somewhere.

mbutrovich: I can revert and as long as CI is green we're good.

native/core/src/execution/jni_api.rs
```diff
@@ -590,11 +590,22 @@ fn prepare_datafusion_session_context(
...
+    // explicitly, so an explicit override isn't silently forced back to `true`.
     if spark_config.get_bool(COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED) {
-        session_config = session_config
-            .set_str("datafusion.execution.parquet.pushdown_filters", "true")
-            .set_str("datafusion.execution.parquet.reorder_filters", "true");
+        const PUSHDOWN_FILTERS_KEY: &str =
+            "spark.comet.datafusion.execution.parquet.pushdown_filters";
+        const REORDER_FILTERS_KEY: &str =
+            "spark.comet.datafusion.execution.parquet.reorder_filters";
+        if !spark_config.contains_key(PUSHDOWN_FILTERS_KEY) {
```
andygrove: Nice catch on the precedence bug. Could this be handled by ordering instead? If the `rowFilterPushdown.enabled` translation moves above the `SPARK_COMET_DF_PREFIX` pass-through loop, then an explicitly-set `spark.comet.datafusion.execution.parquet.pushdown_filters` naturally overwrites the derived value, and you get the same behavior without the two hardcoded key constants or the `contains_key` checks.

It would also cover any future Comet flag that derives a DataFusion setting, rather than needing a new branch per key. As a bonus it removes the `reorder_filters` branch, which currently has no test.

native/core/src/parquet/parquet_exec.rs
```diff
@@ -76,6 +76,10 @@ pub(crate) fn init_datasource_exec(
     use_field_id: bool,
     ignore_missing_field_id: bool,
 ) -> Result<Arc<DataSourceExec>, ExecutionError> {
+    // Computed once and reused below for `try_pushdown_filters`, since `SessionContext::state()`
+    // clones the whole `SessionState` (function registries, optimizer rules, etc.), not just the
+    // config.
+    let state = session_ctx.state();
```
andygrove: Before this change `state()` was only reached when there were data filters. It is now on every scan, including filter-free ones.

`SessionState::clone` copies eight owned `HashMap`s, and `scalar_functions` holds every DataFusion builtin plus the roughly 250 UDFs registered in `prepare_datafusion_session_context`. Since `init_datasource_exec` runs once per native plan creation, that is once per Spark task.

Would `session_ctx.copied_config()` do here? It clones only the `SessionConfig`, and `.options()` gives you the `&ConfigOptions` that both `get_options` and `try_pushdown_filters` need.

native/core/src/parquet/parquet_exec.rs
```diff
@@ -220,8 +225,30 @@ fn get_options(
...
+
+    // Reader options that reach `ParquetSource`'s actual per-file execution path
+    // (`ParquetMorselizer`, built from `table_parquet_options.global`) rather than only
+    // `ParquetFormat::infer_schema`'s schema-inference path, which Comet never calls (Comet
+    // always supplies its own schema, translated from Spark's catalog). Seeded from the
+    // session so `spark.comet.datafusion.execution.parquet.*` (behind
+    // `spark.comet.exec.respectDataFusionConfigs`) and `spark.comet.parquet.
+    // rowFilterPushdown.enabled` actually take effect on the native scan (#4990); a fresh
+    // `TableParquetOptions::new()` ignored the session entirely.
+    table_parquet_options.global.pushdown_filters = session_parquet_options.pushdown_filters;
```
andygrove, line 247: The per-field verification behind this list is genuinely useful, thanks for spelling it out in the description. My concern is what happens six months from now. When DataFusion adds a reader option, nothing here fails, and the symptom is the same silent inertness that #4990 was about.

Would `table_parquet_options.global = session_parquet_options.clone()` work, keeping the `coerce_int96` / `coerce_int96_tz` assignments right after it? From what I can see the extra fields are all harmless. Everything from `data_pagesize_limit` down in `ParquetOptions` is writer-side and this scan never writes. The four `infer_schema` fields you called out are no-ops for the reason you gave. And `global.metadata_size_hint` is not read by `ParquetSource` at all, since it keeps its own `metadata_size_hint` field that the `with_metadata_size_hint(512 * 1024)` call above sets.

That way the Spark-compat overrides still win because they come last, and the allowlist cannot drift.

native/core/src/parquet/parquet_exec.rs
```diff
@@ -220,8 +225,30 @@ fn get_options(
...
+    table_parquet_options.global.reorder_filters = session_parquet_options.reorder_filters;
+    table_parquet_options.global.force_filter_selections =
+        session_parquet_options.force_filter_selections;
+    table_parquet_options.global.bloom_filter_on_read =
+        session_parquet_options.bloom_filter_on_read;
+    table_parquet_options.global.max_predicate_cache_size =
+        session_parquet_options.max_predicate_cache_size;
+    table_parquet_options.global.enable_page_index = session_parquet_options.enable_page_index;
```
andygrove, line 260: Worth a word in the comment about this one. `EagerPageIndexReaderFactory` forces `PageIndexPolicy::Optional` on every fetch and ignores the requested policy, so a session `enable_page_index=false` stops the page index from being used for pruning but does not stop it from being read.

Someone setting this to reduce metadata I/O would not get that. The comment above says these fields reach the real execution path, which is accurate but reads as a stronger guarantee than this particular field has.

spark/src/test/scala/org/apache/comet/exec/CometNativeReaderSuite.scala
```diff
@@ -829,6 +829,78 @@ class CometNativeReaderSuite extends CometTestBase with AdaptiveSparkPlanHelper
     }
   }
 
+  test("row-level pushdown reaches native scan when rowFilterPushdown.enabled is set") {
```
andygrove, line 832: Both new tests are good and they cover the #4990 regression directly.

The Rust unit test proves `get_options` copies the fields, though, which is a narrower claim than a user-set option actually changing what the reader does. Five of the seven newly-plumbed options have no coverage at either level.

It might be worth one more test that flips a non-pushdown option through the escape hatch and asserts observable behavior. `spark.comet.datafusion.execution.parquet.pruning=false` against the multi-row-group setup from `"row-group statistics pruning fires for native Parquet scan dataFilters"` would work, asserting `row_groups_pruned_statistics == 0`. That pins the plumbing at the level users care about, and it would have failed before this PR.

spark/src/main/scala/org/apache/comet/CometConf.scala
```diff
@@ -152,7 +152,10 @@ object CometConf extends ShimCometConf {
           "statistics, page index, bloom filters) is independent of this flag and runs " +
           "whenever Spark's spark.sql.parquet.filterPushdown is enabled. Disabling this " +
           "flag still lets format-level pruning work; the per-row eval falls back to " +
-          "the CometFilter operator above the scan.")
+          "the CometFilter operator above the scan. This flag's derived pushdown_filters " +
+          "/ reorder_filters settings can be overridden by explicitly setting " +
+          "spark.comet.datafusion.execution.parquet.pushdown_filters / reorder_filters " +
+          "when spark.comet.exec.respectDataFusionConfigs is enabled.")
```
andygrove: `COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED` is `CATEGORY_PARQUET`, so this text ends up in the published config reference, while `spark.comet.exec.respectDataFusionConfigs` is `CATEGORY_TESTING` and is described in its own doc as a development and testing option.

Does pointing users at it from a user-facing page read as an endorsement we do not want? The precedence behavior is worth writing down, so maybe the contributor guide is a better home for it.

andygrove (commented):
Thanks for digging into this. The diagnosis is right and the per-field verification in the description is really useful. I checked the DataFusion 54.1.0 source and all seven seeded fields really are read from `table_parquet_options.global` on the execution path, and the four you deliberately left out really are only touched by `ParquetFormat::infer_schema`. The precedence fix is sound too. `CometExecIterator.serializeCometSQLConfs` only forwards `spark.comet.datafusion.*` keys when `respectDataFusionConfigs` is on, so `contains_key` on the Rust side correctly implies the escape hatch is active.

I left comments on a few things. The main ones are that the precedence fix might be doable purely by reordering the two blocks, that the seven-field allowlist has no way to stay current as DataFusion adds reader options, and that hoisting `session_ctx.state()` adds a `SessionState` clone to the filter-free path that was not there before.

No docs action needed, by the way. `configs.md` is generated by `GenerateDocs` and `dev/ci/compute-changes.py` already routes `CometConf.scala` changes into the docs job.

CI is clean so far, 15 passing and nothing failing. `rust-test` passed so the new Rust unit test is green, but the shards that run `CometNativeReaderSuite` are still pending.

native/core/src/execution/jni_api.rs
mbutrovich: I considered changing the ordering here but was afraid of the blast radius a day before cutting the 1.0 branch.

native/core/src/parquet/parquet_exec.rs
mbutrovich: Went through every field in `ParquetOptions` against actual usage in datasource-parquet's `source.rs` and `opener/mod.rs`. Confirmed `schema_force_view_types`, `binary_as_string`, `skip_metadata`, and `global.metadata_size_hint` are only read from `file_format.rs` (the `infer_schema` path), and everything under the writer section (`data_pagesize_limit`, `write_batch_size`, `writer_version`, `statistics_enabled`, `compression`, `dictionary_*`, `max_row_group_size`) is writer-only. So `table_parquet_options.global = session_parquet_options.clone()` would be safe for every field that exists in `ParquetOptions` today.

I'd rather keep the explicit allowlist, though. If DataFusion adds a new reader option to `ParquetOptions.global`, the wholesale copy makes it silently active on every Comet scan the next time we bump the DataFusion dependency, with no Comet-side check on whether that option is safe for Spark semantics. `coerce_int96`/`coerce_int96_tz` are exactly this case: a reader option that cannot be passed through as-is. I'd rather a new field come in as silently ignored, forcing a conscious decision (and a test) before it reaches the scan, than silently active with no gate.

I'll add a comment above the allowlist flagging that it should be checked against `ParquetOptions` whenever we bump the DataFusion version, so it doesn't drift by accident. Applied the rest of your comments (the reordering fix, `copied_config()`, the page-index comment, and the `pruning=false` test).

spark/src/main/scala/org/apache/comet/CometConf.scala
mbutrovich: Should be addressed in https://github.com/apache/datafusion-comet/pull/5107/commits/08017243f08f488f62b82402a3763388046956c0, thanks @andygrove!

spark/src/test/scala/org/apache/comet/exec/CometNativeReaderSuite.scala
mbutrovich: Should be addressed in https://github.com/apache/datafusion-comet/pull/5107/commits/08017243f08f488f62b82402a3763388046956c0, thanks @andygrove!

native/core/src/parquet/parquet_exec.rs
mbutrovich: Should be addressed in https://github.com/apache/datafusion-comet/pull/5107/commits/08017243f08f488f62b82402a3763388046956c0, thanks @andygrove!

native/core/src/parquet/parquet_exec.rs
mbutrovich: Should be addressed in https://github.com/apache/datafusion-comet/pull/5107/commits/08017243f08f488f62b82402a3763388046956c0, thanks @andygrove!

native/core/src/execution/jni_api.rs
mbutrovich: Should be addressed in https://github.com/apache/datafusion-comet/pull/5107/commits/08017243f08f488f62b82402a3763388046956c0, thanks @andygrove!

andygrove (approved):
Thanks @mbutrovich. LGTM, pending CI.

PR #5108, opened 2026-07-28
fix: resolve Comet jar without hardcoding the SNAPSHOT qualifier

andygrove:

N/A

## Rationale for this change

The PyArrow UDF test harness locates the built Comet jar with a glob that hardcodes the `-SNAPSHOT` qualifier:

```
spark/target/comet-spark-{spark_tag}{scala_tag}-*-SNAPSHOT.jar
```

On a release branch the Maven version is the final release version with no `-SNAPSHOT` qualifier, so the glob matches nothing and every PyArrow UDF test errors out with `FileNotFoundError: Comet jar not found`. This is what happened on the 1.0.0 release branch in #5105, where the PyArrow UDF jobs were the only failing checks. The release branch runs the same workflows as `main`, so this class of breakage only surfaces after the release branch is cut and the version is bumped.

Two benchmark helpers have the same problem: the TPC docker build copies `comet-spark-spark3.5_2.12-*-SNAPSHOT.jar` out of the container, and `run_all_benchmarks.sh` defaults `COMET_JAR` to a hardcoded `0.14.0-SNAPSHOT` path that is already stale.

- `spark/src/test/resources/pyspark/conftest.py`: match any version suffix rather than only `-SNAPSHOT`, and widen the artifact filter so sources/tests/javadoc/shaded jars are still excluded.
- `benchmarks/tpc/infra/docker/Dockerfile.build-comet`: drop `-SNAPSHOT` from the jar copy glob.
- `benchmarks/pyspark/run_all_benchmarks.sh`: resolve the default `COMET_JAR` by glob instead of a hardcoded version, and fail with a clear message when no jar is found.
- `docs/source/contributor-guide/release_process.md`: enumerate the files that carry the Maven version (including the Gradle version catalog entry in the Iceberg diffs, which uses a different syntax than the Maven property in the Spark diffs), add a `git grep` verification step after the bump, and note that jar globs must not assume a `-SNAPSHOT` suffix.

Existing CI. The PyArrow UDF workflow triggers on `conftest.py`, so it exercises the new resolution logic on `main`, where the version still carries `-SNAPSHOT`. The non-SNAPSHOT path is verified on the 1.0.0 release branch, where the same `conftest.py` change turns the previously failing PyArrow UDF jobs green in #5105.

### DIFF
--- a/benchmarks/pyspark/run_all_benchmarks.sh
+++ b/benchmarks/pyspark/run_all_benchmarks.sh
@@ -25,10 +25,17 @@ set -e
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 DATA_PATH="${1:-/tmp/shuffle-benchmark-data}"
-COMET_JAR="${COMET_JAR:-$SCRIPT_DIR/../../spark/target/comet-spark-spark3.5_2.12-0.14.0-SNAPSHOT.jar}"
+# Resolve the built jar without hardcoding the version, which changes every
+# development cycle and drops the `-SNAPSHOT` qualifier on release branches.
+COMET_JAR="${COMET_JAR:-$(ls "$SCRIPT_DIR"/../../spark/target/comet-spark-spark3.5_2.12-*.jar 2>/dev/null | grep -v -e sources -e tests | tail -1)}"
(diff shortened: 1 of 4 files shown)


mbutrovich (approved):
LGTM, thanks for updating the release process docs!

Issue #1906, opened 2025-06-18
RangePartitioning does not yield correct results with native shuffle


mbutrovich:
### Describe the bug

As noted by @Kontinuation, the RangePartitioning implementation that just got merged has a bug, most obvious when trying to RangePartition pre-sorted data:

> This implementation of RangePartitioning may be incorrect. RangePartitioning should partition the input DataFrame into partitions with consecutive and non-overlapping ranges, this requires scanning the entire DataFrame to obtain the ranges of each partition before performing the actual shuffle writing.
> 
> Here is the PySpark code to illustrate the difference between the behavior of Comet and Vanilla Spark.
> 
> ```python
> spark.range(0, 100000).write.format("parquet").mode("overwrite").save("range-partitioning")
> 
> df = spark.read.parquet("range-partitioning")
> df_range_partitioned = df.repartitionByRange(10, "id")
> 
> df_range_partitioned.explain()
> 
> # Show the min and max of each range
> def get_partition_bounds(idx, iterator):
>     min = None
>     max = None
>     for row in iterator:
>         if min is None or row.id < min:
>             min = row.id
>         if max is None or row.id > max:
>             max = row.id
>     yield idx, min, max
> 
> partition_bounds = df_range_partitioned.rdd.mapPartitionsWithIndex(get_partition_bounds).collect()
> 
> # Print the results
> for partition_id, min_id, max_id in sorted(partition_bounds):
>     print(f"Partition {partition_id}: min_id={min_id}, max_id={max_id}")
> ```
> 
> **Comet**:
> 
> ```
> == Physical Plan ==
> AdaptiveSparkPlan isFinalPlan=false
> +- CometExchange rangepartitioning(id#17L ASC NULLS FIRST, 10), REPARTITION_BY_NUM, CometNativeShuffle, [plan_id=173]
>    +- CometScan parquet [id#17L] Batched: true, DataFilters: [], Format: CometParquet, Location: InMemoryFileIndex(1 paths)[file:/path/to/range-p..., PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
> 
> Partition 0: min_id=0, max_id=90799
> Partition 1: min_id=753, max_id=91680
> Partition 2: min_id=1527, max_id=92520
> Partition 3: min_id=2399, max_id=93284
> Partition 4: min_id=3274, max_id=94123
> Partition 5: min_id=4053, max_id=94844
> Partition 6: min_id=4851, max_id=95671
[...]

Issue #5109, opened 2026-07-28
Add support for SampleExec (DataFrame.sample, TABLESAMPLE, randomSplit)


andygrove:
## What is the problem the feature request solves?

`SampleExec` is not supported, so any query using `DataFrame.sample`, SQL `TABLESAMPLE`, or `DataFrame.randomSplit` falls back to Spark and breaks up the native execution block.

Spark lowers all three of these to a single physical operator:

```scala
case class SampleExec(lowerBound: Double, upperBound: Double,
                      withReplacement: Boolean, seed: Long, child: SparkPlan)
```

There are two execution paths (`sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala`):

**Without replacement** (the common case; `df.sample(0.1)` gives `lowerBound=0.0, upperBound=0.1`): `BernoulliCellSampler(lb, ub)` seeded with `seed + partitionIndex`, drawing one `XORShiftRandom.nextDouble()` per input row and keeping the row when `x >= lb && x < ub`.

**With replacement**: `PoissonSampler(ub - lb, useGapSamplingIfPossible = false)`, which is commons-math3 `PoissonDistribution.sample()` per row (emitting each row 0..n times), reseeded per partition from `new java.util.Random(seed).nextLong()` drawn `partitionIndex + 1` times.

`SampleExec` is unchanged across Spark 3.4 through 4.2, so no shims are needed.

## Describe the potential solution

Add `SampleExec` as a fully native operator (`nativeExecs`), per `docs/source/contributor-guide/adding_a_new_operator.md`:

- `native/proto/src/proto/operator.proto`: new `Sample` message (`lower_bound`, `upper_bound`, `with_replacement`, `seed`) plus a slot in `oneof op_struct`
- `spark/src/main/scala/org/apache/comet/serde/operator/CometSampleExec.scala`: `CometOperatorSerde[SampleExec]` and a `CometSampleExec` case class extending `CometUnaryExec`, passing through `outputPartitioning` and `outputOrdering`
- `spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala`: register in `nativeExecs`
- `common/src/main/scala/org/apache/comet/CometConf.scala`: `spark.comet.exec.sample.enabled`
- `native/core/src/execution/operators/sample.rs`: the `ExecutionPlan` implementation
- `native/core/src/execution/planner.rs`: match arm for the new `OpStruct`

### RNG fidelity

[...]

PR #2258, opened 2025-08-28
fix: RangePartitioning boundaries with native shuffle

mbutrovich:
## Which issue does this PR close?

Closes #1906.

## Rationale for this change

#1862 tried to implement RangePartitioning with native shuffle. The implementation didn't work because executors calculated their own partition boundaries.

## What changes are included in this PR?

This modifies the flow for the driver to calculate the boundaries (like Spark). At a high level:
- Hoist code from Spark's `ShuffleExchangeExec` for using Spark's `RangePartitioner` to calculate boundary rows.
- Serialize boundary rows to native side.
- Deserialize boundary rows and pass as part of the partitioning scheme. Each executor should have the boundary values now.

## How are these changes tested?

- New test implementing #1906.

## Remaining concerns

- What is the performance implication of iterating over these batches to calculate the bounds at the driver? Are we introducing a bunch of overhead?
- Should we remove the code for calculating boundary rows in native code? It's possible we (or someone else) might want this, so I left the tests and annotated them to allow dead code.
- I change the default to enable this feature, but we can discuss setting it to false. For now I want to see how it does in CI.
- I need to generate new plans.

PR #1862, opened 2025-06-07
feat: support RangePartitioning with native shuffle

mbutrovich:
## Which issue does this PR close?

Closes #458.

## What changes are included in this PR?

- New configs for hash and range partitioning with native shuffle. Range partitioning, due to its random sampling, can yield differently-ordered query results. It's unclear that this would matter in production, but this makes it easier to make Spark SQL tests work without falling back to Spark entirely
- New diffs to turn range partitioning with native shuffle off for tests that expect specific partitioning (bucket scan-related tests)
- Range partitioning added to shuffle_writer benchmark
- Remove use of DF's `Partitioning` since Comet's supported schemes don't match. There's a new `CometPartitioning` enum
- Moved some repeated shuffle_writer logic to individual functions
- RangePartitioner class that performs random sampling of input batch and determines partitioning bounds

## How are these changes tested?

- Native tests for RangePartitioner
- New golden plans that have CometExchange where there used to be CometColumnarExchange
- Updated fuzz and native shuffle tests

PR #5110, opened 2026-07-28
feat: support SampleExec natively for sampling without replacement

andygrove:

Part of #5109.

## Rationale for this change

`SampleExec` is not supported today, so any query using `DataFrame.sample`, SQL `TABLESAMPLE`, or `DataFrame.randomSplit` falls back to Spark and breaks up the surrounding native execution block.

Spark has two sampling paths. Without replacement, it applies `BernoulliCellSampler` to each partition, seeded with `seed + partitionIndex`, drawing one `XORShiftRandom.nextDouble()` per input row and keeping the row when the value falls in `[lowerBound, upperBound)`. With replacement, it draws from a commons-math3 `PoissonDistribution` seeded by `Well19937c`.

This PR covers the first path, which is the common case. Comet already has a bit-exact port of `XORShiftRandom` (used by `rand` / `randn` / `shuffle`), including the murmur3 `hashSeed` that `setSeed` applies, so reproducing Spark's exact draw sequence is a small addition rather than a new RNG port. The with-replacement path needs ports of `Well19937c` and `PoissonDistribution.sample()` and is left as follow-up work, tracked by #5109.

- New `Sample` protobuf message and `SampleExec` native operator (`native/core/src/execution/operators/sample.rs`), which filters each batch by the sampler's decisions and carries generator state across batch boundaries.
- New `BernoulliCellSampler` in the `spark-expr` crate, a port of Spark's sampler built on the existing `XorShiftRandom`.
- `CometSampleExec` operator serde, registered in `CometExecRule.nativeExecs`, gated by the new `spark.comet.exec.sample.enabled` config (default true). `getSupportLevel` returns `Unsupported` when `withReplacement` is set, so that case falls back to Spark.
- The planner applies the partition index to the seed the same way the `rand` / `shuffle` expression builders do.
- Documentation: a Sampling section in the operator compatibility guide covering the row-order caveat below, and a `SampleExec` row in the operator support reference.

Because the sampler consumes one random value per row within a partition, it matches Spark's selection only when rows reach it in the same order. That holds when sampling above a scan, filter, or projection. Above an operator where Comet may emit rows in a different order than Spark, such as a join or an aggregate, the result is still a valid sample of the same expected size but not necessarily the same rows. This is inherent to per-row RNG consumption and already applies to `rand()`; it is documented in the compatibility guide.

New Rust unit tests cover that exactly one value is drawn per row, that complementary ranges partition the input, that an empty range selects nothing, and that splitting the input into multiple batches does not change which rows are selected.

New tests in `CometExecSuite` compare results against Spark for `sample` with a fixed seed, for a multi-partition input (which is what catches a native side that ignores the partition index), and for `randomSplit` (non-zero lower bound), and assert that sampling with replacement and the disabled config both fall back to Spark.

The result-comparison tests were verified to be non-vacuous by temporarily perturbing the native seed by one, which fails all three comparison tests. The suite passes on both the default profile and `-Pspark-3.4`.

### DIFF
--- a/docs/source/user-guide/latest/compatibility/operators.md
+++ b/docs/source/user-guide/latest/compatibility/operators.md
@@ -26,11 +26,10 @@ Comet runs `SampleExec` natively when sampling is performed without replacement,
 reproduces Spark's per-row `XORShiftRandom` draw sequence, so for a given seed it selects the same
 rows as Spark.
 
-Because the sampler consumes one random value per row within a partition, it selects the same rows
-as Spark only when the rows reach it in the same order. Sampling directly above a scan, filter, or
-projection meets that condition. Sampling above an operator where Comet may emit rows in a
-different order than Spark, such as a join or an aggregate, still returns a valid sample of the
-same expected size, but not necessarily the same rows Spark would have selected.
+Because the sampler consumes one random value per row, sampling directly above a scan, filter, or
+projection reproduces Spark's selection. Above an operator where Comet may emit rows in a different
+order than Spark, such as a join or an aggregate, the result is still a valid sample of the same
+expected size, but not necessarily the same rows.
 
 Sampling with replacement (`df.sample(withReplacement = true, ...)`) falls back to Spark, because
 it draws from a Poisson distribution that Comet does not implement natively
--- a/docs/source/user-guide/latest/compatibility/operators.md
+++ b/docs/source/user-guide/latest/compatibility/operators.md
@@ -19,6 +19,23 @@ under the License.
 
 # Operator Compatibility
 
+## Sampling
+
+Comet runs `SampleExec` natively when sampling is performed without replacement, which covers
+`DataFrame.sample`, SQL `TABLESAMPLE`, and `DataFrame.randomSplit`. The native implementation
+reproduces Spark's per-row `XORShiftRandom` draw sequence, so for a given seed it selects the same
+rows as Spark.
+
+Because the sampler consumes one random value per row within a partition, it selects the same rows
+as Spark only when the rows reach it in the same order. Sampling directly above a scan, filter, or
+projection meets that condition. Sampling above an operator where Comet may emit rows in a
+different order than Spark, such as a join or an aggregate, still returns a valid sample of the
+same expected size, but not necessarily the same rows Spark would have selected.
+
+Sampling with replacement (`df.sample(withReplacement = true, ...)`) falls back to Spark, because
+it draws from a Poisson distribution that Comet does not implement natively
+([#5109](https://github.com/apache/datafusion-comet/issues/5109)).
+
 ## Window Functions
 
 Comet runs `WindowExec` natively and it is enabled by default (`spark.comet.exec.window.enabled`). A broad set of
--- a/spark/src/test/scala/org/apache/comet/exec/CometExecSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/exec/CometExecSuite.scala
@@ -97,14 +97,9 @@ class CometExecSuite extends CometTestBase {
   // The sampler is seeded per partition with `seed + partitionIndex`, so a multi-partition input
   // is what catches a native side that ignores the partition index.
   test("sample without replacement over multiple partitions") {
-    withTempDir { dir =>
-      val path = new Path(dir.toURI.toString, "sample.parquet")
-      spark
-        .range(0, 4000)
-        .repartition(4)
-        .write
-        .parquet(path.toString)
-      withParquetTable(path.toString, "tbl") {
+    withTempPath { dir =>
+      spark.range(0, 4000).repartition(4).write.parquet(dir.getCanonicalPath)
+      withParquetTable(dir.getCanonicalPath, "tbl") {
         val df =
           sql("SELECT * FROM tbl").sample(withReplacement = false, fraction = 0.2, seed = 7)
         assert(df.rdd.getNumPartitions > 1)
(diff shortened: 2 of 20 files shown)


docs/source/user-guide/latest/compatibility/operators.md
```diff
@@ -19,6 +19,23 @@ under the License.
...
+rows as Spark.
+
+Because the sampler consumes one random value per row within a partition, it selects the same rows
+as Spark only when the rows reach it in the same order. Sampling directly above a scan, filter, or
+projection meets that condition. Sampling above an operator where Comet may emit rows in a
+different order than Spark, such as a join or an aggregate, still returns a valid sample of the
+same expected size, but not necessarily the same rows Spark would have selected.
+
+Sampling with replacement (`df.sample(withReplacement = true, ...)`) falls back to Spark, because
+it draws from a Poisson distribution that Comet does not implement natively
```
mbutrovich, line 35: For the record, I think I did implement this natively in my early drafts of range partitioning where we did the sampling in native code. It's in that commit history somewhere if we ever want to resurrect it.

spark/src/test/scala/org/apache/comet/exec/CometExecSuite.scala
```diff
@@ -87,6 +87,51 @@ class CometExecSuite extends CometTestBase {
     }
   }
 
+  test("sample without replacement") {
```
parthchandra, line 90: Should we add a SQL test with TABLESAMPLE as well?

parthchandra (commented):
Would it be useful to add a benchmark comparing with Spark?

parthchandra (approved):
My review comments are nice to have rather than required. 
Approving (pending CI)

comphead (approved):
Thanks @andygrove I checked it vs Spark and test coverage, looks solid.

We may want to add some tests though

```
 - Ordering claim untested (CometSampleExec.scala:66, low). The docs and code claim ordering is preserved above scan/filter/project. Add one test: spark.range(N).orderBy('id).sample(false, f, seed) and
  assert monotonicity plus row-for-row match with Spark. Guards against a future refactor that re-batches.
 - No test for Sample above non-scan children (CometExecSuite.scala, low). Docs at compatibility/operators.md:26-33 explicitly caveat that above join/aggregate, only expected size is preserved. Add a
  size-tolerance test above a hash-aggregate or join to lock in the "valid sample of the same expected size" claim.
  - Empty batch interleave in Rust (sample.rs:234, nit). Neither Rust test feeds a 0-row batch alongside non-zero batches. Cheap regression guard: alternate an empty batch with a non-empty batch and assert
  output equals the sample of the concatenated non-empty batches.
```

docs/source/user-guide/latest/compatibility/operators.md
kazuyukitanimura: Let's find the commit and document

andygrove:
The latest commit expands test coverage: ordering preservation when sampling above a sort, sampling above an aggregate, SQL `TABLESAMPLE`, and empty batch interleaving in the native operator. It also adds a `Sample` case to `CometExecBenchmark` comparing against Spark.

Benchmark results (Apple M3 Ultra, OpenJDK 17, Spark 4.1 profile, release build, 10M rows):

```
OpenJDK 64-Bit Server VM 17.0.17+10 on Mac OS X 26.3.1
Apple M3 Ultra
Sample Exec (fraction 0.05):              Best Time(ms)   Avg Time(ms)   Stdev(ms)    Rate(M/s)   Per Row(ns)   Relative
------------------------------------------------------------------------------------------------------------------------
SQL Parquet - Spark                                  35             38           2        303.9           3.3       1.0X
SQL Parquet - Comet                                  22             24           2        478.9           2.1       1.6X

Sample Exec (fraction 0.5):               Best Time(ms)   Avg Time(ms)   Stdev(ms)    Rate(M/s)   Per Row(ns)   Relative
------------------------------------------------------------------------------------------------------------------------
SQL Parquet - Spark                                  62             63           1        168.9           5.9       1.0X
SQL Parquet - Comet                                  36             38           2        289.1           3.5       1.7X

Sample Exec (fraction 0.95):              Best Time(ms)   Avg Time(ms)   Stdev(ms)    Rate(M/s)   Per Row(ns)   Relative
------------------------------------------------------------------------------------------------------------------------
SQL Parquet - Spark                                  48             49           2        220.5           4.5       1.0X
SQL Parquet - Comet                                  49             51           2        215.8           4.6       1.0X
```

Comet is roughly 1.6x to 1.7x faster at low and medium fractions, and on par at fraction 0.95, where nearly every row passes and the cost is dominated by scan and row copying rather than the sampling filter.

docs/source/user-guide/latest/compatibility/operators.md
andygrove: Found it. The native sampling was in `native/core/src/execution/shuffle/range_partitioner.rs`:

- Added by 3aa3dc707821ea7620c50658eb22af90b6b0edb7 (#1862, "feat: support RangePartitioning with native shuffle").
- Removed by 25d5924a72cfbe9f5effeda11f7f27eb09ec26eb (#2258, "fix: distributed RangePartitioning bounds calculation with native shuffle", closes #1906), because bounds were being computed per-executor and so were wrong in a distributed setting. That PR moved bounds calculation to the driver using Spark's own `RangePartitioner` and serialized the boundary rows to native.

To recover the implementation plus its tests (474 lines):

```
git show 25d5924a7^:native/core/src/execution/shuffle/range_partitioner.rs
```

Two caveats on how much this helps the with-replacement path, though:

1. It is reservoir sampling (Algorithm L) for estimating range-partition bounds, i.e. sampling *without* replacement. That is a different problem from what `SampleExec` needs, which is a per-row Bernoulli or Poisson decision.
2. It seeded Rust's `rand::SmallRng` rather than porting Spark's RNG, so it produced a statistically valid sample but not Spark's exact draw sequence. That was fine for bound estimation, where agreement with Spark does not matter, but the `SampleExec` path needs exact agreement.

So it is good prior art for having done sampling in native code, but it does not give us the `Well19937c` and `PoissonDistribution.sample()` ports that the with-replacement case still needs. I have noted the pointer on #5109 so the history is findable when we pick that up.

Issue #5112, opened 2026-07-28
Native columnar-to-row conversion is much slower than the JVM implementation for small batches


andygrove:
### Describe the issue

`CometNativeColumnarToRowExec` (enabled by default) can be much slower than the JVM `CometColumnarToRowExec`, but none of our existing benchmarks can see it. Measuring the conversion in isolation (no parquet scan) shows the native converter is 3.7x slower per row at the default 8192-row batch size, and the gap grows quickly as batches get smaller because the native path carries a large fixed cost per batch.

Isolated microbenchmark (in-memory Arrow batches, 4 columns: long, int, double, string; conversion plus value consumption only; Apple M3 Ultra, Spark 4.1):

| batch size | JVM rowIterator + UnsafeProjection | native converter | native/JVM |
|---|---|---|---|
| 8192 | 8.8 ns/row | 32.6 ns/row | 3.7x slower |
| 512 | 9.1 ns/row | 53.7 ns/row | 5.9x slower |
| 32 | 19.5 ns/row | 306.6 ns/row | 15.7x slower |

Solving the three data points for a fixed-plus-linear cost model gives roughly 10us of fixed overhead per batch for the native path versus roughly 0.2us for the JVM path. Note that the JVM baseline here is the interpreted iterator path; the production JVM operator is `CodegenSupport` and fuses into whole-stage codegen, so it is cheaper still.

### Where the native cost goes

1. Per-batch fixed overhead (~10us, the dominant real-world problem): every batch pays Arrow FFI export on the JVM side (`allocateArrowStructs` plus `Data.exportVector` per column, re-exporting the constant schema every time), the JNI crossing, `from_ffi` import per column in Rust, and on the return path `FindClass` for `NativeColumnarToRowInfo`, object construction, and two freshly allocated `int[]` arrays. `from_ffi` uses `ArrayData::new_unchecked`, so validation is not a factor; it is the number of per-batch allocations and crossings.
2. Per-row cost (3.7x even at the ideal batch size): for any schema containing a var-length column the Rust side takes the general path (per-row buffer resize and zero, two passes over the columns, enum dispatch per column per row), and every row then crosses back through `NativeRowIterator.next()`.
3. The per-row `unsafeRow.copy()` added in #3367 to fix the #3308 SIGSEGV costs only ~4 ns/row on narrow rows, so it is secondary, but it grows with row width, allocates twice per row, and permanently negates the zero-copy premise described in the operator's docs.

Two aggravating factors:

[...]

PR #5113, opened 2026-07-28
test: add isolated and small-batch benchmarks for columnar-to-row conversion

andygrove:

Part of #5112.

## Rationale for this change

#5112 shows that `CometNativeColumnarToRowExec` is 3.7x slower per row than the JVM implementation at the default batch size and up to 15.7x slower on small batches, yet `CometColumnarToRowBenchmark` reports parity in every group. The existing benchmark measures end-to-end parquet queries with a `noop()` sink, so the scan cost (70 to 530 ns/row) swamps the conversion cost and every batch arrives at the full 8192 rows, which amortizes the native path's roughly 10us of fixed per-batch overhead. This PR adds benchmarks that can actually observe conversion cost so that regressions in this class are visible and future improvements can be measured.

- New `CometC2RIsolatedBench`: measures columnar-to-row conversion in isolation on in-memory Arrow batches, comparing the JVM `rowIterator` plus `UnsafeProjection` path against `NativeColumnarToRowConverter` at batch sizes 8192, 512, and 32. Numbers from an Apple M3 Ultra are in #5112.
- Two new groups in `CometColumnarToRowBenchmark`:
  - dictionary-encoded parquet strings (the session default disables dictionary encoding, but real-world parquet data is usually dictionary-encoded)
  - a Scala UDF consumer, which defeats escape analysis and exercises the whole-stage codegen fusion asymmetry between the `CodegenSupport` JVM operator and the iterator-based native operator
- Disable ANSI mode in the benchmark session: on Spark 4.x defaults the data generation queries fail with `CAST_OVERFLOW` on `cast(id as byte)` at id=128, so the benchmark could not run at all against Spark 4.x before this change.

These are benchmark additions, exercised by running them:

```
make benchmark-org.apache.spark.sql.benchmark.CometColumnarToRowBenchmark
```

and `CometC2RIsolatedBench` via `exec:java` on the test classpath. Result tables from both are posted in #5112.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/benchmark/CometC2RIsolatedBench.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/benchmark/CometC2RIsolatedBench.scala
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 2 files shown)


comphead (approved):
Thanks @andygrove for the extra benchmarks

Issue #4440, opened 2026-05-26
Revisit the case for native columnar-to-row?


mbutrovich:
Reading through PR #3221, PR #3266, PR #3649, and epic #3268, I want to surface some questions about whether the rationale for `CometNativeColumnarToRowExec` still holds. I might be missing context, so this is a question rather than a proposal.

### Background

PR #3221 (landing) was upfront that there was no significant throughput win:

> Although this native implementation does not provide significant speed improvements, the main benefit is reduced GC pressure by avoiding Java allocations.

The benchmark table in that PR showed native winning 1 of 8 scenarios, losing 3, tying 4.

PR #3649 (enable by default) retracted its ~10% TPC-H claim as misattributed to native scan, leaving "no regression + GC pressure" as the standing rationale.

I could not find a GC measurement in any of the PR threads, in #3268, in the operator's SQLMetrics, or in `spark/benchmarks/`. The GC pressure benefit appears to be theoretical, not measured.

### A quick re-run on main (Spark 4.1, 1M rows, fixedWidthOnly)

```
Spark (ColumnarToRowExec)                     68 ms
Comet JVM (CometColumnarToRowExec)            65 ms
Comet Native (CometNativeColumnarToRowExec)   64 ms
```

All within ~5%, roughly consistent with the numbers in PR #3221.

### A JFR pass at allocations

Running the same benchmark with `-XX:StartFlightRecording=settings=profile`, I expected to see per-row allocations from `unsafeRow.copy()` in `NativeColumnarToRowConverter.scala:142`. Out of 2,809 `byte[]` allocation samples, zero appeared in any C2R operator's stack. All came from the parquet scan and broadcast/serializer paths shared across implementations.

The most likely explanation is that under the `.noop()` sink the benchmark uses, HotSpot escape analysis elides Native's per-row `.copy()` (and any equivalent allocations on the JVM path). If that is right, this benchmark setup cannot validate the GC pressure claim either way, since the rows never escape the iteration.

### Questions at address

1. Is there a measurement of native C2R's GC benefit that I am overlooking? A pointer to a concrete workload with JFR or `jdk.GarbageCollection` evidence would help me understand the current value.

2. Would adding a retain-mode benchmark variant to `CometColumnarToRowBenchmark` (e.g., `collect`, hash-agg, broadcast build, anything that defeats EA) be worth it, so the GC claim becomes testable?

[...]

PR #3367, opened 2026-02-02
fix: fix memory safety issue in native c2r

andygrove:
## Which issue does this PR close?

Closes #3308

## Rationale for this change

`NativeRowIterator` reuses a single `UnsafeRow` object that points to native memory. The buffer is cleared on each convert() call, so any held row references become stale. Return `unsafeRow.copy()` to copy each row to Java heap memory.

Issue #3308, opened 2026-01-29
Native columnar-to-row: UnsafeRow may reference freed/overwritten native memory


andygrove:
## Summary

The native columnar-to-row conversion (`CometNativeColumnarToRowExec`) returns `UnsafeRow` objects that point directly into a Rust-owned buffer via raw pointer. This buffer is cleared and reused on each `convert()` call, which can cause SIGSEGV if any row reference outlives the current batch.

## Root Cause

In `NativeColumnarToRowConverter.scala`, the `NativeRowIterator` does:

```scala
unsafeRow.pointTo(null, rowAddress, rowSize)
```

This points the `UnsafeRow` at off-heap native memory owned by `ColumnarToRowContext.buffer` on the Rust side. The Rust `convert()` method calls `self.buffer.clear()` at the start of each invocation, invalidating all previous row pointers.

### Broadcast path (high risk)

In `CometNativeColumnarToRowExec.scala` (line ~110-130), multiple batches are lazily converted via `flatMap` and passed to `mode.transform(rows, ...)`. When the second batch's `convert()` runs, any `UnsafeRow` from the first batch that hasn't been copied yet points to overwritten or freed memory. Whether this is safe depends on `mode.transform()` eagerly copying each row before the iterator advances — a fragile assumption.

### Streaming path (lower risk)

The `doExecute()` path uses `flatMap` which naturally consumes one batch's iterator before requesting the next, so rows don't span batches. However, downstream Spark operators that retain `InternalRow` references (e.g., sort with spill, certain joins, caching) could still hold stale pointers.

## Symptoms

- SIGSEGV in CI (intermittent, timing-dependent)
- Likely became more visible after enabling native columnar-to-row by default (#3299)

## Suggested Fix

Return `unsafeRow.copy()` from `NativeRowIterator.next()` to copy each row to Java heap memory. This eliminates dangling pointer risk while preserving the performance benefit of native conversion (the copy is just a `memcpy` of already-serialized row bytes).

A more targeted fix would be to copy only in the broadcast path, or to change the Rust side to allocate a new buffer per batch instead of reusing.

## Related

- #3299 (feat: Enable native columnar-to-row by default)
- `native/core/src/execution/columnar_to_row.rs`
- `spark/src/main/scala/org/apache/comet/NativeColumnarToRowConverter.scala`
- `spark/src/main/scala/org/apache/spark/sql/comet/CometNativeColumnarToRowExec.scala`

PR #3299, opened 2026-01-27
feat: Enable native columnar-to-row by default

andygrove:
## Which issue does this PR close?

Closes https://github.com/apache/datafusion-comet/issues/3300

## Rationale for this change

Enable by default for better performance and reduced GC for JVM shuffle where the child plan is a Comet plan.

## What changes are included in this PR?

- [x] Enable config by default
- [x] Disable config in stability suite (so that I didn't have to update golden files and make this PR huge). I will do this as a separate PR
- [x] Update diffs for some Spark SQL tests because the plan is different now

PR #5114, opened 2026-07-28
feat: disable native columnar-to-row conversion by default

### LINKED ISSUE #5148: avg on decimal raises spurious ARITHMETIC_OVERFLOW for empty or all-null input in ANSI mode
## Describe the bug

In ANSI mode, `avg` on a decimal column raises `[ARITHMETIC_OVERFLOW] Overflow in sum of decimals` when an ungrouped aggregate has nothing to average — either an empty input or an input where every value is null. Spark returns `NULL` in both cases.

The error comes from the non-grouped `AvgDecimalAccumulator` in `native/spark-expr/src/agg_funcs/avg_decimal.rs`:

- `merge_batch` sets `is_empty = partial_counts.len() == partial_counts.null_count()`. A global partial aggregate over zero rows still emits one row with `sum = NULL, count = 0`, and that count is not null, so `is_empty` flips to `false`.
- `evaluate` then sees `sum.is_none() && !is_empty` and reports an overflow. The adjacent `is_not_null` check a few lines below guards with `count > 0`; this one does not.

The `GroupsAccumulator` path guards its equivalent check with `count > 0`, so only aggregates without `GROUP BY` are affected.

## Steps to reproduce

```sql
set spark.sql.ansi.enabled = true;
create table t(a decimal(38, 2)) using parquet;
select avg(a) from t;
```

```
org.apache.spark.SparkArithmeticException: [ARITHMETIC_OVERFLOW] Overflow in sum of decimals.
Use 'try_avg' to tolerate overflow and return NULL instead. If necessary set
"spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22003
```

The same happens for a table that has rows where every `a` value is null.

## Expected behavior

`NULL`, matching Spark. There is no sum to overflow when nothing was accumulated.

## Additional context

Found while investigating the Spark 4.1 SQL test failures on https://github.com/apache/datafusion-comet/pull/5114. `LogicalPlanTagInSparkPlanSuite` q24a/q24b/q24-v2.7 hit this error because that suite creates the TPC-DS tables with no rows and the q24 scalar subquery is `SELECT 0.05 * avg(netpaid) FROM ssales`. The Spark 3.5 SQL lane does not catch it because ANSI is off by default there.

A fix is included in https://github.com/apache/datafusion-comet/pull/5114 (separate commit), since that PR needs it to get the Spark 4.1 lane green.

andygrove:

Part of #5112. Related to #4440. Also fixes #5148.

## Rationale for this change

Isolated benchmarking in #5112 shows the native columnar-to-row converter is 3.7x slower per row than the JVM `CometColumnarToRowExec` at the default 8192-row batch size, and up to 15.7x slower for small batches, because the native path carries roughly 10us of fixed JNI and FFI cost per batch. End-to-end microbenchmarks and the TPC-DS SF1000 comparison in #4440 show no measurable benefit from the native operator, and the original GC-pressure rationale no longer holds since #3367 made the converter deep-copy every row to the JVM heap. Real workloads that reach C2R with small batches (selective filters, small shuffle partitions, broadcast dimension tables) regress, so the safer default is the JVM implementation until the native path is competitive.

- Change the default of `spark.comet.exec.columnarToRow.native.enabled` from `true` to `false`, and update the config doc to mention the per-batch JNI cost.
- Update the `CometNativeColumnarToRowExec` scaladoc: it claimed zero-copy variable-length data and reduced GC pressure, neither of which holds since the #3308 fix.
- Fix the DPP subquery-broadcast rewrite in `CometExecRule` to also strip the JVM `CometColumnarToRowExec`. Unlike `CometNativeColumnarToRowExec`, it is `CodegenSupport`, so by the time the rule sees the subquery plan it has been collapsed into `WholeStageCodegenExec(CometColumnarToRowExec(InputAdapter(cometPlan)))`. Without this the rule failed to recognize the Comet child and the columnar plan was not unwrapped.
- Regenerate the TPC-DS plan-stability golden files: the transition node in every affected plan changes from `CometNativeColumnarToRow` to `CometColumnarToRow`.
- Update `dev/diffs/*.diff` so the Spark `SubquerySuite` partition-pruning assertion accepts the whole-stage-codegen-wrapped `CometColumnarToRowExec` shape in addition to `CometNativeColumnarToRowExec`, and so `CachedBatchSerializerNoUnwrapSuite` looks through the `WholeStageCodegenExec` before classifying the cached plan's transition node. The latter previously matched only on the top node being a Comet operator, which held for the non-codegen `CometNativeColumnarToRowExec` but not for the codegen-wrapped `CometColumnarToRow`.
- Fix a pre-existing decimal `avg` bug (#5148) that this change exposes: in ANSI mode, an ungrouped `avg` over an empty or all-null decimal input raised `[ARITHMETIC_OVERFLOW] Overflow in sum of decimals` instead of returning NULL, because `AvgDecimalAccumulator::merge_batch` derives `is_empty` from the nullness of the merged counts and a global partial aggregate over zero rows emits `sum = NULL, count = 0` with a non-null count. `evaluate` now guards that check with `count > 0`, matching the neighbouring `is_not_null` check and the `GroupsAccumulator` path. This surfaces here because `CometColumnarToRowExec` supports codegen, so `BenchmarkQueryTest.checkGeneratedCode` reaches `WholeStageCodegenExec.doCodeGen`, which prepares the plan and eagerly executes scalar subqueries: `LogicalPlanTagInSparkPlanSuite` q24a/q24b/q24-v2.7 then ran q24's `SELECT 0.05 * avg(netpaid) FROM ssales` against the row-less TPC-DS tables that suite creates. It is only visible in the Spark 4.1 lane because ANSI is off by default in Spark 3.5.
- Update the plan documentation (`understanding-comet-plans.md`, `installation.md`): example plans now show `CometColumnarToRow`, and the operator reference entry for `CometNativeColumnarToRow` no longer claims zero-copy variable-length handling and records that it is now opt-in.

Users can still opt in with `spark.comet.exec.columnarToRow.native.enabled=true`.

Existing tests cover both code paths explicitly: `CometNativeColumnarToRowSuite` enables the config in its own `withSQLConf` blocks, `CometFuzzTestBase` parameterizes it, and plan assertions in `CometExecSuite` and `CometPlanChecker` match both operator variants, so no test depends on the default. The rest of the suite now exercises `CometColumnarToRowExec` by default, which was the default behavior before #3299. The Spark SQL test lanes (`dev/diffs`) and the TPC-DS plan-stability suites cover the plan-shape changes.

The decimal `avg` fix adds `CometAggregateSuite."avg decimal with nothing to average"`, covering the ungrouped and grouped cases over both an empty table and a table whose values are all null, with ANSI enabled and disabled. It fails with `ARITHMETIC_OVERFLOW` without the native fix.

### DIFF
--- a/dev/diffs/3.4.3.diff
+++ b/dev/diffs/3.4.3.diff
@@ -1124,14 +1124,14 @@ index 18123a4d6ec..0fe185baa33 100644
  
    test("non-matching optional group") {
 diff --git a/sql/core/src/test/scala/org/apache/spark/sql/SubquerySuite.scala b/sql/core/src/test/scala/org/apache/spark/sql/SubquerySuite.scala
-index 75eabcb96f2..f8141c28f60 100644
+index 75eabcb96f2..c68f4827bf1 100644
 --- a/sql/core/src/test/scala/org/apache/spark/sql/SubquerySuite.scala
 +++ b/sql/core/src/test/scala/org/apache/spark/sql/SubquerySuite.scala
 @@ -21,10 +21,11 @@ import scala.collection.mutable.ArrayBuffer
  
  import org.apache.spark.sql.catalyst.expressions.SubqueryExpression
  import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, Join, LogicalPlan, Project, Sort, Union}
-+import org.apache.spark.sql.comet.{CometNativeColumnarToRowExec, CometNativeScanExec, CometScanExec}
++import org.apache.spark.sql.comet.{CometColumnarToRowExec, CometNativeColumnarToRowExec, CometNativeScanExec, CometScanExec}
  import org.apache.spark.sql.execution._
  import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanHelper, DisableAdaptiveExecution}
  import org.apache.spark.sql.execution.datasources.FileScanRDD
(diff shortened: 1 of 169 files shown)


mbutrovich (approved):
I think this makes sense for the 1.0 release, and we can continue to revisit optimizing the native implementation and possible dynamic choices based on schema/cardinality.

Thanks @andygrove!

andygrove:
Thanks for the review @mbutrovich. Changing the default exposed a pre-existing bug which had a trivial fix so I have now added this to the PR. See the most recent commit.

mbutrovich (approved):
> Thanks for the review @mbutrovich. Changing the default exposed a pre-existing bug which had a trivial fix so I have now added this to the PR. See the most recent commit.

Nice!

PR #5118, opened 2026-07-29
test: add schema scenarios and allocation measurement to isolated C2R benchmark

andygrove:

Part of #5112.

## Rationale for this change

The isolated columnar-to-row benchmark added in #5113 only measures one schema (long, int, double, string), which is close to the JVM path's best case: every accessor is allocation-free or escape-analyzable, and the schema contains a var-length column so the native converter is forced onto its slower row-at-a-time path. Real workloads such as TPC-H reach C2R with very different schemas, and cluster results there can favor the native converter even though this benchmark says otherwise. The benchmark needs to cover the schema dimensions that decide which implementation wins.

Parameterizes `CometC2RIsolatedBench` over schema scenarios, sweeping each at batch sizes 8192, 512, and 32:

- long, int, double, string (the original baseline)
- 4 x decimal(12,2), date, 2 x string (TPC-H lineitem-like; the JVM accessor allocates a `Decimal` per value)
- 4 x decimal(12,2), date, long (all fixed-width, which engages the native column-wise fast path)
- 2 x decimal(38,10), long (high precision; the JVM accessor allocates a byte[], `BigInteger`, java `BigDecimal`, scala `BigDecimal`, and `Decimal` per value)
- 16 x long (wide primitives)

Also reports JVM heap bytes allocated per row for each path via `ThreadMXBean.getThreadAllocatedBytes`, since the two implementations have very different allocation profiles that an idle-heap microbenchmark's wall-clock numbers do not surface: the JVM path allocates per object-typed value but reuses its output row buffer, while the native path allocates a `byte[]` plus `UnsafeRow` per row for the defensive copy added in #3367.

Results on Apple M3 Ultra, Spark 4.1 (per-row ns, best time; allocation in JVM heap bytes per row at batchSize=8192):

| Scenario | JVM 8192 | Native 8192 | JVM 32 | Native 32 | JVM alloc B/row | Native alloc B/row |
|---|---|---|---|---|---|---|
| long, int, double, string | 8.5 | 32.7 | 16.5 | 309.9 | 24.1 | 100.0 |
| 4 x decimal(12,2), date, 2 x string | 35.5 | 52.6 | 45.9 | 518.6 | 131.2 | 120.8 |
| 4 x decimal(12,2), date, long (all fixed-width) | 20.4 | **16.7** | 30.7 | 451.0 | 152.2 | 146.0 |
| 2 x decimal(38,10), long | 58.7 | 70.5 | 73.6 | 310.5 | 392.1 | 137.0 |
| 16 x long | 23.7 | 31.4 | 36.5 | 966.3 | 24.5 | 228.5 |

The all-fixed-width decimal scenario is the first case where the native converter wins outright, and the decimal scenarios show the JVM path allocating up to 3x more heap per row, a cost that is mostly invisible on an idle benchmark heap but real on a loaded executor.

Ran `make benchmark-org.apache.spark.sql.benchmark.CometC2RIsolatedBench` twice; results above are from the second run and the baseline scenario reproduces the numbers reported in #5112.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/benchmark/CometC2RIsolatedBench.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/benchmark/CometC2RIsolatedBench.scala
@@ -33,35 +33,138 @@ import org.apache.comet.NativeColumnarToRowConverter
  * Isolated columnar-to-row microbenchmark that excludes the parquet scan entirely. Batches are
  * built once in memory; each case converts them to rows and consumes the values so that the
  * conversion itself is what gets measured.
+ *
+ * The schema scenarios exist because the two implementations have very different per-value costs
+ * by data type. The JVM path allocates a Decimal object per non-null decimal value (and a
+ * BigInteger/BigDecimal chain for precision > 18), while the native path writes the unscaled
+ * value directly. The native path also has a column-wise fast path that only engages when every
+ * column in the schema is fixed-width.
  */
 object CometC2RIsolatedBench {
 
   private val totalRows = 1024 * 1024
 
-  private def makeBatches(schema: StructType, batchSize: Int): Array[ColumnarBatch] = {
-    val rows = (0 until totalRows).iterator.map { i =>
-      InternalRow(i.toLong, i, i.toDouble, UTF8String.fromString(s"value_$i"))
-    }
+  private case class Scenario(
+      name: String,
+      schema: StructType,
+      makeRow: Int => InternalRow,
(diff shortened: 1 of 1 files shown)


andygrove:
Folding this into #5120. That PR was stacked on this branch and already contains this commit (`e27384c`) as its first commit, and it targets `main` directly, so the benchmark scenarios and allocation measurement land there along with the converter they motivated. #5120's description now covers this work and carries the baseline numbers from here.

Closing in favor of #5120; no changes are lost.

PR #5127, opened 2026-07-29
fix: honour Spark's legacy `null IN ()` behavior

### LINKED ISSUE #4786: bug: `null IN ()` returns false instead of null under legacy null-in-empty behavior
### Describe the bug

For a null operand against an empty IN list, Comet always returns `false`. Spark returns `null` when `spark.sql.legacy.nullInEmptyListBehavior=true`. Comet ignores the config and the null operand.

### Steps to reproduce

Empty IN lists are not expressible in SQL, so build via the DataFrame API; `ConvertToLocalRelation` and `OptimizeIn` must be disabled or the IN is folded away. Add to a suite extending `CometTestBase` (config exists in Spark 4+):

```scala
test("null IN empty list under legacy behavior") {
  assume(org.apache.comet.CometSparkSessionExtensions.isSpark40Plus)
  withSQLConf(
    "spark.sql.optimizer.excludedRules" ->
      ("org.apache.spark.sql.catalyst.optimizer.ConvertToLocalRelation," +
        "org.apache.spark.sql.catalyst.optimizer.OptimizeIn"),
    "spark.sql.legacy.nullInEmptyListBehavior" -> "true") {
    val data: Seq[(Integer, Integer)] =
      Seq((Integer.valueOf(1), Integer.valueOf(1)), (null, Integer.valueOf(2)))
    withParquetTable(data, "t") {
      val df = sql("SELECT _1 AS a FROM t").select(col("a"), col("a").isin())
      checkSparkAnswer(df)
    }
  }
}
```

```
== Results ==
!== Spark Answer - 2 ==            == Comet Answer - 2 ==
 struct<a:int,(a IN ()):boolean>   struct<a:int,(a IN ()):boolean>
 [1,false]                         [1,false]
![null,null]                       [null,false]
```

### Expected behavior

`null IN ()` returns `null` under legacy behavior, matching Spark. (Under ANSI-on, the Spark 4 default, `false` is expected and Comet already matches, so only the legacy path is affected.)

### Additional context

Found while enabling `CometLocalTableScanExec` by default (#4393), but reproduces over a Parquet scan. Upstream test: `EmptyInSuite` "IN with empty list".

andygrove:

Closes #4786.

## Rationale for this change

Comet's native `in` kernel always returns `false` for any operand against an empty `IN` list. Spark only behaves that way under the non-legacy behavior; under the legacy behavior a `NULL` operand evaluates to `NULL` (SPARK-44550), so Comet returns `false` where Spark returns `NULL`.

That legacy behavior applies more often than the config name suggests:

- Spark 3.4: no config exists and `In`/`InSet` always use the legacy behavior
- Spark 3.5: `spark.sql.legacy.nullInEmptyListBehavior` defaults to `true`
- Spark 4.0+: the config is optional and defaults to `!spark.sql.ansi.enabled`, so the legacy behavior is in effect whenever ANSI mode is disabled

- `CometIn` / `CometInSet` return `Unsupported` for an empty list when the legacy behavior is in effect. Both serdes already mix in `CodegenDispatchFallback`, so these cases run Spark's own `doGenCode` inside the Comet pipeline instead of falling the whole operator back to Spark.
- `CometNot`'s `Not(In(...))` fast path skips the same case so it defers to `CometIn` (otherwise `NOT (null IN ())` returned `true` instead of `NULL`).
- Removed the now-fixed divergence from the compatibility guide, and replaced `CometIn`'s compatible note with an unsupported-case reason.

`OptimizeIn` rewrites empty `IN` lists away, so this only affects plans where that rule is excluded; non-empty `IN` lists are unaffected.

New test in `CometExpressionSuite` covering `a IN ()` and `NOT (a IN ())` over a Parquet scan with a `NULL` operand, across `spark.sql.legacy.nullInEmptyListBehavior` set to `true`, `false`, and unset (which exercises the version- and ANSI-derived default) with ANSI mode both on and off. The test fails on `main` with:

```
!== Spark Answer - 2 ==     == Comet Answer - 2 ==
 [1,false,true]             [1,false,true]
![null,null,null]           [null,false,true]
```

### DIFF
--- a/docs/source/user-guide/latest/compatibility/index.md
+++ b/docs/source/user-guide/latest/compatibility/index.md
@@ -145,9 +145,6 @@ so users hunting an unexpected value have a single place to check:
 - Native `RANGE` window frames with an explicit `PRECEDING` / `FOLLOWING` offset diverge from
   Spark when the boundary arithmetic overflows for `DATE` or `DECIMAL` `ORDER BY` columns
   ([#5022](https://github.com/apache/datafusion-comet/issues/5022)).
-- For an empty `IN` list, Comet always returns `false` for a `NULL` operand. Spark returns `NULL`
-  when `spark.sql.legacy.nullInEmptyListBehavior=true` (Spark 4.0+)
-  ([#4786](https://github.com/apache/datafusion-comet/issues/4786)).
 
 ## Object store cache
 
--- a/spark/src/main/scala/org/apache/comet/serde/predicates.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/predicates.scala
@@ -22,9 +22,10 @@ package org.apache.comet.serde
 import scala.jdk.CollectionConverters._
 
 import org.apache.spark.sql.catalyst.expressions.{And, Attribute, BinaryExpression, EqualNullSafe, EqualTo, Expression, GreaterThan, GreaterThanOrEqual, In, InSet, IsNaN, IsNotNull, IsNull, LessThan, LessThanOrEqual, Literal, Not, Or}
+import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.BooleanType
 
-import org.apache.comet.CometSparkSessionExtensions.withFallbackReason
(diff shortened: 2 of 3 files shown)


comphead (approved):
Thanks @andygrove

Issue #5128, opened 2026-07-29
[EPIC] Optimize native cast expression kernels


andygrove:
## What is the problem the feature request solves?

A review of the native cast kernels in `native/spark-expr/src/conversion_funcs/` found a number of remaining per-row costs: element-wise builder loops that could be vectorized Arrow passes, batch-constant work recomputed per row, per-row heap allocations, and redundant regex matching. This epic tracks optimizing them, following the methodology in `docs/source/contributor-guide/optimizing_expressions.md` (baseline first, bit-identical output, no regression on any benchmark shape).

### Already tuned (do not re-tread)

- `spark_cast_int_to_int` narrowing casts (#4920)
- float/double to decimal (#4940)
- `parse_string_to_decimal` (#4916)
- `date_parser` / string to date (#4917)
- `spark_cast_float64_to_utf8` (#4918)
- `cast_decimal128_to_utf8` (#4924)
- `cast_binary_to_string` (#4763, #4912)

### High impact

- [ ] **String to timestamp / timestamp_ntz: eliminate double regex matching.** `timestamp_parser` (`string.rs`) ORs `is_match` across 14 patterns to detect a direct match, then `timestamp_parser_with_tz` re-matches the same patterns sequentially to dispatch, up to ~28 regex executions per row (~21 for the NTZ variant). A single `regex::RegexSet` pass returning the matched index, reused for both decisions, removes most of this; a byte classifier (the patterns are digit-count/separator shapes, like the tuned `date_parser`) would go further. No benchmark exists for string to timestamp; one must be added first. Traps: first-match-wins pattern order, `RE_YEAR` accepts 4-6 digits while the others accept 4-7, ANSI errors must carry the raw untrimmed value.
- [ ] **Decimal to int: hoist the per-row `10^scale` and vectorize.** The `cast_decimal_to_int*` macros (`numeric.rs`) evaluate `10_i128.pow(scale)` inside the per-row closure in every arm, then collect element by element. Hoist the divisor, then use `unary` for Legacy/Try (both wrap via `as`) and `try_unary` for ANSI. Traps: ANSI error text formats the value via `format_decimal_str` with a `BD` suffix; int8/int16 targets truncate through i32 first. No benchmark exists.
[...]

PR #5130, opened 2026-07-29
perf: classify string-to-timestamp shapes in one byte scan (up to 5.6x faster)

andygrove:

Part of #5128 (first "High impact" item: string to timestamp / timestamp_ntz double regex matching).

## Rationale for this change

Optimize existing expression.

`timestamp_parser` ORed `is_match` across 14 regex patterns just to decide whether the value
already matched a base shape (so that an offset suffix must not be stripped), and then
`timestamp_parser_with_tz` re-matched the same 14 patterns sequentially to pick the parse
routine — up to 28 regex executions per row. `timestamp_ntz_parser` matched three separate
times (7 time-only + 7 date + 7 in the inner parse), for up to 21 per row.

Classify the input shape **once** into a `TimestampPattern` and thread that single result
through every decision that previously re-matched.

ASCII input — effectively all real data — is classified by one left-to-right byte scan, the
same technique as the already-tuned `date_parser`. The 14 patterns are digit-count and
separator shapes, so the scan is a direct transcription of them, including the two
asymmetries that are easy to lose: a bare year takes 4-6 digits while a separated year takes
4-7, and a trailing hour takes 1-2 digits while an hour followed by a minute must be exactly 2.

Non-ASCII input keeps regex semantics rather than the byte scan: the `regex` crate's `\d` is
Unicode-aware, so the patterns accept non-ASCII digits (e.g. Arabic-Indic) that an ASCII scan
does not. Those inputs go through a single `RegexSet` pass instead of the sequential scans, so
that path gets faster too without any behavior change.

Also removes the seven single-line `parse_str_to_*_timestamp` wrappers, which existed only to
be function pointers in the pattern table.

Existing tests — the full `datafusion-comet-spark-expr` suite (545 tests, was 540) and
`CometCastSuite` (162 succeeded, 0 failed, 9 ignored, on the Spark 4.1 profile so the
`is_spark4_plus` leading-whitespace path is covered).

Two new tests pin the classifier:

- `test_timestamp_pattern_classifier_matches_regexes` — a differential test comparing the
  classifier against a sequential scan of the original patterns over an exhaustive
  enumeration of every string of up to 5 shape characters (`019-:.T +`), plus every
  one-character deletion, substitution, and insertion applied to nine realistic timestamp
  strings, plus non-ASCII inputs. Verified to fail if the classifier is perturbed (widening
  the bare-year digit range to 4-7 makes it report `"0202001"`).
- `test_timestamp_pattern_digit_count_asymmetries` — readable anchors for the digit-count
  asymmetries above.

Spark's string-to-timestamp cases are ported into the Rust tests
(`DateTimeUtilsSuite`: `test("string to timestamp")`, `SPARK-35780` full range,
`SPARK-37326` `allowTimeZone`, the `stringToTimestamp` half of `SPARK-15379`, and the
`permuteWithWhitespaceAndControl` helper). Expected values are computed from a
proleptic-Gregorian day count rather than from the parser, so they are independent of the
code under test and reach the years outside `chrono`'s range that the
`Long.MinValue`/`MaxValue` microsecond boundaries need. Each case also runs across all
three eval modes and several session timezones.

All of Spark's expectations hold except two divergences that are unchanged from `main` and
unrelated to the shape classifier, now tracked by #5165: ISO control characters are not
trimmed (`str::trim` strips only Unicode whitespace, where Spark strips whitespace and ISO
control), and a leading `+` that is not a year sign returns null under ANSI where Spark
raises. Both are asserted at their current behavior with a pointer to the issue so they
cannot silently widen.

### Benchmark

New benchmark `native/spark-expr/benches/cast_string_to_timestamp.rs` (no coverage existed
for this cast). Baseline captured on unmodified source. Every shape improved; nothing
regressed.

| shape | main | this PR | change |
| --- | --- | --- | --- |
| `timestamp/canonical` | 1830.8 µs | 483.4 µs | **-73.6%** (3.79x) |
| `timestamp/microseconds` | 2238.4 µs | 576.9 µs | **-74.2%** (3.88x) |
| `timestamp/date_only` | 890.0 µs | 334.0 µs | **-62.5%** (2.66x) |
| `timestamp/time_only` | 2271.3 µs | 1198.2 µs | **-47.2%** (1.90x) |
| `timestamp/iso_z_suffix` | 2896.2 µs | 1127.8 µs | **-61.1%** (2.57x) |
| `timestamp/named_tz_suffix` | 4380.9 µs | 2574.8 µs | **-41.2%** (1.70x) |
| `timestamp/invalid` | 2566.4 µs | 993.8 µs | **-61.3%** (2.58x) |
| `timestamp/sparse_nulls` | 1740.7 µs | 465.2 µs | **-73.3%** (3.74x) |
| `timestamp/dense_nulls` | 203.2 µs | 68.4 µs | **-66.3%** (2.97x) |
| `timestamp/non_ascii_digits` | 1045.1 µs | 709.6 µs | **-32.1%** (1.47x) |
| `timestamp/mixed` | 1805.1 µs | 598.3 µs | **-66.9%** (3.02x) |
| `timestamp_named_zone/canonical` | 2082.2 µs | 722.4 µs | **-65.3%** (2.88x) |
| `timestamp_ntz/canonical` | 2271.3 µs | 406.6 µs | **-82.1%** (5.59x) |
| `timestamp_ntz/microseconds` | 2692.6 µs | 508.0 µs | **-81.1%** (5.30x) |
| `timestamp_ntz/date_only` | 1352.8 µs | 263.3 µs | **-80.5%** (5.14x) |
| `timestamp_ntz/iso_z_suffix` | 2838.5 µs | 1065.2 µs | **-62.5%** (2.66x) |
| `timestamp_ntz/invalid` | 2096.8 µs | 975.1 µs | **-53.5%** (2.15x) |
| `timestamp_ntz/dense_nulls` | 247.8 µs | 61.4 µs | **-75.2%** (4.03x) |
| `timestamp_ntz/mixed` | 2045.5 µs | 530.7 µs | **-74.1%** (3.85x) |

All 19 shapes reported `Performance has improved.` at p = 0.00.

No. Output is bit-identical: same values, same null buffer, same ANSI error text.


native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1619,87 +1574,307 @@ fn extract_offset_suffix(value: &str) -> Option<(&str, timezone::Tz)> {
...
-static RE_TIME_ONLY_HMS: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^T\d{1,2}:\d{1,2}:\d{1,2}$").unwrap());
-static RE_TIME_ONLY_HMSU: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^T\d{1,2}:\d{1,2}:\d{1,2}\.\d+$").unwrap());
-static RE_BARE_HM: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^\d{1,2}:\d{1,2}$").unwrap());
-static RE_BARE_HMS: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^\d{1,2}:\d{1,2}:\d{1,2}$").unwrap());
-static RE_BARE_HMSU: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^\d{1,2}:\d{1,2}:\d{1,2}\.\d+$").unwrap());
+/// The timestamp string shapes the parser recognises, listed in the order they are matched.
```
peterxcli, line 1577: I think this would be better to match the type and corresponding regex line by line but still lazy initialising one `RegexSet`. same perf, better readability.
```rust
#[derive(Clone, Copy, PartialEq, Eq, Debug)]
enum TimestampPattern {
    Year,
    Month,
    Day,
    Hour,
    Minute,
    Second,
    Microsecond,
    TimeOnlyH,
    TimeOnlyHm,
    TimeOnlyHms,
    TimeOnlyHmsu,
    BareHm,
    BareHms,
    BareHmsu,
}

impl TimestampPattern {
    /// Every shape, in the order they are matched. First match wins, so the order matters.
    const ALL: [TimestampPattern; 14] = [
        Self::Year,
        Self::Month,
        Self::Day,
        Self::Hour,
        Self::Minute,
        Self::Second,
        Self::Microsecond,
        Self::TimeOnlyH,
        Self::TimeOnlyHm,
        Self::TimeOnlyHms,
        Self::TimeOnlyHmsu,
        Self::BareHm,
        Self::BareHms,
        Self::BareHmsu,
    ];

    /// The regex equivalent of this shape.
    ///
    /// Only used for the rare non-ASCII input, where the Unicode-aware `\d` class accepts
    /// digits (e.g. Arabic-Indic) that the ASCII classifier does not.
    ///
    /// `Year` allows only 4-6 digits (not 7) because a bare 7-digit string like "0119704"
    /// is ambiguous and Spark rejects it. The others keep `\d{4,7}` because the `-`
    /// separator disambiguates the year, so "0002020-01-01" is validly year 2020 with
    /// leading zeros.
    fn regex_str(self) -> &'static str {
        match self {
            Self::Year => r"^-?\d{4,6}$",
            Self::Month => r"^-?\d{4,7}-\d{2}$",
            Self::Day => r"^-?\d{4,7}-\d{2}-\d{2}$",
            Self::Hour => r"^-?\d{4,7}-\d{2}-\d{2}[T ]\d{1,2}$",
            Self::Minute => r"^-?\d{4,7}-\d{2}-\d{2}[T ]\d{2}:\d{2}$",
            Self::Second => r"^-?\d{4,7}-\d{2}-\d{2}[T ]\d{2}:\d{2}:\d{2}$",
            Self::Microsecond => r"^-?\d{4,7}-\d{2}-\d{2}[T ]\d{2}:\d{2}:\d{2}\.\d+$",
            Self::TimeOnlyH => r"^T\d{1,2}$",
            Self::TimeOnlyHm => r"^T\d{1,2}:\d{1,2}$",
            Self::TimeOnlyHms => r"^T\d{1,2}:\d{1,2}:\d{1,2}$",
            Self::TimeOnlyHmsu => r"^T\d{1,2}:\d{1,2}:\d{1,2}\.\d+$",
            Self::BareHm => r"^\d{1,2}:\d{1,2}$",
            Self::BareHms => r"^\d{1,2}:\d{1,2}:\d{1,2}$",
            Self::BareHmsu => r"^\d{1,2}:\d{1,2}:\d{1,2}\.\d+$",
        }
    }
}

static TIMESTAMP_PATTERN_SET: LazyLock<RegexSet> = LazyLock::new(|| {
    RegexSet::new(TimestampPattern::ALL.map(TimestampPattern::regex_str)).unwrap()
});
```

peterxcli (commented):
other looks good, and we probably can port all spark string2ts test cases here.

native/spark-expr/src/conversion_funcs/string.rs
andygrove: Agreed, that reads much better — a shape and its pattern are declared together instead of relying on two arrays staying in the same order. Applied in e14e086, with `ALL` and `regex_str` on the enum and the `RegexSet` built from `ALL.map(regex_str)`.

I kept the `Year` 4-6 vs `\d{4,7}` note on `regex_str` since that asymmetry is the easiest thing to lose when editing the patterns. The differential test now builds its reference regexes from `regex_str()` too, and I re-confirmed it still fails when the classifier is perturbed.

andygrove:
> we probably can port all spark string2ts test cases here.

Good call — done in e14e086. Ported from `DateTimeUtilsSuite`: `test("string to timestamp")`, `SPARK-35780: support full range of timestamp string`, `SPARK-37326: stringToTimestampWithoutTimeZone with allowTimeZone`, and the `stringToTimestamp` half of `SPARK-15379`, plus a port of the `permuteWithWhitespaceAndControl` helper.

Two things I did differently from a literal transcription:

- Expected values are computed from a proleptic-Gregorian day count rather than from the parser, so they don't depend on the code under test. That also covers the years outside `chrono`'s range, which the `Long.MinValue`/`MaxValue` microsecond boundary cases need — the test asserts those two land exactly on `i64::MIN`/`i64::MAX`.
- Each case runs across all three eval modes and a few session timezones. Spark's suite loops `ALL_TIMEZONES` but only exercises the `Option`-returning function; adding the ANSI dimension is where the second finding below came from.

Every one of Spark's expectations holds except two divergences, both identical on `main` and unrelated to the shape classifier. Filed as #5165:

1. **ISO control characters aren't trimmed.** Spark trims whitespace *and* ISO control before parsing (`getTrimmedStart`/`getTrimmedEnd`); `timestamp_parser` uses Rust's `str::trim`, which only strips Unicode whitespace. So `CAST(CONCAT('2015-03-18 12:03:17', CHAR(3)) AS TIMESTAMP)` is null in Comet and a timestamp in Spark — all 30 of the `permuteWithWhitespaceAndControl` permutations for a valid input diverge. Notably `date_parser` in the same file already trims both, so `CAST(... AS DATE)` accepts input `CAST(... AS TIMESTAMP)` rejects.
2. **Leading `+` under ANSI.** `'+12:12:12'` and `'+'` return null in every mode; Spark raises `CAST_INVALID_INPUT` under ANSI. The non-ANSI value is correct.

I left both out of scope here to keep this PR behavior-preserving, and asserted them at their current behavior with a pointer to #5165 so they can't silently widen. The carve-outs in the ported tests are written so that removing them is the verification for the fix.

`CometCastSuite` on the Spark 4.1 profile: 162 succeeded, 0 failed, 9 ignored. Rust suite is 545 tests (was 540).

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1619,87 +1574,311 @@ fn extract_offset_suffix(value: &str) -> Option<(&str, timezone::Tz)> {
...
+) -> SparkResult<Option<i64>> {
+    let timestamp = match pattern {
+        Some(TimestampPattern::Year) => get_timestamp_values(value, "year", tz)?,
+        Some(TimestampPattern::Month) => get_timestamp_values(value, "month", tz)?,
+        Some(TimestampPattern::Day) => get_timestamp_values(value, "day", tz)?,
+        Some(TimestampPattern::Hour) => get_timestamp_values(value, "hour", tz)?,
+        Some(TimestampPattern::Minute) => get_timestamp_values(value, "minute", tz)?,
+        Some(TimestampPattern::Second) => get_timestamp_values(value, "second", tz)?,
+        Some(TimestampPattern::Microsecond) => get_timestamp_values(value, "microsecond", tz)?,
+        Some(_) => parse_str_to_time_only_timestamp(value, tz)?,
```
parthchandra, line 1879: Not a blocker, but this could be fragile if someone adds a new variant that is not time-only to TimestampPattern. We'd fall thru to the catch-all and produce wrong output.

parthchandra (commented):
Great improvement @andygrove

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -2941,6 +3195,532 @@ mod tests {
...
+            utc_micros(2015, 3, 18, 12, 3, 17, 123_120),
+            plus_730,
+        ));
+        check("2015-03-18T12:03:17.12312+7:30", expected);
+        check("2015-03-18T12:03:17.12312 UT+07:30", expected);
+        check("2015-03-18T12:03:17.12312+0730", expected);
+
+        // Time-only strings are anchored to the current date, so assert against a
+        // today-relative expectation rather than a fixed instant.
+        let today = chrono::Utc::now().date_naive();
```
rich7420: Could we use the input timezone's current date here? After 16:30 UTC, the `+07:30` cases are a day ahead of UTC, so the expected value is off by 24 hours. I reproduced the failure and confirmed that using the date in `+07:30` fixes it.

rich7420:
@andygrove thanks for the patch!

native/spark-expr/src/conversion_funcs/string.rs
andygrove: Fixed in ad2e83017. You're right, and the anchor is the string's own zone rather than UTC — the parser resolves it with `tz.from_utc_datetime(Utc::now())`, and for a string carrying its own offset that `tz` **is** the offset from the string. The expectation was built from `chrono::Utc::now().date_naive()`, so it disagreed for exactly the part of the day you identified.

`today_micros` now derives the date from the offset it is given:

```rust
let today = (chrono::Utc::now() + chrono::Duration::seconds(offset)).date_naive();
```

One thing I wanted to avoid was fixing this and leaving a test that only proves it for part of the day — at 14:35 UTC the `+07:30` cases pass either way, so a green run says nothing. I added two permanent cases with extreme offsets in opposite directions:

```rust
check("T18:12:15+14:00", today_micros(18, 12, 15, 0, 14 * 3600));
check("T18:12:15-12:00", today_micros(18, 12, 15, 0, -12 * 3600));
```

`+14:00` crosses the date line from 10:00 UTC onward and `-12:00` crosses before 12:00 UTC, so between them one is always in the failing window whatever hour CI runs at. Checked against the old code at 14:35 UTC and the `+14:00` case fails by exactly 86,400 s:

```
left: Some(1789013535000000)
right: Some(1788927135000000)
```

native/spark-expr/src/conversion_funcs/string.rs
andygrove: Fixed in ad2e83017 — not a blocker, but you're describing a silent-wrong-answer failure mode, which is the kind worth closing off while it is cheap.

The catch-all is gone; the time-only arm now lists its variants:

```rust
// Listed out rather than caught by a `Some(_)` arm on purpose: a catch-all would route
// any future non-time-only shape into the time-only parser and produce a wrong value
// silently. Spelling out the variants makes adding one a compile error here.
Some(
    TimestampPattern::TimeOnlyH
    | TimestampPattern::TimeOnlyHm
    | TimestampPattern::TimeOnlyHms
    | TimestampPattern::TimeOnlyHmsu
    | TimestampPattern::BareHm
    | TimestampPattern::BareHms
    | TimestampPattern::BareHmsu,
) => parse_str_to_time_only_timestamp(value, tz)?,
```

So adding a variant now stops the build here rather than quietly parsing a date shape as a time.

PR #4707, opened 2026-06-22
perf: cache full Parquet metadata (incl. page index) via DataFusion's CachedParquetFileReaderFactory

mbutrovich:
## Which issue does this PR close?

Part of #3978.

## Rationale for this change

Comet's `native_datafusion` scan used a hand-rolled `CachingParquetReaderFactory` that cached only the Parquet footer. DataFusion's opener loads the page index in a separate step, so it re-fetched the page index from object storage on every open (the footer-only cache never held it). For non-selective `IS NOT NULL` predicates on non-null join keys (e.g. TPC-DS q88), the page index prunes nothing yet is re-read per split, adding gigabytes of wasted I/O at scale. The hand-rolled cache was also unbounded (no eviction).

## What changes are included in this PR?

- Replace `CachingParquetReaderFactory` with DataFusion's `CachedParquetFileReaderFactory`, backed by the per-task `RuntimeEnv` file-metadata cache (bounded LRU, `metadata_cache_limit`). It loads the full metadata including the page index once per file, so the opener no longer re-fetches it.
- Delete `parquet_read_cached_factory.rs`.

Follow-up noted in a code TODO: metadata I/O is not reflected in `bytes_scanned` because `fetch_metadata` reads via `ObjectStore::get_ranges` directly. A byte-counting `ObjectStore` wrapper would surface it.

## How are these changes tested?

New `caches_full_metadata_with_page_index` unit test in `parquet_exec.rs`: writes a Parquet file with a page index, runs a scan, and asserts the `RuntimeEnv` metadata cache holds metadata with the column and offset index loaded.

Issue #3978, opened 2026-04-17
chore: DataFusion 54.0.0


comphead:
### What is the problem the feature request solves?

The placeholder for DF 54.0.0
https://github.com/apache/datafusion/issues/21080

### Internal dependencies  
- [ ] https://github.com/apache/datafusion-comet/issues/3929
- [ ] https://github.com/apache/datafusion-comet/pull/4161

comphead:
@mbutrovich cc

mbutrovich:
I'm trying to maintain a branch for this release so Comet doesn't fall behind and isn't able to test x.0 RCs:

https://github.com/apache/datafusion-comet/pull/3916

I update it every day or two, and file upstream fixes to DataFusion. So far it's found real issues that would have likely had us waiting for a x.1 or x.2 release.

Note that in the DF 54.0 branch I am now enabling SMJ with filter by default as the performance regressions there are now gone after https://github.com/apache/datafusion/issues/20910 and https://github.com/apache/datafusion/issues/21197.

comphead:
Experiments started in #4062

mbutrovich:
Benchmarked SF1000 TPC-H and TPC-DS:

<img width="1000" height="600" alt="Image" src="https://github.com/user-attachments/assets/915ba18c-4085-4df6-ac9b-a8614775795e" />
<img width="3500" height="600" alt="Image" src="https://github.com/user-attachments/assets/5aa48a23-62fd-4126-8432-75ed580a8ced" />
<img width="1000" height="600" alt="Image" src="https://github.com/user-attachments/assets/5bb017fa-f7af-4613-87ec-8ecb91887b19" />
<img width="1500" height="600" alt="Image" src="https://github.com/user-attachments/assets/1535d932-88ff-45c3-a8d8-ef6a10e85665" />

I'm digging into the regressions.

mbutrovich:
Benchmarking the DataFusion 54 branch against main, q88 regresses roughly 12s per iteration (about 21s to 34s). The regression is entirely scan-side, not the join. On the `store_sales` `CometNativeScan` the metrics move like this:

| metric (store_sales scan, summed) | main (DF 53.1) | DF 54 branch |
| --- | --- | --- |
| Time reading and parsing footer metadata | 2.0 m | 18.0 m |
| Wall clock for file opening | 6.0 m | 18.0 m |
| Number of bytes scanned | 586.6 MiB | 3.4 GiB |
[...]

PR #5132, opened 2026-07-29
fix: work around DataFusion 54.1.0 Parquet page-index regression

mbutrovich:

Part of #3978.

## Rationale for this change

Comet's `native_datafusion` Parquet scan regressed on q88-shaped queries (three `IS NOT NULL` predicates on non-null foreign keys against a wide fact table, scanned across thousands of files and partitions) between DataFusion 54.0.0 and 54.1.0. Duration on one benchmark went from 24s to 34s for that query alone.

The cause is upstream, in `datafusion-datasource-parquet`'s opener. apache/datafusion#22857 (backported to branch-54 as #23088, first released in 54.1.0) reordered the opener so the initial metadata load requests `PageIndexPolicy::Skip`, and defers loading the page index until row-group statistics can't already prove the surviving row groups are fully matched. That's a real win when the skip fires. When it doesn't, the deferred load (`load_page_index` in `opener/mod.rs`) reads the page index directly off the `AsyncFileReader`'s byte-range methods, bypassing `ParquetFileReaderFactory::get_metadata` and therefore the `FileMetadataCache` entirely. The result is never written back to the cache the initial load populated.

`store_sales`'s row groups don't carry the `null_count` statistics the skip needs for `IS NOT NULL`, so the skip never fires for this query, and every open of the same file now pays for a fresh, uncached page-index fetch, once per row-group split. Comparing 54.0.0 to 54.1.0 on the same query and data, cumulative `CometNativeScan` metrics for `store_sales` moved:

| metric (summed across the scan's task instances) | 54.0.0 | 54.1.0 |
| --- | --- | --- |
| Wall clock time elapsed for file opening | 79.5 min | 149.8 min (+88%) |
| Wall clock time elapsed for data decompression + decoding | 57.0 min | 77.6 min (+36%) |
| Total time reading and parsing footer metadata | 79.0 min | 89.1 min (+13%) |
| Number of bytes scanned | 3.4 GiB | 3.5 GiB (+3%) |

Bytes scanned barely moved; file-opening wall clock nearly doubled. That's a request-count/latency signature, not a data-volume one, consistent with a second, uncached fetch being added per open rather than more data being read.

This isn't specific to DataFusion 54.1: it's a gap in how `load_page_index` in the opener interacts with `FileMetadataCache`, present in current DataFusion main too. Filed upstream as apache/datafusion#23978. This PR is a Comet-side workaround, not a fix for the upstream gap.

- Add `EagerPageIndexReaderFactory` (`native/core/src/parquet/eager_page_index_reader_factory.rs`), a `ParquetFileReaderFactory` that ignores the `PageIndexPolicy` DataFusion's opener requests and always fetches the page index eagerly, via `DFParquetMetadata::with_page_index_policy(Some(PageIndexPolicy::Optional))`, caching it in the same shared `FileMetadataCache` DataFusion's own factory uses. Once the page index is present in that first load, the opener's own `missing_column_index || missing_offset_index` guard in `load_page_index` sees it's already there and never issues the uncached fetch.
- Wire it into `init_datasource_exec` in `native/core/src/parquet/parquet_exec.rs`, replacing `CachedParquetFileReaderFactory`.
- Update the existing regression test (`caches_full_metadata_with_page_index`, added by #4707) to assert the page index is present in the cache even for a scan with no pruning predicate, since this factory no longer depends on a predicate driving the load.

The tradeoff: this gives up the benefit of #22857's skip for files where it would have legitimately avoided the page index load (selective predicates, row groups with real `null_count` stats). Given `store_sales` never benefits from the skip here regardless of DataFusion version, and the alternative is unbounded repeated I/O at scale, that trade is worth it for now. Revert this once apache/datafusion#23978 is fixed.

Updated `caches_full_metadata_with_page_index` in `parquet_exec.rs`: writes a Parquet file with a page index, runs a scan through `init_datasource_exec` with no pruning predicate, and asserts the `RuntimeEnv` metadata cache holds the column and offset index after the first open.

No new integration/benchmark coverage in this PR; the regression itself was found via TPC-DS q88 wall-clock comparison across DataFusion versions, not a unit test, and isn't something existing CI would catch (bytes-scanned and duration deltas at this scale need the full SF1000-class dataset).

### DIFF
--- a/native/core/src/parquet/eager_page_index_reader_factory.rs
+++ b/native/core/src/parquet/eager_page_index_reader_factory.rs
@@ -61,11 +61,11 @@ use std::sync::Arc;
 #[derive(Debug)]
 pub struct EagerPageIndexReaderFactory {
     store: Arc<dyn ObjectStore>,
-    metadata_cache: Arc<FileMetadataCache>,
+    metadata_cache: Arc<dyn FileMetadataCache>,
 }
 
 impl EagerPageIndexReaderFactory {
-    pub fn new(store: Arc<dyn ObjectStore>, metadata_cache: Arc<FileMetadataCache>) -> Self {
+    pub fn new(store: Arc<dyn ObjectStore>, metadata_cache: Arc<dyn FileMetadataCache>) -> Self {
         Self {
             store,
             metadata_cache,
@@ -111,7 +111,7 @@ struct EagerPageIndexReader {
     store: Arc<dyn ObjectStore>,
     inner: ParquetObjectReader,
     partitioned_file: PartitionedFile,
-    metadata_cache: Arc<FileMetadataCache>,
+    metadata_cache: Arc<dyn FileMetadataCache>,
     metadata_size_hint: Option<usize>,
 }
 
--- a/native/core/src/parquet/eager_page_index_reader_factory.rs
+++ b/native/core/src/parquet/eager_page_index_reader_factory.rs
@@ -29,12 +29,19 @@
 //! every open. At the scale of a wide fact table scanned across many partitions, that is
 //! repeated, unbounded I/O for the same bytes.
 //!
-//! This factory forces `PageIndexPolicy::Optional` on every metadata fetch, ignoring whatever
-//! policy the caller requests, so the page index is always present in the cached
-//! `ParquetMetaData` after the first load. DataFusion's opener checks whether the metadata it
-//! already has includes the page index before issuing its own fetch, so with this factory that
-//! check is always true and the uncached fetch never happens. The tradeoff: files where the
-//! opener's skip heuristic would have avoided the page index load entirely now load it anyway.
+//! This factory forces `PageIndexPolicy::Optional` on every metadata fetch for files with no
+//! decryption properties, ignoring whatever policy the caller requests, so the page index is
+//! always present in the cached `ParquetMetaData` after the first load. DataFusion's opener
+//! checks whether the metadata it already has includes the page index before issuing its own
+//! fetch, so with this factory that check is always true and the uncached fetch never happens.
+//! The tradeoff: files where the opener's skip heuristic would have avoided the page index load
+//! entirely now load it anyway.
+//!
+//! Encrypted files are exempt from the override: `DFParquetMetadata::fetch_metadata` disables
+//! `FileMetadataCache` entirely whenever decryption properties are set, so nothing gets cached
+//! for them either way, and forcing eager loading would only add an unconditional page-index
+//! fetch to encrypted scans that have no pruning predicate at all. Encrypted opens get exactly
+//! the caller's requested policy, unchanged from stock behavior.
 //!
 //! Filed upstream as apache/datafusion#23978. Revert this once the opener merges its deferred
 //! page-index load back into `FileMetadataCache` instead of bypassing it.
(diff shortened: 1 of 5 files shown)


native/core/src/parquet/eager_page_index_reader_factory.rs
```diff
@@ -0,0 +1,176 @@
...
+        let metadata_cache = Arc::clone(&self.metadata_cache);
+        let store = Arc::clone(&self.store);
+        let metadata_size_hint = self.metadata_size_hint;
+        async move {
+            DFParquetMetadata::new(store.as_ref(), &object_meta)
+                .with_file_metadata_cache(Some(metadata_cache))
+                .with_metadata_size_hint(metadata_size_hint)
+                .with_page_index_policy(Some(PageIndexPolicy::Optional))
+                .fetch_metadata()
+                .await
```
andygrove, line 170: `get_metadata` ignores `options`, which drops the file decryption properties DataFusion's opener passes in.

`CachedParquetFileReaderFactory` — the factory this replaces — forwards them (`datafusion-datasource-parquet-54.1.0/src/reader.rs:291`):

```rust
let file_decryption_properties = options
    .and_then(|o| o.file_decryption_properties())
    .map(Arc::clone);

let page_index_policy = options.map(|o| o.column_index_policy());

DFParquetMetadata::new(&self.store, &object_meta)
    .with_decryption_properties(file_decryption_properties)
    ...
```

Comet compiles that path in — `native/core/Cargo.toml` has `datafusion = { workspace = true, features = ["parquet_encryption", "sql"] }` — and `init_datasource_exec` sets `with_encryption_factory(...)` a dozen lines above this factory's wiring. So an encrypted-Parquet scan should now fail its metadata fetch rather than degrade gracefully.

Suggest keeping the forwarding and overriding only the page-index policy, so that override stays the single intentional deviation from the upstream factory:

```rust
fn get_metadata<'a>(
    &'a mut self,
    options: Option<&'a ArrowReaderOptions>,
) -> BoxFuture<'a, parquet::errors::Result<Arc<ParquetMetaData>>> {
    let object_meta = self.partitioned_file.object_meta.clone();
    let metadata_cache = Arc::clone(&self.metadata_cache);
    let store = Arc::clone(&self.store);
    let metadata_size_hint = self.metadata_size_hint;
    let file_decryption_properties = options
        .and_then(|o| o.file_decryption_properties())
        .map(Arc::clone);
    async move {
        DFParquetMetadata::new(store.as_ref(), &object_meta)
            .with_decryption_properties(file_decryption_properties)
            .with_file_metadata_cache(Some(metadata_cache))
            .with_metadata_size_hint(metadata_size_hint)
            // Deliberately ignore the caller-requested policy; see module docs.
            .with_page_index_policy(Some(PageIndexPolicy::Optional))
            .fetch_metadata()
            .await
            .map_err(|e| { /* unchanged */ })
    }
    .boxed()
}
```

(DataFusion gates its block on the `parquet_encryption` feature of its own crate; `datafusion-comet-core` has no equivalent crate feature, so this can be called unconditionally here.)

Also worth confirming whether any existing encrypted-Parquet test actually exercises `init_datasource_exec` — the updated `caches_full_metadata_with_page_index` test won't catch this one.

mbutrovich:
Benchmarked `main` against this PR branch with TPC-DS SF 1000:

tpcds_queries_compare
tpcds_allqueries

mbutrovich:
Q88 on main:

Screenshot 2026-07-29 at 3 44 29 PM

this PR:

Screenshot 2026-07-29 at 3 44 35 PM

andygrove (approved):
Thanks for the quick fix @mbutrovich

PR #5133, opened 2026-07-29
fix: support null calendar interval literals

### LINKED ISSUE #5058: Null CalendarInterval literal throws native exception instead of evaluating to null
### Describe the bug

A null literal of `CalendarIntervalType` in a projection throws a native exception instead of producing a null:

```
org.apache.comet.CometNativeException: General execution error with reason:
Interval(MonthDayNano) is not supported in Comet.
```

#4898 mapped `CalendarIntervalType` to Arrow `Interval(MonthDayNano)` and added it to the data-type support checks, the codegen kernels, and plan serialization. The null-literal path in the native planner was not updated to match. `create_null_literal`'s match in `native/core/src/execution/planner.rs:389-421` handles Utf8, Date32, Timestamp, Decimal128, Struct, Map, List, Time64, Duration and friends, but has no `Interval` arm, so it hits the catch-all at `:418`.

The result is a hard failure rather than a fallback: because the type now passes `isSupportedDataType`, the plan is accepted and only fails once it reaches native execution.

Contrast with `YearMonthIntervalType`, which is not in the supported set and so falls back cleanly:

```
Project [COMET: Unsupported data type YearMonthIntervalType(0,1)]
```

That is the behavior a null calendar interval should have at minimum, and ideally it should just evaluate natively.

### Steps to reproduce

```sql
SELECT CAST(NULL AS INTERVAL);
```

This is easy to hit indirectly. `MakeInterval` is `NullIntolerant`, so `NullPropagation` rewrites any call with a literal NULL argument into a null interval literal, and the original expression disappears before Comet sees it:

```sql
SELECT make_interval(NULL, 2, 3, 4, 5, 6, 7.008009);
```

Both fail with the exception above.

Non-null calendar interval literals are fine, and calendar interval values flowing through a `LocalTableScan` (as in `calendar_interval.sql`) are fine. It is specifically the null literal in a projection.

### Expected behavior

`SELECT CAST(NULL AS INTERVAL)` returns a single null row, matching Spark.

Adding an `Interval(MonthDayNano)` arm producing `ScalarValue::IntervalMonthDayNano(None)` to the null-literal match should cover it. Worth checking `Interval(YearMonth)` and `Interval(DayTime)` at the same time, since they are absent from that match too, though they are currently shielded by the type check.

### Additional context

Found while adding `make_interval` codegen-dispatch support in #5030. That PR works around it by driving its NULL coverage from nullable columns rather than NULL literals, since literal NULL arguments are constant-folded away by `NullPropagation` and never exercise the expression anyway.

Related: #4540 (interval type support epic), #4898 (the PR that added `CalendarIntervalType`).

peterxcli:

Closes #5058.

## Rationale for this change

`CalendarIntervalType` is supported as Arrow `Interval(MonthDayNano)`, but the native planner did not construct a null scalar for that type. As a result, null calendar interval literals passed the JVM support check and then failed during native planning.

Map null `Interval(MonthDayNano)` literals to `ScalarValue::IntervalMonthDayNano(None)` in the native planner.

The existing `calendar_interval.sql` fixture covers `CAST(NULL AS INTERVAL)` and requires native execution.

- `cargo fmt --manifest-path native/Cargo.toml --all --check`
- `make core`
- `./mvnw test -Dtest=none -Dsuites="org.apache.comet.CometSqlFileTestSuite calendar_interval" -Dscalastyle.skip=true` with JDK 17 (1 test passed)


andygrove (commented):
Thanks for the quick fix. The arm itself is right: `Interval(MonthDayNano)` is what `CalendarIntervalType` maps to, and `ScalarValue::IntervalMonthDayNano(None)` is the matching null scalar. Two things I would like to sort out before this merges.

## 1. I do not think the cited fixture covers this

The PR says `calendar_interval.sql` covers `CAST(NULL AS INTERVAL)` and requires native execution. The text is in there, but it sits inside `SELECT * FROM VALUES (...)`, so the null interval arrives as a row value through the local table scan rather than as a null literal in a projection. That is exactly the distinction #5058 draws: interval values flowing through a `LocalTableScan` were already fine, and it is the literal path that hit the catch-all. The fixture is unmodified here and CI on main is green, so it must be passing without this change, which means nothing in the repo would catch a regression.

Could you add the repro from the issue? It needs a projection over a real scan so the literal is serialized and the query runs natively, something like:

```sql
statement
CREATE TABLE test_null_interval(id int) USING parquet

statement
INSERT INTO test_null_interval VALUES (1)

query
SELECT CAST(NULL AS INTERVAL) FROM test_null_interval
```

The `NullPropagation` route from the issue is worth a line too, since that is how this gets hit in practice rather than by someone writing a null cast by hand:

```sql
query
SELECT make_interval(NULL, 2, 3, 4, 5, 6, 7.008009) FROM test_null_interval
```

## 2. This match is the bug, not the missing arm

`native/core/src/execution/planner.rs`

The null branch enumerates types by hand, and every time a type is added to the JVM support checks someone has to remember to come back here. That is how we got here, and the failure mode is bad: the type passes `isSupportedDataType`, the plan is accepted, and it blows up in native planning instead of falling back.

DataFusion already has exactly this function. `ScalarValue::try_new_null(&data_type)` covers every arm currently in the match, including `Struct`, `Map`, `List`, `Time64(Nanosecond)`, `Duration(Microsecond)`, `Decimal128`, `Timestamp` with timezone, and `Null`, and it covers all three interval units on top of that. Its `Struct` arm builds the same null struct array that `ScalarStructBuilder::new_null` gives us. So the whole block could collapse to one call and this class of bug goes away:

```rust
let scalar_value = if literal.is_null {
    ScalarValue::try_new_null(&data_type)?
} else {
    ...
```

One thing to decide if you take this: unsupported types would then report DataFusion's `Can't create a null scalar from data_type "..."` instead of our `... is not supported in Comet`. A `map_err` would keep the Comet wording if you want it.

This also matters for your #5161, which adds an `Interval(YearMonth)` arm to this same match plus the `IntervalUnit` import. As written the two PRs will conflict, and switching to `try_new_null` lets both drop their arm. For what it is worth I checked the third unit and `DayTimeIntervalType` maps to `DurationVector`, so `Duration(Microsecond)` already covers it and `YearMonth` was the only other gap.

If you would rather keep the explicit match, that is a defensible call, but then please add the `Interval(YearMonth)` arm here as well so this file stops being a source of hard failures one type at a time.

peterxcli:
> 2. This match is the bug, not the missing arm

I think we should keep the explicit matching here, whenever we have checked that all types are supported, then we can switch to use `ScalarValue::try_new_null`. switching back...

peterxcli:
@andygrove thanks for the review. added sql tests

andygrove:

The fix itself is obviously right and the SQL coverage is in the correct place. Two suggestions.

**The same gap exists for other types**

That `match data_type` block in `planner.rs` is a hand-maintained list, and `Interval(MonthDayNano)` is not the only thing missing from it. `Interval(YearMonth)` is the immediate one, since `YearMonthIntervalType` is handled all over `CometBatchKernelCodegen*` and a null year-month literal would hit the same `"{dt:?} is not supported in Comet"` error. `LargeUtf8`, `Utf8View`, and `FixedSizeBinary` are others that would fail the same way.

Note that two arms of that match already do the right thing generically: `DataType::Map(f, s) => DataType::Map(f, s).try_into()?` and the `List` arm both go through DataFusion's `ScalarValue: TryFrom<&DataType>`, which returns a typed null for essentially every type. Would it be better to collapse the whole block to `ScalarValue::try_from(&data_type)` and keep explicit arms only where Comet genuinely needs to differ from DataFusion? That closes this class of bug permanently rather than one type at a time.

If you would rather keep this PR minimal, adding `Interval(YearMonth)` alongside `MonthDayNano` here and filing an issue for the general cleanup would also work.

**Test placement**

The two new queries go into `calendar_interval.sql`, which already covers non-null calendar intervals. That is the right file. One thing to check: does `SELECT CAST(NULL AS INTERVAL) FROM test_null_interval` actually reach the native planner, or does Spark's `ConstantFolding` collapse it before serialization? The comment on the `make_interval` case says `NullPropagation` folds it to a null interval literal, which is exactly what we want to exercise, so presumably yes. It would be worth confirming that the test fails on `main` without the planner change, since a fixture that passes either way is not protecting anything.

peterxcli:
Verified the test protects the fix: with the `Interval(MonthDayNano)` arm removed and everything else unchanged, `CometSqlFileTestSuite calendar_interval` fails with exactly the issue's error (`Interval(MonthDayNano) is not supported in Comet`), and passes with it. So `ConstantFolding` doesn't fold the literal away — it reaches the native planner.

On the other types: a null `Interval(YearMonth)` literal can't reach this match today. `CometLiteral.getSupportLevel` gates on `supportedDataType`, which doesn't include `YearMonthIntervalType`, so those literals fall back to Spark rather than failing natively. I'd rather add that arm in #5161, where the type actually becomes reachable. `LargeUtf8`/`Utf8View`/`FixedSizeBinary` can't appear here at all — `to_arrow_datatype` never produces them from the proto. So after this PR the explicit match covers every type the serde can emit, which is the state I wanted before considering the `try_from` collapse (per the earlier review round).

andygrove (approved):
LGTM. Thanks @peterxcli

PR #5135, opened 2026-07-29
fix: support CalendarIntervalType hashing

### LINKED ISSUE #5059: Hashing a CalendarInterval value fails with "Unsupported data type in hasher: Interval(MonthDayNano)"
### Describe the bug

Hashing a `CalendarIntervalType` value fails the query with a native error instead of either working or falling back to Spark:

```
org.apache.comet.CometNativeException: Unsupported data type in hasher: Interval(MonthDayNano).
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/datafusion/issues
	at org.apache.comet.CometExecIterator.$anonfun$getNextBatch$2(CometExecIterator.scala:155)
	...
```

#4898 added `CalendarIntervalType` to `QueryPlanSerde.supportedDataType`:

https://github.com/apache/datafusion-comet/blob/1d3044f35bda457a20f1ccf4324a4699b35f5158/spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala#L530-L534

That is the gate `hash.scala` consults when deciding whether a hash expression's child type is supported:

https://github.com/apache/datafusion-comet/blob/1d3044f35bda457a20f1ccf4324a4699b35f5158/spark/src/main/scala/org/apache/comet/serde/hash.scala#L136-L147

So `Murmur3Hash` / `XxHash64` over a CalendarInterval child now reports `Compatible` and is serialized to native, but `native/spark-expr/src/hash_funcs/utils.rs` has no branch for `Interval(MonthDayNano)` and falls through to the catch-all `DataFusionError::Internal("Unsupported data type in hasher: ...")`.

The ANSI interval types are not in `supportedDataType`, so `hash(dt_interval)` and `hash(ym_interval)` correctly fall back to Spark. Only `CalendarIntervalType` reaches native and fails.

### Steps to reproduce

Against a Parquet-backed table so the expression is not constant-folded away:

```sql
CREATE TABLE t(d int, h int, y int, m int) USING parquet;
INSERT INTO t VALUES (1, 2, 1, 2), (0, 0, 0, 3);

SELECT hash(make_interval(y, m, 0, d, h, 0, 0)) FROM t;
```

Equivalently, as a sql-file test under `spark/src/test/resources/sql-tests/expressions/datetime/`.

A `OneRowRelation` form reproduces it too:

```sql
SELECT hash(make_interval(1, 2, 3, 4, 5, 6, 7.008009));
```

Physical plan at failure time (the whole projection is native):

```
CometNativeColumnarToRow
+- CometProject [hash(make_interval(1, 2, 3, 4, 5, 6, cast(7.008009 as decimal(18,6)), false), 42) ...]
   +- CometSparkRowToColumnar
      +- *(1) Scan OneRowRelation[]
```

### Expected behavior

The query should return the same values as Spark. Failing that, it should fall back to Spark rather than aborting the stage.

Two possible fixes:

1. Implement Spark-compatible hashing for `Interval(MonthDayNano)` (and `Interval(YearMonth)` / `Duration(Microsecond)` while we are there) in `native/spark-expr/src/hash_funcs/utils.rs`. Spark's `HashExpression` hashes a `CalendarInterval` as its `months`, `days` and `microseconds` fields, so the native side needs to match that field order and width exactly.
2. Or narrow the Scala gate so `CalendarIntervalType` is rejected for hash expressions, matching how the ANSI interval types behave today, and leave native hashing for the follow-up that implements it properly.

Either way this needs a regression test covering `hash` and `xxhash64` over all three interval types.

### Additional context

- Found while auditing interval support after #4898 and #4976 landed. Part of #4540.
- Reproduced on `1d3044f35` with the Spark 4.1 profile.
- `QueryPlanSerde.supportedDataType` is also consulted by `CometScalarSubquery` and `CometSink`, so it is worth checking whether a CalendarInterval-typed scalar subquery hits a similar unhandled-type path in the planner.
- Related gap worth noting separately: native shuffle accepts all three interval types after #4976, but `supportedHashPartitioningDataType` in `CometShuffleExchangeExec` rejects them, so hash partitioning on an interval key still falls back. That is a fallback rather than a failure, so it is not part of this issue.

peterxcli:

Closes #5059.

## Rationale for this change

`CalendarIntervalType` is accepted by the Scala hash-expression compatibility check, but the shared native hasher did not support Arrow's `Interval(MonthDayNano)` representation. As a result, native `hash` and `xxhash64` expressions failed at runtime instead of matching Spark.

- Add shared native hashing for `Interval(MonthDayNano)`, covering both Murmur3 and xxHash64.
- Convert Arrow nanoseconds back to Spark microseconds and match Spark's generated hash code, including its omission of the days field.
- Add a focused regression test covering both hash functions, nulls, negative intervals, and days-only intervals.

- `make core`
- `./mvnw test -Dtest=none -Dsuites="org.apache.comet.CometHashExpressionSuite calendar interval" -Dscalastyle.skip=true` (1 test passed)
- `cargo fmt --manifest-path native/Cargo.toml --all --check`
- `cargo clippy --manifest-path native/Cargo.toml -p datafusion-comet-spark-expr --lib --tests -- -D warnings`


native/spark-expr/src/hash_funcs/utils.rs
```diff
@@ -706,6 +706,23 @@ macro_rules! create_hashes_internal {
...
                 }
+                DataType::Interval(arrow::datatypes::IntervalUnit::MonthDayNano) => {
+                    let array = col
+                        .as_any()
+                        .downcast_ref::<IntervalMonthDayNanoArray>()
+                        .unwrap();
+                    for (hash, value) in $hashes_buffer.iter_mut().zip(array.iter()) {
+                        if let Some(value) = value {
+                            // Match Spark's generated hash code, which omits the days field:
+                            // https://github.com/apache/spark/blob/710b3c45aab88bd14e51d49f400e2f31e3b65772/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/hash.scala#L420-L423
```
andygrove: Good find on the field order, but the link needs a rethink. `710b3c45` is a master snapshot, and Spark master changed this eight days ago in SPARK-58236 (`a2616be0f`, 2026-07-22, "Fix hash codegen dropping the CalendarInterval days field"), which adds the missing `days` round to the codegen path so it matches the interpreted path. So the linked lines no longer say what the comment says they say.

The PR is still correct for everything we support: `days` is omitted in `v3.4.3`, `v3.5.8`, `v4.0.1`, `v4.1.2`, `v4.2.0` and on `branch-4.0` / `branch-4.1` / `branch-4.2`, and the fix has not been backported.

Could we point at a release tag such as `v4.2.0` instead of a master SHA, and name SPARK-58236 in the comment so the next person understands this becomes version-dependent once we support a Spark that carries the fix? Worth filing a follow-up issue for that too. Per our convention the version switch would live in `ShimCometConf` and travel through proto rather than becoming a native shim.

native/spark-expr/src/hash_funcs/utils.rs
```diff
@@ -706,6 +706,23 @@ macro_rules! create_hashes_internal {
                         $hash_method
                     );
                 }
+                DataType::Interval(arrow::datatypes::IntervalUnit::MonthDayNano) => {
+                    let array = col
```
andygrove: Would you mind pulling this into a `hash_array_interval_month_day_nano!` macro alongside the other `hash_array_*` macros at the top of the file? The neighbours all split a `null_count() == 0` fast path from the null-checking one, and they use the `unwrap_or_else(|| panic!("Failed to downcast column to {}. Actual data type: {:?}", ...))` form rather than a bare `.unwrap()`, which gives a much better message if a downcast ever goes wrong. This hasher runs on the shuffle and join hot path, so the fast path is worth having.

Small related nit: if you add `IntervalUnit` to the `use arrow::datatypes::{DataType, TimeUnit};` inside the macro, the match arm can read `DataType::Interval(IntervalUnit::MonthDayNano)` like the `TimeUnit` arms above it.

spark/src/test/scala/org/apache/comet/CometHashExpressionSuite.scala
```diff
@@ -128,6 +130,24 @@ class CometHashExpressionSuite extends CometTestBase with AdaptiveSparkPlanHelpe
     }
   }
 
+  test("hash - calendar interval") {
```
andygrove: Could this go in a SQL file test instead? There is already a `sql-tests/expressions/datetime/calendar_interval.sql` using the `VALUES` plus `make_interval` pattern with `spark.comet.exec.localTableScan.enabled=true`, and a `sql-tests/expressions/hash/hash.sql`, so either would host it. That would also cover the exact repro from #5059 (`SELECT hash(make_interval(y, m, 0, d, h, 0, 0)) FROM t`), which the current test does not go through.

spark/src/test/scala/org/apache/comet/CometHashExpressionSuite.scala
```diff
@@ -128,6 +130,24 @@ class CometHashExpressionSuite extends CometTestBase with AdaptiveSparkPlanHelpe
...
+    withTempView("t") {
+      val rows = Seq(
+        Row(new CalendarInterval(14, 25, 18367008009L)),
+        Row(new CalendarInterval(0, 1, 0L)),
+        Row(new CalendarInterval(0, 2, 0L)),
+        Row(new CalendarInterval(-14, -25, -18367008009L)),
+        Row(null))
```
andygrove: Good set of values, especially the days-only pair that pins the omitted field. A few more worth adding.

An all-zero interval is the one I would most want, because `Murmur3_x86_32.hashInt(0, seed) != seed`, so a zero interval still shifts the seed chain and that is exactly where a wrong fold order would show up.

Then months-only and microseconds-only, a non-default seed such as `hash(c, 1)` and `hash(c, 0)`, and a multi-argument form like `hash(c, someInt)` to exercise the seed chaining across columns.

And since the list and struct paths recurse back into this macro, `hash(array(c))` and `hash(struct(c))` should work now too, so a query for those would lock that in.

andygrove (commented):
Thanks for picking this up. The fix is small and the field order is right, and I confirmed against the Spark source that `genHashCalendarInterval` omits `days` in `v3.4.3`, `v3.5.8`, `v4.0.1`, `v4.1.2`, `v4.2.0` and on all the release branches, so this matches every Spark version we support. I also checked the round trip: `IntervalMonthDayNanoWriter` and the codegen kernel both use `Math.multiplyExact(microseconds, 1000L)`, so `nanoseconds / 1_000` recovers the original microseconds exactly or throws rather than wrapping. Nice.

A couple of things worth surfacing beyond the inline comments.

Spark's interpreted path has always folded all three fields. `InterpretedHashFunction.hash` does `hashInt(c.months, hashInt(c.days, hashLong(c.microseconds, seed)))` in every version including 3.5. Matching codegen is the right call here since that is what runs at execution time, but it does mean we are now reproducing a Spark inconsistency that Spark has since decided to fix. See the inline comment on the link.

One thing I liked: because `hash_list_with_primitive_elements!` falls through to the recursive path for unrecognised element types, and `DataType::Struct` recurses through `$recursive_hash_method`, `array<interval>` and `struct<interval>` should start working from this change too. Worth a query or two so that stays covered.

`supportedHashPartitioningDataType` in `CometShuffleExchangeExec.scala` deliberately leaves `CalendarIntervalType` out, so native shuffle will not hash-partition on an interval key. Good, that keeps the `days` divergence away from partitioning correctness, which is where it would actually be dangerous. Not asking you to change that.

Finally, `docs/source/user-guide/latest/datatypes.md` still says "Interval types fall back to Spark today" and marks `CalendarIntervalType` as 🔜. That was already drifting after #4898 and this moves it further along. Happy for it to be a follow-up rather than scope creep here.

One verification note for myself before merge: it would be good to confirm the new test actually fails without the Rust arm. I believe it does, since `getSupportLevel` returns `Compatible()` so the expression really is serialized to native, but our JVM codegen dispatch kernel does support `CalendarIntervalType`, so worth a quick check that nothing quietly absorbs the failure.

CI is fully green across all five Spark profiles.

spark/src/test/scala/org/apache/comet/CometHashExpressionSuite.scala
peterxcli: remove this test and added sql test in https://github.com/apache/datafusion-comet/pull/5135/changes/d069c8a8d195d7b19695b0d269ad911bb9af577d

native/spark-expr/src/hash_funcs/utils.rs
peterxcli: - pulled this into `hash_array_interval_month_day_nano!` macro
- added `IntervalUnit` to the `use arrow::datatypes::{DataType, TimeUnit};`

native/spark-expr/src/hash_funcs/utils.rs
peterxcli: yes, pointed to v4.2.0

spark/src/test/scala/org/apache/comet/CometHashExpressionSuite.scala
peterxcli: included in new sql test.

andygrove:

Good find, and the fix is in the right place. Matching Spark's omission of the `days` field is the kind of detail that would be very easy to get wrong, and I like that the test explicitly covers days-only intervals, since that is where the omission is observable as a collision.

Two things I would like to see addressed.

**This will be wrong on Spark 4.3**

The comment says SPARK-58236 starts including `days` in the hash from Spark 4.3. So as written, the native hasher is correct for every Spark version Comet supports today and silently wrong for the next one, and the failure mode is a hash mismatch, which means wrong shuffle partitioning and wrong hash join results rather than an error.

Comet's convention for this is to resolve the version-dependent policy in Scala (`ShimCometConf`) and pass the resolved boolean through the protobuf, rather than branching per version in Rust. Could this follow that pattern, with an `include_days` flag on the hash expression? Even if 4.3 is not supported yet, wiring the flag now is much cheaper than discovering the mismatch later, and there is already a `spark-4.2` bring-up in flight.

At minimum, if you would rather defer, please file an issue and reference it in the comment so this does not get found the hard way.

**Nanosecond-to-microsecond conversion**

`value.nanoseconds / 1_000` is an integer division that truncates toward zero. That is exact as long as the nanoseconds field is always a multiple of 1000, which it should be since Spark's `CalendarInterval` is microsecond-based. But nothing enforces that, and for a negative value with a non-multiple the truncation direction differs from a floor, giving a hash that does not match Spark.

More importantly, is there a range concern? Spark stores `CalendarInterval.microseconds` as an `i64`. Arrow's `IntervalMonthDayNano` stores nanoseconds, also `i64`, so any interval longer than roughly 292 years cannot round-trip. If that truncation happens earlier in the pipeline, the hash here is computed on an already-wrong value. Is that a pre-existing limitation that is documented somewhere? If so, pointing at it from this macro would help. If not, it is worth knowing whether `make_interval` can produce one.

peterxcli:
@andygrove Thanks for the pass — responses to both points.

**Spark 4.3 / `include_days` flag:** I've filed #5498 to track the version switch. I'd rather not wire the flag in this PR: every supported profile (3.4.3 through 4.2.0 — the SPARK-58236 fix is not on `branch-4.2`) omits `days`, so the flag would have a single value across all of CI and the `include_days = true` arm would be untestable dead code until a Spark 4.3 bring-up exists. The issue captures the `ShimCometConf`-plus-proto design so it's cheap to pick up as part of 4.3 support, and the code comment next to the SPARK-58236 note points at it.

**`nanoseconds / 1_000`:** the division is exact by construction. Both JVM-to-Arrow producers — `ArrowWriters.CalendarIntervalWriter` and the codegen dispatch kernel — convert with `Math.multiplyExact(microseconds, 1000L)`, so the nanoseconds field is always an exact multiple of 1000 (the negative-truncation case can't occur), and an interval whose time component exceeds ~292 years throws `ArithmeticException` at conversion time instead of reaching the hasher with a wrapped value. This matches what you verified in your July review ("recovers the original microseconds exactly or throws rather than wrapping"). I've added a comment on the macro documenting the invariant and where it's enforced, per your suggestion.

andygrove (approved):
Thanks for working through both rounds. The macro, the `v4.2.0` link with SPARK-58236 and #5498, and the documented `nanoseconds / 1_000` invariant cover what I asked for. The second query is the one that really pins the fix, since `DISTRIBUTE BY n` keeps `hash(c)` above a shuffle where the native hasher has to run.

The first query doesn't reach native, though. It hashes `make_interval(y, m, 0, d, h, 0, 0)` straight over an inline `VALUES` table, so `ConvertToLocalRelation` evaluates the whole projection in the optimizer and both runs just scan the precomputed rows. #5059's repro used a Parquet table for that reason. Could this one read from a Parquet table instead, so it covers `make_interval` over columns feeding the native hash?

Issue #4789, opened 2026-07-01
bug: array/map kernels crash on non-null child field via CometLocalTableScanExec


mbutrovich:
### Describe the bug

Several Comet array/map expressions crash natively when the input comes through `CometLocalTableScanExec` with a non-nullable child field (array element, or map key/value). An in-memory `Seq[Int]` / `Map[Int, Int]` column encodes with `containsNull=false` / `valueContainsNull=false`, and the local scan carries that non-null child into the DataFusion kernel, which disagrees with Comet's planned output type. Reproduced locally (Spark 4.0) on two kernels:

- `spark_array_slice` (`slice`): `Assertion failed: result_data_type == *expected_type: Function 'spark_array_slice' returned value of type 'List(non-null Int32)' while the following type was promised at planning time and expected: 'List(Int32)'.`
- ListArray/Struct build (`map_entries`): `InvalidArgumentError("ListArray expected data type Struct("key": non-null Int32, "value": Int32) got Struct("key": non-null Int32, "value": non-null Int32) for "item"")` (panics in `datafusion-functions-nested` `map_entries.rs:126`)

A third signature appeared in CI (Spark 4.1) on `ArrayInsert` (`array prepend`, SPARK-41233) and is very likely the same root cause, though it did not reproduce locally with a scalar `array_insert` on Spark 4.0: `Type mismatch in ArrayInsert: array type is List(Field { data_type: Int32 }) but item type is List(Field { data_type: Int32, nullable: true }).`

It does NOT reproduce over a native Parquet scan (which normalizes children to nullable) or with a SQL `array(...)`/`map(...)` literal (literal children are nullable), so it is specific to the local scan path.

### Steps to reproduce

`ConvertToLocalRelation` must be disabled or the optimizer folds the expression over the `LocalRelation` at plan time and nothing executes natively. Add to a suite extending `CometTestBase`:

```scala
import testImplicits._

private def withLocalTableScanNoFold(f: => Unit): Unit = {
  withSQLConf(
    CometConf.COMET_EXEC_LOCAL_TABLE_SCAN_ENABLED.key -> "true",
    "spark.sql.optimizer.excludedRules" ->
      "org.apache.spark.sql.catalyst.optimizer.ConvertToLocalRelation") {
    f
  }
}

test("slice on non-null element array") {
  withLocalTableScanNoFold {
    val df = Seq(Seq(1, 2, 3), Seq(4, 5)).toDF("x")
[...]

Issue #4528, opened 2026-05-30
CreateArray with nullability-divergent children panics in native make_array


schenksj:
### Problem

DataFusion's `make_array` asserts strict element-type equality in `MutableArrayData::with_capacities` and panics on a mismatch. Spark's `CreateArray` is more permissive: its type coercion compares element types with `sameType`, which ignores nullability, so children that share a surface type but differ only in a nested **struct field's** nullability get no unifying cast. Native execution then panics inside `make_array_inner`:

```
native panic: assertion `left == right` failed: Arrays with inconsistent types passed to MutableArrayData
```

Reproducible with `array(struct(id, ct = lit("a")), struct(id, ct = <nullable expr>))` — one arm's `ct` is non-nullable (literal), the other nullable.

### Scope (narrower than "any nullability difference")

DataFusion **tolerates** container nullability differences — an `ArrayType.containsNull` / `MapType.valueContainsNull` mismatch is coerced and runs natively (e.g. `array(array(int) non-null-elem, array(int) nullable-elem)`). Only a **struct field's** nullability actually panics. So the decline must not fire on the container-only case, or it over-declines and loses native execution for legitimate arrays-of-arrays/maps.

### Proposed fix

In `CometCreateArray`, decline serialization (`withFallbackReason`) only when the children still differ **after normalizing container nullability** (force `ArrayType.containsNull` / `MapType.valueContainsNull` to `true`, keep struct field nullability significant). Those fall back to Spark's JVM evaluator, which has no such strictness.

This tracks upstream apache/datafusion#22366; the caller-side decline can be removed once that fix lands.

### Relationship to the Delta integration

Standalone guard against a native panic. It is **surfaced by** the in-progress Delta Lake contrib integration (Delta's CDC write path builds one struct per change type, leaving `_change_type` nullability divergent across arms), so it would help to prioritize it alongside that work.

Issue #4515, opened 2026-05-29
DataFusion / DataFusion-Spark functions whose Arrow return type drifts from Spark catalyst's declared type


mbutrovich:
I'm running into this in multiple PRs that try to reduce the FFI deep copies in the system (#4393, #4507) so I need to start tracking it since I keep introducing workarounds on those branches.

## Background

Comet serializes Spark catalyst types into a proto schema, and the native planner expects DataFusion physical expressions to produce arrays matching those Arrow types. Several DataFusion / `datafusion-spark` functions return arrays whose Arrow type does not match what Spark catalyst declares.

On `main`, this is masked because batches cross JVM-to-native FFI boundaries inside a stage. Each crossing deep-copies and re-advertises the stream schema, and the consuming `ScanExec.build_record_batch` casts to the declared schema. Together they reshape `Timestamp(us)` -> `Timestamp(us, "UTC")`, `List(nullable Int32)` -> `List(non-null Int32)`, etc. before validation runs.

Reducing FFI hops removes both the deep-copy re-stamp and the scan cast, so the drift surfaces as `Invalid argument error: column types must match schema types, expected ... but found ...`. PRs that hit this work around it with a `Projection(Cast)` and a warning per drifting column. The right fix is to correct the function return types upstream so the cast becomes a no-op.

## Related issue

- DataFusion [apache/datafusion#22602](https://github.com/apache/datafusion/issues/22602): `width_bucket` returns `Int32` instead of `Int64`.

## Observed mismatches

| Spark expression | Catalyst declares | DataFusion / datafusion-spark produces |
| --- | --- | --- |
| `width_bucket` | `LongType` (Int64) | `Int32` |
| `date_trunc(unit, ts)` | `Timestamp(us, "UTC")` | `Timestamp(us)` (no timezone) |
| `collect_set(int)` | `List(non-null Int32)` | `List(nullable Int32)` |

Not exhaustive. Each warning emitted by the workaround cast layer on a branch that reduces FFI hops is a candidate to add here.

## Other victims of the same drift

[...]

PR #5051, opened 2026-07-27
feat: add experimental native support for in-memory cache, disabled by default

andygrove:
## Which issue does this PR close?

Closes #2391.

## Rationale for this change

This PR continues the work started by @pchintar in #4591, who is no longer able to work on it. All credit for the original implementation goes to them. This branch is based on their branch with latest `apache/main` merged in, and picks up the remaining review feedback and CI failures.

Comet currently has limited support for Spark's in-memory cache.

When a table is cached and later read, the cached data cannot be consumed directly by Comet operators. Instead, the execution plan falls back to Spark's cache scan path and introduces an additional `CometSparkColumnarToColumnar` conversion before execution can continue in Comet.

This extra conversion adds overhead to cached table scans and prevents cached data from remaining on a native Comet execution path.

## What changes are included in this PR?

A native cache path for in-memory cached tables behind a new configuration:

`spark.comet.exec.inMemoryCache.enabled`

When enabled:

* Cached data is stored using a Comet-specific cache serializer, one compressed Arrow IPC stream per column, so a scan decodes only the columns it projects.
* Cached data is represented as `CometCachedBatch`.
* Cached tables are scanned using `CometInMemoryTableScanExec`.
* Cached data can be consumed directly by Comet operators without introducing a `CometSparkColumnarToColumnar` conversion.
* Per-batch column statistics (lower bound, upper bound, null count, row count) are written in the format expected by `SimpleMetricsCachedBatchSerializer`, so Spark's `buildFilter` can prune cached batches before they are decoded.

When disabled:

* Spark's existing cache serializer continues to be used.
* Existing cache scan behavior is preserved.

## How are these changes tested?

`CometInMemoryCacheSuite` covers:

- Comet-native cache scan over `CometCachedBatch`
- Fallback behavior when native cache support is disabled
- Multi-partition cached tables
- Empty cached tables
- Projection-only cache reads
- Shuffle execution after cached table scans
- Stats-based batch pruning
- Empty projection scans (`SELECT count(*)`)
- Floating-point pruning with NaN values
- Fallback read path for Spark `DefaultCachedBatch`
- One stream stored per cached column, each carrying data
- A projected read decoding only the projected columns
- A row-count-only read decoding no columns at all
- The scan asking for one cheap column, not the whole schema, when a query needs only the row count
- Per-column sizes reported in the statistics row

The two projection tests corrupt the streams a read must not touch, so they assert on what was decoded rather than on timings, and each is checked to fail against the previous single-stream format.

[...]

PR #5138, opened 2026-07-29
fix: normalize nested field nullability in ShuffleScanExec and ExpandExec

### LINKED ISSUE #5137: ShuffleScanExec and ExpandExec assert on nested field nullability instead of normalizing it
### Describe the bug

Comet has two independent sources of truth for a column's Arrow type:

1. **Declared** — Spark catalyst's type, serialized per nested field over protobuf (`StructInfo.field_nullable`, `ListInfo.contains_null` in `native/proto/src/proto/types.proto`) and rebuilt by `to_arrow_datatype` (`native/core/src/execution/serde.rs`).
2. **Actual** — the Arrow type of the array a kernel, scan, or shuffle block really produced.

Arrow treats nested field nullability as part of `DataType` identity: `DataType::equals_datatype` compares `is_nullable()` at every list element, struct field, and map entry, and `RecordBatch::try_new_with_options` has no option to relax it (`match_field_names: false` only ignores field *names*). So any operator that **stamps** the declared schema onto a runtime array hard-fails when the two disagree by a single nested `nullable` flag, e.g.:

```
Invalid argument error: column types must match schema types,
expected List(Struct("id": Int64, "flag": Boolean))
but found List(Struct("id": Int64, "flag": non-null Boolean))
at column index 0
```

The data is fine in this situation — a non-nullable child is a strict subset of a nullable one, and the batch would produce identical results. It is a metadata assertion failure, not corruption.

Most boundaries already **normalize** instead of asserting, and that is why this drift is usually invisible:

- `ScanExec::build_record_batch` (`native/core/src/execution/operators/scan.rs`) casts every imported column to the declared type. Arrow's cast rewrites nested fields from the target (`cast_struct_to_struct` builds with `to_fields`, `cast_list_values` with `to.clone()`), so the FFI boundary absorbs nullability drift.
- `SchemaAlignExec` (`native/shuffle/src/schema_align.rs`) casts the shuffle writer's input, added for #4515.
- `CometLocalTableScanExec` widens its output with `.asNullable`, added for #4789.

Two stamp points still assert:

- `ShuffleScanExec::poll_next` (`native/core/src/execution/operators/shuffle_scan.rs`) — stamps the catalyst schema onto the decoded shuffle block with no cast. This is the direct-read shuffle path, enabled by default (`spark.comet.shuffle.directRead.enabled`).
- `ExpandStream::expand` (`native/core/src/execution/operators/expand.rs`) — the schema is derived from `projections[0]` only (see `OpStruct::Expand` in `native/core/src/execution/planner.rs`) and then stamped onto the output of *every* projection, so the operator fails if any two projections disagree on a nested nullable flag for the same output column.

Because the surviving stamp points are both inside native plans (past `ScanExec`'s cast), the failure mode is: a native kernel or non-Parquet source produces a nested type that is narrower than what catalyst declared, nothing corrects it, and a downstream `ShuffleScanExec` or `ExpandExec` aborts the task. Retries do not help; it is deterministic for the partition's data shape.

Upstream nested-nullability drift is tracked by #4515 (with concrete kernel instances in #4789 and #4528). This issue is about the boundaries: the same drift should be absorbed rather than fatal, the way the other boundaries already do.

A secondary problem: the arrow error text names neither the operator nor the column name, only `at column index N`. For a wide schema of deeply nested structs this makes the report very hard to act on, since the two printed types can differ by one flag hundreds of characters in.

### Steps to reproduce

Not yet reduced to a standalone test. Both sites need a native plan in which a column's actual nested nullability differs from the catalyst-declared type:

- `ShuffleScanExec`: a native plan reading shuffle output where the block's nested field nullability differs from the consuming stage's declared type.
- `ExpandExec`: a grouping-sets / rollup / multiple-distinct-aggregate plan over an `array<struct<...>>` column where one projection's expression yields a different nested nullable flag than `projections[0]`.

Writing the tests is part of the work here. The `SchemaAlignExec` unit tests are a good model — they pin each drift pair so the test flips to a passthrough assertion once the upstream type is corrected.

### Expected behavior

Neither boundary should fail on a nested nullability difference:

- `ShuffleScanExec` should reconcile the decoded block with the declared schema the way `ScanExec` does (cast per column, or reuse `SchemaAlignExec`'s alignment logic on the read side).
- `ExpandExec` should build its output schema so that it is valid for every projection, not just `projections[0]` — e.g. widen nested nullability across all projections, or align each projection's output to the operator's declared schema before stamping.

Additionally, the error raised when a stamp genuinely cannot be reconciled should name the operator and the offending column path, not just the column index.

### Additional context

Related: #4515 (epic: return-type drift from DataFusion / datafusion-spark), #4789 (non-null nested child fields from local table scans), #4528 (`make_array` with nullability-divergent struct children).

andygrove:

Closes #5137.

## Rationale for this change

Arrow treats nested field nullability as part of `DataType` identity, so `RecordBatch::try_new_with_options` rejects a column whose nested `nullable` flags are narrower than the declared type — even though the data is fine, since a non-null child is a strict subset of a nullable one. It is a metadata assertion failure, not corruption.

Most Comet boundaries already normalize instead of asserting, which is why this drift is usually invisible:

- `ScanExec::build_record_batch` casts every imported column to the declared type, so the FFI boundary absorbs it.
- `SchemaAlignExec` casts the shuffle writer's input (added for #4515).
- `CometLocalTableScanExec` widens its output with `.asNullable` (added for #4789).

Two stamp points still asserted, and because both sit inside native plans (past `ScanExec`'s cast) nothing corrected the drift before them. The task aborted deterministically for the partition's data shape, so retries did not help:

```
Invalid argument error: column types must match schema types,
expected List(Struct("id": Int64, "flag": Boolean))
but found List(Struct("id": Int64, "flag": non-null Boolean))
at column index 0
```

- `ShuffleScanExec::poll_next` stamped the catalyst schema onto the decoded shuffle block with no cast. This is the direct-read shuffle path, enabled by default (`spark.comet.shuffle.directRead.enabled`).
- `ExpandStream::expand` derived its schema from `projections[0]` only and stamped it onto the output of *every* projection, so it failed if any two projections disagreed on a nested nullable flag for the same output column.

A new `cast_and_stamp_schema` helper in `datafusion-comet-common`, used at both stamp points, which casts any column whose Arrow type differs from the declared field type before stamping. Arrow's cast rewrites nested fields from the target (`cast_struct_to_struct` builds with `to_fields`, `cast_list_values` with `to.clone()`), which is exactly how `ScanExec` already absorbs this.

`ExpandExec::build_schema` now derives the output schema by widening nested nullability across **all** projections instead of reading it off `projections[0]`, so the declared schema is valid for every projection. The runtime cast is a second, independent line of defence: each is tested on its own, so neither silently carries the other.

Error messages now name the operator and the dotted path of the offending column:

```
CometExpandExec cannot reconcile col[0] with its declared schema at
col_0.element.flag: expected nullable Boolean, found non-null Boolean
```

instead of only `at column index N`, which for a wide schema of deeply nested structs was very hard to act on — the two printed types could differ by one flag hundreds of characters in.

A narrowing cast (declared non-null child, actual data contains nulls) is not silently accepted: `StructArray::try_new` rejects unmasked nulls for a non-nullable field, so that case still errors loudly, now with the column path named.

This PR is about the boundaries. The upstream nested-nullability drift itself remains tracked by #4515, with concrete kernel instances in #4789 and #4528.

New Rust unit tests, modelled on the existing `SchemaAlignExec` tests as the issue suggested:

- `native/common/src/schema.rs` — 9 tests over `cast_and_stamp_schema`, `describe_type_mismatch`, and `widen_nested_nullability`. `stamps_nested_nullability_drift` first asserts that the plain stamp *does* reject the drift, then that the helper absorbs it with values intact. That assertion pins why the helper exists: if arrow ever relaxes nested nullability comparison, it flips and the call sites can go back to a plain stamp.
- `native/core/src/execution/operators/expand.rs` — 5 tests: `build_schema` widens across projections (driven from the narrow side, so it cannot pass by echoing `projections[0]`), rejects projections of differing width, expands divergent projections end-to-end, and reconciles even when handed the narrow `projections[0]`-derived schema the planner used to build.
- `native/core/src/execution/operators/shuffle_scan.rs` — 2 tests: a decoded block with a non-null nested child against a nullable declared type is reconciled with values preserved, and an unreconcilable column names the operator and column.

I confirmed these are genuine regression tests rather than tests that pass either way: without the fix, `RecordBatch::try_new_with_options` rejects the drift, which the first assertion in `stamps_nested_nullability_drift` now pins permanently.

Also run and passing: full `cargo test --workspace` (135 + 33 native tests; the 3 pre-existing `datafusion-comet-jni-bridge` `errors::tests` panic-handling failures reproduce on a clean checkout of this branch point and are unrelated), `cargo clippy --all-targets --workspace` clean, `cargo fmt`, and on Spark 3.5: `CometExecSuite "expand operator"`, `CometNativeShuffleSuite` (27), `CometShuffleSuite` (44), `CometAggregateSuite` (85, 2 canceled).

I did not reduce this to a Spark-level SQL test. Both sites need a native plan where a kernel actually produces a nested type narrower than catalyst declared, and those kernels are the subject of #4515/#4789/#4528 — so a SQL repro would be pinned to whichever kernel currently drifts and would evaporate when that kernel is fixed, which is the wrong thing to encode. The tests above pin the boundaries directly, independent of which kernel drifts.

## Are these changes tested?

Yes.

No API changes. Queries that previously aborted with `column types must match schema types` in the direct-read shuffle path or in Expand (grouping sets / rollup / multiple distinct aggregates over nested columns) now complete. Errors that remain genuinely unreconcilable are reported with the operator and column path named.

## Follow-up

`ScanExec::build_record_batch` does the same cast-then-stamp but keeps its own copy of the logic and its own `cast_time` metric, so its errors still report only `at column index N`. Switching it to the shared helper would need the helper to accept a timer; I left it out to keep this diff focused on the two boundaries the issue names.

### DIFF
--- a/native/common/src/schema.rs
+++ b/native/common/src/schema.rs
@@ -39,10 +39,17 @@ use std::sync::Arc;
 /// This is the normalizing counterpart to stamping `schema` on directly: it absorbs the
 /// return-type drift — most commonly a nested `nullable` flag — that native kernels and non-Parquet
 /// sources introduce, the same way `ScanExec` absorbs it at the FFI boundary.
+///
+/// Reconciliation follows `schema` in both directions, so it will also narrow a nullable nested
+/// child to non-null when that is what `schema` declares. That is not silently lossy: arrow's
+/// `StructArray::try_new` rejects unmasked nulls under a non-nullable field, so data that cannot
+/// survive the narrowing errors here rather than producing an array that misreports itself.
+/// Callers that must not narrow should widen `schema` first — see [`widen_nested_nullability`],
+/// which is what `SchemaAlignExec` and `ExpandExec` do.
 pub fn cast_and_stamp_schema(
     operator: &str,
     schema: &SchemaRef,
-    columns: &[ArrayRef],
+    mut columns: Vec<ArrayRef>,
     num_rows: usize,
 ) -> Result<RecordBatch, DataFusionError> {
     if columns.len() != schema.fields().len() {
@@ -53,44 +60,28 @@ pub fn cast_and_stamp_schema(
         )));
     }
 
-    let mut aligned: Vec<ArrayRef> = Vec::with_capacity(columns.len());
-    // A cast yields exactly the target type, so after this loop every column's type matches the
-    // declared field. `residual_mismatch` records the one case that would not: a cast that returned
-    // something other than what it was asked for. Tracked here so the error below can name the
-    // column without having to keep `aligned` alive past the stamp.
-    let mut residual_mismatch: Option<(usize, DataType)> = None;
-    for (idx, (column, field)) in columns.iter().zip(schema.fields()).enumerate() {
-        if column.data_type() == field.data_type() {
-            aligned.push(Arc::clone(column));
-            continue;
+    for (idx, (column, field)) in columns.iter_mut().zip(schema.fields()).enumerate() {
+        if column.data_type() != field.data_type() {
+            *column = cast_with_options(column, field.data_type(), &CastOptions::default())
+                .map_err(|e| cast_error(operator, schema, idx, column.data_type(), e))?;
         }
-        let cast = cast_with_options(column, field.data_type(), &CastOptions::default())
-            .map_err(|e| stamp_error(operator, schema, idx, column.data_type(), e))?;
-        if residual_mismatch.is_none() && cast.data_type() != field.data_type() {
-            residual_mismatch = Some((idx, cast.data_type().clone()));
-        }
-        aligned.push(cast);
     }
 
+    // Every column's type now matches its declared field — `cast_with_options` returns either an
+    // error or an array of exactly the requested type — so the stamp can only fail on row counts.
     let options = RecordBatchOptions::new().with_row_count(Some(num_rows));
-    RecordBatch::try_new_with_options(Arc::clone(schema), aligned, &options).map_err(|e| {
-        match residual_mismatch {
-            Some((idx, actual)) => stamp_error(operator, schema, idx, &actual, e),
-            // Types all match, so the stamp can only have failed on row counts.
-            None => DataFusionError::Context(
-                format!(
-                    "{operator} cannot build a batch of {num_rows} rows with its declared schema"
-                ),
-                Box::new(DataFusionError::from(e)),
-            ),
-        }
+    RecordBatch::try_new_with_options(Arc::clone(schema), columns, &options).map_err(|e| {
+        DataFusionError::Context(
+            format!("{operator} cannot build a batch of {num_rows} rows with its declared schema"),
+            Box::new(DataFusionError::from(e)),
+        )
     })
 }
 
 /// Names the operator and the dotted path of the column that could not be reconciled, since
 /// arrow's own message reports only `at column index N` and the two printed types may differ by a
 /// single flag hundreds of characters in.
-fn stamp_error(
+fn cast_error(
     operator: &str,
     schema: &SchemaRef,
     idx: usize,
@@ -108,19 +99,16 @@ fn stamp_error(
 
 /// Describes where `expected` and `actual` first diverge, as a dotted path rooted at `path`.
 /// Returns `None` when the two types are equal.
-pub fn describe_type_mismatch(
-    path: &str,
-    expected: &DataType,
-    actual: &DataType,
-) -> Option<String> {
+///
+/// The nested arms cover the shapes Comet actually builds — see `make_all_fields_nullable` in the
+/// planner and `to_arrow_datatype` in the serde layer, which walk the same set.
+fn describe_type_mismatch(path: &str, expected: &DataType, actual: &DataType) -> Option<String> {
     if expected == actual {
         return None;
     }
     match (expected, actual) {
         (DataType::List(e), DataType::List(a))
-        | (DataType::LargeList(e), DataType::LargeList(a))
-        | (DataType::ListView(e), DataType::ListView(a))
-        | (DataType::LargeListView(e), DataType::LargeListView(a)) => {
+        | (DataType::LargeList(e), DataType::LargeList(a)) => {
             describe_field_mismatch(&format!("{path}.element"), e, a)
         }
         (DataType::FixedSizeList(e, e_len), DataType::FixedSizeList(a, a_len))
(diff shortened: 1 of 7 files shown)


native/core/src/execution/operators/expand.rs
```diff
@@ -65,6 +66,47 @@ impl ExpandExec {
...
+            }
+            for (data_type, expr) in data_types.iter_mut().zip(projection) {
+                *data_type = widen_nested_nullability(data_type, &expr.data_type(input_schema)?);
+            }
+        }
+
+        let fields: Vec<Field> = data_types
+            .into_iter()
+            .enumerate()
+            .map(|(idx, dt)| Field::new(format!("col_{idx}"), dt, true))
```
hsiang-c, line 110: Is it required to keep the field name as is?

andygrove: Not strictly required for correctness, but I kept it deliberately.

`col_{idx}` is exactly what the planner produced before this logic moved into `build_schema`, so keeping it means the only behavior change here is the nullability derivation. There is also no better name available: an Expand projection is an arbitrary `PhysicalExpr` (including literals like the grouping ID), so there is nothing to inherit — and parent operators bind to these columns by index, with a bound reference taking its name from the child schema (`planner.rs`, `ExprStruct::Bound`), so the name is positional either way. It is only visible in explain output.

Added a comment at the call site recording that.

hsiang-c (approved):
Thanks @andygrove LGTM

native/common/src/schema.rs
```diff
@@ -0,0 +1,378 @@
...
+/// messages.
+///
+/// This is the normalizing counterpart to stamping `schema` on directly: it absorbs the
+/// return-type drift — most commonly a nested `nullable` flag — that native kernels and non-Parquet
+/// sources introduce, the same way `ScanExec` absorbs it at the FFI boundary.
+///
+/// Reconciliation follows `schema` in both directions, so it will also narrow a nullable nested
+/// child to non-null when that is what `schema` declares. That is not silently lossy: arrow's
+/// `StructArray::try_new` rejects unmasked nulls under a non-nullable field, so data that cannot
+/// survive the narrowing errors here rather than producing an array that misreports itself.
```
parthchandra, line 46: I don't think this is covered in the test. Maybe add a test that stamps a nullable-child array carrying a real null against a non-null-declared schema and asserts it errors?

parthchandra (approved):
lgtm. One small suggestion (not blocking though).

native/common/src/schema.rs
andygrove: Thanks, will do as follow up. Merging this to unblock #5051

andygrove: Good catch, that claim was untested. Added in https://github.com/apache/datafusion-comet/pull/5285 — it stamps a nullable-child array against a non-null-declared schema both with and without a real null, so the child's `nullable` flag alone does not decide the outcome: the null-free case narrows fine, and the one carrying a null errors with the `c0.element.flag` path named.

PR #5139, opened 2026-07-29
chore: stop enabling incompatible casts in plan stability suite

andygrove:

N/A

## Rationale for this change

TPC-DS had many fallbacks due to incompatible casts from floating-point to decimal. These fallbacks did not show up in the stability suite because we enabled these  incompatible casts in the test. Now that cast supports codegen dispatch fallback, we no longer need to set this config.

- Stop setting config
- Updated golden files

CI


comphead (approved):
Thanks @andygrove the CI fails because of style check

PR #581, opened 2024-06-17
build: Drop Spark 3.2 support

huaxingao:
## Which issue does this PR close?

Closes [565](https://github.com/apache/datafusion-comet/issues/565).

PR #5141, opened 2026-07-29
chore: clean up Scala compiler warnings

andygrove:

No issue filed — this is a mechanical compiler-warning cleanup. Happy to file one if the project would prefer to track it.

## Rationale for this change

Building Comet produces a large number of Scala compiler warnings, which makes it easy to miss a genuinely new warning introduced by a change. Most of them are Scala 2.13 deprecations that have accumulated since the 2.13 / Spark 4.x profiles were added.

The cleanup has been split into reviewable pieces, one per warning category:

| Category | PR | Status |
|---|---|---|
| Symbol literals (`'col` → `$"col"`) in tests | #5168 | merged |
| `Array` → `Seq` round trips | #5170 | merged |
| `.toIterable` in protobuf builder calls | #5173 | open |
| Auto-application to `()` | #5175 | open |

This PR is now the remainder: the long tail of one-off warnings that don't group into a category of their own. It no longer overlaps with any of the above, so it can be reviewed and merged independently of #5173 and #5175 in either order.

Mostly mechanical 2.13 deprecation fixes:

- `Long + String` concatenation replaced with string interpolation in `ConfigHelpers.timeToString` / `byteToString`.
- `Map.filterKeys` (which returns a lazy, non-serializable view) replaced with an explicit `filter` in `CometNativeShuffleWriter`.
- `toIterator` → `iterator` in `CometBatchRDD.compute`.
- Two inferred existential types replaced with explicit types: the `null` / RDD branches in `CometNativeExec` are widened to `RDD[Any]` explicitly, and `CometIcebergRewriteActionSuite.invoke` uses an `isStatic` flag rather than a `Class[_]` / instance tuple whose type only unifies existentially.
- `case Some(_: Ordering[K])` → `case Some(_)` in `CometBlockStoreShuffleReader`; the type test was unchecked and the branch throws regardless.

Two changes are slightly more than mechanical:

- **`GenerateDocs.categoryPages`** triggered a group of *"adapting argument list by creating a 2-tuple"* warnings on 2.12, because `"array" -> ("array.md", () => ...)` parses as a two-argument call to `->`. Introducing `exprPage`/`aggPage` helpers removes both the ambiguity and ~40 lines of duplication. The generated docs are byte-identical.
- **`FuzzDataGenerator`** no longer uses the deprecated `RandomStringUtils.randomAlphabetic`. The replacement draws from the caller's seeded `Random`, so generated strings are now reproducible from the fuzz seed — previously they were not, since `RandomStringUtils` uses its own RNG.

`-processing` is added to the javac `-Xlint` list, silencing the *"No processor claimed any of these annotations"* warnings.

### Warnings deliberately left in place

Each of these needs a shim or a larger change, so they felt out of scope here:

1. **`AliasAwareOutputExpression`** — uses `Stream`, deprecated in 2.13. Can't be fixed without a shim: `multiTransformDown` returns `Stream` on Spark 3.x / 2.12 and `LazyList` on Spark 4.x / 2.13, and `LazyList` doesn't exist in 2.12. Worth flagging separately: **this trait appears to be dead code** — `AliasAwareOutputExpression`, `projectExpression`, and `generateCartesianProduct` have no references anywhere in the repo, and the file hasn't been touched since #581 (Spark 3.2 drop). Deleting it would remove these warnings; I left that decision to reviewers.
2. **`CometSQLQueryTestHelper.getErrorClass`** — renamed to `getCondition` in Spark 4.0, so it needs a shim.
3. **`CometTestBase.withRowGroupSize`** — parquet-version-specific overload; already carries a TODO saying it needs a shim.
4. **`CometTPCDSQuerySuite` early initializer** — the `extends { val tpcdsAllQueries = … }` block exists so that `TPCDSBase`'s constructor sees the query lists. Moving them to an `override val` in the class body would leave the parent reading `null` during construction.
5. **`CometParquetWriterSuite`** — still has `Array` → `Seq` warnings that #5170 didn't reach, because the file is only compiled under `-Pspark-3.5 -Pscala-2.13`, which isn't one of the profiles CI builds. Worth a follow-up, either the fix or adding that profile to CI.

One warning was intentionally not "fixed" as the compiler suggests: `CometFuzzTestSuite` keeps `.equals` rather than `==` when comparing default values, because `==` on boxed floats goes through cooperative equality where `NaN != NaN`, while `Float.equals` compares bits. It is restructured to compare statically typed `AnyRef`s so the lint no longer fires, with a comment explaining why.

No new tests — this is a warning cleanup with no intended behavior change.

- All five profiles `test-compile` cleanly: default (Spark 4.1 / 2.13), `-Pspark-3.4`, `-Pspark-3.5`, `-Pspark-3.5 -Pscala-2.13`, `-Pspark-4.0`.
- `GenerateDocs` output is unchanged (no diff under `docs/` after a full `package`), confirming the `categoryPages` refactor is output-identical.
- Existing suites covering the touched code paths pass — 502 tests: `CometFuzzTestSuite`, `DataGeneratorSuite`, `CometTemporalExpressionSuite`, `CometNativeShuffleSuite`, `CometShuffleSuite`, `CometWindowExecSuite`, `CometNativeReaderSuite`, `ParquetReadV1Suite`, `CometIcebergRewriteActionSuite`, `CometExecSuite`. (Run before the category PRs were split out; the split only reverts files back to `main`.)


andygrove:
I will likely break this out into smaller PRs, once we have 1.0 released.

mbutrovich (approved):
Thanks @andygrove!

PR #5152, opened 2026-07-30
feat: [branch-1.0] expose Comet version as spark.comet.version runtime config (#5049)

andygrove:

Backport of #5049 to `branch-1.0`.

## Rationale for this change

There is currently no convenient way for a Spark user to query which Comet version is loaded. The JVM `org.apache.comet.COMET_VERSION` constant exists but is only reachable via Scala imports, and the native library logs the version on init but that requires driver log access.

Clean cherry-pick of 3c830f4de (no conflicts):

- The Comet driver plugin sets `spark.comet.version` on the SparkConf during `init`, so it is queryable at runtime via `spark.conf.get("spark.comet.version")` or `SET spark.comet.version` in SQL. It is set before the off-heap check so the version is reported even when Comet is otherwise disabled.
- Documents both the new `spark.comet.version` runtime config and the existing `org.apache.comet.COMET_VERSION` (plus `COMET_BRANCH` / `COMET_REVISION`) programmatic accessors in the installation guide.

Existing test from the original PR: `CometPluginsSuite` (`Comet version is exposed as a Spark config`).


mbutrovich (approved):
Thanks for the backport PR, @andygrove!

PR #5154, opened 2026-07-30
docs: drop compatibility notes for bugs that are now fixed

### LINKED ISSUE #4959: cast string to boolean: trim ISO control bytes to match Spark's UTF8String.trimAll
### Describe the bug

Comet's `parse_utf8_to_boolean` (`native/spark-expr/src/conversion_funcs/string.rs`) uses Rust `str::trim`, which strips only Unicode `char::is_whitespace` code points. Spark's `UTF8String.trimAll()` (referenced by `Cast` for `stringToBool`) trims every byte where `Character.isWhitespace(b) || Character.isISOControl(b)` (`UTF8String.java:201-203`, `UTF8String.java:994-1010`).

The two sets differ on the ASCII control bytes `0x00-0x08`, `0x0E-0x1F`, and `0x7F`: Spark trims them, Comet does not.

### Repro (per @mbutrovich on #4914)

```
`\" \x01true\x01 \"` -> Spark casts to `true`, Comet casts to `null` (or ANSI-errors).
```

### Scope

Same divergence likely exists in other `trim`-based cast paths in `conversion_funcs/string.rs`. `UTF8String.trimAll` parity should be assessed systematically.

### Prior context

Reported in the review of #4914; that PR intentionally left the pre-existing behavior alone to keep it a pure perf change.

andygrove:

Documentation-only follow-up to #5085. Does not close an issue.

## Rationale for this change

#5085 documented the then-open `correctness`+`bug` issues. Three of those entries have since been fixed, and none of the follow-up PRs removed the note describing the bug:

- **#5065** was fixed by #5084 (`38c3f8bce`), but `CometIntegralDivide.getCompatibleNotes` still tells users that `Long.MinValue div -1` silently returns the wrapped value under ANSI.
- **#5066** was fixed by #5080 (`a1b022b52`), but `CometMakeDecimal.getCompatibleNotes` still says the native implementation ignores `nullOnOverflow`.
- **#5070** was fixed by #5082 (`e0de5820f`), but `CometRound.getCompatibleNotes` still says `LongType` input with a negative `scale` of `-19` or lower returns `0` instead of overflowing.

This matters because the per-category compatibility pages (`math.md`, `misc.md`, ...) are generated from these strings at docs-build time and are not committed to the tree, so the stale text publishes on the next docs build rather than showing up in a diff. #5127 already set the precedent by removing its `null IN ()` bullet as part of the fix.

Delete the now-inaccurate `getCompatibleNotes` overrides on `CometIntegralDivide` and `CometRound` (`arithmetic.scala`) and `CometMakeDecimal` (`decimalExpressions.scala`). No other changes.

### Dropped from this PR since it was opened

The original version of this PR also rescoped the `CAST(string AS boolean)` trim bullet in `compatibility/index.md`, repointing it from the closed #4959 at #5149 and widening it to the full per-target divergence. **#5150 has since landed and made that change unnecessary** — it fixed the boolean, integral, float, double and decimal targets and rescoped the same bullet to the remaining `timestamp` / `timestamp_ntz` / `to_time` / `try_to_time` targets in the same commit. Main's wording is now the more accurate of the two (this PR's "seven of the eight targets diverge" is no longer true), so the docs hunk is dropped and `compatibility/index.md` is untouched here.

The `#5070` removal is new since the PR was opened; #5070 was still open and its note still accurate when the first version was written.

### Deliberately unchanged

Everything else from #5085 is still accurate and stays as-is: #5073, #4680, #4947, #5071, #5072, #4967, #5068, #5022, #4993, and the datetime-rebase note in `scans.md` — I re-checked the dead-code claim there, and `spark.comet.exceptionOnDatetimeRebase` is still only its `CometConf.scala` definition with no reader, so the note holds.

After this change, the only issue referenced by any `getCompatibleNotes` override anywhere in the tree is #5073, which is still open.

A matching cleanup is needed on `branch-1.0`, which carries the #5065 and #5066 fixes (#5145, #5146) and their stale notes. That will be a separate backport.

Documentation only — no behavior change, so no new tests.

Verified locally, after rebasing onto `ba21f02de`:

- `./mvnw package -Pspark-3.5 -DskipTests -pl common,spark -am` compiles cleanly; scalastyle reports 0 errors and `spotless:check` passes.
- Confirmed by grep that no `getCompatibleNotes` override anywhere in `spark/src/main` or `common/src/main` still references #5065, #5066, or #5070, and that every issue still referenced from one (#5073) is open.
- Confirmed `compatibility/index.md` is byte-identical to main.


mbutrovich (approved):
Makes sense, thanks @andygrove!

PR #5155, opened 2026-07-30
fix: reduce log verbosity at per-task logging callsites

mbutrovich:

Part of #4082.

## Rationale for this change

Ahead of the 1.0 release, Comet's logging should be consistent with Spark's own conventions: informational messages that fire once per query or plan node are fine at INFO, but messages that fire once per task should not be, since they scale with partition count and can produce an unreasonable number of log lines for large jobs. Spark's `FileFormatWriter.executeTask` has no equivalent per-task success log, only a `logError` on abort, so Comet's per-task INFO logs are out of step with that convention.

- `CometExecIterator.getMemoryConfig` logged the computed memory pool configuration at INFO on every task (it runs once per `CometExecIterator` construction, i.e. once per native-exec task per stage). Changed to `logDebug`.
- `CometNativeWriteExec` logged `"Task ... committed successfully"` at INFO for every successful write task. Changed to `logDebug` to match Spark's `FileFormatWriter`, which stays silent on task-commit success.

No behavior change, only log levels. Existing test suite covers the surrounding code paths.


andygrove (approved):
Thanks @mbutrovich

PR #5159, opened 2026-07-31
fix: make collect_list/collect_set argument coercion a normalization barrier

### LINKED ISSUE #5158: collect_list/collect_set can fail with "column types must match schema types" on nested-field nullability drift
### Describe the bug

`collect_list` / `collect_set` over a nested argument can fail at run time with:

```
Invalid argument error: column types must match schema types,
expected List(Struct("flag": Boolean, "items": List(Struct("n": Int32))))
but found  List(Struct("flag": non-null Boolean, "items": List(Struct("n": non-null Int32))))
at column index 1
```

The two types are identical apart from the **nullability of nested fields**. Arrow treats that as fatal: `RecordBatch::try_new` compares with `DataType::equals_datatype`, which does compare nested field nullability.

There are two independent sources of truth for the element type, and nothing forces them to agree:

* **Declared** — `SparkCollectSet::state_fields` / `return_type` copy the aggregate argument's *declared* type verbatim (`datafusion-spark`, `function/aggregate/collect.rs`).
* **Produced** — `DistinctArrayAggAccumulator` stores `ScalarValue::try_from_array(...)` slices of the *actual* argument array, and `ScalarValue::new_list` honors its `data_type` parameter **only when the value set is empty** (`datafusion/common/src/scalar/mod.rs`); otherwise it just concatenates the stored scalars. So the emitted list's element type is whatever the runtime array carried.

If the declared argument type and the runtime array type disagree in nested nullability, `GroupedHashAggregateStream::emit` fails when it validates its output batch. This surfaces on both the normal emit path (validated against `AggregateExec`'s schema, derived from `return_type`) and the spill path (validated against `spill_state.spill_schema`, built from `state_fields()`).

### Why the existing guard does not cover this

`PhysicalPlanner::coerce_collect_child_nullability` (added in #4720) casts the argument to the all-nullable variant of its type, but returns early when the declared type is *already* all-nullable:

```rust
let child_type = child.data_type(schema.as_ref())?;
let nullable_type = make_all_fields_nullable(&child_type);
if child_type.equals_datatype(&nullable_type) {
    Ok(child)          // no cast inserted
} else {
    Ok(Arc::new(CastExpr::new(child, nullable_type, None)))
}
```

That is exactly the failing case: the declared type is all-nullable, the runtime array is not, no cast is inserted, and the drift reaches the accumulator. The cast only helps in the opposite direction.

Separately, the rationale comment on that function is stale for DataFusion 54.1.0. It says the accumulator "build[s] their result list with all element fields marked nullable, regardless of the input's nullability" — that is no longer true; the accumulator preserves the input array's type, so declared and produced already agree when there is no drift. The coercion is currently a normalization, not a defence.

### To Reproduce

Nullability drift is hard to produce from Comet's own operators today, because `ProjectionExec` and `ScanExec` both validate/cast their output batches against their declared schemas. The failure is reachable when an argument expression's `evaluate()` output type disagrees with its `data_type(schema)` — for example `CreateNamedStruct` derives its fields from `batch.schema()` in `evaluate()` but from the plan-time schema in `data_type()` (`native/spark-expr/src/struct_funcs/create_named_struct.rs`), so the two diverge for any batch whose schema object differs from the plan-time schema.

A unit test that builds `collect_set` against a schema with nullable leaves and feeds it a batch whose array has non-nullable leaves reproduces the error message above exactly.

### Expected behavior

The array handed to the `collect_list` / `collect_set` accumulator should have exactly the type the plan declared for it, so the aggregate cannot fail validating its own output.

### Additional context

Two fixes, not mutually exclusive:

1. Make the coercion an unconditional normalization barrier for nested argument types, so the cast forces the runtime array to the declared type regardless of what the argument produced. This is metadata-only — `cast_with_options` shallow-clones when the source and target types already match, and otherwise `cast_struct_to_struct` / `cast_list_values` stamp the target fields and clone child buffers — so it does not copy data. The declared output type is unchanged in every case, so it is not observable from Spark.
2. Harden `CreateNamedStruct` so it cannot drift: resolve its `Fields` once from the plan-time schema at construction instead of recomputing them from `batch.schema()` on every `evaluate()`.

I have a PR ready for (1); (2) is a behaviour-changing signature change and is better as a follow-up.

The underlying defect is upstream: `ScalarValue::new_list` accepts a `data_type` it then discards for non-empty input, so no accumulator built on it can guarantee the type its `state_fields()` advertises. Filing that separately against DataFusion.

andygrove:

Closes #5158.

## Rationale for this change

`SparkCollectList` / `SparkCollectSet` derive their `return_type` and `state_fields` from the aggregate argument's *declared* type, but the accumulator builds its list from the arrays it actually received — `ScalarValue::new_list` honors its `data_type` parameter only when the value set is empty, and otherwise just concatenates the collected scalars. Nothing forces the two to agree, so an argument expression whose `evaluate()` output type differs from its `data_type()` in nested field nullability makes the grouped `AggregateExec` fail validating its own output batch:

```
Invalid argument error: column types must match schema types,
expected List(Struct("flag": Boolean, "items": List(Struct("n": Int32))))
but found  List(Struct("flag": non-null Boolean, "items": List(Struct("n": non-null Int32))))
at column index 1
```

`coerce_collect_child_nullability` (added in #4720) already casts the argument to the all-nullable variant of its type, but returns early when the declared type is *already* all-nullable — which is exactly the case above, where the declared type is all-nullable and the runtime array is not. The cast only helped in the opposite direction.

Cast unconditionally for nested argument types, so the accumulator is guaranteed to see arrays of exactly the type the plan declared, whichever way the drift goes.

Two things worth noting for review:

* **Not observable from Spark.** The declared output type is unchanged in every case. Where the argument type was already all-nullable, `make_all_fields_nullable` is the identity, so the aggregate's declared type is byte-identical to before and only the runtime array is normalized. Where it was not, the cast was already being inserted.
* **Free when there is no drift.** `cast_with_options` returns a shallow clone when the source and target types are equal, so the added cast costs nothing in the common case. When they do differ it is still metadata-only: `cast_struct_to_struct` and `cast_list_values` stamp the target fields and clone the child buffers rather than copying data.

The rationale comment is also refreshed. It described the accumulator as marking all element fields nullable regardless of the input; that is no longer true in DataFusion 54.1.0 — the accumulator preserves the input array's type, so declared and produced already agree when there is no drift, and the coercion was a normalization rather than a defence.

Three new unit tests in `planner.rs`:

* `test_collect_agg_absorbs_nested_nullability_drift` — builds `collect_set` and `collect_list` against a schema with nullable nested leaves and feeds them a batch whose array has non-nullable leaves, then compares `state_fields()[0].data_type()` against the type of the emitted state array. That is exactly the pair `RecordBatch::try_new` compares inside `GroupedHashAggregateStream::emit`. Asserts the drift is present without the coercion and absent with it. Without the fix this test fails, and reproduces the error message above verbatim.
* `test_collect_agg_coercion_widens_all_nested_fields` — the declared element type is the all-nullable variant of the argument's declared type.
* `test_collect_agg_no_coercion_for_primitive_child` — primitives carry no field-level nullability, so no cast is inserted.

Existing coverage: `cargo test -p datafusion-comet --lib` (135 passed) and `CometAggregateSuite` under `-Pspark-3.5` (86 passed, 0 failed, 2 canceled as Spark-4.1-gated), including the existing `collect_list over struct with non-nullable fields` test which covers the original direction.

## Follow-ups (not in this PR)

* The root cause on the Comet side is that `CreateNamedStruct` recomputes its `Fields` from `batch.schema()` in `evaluate()` but from the plan-time schema in `data_type()`. Resolving the `Fields` once at construction would make the expression incapable of drifting, and drop per-batch field construction. That is a signature change with wider blast radius, so it is better separate.
* The underlying defect is upstream: `ScalarValue::new_list` accepts a `data_type` it then discards for non-empty input, so no accumulator built on it can guarantee the type its `state_fields()` advertises. Filed against DataFusion; if fixed there, this coercion can be dropped entirely.


sunchao (approved):
Reviewed the normalization barrier against DataFusion CastExpr evaluation and the nested-field nullability rules. I found no P1/P2 issues. The three focused normalization tests passed in an isolated harness using the exact-head production helpers and pinned dependencies.

The remaining CI failure is an HTTP 429 while downloading Maven 3.9.6, before the build runs. The other exact-head checks passed or were skipped. I did not run the full test matrix locally.

sunchao:
Linking the upstream fix mentioned in the follow-ups: [apache/datafusion#24029](https://github.com/apache/datafusion/pull/24029), merged on August 12, fixes [apache/datafusion#24022](https://github.com/apache/datafusion/issues/24022). It makes the list constructors honor the requested `data_type` for non-empty input, addressing the nested-nullability mismatch described here.

The current head (`dabc891aa426fa4e8767b547ee8de723c9363055`) still pins DataFusion `54.1.0`, which does not include that fix. Once Comet upgrades to a version containing it, we should rerun `test_collect_agg_absorbs_nested_nullability_drift` and reassess whether this additional normalization is still needed.

PR #5161, opened 2026-07-31
feat: support ANSI interval types in native Parquet scans

### LINKED ISSUE #5060: Support reading ANSI interval columns (YearMonthIntervalType / DayTimeIntervalType) in the native Parquet scan

Comet's native Parquet scan does not support the ANSI interval types, so any table with an `INTERVAL YEAR TO MONTH` or `INTERVAL DAY TO SECOND` column falls the **entire scan** back to Spark, including all the other columns in that table.

```sql
CREATE TABLE t(ym INTERVAL YEAR TO MONTH, dt INTERVAL DAY TO SECOND) USING parquet;
INSERT INTO t VALUES (make_ym_interval(1, 2), make_dt_interval(1, 2, 3, 4.5));
SELECT ym, dt FROM t;
```

```
Scan parquet spark_catalog.default.t [COMET: Unsupported schema
  StructType(StructField(ym,YearMonthIntervalType(0,1),true),StructField(dt,DayTimeIntervalType(0,3),true)):
  Unsupported ym of type YearMonthIntervalType(0,1)]
```

The gate is `DataTypeSupport.isTypeSupported`, which `CometScanTypeChecker` delegates to:

https://github.com/apache/datafusion-comet/blob/1d3044f35bda457a20f1ccf4324a4699b35f5158/spark/src/main/scala/org/apache/comet/DataTypeSupport.scala#L51-L56

Note that the list includes `CalendarIntervalType` but not the two ANSI interval types, which is backwards for a scan gate. Spark's `ParquetTable.supportsDataType` accepts only `AtomicType` and friends; `CalendarIntervalType extends DataType` directly, so it can never appear in a Parquet file, while `YearMonthIntervalType` / `DayTimeIntervalType` extend `AnsiIntervalType extends AtomicType` and are written as annotated INT32 / INT64. So the one type listed is unreachable for scans, and the two that are reachable are rejected.

This is now the main thing blocking interval work from reaching real queries. After #4898 and #4976 the types round-trip through codegen dispatch, `CometSink`, `CometLocalTableScanExec` and native shuffle, and I verified that interval columns already work natively for joins, multi-key sorts, `first`/`last`, `coalesce`/`if` and comparisons. But none of that is reachable from a Parquet-backed table today, so the only way to get an interval column into a Comet operator is to construct it inline with `make_dt_interval` / `make_ym_interval`, which is exactly what all the existing tests do.

- Add `YearMonthIntervalType` and `DayTimeIntervalType` to `DataTypeSupport.isTypeSupported`, and confirm the native Parquet reader decodes Spark's physical layout: `YearMonthIntervalType` is INT32 total months and `DayTimeIntervalType` is INT64 microseconds, matching the Arrow mappings already chosen in `native/core/src/execution/serde.rs` (`Interval(YearMonth)` and `Duration(Microsecond)`).
- Verify both scan implementations (`CometScanExec` / V1 and `CometBatchScanExec` / V2) and both `spark.comet.scan.impl` modes.
- Decide what to do with the `CalendarIntervalType` entry. It appears to be unreachable for scans; if it is only needed for `CometLocalTableScanExec` (which extends `DataTypeSupport`), it would be clearer to override there rather than widen the shared scan-facing gate.
- Add scan tests that write ANSI interval columns to Parquet and read them back, including nulls, negative values, the full `startField`/`endField` range (`INTERVAL YEAR`, `INTERVAL MONTH`, `INTERVAL DAY TO HOUR`, ...), and dictionary-encoded and multi-row-group files.

### Additional context

- Part of #4540. Follows #4898 (CalendarIntervalType support) and #4976 (nested values and native shuffle).
- Reproduced on `1d3044f35` with the Spark 4.1 profile.
- Related gates that still exclude the ANSI interval types and will surface as fallbacks once scans work, worth tracking separately:
  - `QueryPlanSerde.supportedScalarSortElementType` already carries a `TODO: Include SparkSQL's YearMonthIntervalType and DayTimeIntervalType`. Single-column sort on an interval falls back, while multi-column sort is not checked at all and produces correct results, so the restriction looks removable.
  - `AggSerde.minMaxDataTypeSupported` / `avgDataTypeSupported` / `sumDataTypeSupported` reject intervals, so `min` / `max` / `sum` / `avg` fall back even though Spark supports all four over ANSI intervals.
  - `CometShuffleExchangeExec.supportedHashPartitioningDataType` and the columnar-shuffle `supportedSerializableDataType` reject all three interval types, so `GROUP BY <interval>` falls back unless shuffle mode is `native`.
  - `arithmetic.scala`'s `supportedDataType` rejects intervals, so `UnaryMinus` and `Abs` fall back, even though `native/spark-expr/src/math_funcs/negative.rs` already has overflow-checked interval negation that is currently unreachable. That one is #4756.

peterxcli:

Closes #5060.

## Rationale for this change

Comet already maps Spark's ANSI interval types to Arrow's native representations, but the native file-scan type checker rejected both types before planning. This caused a Parquet table containing an interval column to fall back to Spark for the entire scan.

The types are admitted in `CometScanTypeChecker` instead of the shared `DataTypeSupport` predicate because the latter also gates the Arrow CSV reader, which cannot read interval or duration columns. `CometScanTypeChecker` is shared by native Parquet and Iceberg scan planning, but Iceberg has no interval type, so the practical behavior change is Parquet-only.

Year-month partition values also need literal-local serialization: Spark represents them as 32-bit month counts, while day-time intervals use 64-bit microseconds.

- Allow `YearMonthIntervalType` and `DayTimeIntervalType` in native file-scan schemas.
- Serialize null and non-null year-month literals through Comet and translate them to DataFusion `IntervalYearMonth` scalars, enabling interval partition columns and year-month predicates.
- Extend the Parquet regression test to cover all 13 ANSI interval ranges, null and negative values, dictionary encoding, multiple row groups, nested struct/array/map leaves, minimum and maximum values, filter pushdown enabled and disabled, Spark-compatible results, and native Comet execution.
- Extend the SQL-file regression test with positive, negative, and null year-month/day-time partition values.
- Update the data-type guide to document native Parquet support for nested and partition ANSI interval columns.

- `make core`
- `./mvnw test -Dtest=none '-Dsuites=org.apache.comet.parquet.ParquetReadV1Suite ANSI interval types' -Dscalastyle.skip=true`
- The same focused suite with `-Pspark-3.4`
- The same focused suite with `-Pspark-4.2`
- `./mvnw test -Dtest=none '-Dsuites=org.apache.comet.CometSqlFileTestSuite ansi_interval_scan' -Dscalastyle.skip=true`
- The same focused SQL suite with `-Pspark-3.4`
- The same focused SQL suite with `-Pspark-4.2`
- `cargo fmt --check`
- `./mvnw scalastyle:check spotless:check -DskipTests`
- `prettier --check docs/source/user-guide/latest/datatypes.md`
- `git diff --check`


spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -855,6 +855,8 @@ case class CometScanTypeChecker() extends DataTypeSupport with CometTypeShim {
       name: String,
       fallbackReasons: ListBuffer[String]): Boolean = {
     dt match {
+      case _: YearMonthIntervalType | _: DayTimeIntervalType =>
```
andygrove, line 915: `isSchemaSupported` is called for `r.partitionSchema` as well as `requiredSchema` (lines 831 and 839), and Spark does allow ANSI intervals as partition columns. `PartitioningUtils.canPartitionOn` takes any `AtomicType`, and `castPartValueToDesiredType` has an explicit `case it: AnsiIntervalType` branch. Once the checker admits the type, `partition2Proto` tries `exprToProto(Literal(value, YearMonthIntervalType))`, and `CometLiteral.getSupportLevel` only whitelists `DayTimeIntervalType` today (`literals.scala:60`), because `QueryPlanSerde.supportedDataType` lists neither interval type. So the literal comes back as `None` and the `assert` at `serde/operator/package.scala:64` fires.

I reproduced this on your branch with Spark 4.1:

```
java.lang.AssertionError: assertion failed: Unsupported partition value: 38, type: YearMonthIntervalType(0,1)
  at org.apache.comet.serde.operator.package$.partition2Proto(package.scala:52)
  at org.apache.spark.sql.comet.CometNativeScanExec.$anonfun$serializedPartitionData$10(...)
```

Reverting just this hunk goes back to a clean fallback with `Unsupported p of type YearMonthIntervalType(0,1)`. Day-time interval partition columns are fine, because the literal path already handles them.

I think this is also what the Spark 4.1 `sql_core-1` job is hitting. `PartitionedWriteSuite` has `"SPARK-37231, SPARK-37240: Dynamic writes/reads of ANSI interval partitions"`, which writes a Parquet dir partitioned by `YearMonthIntervalType(YEAR)` and `YearMonthIntervalType(YEAR, MONTH)` and reads it back. The suite is untagged so it runs in that shard, and it is not excluded in `dev/diffs/4.1.2.diff`.

Could you either teach `CometLiteral` and the native `IntVal` branch about `YearMonthIntervalType`, or keep the partition-schema gate narrow so interval partition columns still fall back? Either way a test for both interval types as partition columns would be good, since nothing covers that today.

Separately, the `assert` in `partition2Proto` turning a planner gap into an `AssertionError` rather than a fallback looks fragile. That seems worth a follow-up issue of its own.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -855,6 +855,8 @@ case class CometScanTypeChecker() extends DataTypeSupport with CometTypeShim {
...
+        true
```
andygrove, line 916: Did you look at what happens when the Parquet physical type does not match the requested interval type? Spark's `ParquetVectorUpdaterFactory` accepts any INT32 as `YearMonthIntervalType` via `IntegerUpdater` and any INT64 as `DayTimeIntervalType` via `LongUpdater`, and raises `PARQUET_COLUMN_DATA_TYPE_MISMATCH` for anything else. Comet's `parquet_convert_array` goes through arrow `can_cast_types` and then the `_ => Ok(array)` fallthrough, which gives different answers in two cases. I probed these on your branch:

| file column | requested schema | Spark | Comet |
| --- | --- | --- | --- |
| `date` | `interval year to month` | succeeds (`P1522Y1M`, ...) | `Invalid argument error: column types must match schema types, expected Interval(YearMonth) but found Date32` |
| `int` | `interval day to second` | `PARQUET_COLUMN_DATA_TYPE_MISMATCH` | succeeds, values reinterpreted as micros |
| `timestamp` | `interval day to second` | error | error |
| `long` | `interval year to month` | error | error |

The `int` row is the one I would want resolved, since Comet returns rows where Spark raises. arrow's `(_, Duration(_)) if from_type.is_numeric()` happily accepts that cast. The `date` row is the mirror image, a hard native failure where Spark succeeds.

Both need schema-on-read to trigger, so I do not think they have to be fixed in this PR. Could you file a tracking issue and link it here? Left unfiled I suspect it will not get picked up.

spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala
```diff
@@ -125,6 +127,64 @@ abstract class ParquetReadSuite extends CometTestBase {
     }
   }
 
+  test("ANSI interval types") {
```
andygrove, line 130: This is a thorough test. I like that you assert the row group count and dictionary encoding rather than hoping the write options took effect.

A few cases that are newly reachable through this change and might be worth covering, since `DataTypeSupport.isTypeSupported` recurses into complex types via your override:

- interval leaves nested in a struct, an array and a map
- a filter on an interval column, with and without `spark.sql.parquet.filterPushdown`
- the interval min and max (`INTERVAL '178956970-7' YEAR TO MONTH` and the day-time equivalents)
- both interval types as partition columns

I checked all of these on your branch and everything except the year-month partition column produces Spark-matching results, so mostly this is about locking the behavior in rather than chasing a bug.

One thing worth knowing: a `ym` predicate is currently dropped from the native data filters because the year-month literal is unsupported, so that column gets no row group pruning. Results stay correct but the filter is not accelerated. Same root cause as the partition value problem.

docs/source/user-guide/latest/datatypes.md
```diff
@@ -82,15 +82,15 @@ the tables below and may be reconsidered based on demand:
 
 All three interval types are mapped to Arrow and flow through serde, native shuffle, and the
 codegen dispatcher, so interval columns and the interval-producing expressions run natively.
-Several operators still gate on the type and fall back: Parquet scans of ANSI interval columns,
-single-column sorts, hash aggregates (`min` / `max` / `sum` / `avg`), `GROUP BY`, window
-functions, and hashing a `CalendarInterval`. Remaining work is tracked by
+Comet's V1 native Parquet scan also supports both ANSI interval types. Several operators still
```
andygrove: Comet has a single native Parquet scan path now, so "V1 native Parquet scan" may read as if there is a V2 variant with different support. Could this just say "Comet's native Parquet scan"?

Two more things for this section. The notes should call out whatever the partition column outcome ends up being. And #5060 asked whether the `CalendarIntervalType` entry in the shared `DataTypeSupport.isTypeSupported` should move to `CometLocalTableScanExec` or go away, given it can never appear in a Parquet file. Did you decide to leave that for a follow-up? A linked issue would help if so.

Unrelated to the docs, but worth adding to the PR description: `CometScanTypeChecker` also gates the native Iceberg scan (`CometScanRule.scala:366` and `:525`), not just Parquet V1. That is moot in practice since Iceberg has no interval type, but the current wording reads as if the change is Parquet-only and a future reader may not realize the checker is shared.

andygrove (commented):
Thanks for this. The mechanism looks sound to me. Spark writes year-month intervals as bare INT32 and day-time as bare INT64 (`ParquetSchemaConverter.scala:685,688`), Comet already maps those to `Interval(YearMonth)` and `Duration(Microsecond)` in `native/core/src/execution/serde.rs`, and arrow-rs has `(Int32, Interval(YearMonth))` as a reinterpret cast plus `(_, Duration(_))` for numerics. So `parquet_convert_array` resolves both with no new native code. I checked the interval min and max values on your branch and they read back correctly.

I built the branch locally and ran some probes, and I think the Spark 4.1 `sql_core-1` failure is real. That job passed on each of the last four commits on `main`. My network cannot reach the GitHub log blob host so I could not read the log, but I believe I found the cause. Details are in the inline comment on `CometScanRule.scala`.

On the other CI failure, Spark 3.5 `sql_core-1` died in "Initialize containers" after about a minute, so that one is infrastructure rather than your change. It would probably have hit the same test though.

Everything else is green, including all four scan shards on five Spark versions, Iceberg, TPC-H and TPC-DS.

spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala
peterxcli: addded:
```sql
"named_struct('ym', INTERVAL '1-2' YEAR TO MONTH, " +
"'dt', INTERVAL '1 02:03:04.5' DAY TO SECOND) AS interval_struct",
"array(INTERVAL '178956970-7' YEAR TO MONTH, " +
"INTERVAL '-178956970-8' YEAR TO MONTH) AS interval_array",
"map('max', INTERVAL '106751991 04:00:54.775807' DAY TO SECOND, " +
"'min', INTERVAL '-106751991 04:00:54.775808' DAY TO SECOND) AS interval_map")
```

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
peterxcli: thanks for the review, filed https://github.com/apache/datafusion-comet/issues/5189

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
peterxcli: filed: https://github.com/apache/datafusion-comet/issues/5188

andygrove:

Nice, targeted change. The reasoning for putting the exception in `CometScanTypeChecker` rather than the shared `DataTypeSupport` predicate is well argued, and the extra `IntVal` arm for `Interval(YearMonth)` plus the improved error message are the right supporting pieces. Extending the Parquet test to all 13 ANSI interval ranges is thorough.

A few questions.

**The new case short-circuits every other check**

```scala
dt match {
  case _: YearMonthIntervalType | _: DayTimeIntervalType => true
  case ShortType if ... =>
  case dt if isStringCollationType(dt) =>
  ...
}
```

Putting the interval arm first means these two types skip every subsequent guard in `isTypeSupported`. That is almost certainly fine today, since an interval is not a `ShortType`, not a collated string, and not complex. But it reads as "intervals are exempt from all checks" rather than "intervals are supported", and if a future check needs to apply to all leaf types it will silently not apply to these. Would it be better to place the arm at the end, just before the default, so it reads as the fallthrough it actually is?

**Only V1 is tested**

`ansi_interval_scan.sql` sets `spark.sql.sources.useV1SourceList=parquet`, so the fixture covers `FileSourceScanExec` and not `CometBatchScanExec`. `CometScanTypeChecker` is used on the V2 path too. Was V2 tested? If the answer is "it works but the fixture pins V1 for determinism", saying so in a comment would help. If V2 is untested, it is worth adding, since the partition-value serialization path in particular differs between the two.

**Is the year-month literal path exercised beyond partition values?**

The description says year-month literal serialization is needed for partition columns and year-month predicates. The SQL fixture covers partition columns. Does anything cover a predicate, for example `WHERE ym > INTERVAL '1-0' YEAR TO MONTH`, running natively? That is the case where a wrong literal encoding gives a wrong answer rather than an error, so it is worth pinning explicitly.

**Docs formatting**

The `datatypes.md` table rows changed length but the separator row did not. Prettier will want to realign that table, and prettier drift fails CI. Worth running `make format` plus prettier before pushing so this does not come back as a CI failure unrelated to the change.

peterxcli:
Thanks for the pass — point by point:

**Match arm ordering** — agreed it read as "exempt from everything" rather than "supported." Moved the arm to just above the `super.isTypeSupported` fallthrough (it has to stay above the super call, since the shared default rejects intervals) with a comment. No behavior change: an interval can never match the `ShortType`, collation, or struct guards.

**V2 coverage** — the V2 Parquet path doesn't go through `CometScanTypeChecker`: `transformV2Scan` gates on `CometBatchScanExec.isSchemaSupported`, which uses the shared `DataTypeSupport` default and still rejects intervals, so a V2 interval scan falls back to Spark exactly as before this PR. (`CometScanTypeChecker` is shared with the V2 Iceberg path, but Iceberg has no interval type.) The fixture's `Config:` line only pins Spark's default — parquet is already in the default `useV1SourceList` — and I've added a comment in the fixture saying so.

**Native predicate coverage** — `ParquetReadSuite`'s "ANSI interval types" test already filters on `i2 = INTERVAL '1-2' YEAR TO MONTH AND i9 = INTERVAL '1 02:03:04.5' DAY TO SECOND` under `checkSparkAnswerAndOperator`, asserts `dataFilters` is non-empty, and checks `output_rows` == 333 with pushdown vs 1000 without, so both the literal encoding and row-group pruning are pinned against Spark's answer. I've additionally added `ym >` / `dt >` predicate queries to `ansi_interval_scan.sql` so the SQL golden harness covers it across Spark versions too.

**Prettier** — `npx prettier --check docs/source/user-guide/latest/datatypes.md` passes on the branch as-is, and the markdown CI check is green on the latest push, so no realignment is pending.

Separately, the one failing check (`PR Build (Linux) / Spark 4.1, JDK 17 [exec]`) was an HTTP 429 fetching the Maven distribution during setup — infra flake; a re-run should clear it.

Both targeted suites pass locally on Spark 4.1 / JDK 17 after the change: `ParquetReadV1Suite "ANSI interval types"` and `CometSqlFileTestSuite "sql-file: expressions/datetime/ansi_interval_scan.sql"`.

PR #5162, opened 2026-07-31
fix: skip null slots when checking overflow in unary negation

### LINKED ISSUE #5071: ANSI arithmetic overflow errors: wrong error class for Byte/Short, wrong type names, missing try_ suggestions
### Describe the bug

For several arithmetic expressions, Comet's ANSI throw/NULL decision is correct, but the error class, error message, or exception parameters differ from Spark. All of these surface to users who catch or match on Spark error conditions.

1. **Byte/Short overflow raises the wrong error class.** `Add`/`Subtract`/`Multiply` on TINYINT/SMALLINT under ANSI: Spark 4.1 raises `BINARY_ARITHMETIC_OVERFLOW` carrying the two operand values and the operator symbol (`types/numerics.scala`, codegen in `arithmetic.scala`). Comet raises `ARITHMETIC_OVERFLOW` with `from_type: "integer"` (`native/spark-expr/src/math_funcs/checked_arithmetic.rs:69-71, 93-96`). `SparkError::BinaryArithmeticOverflow` already exists (`native/common/src/error.rs:87-93`) and is wired through `ShimSparkErrorConverter`, but no kernel constructs it.
2. **Long overflow says "integer overflow".** `from_type` is hard-coded to `"integer"` regardless of Int32 vs Int64 (`checked_arithmetic.rs:70, 94`). Spark says `"long overflow"` for BIGINT. The same applies to `SumInteger` (`native/spark-expr/src/agg_funcs/sum_int.rs`), where Spark's sum result type is always LONG.
3. **The `try_` suggestion is always empty.** `ShimSparkErrorConverter.scala:118` passes `""` for the alternative, so Comet's message omits ``Use `try_add` to tolerate overflow and return NULL instead.``
4. **`UnaryMinus` scalar path for Byte/Short** passes `" caused"` as the type (`native/spark-expr/src/math_funcs/negative.rs:157, 160`), producing `"[ARITHMETIC_OVERFLOW]  caused overflow."` instead of `"byte overflow"` / `"short overflow"`. The array path uses the correct names.
5. **`Abs`** passes Rust type names (`"Int8"`, `"Int64"`, `"Decimal128"`, ...) as `from_type` (`native/spark-expr/src/math_funcs/abs.rs:98, 106, 114, 122, 156, 186`), so Comet says `"Int64 overflow"` where Spark says `"long overflow"`. Contrast with `negative.rs`, which uses the Spark names.
6. **`avg` overflow suggestion**: `avg_decimal.rs` uses `decimal_sum_overflow_error(...)`; verify the function-name parameter produces `try_avg` (Spark's message) rather than `try_sum` (`ShimSparkErrorConverter.scala:123-127` builds ``try_$functionName``).

### Steps to reproduce

```sql
SELECT CAST(127 AS TINYINT) + CAST(1 AS TINYINT);      -- Spark: BINARY_ARITHMETIC_OVERFLOW; Comet: ARITHMETIC_OVERFLOW "integer overflow"
SELECT CAST(9223372036854775807 AS LONG) + 1L;          -- Spark: "long overflow"; Comet: "integer overflow"
SELECT -CAST(-128 AS TINYINT);                          -- literal/scalar path: Comet message is "[ARITHMETIC_OVERFLOW]  caused overflow."
SELECT abs(CAST(-9223372036854775808 AS LONG));         -- Spark: "long overflow"; Comet: "Int64 overflow"
```

### Expected behavior

Error classes, type names, and `try_` suggestions match Spark exactly. Existing tests pass because they assert message substrings only; assertions on `getCondition()` / error class would catch these.

### Additional context

Found by an ANSI-mode audit of all native expressions against Spark 4.1.1. Possibly related to the Spark 4.2 ANSI test failures in #4967.

### LINKED ISSUE #5093: negative: ANSI overflow check reads null slots and can raise spurious overflow errors
### Describe the bug

In `native/spark-expr/src/math_funcs/negative.rs`, the ANSI-mode overflow check (`check_overflow!` macro) loops over Int8/16/32/64 and interval arrays comparing every slot's value to `MIN`, then calls `neg_wrapping`. The scan reads `value(i)` without checking validity, so a `MIN` value sitting in a null slot (arbitrary bytes left behind by filter/slice/FFI) raises a spurious ARITHMETIC_OVERFLOW error even though the row is null.

It is also a two-pass re-implementation of arrow's checked `neg` kernel, which negates with overflow detection in a single pass and correctly skips null slots.

### Steps to reproduce

Construct an Int64 array where a null slot's underlying value is `i64::MIN` (for example by masking a valid `i64::MIN` entry to null via a validity buffer) and evaluate `negative()` on it in ANSI mode. The expression errors; Spark returns null for the null row and succeeds.

### Expected behavior

Null slots must be ignored by the overflow check. Replacing the ANSI path with `arrow::compute::kernels::numeric::neg` fixes this, since the kernel skips invalid slots.

### Additional context

Notes for the fix:
- The Spark error message per type must be reconstructed in a `map_err`; if the exact offending value is required for Int8/Int16 wording, a rescan on the (cold) error path can recover it.
- `IntervalUnit::MonthDayNano` is deliberately not overflow-checked today, and arrow's checked `neg` would check it, so that type must keep `neg_wrapping` unconditionally.
- Legacy (non-ANSI) mode keeps `neg_wrapping` as today.

Found during an audit of native code that replicates existing arrow-rs kernels. The corresponding null-slot behavior is already pinned correctly for add in `checked_arithmetic.rs` tests.

Smallfu666:
Closes #5093

> Also addresses item 4 of #5071 (the `UnaryMinus` byte/short scalar overflow message). This PR
> does **not** close #5071 — its other items, including `abs` (item 5), remain tracked there.

### Summary
Replace the validity-blind ANSI overflow pre-scan in `NegativeExpr` with Arrow's checked `neg` kernel, which skips null slots natively.

The old pre-scan walked `typed_array.value(i)` for every slot without consulting the null buffer, so a MIN sentinel left behind in a **null** slot — by `filter`, `slice`, or FFI — raised a spurious ANSI overflow. `neg` goes through `PrimitiveArray::try_unary`, which applies the closure only to valid indices.

### What changes, precisely

**Fixed: false positives on null slots.** Verified against `main`, all of these errored before and are now correctly skipped:

| input (slot 0 null, holding the sentinel) | before | after |
| --- | --- | --- |
| `Int8`/`Int16`/`Int32`/`Int64` at `MIN` | `ArithmeticOverflow` | null preserved |
| `Interval(YearMonth)` at `i32::MIN` | `ArithmeticOverflow` | null preserved |
| `Interval(DayTime)` at `IntervalDayTime::MIN` | `ArithmeticOverflow` | null preserved |

**Fixed: error mapping for single-component `DayTime` overflow.** An `Interval(DayTime)` with only one component at `i32::MIN` — `(i32::MIN, 0)` or `(0, i32::MIN)` — was not matched by the pre-scan, which compared against the whole `IntervalDayTime::MIN`. It then fell through to `neg_wrapping`, which wraps only for integers: `downcast_integer! { ..., _ => neg(array) }` means every other type, intervals included, is already routed to the checked kernel. So these values **did** fail before, with a raw `DataFusionError::ArrowError(ArithmeticOverflow(..))` instead of a Spark error.

**The `DayTime` failure set is therefore unchanged** (`days == i32::MIN || ms == i32::MIN`, before and after) — this PR only unifies the message, matching the scalar path at `negative.rs`, which already checked the components individually.

**Unchanged:** legacy (non-ANSI) mode, and the `Interval(MonthDayNano)` dispatch.

### Also in this PR (from review)

**Byte/short type names.** The scalar `Int8`/`Int16` arms passed a bare `" caused"` as `from_type`, rendering as `[ARITHMETIC_OVERFLOW]  caused overflow.` — no type, no value, double space. Both paths now pass `"byte"` / `"short"`.

No single string can match every supported Spark version here, because Spark changed both the wording and the error class:

| Spark | throws | message |
| --- | --- | --- |
| 3.4, 3.5 | `QueryExecutionErrors.unaryMinusCauseOverflowError` | `[_LEGACY_ERROR_TEMP_2043] - -128 caused overflow.` |
| 4.0+ | `MathUtils.negateExact` → `arithmeticOverflowError("byte overflow")` | `[ARITHMETIC_OVERFLOW] byte overflow. If necessary set …` |

`SparkError::ArithmeticOverflow` always renders as `ARITHMETIC_OVERFLOW` — all three shims build the message as `fromType + " overflow"` — so the 3.x error class is unreachable regardless of the string. `"byte"` / `"short"` is byte-identical to Spark 4.x and no worse than the previous values on 3.x, and it drops the need to hardcode the value.

**Test assertions now pin the documented divergence instead of skipping it.** `checkOverflow` in `CometExpressionSuite`'s `unary negative integer overflow test` takes separate Spark and Comet expected substrings:

- **Spark 3.4 / 3.5:** it asserts Spark's message contains `"caused overflow"` while Comet's contains `"byte overflow"` / `"short overflow"` — the previously vacuous `dtype = ""` placeholder (which only checked for `" overflow"`) is gone.
- **Spark 4.x:** both sides assert the type name, which now match exactly.

The dead interval sub-block in that test (its DataFrames were built and discarded, so it re-ran `select -(_1) FROM tbl` and asserted nothing) was removed.

### Tests
Native tests are the right level for the null-slot fix itself: the bug needs garbage in a null slot, which is not reachable from SQL.

- `cargo fmt --check` — clean
- `cargo clippy -p datafusion-comet-spark-expr --all-targets -- -D warnings` — clean
- `spotless:check` — clean
- `cargo test -p datafusion-comet-spark-expr` — 610 passed
- `CometExpressionSuite` — 140 tests, 0 failures (4 ignored) on **Spark 3.5** (`-Pspark-3.5 -Pscala-2.12`) and on **Spark 4.0** (`-Pspark-4.0 -Pscala-2.13`), both against a freshly built `libcomet.so` verified byte-identical to the one loaded by the JVM.


andygrove (commented):
Thanks for the fix! I verified the mechanism rather than taking it on faith, and also built the branch and ran the tests, clippy, and fmt — all clean.

## The fix is correct

- `arrow_arith::numeric::neg` dispatches Int8/16/32/64 and `Interval(YearMonth|DayTime)` through `neg_checked!` → `PrimitiveArray::try_unary`, which (arrow-array 58.4.0, `primitive_array.rs:986-989`) does `nulls.try_for_each_valid_idx(f)` when a null buffer is present. Null slots are genuinely skipped, so the spurious-overflow bug is gone.
- The `-128 caused` / `-32768 caused` hardcodes are provably faithful: checked negation of `iN` overflows only at `iN::MIN`, so the old `format!("{:?} caused", value)` could only ever produce those exact strings. This is cleaner than the cold-path rescan the issue suggested.
- No other callers of the deleted `check_overflow!` macro.
- `cargo test -p datafusion-comet-spark-expr --lib math_funcs::negative` → 8/8 pass; `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean.

Bonus: this collapses two passes (scan + wrap) into one fused checked pass in ANSI mode.

## Issues

**1. Undocumented behavior change in the `Interval(DayTime)` branch.** The description says only legacy and `MonthDayNano` are unchanged, but `DayTime` semantics changed materially:

- Before: error only when `value == IntervalDayTime::MIN`, i.e. **both** `days == i32::MIN` **and** `ms == i32::MIN`.
- After: arrow's kernel calls `neg_checked()` on each component (`numeric.rs:148-157`), so **either** component at `i32::MIN` errors.

`test_ansi_interval_day_time_component_overflow_and_null_controls` pins the new, wider behavior. This is low risk and arguably a fix — it now matches the scalar path at `negative.rs:171` (`days == i32::MIN || ms == i32::MIN`), and `supportedDataType` (`QueryPlanSerde.scala:527-542`) excludes `DayTimeIntervalType`, so `CometUnaryMinus.getSupportLevel` rejects these columns and the branch is unreachable from the planner today. But please call it out in the description, ideally with a comment noting the array/scalar alignment.

**2. Substantial duplication in the new `evaluate` arm.** Five near-identical `let result = ...; Ok(ColumnarValue::Array(result))` blocks. Collapsing to a type-name lookup halves the branch and makes the legacy/checked split obvious:

```rust
ColumnarValue::Array(array) => {
    if !self.fail_on_error {
        return Ok(ColumnarValue::Array(neg_wrapping(array.as_ref())?));
    }
    // Checked negation overflows only at MIN, so the byte/short values are fixed.
    let from_type = match array.data_type() {
        DataType::Int8 => "-128 caused",
        DataType::Int16 => "-32768 caused",
        DataType::Int32 => "integer",
        DataType::Int64 => "long",
        DataType::Interval(IntervalUnit::YearMonth | IntervalUnit::DayTime) => "interval",
        // MonthDayNano is deliberately not overflow-checked (pre-existing behavior);
        // no other type has a Spark overflow message here.
        _ => return Ok(ColumnarValue::Array(neg_wrapping(array.as_ref())?)),
    };
    Ok(ColumnarValue::Array(
        neg(array.as_ref()).map_err(|e| map_neg_error(e, from_type))?,
    ))
}
```

Also, `map_neg_error`'s `type_name: &'static str` parameter now receives value strings like `"-128 caused"` — rename it `from_type` to match the `SparkError` field.

**3. The scalar path is left inconsistent with the array path.** `negative.rs:156-161` still passes bare `" caused"` for `Int8`/`Int16`, producing `[ARITHMETIC_OVERFLOW]  caused overflow.` — double space, no value. The array path now says `-128 caused`. Since this PR hardcodes those exact strings a few lines above, fixing the scalar arms is a two-line change that closes a real message bug. Worth doing here.

**4. Stale comment.** `negative.rs:110-111` ("datatypes supported are byte, short, integer, long, float, interval") is now more misleading than before, since the restructured match makes float/decimal explicitly fall to the `_` wrapping arm.

## Test coverage

Good coverage of the actual fix — the null-slot-holds-MIN case is pinned for all four int widths plus both `IntervalDayTime` components, and `test_legacy_mode_wraps_min_values` guards the non-ANSI path. Native tests are the right level here; the bug is latent (garbage in null slots from filter/slice/FFI), so it isn't reachable from SQL and no Scala test is needed.

Nits:

- Heavy copy-paste: the Int8/16/32/64 blocks are repeated verbatim in two tests (~120 of the 280 lines). A small macro or a loop over a helper closure would cut this a lot.
- Bare `panic!()` in the `let ... else` arms gives no diagnostic. `panic!("expected array result")` costs nothing.
- `test_map_neg_error_preserves_non_overflow_errors` is testing a two-line `match` — low value relative to its size.

## Follow-up (not this PR)

`CalendarIntervalType` *is* in `supportedDataType`, and it maps to Arrow `IntervalMonthDayNano` — the one branch this PR deliberately keeps on `neg_wrapping`. Spark's `UnaryMinus` under ANSI uses `IntervalUtils.negateExact` for `CalendarIntervalType` (`arithmetic.scala:101-102`), which throws on overflow, so Comet silently wraps where Spark errors. Pre-existing and correctly out of scope, but arrow's `neg` would now check `MonthDayNano` too, so this is a cheap follow-up — worth an issue and a code comment explaining *why* the branch opts out.

## Verdict

The core fix is sound, verified, and well-targeted. Happy to approve once items 2–4 (all mechanical) are addressed and the `DayTime` change from item 1 is documented.

Smallfu666:
Thanks for the thorough review — all addressed in `f1acb5e50`:

- collapsed the duplicated checked-negation branches (and renamed `map_neg_error`'s param to `from_type`)
- fixed the scalar Int8/Int16 messages to match the array path
- updated the stale comment, cut the test copy-paste
- added the null-slot regressions that were actually missing: `IntervalYearMonth(i32::MIN)` and the whole `IntervalDayTime::MIN`

On the `DayTime` change in your item 1 — I checked this on `main` and I don't think the behavior widened. `neg_wrapping` is `downcast_integer! { ..., _ => neg(array) }`, so every non-integer type, intervals included, already went through the checked kernel. `(i32::MIN, 0)` and `(0, i32::MIN)` did fail before this PR:

```
DayTime valid (MIN,0) => ArrowError(ArithmeticOverflow("Overflow happened on: - -2147483648"))
```

The failure set is the same; what changes is that these now map to `SparkError::ArithmeticOverflow` like every other overflow here. I've reframed that test accordingly and rewritten the description.

That also means the `MonthDayNano` follow-up doesn't hold — nothing silently wraps there, since `neg_wrapping` is `neg` for that type too. So I dropped it.

The workflows are still waiting on a maintainer to approve the run.

native/spark-expr/src/math_funcs/negative.rs
```diff
@@ -107,57 +97,51 @@ impl PhysicalExpr for NegativeExpr {
...
+                if !self.fail_on_error {
+                    return Ok(ColumnarValue::Array(neg_wrapping(array.as_ref())?));
                 }
-                let result = neg_wrapping(array.as_ref())?;
-                Ok(ColumnarValue::Array(result))
+                // Checked negation of `iN` overflows only at `iN::MIN`, so the value
+                // Spark reports for byte/short is fixed and can be hardcoded here.
+                let from_type = match array.data_type() {
+                    DataType::Int8 => "-128 caused",
+                    DataType::Int16 => "-32768 caused",
```
andygrove: Nit, not blocking. Worth knowing that these two strings cannot match every supported Spark version, because Spark changed both the wording and the error class here:

- Spark 3.4 and 3.5 throw `QueryExecutionErrors.unaryMinusCauseOverflowError`, which is error class `_LEGACY_ERROR_TEMP_2043` with the template `- <sqlValue> caused overflow.`, so the message is `[_LEGACY_ERROR_TEMP_2043] - -128 caused overflow.`
- Spark 4.0 onward routes byte and short through `MathUtils.negateExact`, which throws `arithmeticOverflowError("byte overflow")`, so the message is `[ARITHMETIC_OVERFLOW] byte overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error.`

`SparkError::ArithmeticOverflow` always renders as `ARITHMETIC_OVERFLOW`, since the shims build the message as `fromType + " overflow"`, so the 3.x class is out of reach no matter what string goes here. That makes `"byte"` and `"short"` the better choice: byte-identical to Spark 4.x, and no worse than the current values on 3.x. It also drops the need to hardcode the value, and the `-128` in the comment above.

If you take it, the matching test expectations move to `"byte"` / `"short"`, and the `dtype = ""` placeholders for the byte and short cases in `CometExpressionSuite`'s `unary negative integer overflow test` could become real assertions under `isSpark40Plus`. Happy for this to land as is either way.

Smallfu666: Took it — `6ee228bb8`. Both the array and scalar paths now pass `"byte"` / `"short"`, which also drops the hardcoded value.

You're right that no single string satisfies every version. I checked the 4.x side directly: in `spark-sql-api_2.13-4.0.2.jar`, `MathUtils.negateExact(byte)` uses the literal `"byte overflow"`, so with the shims' `fromType + " overflow"` the message is byte-identical there. On 3.4/3.5 the legacy `_LEGACY_ERROR_TEMP_2043` formatting stays different — the description now says that explicitly rather than leaving it implied.

I took the second half of your suggestion too: the byte/short `dtype` placeholders in `unary negative integer overflow test` are real assertions under `isSpark40Plus` now, for both the array cases and the scalar `cast(… as byte/short)` ones. That adds `CometExpressionSuite.scala` to the PR. It's 143 tests / 0 failures on both `-Pspark-3.5 -Pscala-2.12` and `-Pspark-4.0 -Pscala-2.13`.

andygrove:
Thanks for the revision. All four items from my earlier review are addressed, and I built the branch and re-ran everything rather than taking the results on faith.

## Your correction on `DayTime` is right, and my follow-up was wrong

`neg_wrapping` in arrow-arith 58.4.0 (`numeric.rs:179-184`) is:

```rust
downcast_integer! {
    array.data_type() => (neg_wrapping, array),
    _ => neg(array),
}
```

So every non-integer type, intervals included, already went through the checked kernel. The `DayTime` failure set really is unchanged, and my `MonthDayNano` follow-up rested on a false premise. Nothing silently wraps there. Sorry for the noise, and good catch.

## What else I verified

**The byte/short reasoning holds.** All three shims build the message as `fromType + " overflow"` under `ARITHMETIC_OVERFLOW` (`spark-3.4/.../ShimSparkErrorConverter.scala:98`, `spark-3.5/...:98`, `spark-4.x/...:118`). Spark 4.x `MathUtils.negateExact` throws literally `"byte overflow"` and `"short overflow"` (`sql/api/.../MathUtils.scala:64-76`), while Spark 3.5's `UnaryMinus` uses `QueryExecutionErrors.unaryMinusCauseOverflowError` and lands on `_LEGACY_ERROR_TEMP_2043`, which Comet cannot emit. Exact on 4.x, out of reach on 3.x either way.

**Local runs, since the workflows still have not been approved:**

- `cargo test -p datafusion-comet-spark-expr` gives 574 passed, matching your number
- `cargo clippy -p datafusion-comet-spark-expr --all-targets -- -D warnings` is clean
- `cargo fmt --check` is clean
- `CometExpressionSuite` `unary negative integer overflow test` passes on `-Pspark-3.5 -Pscala-2.12` and on `-Pspark-4.0 -Pscala-2.13`

I also checked whether the `Interval(MonthDayNano)` fallthrough is reachable, since Spark's legacy `IntervalUtils.negate` wraps while Arrow's `neg` is checked for that type. It is not: `make_interval` falls back, and there is no Parquet source for `CalendarIntervalType`, so the projection stays in Spark. Not raising it.

## Three things left

**1. The interval block at `CometExpressionSuite.scala:2379-2391` does not test anything.**

The three DataFrames are built and then discarded. `Seq(longDf, yearMonthDf, dayTimeDf).foreach { _ => ... }` ignores its argument, so this runs `select -(_1) FROM tbl` three times against `tbl`, whose `_1` is `i % 5`. No overflow, so `checkOverflow` lands in the `(None, None)` branch every time.

Worth cleaning up while you are in this method, especially since it looks like the interval coverage for the very behavior this PR changes. There is not much to salvage either: `supportedDataType` (`QueryPlanSerde.scala:528-531`) omits `YearMonthIntervalType` and `DayTimeIntervalType`, so `CometUnaryMinus.getSupportLevel` returns `Unsupported` and those columns fall back today. Deleting the block seems more honest than leaving it in place. Would you mind removing it, or filing an issue if you would rather keep a marker for real interval coverage?

**2. The Spark 3.x half of the placeholder in `checkOverflow` (`CometExpressionSuite.scala:2329-2340`) is still a non-assertion.**

With `dtype = ""` it asserts that both messages contain `" overflow"`. Spark 3.5's `"- -128 caused overflow."` and Comet's `"byte overflow."` both satisfy that, and so would any other type name, so it cannot catch a regression. Given that you have pinned down exactly how 3.x diverges, could that divergence be asserted rather than skipped? Something like having `checkOverflow` take separate expected substrings, then asserting `"caused overflow"` for Spark and `"byte overflow"` for Comet on 3.x. That turns the documented incompatibility into something the suite actually guards.

**3. `abs` has the same message bug this PR just fixed for negation, and is now inconsistent with it. Could you file an issue?**

`abs.rs` passes `"Int8"`, `"Int16"`, `"Int32"`, `"Int64"`, `"Decimal128"` and `"Decimal256"` as `from_type` (lines 98, 106, 114, 122, 156, 186, plus the scalar arms at 210-302), so ANSI `abs(-128)` reports `[ARITHMETIC_OVERFLOW] Int8 overflow.` where Spark 4.x reports `byte overflow`. Spark's `Abs` routes byte, short, int and long through the same `MathUtils.negateExact` as `UnaryMinus` does (the `ByteType | ShortType | IntegerType | LongType if failOnError` arm in `arithmetic.scala`), so `"byte"`, `"short"`, `"integer"` and `"long"` is what it should be saying.

That is a different expression and file, so it is out of scope here. Now that negation uses the Spark names it would be good to have it tracked though. Could you open an issue and link it? The regex in `CometMathExpressionSuite:65` is loose enough that it passes today, so nothing is catching it.

## Verdict

The core fix is sound and verified. Items 1 and 2 are contained test changes and item 3 is just an issue to file, so this is close. The workflows also still need a maintainer to approve the run before merge.

Smallfu666:
Thanks! All three are handled in the latest revision:

1. Dropped the dead interval block — those DataFrames were built but never actually used by the query, so it wasn't covering anything.

2. Reworked `checkOverflow` to assert the Spark and Comet substrings separately:
   - Spark 3.4/3.5: Spark `"caused overflow"` vs Comet `"byte overflow"` / `"short overflow"`
   - Spark 4.x: both sides `"byte overflow"` / `"short overflow"`

   So the old empty-string assertion is gone.

3. I didn't open a separate `abs` issue — it's already tracked as item 5 of #5071, and this PR covers the `UnaryMinus` half (item 4 there).

Rebased onto the latest `main` and reran the native and Spark 3.5 / 4.0 suites — all green.

native/spark-expr/src/math_funcs/negative.rs
```diff
@@ -107,57 +97,55 @@ impl PhysicalExpr for NegativeExpr {
...
+                    DataType::Int8 => "byte",
+                    DataType::Int16 => "short",
+                    DataType::Int32 => "integer",
+                    DataType::Int64 => "long",
+                    // `neg` checks each `DayTime` component, so either `days` or `ms` at
+                    // `i32::MIN` overflows -- the same failure set as before. The pre-scan
+                    // this replaced only matched the whole `IntervalDayTime::MIN`, but the
+                    // single-component values then overflowed inside `neg_wrapping`
+                    // anyway; routing them here only maps the raw Arrow error onto Spark's,
+                    // and matches the scalar path below.
```
andygrove: Some of the comments seem to refer to the previous implementation - seems like info that is not useful once this merges?

andygrove:
Thanks for the revision. Two things left.

**1. Drop the empty-string sentinel in `checkOverflow` and fold the version branch into a helper.**

`CometExpressionSuite.scala:2327-2344`. The comment explaining why a local `def` cannot take
`= sparkExpected` is a sign the signature is fighting the language. The version-dependent Spark
string is also written three times, once in the tuple list and once in each of the byte and short
scalar loops. Both go away with a small helper:

```scala
// Spark 3.4/3.5 throw `_LEGACY_ERROR_TEMP_2043` ("- <sqlValue> caused overflow.") for byte and
// short. Spark 4.x routes them through `MathUtils.negateExact` and agrees with Comet, which
// always renders `SparkError::ArithmeticOverflow` with the Spark type name.
def sparkOverflowMsg(dtype: String): String =
  if (isSpark40Plus) s"$dtype overflow" else "caused overflow"

def checkOverflow(query: String, sparkExpected: String, cometExpected: String): Unit = ...
```

Every call site then passes both strings, the byte case reads
`checkOverflow(query, sparkOverflowMsg("byte"), "byte overflow")`, and the `if (isSpark40Plus)`
branching disappears from all three places.

**2. The float rows never exercise `checkOverflow`.**

`CometExpressionSuite.scala:2423-2425`. Float negation cannot overflow, so those two queries always
land in the `(None, None)` branch and the `"float overflow"` argument is never asserted.
`checkSparkAnswerAndOperator` still verifies the result and native execution, so nothing is wrong,
but the argument implies a check that never happens. Since you just removed the dead interval block
for the same reason, could these move to a plain `checkSparkAnswerAndOperator` call, or get a
one-line comment saying they are there to confirm float negation stays native and does not throw?

Smallfu666:
Thanks. I addressed the remaining feedback.

`checkOverflow` now takes explicit Spark and Comet substrings, with the version-specific Spark message handled in one `sparkOverflowMsg` helper.

The float cases now use `checkSparkAnswerAndOperator`, since float negation does not overflow.

I also removed the comments that described the previous implementation.

The native tests and `CometExpressionSuite` pass locally on Spark 3.5 and 4.0.

andygrove (approved):
LGTM pending CI. Thanks @Smallfu666

PR #5166, opened 2026-07-31
fix: normalize floating-point values in native collect_set

### LINKED ISSUE #4966: Spark 4.2: Comet native collect_set does not normalize NaN / -0.0 (SPARK-57298)
Spark 4.2 (SPARK-57298) makes `collect_set` normalize `NaN` and `-0.0` for floating-point types. Comet's native `collect_set` aggregate does not perform this normalization, so results differ from Spark.

Failing Spark SQL test (currently marked `IgnoreComet` in `dev/diffs/4.2.0.diff`):
- `DataFrameAggregateSuite`: "SPARK-57298: collect_set normalizes NaN and -0.0 for floating-point types"

Note: the array-expression normalization path (`array_distinct` / `array_except`) is already handled; this issue tracks the aggregate `collect_set` path.

peterxcli:

Closes #4966.

## Rationale for this change

Spark 4.2 (SPARK-57298, apache/spark@49908a2cf92) changed `collect_set` to normalize NaN values and signed zero before deduplication, inside `CollectSet.convertToBufferElement`/`eval` — so the normalization is invisible in the plan Comet receives. Comet passed the raw input to the native partial aggregate, whose set distinguishes floating-point bit patterns, so its result could differ from Spark for NaN and `-0.0` values.

For array-typed inputs, Spark's recursive normalizer produces `ArrayTransform`, which Comet executes through the JVM codegen dispatcher rather than end-to-end native execution. Scalar normalization remains native, as does the plain struct path. If `spark.comet.exec.scalaUDF.codegen.enabled=false`, array-typed cases safely fall back to Spark. #5312 tracks whether to normalize signed zero on Spark 3.4 through 4.1 while accounting for this nested-array codegen cost.

- Reuse Spark's recursive floating-point normalizer for Spark 4.2+ `collect_set` raw-row aggregation (`Partial` and `Complete`). Normalization is idempotent: `needNormalize` short-circuits on `KnownFloatingPointNormalized`, so an already-normalized child is never wrapped twice.
- Keep final aggregation state merging unchanged and preserve the strict floating-point fallback on older Spark versions.
- Document the array-input JVM codegen dispatch trade-off via `getCompatibleNotes()` so it appears in the generated Compatibility Guide.
- Document the compatibility risk of relying on Spark's `private[sql]` `NormalizeFloatingNumbers.normalize`: it exists in all supported Spark versions (3.4–4.2) and any upstream rename fails the profile build loudly since this is a direct compile-time reference.
- Add Spark-version-specific SQL coverage for scalar, array, and struct values across partial aggregation buffers, gated by `-- MinSparkVersion: 4.2` / `-- MaxSparkVersion: 4.1` headers.

- `make core`
- `./mvnw test -Pspark-4.2 -Dtest=none -Dsuites='org.apache.comet.CometSqlFileTestSuite collect_set' -Dscalastyle.skip=true`
- `./mvnw test -Pspark-4.1 -Dtest=none -Dsuites='org.apache.comet.CometSqlFileTestSuite collect_set_floating_fallback' -Dscalastyle.skip=true`


andygrove (commented):
Thanks for taking this on. Reusing Spark's own `NormalizeFloatingNumbers.normalize` rather than reimplementing the normalization in the native set is the right call, and gating on `Partial` mirrors Spark, where normalization only happens in `update()` via `convertToBufferElement`.

I checked a couple of things that could have gone wrong and they look fine. The nested paths are already reachable, since `CometKnownFloatingPointNormalized` in `contraintExpressions.scala` handles both the scalar `NormalizeNaNAndZero` case and the nested `ArrayTransform` / `CreateNamedStruct` cases. The mixed Comet-partial / JVM-final buffer hazard is already guarded by `hasNativeArrayBufferAgg` plus the `COMET_UNSAFE_PARTIAL` tagging in `CometExecRule`, so there is no new interop concern here. CI is fully green including all four Spark 4.2 shards.

**1. `aggregates.scala:855` - version-gate the incompatible reason.**

On Spark 4.2+ `getSupportLevel` now returns `Compatible()`, but `getIncompatibleReasons()` still returns a bullet unconditionally. `GenerateDocs` writes those bullets into the per-Spark-version compatibility page, so the 4.2 aggregate page would show `collect_set` carrying an incompatibility that this PR just removed. Could this mirror the support level?

```scala
override def getIncompatibleReasons(): Seq[String] =
  if (isSpark42Plus) Nil else Seq(...)
```

That would also let the wording drop the "Before Spark 4.2" qualifier, which reads a little oddly on a page that is already version-specific.

**2. `aggregates.scala:889` - should `Complete` normalize too?**

Spark normalizes inside `convertToBufferElement`, which is reached from `update()`, and that runs in both `Partial` and `Complete` mode. Comet rejects `Complete` today in `operators.scala:1737`, so this is not reachable, but I think it is worth writing `case Partial | Complete` now with a one-line comment. Spark 4.3 adds `spark.sql.execution.bypassPartialAggregation`, which plans exactly a single `Complete`-mode aggregate in `AggUtils.planAggregateWithoutDistinct`, and `CometWindowExec` already asserts `Complete` for window aggregates. If either path gains `collect_set` support later, the current gate silently returns wrong results rather than falling back.

**3. `collect_set_normalization.sql` - assert values, not just sizes.**

Most queries here check `size(collect_set(...))`. That confirms the dedup count but not which representative survived, which is the actual thing SPARK-57298 changed. Spark's own test asserts values (`Row(Seq(0.0d, 1.0d, Double.NaN))`). A `size` of 3 for the `mixed` group would pass even if Comet kept a non-canonical NaN payload. Could these use `sort_array(collect_set(...))` so the framework compares the values against Spark?

**4. `collect_set.sql` - float and double coverage is now missing on Spark 4.2+.**

The removed `cs_src_float` / `cs_src_double` blocks covered `NULL`, `Infinity`, `-Infinity` and ordinary values, and they ran under that file's `ConfigMatrix: parquet.enable.dictionary=false,true` sweep. They now only live in `collect_set_floating_fallback.sql`, which is `MaxSparkVersion: 4.1` and `expect_fallback`. So on 4.2+, where these cases finally execute natively, none of them are exercised, and the dictionary sweep is gone. Would it work to bring those two tables into `collect_set_normalization.sql` as native `query` cases with `sort_array`, so 4.2+ gets strictly more coverage than before rather than less?

**5. `collect_set_normalization.sql` - nulls and deeper nesting on the nested path.**

`NormalizeFloatingNumbers.normalize` takes visibly different branches depending on shape. A nullable struct becomes `KnownFloatingPointNormalized(If(IsNull(expr), Literal(null, structType), CreateNamedStruct(...)))`, and an array becomes `ArrayTransform` with a lambda. `cs_norm_nested` has no NULL struct row, no NULL array row, no array containing a NULL element, and no double nesting such as `array<struct<v:double>>` or `struct<a:array<double>>`. Since a serde gap in any of those produces a silent fallback rather than a wrong answer, `query` cases are what actually prove the native path stays native. Would you mind adding a few rows?

**6. `collect_set_normalization.sql` - add the global aggregate case.**

Both repartitioned queries have a `GROUP BY`. Spark's test uses the no-`GROUP BY` form, `Seq(Double.NaN, Double.NaN, 0.0d, -0.0d, 1.0d).toDF("v").repartition(3)` then `sort_array(collect_set(v))`, which goes through a different aggregate shape. The `grp = 'zero'` query here has no `GROUP BY` but also no repartition, so it likely stays in one partition and never merges across partial buffers. Adding the repartitioned global variant would close that.

**7. Confirm the Spark SQL test is unblocked.**

The issue says `DataFrameAggregateSuite` "SPARK-57298: collect_set normalizes NaN and -0.0 for floating-point types" is marked `IgnoreComet` in `dev/diffs/4.2.0.diff`. That file does not exist on this branch (only `3.4.3`, `3.5.8`, `4.0.2`, `4.1.2`), so there may be nothing to remove. Could you confirm? The fix is only proven end to end once that test runs against Comet, and there is a second one to account for, "SPARK-57298: collect_set normalizes NaN and -0.0 nested in complex types". If the 4.2 diff is landing separately, a tracking issue linked from here would keep it from getting lost.

andygrove (commented):
A follow-up pass on top of my review above. I went through the Spark 4.2 source and the native side, and most of what I chased came back clean, so let me record that first and then raise one thing and answer my own point 7.

Things I checked that turned out correct, so nobody needs to re-verify them:

- Struct field names survive normalization. The struct branch of `normalize` builds its replacement with `fieldNames.zipWithIndex` and an explicit `Literal(name)`, so `struct<v:double>` stays `struct<v:double>`. The `serializeDataType(expr.dataType)` computed from the original child still matches what native produces.
- Spark 4.2 changed `CollectSet.bufferElementType` to `LongType` / `IntegerType` for double / float, since `convertToBufferElement` now keys on `doubleToLongBits`. That does not leak into Comet, because `adjustOutputForNativeState` in `operators.scala` already rewrites the `CollectSet` buffer attribute to `ArrayType(cs.children.head.dataType)`. Together with the `hasNativeArrayBufferAgg` guard forbidding mixed execution, there is no cross-engine buffer hazard.
- `normalize_float` in `native/spark-expr/src/math_funcs/internal/normalize_nan.rs` maps any NaN to `T::nan()` and `-0.0` to `0.0`, matching Spark's `DOUBLE_NORMALIZER` and `FLOAT_NORMALIZER` exactly. A non-canonical NaN payload will not slip through.
- The nullable-struct branch of `normalize` emits `Literal(null, structType)`, and `CometLiteral.getSupportLevel` sets `allowComplex = expr.value == null`, so that literal serializes and the nested path does not silently fall back.
- `normalize` does have a `TransformValues` branch for maps, but it is unreachable here, because `CollectSet.checkInputDataTypes` rejects any type containing `MapType`.
- There is no cost for non-floating-point children, since `needNormalize` short-circuits and `normalize` returns the expression untouched.

**8. `aggregates.scala:855` - the pre-4.2 reason documents one divergence when there are two.**

This builds on point 1. Since that string is being rewritten anyway, it is worth making it complete. On 3.4 through 4.1 there are two differences and they point in opposite directions.

Comet's native set is `HashSet<ScalarValue>` by way of `DistinctArrayAggAccumulator`, and `ScalarValue` compares and hashes `Float32` and `Float64` by `to_bits()`, so `-0.0` and `0.0` stay as two entries. Spark before 4.2 buffers into `mutable.HashSet[Any]` of boxed doubles where Scala's `==` is IEEE, so `-0.0 == 0.0` collapses to one entry while `NaN == NaN` is false and the NaNs are kept apart. That gives:

| case | Spark pre-4.2 | Comet pre-4.2 |
| --- | --- | --- |
| two NaNs | 2 entries | 1 entry |
| `-0.0` and `0.0` | 1 entry | 2 entries |

Only the NaN row is documented today. `spark.comet.exec.strictFloatingPoint` defaults to `false`, so both are live by default on those versions, and this bullet is the only warning a user gets. The config's own doc already uses "comparing or sorting -0.0 and 0.0" as its headline example, which makes the omission read as accidental. Could signed zero be added here, and to the `what` string passed to `strictFloatingPointReason` on line 872, which carries the same wording?

Related: the `'d'` group in the `cs_src_float` and `cs_src_double` tables this PR moves into `collect_set_floating_fallback.sql` is `0.0, -0.0, 1.0, NULL`, which is exactly the signed-zero case, and in that file it only asserts fallback. So the reason string is the only record of that behavior, which is a further argument for getting its wording right.

**9. Answering my own point 7 so you do not have to.**

There is no `dev/diffs/4.2*.diff` in the repo. `dev/diffs/` holds 3.4.3, 3.5.8, 4.0.2 and 4.1.2, and `ci.yml` wires `spark_sql_test_reusable.yml` for those same four versions, so there is no Spark SQL Tests job for 4.2 at all. Nothing to un-ignore.

The consequence matters. The SQL file tests in this PR are the only thing verifying this fix, so points 3 through 6 are carrying all of the weight. Could you file a tracking issue for enabling the two `DataFrameAggregateSuite` tests, "SPARK-57298: collect_set normalizes NaN and -0.0 for floating-point types" and "SPARK-57298: collect_set normalizes NaN and -0.0 nested in complex types", once a 4.2 diff lands, and link it here?

**Concrete vectors for points 3, 5 and 6.**

Spark's own two tests give you most of what those points ask for:

```sql
-- value-asserting scalar cases
SELECT collect_set(v) FROM VALUES (double('NaN')), (double('NaN')) AS t(v)     -- Seq(NaN)
SELECT collect_set(v) FROM VALUES (-0.0D), (0.0D) AS t(v)                      -- Seq(0.0)
SELECT collect_set(v) FROM VALUES (float(-0.0)), (float(0.0)) AS t(v)          -- Seq(0.0f)
-- Seq(NaN, NaN, 0.0, -0.0, 1.0) repartitioned to 3, then sort_array           -- Seq(0.0, 1.0, NaN)

-- value-asserting nested cases
SELECT collect_set(named_struct('a', v)) FROM VALUES (-0.0D), (0.0D) AS t(v)   -- Seq(Row(0.0))
SELECT collect_set(a) FROM VALUES (array(-0.0D)), (array(0.0D)) AS t(a)        -- Seq(Seq(0.0))
SELECT collect_set(a) FROM VALUES (array(float(-0.0))), (array(float(0.0))) AS t(a)
SELECT collect_set(named_struct('a', v)) FROM VALUES (double('NaN')), (double('NaN')) AS t(v)
SELECT collect_set(a) FROM VALUES (array(double('NaN'))), (array(double('NaN'))) AS t(a)
SELECT collect_set(a) FROM VALUES (array(float('NaN'))), (array(float('NaN'))) AS t(a)
```

Note that `array<float>` NaN and the struct NaN guardrail are in Spark's set but not in `cs_norm_nested`. The NULL rows and the double nesting from point 5 are missing from Spark's coverage too, so those still need writing from scratch.

peterxcli:
@andygrove thanks for the review! review change is pushed, issue is opened per your review. please take another look, thanks!

> 1. Gate incompatibility reasons by Spark version

`getIncompatibleReasons()` now returns `Nil` on Spark 4.2+, while retaining the compatibility warning for Spark 3.4–4.1.

> 2. Normalize `Complete` aggregates

Floating-point normalization now applies to both `Partial` and `Complete` modes. Merge modes remain unchanged because they consume aggregate buffers rather than raw rows.

> 3. Assert collected values

The regression queries now compare `sort_array(collect_set(...))` results instead of only checking collection sizes.

> 4. Restore scalar coverage

Added float and double cases covering ordinary values, NULL, NaN, positive/negative infinity, and signed zero. Tests run with Parquet dictionary encoding both enabled and disabled.

> 5. Expand nested-value coverage

Added null structs, null arrays, arrays containing null elements, `array<struct<double>>`, and `struct<array<double>>`, including NaN and signed-zero values.

> 6. Test global aggregation

Added a repartitioned `collect_set` query without `GROUP BY` to exercise partial-buffer merging into a global aggregate.

> 7. Check Spark 4.2 SQL-test diffs

Confirmed that no `dev/diffs/4.2*.diff` or Spark 4.2 Spark SQL Tests job currently exists. No diff file was modified.

> 8. Document both pre-4.2 divergences

The compatibility messages now explain that Comet deduplicates NaNs but distinguishes `-0.0` from `0.0`, while pre-4.2 Spark does the opposite. Negative-zero fixtures now use string casts so the sign is preserved.

> 9. File a Spark 4.2 tracking issue

Filed https://github.com/apache/datafusion-comet/issues/5209 to enable the two SPARK-57298 `DataFrameAggregateSuite` tests once the Spark 4.2 diff and CI job are added.

> 10. Keep the original compatibility wording

Restored the original “Before Spark 4.2…” wording and added only the missing signed-zero explanation.

> 11. Remove the mode explanation comment

Removed the added explanatory comment above `normalizeFloatingNumbers`; the `Partial | Complete` behavior remains unchanged.

spark/src/main/scala/org/apache/comet/serde/aggregates.scala
```diff
@@ -861,7 +872,12 @@ object CometCollectSet extends CometAggregateExpressionSerde[CollectSet] {
       inputs: Seq[Attribute],
       binding: Boolean,
       conf: SQLConf): Option[ExprOuterClass.AggExpr] = {
-    val child = expr.children.head
+    val child = aggExpr.mode match {
+      case Partial | Complete if isSpark42Plus =>
+        CometExecUtils.normalizeFloatingNumbers(expr.children.head)
```
andygrove, line 889: One consequence of reusing `normalize` that I would like called out somewhere.

For any array-typed child it produces `KnownFloatingPointNormalized(ArrayTransform(expr, lambda))`, and `CometArrayTransform` is a `CometCodegenDispatch` rather than a native serde. So on 4.2+, `collect_set` over `array<float>`, `array<double>`, `array<struct<v:double>>` and `struct<a:array<double>>` now compiles Spark's `doGenCode` into a per-batch JVM kernel instead of running natively end to end. Before this PR those ran purely natively, just with the wrong answer, so this is the right trade. The scalar path is unaffected, since it maps onto the native `NormalizeNaNAndZero`. The plain struct path is unaffected too, since `If`, `IsNull`, `CreateNamedStruct` and `GetStructField` all have native serdes. It is specifically the array shapes.

Two follow-ups. Could the PR description mention this so the execution-mode change is on the record? And would you file an issue to teach the native `NormalizeNaNAndZero` in `native/spark-expr/src/math_funcs/internal/normalize_nan.rs` to recurse into `List` and `Struct`, so the serde can emit a single native node for the nested shapes?

Worth noting in that issue that with `spark.comet.exec.scalaUDF.codegen.enabled=false` the `ArrayTransform` serializes to `None`, which makes `convert` return `None` and the whole aggregate fall back to Spark. That degrades safely, which is good, but it is slower than it needs to be.

andygrove (commented):
Thanks for turning all nine points around. The tests are much stronger now: asserting values through `sort_array` instead of `size`, float and double scalar coverage back on 4.2+ under the dictionary matrix, the nested file gaining NULL structs, NULL arrays, NULL elements and both double-nesting shapes, and the repartitioned global aggregate. #5209 captures the Spark 4.2 diff work. CI is fully green across all four 4.2 shards plus 3.4, 3.5, 4.0, 4.1, Iceberg, TPC-H and TPC-DS.

I went back through Spark master and the Comet side again. A batch of things came back clean, so let me record them first so nobody re-verifies them.

- `normalize` short-circuits on `KnownFloatingPointNormalized`, so a child the optimizer already normalized will not be double-wrapped.
- Spark's `complexNormalizer` builds its projection from `BoundReference(0, child.dataType, nullable = true)` and so always takes the `If(IsNull(...))` struct branch. Comet passes the real child, which may be non-nullable and take the plain `KnownFloatingPointNormalized(struct)` branch. Those are semantically identical, and the reconstructed struct and array types match the original nullability, so `serializeDataType(expr.dataType)` still agrees with what the normalized child produces.
- The `TransformValues` map branch stays unreachable, since `CollectSet.checkInputDataTypes` rejects anything containing a `MapType`.
- `Complete` is genuinely defensive today. `CometWindowExec` only accepts `Average`, `Count`, `First`, `Last`, `Max`, `Min` and `Sum` as window aggregates, so `CollectSet` cannot reach the `Complete` assert there.
- The `nan` groups in the new tests cannot fail. `CAST('NaN' AS DOUBLE)` produces the canonical payload, which Comet's bit-keyed set already collapsed before this PR. The `zero` groups are what actually exercise the change, and they are present in the scalar case and in every nested column, so coverage is fine as written.
- `checkSparkAnswerAndOperator` runs with `assertCometNative = true`, so the new `query` cases really do prove the aggregate stayed native. They are not vacuous.
- Dropping `Config: spark.comet.exec.strictFloatingPoint=true` from `collect_set.sql` is fine, since no floating-point query remains in that file.

Three things I would still like addressed, left as inline comments.

spark/src/main/scala/org/apache/comet/serde/aggregates.scala
```diff
@@ -861,7 +872,12 @@ object CometCollectSet extends CometAggregateExpressionSerde[CollectSet] {
       inputs: Seq[Attribute],
       binding: Boolean,
       conf: SQLConf): Option[ExprOuterClass.AggExpr] = {
-    val child = expr.children.head
+    val child = aggExpr.mode match {
+      case Partial | Complete if isSpark42Plus =>
```
andygrove, line 888: Could you say a bit about why the normalization is gated on `isSpark42Plus`?

Normalizing NaN is a no-op for Comet on every version, since the native set already keys on `to_bits()` and every NaN Spark can hand it is canonical. So dropping the gate would change exactly one thing on 3.4 through 4.1: `-0.0` and `0.0` would collapse to a single entry, which is what Spark does on those versions. That is precisely the divergence your new `getIncompatibleReasons()` text documents. The NaN difference would remain and `collect_set` would stay `Incompatible` there, but users would be one divergence away instead of two.

The cost is that array-typed children would start paying the codegen dispatch from my other comment on the older versions as well, so I can see this going either way. If you would rather not take it on here, a tracking issue linked from this PR works. I would just like the reasoning captured somewhere rather than left implicit in the gate.

spark/src/main/scala/org/apache/comet/serde/aggregates.scala
```diff
@@ -844,12 +853,14 @@ object CometCollectSet extends CometAggregateExpressionSerde[CollectSet] {
     // analysis time, and CometCollectShim.ignoreNulls hardcodes true, making this a no-op.
     if (!CometCollectShim.ignoreNulls(expr)) {
       Unsupported(Some("collect_set with RESPECT NULLS (ignoreNulls = false) is not supported"))
```
andygrove, line 864: Now that `getIncompatibleReasons()` returns `Nil` on 4.2+, there is nothing left to render on the Spark 4.2 compatibility page, so `collect_set` will show up there as fully supported with no caveats at all. But `RESPECT NULLS` still falls back on this line.

Could you add the matching `getUnsupportedReasons()`, gated the same way so it does not appear on the 3.4 through 4.1 pages where the field does not exist?

```scala
override def getUnsupportedReasons(): Seq[String] =
  if (isSpark42Plus) {
    Seq("`collect_set` with `RESPECT NULLS` falls back to Spark, since the native " +
      "implementation always drops null inputs.")
  } else {
    Nil
  }
```

spark/src/main/scala/org/apache/comet/serde/aggregates.scala
peterxcli: > Two follow-ups. Could the PR description mention this so the execution-mode change is on the record? And would you file an issue to teach the native `NormalizeNaNAndZero` in `native/spark-expr/src/math_funcs/internal/normalize_nan.rs` to recurse into `List` and `Struct`, so the serde can emit a single native node for the nested shapes?

1. 
2. teach the native `NormalizeNaNAndZero` to recurse into `List` and `Struct`: plan to address this issue along with my new review: 
   1. https://github.com/apache/datafusion-comet/pull/5235#pullrequestreview-4889228174, 
   2. https://github.com/apache/datafusion-comet/pull/5235#discussion_r3741122515, 
   3. https://github.com/apache/datafusion-comet/pull/5235#discussion_r3741131088

> Worth noting in that issue that with `spark.comet.exec.scalaUDF.codegen.enabled=false` the `ArrayTransform` serializes to `None`, which makes `convert` return `None` and the whole aggregate fall back to Spark. That degrades safely, which is good, but it is slower than it needs to be.

I added this in https://github.com/apache/datafusion-comet/issues/5312 follow-up per another inline review, https://github.com/apache/datafusion-comet/pull/5166#discussion_r3732293880

spark/src/main/scala/org/apache/comet/serde/aggregates.scala
peterxcli: filed a https://github.com/apache/datafusion-comet/issues/5312 as follow-up

andygrove:

Reusing Spark's own `NormalizeFloatingNumbers.normalize` instead of reimplementing the recursion is the right instinct, and gating it to `Partial` and `Complete` so that merge paths are untouched looks correct to me. Splitting the SQL coverage into a version-specific normalization file and a fallback file is a good way to keep both behaviors pinned.

A few things.

**The array case now costs a JVM round trip on every batch**

For an `ARRAY<DOUBLE>` input, `normalize` produces an `ArrayTransform`, which Comet runs through the codegen dispatcher. So on Spark 4.2, `collect_set` over a float-containing array goes from fully native to per-batch JVM dispatch. That is a correctness-driven trade and I am not arguing against it, but it should be visible to users rather than only in this PR description. Could `getIncompatibleReasons` or the compatibility docs say that array-typed `collect_set` on floating-point elements is no longer end-to-end native on 4.2+?

A rough number would also help. How much slower is the dispatched path for a `collect_set(array<double>)` over a realistic batch?

**Depending on a Spark internal**

`NormalizeFloatingNumbers.normalize` is `private[sql]` with no stability guarantee, and `CometExecUtils` gets at it by living in `org.apache.spark.sql.comet`. That is a legitimate technique and Comet does it elsewhere, but it is worth a note in the doc comment saying which Spark versions the method exists in and what to do if it changes shape. Right now the comment explains what the wrapper does but not that it is a compatibility risk.

Related: is there a test that would fail loudly if a future Spark renamed or moved it? A compile failure is fine as a signal, but only if some profile actually compiles this path.

**How does the version-specific SQL coverage get gated?**

`collect_set_normalization.sql` is described as 4.2+ and `collect_set_floating_fallback.sql` as pre-4.2. What mechanism keeps each from running on the wrong version? If it is a `-- Config:` header or a directory convention, saying so in a comment at the top of each file would help, otherwise the next person adding a version-specific fixture has to reverse-engineer it.

**One check on idempotency**

If Spark 4.2 performs this normalization itself somewhere in the analyzer or optimizer, Comet would be normalizing an already-normalized child. `needNormalize` short-circuits on `KnownFloatingPointNormalized`, so I believe it is a no-op, but it would be worth confirming with a test that asserts the serialized plan contains exactly one `NormalizeNaNAndZero` rather than two. Where in Spark 4.2 does the normalization actually live? A link in the comment would make the whole change much easier to verify.

peterxcli:
Thanks for the detailed look. Point-by-point:

**Array JVM dispatch**: fair ask on visibility — I've added a `getCompatibleNotes()` entry (which feeds the generated Compatibility Guide) documenting that array inputs containing floats route through the JVM codegen dispatcher on 4.2+, while scalar/struct inputs stay native, and that `spark.comet.exec.scalaUDF.codegen.enabled=false` falls those cases back to Spark. One framing correction: pre-PR the array path wasn't "fully native and fast" in any useful sense — it was fully native and *wrong* (#4966), since `spark.comet.exec.strictFloatingPoint` defaults to `false`. Dispatch is the fastest correct option today. Benchmarking the dispatch cost is tracked in #5312, which also covers whether to extend normalization to 3.4–4.1.

**Spark internal**: expanded the doc comment on `CometExecUtils.normalizeFloatingNumbers` to note it's `private[sql]`, exists in all supported versions (3.4–4.2), and that any upstream rename fails the profile build loudly — it's a direct compile-time reference (not reflection) compiled against every Spark profile in CI, so the "loud test" is the build itself. Added guidance to shim per-version if it ever changes shape.

**SQL gating**: the mechanism is the `-- MinSparkVersion: 4.2` / `-- MaxSparkVersion: 4.1` headers at the top of each file — they're parsed directives (`SqlFileTestParser`), documented in the [SQL file tests contributor guide](https://datafusion.apache.org/comet/contributor-guide/sql-file-tests.html), not a directory convention.

**Idempotency**: Spark 4.2's normalization (SPARK-57298, apache/spark@49908a2cf92) lives inside `CollectSet.convertToBufferElement`/`eval`, not in the analyzer/optimizer — so the plan Comet serializes never arrives pre-wrapped and double-normalization cannot occur; `needNormalize` short-circuiting on `KnownFloatingPointNormalized` covers the hypothetical anyway. I've added the SPARK-57298 reference to the code comment. Given the wrap can't happen by construction, I'd skip the serialized-plan assertion test.

andygrove (approved):
Everything from both earlier rounds is in and CI is green across the profiles.

I checked the one thing I was unsure about. `NormalizeFloatingNumbers.normalize` on a nullable struct produces `If(IsNull(e), Literal(null, st), CreateNamedStruct(...))`, and that shape has failed natively before now when the constructed fields are non-nullable while the batch carries nulls in them. It is fine here. The children come from `GetStructField` on a nullable parent, so they are nullable in the Spark type too and the mismatch cannot arise. `collect_set_normalization.sql` covers the null struct row and `struct<array<double>>` directly on 4.2, so it is pinned rather than argued.

The `getCompatibleNotes()` entry reads well on the generated page, and gating `getUnsupportedReasons()` on `isSpark42Plus` lines up with where `CometCollectShim.ignoreNulls` can actually return false.

Approving.

peterxcli:
@andygrove Thanks!

PR #5167, opened 2026-07-31
fix: surface next_day and make_date ANSI errors as Spark exceptions

### LINKED ISSUE #5073: next_day and make_date ANSI errors surface as CometNativeException instead of Spark exception classes
### Describe the bug

`next_day` and `make_date` honor ANSI mode correctly (throw vs NULL), and `make_date` even reproduces the `java.time` message text and validation order faithfully. But both raise plain `DataFusionError::Execution` strings instead of `SparkError` variants, so under ANSI they surface as `CometNativeException` instead of the Spark exception types:

- `next_day` with an invalid day name (`native/spark-expr/src/datetime_funcs/next_day.rs:146-155`): Spark throws `SparkIllegalArgumentException` with error class `ILLEGAL_DAY_OF_WEEK`, SQLSTATE `22009`. Comet throws a generic exception with only the message substring `"Illegal input for day of week: ..."`.
- `make_date` with out-of-range fields (`native/spark-expr/src/datetime_funcs/make_date.rs:177-187`): Spark throws `SparkDateTimeException` with error class `DATETIME_FIELD_OUT_OF_BOUNDS.WITH_SUGGESTION`, SQLSTATE `22023`, including the `spark.sql.ansi.enabled` suggestion sentence. Comet emits only the inner range message.

`native/common/src/error.rs` has no variants for these two error classes; adding e.g. `IllegalDayOfWeek { input }` and `DatetimeFieldOutOfBounds { range_message }` and routing them through the existing JSON conversion path (as done for cast and arithmetic ANSI errors, `ShimSparkErrorConverter`) closes the gap.

Existing tests (`spark/src/test/resources/sql-tests/expressions/datetime/next_day_ansi.sql`, `make_date_ansi.sql`) assert message substrings only, so they pass despite the wrong exception type.

Minor related note: Spark flips `MakeDate.nullable` to depend on child nullability when `failOnError` is set (`datetimeExpressions.scala:2570`); Comet does not model this.

### Steps to reproduce

```sql
SELECT next_day(date'2024-01-01', 'NOT_A_DAY');   -- Spark: SparkIllegalArgumentException [ILLEGAL_DAY_OF_WEEK]
SELECT make_date(2024, 13, 1);                     -- Spark: SparkDateTimeException [DATETIME_FIELD_OUT_OF_BOUNDS.WITH_SUGGESTION]
```

Comet throws `CometNativeException` in both cases.

### Expected behavior

Exception type, error class, and SQLSTATE match Spark.

### Additional context

peterxcli:

Closes #5073.

## Rationale for this change

Under ANSI mode, native `next_day` and `make_date` return generic DataFusion execution errors. Those bypass Comet's structured `SparkError` conversion at the JNI boundary and surface as `CometNativeException`, losing Spark's exception class, error class, and SQLSTATE.

- Add typed `IllegalDayOfWeek` and `DatetimeFieldOutOfBounds` Spark errors and emit them from the native functions.
- Convert the errors through the version-specific Spark shims, preserving Spark 3.x legacy errors and Spark 4.x structured errors.
- Add an end-to-end regression that verifies native execution and matches Spark's exception class, error class, and SQLSTATE without `CometNativeException`.
- Remove the resolved compatibility notes and correct the Spark-version notes in the existing ANSI SQL fixtures.

- `make core`
- Focused `CometTemporalExpressionSuite` regression on Spark 3.4, 3.5, 4.0, 4.1, and 4.2: `next_day and make_date ANSI errors match Spark exceptions`
- Maven Spotless and Scalastyle checks


andygrove (commented):
Thanks for this, and thanks for the thorough PR description. I checked the shim mapping against Spark 3.4.3, 3.5.8, 4.0.1, 4.1.0, and master, and the version handling is correct:

- `NextDay` under ANSI on 3.4/3.5 throws a plain `IllegalArgumentException` from `DateTimeUtils.getDayOfWeekFromString` and wraps it in `QueryExecutionErrors.ansiIllegalArgumentError` (`_LEGACY_ERROR_TEMP_2000`). On 4.0+ it throws `SparkIllegalArgumentException` with `ILLEGAL_DAY_OF_WEEK` and a `string` parameter. Both shim branches match exactly, including the reconstructed 3.x message text.
- `MakeDate` under ANSI uses `ansiDateTimeError` on 3.4/3.5 and `ansiDateTimeArgumentOutOfRange` on 4.x. Delegating to the same `QueryExecutionErrors` helper Spark itself calls is the right call, because the error class behind that helper has moved three times (`DATETIME_FIELD_OUT_OF_BOUNDS` on 4.0, `.WITH_SUGGESTION` on 4.1, `.WITHOUT_SUGGESTION` on master) and the shim tracks it for free.
- Neither Spark error carries a `QueryContext`, and the shim correctly ignores the one Comet threads through.

I also went through every branch of `invalid_date_message` against `LocalDate.of`. It is a careful reproduction of the `java.time` wording, including the uppercase month enum name in `Invalid date 'APRIL 31'` and the differently cased `Invalid date 'February 29' as '2023' is not a leap year`. One field looks uncovered, which is my main comment below.

**`invalid_date_message` is missing the `Year` field**

`LocalDate.of` validates `YEAR` before `MONTH_OF_YEAR`, and the valid range is -999999999 to 999999999. So `make_date(1000000000, 1, 1)` gives Spark:

```
Invalid value for Year (valid values -999999999 - 999999999): 1000000000
```

whereas `invalid_date_message` passes the month and day checks and falls through to `Invalid date 'JANUARY 1'`. Before this PR that string was buried inside a `CometNativeException`, but now it becomes the `rangeMessage` parameter of a real `DATETIME_FIELD_OUT_OF_BOUNDS` error, so it is the message the user actually reads. Would you mind adding a `Year` branch as the first check, so the field ordering matches `java.time` as well?

**`make_date` is now advertised as unqualified `Compatible()`**

`CometMakeDate` has no `getSupportLevel` override, so removing `getCompatibleNotes()` leaves the expression with no caveats at all. I think there is still a divergence hiding there. `NaiveDate::from_ymd_opt` is bounded by chrono's `MAX_YEAR`, which is `(i32::MAX >> 13) - 1`, or 262142. That is far narrower than `java.time`'s 999999999. So `make_date(300000, 6, 15)` is a perfectly valid date to Spark, but Comet's `make_date` returns `None` for it. That means NULL with ANSI off, and with this PR it becomes a well formed `SparkDateTimeException` claiming the date is out of bounds when Spark would have returned a value.

That part is pre-existing and I am not asking you to fix chrono's range here. Could you confirm the behavior and file a tracking issue, then link it either from a `getSupportLevel` note or from the compatibility index? My concern is that this PR makes the wrong answer look authoritative while removing the last remaining caveat on the expression.

**The new test does not assert the message**

`next_day and make_date ANSI errors match Spark exceptions` checks the exception class, error class, and SQLSTATE, but not the message. The message is the one dimension where I found an actual gap, so it would be worth closing. Comparing `getMessageParameters` would pin `rangeMessage` and `string` directly and would catch the year case above. A third query with a year outside `java.time`'s range would exercise it.

The `assert(!causeChain(cometFailure).exists(_.isInstanceOf[CometNativeException]))` guard is a good touch. It is what makes this a real regression test rather than a coincidence.

**Consider moving the assertion helpers into `CometTestBase`**

`causeChain` and `deepestSparkThrowable` are the reusable part of this test, and the two remaining bullets in the same compatibility index section (5072 and 4967) are the same class of problem for other datetime and arithmetic functions. Could they move into `CometTestBase` as something like `checkSparkExceptionMatches(df)`? Otherwise the next ANSI parity fix will copy them and the two copies will drift.

**The corrected note in `make_date_ansi.sql` will go stale again**

It says `DATETIME_FIELD_OUT_OF_BOUNDS.WITH_SUGGESTION on 4.1+`, but `ansiDateTimeArgumentOutOfRange` on Spark master switched to `DATETIME_FIELD_OUT_OF_BOUNDS.WITHOUT_SUGGESTION`, and CI runs 4.2. Since the shim delegates to the helper rather than hardcoding a class, the behavior is version agnostic. Could the comment say the same? Something like "the exact subclass varies by version (`DATETIME_FIELD_OUT_OF_BOUNDS` on 4.0, `.WITH_SUGGESTION` on 4.1, `.WITHOUT_SUGGESTION` on 4.2+)" would stop it going stale on the next Spark bump.

peterxcli:
@andygrove thanks for the review, push review change and filed two issue per your review!

> `invalid_date_message` should validate the year before the month and day.

Added Java `LocalDate` year-range validation before month/day validation, including Spark’s exact error message. Added a combined invalid-input case to verify year-first precedence.

> Compare the exception message parameters and add an out-of-range year test.

The test now compares exception class, error class, SQL state, `getMessageParameters`, and the full message. Added cases for `make_date(1000000000, 1, 1)` and an invalid year/month/day combination.

> Document the narrower chrono year range and create a tracking issue.

Filed https://github.com/apache/datafusion-comet/issues/5208 and restored `CometMakeDate.getCompatibleNotes()` documenting chrono’s `[-262143, 262142]` limitation.

> Consider moving `causeChain` and `deepestSparkThrowable` into `CometTestBase`.

Found five equivalent cause-chain traversals across four suites. Filed https://github.com/apache/datafusion-comet/issues/5223) to centralize `causeChain` in `CometTestBase`. `deepestSparkThrowable` remains local because its selection policy is unique, but it can reuse the shared helper.

> Avoid hardcoding the Spark-version-specific `DATETIME_FIELD_OUT_OF_BOUNDS` subclass.

Reworded the SQL test comment to state that the subclass and message parameters vary by Spark version, without pinning a specific version boundary.

andygrove (commented):
Thanks for the detailed reply, and thanks for filing #5208 and #5223. Both issues are well written. #5208 captures the chrono range accurately, and #5223 does a careful job separating the five genuine full-chain traversals from the call sites that intentionally unwrap only one level. For the record the `[-262143, 262142]` range is correct. chrono 0.4.45 has `MIN_YEAR = (i32::MIN >> 13) + 1` and `MAX_YEAR = (i32::MAX >> 13) - 1` in `naive/date/mod.rs`, so those are the right astronomical year numbers.

I think the code changes did not make it up, though. The branch is still at `05c2828ba`, committed `2026-07-31T14:07:56Z`, which is before this review, and there is no second commit or force-push. Checking each item against `pull/5167/head`:

- `invalid_date_message` still begins at the month check (`native/spark-expr/src/datetime_funcs/make_date.rs:75`). No `Year` branch.
- The test still asserts only class, error class, and SQL state (`spark/src/test/scala/org/apache/comet/CometTemporalExpressionSuite.scala:69-71`). No message or `getMessageParameters` comparison.
- `CometMakeDate` has no `getCompatibleNotes` and no `getSupportLevel` (`spark/src/main/scala/org/apache/comet/serde/datetime.scala:471`).
- `make_date_ansi.sql:20` still reads `DATETIME_FIELD_OUT_OF_BOUNDS.WITH_SUGGESTION on 4.1+`.

Could you push the commit? Everything else below is meant to help when you do.

## Re-verified against Spark master, branch-4.1, and branch-3.5

The shim mapping is still exactly right, and two details are worth passing on.

`QueryExecutionErrors.ansiDateTimeArgumentOutOfRange` on Spark master now produces `DATETIME_FIELD_OUT_OF_BOUNDS.WITHOUT_SUGGESTION` (`QueryExecutionErrors.scala:308`), and 4.1 added a separate `ansiDateTimeArgumentOutOfRangeWithoutSuggestion` alongside the `.WITH_SUGGESTION` one. Delegating to the helper rather than naming a subclass is confirmed correct on 3.5, 4.1, and master, which is exactly why the fixture comment should stop pinning a version. Your version-agnostic rewording is the right call.

More importantly, a boundary caution for the year test. On master the ANSI path moved into `DateTimeExpressionUtils.makeDateExact`, and its Javadoc is explicit that only `DateTimeException` is converted. A `toIntExact` day-count overflow from `localDateToDays` "propagates to the caller unchanged." So `make_date(999999999, 1, 1)` throws a plain `java.lang.ArithmeticException` on the Spark side with no `SparkThrowable` anywhere in the cause chain, and `deepestSparkThrowable` would call `fail()` on it. `make_date(1000000000, 1, 1)` is the correct value because `YEAR.checkValidValue` rejects it before the day-count conversion ever runs. Could you add a short comment in the test explaining why that specific value and not `999999999`? Otherwise someone will reasonably tighten it to the real `java.time` maximum and get a confusing failure.

## The chrono gap produces a factually wrong message, not just a wrong value

This is for the restored note, and it sharpens the concern I raised about removing the last caveat.

`native/spark-expr/src/datetime_funcs/make_date.rs:81` is reached whenever `make_date()` returned `None`, which includes years outside chrono's range. So `make_date(300000, 2, 29)` yields `Invalid date 'February 29' as '300000' is not a leap year`. 300000 is divisible by 400, so it is a leap year. Before this PR that sentence was buried inside a `CometNativeException`. After it, it becomes the `rangeMessage` of a well formed `SparkDateTimeException`, so a user reads a false statement presented as a real Spark error.

Could the restored `getCompatibleNotes()` say that the message itself can be wrong for out-of-range years, not just that the throw/NULL decision differs? That is the part that gets worse rather than better here. It would be worth being explicit about in #5208 too.

## Where the note should live

Confirming it belongs on `CometMakeDate` (`spark/src/main/scala/org/apache/comet/serde/datetime.scala:471`) and should link #5208. As it stands that object has neither `getSupportLevel` nor `getCompatibleNotes`, so the generated `datetime.md` will show `make_date` with no caveats at all once this merges. The "Known result-value divergences" section of `compatibility/index.md` also looks like a natural second home for it, since that section exists precisely for silently wrong values, whereas the bullet you removed was in the ANSI error-class section.

## Things I checked that are fine

`ILLEGAL_DAY_OF_WEEK` on master is thrown directly by `DateTimeUtils.getDayOfWeekFromString` with no `QueryExecutionErrors` helper, so the 4.x shim constructing `SparkIllegalArgumentException` inline mirrors Spark rather than diverging from the pattern. The 3.5 message reconstruction is byte exact, including the absent trailing period (`DateTimeUtils.scala:368`). `params` is a `Map[String, Any]` from json4s, so string values interpolate cleanly with no quoting artifact. `getErrorClass` still delegates to `getCondition` on master, so the test compiles and behaves across 3.4 through 4.2. The `test_error_class_mapping` and `test_exception_class_mapping` tests in `error.rs` are spot checks rather than exhaustive tables, so not extending them is fine.

CI is fully green across all five Spark versions, but note that run is on `05c2828ba`.

peterxcli:
> I think the code changes did not make it up, though. The branch is still at `05c2828ba`, committed `2026-07-31T14:07:56Z`, which is before this review, and there is no second commit or force-push. Checking each item against `pull/5167/head`:

sorry I forgot to push local commit. also thanks for the new round of review, addressing them!

peterxcli:
@andygrove thanks for the another round of review.

Addressed all required feedback:

- Added the `1000000000` rationale in `CometTemporalExpressionSuite.scala`.
- Documented the incorrect leap-year message in `datetime.scala`.
- Updated https://github.com/apache/datafusion-comet/issues/5208 with the reproduction and wrong-message behavior.

andygrove (approved):
LGTM. Thanks @peterxcli

PR #5168, opened 2026-07-31
chore: replace deprecated symbol literals with $"col" in tests

andygrove:

No issue filed — this is a mechanical compiler-warning cleanup, split out of #5141 because it accounts for the largest single group of warnings on its own.

## Rationale for this change

Symbol literals (`'col`) are deprecated in Scala 2.13. They account for **63 of the 91 Scala warnings** in the test sources under the 2.13 profiles — by far the largest single group of Scala compiler warnings in the build, and all of them are in two files. Splitting them out of #5141 keeps that PR reviewable and lets this purely mechanical change go in on its own.

`'col` replaced with the equivalent `$"col"` string interpolator in:

- `spark/src/test/scala/org/apache/spark/sql/TPCH.scala` (62 occurrences — the TPC-H table schema definitions)
- `spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala` (1 occurrence)

No behavior change: both forms resolve through the same Catalyst DSL implicits to the same attribute. `Table("region", ...)` is reformatted onto multiple lines by spotless because the longer names push it past the line limit — that is the only non-substitution change.

These are the last symbol literals in the repo, so this group of warnings drops to zero.

No new tests — mechanical cleanup with no intended behavior change.

- Test sources warning count under the default profile (Spark 4.1 / Scala 2.13): **91 → 28**, with all 63 symbol-literal warnings gone.
- `test-compile` succeeds on the default profile, `-Pspark-3.5`, and `-Pspark-3.4`.
- Relying on CI to run the affected suites.


mbutrovich (approved):
Thanks @andygrove!

Issue #4124, opened 2026-04-28
SPARK-53968 SQLViewSuite: decimal arithmetic returns ~10x smaller values through view CTE on Spark 4.1.1


andygrove:
## Describe the bug

The new Spark 4.1 test `SPARK-53968 reading the view after allowPrecisionLoss is changed` in `SQLViewSuite` fails with Comet enabled. The test stores `DECIMAL(38, 18)` values and computes `unit_price + COALESCE(shipping_price, 0)` through a CTE wrapped in a view. Comet returns values approximately 10x smaller than expected:

| Row | Expected | Actual (with Comet) |
|---|---|---|
| part1 | 100.00000000000000000 | 10.00000000000000000 |
| part2 | 100.00000000000000000 | 10.00000000000000000 |
| part3 | 300.23000000000000000 | 30.02300000000000000 |

The plan involves `CometBroadcastHashJoin → CometProject(unit_price + COALESCE(shipping_price, 0E-18)) → CometExchange(rangepartitioning) → CometSort`.

## Steps to reproduce

Run Spark 4.1.1's `SQLViewSuite` with Comet enabled. The reproducer is the test body itself (see `sql/core/src/test/scala/org/apache/spark/sql/execution/SQLViewSuite.scala` `SPARK-53968 reading the view after allowPrecisionLoss is changed`).

## Expected behavior

`unit_price + shipping_price` should match Spark's result (100, 100, 300.23).

## Workaround

The test is currently tagged `IgnoreComet(...)` in `dev/diffs/4.1.1.diff`.

## Additional context

PR #4093 enables Spark 4.1.1 in the `Spark SQL Tests` workflow. The 10x discrepancy and the `DECIMAL(38, 18)` schema strongly suggest a precision/scale handling bug in Comet's decimal addition or in the view-resolution path that drops 1 digit of scale.

PR #5171, opened 2026-07-31
fix: use per-expression eval mode for decimal promotion

### LINKED ISSUE #5075: DecimalPrecision.promote uses session ANSI flag instead of per-expression eval mode
### Describe the bug

`QueryPlanSerde` derives the `nullOnOverflow` flag for decimal-promotion `CheckOverflow` wrappers from the live session config instead of the per-expression eval mode:

```scala
// spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala:787
val newExpr = DecimalPrecision.promote(expr, !SQLConf.get.ansiEnabled)
```

Spark 4.1 captures the eval mode per expression (`NumericEvalContext` on `Add`/`Subtract`/`Multiply`/`Divide`/`IntegralDivide`/`Remainder`/`Pmod`/`Sum`), precisely because the expression's mode and the session's current ANSI setting can disagree: views created under a different ANSI setting, or `try_` operations inside an ANSI session (`try_add` lowers to `Add(..., EvalMode.TRY)`, not a separate node).

This is currently latent rather than live, because the reachable decimal paths all get the mode from a per-expression source that fires first:

- overflow-capable decimal add/sub/mul takes the wide-decimal path, whose `CheckOverflow` is deliberately elided (`native/core/src/execution/planner.rs:519-526`) in favor of `WideDecimalBinaryExpr`'s own per-expression `eval_mode`;
- decimal divide gets an inner `CheckOverflow` from `CometDivide` carrying `expr.evalMode == EvalMode.ANSI` (`spark/src/main/scala/org/apache/comet/serde/arithmetic.scala:279-288`), which fires before the outer promotion wrapper;
- remainder result precision is always exact, so no overflow.

But any future change to those paths turns this into a live wrong-mode bug, and the current code reads as if the session flag were the right source.

### Expected behavior

The flag passed to `DecimalPrecision.promote` should come from the expression's own eval mode (via the shim accessors already used elsewhere, e.g. `CometEvalModeUtil`), not `SQLConf.get.ansiEnabled`.

### Additional context

peterxcli:

Closes #5075.

## Rationale for this change

`DecimalPrecision.promote` currently derives `nullOnOverflow` from the live session ANSI setting. Spark expressions capture their own evaluation mode, and one expression tree can contain mixed modes, such as a `TRY` operation nested inside an ANSI expression. Each generated `CheckOverflow` wrapper must therefore follow the evaluation mode of the arithmetic expression it guards.

- Remove the session-derived `nullOnOverflow` argument from `DecimalPrecision.promote`.
- Derive `nullOnOverflow` from each decimal arithmetic node's evaluation mode.
- Add a Spark 4.1 regression test that verifies nested ANSI and TRY expressions serialize with independent overflow behavior.

- `make core`
- `./mvnw test -Pspark-4.1 -Dtest=none -Dsuites=org.apache.spark.sql.comet.CometDecimalArithmeticViewSuite -Dscalastyle.skip=true` (1 suite, 2 tests)
- `./mvnw scalastyle:check -Pspark-4.1`
- `./mvnw spotless:apply -Pspark-4.1`


andygrove (commented):
Thanks for picking this up. The change is correct and I traced through it to confirm the reasoning.

`CometCheckOverflow.convert` serializes `failOnError = !expr.nullOnOverflow` (`math.scala:248`), so `nullOnOverflow = evalMode != EvalMode.ANSI` becomes `failOnError = evalMode == EvalMode.ANSI`. That lines up with what `CometDivide` already does at `arithmetic.scala:284` and with what `WideDecimalBinaryExpr` gets through `eval_mode`. `EvalMode` has three values and both `LEGACY` and `TRY` should return null on decimal overflow, so folding them together is right.

I also confirmed the bug stays latent, matching the analysis in #5075. For `Plus` / `Minus` / `Multiply` the native planner only builds `WideDecimalBinaryExpr` when operand precision can actually reach 38 (`planner.rs:912-919`), and there `planner.rs:523-527` elides the `CheckOverflow` because the declared type always equals `expr.dataType` on both sides. Below that threshold the result precision cannot overflow. `Divide` has `CometDivide`'s inner wrapper firing first, and `Remainder` is precision-exact. So there is no end-to-end repro available today and a proto-level unit test is the right shape of test here.

A few things I would like to see addressed before merge.

**1. The new regression test does not run on Spark 4.2.**

`spark/pom.xml:538-539` adds only `src/test/${shims.majorVerSrc}` and `src/test/${shims.minorVerSrc}` as test sources. The `spark-4.2` profile sets `shims.minorVerSrc` to `spark-4.2` (`pom.xml:747`), and there is no `spark/src/test/spark-4.2` directory, so everything under `spark/src/test/spark-4.1/` is invisible to the 4.2 build. The whole motivation for this fix is the per-expression eval mode introduced in 4.1 and carried forward, so it would be good for the test to cover the newest supported Spark too.

Would you be up for adding `src/test/${shims.minorPlusVerSrc}` to the `add-test-source` list and moving `CometDecimalArithmeticViewSuite` to `spark/src/test/spark-4.1+/`? That picks up 4.2 and anything after it. The existing `issue #4124` test gets the same benefit.

**2. The `TRY` assertion in the new test is currently vacuous.**

The test runs under `SQLConf.ANSI_ENABLED -> "false"`. On the old code that produced `nullOnOverflow = true` for every wrapper, so `assert(!tryAddProto.getCheckOverflow.getFailOnError)` would have passed before the fix too. Only the outer `assert(ansiOverflow.getFailOnError)` actually catches the bug.

Could the test run the same expression tree under both `ANSI_ENABLED -> "true"` and `"false"` and assert the same two flags each time? That is really the property being fixed, that the session setting no longer participates at all, and it makes both assertions load-bearing in one direction or the other.

**3. Only `Add` is covered, but the change touches five operators.**

Would it be worth looping the new test over `Add`, `Subtract`, `Multiply`, `Divide` and `Remainder`? `Divide` is the interesting one, since `CometDivide.convert` emits its own `CheckOverflow` and the proto shape is a nested pair rather than a single wrapper. A regression that changed only the `div` case would slip past the test as written.

**4. Please document why the flag comes from the expression.**

The scaladoc on `DecimalPrecision` goes into real detail on why the target type must come from `expr.dataType` rather than `SQLConf`. #5075 makes the point that the old code reads as if the session flag were the right source. A sentence in that same comment saying `nullOnOverflow` follows each node's captured `evalMode`, because a `TRY` operation can be nested inside an ANSI expression, would stop the same thing being reintroduced later.

**5. The `Divide` wrapper is now provably redundant, and I would like a tracking issue for it.**

After this change `promote` wraps a decimal `Divide` in `CheckOverflow(div, div.dataType, div.evalMode != ANSI)`, which serializes to `failOnError = evalMode == ANSI` with datatype `div.dataType`. `CometDivide.convert` (`arithmetic.scala:281-288`) then emits an inner `CheckOverflow` with exactly the same `failOnError` and the same datatype around the same node. The native dedup at `planner.rs:523` only recognizes a `WideDecimalBinaryExpr` child, and the fusion at `planner.rs:530` only recognizes a `Cast` child, so neither collapses a nested `CheckOverflow`. Every decimal division therefore runs an identical precision check twice per batch.

Before this PR the two flags could in principle disagree, so the duplication was arguably defensive. Now it is not. Cleaning it up is separate work, so could you open an issue and link it here rather than leaving it unrecorded?

peterxcli:
@andygrove Thanks for the review!

> 1. The new regression test does not run on Spark 4.2.

- moved `CometDecimalArithmeticViewSuite` to `spark/src/test/spark-4.1+/`
- added `src/test/${shims.minorPlusVerSrc}` into `pom.xml`

> 2. The TRY assertion in the new test is currently vacuous.

added scala test

> 3. Only Add is covered, but the change touches five operators.

added Subtract, Multiply, Divide and Remainder scala tests.

> 4. Please document why the flag comes from the expression.

added comment.

> 5. The Divide wrapper is now provably redundant, and I would like a tracking issue for it.

filed https://github.com/apache/datafusion-comet/issues/5190

andygrove (approved):
LGTM. Thanks @peterxcli

peterxcli:

PR #5173, opened 2026-07-31
chore: drop deprecated .toIterable from protobuf builder calls

andygrove:

No issue filed — mechanical compiler-warning cleanup, split out of #5141. Follows #5168 and #5170.

## Rationale for this change

Scala 2.13 deprecates `Iterable.toIterable` (it is being made protected). Comet calls it at 12 sites, all of them `xs.toIterable.asJava` when handing a collection to a protobuf `addAll*` builder — **12 of the 127 warnings** under the 2.13 profiles.

The interesting part is what *not* to do. At all 12 sites the receiver is an `Array`, so `.toIterable` resolves through `Predef.wrapRefArray` (the non-deprecated `Array` → `mutable.ArraySeq` wrap) and then `Iterable.toIterable` returns `this`. **The call is already free**, and the obvious fix of `.toSeq` would *add* a copy on 2.13, since `ArrayOps.toSeq` is `toIndexedSeq`. `.asJava` cannot be called on an `Array` directly either — that would need two chained implicit conversions.

So rather than convert the arrays, this PR removes them. Every one of these `addAll*` calls then takes `.asJava` on a `Seq`, which is what the neighbouring `addAll*` calls in the same methods already do (e.g. `addAllFields(scanTypes.asJava)`, `addAllPartitionByList(partitionExprs.map(_.get).asJava)`).

- **`schema2Proto`** takes and returns a `Seq[StructField]` / `Seq[SparkStructField]`. All five callers already hold a `StructType`, which *is* a `Seq[StructField]`, so they pass it straight in instead of unwrapping `.fields`; `CometNativeShuffleWriter` drops a `.toArray` on the way in. This clears 6 of the 12 warnings on its own (`CometNativeScan` ×3, `CometCsvNativeScanExec` ×2, and it is the same array that `addAllDataSchema` and friends were wrapping).
- **`QueryPlanSerde.serializeDataType`** maps over the `StructType` instead of `s.fields` for the struct field names/datatypes/nullability. That also drops a `.toSeq` (a real copy) on `fieldDatatypes`.
- **Projection vectors** are built from the `StructType` in both scan serdes, and in `CometNativeScan` from a `Range` rather than `Array.range` — same values, no array.
- **`CometNativeScan` default values**: the `zipWithIndex`/`filter`/`map`/`unzip` chain runs on an iterator and materializes once, so the three intermediate arrays are gone.
- **`CometWindowExec.winExprs`** keeps the `Seq` that `op.windowExpression.map` returns instead of `.toArray`-ing it; it is only used for `.length` and `.map`.

Net effect: 12 deprecation warnings gone, several array copies removed from the per-plan serde path, and no conversion added anywhere.

No new tests — no intended behavior change; the protobuf messages built are identical.

- Warnings under the default profile (Spark 4.1 / Scala 2.13): **127 → 115**, all 12 of these gone and none added, verified by diffing the full sorted warning list before and after.
- Scala 2.12 (`-Pspark-3.5`): warning list **byte-identical** (15 → 15).
- 333 tests pass across the suites covering every touched serde path: `CometExecSuite` and `CometWindowExecSuite` (window serde), `CometCsvNativeReadSuite` and `CometCsvExpressionSuite` (`CometCsvNativeScanExec`, including its projection vector and both schema protos), `CometNativeShuffleSuite` (`schema2Proto` via `addExpectedOutputSchema`), `ParquetReadV1Suite` and `CometNativeReaderSuite` (`CometNativeScan` schemas, projection vector and default values, plus `serializeDataType` on nested structs).

### DIFF
--- a/spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
@@ -626,9 +626,9 @@ object QueryPlanSerde extends Logging with CometExprShim with CometTypeShim {
         val info = DataTypeInfo.newBuilder()
         val struct = StructInfo.newBuilder()
 
-        val fieldNames = s.fields.map(_.name).toIterable.asJava
-        val fieldDatatypes = s.fields.map(f => serializeDataType(f.dataType)).toSeq
-        val fieldNullable = s.fields.map(f => Boolean.box(f.nullable)).toIterable.asJava
+        val fieldNames = s.map(_.name).asJava
+        val fieldDatatypes = s.map(f => serializeDataType(f.dataType))
+        val fieldNullable = s.map(f => Boolean.box(f.nullable)).asJava
 
         if (fieldDatatypes.exists(_.isEmpty)) {
           return None
--- a/spark/src/main/scala/org/apache/comet/serde/operator/CometNativeScan.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/operator/CometNativeScan.scala
@@ -143,17 +143,18 @@ object CometNativeScan extends CometOperatorSerde[CometScanExec] with Logging {
         // Our schema has default values. Serialize two lists, one with the default values
         // and another with the indexes in the schema so the native side can map missing
         // columns to these default values.
-        val (defaultValues, indexes) = possibleDefaultValues.zipWithIndex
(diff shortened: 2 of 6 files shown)


hsiang-c (approved):
Thanks @andygrove LGTM

mbutrovich (approved):
LGTM, thanks @andygrove!

PR #5177, opened 2026-07-31
fix: prevent silent overflow when reading Parquet TIMESTAMP_MILLIS values

### LINKED ISSUE #5090: Use arrow cast for hand-rolled temporal unit conversion loops

Several hand-rolled per-row loops implement temporal unit conversions that arrow's `cast` kernel performs bit-identically (verified against the arrow-cast 58.4.0 source):

- `native/core/src/parquet/cast_column.rs` (`cast_timestamp_micros_to_millis_array` / `_scalar`): `unary(|v| v / 1000)` plus manual timezone re-attachment. Arrow's Timestamp-to-Timestamp cast performs the same truncating division and attaches the target tz.
- `native/spark-expr/src/conversion_funcs/temporal.rs` (NTZ arm of `cast_date_to_timestamp`): per-row `(d as i64) * 86_400 * 1_000_000` into a builder. Arrow's `(Date32, Timestamp(Microsecond, _))` cast computes exactly `(x as i64) * MICROSECONDS_IN_DAY`.
- `native/spark-expr/src/utils.rs` (millis-to-micros): `arity::unary(|v| v * 1000)`. Arrow's timestamp-unit cast performs the same multiply.
- `native/spark-expr/src/datetime_funcs/date_from_unix_date.rs`: manual Int32-to-Date32 reinterpretation plus a scalar match arm. Arrow dispatches `(Int32, Date32)` to a zero-copy reinterpret cast; the scalar path can use `ScalarValue::cast_to`.

Replace each loop with `arrow::compute::cast_with_options` to the target type.

### Additional context

Caveats to preserve:
- micros-to-millis: safe when source and target tz are both Some or both None. If the source is NTZ and the target has a tz, arrow wall-clock-adjusts values, so that combination must keep the current relabel behavior.
- Date32-to-Timestamp: only the `target_tz.is_none()` branch is replaceable. The session-tz branch implements Spark's DST ambiguity/gap rules (earlier occurrence, pre-transition offset), while arrow's adjustment returns null on both, so it must stay.

peterxcli:
## Rationale for this change

Spark's vectorized Parquet reader converts TIMESTAMP_MILLIS values with the checked `millisToMicros` (`Math.multiplyExact`), so overflow throws independently of ANSI mode. Comet's reader used an unchecked `v * 1000`, silently wrapping the value.

Spark can also discard filtered-out values before timestamp conversion using statistics, dictionary, and row-level filters. Comet cannot mirror every pruning path, so checked conversion must not run on values Spark would never read.

- Add a checked top-level `Timestamp(Millisecond) -> Timestamp(Microsecond)` conversion in `parquet_convert_array`, the path the native Parquet scan actually takes. Unfiltered overflow now raises an error instead of silently wrapping, matching Spark in every eval mode.
- Retain the existing safe cast for filtered scans. This preserves Spark-successful queries when pruning differs, including dictionary-only exclusion, timestamp `IN` lists above DataFusion's 20-value pruning cutoff, nested predicates, and schema-evolution casts on unrelated filter columns.
- Remove the predicate-domain rewrite. The filtered-scan fallback makes it unnecessary and avoids changing empty `NOT IN` NULL semantics.
- Keep nested timestamp fields on the existing safe conversion because DataFusion cannot prune nested-field predicates before conversion.
- Remove the unreachable millis-to-micros arm from `array_with_timezone`: Spark logical timestamps are always microseconds, and Parquet conversion now happens on the actual reader path.
- Reject a microsecond physical timestamp with a top-level millisecond Spark logical target during planning (`CometCastColumnExpr::try_new`), since Spark read schemas represent logical timestamps in microseconds.
- Use Arrow's zero-copy cast for `Int32 -> Date32` in `date_from_unix_date`, with consistent `DEFAULT_CAST_OPTIONS` for scalar and array inputs.

Follow-up #5517 tracks surfacing overflow as Spark's `ArithmeticException("long overflow")` instead of a raw Arrow compute error. Follow-up #5553 tracks checked nested-field conversion once DataFusion supports nested pruning.

- `make core`
- `DYLD_LIBRARY_PATH=$JAVA_HOME/lib/server cargo test -p datafusion-comet test_millis_to_micros_overflow_checked_only_at_top_level --lib` — 1 passed
- `./mvnw test -Dtest=none -Dsuites="org.apache.comet.parquet.ParquetReadV1Suite filtered TIMESTAMP_MILLIS" -Dscalastyle.skip=true` — 2 tests passed in 1 suite
- `./mvnw test -Dtest=none -Dsuites="org.apache.comet.parquet.ParquetReadV1Suite TIMESTAMP_MILLIS overflow fails in native scan" -Dscalastyle.skip=true` — 1 test passed in 1 suite
- `./mvnw spotless:apply`

The Spark regressions cover direct and dictionary pages, positive/negative overflow, ANSI on/off, row-filter pushdown on/off, the 20/21-item `IN` boundary, dictionary-only exclusion, widened filter columns, nested predicates, and empty `NOT IN` with a NULL input.


andygrove (commented):
Thanks for working on this. Replacing the hand-written loops with Arrow kernels is a good direction, and I especially like that you spotted Arrow's timezone adjustment in `cast_with_options` and worked around it with a metadata-only relabel. The comment explaining why the relabel has to happen first is really helpful, and the new `+07:00` test cases genuinely guard it, since without the relabel they would be off by seven hours.

I have a few things I would like to work through before this goes in.

## 1. Micros to millis: Arrow truncates, Spark floors

`native/core/src/parquet/cast_column.rs`

Arrow's timestamp downscale is plain integer division (`time_array.unary(|o| o / divisor)` in `arrow-cast`), so it truncates toward zero. Spark's `SparkDateTimeUtils.microsToMillis` is `Math.floorDiv(micros, MICROS_PER_MILLIS)`, and there is a comment there specifically about pre-1970 timestamps needing that adjustment. So for `-1_500_001` micros Spark produces `-1501` and we produce `-1500`.

The old `unary(|v| v / 1000)` had the same behavior, so this is not something the PR introduces. My concern is that the new test asserts `-1_500_001 -> -1500`, which bakes the divergence into a test as though it were intended. Moving to Arrow's cast also takes away the easy fix, since the hand-written closure could have simply become `v.div_euclid(1000)`.

Would it make sense to keep a small kernel for this one case so the floor semantics can be matched? If you would rather keep the Arrow cast, could we file an issue and reference it next to the negative assertion so the expected value does not read as deliberate?

## 2. `CastOptions::default()` is `safe: true`

`native/spark-expr/src/utils.rs`

Could the millis to micros cast use `safe: false` instead of `CastOptions::default()`? Arrow branches on that flag for the upscale path. With `safe: true` it takes `unary_opt(|o| o.checked_mul(mul))`, which allocates a fresh null buffer and checks every element. With `safe: false` it takes `try_unary`, which reuses the input null buffer. So the default is doing an extra pass per batch compared to the `unary` this replaces.

There is a behavior argument too. Spark's `millisToMicros` is `Math.multiplyExact` and throws on overflow, so raising an error is closer to Spark than silently producing NULL. It would also line up with DataFusion's `DEFAULT_CAST_OPTIONS`, which is what `cast_column.rs` uses in this same PR.

## 3. Inconsistent cast options in `date_from_unix_date`

`native/spark-expr/src/datetime_funcs/date_from_unix_date.rs`

The two branches end up with different options. The array path gets `CastOptions::default()`, which is `safe: true`, while `scalar.cast_to(...)` resolves to `cast_to_with_options(target, &DEFAULT_CAST_OPTIONS)`, which is `safe: false`. `Int32 -> Date32` goes through `cast_reinterpret_arrays`, so it cannot fail either way today. But if the `Signature::exact(vec![Int32])` is ever widened, the two paths would quietly disagree, one nulling and one erroring. Worth making them match while it is cheap to do?

For what it is worth, I checked the two things in this file that looked riskiest and both are fine. `Int32 -> Date32` stays zero-copy, so there is no regression versus the manual `Date32Array::new`. And dropping the explicit `ScalarValue::Null` arm is safe, because `can_cast_types` has `(Null, _) => true` and the cast returns `new_null_array`.

## 4. Test coverage in `cast_column.rs`

Both `evaluate` tests moved from `Timestamp(ms, None)` to `Timestamp(ms, Some("+07:00"))`, and the three deleted unit tests were the ones covering `target_tz = None`. I think that leaves the no-timezone case uncovered, which is the branch where `relabel_array` early-returns because the types already match. Could one of these keep a `None` target?

It would also be good to have a case where the input array already carries a timezone, say `Timestamp(us, Some("UTC"))` to `Timestamp(ms, Some("America/New_York"))`. `relabel_array` overwrites whatever timezone the input has, and "relabel, do not shift" is exactly the property the workaround is protecting, so having that pinned down would help.

## A note, not a request

In `cast_date_to_timestamp`, Arrow's `Date32 -> Timestamp(us)` is a plain `unary(|x| (x as i64) * MICROSECONDS_IN_DAY)` that ignores the safe flag, so a large `Date32` wraps silently. Spark's `daysToMicros` uses `Math.multiplyExact`. That is identical to the old `(d as i64) * 86_400 * 1_000_000`, so I am not asking for anything here. I only wanted to note it so it is not mistaken for something the Arrow cast fixed.

peterxcli:

> 1. Micros → millis rounding

Changed this one conversion back to a small custom kernel because Arrow truncates negative values toward zero, while Spark uses floor division.

- Before: `-1_500_001 / 1_000 = -1500`
- Now: `-1_500_001.div_euclid(1_000) = -1501`

> 2. Millis → micros cast options for `array_with_timezone` in `utils.rs`

Changed to use `DEFAULT_CAST_OPTIONS`.

I also added a regression assertion using `i64::MAX` to confirm overflow returns an error from spark: https://github.com/apache/spark/blob/v4.2.0/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/util/DateTimeUtilsSuite.scala#L969-L972

> 3. `date_from_unix_date` option consistency

Changed to use `DEFAULT_CAST_OPTIONS`.

> 4. `cast_column.rs` test coverage

Expanded the array test to cover all three relevant timezone layouts:

- No timezone → no timezone
- No timezone → `+07:00`
- `UTC` → `America/New_York`

andygrove (commented):
Thanks for the quick turnaround. All four points from the last round look addressed, and pinning the new test values to Spark's own `DateTimeUtilsSuite` example is a nice touch. I ran the touched tests locally and they pass, and I re-checked the claims against arrow-cast 58.4.0 and Spark master.

A couple of things I confirmed so nobody has to re-derive them. `SparkDateTimeUtils.microsToMillis` really is `Math.floorDiv`, with a comment about pre-1970 timestamps, so `div_euclid` is the right call. `millisToMicros` really is `Math.multiplyExact`, so `safe: false` in `utils.rs` is the Spark-faithful choice. And there is no performance regression anywhere: the millis-to-micros upscale reinterprets to `Int64` zero-copy and then uses `try_unary`, which reuses the input null buffer, so it is the same single allocation the old `unary` did. `Date32 -> Timestamp(us)` is one `unary` plus a same-type cast that early-returns, and `Int32 -> Date32` is still `cast_reinterpret_arrays`. I also grepped for other micros-to-millis sites that might share the truncation bug, and the only ones are the two this PR fixes.

One thing on packaging that I would like to sort out before this merges. The `div_euclid` change is a real behavior fix rather than a refactor, since any pre-1970 timestamp with a sub-second component now yields a different value than before. Our changelog is generated from PR titles, so as `refactor:` this lands with no signal that timestamp results changed. Could we retitle to `fix:`, or split the floor fix out so it gets its own entry? Either way it would help to have the divergence tracked in an issue we can link from the code comment.

The description needs a refresh too. The first bullet still says we use Arrow casts for the Parquet micros-to-millis conversion, and that is the one case that ended up keeping a hand-written kernel. It would also be good to mention that millis-to-micros overflow went from silently wrapping to raising an error, since that is user visible as well.

Nothing needed on docs. No serde or expression registration changed, so the compatibility pages and `expressions.md` stay as they are. `date_from_unix_date` already has good SQL test coverage including the Spark min and max date boundaries, and both of its branches are now consistent, so I have nothing to raise there.

The rest of my comments are inline.

native/core/src/parquet/cast_column.rs
```diff
@@ -279,12 +245,24 @@ impl PhysicalExpr for CometCastColumnExpr {
...
-                    let casted = cast_timestamp_micros_to_millis_array(&array, target_tz.clone());
-                    Ok(ColumnarValue::Array(casted))
+                    let micros = array
+                        .as_any()
+                        .downcast_ref::<TimestampMicrosecondArray>()
+                        .expect("Expected TimestampMicrosecondArray");
+                    // Spark floors when downscaling negative timestamps; Arrow truncates.
+                    // [SparkDateTimeUtils.scala](https://github.com/apache/spark/blob/v4.2.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L92-L101)
+                    let millis: TimestampMillisecondArray =
+                        arity::unary(micros, |v| v.div_euclid(1_000));
```
andygrove: Do you know whether this arm is reachable from a Spark or Iceberg query? I traced it back to the DataFusion 52 Iceberg migration and could not find anything on the JVM side that asks for a `Timestamp(Millisecond)` logical field, so I could not tell whether the floor fix here is observable end to end.

If it is reachable, a Spark-level test reading a pre-1970 sub-second timestamp would be worth having, since that is the case that was wrong before. If it is only defensive for a schema shape we do not currently generate, a comment saying that would be just as useful.

native/core/src/parquet/cast_column.rs
```diff
@@ -476,15 +383,11 @@ mod tests {
         ));
         let schema = Schema::new(vec![Arc::clone(&input_field)]);
 
-        // Create target field with TimestampMillisecond
-        let target_field = Arc::new(Field::new(
-            "ts",
-            DataType::Timestamp(TimeUnit::Millisecond, None),
-            true,
-        ));
+        let target_type = DataType::Timestamp(TimeUnit::Millisecond, None);
```
andygrove: The array test above now loops over three timezone layouts, which is a good improvement. This scalar test is still only `target_tz = None`, and `test_cast_timestamp_micros_to_millis_scalar`, which got deleted, was the one checking that a target timezone lands on the scalar.

Could this test loop over the same timezone cases as the array one, so we pin down that the resulting `ScalarValue::TimestampMillisecond` actually carries `target_tz`?

native/spark-expr/src/conversion_funcs/temporal.rs
```diff
@@ -39,49 +40,45 @@ pub(crate) fn cast_date_to_timestamp(
     cast_options: &SparkCastOptions,
     target_tz: &Option<Arc<str>>,
 ) -> SparkResult<ArrayRef> {
+    if target_tz.is_none() {
+        return Ok(cast_with_options(
+            array_ref,
+            &DataType::Timestamp(TimeUnit::Microsecond, None),
+            &CastOptions::default(),
```
andygrove: This one is still `CastOptions::default()` while the other two sites moved to `DEFAULT_CAST_OPTIONS`. Could we make it three for three?

I know it makes no behavioral difference today, since Arrow's `Date32 -> Timestamp(us)` arm is a plain `unary` that ignores the safe flag. But leaving one call site on `safe: true` reads like it was a deliberate choice, and if Arrow ever adds the overflow check then `safe: false` is what we want anyway, because Spark's `daysToMicros` goes through `Math.multiplyExact`.

native/spark-expr/src/utils.rs
```diff
@@ -81,12 +82,8 @@ pub fn array_with_timezone(
...
                     timestamp_ntz_to_timestamp(array, timezone.as_str(), Some(target_tz.as_ref()))
                 }
-                Some(DataType::Timestamp(TimeUnit::Microsecond, None)) => {
-                    // Convert from Timestamp(Millisecond, None) to Timestamp(Microsecond, None)
-                    let millis_array = as_primitive_array::<TimestampMillisecondType>(&array);
-                    let micros_array: TimestampMicrosecondArray =
-                        arrow::compute::kernels::arity::unary(millis_array, |v| v * 1000);
-                    Ok(Arc::new(micros_array))
+                Some(to_type @ DataType::Timestamp(TimeUnit::Microsecond, None)) => {
+                    cast_with_options(array.as_ref(), to_type, &DEFAULT_CAST_OPTIONS)
```
andygrove: I want to make sure I understand what the new `is_err()` assertion in the test is locking in. Over in `cast.rs`, `array_with_timezone` is called on line 262, one line before `native_cast_options` is built with `safe: !matches!(eval_mode, EvalMode::Ansi)`. So this cast now hard errors on overflow even under `try_cast` or legacy mode, where the rest of `cast_array` would produce NULL.

My guess is that this is fine, because the arm looks like it serves Iceberg schema adaptation rather than a user-written cast, and Spark's Parquet reader calls `millisToMicros` unconditionally and throws regardless of ANSI. If that is right, could we add a short comment saying so? It would stop someone from later "fixing" this to respect eval mode.

peterxcli:
@andygrove thanks for the review! review change is pushed. please take another look, thanks!


> 1. `div_euclid` is a behavior fix, not a refactor.

Retitled the PR. The final patch rejects that invalid read-schema pair instead of using `div_euclid`.

> 2. Refresh the PR description and mention overflow.

Updated it to describe the planning rejection and that millis→micros overflow now errors instead of wrapping.

> 3. Is micros→millis reachable from Spark or Iceberg?

No. Spark read schemas use microsecond logical timestamps. Construction now returns `DataFusionError::Plan`, with a Spark source link.

> 4. Keep scalar target-timezone coverage.

The pair is now rejected before scalar or array evaluation. The planning test covers timezone-free and timezone-bearing fields.

> 5. Use `DEFAULT_CAST_OPTIONS` at the third site.

Done. The timezone-free `Date32 -> Timestamp` path now uses `DEFAULT_CAST_OPTIONS`.

> 6. Explain why millis→micros always errors on overflow.

Added links to Spark’s [Parquet call site](https://github.com/apache/spark/blob/v4.2.0/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/ParquetVectorUpdaterFactory.java#L817-L833) and checked [`millisToMicros`](https://github.com/apache/spark/blob/v4.2.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L103-L108) implementation.

native/spark-expr/src/utils.rs
```diff
@@ -81,12 +82,14 @@ pub fn array_with_timezone(
...
-                        arrow::compute::kernels::arity::unary(millis_array, |v| v * 1000);
-                    Ok(Arc::new(micros_array))
+                Some(to_type @ DataType::Timestamp(TimeUnit::Microsecond, None)) => {
+                    // This defensive conversion intentionally errors in every CAST eval mode:
+                    // Spark's vectorized Parquet reader calls `millisToMicros` for both direct
+                    // and dictionary values, independent of CAST evaluation.
+                    // https://github.com/apache/spark/blob/v4.2.0/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/ParquetVectorUpdaterFactory.java#L817-L833
+                    // `millisToMicros` uses `Math.multiplyExact`:
+                    // https://github.com/apache/spark/blob/v4.2.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L103-L108
+                    cast_with_options(array.as_ref(), to_type, &DEFAULT_CAST_OPTIONS)
```
andygrove: I went to confirm the overflow fix end to end and I do not think it reaches the Parquet reader. I wrote a scratch test driving `CometCastColumnExpr::evaluate` with an `i64::MAX` millisecond value on this branch, and it returns `null` rather than erroring, for both `Timestamp(ms, None)` and `Timestamp(ms, Some("UTC"))`.

The reason is that the scan never goes through `array_with_timezone`. `evaluate` falls through the relabel arm, because `types_differ_only_in_field_names` is `false` for a flat timestamp pair. It lands in `spark_parquet_convert`. `parquet_convert_array` has no timestamp-unit arm, so it reaches the `can_cast_types` fallthrough at `parquet_support.rs:257` and calls `cast_with_options` with `PARQUET_OPTIONS`. That static is `safe: true` at `parquet_support.rs:57`. I also grepped for callers, and `array_with_timezone` is only reached from `cast.rs:262` and the three datetime functions. Nothing under `native/core/src/parquet/` calls it.

On the Spark side this is a real divergence. `ParquetVectorUpdaterFactory:155-170` sends both `TimestampType` and `TimestampNTZType` with MILLIS to `LongAsMicrosUpdater`, which calls `millisToMicros`, which is `Math.multiplyExact`. So Spark throws regardless of ANSI mode, and we return null.

Would you be willing to add an explicit `(Timestamp(ms, _), Timestamp(us, _))` arm to `parquet_convert_array` that does the checked multiply? I would avoid flipping `PARQUET_OPTIONS` to `safe: false` wholesale, since that static also covers narrowing casts where returning null is what Spark does, so a targeted arm seems safer than changing the global.

If you would rather keep the scope as it is, could we retitle away from `fix:` and open an issue for the Parquet path? Our changelog is generated from PR titles, and as written this would tell users the Parquet overflow is fixed when it is not.

native/spark-expr/src/utils.rs
```diff
@@ -81,12 +82,14 @@ pub fn array_with_timezone(
...
                     timestamp_ntz_to_timestamp(array, timezone.as_str(), Some(target_tz.as_ref()))
                 }
-                Some(DataType::Timestamp(TimeUnit::Microsecond, None)) => {
-                    // Convert from Timestamp(Millisecond, None) to Timestamp(Microsecond, None)
-                    let millis_array = as_primitive_array::<TimestampMillisecondType>(&array);
-                    let micros_array: TimestampMicrosecondArray =
-                        arrow::compute::kernels::arity::unary(millis_array, |v| v * 1000);
-                    Ok(Arc::new(micros_array))
+                Some(to_type @ DataType::Timestamp(TimeUnit::Microsecond, None)) => {
+                    // This defensive conversion intentionally errors in every CAST eval mode:
```
andygrove: Related to my other comment, do you know whether this arm is reachable today?

Spark logical timestamps are always microseconds, per `to_arrow_datatype` at `serde.rs:93-96` and `Utils.scala:157-160`. The only millisecond arrays I can find come from the Parquet physical schema, and those are converted inside `spark_parquet_convert` rather than `cast_array`. So I could not find a path that gets here with a millisecond input and a `Timestamp(Microsecond, None)` target.

If it is dead, then the new test is pinning behavior nothing exercises, and a note saying so would help the next reader. If you can point me at a query that reaches it, I am happy to be wrong.

native/core/src/parquet/cast_column.rs
```diff
@@ -214,20 +176,41 @@ impl Hash for CometCastColumnExpr {
...
+    ) -> DataFusionResult<Self> {
+        let physical_type = physical_field.data_type();
+        let target_type = target_field.data_type();
+        // `target_field` is the Spark logical field, while `physical_field` comes from the
+        // Parquet or Iceberg file. Comet represents Spark's TimestampType and TimestampNTZType
+        // as Arrow microseconds, and Spark maps both TIMESTAMP_MICROS and TIMESTAMP_MILLIS files
+        // to those logical types. A millisecond target is therefore invalid at this read-adapter
+        // boundary:
+        // https://github.com/apache/spark/blob/v4.2.0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala#L318-L324
+        if matches!(
```
andygrove, line 194: The rejection reads well, and I confirmed the claim holds. Both the Parquet scan and `IcebergScanExec` build their logical schema through `convert_spark_types_to_arrow_schema`, so a millisecond target cannot occur.

One scoping question. The check looks at the top-level pair only, so a target like `Struct<ts: Timestamp(ms)>` passes `try_new` and goes on to `spark_parquet_convert`. Since the invariant you are asserting is about the whole read schema rather than just flat columns, the guard currently reads more complete than it is. Could we either walk nested types or narrow the comment to say it covers the top level? Either is fine. I mostly want the next person not to trust it further than it goes.

andygrove (commented):
Thanks for the rework. Rejecting the pair outright is a better answer than either version of the division fix, and the `ParquetSchemaConverter` link makes the invariant checkable. CI is fully green, including Iceberg 1.11, which is the PR-gated Iceberg job, so the scan paths that matter here did get exercised.

A few things I confirmed independently so you do not have to re-derive them.

Your reachability claim is correct. `to_arrow_datatype` at `native/core/src/execution/serde.rs:93-96` maps Spark timestamps to microseconds only, and the JVM side agrees at `Utils.scala:157-160`. `IcebergScanExec` builds its schema from the same `convert_spark_types_to_arrow_schema` at `planner.rs:1651`. A millisecond logical target genuinely cannot occur.

Deleting the three tests costs no coverage. `types_differ_only_in_field_names` returns `false` for a flat `(Timestamp, Timestamp)` pair, so `relabel_array` was never on the timestamp path, and its List, Map, and Struct coverage is intact.

`temporal.rs` and `date_from_unix_date.rs` are now consistent on `DEFAULT_CAST_OPTIONS`, and both look right to me.

Nothing needed on docs. No serde or expression registration changed, so the compatibility pages and `expressions.md` stay as they are.

The one thing I would like to sort out before this merges is whether the overflow fix actually reaches the Parquet reader. I do not think it does. Details inline.

native/spark-expr/src/utils.rs
peterxcli: Good catch. I moved the checked millis-to-micros conversion into `parquet_convert_array`, which is the actual Parquet reader path. It now uses `try_unary` with `mul_checked(1_000)`, so overflow returns an error instead of `null`.

I also removed the unreachable conversion from `array_with_timezone` and added an end-to-end native Parquet scan test covering TimestampType and TimestampNTZType, positive and negative overflow, dictionary on/off, and ANSI on/off.

Since the Parquet path is now fixed, I kept the `fix:` title and will update the PR description accordingly. No follow-up issue should be necessary.

peterxcli:
@andygrove thanks for another round of review, addressed all of your review. please take another look. TIA!

> The overflow fix may not reach the actual Parquet reader.

Moved the checked millis -> micros conversion into `parquet_convert_array`, the real Parquet reader path. It now uses `try_unary` with `mul_checked(1_000)`, so overflow returns an error instead of `null`.

> Please verify the fix end to end against Spark behavior.

Added a native Parquet scan regression covering TimestampType and TimestampNTZType, positive and negative overflow, dictionary on/off, and ANSI on/off. It verifies both Spark and Comet report overflow.

> Is the millis→micros arm in `array_with_timezone` reachable?

Confirmed it is unreachable for Parquet reads. Removed the dead arm, its imports, and its unit test. The conversion is now tested where it actually runs.

> The millisecond-target rejection only checks top-level timestamps.

Narrowed the comment to explicitly state that the guard applies to top-level timestamp columns, so it does not imply nested timestamp validation.

andygrove:

Turning the silent `v * 1000` wrap into a checked multiply is the important part here, and linking to `ParquetVectorUpdaterFactory` and `SparkDateTimeUtils.millisToMicros` in the comment makes it easy to verify against Spark. Deleting the duplicate `cast_timestamp_micros_to_millis_*` helpers is a good cleanup.

Several things.

**This conflicts directly with #5457**

#5457 rewrites `cast_date_to_timestamp` in `temporal.rs` as well, replacing the chrono region-zone path with fixed-offset arithmetic and a plan-time fallback. This PR keeps that path and adds an Arrow fast path for the NTZ case. They cannot both land as written. Worth coordinating with @sunchao on which goes first.

**`date_from_unix_date` loses a zero-copy path**

The old code was:

```rust
Date32Array::new(int_array.values().clone(), int_array.nulls().cloned())
```

which is O(1): cloning an Arrow `Buffer` is a refcount bump. The new code calls `cast_with_options(arr, &DataType::Date32, ...)`. Depending on the Arrow version, `Int32 -> Date32` may or may not take a reinterpret fast path. If it does not, this turns a free operation into a full array copy on every batch.

Could you confirm which it is? If Arrow copies, the old code was better and I would keep it. The general principle of delegating to Arrow is right, but not when the hand-written version is asymptotically cheaper.

Separately, the new version accepts any type Arrow can cast to `Date32`, where the old one rejected anything that was not `Int32`. Spark's `DateFromUnixDate` only accepts `IntegerType`, so this should not matter, but it does remove a guard against a planner bug. Worth a comment saying the input type is guaranteed by the serde.

**Removing the millis-to-micros arm from `array_with_timezone`**

That arm is gone, so the same conversion in `array_with_timezone` now falls to `Err("Not supported")`. `array_with_timezone` is called from `cast_array`, not just the Parquet reader. Is there any path where a Comet `CAST` sees a `Timestamp(Millisecond, None)` input, for example from an Iceberg scan or a UDF return value? If there is, this changes a working conversion into an error. If there is not, it would be good to say so in the commit message, since the deletion looks unrelated to the Parquet fix at first glance.

**What does the new overflow error look like to a user?**

`try_unary(... mul_checked ...)` produces an `ArrowError`, which as far as I know surfaces as `CometNativeException` rather than the `ArithmeticException("long overflow")` Spark's `Math.multiplyExact` throws. #5169 is doing exactly this kind of error-fidelity work for the decimal paths. Is it worth making this one a typed `SparkError` from the start, rather than adding it to the list of raw Arrow errors that need converting later?

**`try_new` returning a plan error**

The new microsecond-physical to millisecond-target check returns `DataFusionError::Plan` from inside the schema adapter. If a file ever does present that shape, the query fails rather than falling back to Spark. The comment argues Spark's schema converter makes it unreachable, and the Spark link supports that. Could the message say what a user should do if they somehow hit it, or would it be better to fall back rather than fail?

peterxcli:
@andygrove thanks for the review! Addressed three points and pushing back on two, details below.

> **This conflicts directly with #5457**

Good catch. Since #5457 rewrites `cast_date_to_timestamp` as a safety fix and my change there was behavior-neutral (Arrow's NTZ arm compiles to the same unchecked `days * MICROS_PER_DAY`), I dropped the `temporal.rs` hunk from this PR entirely. #5457 can own that function; this PR now touches only the Parquet overflow paths, which also fits the `fix:` title better.

> **`date_from_unix_date` loses a zero-copy path**

Confirmed on arrow-cast 58.4.0: `(Int32, Date32)` takes `cast_reinterpret_arrays::<Int32Type, Date32Type>` ([mod.rs#L1631](https://github.com/apache/arrow-rs/blob/58.4.0/arrow-cast/src/cast/mod.rs#L1631)), which is a zero-copy buffer reinterpret — same cost as the old `Date32Array::new`, no regression. (Also matches your July 31 review note.) Added a comment documenting that the input is guaranteed `Int32` by `Signature::exact` and the serde.

> **Removing the millis-to-micros arm from `array_with_timezone`**

This was settled in the previous round — the arm was removed at your suggestion after we verified it unreachable: Spark logical timestamps are exclusively microseconds (`serde.rs:95-98`, `Utils.scala:157-160`), and both Iceberg scans and UDF return types flow through the same schema conversion, so `cast_array` can never see a millisecond input. I updated the PR description to note the rationale so the deletion doesn't look unrelated in the squashed commit.

> **What does the new overflow error look like to a user?**

I'd rather not convert it in this PR. Spark's exception here is an *untyped* `java.lang.ArithmeticException("long overflow")` from `Math.multiplyExact` — there is no Spark error class to map to — and reusing `SparkError::ArithmeticOverflow` would emit "set spark.sql.ansi.enabled to false to bypass this error", which is actively misleading since this path throws regardless of ANSI (the new test pins that with ANSI both on and off). Proper fidelity needs a new native variant plus `ShimSparkErrorConverter` mappings per Spark version, which belongs with the #5169 error-fidelity effort. Filed #5517 to track it.

> **`try_new` returning a plan error**

Fallback isn't reachable from there — by the time the native schema adapter runs, the JVM has already committed to the native scan. I extended the message to state it indicates a Comet bug, link the issue tracker, and suggest `spark.comet.scan.enabled=false` as a workaround.

sunchao:
> Worth coordinating with @sunchao on which goes first.

Feel free to merge this first!

spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala
```diff
@@ -261,6 +261,56 @@ abstract class ParquetReadSuite extends CometTestBase {
...
+            |message root {
+            |  optional int64 ts(TIMESTAMP_MILLIS);
+            |  optional int64 ts_ntz(TIMESTAMP(MILLIS,false));
+            |}
+            |""".stripMargin)
+          val writer = createParquetWriter(schema, path, dictionaryEnabled)
+          val record = new SimpleGroup(schema)
+          record.add(0, millis)
+          record.add(1, millis)
+          writer.write(record)
```
sunchao: **[P3] Exercise dictionary encoding in the overflow regression**

With this single row, Parquet 1.16.0 writes PLAIN pages even when `dictionaryEnabled=true`, so the dictionary-enabled iteration does not cover a dictionary read. I reproduced this for both timestamp columns and both overflow signs. Could we write repeated rows and assert `getEncodingStats.hasDictionaryEncodedPages`? A temporary 16-row variant produced dictionary pages and passed this regression on Spark 4.1.3 with ANSI both on and off.

sunchao (approved):
Reviewed `afa3673` with five specialist agents; no blocking correctness findings. The inline dictionary fixture note does not block approval.

Local validation passed: 82 native Parquet tests, 643 native expression tests, and the Spark 4.1.3 overflow regression. A temporary 16-row variant also passed after verifying dictionary pages for both timestamp types and both signs. Native validation excluded HDFS; the full Spark suite was not run locally.

spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala
peterxcli: Good catch — confirmed: with one row parquet-mr's dictionary writer falls back to PLAIN because a one-entry dictionary page isn't smaller than the raw values. Changed the test to write 16 repeated rows and to assert `getEncodingStats.hasDictionaryEncodedPages == dictionaryEnabled` for every column chunk of the footer, so the dictionary leg can't silently degrade again. Verified locally: the suite passes with dictionary pages present when enabled and PLAIN when disabled, across both overflow signs and ANSI on/off. (c833b72a6)

native/core/src/parquet/parquet_support.rs
```diff
@@ -195,6 +195,21 @@ fn parquet_convert_array(
...
+        ) => {
+            // Spark's Parquet reader calls the checked `millisToMicros` conversion for both
+            // direct and dictionary values, independent of CAST evaluation mode:
+            // https://github.com/apache/spark/blob/v4.2.0/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/ParquetVectorUpdaterFactory.java#L817-L833
+            // `millisToMicros` uses `Math.multiplyExact`:
+            // https://github.com/apache/spark/blob/v4.2.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L103-L108
+            let micros = array
+                .as_primitive::<TimestampMillisecondType>()
+                .try_unary::<_, TimestampMicrosecondType, _>(|value| value.mul_checked(1_000))?
+                .with_timezone_opt(target_tz.clone());
```
sunchao, line 237: **[P2] Preserve timestamp pruning before checked conversion**

With a `TimestampType` column stored as `TIMESTAMP_MILLIS` and containing `9223372036854776` milliseconds, `WHERE ts < TIMESTAMP '1970-01-01 00:00:00'` returns no rows in Spark 4.1.3 and the base native build (`c067e4e`), but `5c8de42` throws `Overflow happened on: 9223372036854776 * 1000`. I reproduced this with plain/dictionary encoding, ANSI on/off, and Comet row-filter pushdown on/off: all eight Spark/base cases succeed and all eight head cases fail.

The existing `CometCastColumnExpr` prevents DataFusion from recognizing the timestamp statistics predicate, so Comet converts values from row groups Spark skips. This new error turns that pruning limitation into a query failure. Please preserve pruning before the checked conversion and add this filtered case as a regression test, while retaining overflow errors for values actually read.

sunchao (commented):
Re-reviewed `5c8de42` with five specialist agents. One additional regression is described inline; the earlier dictionary fixture issue is fixed.

Local validation: 82 native Parquet tests, 660 native expression tests, and the Spark 4.1.3 overflow regression passed. The additional filtered-read regression passes with base `c067e4e` and fails on this head. Native builds excluded HDFS; the full Spark suite was not rerun locally.

native/core/src/parquet/parquet_support.rs
peterxcli: Confirmed and fixed — thanks for the thorough repro. The predicate's column was wrapped in `CometCastColumnExpr`, which is opaque to DataFusion's pruning analyzer, so the row group Spark prunes from millisecond statistics was being read and converted. The fix rewrites predicate comparisons over a `TIMESTAMP_MILLIS` file column into the millisecond domain (exact integer rescaling of the literal, like Spark's `ParquetFilters` pushing predicates in the file's physical unit), plus `IS NULL`/`IS NOT NULL` unwrapping. Pruning works again, predicate evaluation never converts file values, and the scan output conversion stays checked, so values actually read still fail like `millisToMicros`.

Added your filtered case as a regression test over dictionary × ANSI × `rowFilterPushdown` (8 configs), and native unit tests pinning the rounding table for all six comparison operators, both operand orders, and negative/sub-millisecond literals.

One deliberate divergence to note: with row-filter pushdown on and a non-pruned row group, rows the filter discards are no longer converted, so Comet can succeed where Spark (which converts the whole row group during decode) throws — the benign direction of "errors only for values actually read." (d46519298)

sunchao: **[P2] Preserve pruning for IN, null-safe equality, and nested timestamp predicates**

Rechecked `0d78875`: the original `ts < epoch` reproducer now passes, but this issue remains for three supported predicate forms. With 32 repeated `9223372036854776` millisecond values in a `TIMESTAMP_MILLIS` column (and the same value in nested `s.ts`), these predicates return zero rows in Spark 4.1.3 and base `c067e4e`, while the PR head throws `Overflow happened on: 9223372036854776 * 1000`:

```sql
ts IN (TIMESTAMP '1970-01-01 00:00:00', TIMESTAMP '1970-01-02 00:00:00')
ts <=> TIMESTAMP '1970-01-01 00:00:00'
s.ts < TIMESTAMP '1970-01-01 00:00:00'
```

All 24 cases reproduce across plain/dictionary encoding, ANSI on/off, and row-filter pushdown on/off; dictionary encoding was verified from the footer. The eight ordinary `<` controls pass on all three builds.

The [new rewrite](https://github.com/apache/datafusion-comet/blob/0d78875ab6d9ec99d0a1426348c86e7da9208648/native/core/src/parquet/schema_adapter.rs#L596-L634) skips `InListExpr`, does not handle `IsNotDistinctFrom`, and only matches a direct `CometCastColumnExpr`, so the nested-field predicate also retains the opaque conversion. These row groups are still read and converted even though Spark prunes them.

Could we extend pruning to these forms before checked conversion and add regression coverage? This is the failing-query direction of the original issue, separate from the documented case where Comet skips an error on a row discarded by its filter.

Validation: independent native Parquet scans plus focused Spark 4.1.3/JDK17 comparisons. The base control reused a hash-verified `c067e4e` native library; production JVM/proto sources are identical. Other Spark versions were not run locally.

peterxcli: Verified all three and fixed. `IN` and `<=>` now rewrite into the millisecond domain (list elements rescale with IN's null semantics preserved; null-safe equality rescales or folds to a constant, and `<=> NULL` becomes `IS NULL`) — I confirmed DataFusion 54.1's PruningPredicate analyzes both `InListExpr` and `IsNotDistinctFrom`, so pruning covers them in every config.

The nested case turned out deeper than the rewrite: DataFusion can neither prune nested-field predicates ([pruning_predicate.rs](https://github.com/apache/datafusion/blob/54.1.0/datafusion/pruning/src/pruning_predicate.rs#L1131) — "PruningPredicate does not support pruning on nested fields yet") nor evaluate them as row filters (`can_expr_be_pushed_down_with_schemas` classifies struct columns non-pushable), and the failing conversion was actually the flat `ts` column being materialized from row groups nothing could prune. Since Spark only avoids the error via nested statistics pruning we don't have, scans whose data filters reference nested fields now fall back to the safe conversion (overflow → NULL, main's behavior) and the filter discards the rows — matching Spark's zero-row answers across your matrix. Checked conversion stays for all other scans, and is scoped to top-level columns for the same reason.

Extended the regression test to your three forms plus the flat control (4 predicates × dictionary × ANSI × rowFilterPushdown), and added native tests for the IN/null-safe rewrites and the top-level/nested/flag-off conversion split. Remaining divergence, documented in the code: a nested-predicate scan Spark *fails* to prune errors in Spark but NULLs in Comet, as does a direct nested read of overflow — both pre-existing behavior. Filed #5553 to lift both once DataFusion grows nested-field pruning. (917286124)

native/core/src/parquet/parquet_support.rs
```diff
@@ -195,6 +217,28 @@ fn parquet_convert_array(
...
+            //
+            // The checked conversion is limited to TOP-LEVEL columns. Spark only avoids the
+            // error for filtered-out values through row-group statistics pruning, and
+            // DataFusion's PruningPredicate does not support nested fields yet, so a checked
+            // conversion on a nested field would fail queries whose predicates Spark prunes
+            // (e.g. `WHERE s.ts < X` over an all-overflowing file). Nested fields keep the
+            // pre-existing safe-cast behavior below (overflow -> NULL).
+            let micros = array
+                .as_primitive::<TimestampMillisecondType>()
+                .try_unary::<_, TimestampMicrosecondType, _>(|value| value.mul_checked(1_000))?
```
sunchao, line 236: **[P2] Preserve dictionary pruning before checked timestamp conversion**

A row group can be excluded by Spark's dictionary filter even when its min/max statistics overlap the predicate. I reproduced this with one 32-row group containing `id: INT32` alternating between 1 and 3, and `ts: TIMESTAMP_MILLIS = 9223372036854776`. For `SELECT ts WHERE id = 2`, Spark 4.1.3 and base `215ab706` return zero rows, but this head throws `Overflow happened on: 9223372036854776 * 1000` with `spark.comet.parquet.rowFilterPushdown.enabled=false` (the default).

The footer verifies dictionary-only data pages and `id` min/max of 1/3; there is no schema widening. With plain encoding, both Spark and the head overflow, while enabling Comet row-filter pushdown makes the dictionary case succeed. This isolates a pruning gap beyond the timestamp-expression rewrite: checked conversion now runs on data Spark never decodes. Please preserve dictionary-based elimination, or use a fallback that preserves the successful query, before applying checked conversion to these scans; add the dictionary-only exclusion as a regression test.

native/core/src/parquet/schema_adapter.rs
```diff
@@ -522,7 +534,220 @@ impl PhysicalExprAdapter for SparkPhysicalExprAdapter {
...
+            return Ok(Transformed::yes(Arc::new(BinaryExpr::new(
+                inner,
+                op,
+                Arc::new(Literal::new(ScalarValue::TimestampMillisecond(
+                    Some(sentinel),
+                    file_tz,
+                ))),
+            ))));
+        }
+        let rewritten = in_list(
```
sunchao: **[P2] Preserve pruning for timestamp IN lists longer than 20 values**

Returning a raw `InListExpr` only restores statistics pruning up to DataFusion 54.1's `MAX_LIST_VALUE_SIZE_REWRITE = 20`. I reproduced the boundary with 32 repeated `9223372036854776` millisecond values and `ts IN (...)` containing 21 timestamps from `1970-01-01 00:00:00` through `00:00:20`: Spark 4.1.3 and base `215ab706` return zero rows, but this head throws checked timestamp overflow when row-filter pushdown is disabled (the default).

All four plain/dictionary × ANSI combinations fail on the head; the 20-value and 2-value controls pass, and enabling row-filter pushdown avoids the error. The rewritten list is still not statistics-prunable above the dependency's cutoff. Please provide a prunable representation or a compatible fallback for longer lists, and cover the 20/21 boundary with the projected overflowing timestamp.

sunchao (commented):
Re-reviewed `917286124cf39cd0c5f6eed5a1a06676e9679f28` against base `215ab706aa9af074e589f400cd6843dee9938030`. The checked conversion still makes some queries fail after Spark would prune the offending data, and the empty `NOT IN` rewrite can discard NULL rows. Four P2 findings are detailed inline: dictionary pruning, timestamp IN lists longer than 20 values, widened filter columns, and originally empty NOT IN.

The ordinary-comparison, short-IN, null-safe-equality, and nested-predicate cases discussed previously now pass all 64 comparison executions. The dictionary-overflow fixture also now verifies actual dictionary encoding.

Validation:

- Fresh native rebuild of the reviewed head, with HDFS disabled; independent review scopes covered predicate semantics, scan/planner integration, conversion/ownership/errors, Spark execution, and test coverage.
- Spark 4.1.3 / JDK 17: both PR-added TIMESTAMP_MILLIS tests passed. The additional 152-query matrix ran separately on the head and base: 11 mismatches on the head (8 overflow failures and 3 row-loss cases), none on the base.
- A separate 8-query dictionary-pruning matrix per revision confirmed another failure: Spark and the base return zero rows for a dictionary-excluded value, while the head overflows with row-filter pushdown disabled. Plain-encoding and row-filter controls were also checked.
- 26 conversion/compatibility component tests passed. Independent predicate checks also passed for ordinary comparisons, nonempty IN/NOT IN, timezone combinations, signed boundaries, and boolean composition; originally empty lists reproduced the NULL-semantic defect.
- Base comparisons reused an archived build of the exact base, verified by SHA-256. The production JVM/protocol sources match between these revisions, and the packaged native-library hashes were verified for the accepted runs.

Current [CI checks](https://github.com/apache/datafusion-comet/pull/5177/checks): 62 successful, 9 skipped. Local validation did not cover the full Spark suite, other Spark versions, or HDFS.

native/core/src/parquet/parquet_support.rs
```diff
@@ -195,6 +217,28 @@ fn parquet_convert_array(
                 list_arr.nulls().cloned(),
             )))
         }
+        (
+            Timestamp(TimeUnit::Millisecond, _),
+            Timestamp(TimeUnit::Microsecond, target_tz),
+        ) if top_level && parquet_options.checked_timestamp_overflow => {
```
sunchao, line 221: **[P2] Preserve pruning when an unrelated filter column is widened**

The checked conversion also turns an existing pruning limitation on a different column into a query failure. With a file containing `id: INT32 = 1` and `ts: TIMESTAMP_MILLIS = 9223372036854776`, read `id` using `LongType` and run `SELECT ts WHERE id < 0`. Spark 4.1.3 and base `215ab706` return zero rows, but this head throws timestamp overflow with row-filter pushdown disabled (the default).

The widening filter remains wrapped in an opaque `CometCastColumnExpr`, so DataFusion cannot use its statistics to discard the group before this new checked timestamp conversion. All four plain/dictionary × ANSI cases reproduce; reading `id` as its original IntegerType passes, as does enabling row-filter pushdown. Please preserve pruning for these supported widening casts, or fall back safely for affected scans, and add this unrelated-filter/schema-evolution case alongside the timestamp-predicate tests.

native/core/src/parquet/schema_adapter.rs
```diff
@@ -522,7 +534,220 @@ impl PhysicalExprAdapter for SparkPhysicalExprAdapter {
...
+                    ScalarValue::TimestampMillisecond(None, file_tz.clone()),
+                ))),
+                Some(v) if v % 1_000 == 0 => millis_list.push(Arc::new(Literal::new(
+                    ScalarValue::TimestampMillisecond(Some(v / 1_000), file_tz.clone()),
+                ))),
+                Some(_) => {}
+            }
+        }
+        let negated = in_list_expr.negated();
+        if millis_list.is_empty() {
```
sunchao: **[P2] Preserve NULL rows for an originally empty NOT IN list**

This branch conflates an originally empty list with a nonempty list emptied by dropping sub-millisecond literals. For original empty `NOT IN`, Spark's nonlegacy semantics and DataFusion return true even for a NULL probe; the generated `ts >= i64::MIN` instead returns NULL and can remove matching rows.

I reproduced this on Spark 4.1.3 with `df.filter(!col("ts").isin())`, `spark.sql.legacy.nullInEmptyListBehavior=false`, and `OptimizeIn` plus `ConvertToLocalRelation` excluded so the empty expression survives optimization. For `[NULL, epoch]`, row-filter pushdown drops the NULL row; for an all-NULL group, the head returns zero rows with pushdown either on or off. Spark and base `215ab706` retain both rows in all cases. The direct expression check likewise changes `[true, true]` to `[NULL, true]`.

Please distinguish an originally empty list before using the sentinel comparison, and add NULL-input coverage. The normal optimizer folds this case away, so the Spark exposure is conditional on the stated optimizer configuration.

native/core/src/parquet/parquet_support.rs
sunchao: Rechecked `9172861`: all four predicate forms in your last reply (ordinary comparison, short IN, null-safe equality, and nested timestamp filtering) now pass the 64-query comparison matrix on both the head and base.

The broader pruning issue still has three remaining cases: IN lists above DataFusion's 20-value cutoff, an unrelated INT32 filter column widened to LongType, and dictionary-based exclusion that cannot be inferred from min/max statistics. Each has a Spark/base-success versus head-overflow reproduction with row-filter pushdown disabled. I posted the distinct reproducers and validation in [the follow-up review](https://github.com/apache/datafusion-comet/pull/5177#pullrequestreview-5094649307), along with a separate empty-NOT-IN NULL-row finding. The nested surviving-row overflow difference you documented remains pre-existing and is not one of these new regressions.

native/core/src/parquet/parquet_support.rs
peterxcli: Fixed in c0ad32ab9. Filtered scans now retain the existing safe timestamp cast, so checked conversion cannot run before a Spark-only pruning path. The regression alternates id values 1 and 3, verifies the enabled case has dictionary-only data pages, and checks id = 2 across ANSI and row-filter settings. The focused Spark suite passed.

native/core/src/parquet/parquet_support.rs
peterxcli: Fixed in c0ad32ab9. Every filtered scan now keeps the existing safe cast, including filters whose physical column needs schema-evolution adaptation. The regression reads INT32 id as LongType, filters id < 0, projects the overflowing timestamp, and runs across the same configuration matrix when schema evolution is enabled. The focused Spark suite passed.

native/core/src/parquet/schema_adapter.rs
peterxcli: Fixed in c0ad32ab9 by removing the finite predicate rewrite and using the filtered-scan fallback before conversion. The regression covers both 20- and 21-value timestamp IN lists across plain/dictionary, ANSI, and row-filter settings, with the overflowing timestamp projected. The focused Spark suite passed.

native/core/src/parquet/parquet_support.rs
peterxcli: Superseded in c0ad32ab9 by the simpler filtered-scan fallback. Ordinary comparison, short and long IN, null-safe equality, nested predicates, dictionary exclusion, and widened filter columns no longer checked-convert values before pruning; unfiltered reads remain checked. The focused Spark regressions pass.

native/core/src/parquet/schema_adapter.rs
peterxcli: Fixed in c0ad32ab9 by deleting the millis-domain predicate rewrite, so an originally empty NOT IN keeps DataFusion and Spark semantics. The regression uses NULL plus epoch, disables OptimizeIn and ConvertToLocalRelation, sets nonlegacy empty-list behavior, and passes with row-filter pushdown both off and on.

native/core/src/parquet/parquet_exec.rs
```diff
@@ -94,6 +94,14 @@ pub(crate) fn init_datasource_exec(
     );
     spark_parquet_options.use_field_id = use_field_id;
     spark_parquet_options.ignore_missing_field_id = ignore_missing_field_id;
+    // Spark can discard filtered-out values before timestamp conversion using statistics,
+    // dictionary, and row-level filters. Comet cannot mirror every pruning path, so applying
+    // checked conversion in a filtered scan can fail on values Spark never reads. Preserve the
+    // existing safe cast for filtered scans and use checked conversion only when every value is
+    // necessarily read.
+    spark_parquet_options.checked_timestamp_overflow = !data_filters
```
sunchao: **[P2] Preserve the filtered-scan fallback when native serialization drops a Spark filter**

`checked_timestamp_overflow` is based on the serialized `data_filters`, but that list can be empty even when Spark has a Parquet-pushable filter. For example, with `spark.comet.expression.IsNull.enabled=false`, consider `SELECT ts ... WHERE id IS NULL` over a row group with optional `id` values that are all non-null and an overflowing `TIMESTAMP_MILLIS` `ts`. Spark pushes `IsNull(id)` into Parquet and prunes the group using its zero null count. `CometNativeScan` omits the disabled expression from its native payload without rejecting the scan, so this flag becomes true and the checked conversion errors before the remaining Spark filter can return zero rows. The base safe cast would return zero rows.

Could we retain whether Spark originally supplied data filters even if none serialize, and add a regression for that case? I verified the source paths but did not run this exact configuration end to end.

sunchao (commented):
I found one source-traced regression in the latest revision. Details inline.

spark/src/main/scala/org/apache/comet/serde/operator/CometNativeScan.scala
```diff
@@ -142,8 +142,10 @@ object CometNativeScan extends CometOperatorSerde[CometScanExec] with CometTypeS
       builder.clearChildren()
 
       if (scan.conf.getConf(SQLConf.PARQUET_FILTER_PUSHDOWN_ENABLED)) {
+        val supportedDataFilters = scan.supportedDataFilters
+        commonBuilder.setHasDataFilters(supportedDataFilters.nonEmpty)
```
sunchao, line 146: **[P2] Carry runtime scalar-subquery filters into the overflow fallback**

`scan.supportedDataFilters` excludes filters containing a `PlanExpression`, so this bit remains false when the only filter is `id <=> (SELECT max(k) FROM lookup)`. At execution, `CometNativeScanExec` resolves that scalar subquery and tries to append the predicate. With `spark.comet.expression.EqualNullSafe.enabled=false`, `exprToProto` drops the resolved predicate without setting `has_data_filters`. Both the bit and the native filter list stay empty, so the native planner passes `None` and enables checked timestamp conversion before Spark's remaining filter runs.

I reproduced the Spark 4.1.3 side with dictionary-only `id` values 1 and 3, a subquery result of 2, and an overflowing `TIMESTAMP_MILLIS` `ts`. Spark's scan shows `PushedFilters: [EqualNullSafe(id,ScalarSubquery#...)]` and returns no rows; an unfiltered read of `ts` raises `long overflow`. The current Comet outcome follows from the planning and serialization paths but was not run end to end locally. Could we carry the presence of these late-bound filters through runtime serialization, even if they cannot be serialized, and add this case to the regression tests?

sunchao (commented):
The previous omitted-filter case is fixed at this head. One late-bound scalar-subquery case remains, detailed inline.

sunchao (approved):

sunchao:
Merged, thanks @peterxcli

PR #4752, opened 2026-06-29
feat: support Iceberg metadata columns _pos, _spec, _file, and _partition

parthchandra:
## Which issue does this PR close?

Adds support for iceberg metadata columns

Closes https://github.com/apache/datafusion-comet/issues/5054

PR #5180, opened 2026-07-31
fix: disambiguate Iceberg scans that share a metadata_location

mbutrovich:

No linked issue. Found via `TestMergeOnReadMerge#testSelfMerge` failing on Iceberg 1.8.1 / Spark 3.4 CI in #4752.

## Rationale for this change

`PlanDataInjector` keys a native Iceberg scan's planning data (schema, file scan tasks, pools) by the scan's `metadata_location` alone. A query can read the same table twice within one native plan (a self-join, or a self-merge's target and source reads) via two scans that are logically different: different projected columns, or different pushed filters. Both scans compute the same key, so one scan's serialized data silently overwrites the other's, and the wrong data gets injected into one of the two scan operators.

This surfaced as `TestMergeOnReadMerge#testSelfMerge`/`testSelfMergeWithCaching` failing on Iceberg 1.8.1 / Spark 3.4 CI with `CometNativeException: Column index N is out of bound` once metadata columns (`_file`, `_pos`, `_spec_id`, `_partition`) stopped forcing a fallback to Spark. The underlying collision is not specific to that feature: any two native Iceberg scans of the same table with different pushed filters or projections, landing in one native plan without an intervening shuffle, can hit it.

- Add `scan_hash_code` to `IcebergScanCommon` (native/proto), set from Iceberg's own `SparkScan.hashCode()`, which already folds in pushed filters, snapshot, branch, and read schema.
- `IcebergPlanDataInjector.getKey`, `PlanDataInjector.findAllPlanData`'s Iceberg case, and `CometIcebergNativeScanExec.doExecuteColumnar` now key on `(metadata_location, scan_hash_code)` instead of `metadata_location` alone.

- New tests in `PlanDataInjectorSuite` construct two Iceberg scan operators sharing a `metadata_location` with different `scan_hash_code`s and assert each receives its own planning data rather than the other's.
- Verified against the original regression: reproduced `TestMergeOnReadMerge#testSelfMerge` and `testSelfMergeWithCaching` failing on Iceberg 1.8.1 / Spark 3.4 without this fix, and passing with it applied.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/comet/CometIcebergNativeScanExec.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/CometIcebergNativeScanExec.scala
@@ -220,11 +220,15 @@ case class CometIcebergNativeScanExec(
   override def doExecuteColumnar(): RDD[ColumnarBatch] = {
     val nativeMetrics = CometMetricNode.fromCometPlan(this)
     val serializedPlan = CometExec.serializeNativePlan(nativeOp)
+    // Key by the same (metadata_location, scan_hash_code) pair PlanDataInjector.injectPlanData
+    // looks up via IcebergPlanDataInjector.getKey (see the scan_hash_code field comment in
+    // operator.proto for why metadata_location alone cannot identify a scan).
+    val injectorKey = IcebergPlanDataInjector.getKey(nativeOp).getOrElse(metadataLocation)
     new CometExecRDD(
       sparkContext,
       inputRDDs = Seq.empty,
-      commonByKey = Map(metadataLocation -> commonData),
-      perPartitionByKey = Map(metadataLocation -> perPartitionData),
+      commonByKey = Map(injectorKey -> commonData),
+      perPartitionByKey = Map(injectorKey -> perPartitionData),
       serializedPlan = serializedPlan,
       defaultNumPartitions = perPartitionData.length,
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/org/apache/spark/sql/comet/CometIcebergNativeScanExec.scala
```diff
@@ -220,11 +220,15 @@ case class CometIcebergNativeScanExec(
   override def doExecuteColumnar(): RDD[ColumnarBatch] = {
     val nativeMetrics = CometMetricNode.fromCometPlan(this)
     val serializedPlan = CometExec.serializeNativePlan(nativeOp)
+    // Key by the same (metadata_location, scan_hash_code) pair PlanDataInjector.injectPlanData
+    // looks up via IcebergPlanDataInjector.getKey (see the scan_hash_code field comment in
+    // operator.proto for why metadata_location alone cannot identify a scan).
+    val injectorKey = IcebergPlanDataInjector.getKey(nativeOp).getOrElse(metadataLocation)
```
parthchandra, line 226: nit: `getKey` on an Iceberg op always returns` Some`, so `.getOrElse(metadataLocation)` never actually fires. probably safe to have it though.

parthchandra (approved):
Awesome fix Matt. Thank you! 
Approved pending ci.

PR #5185, opened 2026-07-31
feat: remove native cast from boolean to decimal

### LINKED ISSUE #5334: Boolean-to-decimal cast produces invalid Decimal128 when 10^scale does not fit precision
### Describe the bug 
`cast_boolean_to_decimal` (`native/spark-expr/src/conversion_funcs/boolean.rs`) can return `Ok` with a  `Decimal128Array` whose values do not fit the declared precision.

For `true`, the implementation writes the unscaled value `10^scale`. For `DECIMAL(3,3)`, that is `1000`, which represents `1.000`. But `DECIMAL(3,3)` only allows `-0.999` to `0.999` (unscaled absolute value ≤ `999`).

The code intends to catch this via `with_precision_and_scale` and remap to `decimal_overflow_error`, but Arrow's `with_precision_and_scale` (arrow-array 58.4.0, `primitive_array.rs:1615`) only validates type metadata (`scale ≤ precision ≤ 38`). It does **not** validate per-value precision, so the overflow remap never runs and invalid data is returned as success.

By contrast, `cast_int_to_decimal128_internal` in the same crate checks each value with `is_validate_decimal_precision`.

This is related to but distinct from #5068. 
#5068 describes the same code as "unconditionally raises `SparkError::NumericValueOutOfRange` when the value does not fit". In fact the code does not raise at all for this case, it returns invalid data.                                                         

  ### Steps to reproduce

  Rust unit test (current behavior returns `Ok`, not an error):
  ```rust
  let array: ArrayRef = Arc::new(BooleanArray::from(vec![Some(true)]));
  let result = cast_boolean_to_decimal(&array, 3, 3).unwrap();                                                  
  // result contains unscaled 1000 labeled as Decimal128(3, 3)
  ```                                                                                                           
                                                                                                              
  Or SQL:                                                                                                       
  ```sql
  SELECT CAST(true AS DECIMAL(3,3));                                                                            
  ```
Spark rejects this value (ANSI: error; legacy/try: NULL — see #5068). 
Comet currently accepts it and produces an invalid decimal.

### Expected behavior
When `10^scale` does not fit the target precision and the batch contains `true`, Comet should not return a successful array containing that unscaled value.

Prefer matching Spark: NULL in legacy/try, `NUMERIC_VALUE_OUT_OF_RANGE` under ANSI (#5068).                 
At minimum, do not emit a `Decimal128(p, s)` whose values violate precision.

1` (scanning for `true` when it fails) or validate each value with `is_validate_decimal_precision`. 

### Additional context
Found while working on #5095 (delegate int/float/boolean→decimal to Arrow safe cast). 

Existing Rust coverage in `boolean.rs` only exercises `DECIMAL(10,4)` (single test `test_bool_to_decimal_cast`). The Scala `CometNativeCastSuite` only exercises non-overflowing targets `DECIMAL(10,2)`, `(14,4)`, `(30,0)`. That is why this was not caught earlier.

andygrove:

Closes #5334

## Rationale for this change

Comet has a hand-written native kernel for casting `boolean` to `decimal`. This is not worth
carrying:

- **Nobody uses it.** Casting a boolean to a decimal is a genuine edge case. There is no
  realistic query shape where a user needs `1`/`0` as a scaled decimal rather than as an
  integer or a double.
- **It is disproportionately complex to implement.** The native kernel has to reproduce
  Spark's full `Decimal.toPrecision` semantics: pick the unscaled value for the target scale,
  decide whether it fits the target precision, and then distinguish non-ANSI (return NULL),
  ANSI (throw a Spark-compatible `NUMERIC_VALUE_OUT_OF_RANGE` error), and `try_cast` (return
  NULL) behavior. Overflow is value-dependent -- `false` scales to `0`, which fits every
  decimal type, while `true` overflows anything that cannot hold `1` (`decimal(1,1)`,
  `decimal(2,2)`, `decimal(38,38)`). That is a lot of surface area, and a lot of ways to
  silently diverge from Spark, for a cast with no real users.

Since `CometCast` mixes in `CodegenDispatchFallback`, marking the cast unsupported does not
push the plan back to Spark. The JVM codegen dispatcher runs Spark's own generated code for
the cast inside the Comet pipeline, so the enclosing projection still executes natively and
the results are Spark-compatible by construction.

- `CometCast.canCastFromBoolean` no longer reports `DecimalType` as `Compatible`, so
  boolean -> decimal is routed through the codegen dispatcher.
- Removed the native `cast_boolean_to_decimal` kernel, its dispatch arm in `cast_array`, and
  its Criterion benchmark.
- The Rust unit test now asserts that the native cast rejects boolean -> decimal, pinning the
  fact that the planner must not send it to the native path.
- `CometCastSuite`: the `cast BooleanType to DecimalType(10,2)` test is ignored to satisfy the
  matrix-consistency assertion in `all valid cast combinations covered`, and a new test pins
  the `Unsupported` support level. The `DecimalType(14,4)` and `DecimalType(30,0)` tests stay
  active and now exercise the dispatch path.
- New SQL file tests `cast_boolean_to_decimal.sql` and `cast_boolean_to_decimal_ansi.sql`
  covering both modes. Non-ANSI covers the long fast path (`precision <= 18`) and the
  `BigDecimal` path (`precision > 18`), the tightest representable types, overflow returning
  NULL, literals, `try_cast`, predicates, and aggregates. The ANSI file covers the edge cases:
  overflow throwing `NUMERIC_VALUE_OUT_OF_RANGE` for `decimal(1,1)`, `decimal(2,2)`, and
  `decimal(38,38)`; overflow raised from inside an aggregate and from inside a filter
  predicate; `false` and `NULL` inputs *not* throwing even for those types; and `try_cast`
  suppressing the error. The non-error queries double as sentinels -- they assert fully native
  execution, so a silent fallback to Spark would fail the test rather than let the
  `expect_error` queries pass vacuously.

New SQL file tests:

```
./mvnw test -Dsuites="org.apache.comet.CometSqlFileTestSuite cast_boolean_to_decimal" -Dtest=none
```

Both files pass, which also confirms the cast stays inside the Comet pipeline via the codegen
dispatcher rather than falling the plan back to Spark. Rust unit tests in
`conversion_funcs::boolean` also pass.


spark/src/test/scala/org/apache/comet/CometCastSuite.scala
```diff
@@ -177,10 +177,25 @@ class CometCastSuite extends CometTestBase with AdaptiveSparkPlanHelper {
     castTest(generateBools(), DataTypes.DoubleType)
   }
 
-  test("cast BooleanType to DecimalType(10,2)") {
+  // Boolean -> Decimal has no native path and is routed through the JVM codegen dispatcher, so
+  // `CometCast.isSupported` reports it as unsupported and the matrix-consistency test above
+  // requires this one to be ignored. Execution coverage, including the ANSI overflow edge cases,
+  // lives in `sql-tests/expressions/cast/cast_boolean_to_decimal{,_ansi}.sql`.
+  ignore("cast BooleanType to DecimalType(10,2)") {
```
comphead, line 188: do we need this ignored test? are we planning to switch it back after future changes?

andygrove: CometCastSuite was implemented before we added support for codegen fallback. If we unignore the test, the suite fails because the expr is marked as unsupported. We'll need to improve this suite now that we have codgen dispatch fallback.  I filed https://github.com/apache/datafusion-comet/issues/5186

andygrove: I updated the comment on the test to explain why it is ignored

andygrove: I created https://github.com/apache/datafusion-comet/pull/5268 to rename `CometCastSuite` to `CometNativeCastSuite`.

mbutrovich (approved):
Approved pending merge conflicts/CI.

PR #5192, opened 2026-08-01
feat: support `explode_outer`

### LINKED ISSUE #2838: Add support for `explode_outer`

https://github.com/apache/datafusion-comet/pull/2836 adds `explode_outer` but it is marked as incompatible due to `UnnestExec` having different semantics to Spark for empty arrays. See https://github.com/apache/datafusion/issues/19053 for more information.

This issue is to track implementing the correct behavior in Comet so that this expression can be enabled by default. 

There are existing tests in `CometGenerateExecSuite`, which are currently ignored.

_No response_

### Additional context

_No response_

### LINKED ISSUE #5224: ListPositionsExpr panics on sliced list input, breaking native posexplode over LIMIT with OFFSET
## Describe the bug

`ListPositionsExpr` (`native/core/src/execution/expressions/list_positions.rs`), which backs the `pos` column for native `posexplode` / `posexplode_outer`, panics when its input `ListArray` has a non-zero offset base (i.e. a sliced list). The panic surfaces as:

```
org.apache.comet.CometNativeException: called `Result::unwrap()` on an `Err` value:
InvalidArgumentError("Max offset of 9 exceeds length of values 6")
```

## To Reproduce

Any query where a native limit with a non-zero offset feeds a native `posexplode` in the same native plan. Reproduced against `main` (3df58d528):

```scala
test("posexplode over limit with offset") {
  withSQLConf(
    "spark.sql.adaptive.enabled" -> "false",
    "spark.sql.leafNodeDefaultParallelism" -> "1",
    CometConf.COMET_EXEC_LOCAL_TABLE_SCAN_ENABLED.key -> "true",
    CometConf.COMET_EXEC_EXPLODE_ENABLED.key -> "true") {
    Seq((1, Array(1, 2, 3)), (2, Array(4, 5)), (3, Array(6)), (4, Array(7, 8)), (5, Array(9)))
      .toDF("id", "arr")
      .createOrReplaceTempView("t")
    val df = spark.sql(
      "SELECT id, posexplode(arr) FROM (SELECT id, arr FROM t LIMIT 4 OFFSET 1)")
    checkSparkAnswerAndOperator(df)
  }
}
```

`spark.sql.leafNodeDefaultParallelism = 1` matters only so that all rows arrive in a single batch. With the default parallelism each partition produces a one-row batch, `LimitStream` discards whole batches instead of slicing, and the bug is masked.

The resulting native plan is:

```
UnnestExec
  ProjectionExec: expr=[col_0@0 as col_0, list_positions(col_1@1) as pos, col_1@1 as col_1]
    GlobalLimitExec: skip=1, fetch=None
      ScanExec: source=[Exchange (unknown)], schema=[col_0: Int32, col_1: List(non-null Int32)]
```

Note that the `CometFilter` Spark inserts above the limit via `InferFiltersFromGenerate` does not mask the bug. All rows pass the predicate, so Arrow's `filter` returns the input arrays untouched and the slice survives.

## Root cause

`ListPositionsExpr::evaluate` builds a fresh values array numbered from zero, but reuses the input's original offset buffer:

https://github.com/apache/datafusion-comet/blob/main/native/core/src/execution/expressions/list_positions.rs#L96-L114

```rust
let offsets = list.offsets();
let total_len = *offsets.last().unwrap() as usize;

let mut values: Vec<i32> = Vec::with_capacity(total_len);
for window in offsets.windows(2) {
    let start = window[0];
    let end = window[1];
    for i in 0..(end - start) {
        values.push(i);
    }
}

let result = ListArray::new(
    element_field,
    offsets.clone(),          // <-- still based at the original offset
    Arc::new(Int32Array::from(values)),   // <-- length is only the slice span
    list.nulls().cloned(),
);
```

`GenericListArray::slice` slices `value_offsets` and `nulls` but leaves `values` unsliced, so a sliced list has `offsets[0] > 0` while `values()` stays full length. `ListPositionsExpr` pushes only `offsets.last() - offsets[0]` values but keeps offsets that run up to `offsets.last()`, so `ListArray::try_new` rejects it on `offsets.last() > values.len()` and `ListArray::new` unwraps the error into a panic.

`DataFusion`'s `LimitStream` produces exactly this shape: `poll_and_skip` does `batch.slice(self.skip, batch.num_rows() - self.skip)`. `planner.rs` maps a Comet limit with a non-zero offset onto `GlobalLimitExec` with that skip. The plain `LIMIT` path (`skip == 0`) slices from zero, keeps the base at zero, and is unaffected.

## Suggested fix

Rebase the offsets to zero so they match the newly built values array:

```rust
let base = offsets[0];
let rebased = OffsetBuffer::new(offsets.iter().map(|o| o - base).collect::<Vec<_>>().into());
```

A unit test that fails before the fix and passes after:

```rust
#[test]
fn sliced_input_with_non_zero_offset_base() {
    // 5-row list, then slice out rows 1..4. `offsets()` becomes [2, 3, 3, 5]
    // (base 2) while `values()` stays unsliced at length 6.
    let values = Int32Array::from(vec![1, 2, 3, 4, 5, 6]);
    let offsets = OffsetBuffer::new(vec![0, 2, 3, 3, 5, 6].into());
    let element_field = Arc::new(Field::new("item", DataType::Int32, true));
    let input = ListArray::new(element_field, offsets, Arc::new(values), None);
    let sliced: ArrayRef = Arc::new(input.slice(1, 3));

    let schema = Schema::new(vec![Field::new("arr", sliced.data_type().clone(), true)]);
    let batch = RecordBatch::try_new(Arc::new(schema), vec![sliced]).unwrap();

    let expr = ListPositionsExpr::new(Arc::new(Column::new("arr", 0)));
    let result = expr.evaluate(&batch).unwrap();
    let out = result.into_array(batch.num_rows()).unwrap();
    let list = out.as_any().downcast_ref::<ListArray>().unwrap();

    assert_eq!(list.len(), 3);
    assert_eq!(
        list.value(0).as_any().downcast_ref::<Int32Array>().unwrap(),
        &Int32Array::from(vec![0])
    );
    assert_eq!(list.value(1).len(), 0);
    assert_eq!(
        list.value(2).as_any().downcast_ref::<Int32Array>().unwrap(),
        &Int32Array::from(vec![0, 1])
    );
}
```

Without the fix this test panics with `InvalidArgumentError("Max offset of 5 exceeds length of values 3")`.

It is worth auditing the other native expressions that rebuild a `ListArray` from `offsets()` plus a freshly built values array for the same assumption.

## Additional context

Affects `posexplode` today. It will also affect `posexplode_outer` once #5192 lands, since that PR makes the outer variant native by default. On `main` today `posexplode_outer` is `Incompatible` and falls back to Spark, so the query above succeeds for the outer variant unless `spark.comet.operator.GenerateExec.allowIncompatible=true` is set. With that config set it panics identically, which is the behavior #5192 makes the default.

`ListEmptyToNullExpr`, added in #5192, handles the sliced case correctly (it preserves the original offsets alongside the unsliced `values()`), but it also preserves the non-zero offset base and passes it straight into `ListPositionsExpr` in the `posexplode_outer` plan.

comphead:

Closes #2838. 
Closes #5224 

Fixes the native `explode_outer` / `posexplode_outer` gap tracked in that issue and referenced from DataFusion [#19053](https://github.com/apache/datafusion/issues/19053).

  ## Rationale for this change

  Comet previously routed `GenerateExec` with `outer = true` back to Spark (`Incompatible`) because DataFusion's `UnnestExec` with `preserve_nulls = true` emits one null row for a NULL list but drops rows
  whose list is empty. Spark's `explode_outer` / `posexplode_outer` must emit one null row for *both* cases, so anything containing empty arrays fell back to JVM whole-stage codegen.

  Native:
  - New `ListEmptyToNullExpr` (`native/core/src/execution/expressions/list_empty_to_null.rs`) rewrites a `List<T>` to mark every empty row as null while preserving the original offsets, values, and column
  name.
  - `planner.rs` wraps the array child with `ListEmptyToNullExpr` when `explode.outer` is true, before positions are computed and before the projection feeds `UnnestExec`. `ListPositionsExpr` inherits the
  modified null bitmap so `pos` and `value` stay aligned for `posexplode_outer`.

  Serde:
  - `CometExplodeExec.getSupportLevel` no longer returns `Incompatible` for `op.outer`. Unsupported cases (maps, non-deterministic generators, multi-input generators, `COMET_EXEC_EXPLODE_ENABLED = false`)
  still fall back to Spark whole-stage codegen through the standard `Unsupported` path.

  Tests:
  - Un-ignored `explode_outer with empty array`, `explode_outer with nullable projected column`, `explode_outer with mixed null, empty, and non-empty arrays` in `CometGenerateExecSuite`.
  - Dropped the `WHERE id != 4` workaround and stale `allowIncompatible` `Config:` directive in `posexplode.sql`.
  - Added `sql-tests/expressions/array/explode.sql` covering `explode` / `explode_outer` (plus `LATERAL VIEW` and `LATERAL VIEW OUTER`) across every primitive element type (boolean, tinyint/smallint/int/bigint
  at min/max, float and double with NaN / ±0 / ±Inf / NULL, decimal(18,4) and decimal(38,10) at boundaries, string with empty and unicode, binary, date, timestamp), nested `array<array<int>>`,
  `array<struct>`, NULLs in id and array columns, literal arrays, empty tables, and an `expect_fallback` for `map` input.

  ## Are these changes tested?

  Yes. New `sql-tests/expressions/array/explode.sql` runs through `CometSqlFileTestSuite` and previously ignored tests in `CometGenerateExecSuite` are now enabled.

  `explode_outer` and `posexplode_outer` now run natively without requiring `spark.comet.operator.GenerateExec.allowIncompatible = true`. No behavioral change for `explode` / `posexplode`.

### DIFF
--- a/docs/source/contributor-guide/native_shuffle.md
+++ b/docs/source/contributor-guide/native_shuffle.md
@@ -43,7 +43,7 @@ Comet Native (columnar) → ColumnarToRowExec → rows → JVM Shuffle → Arrow
 
 Native shuffle (`CometExchange`) is selected when all of the following conditions are met:
 
-1. **Shuffle mode allows native**: `spark.comet.shuffle.mode` is `native` or `auto`.
+1. **Shuffle mode allows native**: `spark.comet.exec.shuffle.mode` is `native` or `auto`.
 
 2. **Child plan is a Comet native operator**: The child must be a `CometPlan` that produces
    columnar output. Row-based Spark operators require JVM shuffle.
@@ -69,9 +69,8 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
                                       ▼
 ┌─────────────────────────────────────────────────────────────────────────────┐
 │                         CometNativeShuffleWriter                             │
-│  - Builds protobuf operator plan: ShuffleWriter(child = childNativeOp)       │
-│  - Reads per-partition leaf iterators from CometNativeShuffleInputIterator   │
-│  - Drives one CometExecIterator per partition                                │
+│  - Constructs protobuf operator plan                                         │
+│  - Invokes native execution via CometExec.getCometIterator()                 │
 └─────────────────────────────────────────────────────────────────────────────┘
                                       │
                                       ▼ (JNI)
@@ -104,14 +103,13 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
 
 ### Scala Side
 
-| Class                          | Location                                         | Description                                                                                                                                         |
-| ------------------------------ | ------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `CometShuffleExchangeExec`     | `.../shuffle/CometShuffleExchangeExec.scala`     | Physical plan node. Validates types and partitioning, creates `CometShuffleDependency`.                                                             |
-| `CometNativeShuffleWriter`     | `.../shuffle/CometNativeShuffleWriter.scala`     | Implements `ShuffleWriter`. Builds the unified `ShuffleWriter(child = childNativeOp)` plan and runs it in one `CometExecIterator` per partition.    |
-| `CometShuffleDependency`       | `.../shuffle/CometShuffleDependency.scala`       | Extends `ShuffleDependency`. Holds shuffle type, schema, range partition bounds, and (native shuffle only) a `NativeShuffleSpec`.                   |
-| `CometNativeShuffleInputRDD`   | `.../shuffle/CometNativeShuffleInputRDD.scala`   | Thin scheduling-anchor RDD on the native-shuffle path. `compute` returns a `CometNativeShuffleInputIterator` carrying per-partition leaf iterators. |
-| `CometBlockStoreShuffleReader` | `.../shuffle/CometBlockStoreShuffleReader.scala` | Reads shuffle blocks via `ShuffleBlockFetcherIterator`. Decodes Arrow IPC to `ColumnarBatch`.                                                       |
-| `NativeBatchDecoderIterator`   | `.../shuffle/NativeBatchDecoderIterator.scala`   | Reads compressed Arrow IPC from input stream. Calls native decode via JNI.                                                                          |
+| Class                          | Location                                         | Description                                                                                   |
+| ------------------------------ | ------------------------------------------------ | --------------------------------------------------------------------------------------------- |
+| `CometShuffleExchangeExec`     | `.../shuffle/CometShuffleExchangeExec.scala`     | Physical plan node. Validates types and partitioning, creates `CometShuffleDependency`.       |
+| `CometNativeShuffleWriter`     | `.../shuffle/CometNativeShuffleWriter.scala`     | Implements `ShuffleWriter`. Builds protobuf plan and invokes native execution.                |
+| `CometShuffleDependency`       | `.../shuffle/CometShuffleDependency.scala`       | Extends `ShuffleDependency`. Holds shuffle type, schema, and range partition bounds.          |
+| `CometBlockStoreShuffleReader` | `.../shuffle/CometBlockStoreShuffleReader.scala` | Reads shuffle blocks via `ShuffleBlockFetcherIterator`. Decodes Arrow IPC to `ColumnarBatch`. |
+| `NativeBatchDecoderIterator`   | `.../shuffle/NativeBatchDecoderIterator.scala`   | Reads compressed Arrow IPC from input stream. Calls native decode via JNI.                    |
 
 ### Rust Side
 
@@ -125,19 +123,11 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
 
 ### Write Path
 
-1. **Plan construction**: `CometNativeShuffleWriter` builds a protobuf operator tree with a
-   `ShuffleWriter` operator at the root and `childNativeOp` as its child. `childNativeOp` takes
-   one of two shapes:
-   - The child plan's `nativeOp` directly, when `CometShuffleExchangeExec`'s child is a
-     `CometNativeExec` subtree. The upstream operators run inside the same `CometExecIterator`
-     as the writer, with no JVM-to-native batch boundary between them.
-   - A synthetic `Scan("ShuffleWriterInput")` placeholder, when the dep was built via the
-     convenience `prepareShuffleDependency(rdd, ...)` overload (used by
-     `CometCollectLimitExec` and `CometTakeOrderedAndProjectExec`, or when the
-     exchange's child is a non-native `CometPlan` such as `CometSparkToColumnarExec`). Native
-     code reads `ColumnarBatch`es from the JVM input iterator via Arrow C Stream Interface.
+1. **Plan construction**: `CometNativeShuffleWriter` builds a protobuf operator plan containing:
+   - A scan operator reading from the input iterator
+   - A `ShuffleWriter` operator with partitioning config and compression codec
 
-2. **Native execution**: A single `CometExecIterator` per partition runs the unified plan.
+2. **Native execution**: `CometExec.getCometIterator()` executes the plan in Rust.
 
 3. **Partitioning**: `ShuffleWriterExec` receives batches and routes to the appropriate partitioner:
    - `MultiPartitionShuffleRepartitioner`: For hash/range/round-robin partitioning
@@ -193,10 +183,8 @@ For range partitioning:
 
 ### Single Partition
 
-The simplest case: all rows go to partition 0. Uses `SinglePartitionShufflePartitioner`, which
-streams each batch straight to the writer, whose `BatchCoalescer` combines small batches up to the
-configured batch size. Batches already at least that size pass through unchanged, so a large input
-batch is written as a single block that may exceed the batch size.
+The simplest case: all rows go to partition 0. Uses `SinglePartitionShufflePartitioner` which
+simply concatenates batches to reach the configured batch size.
 
 ### Round Robin Partitioning
 
@@ -216,9 +204,7 @@ sizes.
 Native shuffle uses DataFusion's memory management with spilling support:
 
 - **Memory pool**: Tracks memory usage across the shuffle operation.
-- **Spill triggers**: Partitions spill to disk when the memory pool denies an allocation, or
-  when the buffered bytes reach `spark.comet.shuffle.native.maxBufferBytes`. That config defaults to
-  0, which disables the fixed limit and leaves memory pressure as the only trigger.
+- **Spill threshold**: When buffered data exceeds the threshold, partitions spill to disk.
 - **Per-partition spilling**: Each partition has its own spill file. Multiple spills for a
   partition are concatenated when writing the final output.
 - **Scratch space**: Reusable buffers for partition ID computation to reduce allocations.
@@ -232,7 +218,7 @@ The `MultiPartitionShuffleRepartitioner` manages:
 ## Compression
 
 Native shuffle supports multiple compression codecs configured via
-`spark.comet.shuffle.compression.codec`:
+`spark.comet.exec.shuffle.compression.codec`:
 
 | Codec    | Description                                            |
 | -------- | ------------------------------------------------------ |
@@ -246,14 +232,14 @@ independently compressed, allowing parallel decompression during reads.
 
 ## Configuration
 
-| Config                                       | Default | Description                              |
-| -------------------------------------------- | ------- | ---------------------------------------- |
-| `spark.comet.shuffle.enabled`                | `true`  | Enable Comet shuffle                     |
-| `spark.comet.shuffle.mode`                   | `auto`  | Shuffle mode: `native`, `jvm`, or `auto` |
-| `spark.comet.shuffle.compression.codec`      | `zstd`  | Compression codec                        |
-| `spark.comet.shuffle.compression.zstd.level` | `1`     | Zstd compression level                   |
-| `spark.comet.shuffle.native.writeBufferSize` | `1MB`   | Write buffer size                        |
-| `spark.comet.shuffle.jvm.batchSize`          | `8192`  | Target rows per batch                    |
+| Config                                            | Default | Description                              |
+| ------------------------------------------------- | ------- | ---------------------------------------- |
+| `spark.comet.exec.shuffle.enabled`                | `true`  | Enable Comet shuffle                     |
+| `spark.comet.exec.shuffle.mode`                   | `auto`  | Shuffle mode: `native`, `jvm`, or `auto` |
+| `spark.comet.exec.shuffle.compression.codec`      | `zstd`  | Compression codec                        |
+| `spark.comet.exec.shuffle.compression.zstd.level` | `1`     | Zstd compression level                   |
+| `spark.comet.shuffle.write.buffer.size`           | `1MB`   | Write buffer size                        |
+| `spark.comet.columnar.shuffle.batch.size`         | `8192`  | Target rows per batch                    |
 
 ## Comparison with JVM Shuffle
 
--- a/docs/source/contributor-guide/native_shuffle.md
+++ b/docs/source/contributor-guide/native_shuffle.md
@@ -49,7 +49,6 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
    columnar output. Row-based Spark operators require JVM shuffle.
 
 3. **Supported partitioning type**: Native shuffle supports:
-
    - `HashPartitioning`
    - `RangePartitioning`
    - `SinglePartition`
@@ -129,7 +128,6 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
 1. **Plan construction**: `CometNativeShuffleWriter` builds a protobuf operator tree with a
    `ShuffleWriter` operator at the root and `childNativeOp` as its child. `childNativeOp` takes
    one of two shapes:
-
    - The child plan's `nativeOp` directly, when `CometShuffleExchangeExec`'s child is a
      `CometNativeExec` subtree. The upstream operators run inside the same `CometExecIterator`
      as the writer, with no JVM-to-native batch boundary between them.
@@ -142,7 +140,6 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
 2. **Native execution**: A single `CometExecIterator` per partition runs the unified plan.
 
 3. **Partitioning**: `ShuffleWriterExec` receives batches and routes to the appropriate partitioner:
-
    - `MultiPartitionShuffleRepartitioner`: For hash/range/round-robin partitioning
    - `SinglePartitionShufflePartitioner`: For single partition (simpler path)
 
@@ -150,13 +147,11 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
    exceeds the threshold, partitions spill to temporary files.
 
 5. **Encoding**: `ShuffleBlockWriter` encodes each partition's data as compressed Arrow IPC:
-
    - Writes compression type header
    - Writes field count header
    - Writes compressed IPC stream
 
 6. **Output files**: Two files are produced:
-
    - **Data file**: Concatenated partition data
    - **Index file**: Array of 8-byte little-endian offsets marking partition boundaries
 
@@ -168,7 +163,6 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
 1. `CometBlockStoreShuffleReader` fetches shuffle blocks via `ShuffleBlockFetcherIterator`.
 
 2. For each block, `NativeBatchDecoderIterator`:
-
    - Reads the 8-byte compressed length header
    - Reads the 8-byte field count header
    - Reads the compressed IPC data
--- a/docs/source/contributor-guide/native_shuffle.md
+++ b/docs/source/contributor-guide/native_shuffle.md
@@ -43,12 +43,13 @@ Comet Native (columnar) → ColumnarToRowExec → rows → JVM Shuffle → Arrow
 
 Native shuffle (`CometExchange`) is selected when all of the following conditions are met:
 
-1. **Shuffle mode allows native**: `spark.comet.exec.shuffle.mode` is `native` or `auto`.
+1. **Shuffle mode allows native**: `spark.comet.shuffle.mode` is `native` or `auto`.
 
 2. **Child plan is a Comet native operator**: The child must be a `CometPlan` that produces
    columnar output. Row-based Spark operators require JVM shuffle.
 
 3. **Supported partitioning type**: Native shuffle supports:
+
    - `HashPartitioning`
    - `RangePartitioning`
    - `SinglePartition`
@@ -69,8 +70,9 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
                                       ▼
 ┌─────────────────────────────────────────────────────────────────────────────┐
 │                         CometNativeShuffleWriter                             │
-│  - Constructs protobuf operator plan                                         │
-│  - Invokes native execution via CometExec.getCometIterator()                 │
+│  - Builds protobuf operator plan: ShuffleWriter(child = childNativeOp)       │
+│  - Reads per-partition leaf iterators from CometNativeShuffleInputIterator   │
+│  - Drives one CometExecIterator per partition                                │
 └─────────────────────────────────────────────────────────────────────────────┘
                                       │
                                       ▼ (JNI)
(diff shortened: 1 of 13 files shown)


andygrove (commented):
Thanks for tackling this. The core approach looks right to me. I traced the null-bitmap logic and it handles the cases that matter: empty rows that are already null are excluded from the fast-path scan, the combined bitmap is `existing & non_empty` so a null row with garbage offsets stays null, sliced input arrays reconstruct correctly because `offsets()` is sliced while `values()` is not, and a zero-row batch takes the fast path. I also checked `GenerateExec` in Spark and the outer null row does put null in `pos` as well as `col`, and the analyzer makes those attributes nullable when `outer` is set, which matches `Field::new("pos", Int32, explode.outer)` in the planner.

The gaps I found are mostly around the new pre-projection wiring and the Rust expression itself.

### Test coverage for the new pre-projection wiring

The type coverage in `explode.sql` is really thorough, thank you for that. Two shapes I could not find, and both exercise the new pre-projection specifically.

First, a query that carries the array column through alongside its own explosion, like `SELECT id, arr, explode_outer(arr) FROM test_explode_int`. That is the only place where the difference between the original array and the null-marked copy is observable, since the passthrough should still show `[]` for the empty row while the exploded value is NULL. Second, a query with no passthrough columns at all, like `SELECT explode_outer(arr) FROM test_explode_int`, which drives `project_list` empty so `projections.len()` is zero in the planner.

Could you add both? The same pair for `posexplode_outer` in `posexplode.sql` would be good too.

### Unnecessary `unsafe`

In `native/core/src/execution/expressions/list_empty_to_null.rs` (the `new_nulls` match), `NullBuffer::new` already does `len - buffer.count_set_bits()` internally, so the `new_unchecked` call is doing the same work behind an `unsafe` block. Could this collapse to the safe version?

```rust
let combined = match existing_nulls {
    None => non_empty,
    Some(existing) => existing.inner() & &non_empty,
};
let new_nulls = NullBuffer::new(combined);
```

### Unit tests for `ListEmptyToNullExpr`

It would be good to have a `#[cfg(test)]` module in `list_empty_to_null.rs`. The bitmap combination is the heart of the change and the SQL tests only reach it through a full plan, so a failure there is a lot harder to diagnose. Worth covering the fast path returning the input untouched when no valid row is empty, a mix of empty, NULL, and non-empty rows, empties combined with a pre-existing null bitmap, a zero-row batch, and a sliced input `ListArray` with a non-zero offset. That last one matters because `evaluate` rebuilds the array from `offsets()`, `values()`, and `nulls()`, and those have different slicing semantics.

### Duplicate field name in the pre-projection

In `planner.rs`, `wrapped_name` comes from the child field, so for `explode_outer(arr)` the pre-projection schema ends up with two fields both named `arr`. It works today because `Column::evaluate` only bounds-checks the index and never compares the name, but it makes `EXPLAIN` output confusing and it would break the moment anything downstream resolves that schema by name. Could the pre-projection column get a reserved name like `__comet_explode_outer_arr`, with the original child field name kept for the output column name in the second projection? That keeps the final schema unchanged.

### Pre-projection when there is no positions column

The comment explains the pre-projection exists so `ListPositionsExpr` and the array passthrough share one evaluation. That reasoning only applies to `posexplode_outer`. For plain `explode_outer`, `child_expr` appears exactly once in `project_exprs`, so the extra `ProjectionExec` is per-batch overhead with nothing to share. Would gating the pre-projection on `explode.position` work, wrapping inline otherwise?

Related: since this PR makes `explode_outer` run natively by default where it used to fall back, do you have any numbers? There is no `GenerateExec` benchmark in the repo today, so even ad-hoc timings in the PR description for an array column with a mix of empty and non-empty rows would help confirm the native path is a win and quantify what the extra pass costs.

### Tracking the upstream fix

Once [datafusion#19053](https://github.com/apache/datafusion/issues/19053) is fixed upstream, `ListEmptyToNullExpr` and the pre-projection become dead weight. Could you file a Comet issue to remove them when that lands, and reference it from the comment in `planner.rs`? Otherwise this workaround will quietly outlive the bug it works around.

### Docs

The blank-line removals in `docs/source/contributor-guide/native_shuffle.md` look unrelated to this change. I checked and main is already clean under prettier, and current prettier accepts the file either way, so nothing in CI is asking for them. Could you revert that file to keep the diff focused?

In `docs/source/user-guide/latest/expressions.md`, the new text says "Requires `spark.comet.exec.explode.enabled=true`". That config defaults to `true`, so "requires" reads like the user has to opt in. Maybe "enabled by default via `spark.comet.exec.explode.enabled`" instead?

comphead:
Thanks @andygrove for the review. Addressed the comments in https://github.com/apache/datafusion-comet/pull/5192/commits/60fe5c0a12d3a876c847b941f09de6f9b063b4b1

docs/source/contributor-guide/native_shuffle.md
```diff
@@ -43,13 +43,12 @@ Comet Native (columnar) → ColumnarToRowExec → rows → JVM Shuffle → Arrow
 
 Native shuffle (`CometExchange`) is selected when all of the following conditions are met:
 
-1. **Shuffle mode allows native**: `spark.comet.shuffle.mode` is `native` or `auto`.
+1. **Shuffle mode allows native**: `spark.comet.exec.shuffle.mode` is `native` or `auto`.
```
andygrove: This is the deprecated name for the config. Content in main is correct here.

andygrove: Actually, there probably shouldn't be updates the native shuffle documentation in the contributor guide as part of the `explode_outer` PR?

native/core/src/execution/planner.rs
```diff
@@ -1838,14 +1839,74 @@ impl PhysicalPlanner {
...
+                    .name()
+                    .to_string();
+
+                // Bridge Spark's outer semantics: DataFusion's `UnnestExec` with
+                // `preserve_nulls = true` emits one null row for a NULL list but drops rows
+                // whose list is empty. Spark's `explode_outer`/`posexplode_outer` must emit
+                // exactly one null row in both cases, so we mark empty rows as null before
+                // unnesting. See https://github.com/apache/datafusion/issues/19053. Once
+                // that upstream fix lands, `ListEmptyToNullExpr` and the pre-projection
+                // below can be removed (TODO: link the Comet tracking issue here).
```
andygrove: does this TODO still need to be addressed?

comphead: I feel we need to keep track DF implementation, and after Comet switch to DF, the current implementation can be drastically simplified

docs/source/contributor-guide/native_shuffle.md
comphead: Right, Claude is hallucinating here

native/core/src/execution/planner.rs
comphead: more correctly would be to track #5210 though, I can address it in follow up PR

andygrove:
I dug into the `ListPositionsExpr` concern from my earlier review and filed https://github.com/apache/datafusion-comet/issues/5224 for it.

Short version: `ListPositionsExpr` panics when its input `ListArray` has a non-zero offset base. It builds a fresh values array numbered from zero but reuses the input's original offset buffer, so `ListArray::new` unwraps an `InvalidArgumentError` from Arrow. `GlobalLimitExec` with a non-zero skip produces exactly that shape, since `LimitStream` does `batch.slice(self.skip, ...)`.

This is pre-existing and not something you introduced. It reproduces on `main` today with plain `posexplode`. I had guessed the `CometFilter` that Spark inserts above the limit via `InferFiltersFromGenerate` would reset the offsets and keep the non-outer path safe, but it does not. All rows pass the predicate, so Arrow's `filter` returns the input arrays untouched and the slice survives.

What this PR changes is the blast radius. `posexplode_outer` falls back today, so it is safe by default. Once it runs natively it hits the same panic. Here is a test that shows the difference. It passes on `main` and fails with this PR:

```scala
test("posexplode_outer over limit with offset") {
  withSQLConf(
    "spark.sql.adaptive.enabled" -> "false",
    "spark.sql.leafNodeDefaultParallelism" -> "1",
    CometConf.COMET_EXEC_LOCAL_TABLE_SCAN_ENABLED.key -> "true",
    CometConf.COMET_EXEC_EXPLODE_ENABLED.key -> "true") {
    Seq((1, Array(1, 2, 3)), (2, Array(4, 5)), (3, Array(6)), (4, Array(7, 8)), (5, Array(9)))
      .toDF("id", "arr")
      .createOrReplaceTempView("t")
    val df = spark.sql(
      "SELECT id, posexplode_outer(arr) FROM (SELECT id, arr FROM t LIMIT 4 OFFSET 1)")
    checkSparkAnswerAndOperator(df)
  }
}
```

With this PR the plan becomes `CometExplode` directly over `CometGlobalLimit -1, 1` and the query fails with:

```
org.apache.comet.CometNativeException: called `Result::unwrap()` on an `Err` value:
InvalidArgumentError("Max offset of 9 exceeds length of values 6")
```

I verified the same failure on `main` by setting `spark.comet.operator.GenerateExec.allowIncompatible=true`, which is the behavior this PR makes the default.

One note on the test. `spark.sql.leafNodeDefaultParallelism = 1` is required rather than cosmetic. With the default parallelism each partition produces a one-row batch, `LimitStream` discards whole batches instead of slicing, and the bug is masked. AQE off just keeps the plan readable.

I would suggest the Scala form over a `posexplode.sql` entry here. The SQL harness does support `-- Config:`, but the reproduction depends on how many files the `INSERT` writes and how the shuffle reader batches them, and that is harder to pin down from a `.sql` file.

Could you either pick up the fix from #5224 as a prerequisite and include this test passing, or add it as `ignore` referencing #5224? The fix itself is small, rebasing the offsets to zero so they line up with the newly built values array:

```rust
let base = offsets[0];
let rebased = OffsetBuffer::new(offsets.iter().map(|o| o - base).collect::<Vec<_>>().into());
```

I lean toward fixing it in this PR since this is the change that makes it reachable by default, but I am fine either way as long as it is tracked rather than silent.

comphead:
It would be great having this test, so we can also validate DF native implementation when the time comes

comphead:
@andygrove are you thinking on having a fix for both outer and regular explodes in this PR?

native/core/src/execution/planner.rs
```diff
@@ -1926,22 +1986,15 @@ impl PhysicalPlanner {
...
                     .collect();
 
-                let array_field = child_expr
-                    .return_field(&child_schema)
-                    .expect("Failed to get field from array expression");
-                let array_col_name = array_field.name().to_string();
-
                 if explode.position {
                     let positions_expr: Arc<dyn PhysicalExpr> =
                         Arc::new(ListPositionsExpr::new(Arc::clone(&child_expr)));
```
mbutrovich, line 1994: This is the call site that runs `ListPositionsExpr` on the wrapped array. Per the discussion on #5224, `ListPositionsExpr::evaluate` reuses the input's original offset buffer against a freshly zero based values array, and that panics when the array has a non zero base offset, for example after `GlobalLimitExec` with a non zero skip.

Before this PR `posexplode_outer` fell back to Spark, so that path was unreachable by default. This PR removes the `allowIncompatible` gate, so it becomes reachable by default here.

Can we either pull in the offset rebasing fix from #5224, or add a regression test marked `ignore` referencing #5224, before this merges?

native/core/src/execution/planner.rs
```diff
@@ -1897,14 +1898,74 @@ impl PhysicalPlanner {
...
+                    .name()
+                    .to_string();
+
+                // Bridge Spark's outer semantics: DataFusion's `UnnestExec` with
+                // `preserve_nulls = true` emits one null row for a NULL list but drops rows
+                // whose list is empty. Spark's `explode_outer`/`posexplode_outer` must emit
+                // exactly one null row in both cases, so we mark empty rows as null before
+                // unnesting. See https://github.com/apache/datafusion/issues/19053. Once
+                // that upstream fix lands, `ListEmptyToNullExpr` and the pre-projection
+                // below can be removed (TODO: link the Comet tracking issue here).
```
mbutrovich: This TODO is still a placeholder: `(TODO: link the Comet tracking issue here)`. Can you file the tracking issue for removing `ListEmptyToNullExpr` and this pre-projection once datafusion#19053 lands, and link it here?

andygrove:
Thanks for the updates. Everything from my last round is addressed, and addressed well.

- The `unsafe` `NullBuffer::new_unchecked` is gone in favor of the safe `NullBuffer::new(combined)`.
- `list_empty_to_null.rs` now has a seven-test module covering both fast paths, mixed empty/null/non-empty rows, empties on top of a pre-existing bitmap, a zero-row batch, sliced input with a non-zero offset, and `return_field` nullability.
- The pre-projection column is named `__comet_explode_outer_<name>`, so the duplicate field name is gone.
- The pre-projection is gated on `explode.position`, so plain `explode_outer` no longer pays for a `ProjectionExec` it cannot share.
- Both query shapes I asked about are in `explode.sql` and mirrored in `posexplode.sql`, plus the two new batch-boundary tests with `COMET_BATCH_SIZE = 4`. Nice addition.
- The `native_shuffle.md` churn is reverted and the `expressions.md` wording now reads "Enabled by default".

Four things left.

### CI: scalafix

The four `Lint Java` jobs are failing in `Run scalafix check`, and it is this PR. Removing the `op.outer` branch from `CometExplodeExec.getSupportLevel` left `Incompatible` imported but unused in `spark/src/main/scala/org/apache/spark/sql/comet/operators.scala` at line 61. The only other match in the file is a string literal. `.scalafix.conf` enables `RemoveUnused`, so dropping `Incompatible` from that import list should clear all four jobs.

The `Spark SQL Tests (Spark 3.5) / spark-sql-sql_core-3` failure is not yours. It failed in `Setup Spark` after 1m36s while the other six 3.5 shards ran the same step and passed. That one just needs a re-run.

### #5224 is now reachable by default

To answer your question above: yes, I would fix both in this PR rather than split it. `native/core/src/execution/expressions/list_positions.rs` is unchanged here, so it still builds a fresh values array numbered from zero while reusing the input's original offset buffer. A `ListArray` with a non-zero offset base panics inside `ListArray::new`. Today `posexplode_outer` falls back so users are safe. Once it runs natively it hits the panic. The fix is small and it covers plain `posexplode` at the same time:

```rust
let offsets = list.offsets();
let base = offsets.first().copied().unwrap_or(0);
let total_len = (*offsets.last().unwrap() - base) as usize;
// ... build values as today ...
let rebased = OffsetBuffer::new(offsets.iter().map(|o| o - base).collect::<Vec<_>>().into());
let result = ListArray::new(
    element_field,
    rebased,
    Arc::new(Int32Array::from(values)),
    list.nulls().cloned(),
);
```

Pair it with the `posexplode_outer over limit with offset` test from my earlier comment. If you would rather keep the fix separate, please add that test as `ignore` referencing #5224 so it is tracked rather than silent.

### The TODO placeholder is still literal

`planner.rs` still says `(TODO: link the Comet tracking issue here)`. I searched and no such issue exists yet. Could you file one for removing `ListEmptyToNullExpr` and the pre-projection once [datafusion#19053](https://github.com/apache/datafusion/issues/19053) lands, then drop the link in place of the placeholder? A `TODO: link` shipped in a comment leaves nothing pointing at the workaround.

### Benchmark numbers

Still missing. This change flips `explode_outer` and `posexplode_outer` from falling back to running natively by default, and there is no `GenerateExec` benchmark in the repo. Even ad-hoc timings in the description for an array column with a mix of empty and non-empty rows would confirm the native path is a win and quantify what the extra pass costs.

native/core/src/execution/expressions/list_empty_to_null.rs
```diff
@@ -0,0 +1,313 @@
...
+        let has_valid_empty = (0..len)
+            .any(|i| offsets[i + 1] == offsets[i] && existing_nulls.is_none_or(|n| n.is_valid(i)));
+        if !has_valid_empty {
+            return Ok(ColumnarValue::Array(Arc::clone(&array)));
+        }
+
+        let non_empty = BooleanBuffer::collect_bool(len, |i| offsets[i + 1] > offsets[i]);
+        let combined = match existing_nulls {
+            None => non_empty,
+            Some(existing) => existing.inner() & &non_empty,
```
mbutrovich: When `existing_nulls` is `Some`, this allocates `non_empty` and then allocates a second buffer for `existing.inner() & &non_empty`. Both conditions are already known per row from the `has_valid_empty` scan above. Can this collapse into one `collect_bool` pass:

```rust
let combined = BooleanBuffer::collect_bool(len, |i| {
    offsets[i + 1] > offsets[i] && existing_nulls.is_none_or(|n| n.is_valid(i))
});
```

That drops one allocation on the path where the input already carries a null bitmap.

I also looked at `NullBuffer::union`/`union_many` (arrow-rs #9692) as an alternative. They don't help here, since one side of the combination is a predicate rather than an existing null buffer, so materializing it first just to union it costs the same allocation this avoids.

mbutrovich (commented):
First pass, thanks @comphead!

comphead:
```
java.lang.RuntimeException: Error executing SQL 'SELECT id, explode(array(rand(0))) FROM test_explode_int WHERE id = 1' nondeterministic expressions are only allowed in Project, Filter, Aggregate or Window, found:
  explode(array(rand(0))),col
  in operator Generate explode(array(rand(0))), false, [col#93426].; line 1 pos 0;
```
However the test passed in pure Spark

spark/src/test/resources/sql-tests/expressions/array/explode.sql
```diff
@@ -0,0 +1,506 @@
...
+  (3, 60, NULL, 70)
+
+query
+SELECT id, explode(array(x, y, z)) AS v FROM test_explode_array_ctor
+
+query
+SELECT id, explode_outer(array(x, y, z)) AS v FROM test_explode_array_ctor
+
+-- ===== Non-deterministic generator child. Spark's
+-- `RewriteGeneratorNondeterministicExpressions` rewrites `explode(array(rand(0)))`
```
andygrove: is this a real Spark rule? I could not find any mention of it in Spark source

native/core/src/execution/planner.rs
andygrove: @comphead Yes, the TODO can stay but should link to the actual issue(s)

andygrove:
I built this branch locally and ran benchmarks so we have numbers on the record for the outer variants.

There was no synthetic-data benchmark for `GenerateExec`, only the TPC-DS micro query in `CometTPCDSMicroBenchmark`, which needs generated TPC-DS data. So I added an `explodeExecBenchmark` to `CometExecBenchmark` and ran that. I am happy to push it as a separate PR, or you are welcome to fold it into this one.

## Setup

Apple M3 Max, JDK 17.0.10, Spark 4.1.3 / Scala 2.13 (the repo default), release native build with `RUSTFLAGS="-Ctarget-cpu=native"`, `local[5]`, 10M input rows written to snappy Parquet.

Two data shapes:

- **mixed**: one row in five holds a NULL array, one in five holds an empty array, the rest hold 1 to 3 elements. This is the shape the outer variants exist to handle and the one that drives `ListEmptyToNullExpr` down its slow path.
- **all non-empty**: every array holds 1 to 3 elements, so the outer cases take the `ListEmptyToNullExpr` fast path.

The benchmark checks `findFirstNonCometOperator` on the executed plan before timing, so none of these numbers are accidentally measuring a silent fallback. All six cases planned fully natively.

Two independent runs, best time in ms:

| query | shape | Spark | Comet | Comet relative |
| `explode` | mixed | 164 / 152 | 146 / 145 | 1.1X / 1.0X |
| `explode_outer` | mixed | 168 / 169 | 122 / 122 | **1.4X / 1.4X** |
| `posexplode_outer` | mixed | 165 / 161 | 171 / 169 | **1.0X / 1.0X** |
| `explode` | all non-empty | 182 / 171 | 91 / 95 | 2.0X / 1.8X |
| `explode_outer` | all non-empty | 166 / 158 | 89 / 95 | 1.9X / 1.7X |
| `posexplode_outer` | all non-empty | 177 / 169 | 129 / 133 | 1.4X / 1.3X |

## What I take from this

**`explode_outer` is a clear win.** 1.4X on the mixed shape and 1.7X to 1.9X when the arrays are dense. That was the open question on this PR and I think it is answered.

**`ListEmptyToNullExpr` is not costing anything measurable.** On the dense shape, plain `explode` with no wrapper and `explode_outer` with the wrapper come out the same, 91 and 95 against 89 and 95. So the extra per-batch scan is effectively free even when it finds nothing to do. On the mixed shape `explode_outer` is actually faster than plain `explode`, 122 against 145. I believe that is DataFusion's `preserve_nulls = false` filtering cost rather than anything this PR introduced, since it shows up on the non-outer path.

**`posexplode_outer` on the mixed shape shows no speedup.** Comet's best time came out marginally behind Spark in both runs, 171 and 169 against 165 and 161. Spark's stdev is 16 to 18ms there so the gap itself sits inside the noise, but the direction was consistent across both runs and there is clearly no gain.

**The `pos` branch looks like where the time goes.** Comparing within the same run on the same data, `explode_outer` at 122ms against `posexplode_outer` at 169ms, and 89 to 95ms against 129 to 133ms on the dense shape. That is roughly plus 40ms either way, about 40% on top. `ListPositionsExpr::evaluate` fills a `Vec<i32>` with a scalar `push` per element in a nested loop, which is a fair amount of per-element work for what is just a repeated ramp. That is pre-existing rather than something you added here, and it also affects plain `posexplode`.

I do not think any of this blocks the PR. `explode_outer` is the headline feature and it delivers. But it seemed worth knowing that `posexplode_outer` is at parity on the shape that matters, and that the positions builder rather than the empty-to-null pass is the thing to look at if we want to improve it. Would you be up for filing a follow-up issue for `ListPositionsExpr` so it does not get lost?

## Caveats on reading the table

Cases land in the 90 to 180ms range and Spark's stdev ranges from 3 to 25ms, so anything under about 10ms of difference is noise. The two shapes use different Parquet files, so cross-shape comparisons carry some scan-cost difference as well. The within-run, same-file comparisons are the trustworthy ones, which is why I leaned on the `explode_outer` against `posexplode_outer` delta rather than on absolute numbers.

native/core/src/execution/planner.rs
mbutrovich: The `(TODO: link the Comet tracking issue here)` placeholder can point at #5210 now, not a hypothetical future issue. DataFusion closed #19053 via apache/datafusion#22100 ("Add support for `unnest_outer` function for arrays"), merged 2026-07-31, the day before this PR opened. That PR replaces `UnnestOptions.preserve_nulls: bool` with `datafusion::common::unnest::NullHandling { Drop, Preserve, PreserveAndExpandEmpty }`. `NullHandling::PreserveAndExpandEmpty` is this exact bridge: it treats an empty list identically to a NULL list.

That release isn't in datafusion 54.1.0 (native/Cargo.toml:41) yet, so it doesn't help this PR today, and you already filed #5210 for the follow-up on 2026-08-02. Once Comet moves to a datafusion release carrying #22100, `ListEmptyToNullExpr`, the whole `(true, true)` / `(true, false)` pre-projection match here, and `preserve_nulls: explode.outer` at line 2051 collapse to:

```rust
null_handling: if explode.outer {
    NullHandling::PreserveAndExpandEmpty
} else {
    NullHandling::Drop
},
```

Linking #5210 here instead of the placeholder answers @andygrove's "does this TODO still need to be addressed" thread directly.

native/core/src/execution/expressions/list_positions.rs
```diff
@@ -105,10 +116,16 @@ impl PhysicalExpr for ListPositionsExpr {
             }
         }
```
mbutrovich, line 115: Given this morning's benchmark: `explode_outer` is 1.4x-1.9x over Spark, `posexplode_outer` is at parity, and the delta between the two on the same data (122ms vs 169ms mixed, 89-95ms vs 129-133ms dense) points at this loop. Before filing the follow-up issue you and @andygrove discussed, worth a quick check of whether replacing the inner `for i in 0..(end - start) { values.push(i) }` with `values.extend(0..(end - start))` moves anything. `Vec::push` re-checks `len == capacity` every call even though `with_capacity` already reserved the exact size; `extend` from a `Range` doesn't carry that per-element branch. Five minutes to check before opening the issue, and if it's noise, the issue can say so with a number attached instead of a guess.

mbutrovich (changes requested):
Two things worth doing before merge: the optimization I listed in the first review, and swap the planner.rs TODO for a link to #5210, which already exists and already explains why (apache/datafusion#22100 shipped `NullHandling::PreserveAndExpandEmpty` upstream, just not in a release yet). The `ListPositionsExpr` loop is worth a five-minute check against `Vec::extend` before filing the perf follow-up, but that's not blocking.

mbutrovich (commented):
I pushed my suggested changes so we can try to get this into 1.0.

andygrove (approved):
LGTM. Thanks @comphead and @mbutrovich

mbutrovich (approved):
Removing my "Requested changes" with a token "Approval" now that @andygrove took a look at my changes.

andygrove:
> Removing my "Requested changes" with a token "Approval" not that @andygrove took a look at my changes.

*now

spark/src/test/resources/sql-tests/expressions/array/explode.sql
comphead: Claude is ingenious sometimes.

comphead:
THanks @andygrove and @mbutrovich for the review. Added 1 more commit to remove Claude's creativity on non-existent rules and also point to `non-deterministic support` in generators which added starting from Spark 3.5. https://github.com/apache/datafusion-comet/pull/5192/commits/eb194591b967212f13f99583636f97bfe603c2cb

Issue #1084, opened 2024-11-14
with datafusion comet，no performance improvement.


shaileneF:
### env：
```
host
os：CentOS Linux release 8.2.2004 (Core)
kernel：4.18.0-193.el8.x86_64
memory：1T
jdk：1.8
maven：3.9.6
spark：3.4
scala：2.12
container：
os：CentOS Linux release 7.4.1708 (Core)
kernel：4.18.0-193.el8.x86_64
cpu cores：128
spark: 3.4.3
memory：1T
jdk：11
maven：3.9.6
spark：3.4
scala：2.12
```
----------------------------------------------
data：TPCDS 100G/1T
----------------------------------------------
### with datafusion comet, spark-submit shell:
```
export COMET_JAR=/export/datafusion-test/comet-spark-spark3.4_2.12-0.3.0.jar

$SPARK_HOME/bin/spark-submit \
    --master local \
    --name comet-tpcbench \
    --driver-memory 20G \
    --conf spark.driver.memory=20G \
    --conf spark.executor.instances=16 \
    --conf spark.executor.memory=40G \
    --conf spark.executor.cores=8 \
    --conf spark.cores.max=128 \
    --conf spark.task.cpus=1 \
    --conf spark.executor.memoryOverhead=3G \
    --conf spark.memory.offHeap.enabled=true \
    --conf spark.memory.offHeap.size=40G \
    --jars $COMET_JAR \
    --conf spark.executor.extraClassPath=$COMET_JAR \
    --conf spark.driver.extraClassPath=$COMET_JAR \
    --conf spark.plugins=org.apache.spark.CometPlugin \
    --conf spark.sql.extensions=org.apache.comet.CometSparkSessionExtensions \
    --conf spark.comet.enabled=true \
    --conf spark.comet.exec.enabled=true \
    --conf spark.comet.exec.all.enabled=true \
    --conf spark.comet.cast.allowIncompatible=true \
    --conf spark.comet.exec.shuffle.enabled=true \
    --conf spark.comet.exec.shuffle.mode=auto \
    --conf spark.shuffle.manager=org.apache.spark.sql.comet.execution.shuffle.CometShuffleManager \
    /export/datafusion-test/datafusion-benchmarks/runners/datafusion-comet/tpcbench.py \
    --benchmark tpcds \
    --data /export/cy/test-data-100G-1024/ \
    --queries /export/datafusion-test/datafusion-benchmarks/tpcds/queries-spark \
[...]

Issue #174, opened 2024-03-07
Explore integration with Delta Lake


sunchao:
### What is the problem the feature request solves?

Comet currently only support either Spark's built-in data sources, or Iceberg ([WIP](https://github.com/apache/iceberg/pull/9841)). We should also consider supporting [Delta Lake](https://delta.io/) in future especially given it already has a Rust implementation [`delta-rs`](https://github.com/delta-io/delta-rs). To achieve that, however, we may need to first move away from our hybrid Parquet reader implementation to a fully native one. 

cc @dennyglee per our discussion

### Describe the potential solution

Integrate Comet with `delta-rs`, so Spark queries reading from Delta Lake tables can also leverage Comet native execution.

viirya:
I'm thinking if we move to fully native reader, does it mean we need to drop current JVM-based source (built-in and Iceberg)? Or we are going to have two types of readers? I ask this because I think they might affect how we handle native execution and they might be conflicting each other.

rtyler:
Hello friends! Checking in from the delta-rs project :smile:  We rely heavily on the [parquet](https://docs.rs/parquet/latest/parquet/) which now has support for almost all the data types I have every seen in the wild with Apache Parquet. We recently broke up subcrates so if you have your own file reading tools, pulling in deltalake-core will give you the smallest dependency surface area necessary to process a Delta table.

We do also publish deltalake-aws, deltalake-azure, deltalake-gcp which have storage specific requirements handled within them, such as some of the silly hacks to work around atomic renames in S3.  

You can also take the metacrate `deltalake` if you want the whole bucket of fun :smile:

andygrove:
Now that Comet supports DataFusion's `DataSourceExec` (when `native_datafusion` scan is enabled) it should be much easier to support `delta-rs`.

dennyglee:
Oh brilliant!  Thanks!

tglanz:
> Now that Comet supports DataFusion's `DataSourceExec` (when `native_datafusion` scan is enabled) it should be much easier to support `delta-rs`.

Is the suggestion means to add information to the CometNativeScan that it is a Delta's Parquet, then in the native to delegate the actual table reading to `delta-rs`?

Nassiel:
Is there any pull request or roadmap to include this or for the moment is just a nice to have in this thread? 

We heavily use spark with deltalake but despite our heavy interest (as company) to use it, we check with moving into pure parquet or iceberg with marginal benefits.

[...]

PR #5196, opened 2026-08-01
docs: add a Frequently Asked Questions page

andygrove:

Closes #.

## Rationale for this change

We get the same handful of user questions repeatedly, and while most of the answers already exist in the docs, they are spread across the tuning guide, the plan guide, the compatibility guide, the roadmap, and the Gluten comparison. A new user with a question phrased in their own words has no entry point to them.

Looking through GitHub discussions and issues for the recurring themes, these came up most:

| Source | Already answered somewhere? |
| --- | --- |
| #3199, #1084 — benchmarked and Spark was faster / no speedup | scattered across tuning + plan guides |
| #5001 — native OOM on large joins/distinct where Spark succeeds | partly in the tuning guide's memory section |
| #1018, #4648, #174 — plans for Delta Lake / the data lake ecosystem | roadmap, but that lives under the *contributor* guide |
| #3043 — Spark 3.3 / JDK 8 support | implied by the supported-versions table, never stated as a "no" |
| #1196 — does Comet use SIMD | nowhere |
| #994 — comparison with Gluten | yes, `about/gluten_comparison.md` |
| #1014 — jars for Apple Silicon | yes, installation guide |

Adds `docs/source/faq.md` with 15 questions in four sections (Getting Started, Performance, Compatibility and Support, Project and Community), and links it from the top-level toctree in `index.md` and from the homepage "Getting Started" blurb.

Two things worth calling out for review:

**The page is a routing layer, not a new source of truth.** Every answer is a few sentences that ends in a link into the relevant guide. Nothing is documented only here. That is the property that stops the page drifting out of sync with the versioned user guide, and I'd suggest we keep it as the rule for future additions.

**The page is unversioned, and links into `user-guide/latest`.** Several answers are roadmap- and project-level ("do you plan to support Delta Lake?", "what's coming next?"), which would go stale if frozen into a per-release snapshot, and versioning the page would mean maintaining N copies of it. Linking to `latest` follows what the homepage CTAs and `gluten_comparison.md` already do — but note `latest` is the development snapshot rather than the current stable release, so the FAQ points users slightly ahead of what most of them are running. I followed the existing convention rather than diverging, and said so in the page intro. Happy to change it if we'd rather unversioned pages targeted stable, though that feels like a separate site-wide decision.

On the Spark/JDK question specifically: the JDK 11 and Spark 3.4 deprecation is already documented in three places (`installation.md`, `compatibility/spark-versions.md`, and `versioning_policy.md`), so the FAQ restates it as a warning admonition and links out rather than becoming a fourth place to keep in sync.

Docs only, no code changes.

- Verified all 14 relative file links resolve and all 10 cross-file heading anchors match a real heading in the target file.
- `npx prettier "**/*.md"` reports both files clean.

I was not able to build the docs locally — `docs/requirements.txt` fails to install in my environment — so the toctree entry is unexercised until CI builds the site. Worth a look at the rendered output before this comes out of draft.

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -1,0 +1,231 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Frequently Asked Questions
+
+Short answers to the questions we see most often from Comet users, each linking to the guide that
+covers the topic in full. As with the rest of the unversioned pages on this site, links point into
+the development snapshot of the user guide. If you are running a released version, the equivalent
+pages for it are linked from the [User Guide](user-guide/index.md) index.
+
+## Getting Started
+
+### Do I need to change my queries or my code to use Comet?
+
+No. Comet is a Spark plugin. You add the Comet jar to an existing Spark job, set
+`spark.plugins=org.apache.spark.CometPlugin`, and Comet rewrites the physical plan behind the scenes.
+Your DataFrame and SQL code is unchanged. See the [installation guide](user-guide/latest/installation.md)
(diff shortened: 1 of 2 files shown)


docs/source/faq.md
```diff
@@ -0,0 +1,231 @@
...
+
+### How do I confirm that Comet is actually accelerating my query?
+
+Three checks, from cheapest to most detailed:
+
+1. Look for `INFO core/src/lib.rs: Comet native library version <version> initialized` in the
+   executor logs. If it is absent, the native library never loaded and nothing is being accelerated.
+2. Read `spark.conf.get("spark.comet.version")` to confirm which build is loaded.
+3. Run `EXPLAIN FORMATTED` on the query and look for `Comet*` operators in the plan. Anything without
+   the `Comet` prefix is running on unmodified Spark.
```
comphead, line 45: would be nice having an example that shows a plan with fallbacks

```
val plan = df.queryExecution.executedPlan
val info = new org.apache.comet.ExtendedExplainInfo()
println(info.generateExtendedInfo(plan))
```

PR #4234, opened 2026-05-06
feat: Add experimental support for accelerated PyArrow UDFs

andygrove:
## Which issue does this PR close?

Closes #957
Closes #4240

## Rationale for this change

Spark's `mapInArrow` / `mapInPandas` plans insert `ColumnarToRow` between a Comet scan and the Python operator, and `MapInBatchExec` then projects the Python output back to rows. Both projections copy data per row even though the upstream Comet plan and the Python runner both speak Arrow. Inside the Python runner, `ArrowPythonRunner` then re-encodes those rows into Arrow IPC, so the data round-trips Arrow → Row → Arrow on the JVM side before ever reaching Python.

This PR replaces the row-based plumbing with a custom `CometArrowPythonRunner` that consumes `Iterator[ColumnarBatch]` directly. The destination `VectorSchemaRoot` (in Spark's allocator, used for the IPC stream) is populated per batch by a single `Unsafe.copyMemory` per buffer per column, copying straight from Comet's vectors into the runner's root. No row materialization, no `ArrowWriter.write(InternalRow)` loop.

Not _zero-copy_ in the strict sense: Comet's Parquet readers each construct their own `RootAllocator`, separate from `ArrowUtils.rootAllocator`, so Arrow's `TransferPair` cannot share buffers across the boundary. Bulk per-buffer memcpy is the next-best alternative and is materially faster than the per-row write loop, especially on wide rows where the row path is dominated by the `InternalRow` accessor overhead. A future PR that unifies the allocator parent would unlock true zero-copy via `TransferPair`.

### I/O asymmetry: why memcpy on input but not on output

The input path memcpys; the output path does not, and that is structural rather than an oversight.

- **Input (JVM → Python).** Two Arrow vector trees already exist in different allocators. The source is Comet's `ColumnarBatch`, allocated from each Parquet reader's private `RootAllocator`. The destination is Spark's persistent IPC root, allocated from `ArrowUtils.rootAllocator`. Because the two roots are unrelated, Arrow's `TransferPair` / `VectorLoader.load` cannot rebind buffers across the boundary. `CometColumnarPythonInput.copyVector` walks the trees in lockstep and `setBytes`-copies each buffer. This is the bridge between the two allocator trees.
[...]

PR #5197, opened 2026-08-01
refactor: rename pyarrowUdf config to pyarrowUDF and fix stale config docs

andygrove:

Part of #4978.

## Rationale for this change

`spark.comet.exec.pyarrowUdf.enabled` and `spark.comet.exec.scalaUDF.codegen.enabled` spell the same acronym two different ways. The [configuration conventions](https://datafusion.apache.org/comet/contributor-guide/config_conventions.html) guide calls this out by name as a red flag, so the guide was citing a live inconsistency in our own config surface.

The key was added in #4234 and has not shipped in a release (it is not in 0.17.1), so it can be renamed outright with **no deprecated alias** — which is only true until 1.0 goes out. Doing it now costs nothing; doing it later means carrying a `withAlternative` through all of 1.x.

While in the conventions doc I swept it for other references that no longer match the code.

Rename (no `withAlternative`, key is unreleased):

- `spark.comet.exec.pyarrowUdf.enabled` → `spark.comet.exec.pyarrowUDF.enabled`

The Scala symbol `COMET_PYARROW_UDF_ENABLED` was already correct, so all Scala callsites and `CometMapInBatchSuite` are untouched. Only the key string changes, in `CometConf.scala`, one doc comment, the user guide, and the two pyspark scripts that set the key by name.

Stale references fixed in `config_conventions.md`:

| Claim in the doc | Reality |
| --- | --- |
| `columnar` is a category segment | No `spark.comet.columnar.*` key remains — unified under `spark.comet.shuffle.*` in #4986 |
| (`debug` absent from the category list) | `spark.comet.debug.enabled` exists |
| `spark.comet.parquet.rowFilterPushdown.enabled` → `COMET_ROW_FILTER_PUSHDOWN` | Actual symbol is `COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED` |
| Acronym rule cites `pyarrowUdf` as mixed-casing example | Fixed by this PR |
| Behavior-change section says to use `.category(CATEGORY_LEGACY)` | No legacy config and no `CATEGORY_LEGACY` constant exists yet; now notes the first one to land must add it |

`mvnw compile -Pspark-4.0` (the profile that owns the PyArrow UDF path) is clean, as is `spotless:check`. The rename is a pure key-string change with no behavior change; existing `CometMapInBatchSuite` coverage goes through the Scala symbol and is unaffected.

### DIFF
--- a/docs/source/contributor-guide/config_conventions.md
+++ b/docs/source/contributor-guide/config_conventions.md
@@ -34,9 +34,9 @@ broadest (the prefix) to narrowest (the specific setting).
   prefix like `spark.sql.comet.`.
 - **Category segment.** The first segment after the prefix names a broad area:
   `exec` (execution and expressions), `scan` (source readers), `parquet` (Parquet-specific
-  settings), `shuffle` (shuffle behavior), `columnar` (columnar shuffle), `explain` (plan
-  explain/logging), `metrics`, `tracing`, `testing`, `convert` (Spark → Comet source
-  conversion), or `expression` / `operator` (per-expression / per-operator flags).
+  settings), `shuffle` (shuffle behavior), `explain` (plan explain/logging), `metrics`,
+  `tracing`, `debug`, `testing`, `convert` (Spark → Comet source conversion), or
+  `expression` / `operator` (per-expression / per-operator flags).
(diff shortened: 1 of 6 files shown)


mbutrovich (approved):
Thanks @andygrove!

PR #4870, opened 2026-07-08
feat: support `WindowGroupLimitExec`

comphead:
## Which issue does this PR close?

Closes #4837 .

## Rationale for this change

 Adds native support for Spark's `WindowGroupLimitExec` (SPARK-37099, Spark 3.5+), which computes per-partition top-K for `ROW_NUMBER`, `RANK`, and `DENSE_RANK`.

  - New native `PartitionedRankLimitExec` (`native/core/src/execution/operators/rank_limit.rs`) does a single streaming pass over the child sorted by `[partition_keys..., order_keys...]`, matching Spark's
  `SimpleLimitIterator` / `RankLimitIterator` tie semantics exactly.
  - Planner (`native/core/src/execution/planner.rs`) routes `ROW_NUMBER` without `PARTITION BY` to `LocalLimitExec` (the sorted child makes first-K equivalent to top-K); every other combination goes to
  `PartitionedRankLimitExec`.
  - New proto message `WindowGroupLimit` and enum `RankLikeFunction` (`native/proto/src/proto/operator.proto`).
  - Scala serde `CometWindowGroupLimitExec` plus per-version `ShimCometWindowGroupLimit` (no-op on 3.4, real on 3.5 / 4.0 / 4.1). Registration in `CometExecRule.nativeExecs` is gated on the shim so the 3.4
  build stays clean.
  - New config `spark.comet.exec.windowGroupLimit.enabled` (default `true`) via `CometConf.COMET_EXEC_WINDOW_GROUP_LIMIT_ENABLED`.
  - Spark diff patches for 3.5.9 / 4.0.4 / 4.1.3 teach `RemoveRedundantWindowGroupLimitsSuite` to also count `CometWindowGroupLimitExec`.
  - SQL file tests under `spark/src/test/resources/sql-tests/expressions/window/` cover `ROW_NUMBER` / `RANK` / `DENSE_RANK`, datatypes, edge cases, and scalar-subquery-shaped rewrites.
  - TPC-DS golden plans updated for q44, q67, q70 (v1_4 3.5 / 4.0) and q67a, q70a (v2_7 3.5).

PR #5201, opened 2026-08-01
feat: report native vs codegen-dispatch expression coverage in extended explain

### LINKED ISSUE #5229: Expression explain tags leak between unrelated plans via the shared Literal.TrueLiteral singleton
### Describe the bug

`CometExecRule.rollUpInfoMessages` lifts `EXTENSION_INFO` and `CODEGEN_DISPATCH_EXPRS` tags off every expression node of an operator:

https://github.com/apache/datafusion-comet/blob/main/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala#L751-L767

```scala
val allExprs = op.expressions.flatMap(_.collect { case e: Expression => e })

val infos =
  op.getTagValue(CometExplainInfo.EXTENSION_INFO).getOrElse(Set.empty[String]) ++
    allExprs.flatMap(_.getTagValue(CometExplainInfo.EXTENSION_INFO)).flatten
infos.foreach(msg => withInfo(exec, msg))
```

Some of the nodes in `allExprs` are process-wide singletons, most importantly `Literal.TrueLiteral`. Catalyst's `TreeNode.copyTagsFrom` copies a rewritten node's tags onto its replacement whenever the replacement has no tags of its own (SPARK-32753). So the first time Comet tags an expression that Catalyst later rewrites into `Literal.TrueLiteral`, the tag is welded onto that singleton for the lifetime of the JVM.

Every plan planned afterwards in the same JVM that contains `TrueLiteral` then inherits the tag. `dynamicpruningexpression(true)`, which `CleanupDynamicPruningFilters` puts on partitioned scans, is the common carrier. The result is a `[COMET-INFO: ...]` message attached to an operator that has nothing to do with the message, in an unrelated query, in an unrelated session.

The write side already guards against structural nodes (`QueryPlanSerde.isStructuralExpr` excludes `Attribute`, `BoundReference`, `Literal`, `Alias`), but the read side above does not, so a tag that arrived on a literal by copying is read back as if the operator itself had produced it.

### Steps to reproduce

Plant a tag on the singleton to stand in for whatever earlier query contaminated it, then plan an unrelated query whose scan carries a cleaned-up DPP filter:

```scala
class TagLeakSuite extends CometTestBase {
  test("EXTENSION_INFO on the TrueLiteral singleton leaks into unrelated plans") {
    Literal.TrueLiteral.setTagValue(CometExplainInfo.EXTENSION_INFO, Set("PLANTED_LEAK"))

    withSQLConf(
      CometConf.COMET_EXTENDED_EXPLAIN_FORMAT.key ->
        CometConf.COMET_EXTENDED_EXPLAIN_FORMAT_VERBOSE,
      "spark.sql.optimizer.dynamicPartitionPruning.enabled" -> "true",
      "spark.sql.autoBroadcastJoinThreshold" -> "-1",
      "spark.sql.adaptive.enabled" -> "false") {
      withTable("fact", "dim") {
        sql("create table fact (v int, p int) using parquet partitioned by (p)")
        sql("insert into fact values (1, 1), (2, 2)")
        sql("create table dim (k int, s string) using parquet")
        sql("insert into dim values (1, 'a'), (2, 'b')")
        val plan = sql("select * from fact join dim on fact.p = dim.k where dim.s = 'a'")
          .queryExecution.executedPlan
        println(new ExtendedExplainInfo().generateExtendedInfo(plan))
      }
    }
  }
}
```

Output, on the `fact` scan, which never saw the tagged expression:

```
CometColumnarToRow
+- CometSortMergeJoin
   :- CometSort
   :  +- CometExchange
   :     +-  CometNativeScan parquet spark_catalog.default.fact [COMET-INFO: PLANTED_LEAK]
   +- CometSort
      +- CometExchange
         +- CometFilter
            +- CometNativeScan parquet spark_catalog.default.dim
```

The contamination step is not hypothetical. It was found while investigating a CI-only plan diff on #5201, which adds a `NATIVE_EXPRS` tag read through this same code path. `CometJoinSuite`'s `BroadcastNestedLoopJoin ... with inequality` tests tag their `GreaterThanOrEqual` conditions; a later test in that suite plans a join whose condition is `TrueLiteral`, and from that point on every TPC-DS plan in the JVM reported a `greaterthanorequal` that the query does not contain:

```
op=CometScanExec exprs=ss_item_sk | ss_ext_sales_price | ss_sold_date_sk
    | isnotnull(ss_sold_date_sk) | dynamicpruningexpression(true) | isnotnull(ss_item_sk)
  carrier=Literal: true
```

That is why `CometTPCDSV1_4_PlanStabilitySuite` produces different output when it runs alone (as `dev/regenerate-golden-files.sh` runs it) than when it runs inside CI's 40-suite `[exec]` batch, where `CometJoinSuite` has already polluted the singleton.

### Expected behavior

A `[COMET-INFO: ...]` message should appear only on the operator whose own expressions produced it, and explain output for a query should not depend on which queries were planned earlier in the same JVM.

### Additional context

Suggested fix: `withInfo`, `withCodegenDispatchExpr` (and `withNativeExpr` on #5201) only ever tag a node with a value derived from that node. When reading, keep only values the carrier could have produced, e.g. filter to `exprDisplayName(e)` for the name-valued tags, or skip structural nodes on the read side to mirror `isStructuralExpr` on the write side. Filtering by carrier is the stronger of the two: it holds for any shared node, not just literals.

Worth checking whether `FALLBACK_REASONS` is exposed to the same copying, since it is also set on expressions.

Impact today is confined to explain output; no query results are affected. It does mean `[COMET-INFO: ...]` segments in an EXPLAIN can be misleading, and it makes any golden file that captures extended explain sensitive to suite ordering.

andygrove:

Closes #5229.

## Rationale for this change

The extended explain summary reports operator coverage and the number of
Spark/Comet transitions, but says nothing about expressions. There was no way to
see how much of a plan's expression evaluation ran in native DataFusion kernels
versus Spark's own generated code inside the JVM codegen dispatcher.

Expression coverage is tracked on the expression tree with a new `NATIVE_EXPRS`
tag, the counterpart of the existing `CODEGEN_DISPATCH_EXPRS`, and both are
rolled up per operator onto the converted Comet plan node alongside the info
messages. The summary line now ends with the counts:

```
Comet accelerated 55 out of 58 eligible operators (94%). Final plan contains 3 transitions between Spark and Comet. Comet accelerated 14 expressions (14 native, 1 codegen dispatch).
```

Two new accessors on `ExtendedExplainInfo` give the same information as data:

```scala
val native: Seq[String] = info.getNativeExpressions(plan)
val dispatched: Seq[String] = info.getCodegenDispatchExpressions(plan)
```

Notes on the counting:

- Counts are **distinct expression names per plan**. A name can appear in both
  buckets when the same function is lowered natively for one set of arguments and
  dispatched for another, so the total is the union rather than the sum. `cast` in
  the TPC-DS goldens is a real example.
- Structural nodes (attribute references, literals, aliases, bound references) are
  excluded. They appear in nearly every expression tree and would swamp the names
  that matter, and excluding literals also keeps the dispatcher's
  closure-serialized payload out of the native count.
- Names are the expression's `prettyName` lowercased (the UDF name for a
  `ScalaUDF`), matching the existing `[COMET-INFO: JVM codegen dispatcher: ...]`
  naming.

Dispatch tagging is no longer gated on `spark.comet.explain.codegen.enabled`.
That config now controls only whether the `[COMET-INFO: ...]` segment is
rendered, so the counts are available by default. The config is now enabled in
`CometPlanStabilitySuite` so the goldens record *which* expressions took the
dispatcher path, not just how many, across TPC-DS that is `cast` and `upper`.

This also carries the fix for #5229, found while investigating a CI-only plan
diff here. Catalyst's `TreeNode.copyTagsFrom` copies a rewritten node's tags onto
its replacement, so an expression that Catalyst rewrites into `Literal.TrueLiteral`
brands that process-wide singleton for the lifetime of the JVM, and every later
plan containing `dynamicpruningexpression(true)` inherits the tag. The read side
now keeps only nodes the serde actually writes to, which fixes `EXTENSION_INFO`
and `CODEGEN_DISPATCH_EXPRS` on main as well as the new `NATIVE_EXPRS`.

New tests in `CometCodegenSuite` cover the split counts and both accessors
(including that they are independent of `explain.codegen.enabled`), the
all-fallback case, the decimal promotion rewrite (where `DecimalPrecision.promote`
rebuilds the tree and the tags have to be lifted back onto the node the operator
holds), and the #5229 tag leak.

TPC-DS plan stability goldens were regenerated for Spark 3.4, 3.5, 4.0, 4.1 and
4.2. All 158 changed `extended.txt` files differ only in summary lines and
`COMET-INFO` segments, no plan shapes changed. The suites were re-run against
the regenerated goldens for 3.4, 3.5, 4.0 and the default profile.


comphead (commented):
Thanks @andygrove would it also address issue I came across recently here https://github.com/apache/datafusion-comet/pull/4870/changes#diff-9d04b74555b859929052456a8fd83506fd2154fa2a5dbf246ec4f1a04b807b11R40

So plan looks fully native however coverage is 97% and 1 transition

andygrove:
> Thanks @andygrove would it also address issue I came across recently here https://github.com/apache/datafusion-comet/pull/4870/changes#diff-9d04b74555b859929052456a8fd83506fd2154fa2a5dbf246ec4f1a04b807b11R40
> 
> So plan looks fully native however coverage is 97% and 1 transition

The root of that plan has a c2r, so that explains the transition and one of the reported fallbacks.

```
CometNativeColumnarToRow
+- CometTakeOrderedAndProject
   +- CometProject
```

There is also a `ReusedSubquery` and perhaps that is not correctly counted as native? I will take a look

comphead:
> There is also a `ReusedSubquery` and perhaps that is not correctly counted as native? I will take a look

For this specific file https://github.com/comphead/arrow-datafusion-comet/blob/d781362f4ce284d5e87997be385fd68d31860b30/spark/src/test/resources/tpcds-plan-stability/approved-plans-v1_4-spark3_5/q67/extended.txt

there is no `ReuseQuery`, plan is fully Cometed, but coverage 97%, most likely because of top C2R

andygrove:
> > There is also a `ReusedSubquery` and perhaps that is not correctly counted as native? I will take a look
> 
> For this specific file https://github.com/comphead/arrow-datafusion-comet/blob/d781362f4ce284d5e87997be385fd68d31860b30/spark/src/test/resources/tpcds-plan-stability/approved-plans-v1_4-spark3_5/q67/extended.txt
> 
> there is no `ReuseQuery`, plan is fully Cometed, but coverage 97%, most likely because of top C2R

yes, there is bug with counting comet reused subquery. I filed an issue https://github.com/apache/datafusion-comet/issues/5203

mbutrovich:
The diff fixes the tag-leak bug described in issue #5229: `Literal.TrueLiteral` is a process-wide singleton, Catalyst's `TreeNode.copyTagsFrom` welds a tag onto it the first time some expression that gets rewritten into it carries one, and every later plan in the same JVM with `dynamicpruningexpression(true)` then inherits that tag. The fix is in `ExtendedExplainInfo.scala:277-294` (`collectExprTagValues` / `isNeverTagged`), it is exercised by the dedicated regression test `CometCodegenSuite.scala:2516` ("tags copied onto the shared TrueLiteral do not leak into unrelated plans"), and it benefits `EXTENSION_INFO` and `CODEGEN_DISPATCH_EXPRS`, not just the new `NATIVE_EXPRS` tag. The description lists "Which issue does this PR close? N/A" and #5229 is still open. Can you add `Closes #5229` (or `Fixes #5229`) so the issue closes on merge and reviewers see this PR carries a bug fix, not only the explain feature.

spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
```diff
@@ -786,7 +791,26 @@ object QueryPlanSerde extends Logging with CometExprShim with CometTypeShim {
...
+      // `promote` rebuilt the tree, so the coverage tags landed on copies that the operator does
+      // not hold. Lift them onto `expr` so `CometExecRule.rollUpInfoMessages`, which walks the
+      // operator's own expressions, still sees them. Skipped in the common case where `promote`
+      // returned the same tree and there is nothing to lift.
+      liftCoverageTags(newExpr, expr)
+    }
+    result
+  }
+
+  private def liftCoverageTags(from: Expression, to: Expression): Unit = {
```
mbutrovich, line 805: `liftCoverageTags` re-homes the `NATIVE_EXPRS` / `CODEGEN_DISPATCH_EXPRS` tags onto `expr` when `DecimalPrecision.promote` (called at `QueryPlanSerde.scala:788`) rebuilds the tree and the tags land on the promoted copy instead of the tree the operator actually holds. None of the new tests in `CometCodegenSuite.scala` exercise this path, the only coverage is incidental to whatever the TPC-DS goldens happen to hit, and if this silently dropped tags no test would fail, it would just undercount native expressions in any plan with decimal arithmetic. Could you add a test with a decimal expression that actually triggers `DecimalPrecision.promote` (for example, adding two `DECIMAL` columns with different precision/scale in a projection), asserting `ExtendedExplainInfo.getNativeExpressions` reports the arithmetic expression.

mbutrovich (commented):
Minor PR description comment and a test suggestion.

Otherwise LGTM, thanks @andygrove!

mbutrovich (approved):
Thanks @andygrove!

andygrove:
Merged. Thanks @mbutrovich

PR #5202, opened 2026-08-01
fix: fail rather than silently return unrebased datetimes from Parquet

### LINKED ISSUE #5195: spark.comet.exceptionOnDatetimeRebase is dead code: wire it up or remove it
### Describe the bug

`spark.comet.exceptionOnDatetimeRebase` is declared but never read. It publishes to the
user-facing configs page and promises a safety guarantee that Comet does not provide.

The config is declared at
[`CometConf.scala:718-728`](https://github.com/apache/datafusion-comet/blob/main/spark/src/main/scala/org/apache/comet/CometConf.scala#L718),
is not marked `.internal()`, and sits in `CATEGORY_EXEC`, so it appears in the generated
configuration documentation. Its doc string says:

> When this is true, Comet will throw exceptions when seeing these dates/timestamps that were
> written by Spark version before 3.0.

Grepping the tree, `COMET_EXCEPTION_ON_LEGACY_DATE_TIMESTAMP` has no reader outside its own
declaration. `SparkParquetOptions::use_legacy_date_timestamp_or_ntz`
([`native/core/src/parquet/parquet_support.rs:78`](https://github.com/apache/datafusion-comet/blob/main/native/core/src/parquet/parquet_support.rs#L78)),
which appears to be its intended destination, is only ever initialized to `false` (lines 111 and
127) and is never consulted.

The consequence is worse than a no-op config. A user who reads the scan limitation in
`compatibility/scans.md`, learns that Comet mis-reads legacy-calendar data, and then sets this
config specifically to be protected from that, gets silence and wrong results instead of the
promised exception.

This is carved out of #5010 so that issue can stay focused on the underlying scan limitation
(implementing rebase, or falling back per scan). The two have different fixes: this one is a
small, self-contained decision about a published config, and does not depend on rebase support
landing.

### Steps to reproduce

```scala
val path = "/tmp/legacy_dates"

// Written by Spark 4.1 in legacy-calendar mode.
spark.conf.set("spark.sql.parquet.datetimeRebaseModeInWrite", "LEGACY")
spark.sql("SELECT cast(s as date) as d FROM VALUES ('1000-01-01'),('1990-01-01') AS v(s)")
spark.conf.unset("spark.sql.parquet.datetimeRebaseModeInWrite")

// Ask Comet to protect us from exactly this data.
spark.conf.set("spark.comet.exceptionOnDatetimeRebase", "true")
spark.read.parquet(path).show()
// +----------+
// |         d|
// +----------+
// |1000-01-06|   <-- silently shifted, no exception raised
// |1990-01-01|
// +----------+
```

Setting the config to `false` produces identical output, which is the point: it has no effect in
either direction.

### Expected behavior

Either of:

1. Wire the config up, so that `true` raises when the native scan encounters a Parquet file
   carrying `org.apache.spark.legacyDateTime` metadata (or otherwise holding hybrid-calendar
   values), matching the doc string; or
2. Remove the config, so Comet stops advertising a guarantee it does not implement. The scan
   limitation itself is documented in
   [`compatibility/scans.md`](https://github.com/apache/datafusion-comet/blob/main/docs/source/user-guide/latest/compatibility/scans.md)
   and tracked by #5010, so removing the config loses no user-facing information.

If option 2 is chosen, the sentence in `scans.md` that currently describes the config as dead code
should be dropped along with it.

### Additional context

* Verified on `apache/main` @ `ba21f02de`, default Maven profile (Spark 4.1.2), macOS aarch64,
  debug build.
* Carved out of #5010. Found while re-running the config audit for #4180.

andygrove:

Closes #5195. Substantially narrows #5010 (Comet no longer returns silently-wrong values for legacy-calendar data; it refuses those reads).

## Rationale for this change

`spark.comet.exceptionOnDatetimeRebase` was declared but never read. It published to the configs page and promised a safety guarantee Comet did not provide: a user who read the scan limitation in `compatibility/scans.md` and set this config specifically to be protected got silence and wrong results instead of the promised exception.

Rather than just wiring it up as an opt-in, this makes the safe behaviour the default, so Comet does not silently return values shifted by up to ten days unless a user explicitly asks for that. Implementing actual rebasing remains #5010.

Two commits: the first wires the config up as an opt-in; the second flips the default, refines the guard with row-group statistics, and promotes the error type.

### The guard

A read is refused only when **all** of the following hold. Each condition exists to avoid rejecting reads that are actually fine.

1. **The scan reads a date or timestamp column**, walking recursively through structs, lists, maps, dictionaries, and unions. Only the requested top-level columns are considered, so reading a modern date column out of a file that also holds an ancient one is fine, and reading no calendar-sensitive column at all never fails.
2. **The footer does not prove the values are already Proleptic Gregorian.** It proves it for an `org.apache.spark.version` >= 3.0.0 (>= 3.1.0 for INT96) with no `org.apache.spark.legacyDateTime` / `org.apache.spark.legacyINT96` marker. The two markers and the two switch versions are tracked separately, as Spark does.
3. **Row-group statistics show the column actually holds an affected value**, or cannot show that it does not.

Condition 3 is what makes a default-on guard tolerable. Spark stamps the legacy marker on a *whole file* whenever the write mode was LEGACY, regardless of whether any value is old enough to rebase — and dates from 1582-10-15 onward rebase to themselves. Rejecting on the marker alone would fail a large number of reads that return perfectly correct results today.

Conservative cases resolve to refusing: a row group with no statistics, and INT96 columns (the Parquet spec gives their 12 bytes no meaningful ordering, so writers produce no usable min/max). An all-null row group is cleared — it has no value to rebase — and so is an empty one.

### Credit

The per-read-option gap was spotted by @peterxcli while reviewing #5048: Spark resolves the rebase modes through `ParquetOptions`, so `.option("datetimeRebaseMode", "CORRECTED")` on the reader overrides the session conf. This PR originally read `SQLConf` directly and ignored the option; it now reads `ParquetOptions`, the same source Spark's own `ParquetFileFormat` uses.

#5048 tackles the same issue with a different tradeoff — it detects legacy files on the driver at plan time and falls back to Spark, so it always returns correct results, where this PR raises. The plan-time footer reads are what this PR is trying to avoid; which tradeoff is right is worth other opinions.

### Version-less files, and configs Comet used to ignore

Reframing condition 2 around *provably Gregorian* rather than *provably legacy* fixed a real miss the fixture tests caught: **Spark 2.4.5 and earlier wrote no `org.apache.spark.version` at all**, so the canonical legacy files carry no marker whatsoever, and the opt-in version of this PR cleared them. Spark resolves those through `spark.sql.parquet.datetimeRebaseModeInRead` / `int96RebaseModeInRead` (default `EXCEPTION`, under which Spark raises on an ancient value). Comet now honors both — configs the docs previously said it ignored.

The result lines up with Spark's own defaults: the reads Comet refuses are the reads Spark either rebases or refuses too. **One narrow divergence remains**: under a read mode of `LEGACY`, Spark rebases a version-less file's values and returns them correctly, whereas Comet has no rebasing to do it with and refuses. That is documented on the `ReadModes` type.

### Thresholds come from Spark, not from constants here

`RebaseDateTime.lastSwitchJulianDay` and `lastSwitchJulianTs` are read on the JVM and sent through the proto. `lastSwitchJulianTs` in particular is derived by Spark from its per-timezone rebase tables (the maximum switch point across all zones) and is not something to re-derive natively. Sending them keeps the thresholds exact for whichever Spark version is in use.

### Error type promoted

`SparkError::LegacyDatetimeRebase`, carried out of the reader boxed in `ParquetError::External` (`AsyncFileReader::get_metadata` can only return a `ParquetError`) and recovered **by downcast, not message matching**, in `try_classify_file_read_error`. The three shims translate it into Spark's `SparkUpgradeException` with the `INCONSISTENT_BEHAVIOR_CROSS_VERSION.READ_ANCIENT_DATETIME` class — the same exception Spark raises for this data, and one its `FileScanRDD` deliberately rethrows rather than wrapping in `FAILED_READ_FILE`, because the file is not corrupt.

```
org.apache.spark.SparkUpgradeException: [INCONSISTENT_BEHAVIOR_CROSS_VERSION.READ_ANCIENT_DATETIME]
You may get a different result due to the upgrading to Spark >= 3.0: reading dates before 1582-10-15
or timestamps before 1900-01-01T00:00:00Z from Parquet files can be ambiguous ... SQLSTATE: 42K0B
Caused by: java.lang.RuntimeException: Comet cannot read this Parquet file: it holds dates or
timestamps written in the legacy hybrid (Julian + Gregorian) calendar, which Comet's native scan does
not rebase ... Set spark.comet.exceptionOnDatetimeRebase=false to read these values as-is without
rebasing, or disable Comet for this query so that Spark rebases them. File: <path>
```

**Worth a reviewer's opinion:** the templated Spark message advises setting `datetimeRebaseModeInRead` to `LEGACY`/`CORRECTED`, which is not Comet's remedy, so Comet's actual guidance travels as the cause. The alternative is a Comet-specific exception type whose message is primary but which is not a `SparkUpgradeException`. I picked matching Spark's type; happy to invert it.

### Enforcement point

The check lives in the Parquet reader factory's `get_metadata`, which runs for every file. Not in the expression adapter: DataFusion only creates that when the logical and physical schemas differ or a predicate is pushed down, so a plain `SELECT d FROM ...` could skip it. Since that factory now has two Comet-specific jobs, it is renamed `EagerPageIndexReaderFactory` -> `CometParquetFileReaderFactory`, module docs split in two, with the upstream apache/datafusion#23978 revert note preserved.

### Cost

Nothing when disabled: the guard is an `Option`, `None` when the config is off or no calendar-sensitive column is read, so it costs one check per file open. When armed, footer and statistics inspection happen once per file open on metadata that was already fetched and cached — no extra I/O, nothing per-row or per-row-group.

### Other changes

- Removed the dead `SparkParquetOptions::use_legacy_date_timestamp_or_ntz` field the issue identified (only ever initialized to `false`, never consulted).
- Moved the config from `CATEGORY_EXEC` to `CATEGORY_SCAN`, alongside the other scan safety toggles, and rewrote its doc string.
- `scans.md`: the limitation moved from "may produce incorrect results" to "raises an error at scan time", with the full trigger conditions spelled out.
- `ParquetReadSuite`'s existing `reading ancient dates before 1582` test asserted the old wrong behaviour ("no rebase, no exception"); it now covers the opt-out path explicitly.

**`ParquetDatetimeRebaseSuite`, 55 tests, passing on Spark 4.1 and Spark 3.4.**

Behaviour, on synthetic files:

- ancient legacy dates raise by default, with no config set — via full read, a date-only projection, and a query that only *filters* on the date column
- a legacy-marked file holding **only modern dates** is read normally, and `checkSparkAnswerAndOperator` confirms it matches Spark. Same for modern timestamps. This is the case that makes the default safe to flip
- two date columns, one ancient: reading the modern one succeeds and matches Spark; reading the ancient one raises
- a projection of only non-datetime columns does not raise
- CORRECTED-written files (the common case) do not raise
- the opt-out reads `1000-01-06` — exactly the silently-wrong value from the issue report
- the rejection is a `SparkUpgradeException` and the chain does **not** contain `Encountered error while reading file`

Behaviour, across the 15 checked-in `before_1582_*` fixtures (date / timestamp micros / millis / INT96 plain / INT96 dict, written by Spark 2.4.5, 2.4.6, and 3.2.0), covering every physical encoding and both the version-stamped and version-less footer paths:

- all 15 raise under `EXCEPTION` read mode
- the 5 version-less ones **match Spark exactly** under `CORRECTED` (`checkSparkAnswerAndOperator`), proving the read-mode configs are honored
- the 10 footer-marked ones still raise under `CORRECTED`, because the footer is authoritative — matching Spark, which ignores the read mode for them
- all 15 are readable with the opt-out

**19 Rust unit tests** over the footer provenance rules (modern CORRECTED writer cleared; pre-switch writer not; each legacy marker independently; the two markers tracked separately; version-less not cleared by default; CORRECTED clearing version-less per encoding; CORRECTED not overriding an explicit marker), threshold scaling (millis rounding toward the unsafe direction, exact micros/nanos, saturation instead of overflow), statistics extraction, and guard arming/disarming including nested types and case-insensitive roots.

Regression: `parquet` Rust module 104/104; `ParquetReadV1Suite` 51/51; `SparkErrorConverterSuite`, `CometConfSuite`, `CometPublicApiSuite`, `ParquetTimestampLtzAsNtzSuite`. `test-compile` clean on spark-3.4, spark-3.5, spark-4.0; `cargo fmt`, `clippy -D warnings`, `spotless:check`, `scalastyle:check` clean.

## Note for reviewers

This changes default behaviour: queries that previously read legacy-calendar data and returned wrong values will now fail. That is the intent, and the guard is deliberately narrow so it only fires where the values really are affected — but it is a behaviour change worth a second opinion on, and `spark.comet.exceptionOnDatetimeRebase=false` is the escape hatch.

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -307,6 +307,7 @@ jobs:
               org.apache.comet.parquet.ParquetReadV2Suite
               org.apache.comet.parquet.ParquetReadFromFakeHadoopFsSuite
               org.apache.comet.parquet.ParquetTimestampLtzAsNtzSuite
+              org.apache.comet.parquet.ParquetDatetimeRebaseSuite
               org.apache.spark.sql.comet.ParquetDatetimeRebaseV1Suite
               org.apache.spark.sql.comet.ParquetDatetimeRebaseV2Suite
               org.apache.spark.sql.comet.ParquetEncryptionITCase
--- a/.github/workflows/pr_build_macos.yml
+++ b/.github/workflows/pr_build_macos.yml
@@ -123,6 +123,7 @@ jobs:
               org.apache.comet.parquet.ParquetReadV2Suite
               org.apache.comet.parquet.ParquetReadFromFakeHadoopFsSuite
               org.apache.comet.parquet.ParquetTimestampLtzAsNtzSuite
+              org.apache.comet.parquet.ParquetDatetimeRebaseSuite
               org.apache.spark.sql.comet.ParquetDatetimeRebaseV1Suite
               org.apache.spark.sql.comet.ParquetDatetimeRebaseV2Suite
               org.apache.spark.sql.comet.ParquetEncryptionITCase
--- a/docs/source/user-guide/latest/compatibility/scans.md
+++ b/docs/source/user-guide/latest/compatibility/scans.md
@@ -56,11 +56,14 @@ The following limitation may produce incorrect results without falling back to S
   data written before Spark 3.0, using the hybrid Julian/Gregorian calendar), Comet reads them as
   if they were written using the Proleptic Gregorian calendar. This produces silently-wrong
   values for dates before October 15, 1582 in both projections and predicates. Comet also
-  ignores `spark.sql.parquet.datetimeRebaseModeInRead` and the file-level
-  `org.apache.spark.legacyDateTime` metadata that would tell it to rebase. The
-  `spark.comet.exceptionOnDatetimeRebase` config is currently dead code and does not raise on
-  legacy-calendar data. Tracked by
-  [#5010](https://github.com/apache/datafusion-comet/issues/5010).
+  ignores `spark.sql.parquet.datetimeRebaseModeInRead`. Tracked by
+  [#5010](https://github.com/apache/datafusion-comet/issues/5010). Set
+  `spark.comet.exceptionOnDatetimeRebase=true` to fail such a scan instead of returning wrong
+  values: Comet then raises when a scan reads a date or timestamp column from a Parquet file whose
+  footer says it was written in the hybrid calendar (`org.apache.spark.legacyDateTime` or
+  `org.apache.spark.legacyINT96` metadata, or an `org.apache.spark.version` before 3.0). Files
+  with no Spark writer version in the footer carry no rebasing signal, so they are read as-is
+  regardless of this config.
 
 The following limitations raise an error at scan time rather than falling back to Spark:
 
--- a/docs/source/user-guide/latest/compatibility/scans.md
+++ b/docs/source/user-guide/latest/compatibility/scans.md
@@ -65,16 +65,27 @@ The following limitations raise an error at scan time rather than falling back t
      out of an affected file is fine.
   2. The Parquet footer does not prove those values are already Proleptic Gregorian. It proves it
      for an `org.apache.spark.version` of 3.0 or later (3.1 for `INT96`) with no
-     `org.apache.spark.legacyDateTime` / `org.apache.spark.legacyINT96` marker. Files with no
-     recorded writer version — Spark 2.4.5 and earlier, and non-Spark writers — are resolved
-     through `spark.sql.parquet.datetimeRebaseModeInRead` and
-     `spark.sql.parquet.int96RebaseModeInRead` exactly as Spark does, so `CORRECTED` reads them
-     as-is and the `EXCEPTION` default does not clear them.
-  3. Parquet row-group statistics show the column actually holds a date before October 15, 1582 or
-     a timestamp before 1900-01-01T00:00:00Z, or cannot show that it does not. A legacy-written
-     file whose values are all newer than that rebases to itself and is read normally. `INT96`
-     columns have no usable statistics (the Parquet spec gives their bytes no ordering), so they
-     are always refused when the footer does not clear them.
+     `org.apache.spark.legacyDateTime` / `org.apache.spark.legacyINT96` marker, and also when
+     `spark.sql.parquet.datetimeRebaseModeInRead` / `spark.sql.parquet.int96RebaseModeInRead` is set
+     to `CORRECTED`, which Comet honors for version-less files exactly as Spark does.
+  3. Parquet row-group statistics show the column actually holds a date before October 15, 1582 or a
+     timestamp before 1900-01-01T00:00:00Z. A legacy-written file whose values are all newer than
+     that rebases to itself and is read normally.
+
+     How a column with no usable statistics is treated depends on what the footer said. If the
+     footer positively marks the file as legacy — an `org.apache.spark.legacyDateTime` /
+     `org.apache.spark.legacyINT96` marker, or a Spark older than the switch version — the read is
+     refused, and that includes every `INT96` column, whose bytes the Parquet spec gives no
+     ordering. If the footer records no writer version at all, the read goes ahead.
+
+  That last distinction matters for files Spark did not write. Hive, Impala, Trino and plain
+  parquet-mr leave `org.apache.spark.version` unset, and Hive writes `TIMESTAMP` as `INT96`.
+  Refusing every unprovable column would make all such timestamp data unreadable through Comet
+  whatever its values, so Comet refuses only what it can positively show is affected. The
+  consequence is one gap: an `INT96` column, or a column written with statistics disabled, in a file
+  with no recorded writer version _and_ holding genuinely ancient values is read unrebased, where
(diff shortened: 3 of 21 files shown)


andygrove:
There is an alternate PR open for this - https://github.com/apache/datafusion-comet/pull/5048 - we should compare them

andygrove:
This is turning out to be more complex that it first appeared. Let's just remove the unused config in 1.0.0 and revisit this later. Superseded by https://github.com/apache/datafusion-comet/pull/5221

Issue #5199, opened 2026-08-01
[EPIC] Planner, serde, and optimizer rule performance audit


andygrove:
## What / Why

We spend a lot of effort profiling and optimizing native *execution*, but comparatively little on the JVM-side **planner**: `CometScanRule` / `CometExecRule`, the `QueryPlanSerde` expression and operator serde, and the post-columnar rules. That cost is paid on the driver for every query, and again for **every query stage** under AQE (see the rule injection points documented in `CometSparkSessionExtensions`), so it shows up as latency on short queries and on workloads with many small stages.

This EPIC collects findings from a first pass over that code. Measurements below are from a local probe on a modest 2-table join + aggregate query (778-char SQL, 90 `Expr` nodes in the serialized plan), Spark 4.1 / JDK 17, debug native build:

```
executedPlan w/ Comet          = 23.53 ms
executedPlan w/o Comet         = 19.05 ms
CometScanRule                  =  0.23 ms
CometExecRule                  =  1.13 ms
total serialized plan bytes    = 58445
query_context serialized bytes = 56227  (96.2% of plan)
```

So the rules themselves are a modest ~4.5 ms of overhead here, but the **serialized plan is 32x larger than it needs to be**, which is paid per task on every executor.

### 1. Serialized plan duplicates the full SQL text on every expression

Tracked separately in https://github.com/apache/datafusion-comet/issues/5200 — 96% of plan bytes, 32x plan-size inflation, 2.4x per-task protobuf round-trip cost. This is the largest item by a wide margin.

### 2. Uncached `Class.getMethod` throughout the Iceberg reflection paths

Likely the largest planner-time cost on Iceberg workloads. `IcebergReflection.extractFileLocation` resolves `location()` on *every* file and detects older Iceberg by **catching** `NoSuchMethodException`, so on `path()`-only versions every call constructs an exception with a stack trace. It is called once per file scan task (`CometIcebergNativeScan.scala`, in the `tasksCollection.asScala.foreach { task => ... }` loop) and once per delete file. In the same loop, `extractDeleteFilesList` re-resolves `keyMetadata()` per task, and `serializePartitionInfo` resolves `spec`, `partitionType`, `fields`, `fieldId` and `get` per task *and* per partition field. `CometScanRule.validateIcebergFileScanTasks` does `getMethod("transform")` + `setAccessible` inside its per-task loop. `IcebergReflection.getFieldInfo` resolves three methods per column.

[...]

PR #3580, opened 2026-02-23
feat: [ANSI] Ansi sql error messages

parthchandra:
## Which issue does this PR close?

Closes parts of https://github.com/apache/datafusion-comet/issues/551 
Closes https://github.com/apache/datafusion-comet/issues/2215 
Closes https://github.com/apache/datafusion-comet/issues/3375

## Rationale for this change

With Spark 4.0 (And Spark 3.5 with Ansi mode), Spark produces ansi compliant error messages that have an error code and in many cases include the original SQL query. When we encounter errors in native code, Comet throws a SparkException or CometNativeException that do not conform to the expected error reporting standard.

## What changes are included in this PR?
This PR introduces a framework to report ansi compliant error messages from native code.

Summary of error propagation - 

  1. Spark-side query context serialization : For every serialized expression and aggregate expression, a unique `expr_id` is generated. If the expression's origin carries a `QueryContext` (SQL text, line, column, object name), it is extracted and attached to the protobuf. This is done for both `Expr` and `AggExpr`. 
  2. Native planner (planner.rs): The PhysicalPlanner now holds a `QueryContextMap`. When planning `Expr` and `AggExpr` nodes, if `expr_id` and `query_context` are present,  the context is registered in the map. When creating physical expressions for Cast, CheckOverflow, ListExtract, SumDecimal, AvgDecimal, and arithmetic binary expressions, the relevant QueryContext is looked up and passed to the constructor.
  3. Native errors : The `SparkError` enum is extended with new variants will all the Spark ANSI errors (from https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/errors/QueryExecutionErrors.scala). A new `SparkErrorWithContext` type wraps a `SparkError` with a `QueryContext`. All affected expression implementations look up the context and produce a `SparkErrorWithContext` when available. 
The `SparkError` implementation also has new `to_json()` and `exception_class()` methods for JNI serialization.
  4. JNI boundary (errors.rs -> CometQueryExecutionException): The `throw_exception` function now checks for `SparkErrorWithContext` or `SparkError` and throws `CometQueryExecutionException`. `CometQueryExecutionException` carries the entire `SparkErrorWithContext` as a JSON message. On the Scala side, `CometExecIterator` catches this exception and calls `SparkErrorConverter.convertToSparkException()` to convert to the appropriate Spark exception. If the JSON message contained the `QueryContext`, the exception will contain the query, otherwise it will not.
[...]

PR #5204, opened 2026-08-01
perf: intern QueryContext SQL text into a per-plan pool (up to 20x smaller serialized plans for TPC-DS)

### LINKED ISSUE #5200: Serialized native plan duplicates the full SQL text on every expression (96% of plan bytes)
## Describe the bug

`QueryPlanSerde` attaches a `QueryContext` to **every** expression it serializes, and `QueryContext.sql_text` is the *full* SQL text of the query. The same string is therefore embedded once per expression in the serialized native plan.

On a modest 2-table join + aggregate query (778-char SQL), 71 of the 90 `Expr` nodes in the plan carried a copy of the query text:

```
total serialized plan bytes    = 58445
Expr nodes in plan             = 90
Expr nodes with query_context  = 71
query_context serialized bytes = 56227  (96.2% of plan)
distinct sql_text strings      = 1
plan bytes with ctx stripped   = 1803          <- 32x smaller
per-task proto round trip full = 0.153 ms
per-task proto round trip slim = 0.063 ms      <- 2.4x faster
```

TPC-DS queries are 1-3 KB of SQL, so the ratio gets worse, not better.

## Where it comes from

- `QueryPlanSerde.exprToProtoInternal` ends with a `.map` that calls `extractQueryContext(expr)` for every converted expression; `aggExprToProto` does the same for aggregates.
- `QueryContext.sql_text` in `native/proto/src/proto/expr.proto` is declared as the full query text, and `SQLQueryContext.sqlText` is what gets copied into it.

## Why it costs more than just bytes

1. **Plan size.** The `Array[Byte]` is captured by `CometExecRDD` and shipped in the task binary for the stage.
2. **Per-task JVM protobuf round trip.** Whenever a plan contains a native scan (i.e. essentially every scan-rooted Comet stage), `CometExecRDD.compute` does `Operator.parseFrom(serializedPlan)` then `PlanDataInjector.injectPlanData` then `serializeOperator` — **per task**. Measured at 0.153 ms vs 0.063 ms with the duplication removed. At 10k tasks that is ~1.5 s of JVM protobuf churn per stage, ~60% of it avoidable.
3. **Native side.** `PhysicalPlanner::create_expr` and `create_agg_expr` do `ctx_proto.sql_text.clone()` into a fresh `String` and register it in `QueryContextMap`, per expression, per task. That is 71 identical heap strings per plan per task, held for the lifetime of the query. Note `QueryContext.sql_text` is already `Arc<String>` internally, so the sharing was anticipated — it is just never realised because each proto entry decodes to its own `String`.

## Proposed fix: intern the SQL text into a pool

Truncating to just the relevant fragment is *not* an option: `QueryContext::format_summary` prints the entire `sql_text` (that is what Spark's `== SQL (line N, position M) ==` block shows), so the full text is genuinely needed for the message. The duplication is the problem, not the content.

- Add `repeated string sql_text_pool` to `Operator`, populated only on the root operator of a serialized native block.
- Add `optional int32 sql_text_idx` to `QueryContext`; when set, `sql_text` is left empty.
- Intern as a post-pass in `CometNativeExec.convertBlock()`, which is the point where the full operator tree for a block is in hand. Doing it there means none of the ~90 serde call sites change and no plan-scoped mutable state has to be threaded through `exprToProto`.
- Keep `sql_text` working as a fallback when `sql_text_idx` is absent, so the paths that serialize `nativeOp` directly without going through `convertBlock()` (e.g. `CometNativeWriteExec`) keep working unchanged.
- On the native side, resolve `sql_text_idx` against the pool and share one `Arc<String>` across all contexts that reference it, which also removes the per-expression `String` clone.

Error messages are unchanged; this is purely a wire-format deduplication.

## Steps to reproduce

Plan any SQL query with Comet enabled, take `CometNativeExec.serializedPlanOpt.plan.get`, parse it with `OperatorOuterClass.Operator.parseFrom`, and walk the tree summing `Expr.getQueryContext.getSerializedSize` against the total byte count.

## Expected behavior

The serialized plan should contain each distinct SQL text once, not once per expression.

## Additional context

Found while doing a broader planner / serde / optimizer-rule performance audit: https://github.com/apache/datafusion-comet/issues/5199

andygrove:

Closes https://github.com/apache/datafusion-comet/issues/5200. Part of the planner performance audit epic https://github.com/apache/datafusion-comet/issues/5199.

## Rationale for this change

`QueryPlanSerde` attaches a `QueryContext` to **every** expression it converts, and `QueryContext.sql_text` is the *full* SQL text of the query. The same string therefore ends up embedded once per expression in the serialized native plan. On real TPC-DS queries the query text is 85-95% of the plan bytes.

The context itself is needed — `QueryContext::format_summary` prints the whole `sql_text` to build Spark's `== SQL (line N, position M) ==` error block, so it cannot be trimmed to a fragment. The *duplication* is the problem.

This costs more than just bytes:

1. **Plan size.** The `Array[Byte]` is captured by `CometExecRDD` and shipped in the stage's task binary.
2. **Per-task JVM protobuf round trip.** Whenever a plan contains a native scan — essentially every scan-rooted Comet stage — `CometExecRDD.compute` does `Operator.parseFrom(serializedPlan)` → `PlanDataInjector.injectPlanData` → `serializeOperator`, **per task**.
3. **Per-task native decode.** `PhysicalPlanner::create_expr` / `create_agg_expr` did `ctx_proto.sql_text.clone()` into a fresh `String` per expression and registered it in `QueryContextMap`. For q64 below that is 867 identical heap copies of a 3 KB query, per task, held for the lifetime of the query.

## Benchmarks

Real TPC-DS queries (v1.4 SQL from the Spark test resources) planned against empty TPC-DS tables — planning only, no data needed. "before" is the exact byte count obtained by inlining the pooled text back onto every context, so it is the real pre-patch serialization, not an estimate. "round trip" is the parse + re-serialize that `CometExecRDD.compute` performs per task.

| query | SQL chars | `Expr` nodes | with context | plan bytes before | plan bytes after | reduction | per-task round trip before | after |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| q23a | 1654 | 178 | 168 | 284 KB | 32 KB | 8.9x | 0.84 ms | 0.37 ms |
| q23b | 2009 | 240 | 221 | 441 KB | 47 KB | 9.3x | 0.75 ms | 0.32 ms |
| q64 | 3051 | 1112 | 867 | 2,360 KB | 109 KB | 21.5x | 3.02 ms | 0.73 ms |
| q67 | 884 | 258 | 249 | 247 KB | 24 KB | 10.3x | 0.41 ms | 0.20 ms |
| q72 | 1261 | 282 | 263 | 381 KB | 38 KB | 9.8x | 0.59 ms | 0.23 ms |
| q95 | 976 | 168 | 137 | 153 KB | 24 KB | 6.2x | 0.31 ms | 0.17 ms |
| q14a | 3716 | 513 | 393 | 1,446 KB | 95 KB | 15.2x | 1.87 ms | 0.47 ms |
| q47 | 1731 | 306 | 248 | 433 KB | 29 KB | 14.5x | 0.65 ms | 0.34 ms |
| **total** | | | | **5,749 KB** | **401 KB** | **14.3x** | | |

The worst case, q64, went from a **2.4 MB** serialized plan to 109 KB. Its per-task parse + re-serialize drops from 3.0 ms to 0.7 ms; at 10k tasks that is ~23 s of JVM protobuf work per stage removed, before counting the native-side decode and the 867 string clones per task.

Environment: Spark 4.1 / Scala 2.13, JDK 17, macOS arm64, debug native build. Measured with a throwaway ScalaTest suite (not committed — a proper planner benchmark is tracked in the epic); the committed `QueryContextInternerSuite` asserts the size reduction so the property does not silently regress.

Intern the SQL texts into a pool instead of repeating them:

- **proto**: `Operator` gains `repeated string sql_text_pool`, populated only on the root operator of a serialized native block. `QueryContext` gains `optional int32 sql_text_idx`; `sql_text` is cleared when the index is set.
- **`QueryContextInterner`** (new): rewrites an operator tree so each distinct SQL text appears once in the root's pool. Applied in `CometNativeExec.convertBlock()`, which is where the whole block is in hand and where the pool indices are scoped. Doing it as a post-pass means none of the ~90 `QueryPlanSerde` call sites change and no plan-scoped mutable state has to be threaded through `exprToProto`.
- **native**: `PhysicalPlanner` loads the pool from the root operator (`with_sql_text_pool`) and `build_query_context` resolves `sql_text_idx` against it, handing every context a clone of one shared `Arc<String>`. `QueryContext::new` now takes `impl Into<Arc<String>>` so the `Arc` is shared rather than re-allocated per expression — the struct already stored `Arc<String>`, that sharing was just never realised.
- `sql_text` remains a working fallback when `sql_text_idx` is absent, so the paths that serialize an operator directly without going through `convertBlock` (e.g. `CometNativeWriteExec`) are unaffected.
- **`sourceKey` derivation**: `CometNativeScanExec.apply` hashes a scan's `NativeScanCommon` on the driver to produce the key that `NativeScanPlanDataInjector.getKey` recomputes on the executor from the *serialized* plan. Since only the executor's copy is interned, that hash no longer matched and the scan failed with `Missing planning data for key: ...`. Both sides now hash the context-free form via `QueryContextInterner.stripQueryContexts`, which also makes the key robust to any future change in how contexts are encoded.

The interner walk is descriptor-driven rather than a hand-written recursion over `Expr`'s ~75-way `oneof`: a `QueryContext` can sit on any `Expr` or `AggExpr` at any nesting depth, and a hand-written walk would silently miss newly added variants. `QueryContextInternerSuite` asserts that no un-pooled context survives in a real plan, which is what keeps the walk honest.

New `QueryContextInternerSuite`:
- every `QueryContext` in a real plan carries `sql_text_idx` and no inline `sql_text`, and the pool holds the text once;
- the plan is materially smaller than the inlined equivalent;
- an ANSI `DIVIDE_BY_ZERO` error still reports the full SQL text end-to-end, i.e. the pooled text round-trips JVM → proto → native → error JSON → Spark exception unchanged.

Regressions run locally on Spark 4.1 / Scala 2.13: `CometExpressionSuite`, `SparkErrorConverterSuite`, `CometExecSuite`, `CometAggregateSuite`, `CometCastSuite`, `CometJoinSuite`, `CometNativeReaderSuite`, `ParquetReadSuite`, and the full `cargo test` suite.

Error messages are unchanged; this is purely a wire-format deduplication.

## Additional context

The per-expression `QueryContext` was introduced in https://github.com/apache/datafusion-comet/pull/3580 ("feat: [ANSI] Ansi sql error messages"), which added the SQL context needed for Spark-compatible ANSI error messages. That behaviour is preserved here — only the encoding changes.

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -343,6 +343,7 @@ jobs:
               org.apache.comet.CometNativeSuite
               org.apache.comet.CometConfSuite
               org.apache.comet.CometPublicApiSuite
+              org.apache.comet.QueryContextInternerSuite
               org.apache.comet.CometSetOpWithGroupBySuite
               org.apache.comet.CometSparkSessionExtensionsSuite
               org.apache.spark.CometPluginsSuite
--- a/.github/workflows/pr_build_macos.yml
+++ b/.github/workflows/pr_build_macos.yml
@@ -159,6 +159,7 @@ jobs:
               org.apache.comet.CometNativeSuite
               org.apache.comet.CometConfSuite
               org.apache.comet.CometPublicApiSuite
+              org.apache.comet.QueryContextInternerSuite
               org.apache.comet.CometSetOpWithGroupBySuite
               org.apache.comet.CometSparkSessionExtensionsSuite
               org.apache.spark.CometPluginsSuite
--- a/native/common/src/query_context.rs
+++ b/native/common/src/query_context.rs
@@ -59,9 +59,12 @@ pub struct QueryContext {
 }
 
 impl QueryContext {
+    /// `sql_text` accepts either an owned `String` or an already-shared `Arc<String>`. The plan
+    /// deserializer passes an `Arc` cloned from the plan's SQL text pool so that all the contexts
+    /// in one plan share a single allocation instead of each cloning the full query text.
     #[allow(clippy::too_many_arguments)]
     pub fn new(
-        sql_text: String,
+        sql_text: impl Into<Arc<String>>,
         start_index: i32,
         stop_index: i32,
         object_type: Option<String>,
@@ -70,7 +73,7 @@ impl QueryContext {
         start_position: i32,
     ) -> Self {
         Self {
-            sql_text: Arc::new(sql_text),
+            sql_text: sql_text.into(),
             start_index,
             stop_index,
             object_type,
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -781,6 +781,7 @@ pub unsafe extern "system" fn Java_org_apache_comet_Native_executePlan(
                 let planner =
                     PhysicalPlanner::new(Arc::clone(&exec_context.session_ctx), partition)
                         .with_exec_id(exec_context_id)
+                        .with_sql_text_pool(&exec_context.spark_plan)
                         .with_task_context(exec_context.task_context.clone());
                 let (scans, shuffle_scans, root_op) = planner.create_plan(
                     &exec_context.spark_plan,
--- a/native/core/src/execution/planner.rs
+++ b/native/core/src/execution/planner.rs
@@ -240,6 +240,11 @@ pub struct PhysicalPlanner {
     partition: i32,
     session_ctx: Arc<SessionContext>,
     query_context_registry: Arc<datafusion_comet_spark_expr::QueryContextMap>,
+    /// SQL texts referenced by `QueryContext.sql_text_idx`, taken from the root operator's
+    /// `sql_text_pool`. Held as `Arc`s so that every context in the plan shares one allocation
+    /// rather than each cloning the full query text out of the proto. Empty when the plan was
+    /// serialized without interning, in which case contexts carry `sql_text` inline.
+    sql_text_pool: Vec<Arc<String>>,
     /// Captured at `createPlan` time on `ExecutionContext`; see that struct for the
     /// propagation rationale. `None` when no driving Spark task is available.
     task_context: Option<Arc<Global<JObject<'static>>>>,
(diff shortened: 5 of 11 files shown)


comphead (commented):
thanks @andygrove good catch

CI status doesn't look correct. only 3 checks triggered

andygrove:
Follow-up notes from a self-review pass, for reviewer context:

**Coverage.** The interning applies to plans serialized via `CometNativeExec.convertBlock` — the main scan-rooted path, which is what the benchmark numbers measure. These paths still serialize inline `sql_text` and are **not** covered here: `CometNativeShuffleWriter` (per task), `CometIcebergNativeScanExec`'s `convertBlock` override, `CometNativeWriteExec`, `CometTakeOrderedAndProjectExec`, and the bare-`Expr` Parquet filter path (which has no root operator to hang a pool on, so it will always need the inline form). Routing the operator paths through the interner as well is a small change — move the `intern` call into `CometExec.serializeNativePlan` — but it adds a tree walk to the per-task shuffle-write path, so I left it out of this PR rather than ship it untested. Worth a follow-up.

**`sourceKey` altitude.** The scan plan-data key is a content hash over `Message.toString` of several `NativeScanCommon` fields, computed independently on the driver and the executor. That is what broke when interning changed the context encoding, and it will be fragile against any future encoding change. This PR collapses the two copies into one function so it can only be got wrong in one place, but the deeper fix is to carry an explicit key in the proto the way `IcebergScanCommon.scan_hash_code` already does — which would also delete `stripQueryContexts`' only caller. Tracked as a follow-up under the audit epic.

comphead (approved):
Thanks @andygrove the PR brings significant benefits

PR #5205, opened 2026-08-01
chore: reduce runs for  `Lint Java`

comphead:

Closes #.

## Rationale for this change

Comet currently running `Lint Java` for a matrix which doesn't make a lot of sense. It would be enough to run the Lint once for latest Spark versions & latest JDK

image


andygrove:
what if lint error is one spark version specific shim?

comphead:
> what if lint error is one spark version specific shim?

this is true, I was confused with another one called `Lint Scala`. 
Indeed `Java lint` check Scala shims

I feel  we still can cut back on `lint java` a little bit and run it for latest JDK and Spar3/4 version, WDYT?

Instead of running 4 times it is supposed to run twice. Once for Spark 3.5 JDK 17 and Spark 4.0 JDK 21?

PR #5215, opened 2026-08-02
fix: support Utf8/LargeUtf8/Utf8View in native RLike without panicking

### LINKED ISSUE #5102: rlike: consider arrow regexp_is_match kernel and fix non-StringArray panic

`RLike::is_match` in `native/spark-expr/src/predicate_funcs/rlike.rs` loops over a `StringArray` calling `regex::Regex::is_match` per row into a `BooleanBuilder`. Arrow's `regexp_is_match` kernel does the same thing: it is built on the same Rust `regex` crate, so the documented divergence from Java regex semantics is unchanged, and null propagation matches.

The kernel would also make the array path more robust: it supports Utf8/LargeUtf8/Utf8View, while the current code hard-`expect`s a `StringArray` and would panic on other string layouts.

Replace the row loop with `regexp_is_match(array, &StringArray::new_scalar(pattern), None)`.

Trade-offs to resolve:
- The struct currently pre-compiles the `Regex` once at plan build; the kernel takes a pattern string and compiles per batch. For complex patterns this is a measurable regression; there is no arrow API accepting a pre-built `Regex`. Either accept per-batch compilation or keep the loop and only fix the downcast robustness.
- The dictionary fast path (run on dictionary values then `take`) and the scalar-input branch stay as wrappers.

### Additional context

sam-1112:

Closes #5102.

## Rationale for this change

Native `RLike` array evaluation hard-downcasted to `StringArray` and panicked on other Arrow string layouts (`LargeUtf8` / `Utf8View`). This PR keeps the plan-time compiled `Regex` (instead of Arrow's per-batch `regexp_is_match`) and hardens layout / dictionary handling so those inputs return correct results or a typed error instead of panicking.

Today's common Spark `string` / Parquet path still shows up as `Utf8`, and the planner currently casts some `Utf8View` UDF results back to `Utf8`, so `LargeUtf8` / `Utf8View` may not be easy to hit end-to-end yet. The change is still useful hardening for the layouts called out in #5102, plus correct dictionary key-type handling.

- Generalize array matching over `Utf8` / `LargeUtf8` / `Utf8View` via `StringArrayType`.
- Use `as` values longer than 12 bytes to exercise the out-of-line `Utf8View` representation.
- Use `as_any_dictionary()` on the dictionary path so any key type works (match dictionary values, then `take`); scalar input still uses the precompiled `Regex`.
- Return `internal_err!` for non-string inputs on both the array and scalar paths (invalid plan rather than a user-facing execution error).
- Simplify `is_match` to `iter` / `map` / `collect`, matching the `process_parse_url` style.
- Keep the Spark-compatibility caveat on the struct rustdoc; keep the "why not Arrow kernel" note only on`is_match`.

## Performance notes

The RLike benchmark added by #5598 (originally requested in #5246) was run during review against the previous and current is_match loops.

Both binaries were built from the same source tree with only `rlike.rs` swapped, then measured in an interleaved run at 65,536 rows:

| Input | Previous loop | This PR | Difference |
|---|---:|---:|---:|
| `no_nulls` | 2.820 ms | 2.924 ms | +3.7% |
| `sparse` | 2.640 ms | 2.687 ms | +1.8% |
| `all_null` | 142.4 µs | 74.2 µs | 1.9× faster |

The iterator-based implementation has a small cost for non-null inputs and a substantial improvement for all-null input. This is considered a reasonable trade-off for the broader layout and dictionary support and the removal of panic paths.

These are local, directional measurements rather than CI performance thresholds. Interleaving the two pre-built binaries was used because sequential Criterion baseline comparisons were too noisy to resolve differences of this size reliably.

- Unit coverage for the `Utf8`, `LargeUtf8`, and `Utf8View` layouts.
- Direct and dictionary `Utf8View` cases with values longer than 12 bytes, exercising the out-of-line representation.
- An all-valid input assertion confirming that the result remains null-buffer-free.
- Dictionary coverage for:
  - `Dictionary(Int32, Utf8)`
  - `Dictionary(Int32, Utf8View)`
  - `Dictionary(Int8, Utf8)`
  - `Dictionary(UInt64, Utf8)`
  - nulls in dictionary values
  - a dictionary sliced with a non-zero offset
- `cargo fmt --check`
- `cargo test -p datafusion-comet-spark-expr --lib rlike`
- `cargo clippy -p datafusion-comet-spark-expr --lib --tests -- -D warnings`

### DIFF
--- a/native/spark-expr/src/predicate_funcs/rlike.rs
+++ b/native/spark-expr/src/predicate_funcs/rlike.rs
@@ -18,10 +18,15 @@
 use crate::SparkError;
 use arrow::array::builder::BooleanBuilder;
 use arrow::array::types::Int32Type;
-use arrow::array::{Array, BooleanArray, DictionaryArray, RecordBatch, StringArray};
+use arrow::array::{
+    Array, ArrayAccessor, ArrayRef, BooleanArray, RecordBatch, StringArrayType,
+};
 use arrow::compute::take;
 use arrow::datatypes::{DataType, Schema};
-use datafusion::common::{internal_err, Result, ScalarValue};
+use datafusion::common::cast::{
+    as_dictionary_array, as_large_string_array, as_string_array, as_string_view_array,
+};
+use datafusion::common::{exec_err, internal_err, Result, ScalarValue};
 use datafusion::physical_expr::PhysicalExpr;
 use datafusion::physical_plan::ColumnarValue;
 use regex::Regex;
@@ -37,6 +42,11 @@ use std::sync::Arc;
 /// regular expression engine, which are documented at:
 ///
 /// https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html
+///
+/// Array matching keeps the plan-time compiled [`Regex`] and loops over Utf8 /
+/// LargeUtf8 / Utf8View inputs. Arrow's `regexp_is_match(_scalar)` was considered
+/// but recompiles the pattern per batch; criterion benches showed regressions on
+/// common patterns such as character classes (see issue #5102).
 #[derive(Debug)]
 pub struct RLike {
     child: Arc<dyn PhysicalExpr>,
@@ -71,7 +81,14 @@ impl RLike {
         })
     }
 
-    fn is_match(&self, inputs: &StringArray) -> BooleanArray {
+    /// Match the pre-compiled pattern against a string array of any Arrow string layout.
+    ///
+    /// Keeps the plan-time compiled [`Regex`] rather than calling Arrow's
+    /// `regexp_is_match(_scalar)`, which recompiles the pattern on every batch.
+    fn is_match<'a, S>(&'a self, inputs: &'a S) -> BooleanArray
+    where
+        &'a S: StringArrayType<'a>,
+    {
         let mut builder = BooleanBuilder::with_capacity(inputs.len());
         if inputs.is_nullable() {
             for i in 0..inputs.len() {
@@ -87,6 +104,15 @@ impl RLike {
             }
         }
         builder.finish()
+    }
+
+    fn is_match_array(&self, array: &ArrayRef) -> Result<BooleanArray> {
+        match array.data_type() {
+            DataType::Utf8 => Ok(self.is_match(as_string_array(array)?)),
+            DataType::LargeUtf8 => Ok(self.is_match(as_large_string_array(array)?)),
+            DataType::Utf8View => Ok(self.is_match(as_string_view_array(array)?)),
+            other => exec_err!("RLike requires string type for input, got {other:?}"),
+        }
     }
 }
 
@@ -111,29 +137,19 @@ impl PhysicalExpr for RLike {
 
     fn evaluate(&self, batch: &RecordBatch) -> Result<ColumnarValue> {
         match self.child.evaluate(batch)? {
-            ColumnarValue::Array(array) if array.as_any().is::<DictionaryArray<Int32Type>>() => {
-                let dict_array = array
-                    .as_any()
-                    .downcast_ref::<DictionaryArray<Int32Type>>()
-                    .expect("dict array");
-                let dict_values = dict_array
-                    .values()
-                    .as_any()
-                    .downcast_ref::<StringArray>()
-                    .expect("strings");
+            ColumnarValue::Array(array)
+                if matches!(array.data_type(), DataType::Dictionary(_, _)) =>
+            {
+                let dict_array = as_dictionary_array::<Int32Type>(&array)?;
                 // evaluate the regexp pattern against the dictionary values
-                let new_values = self.is_match(dict_values);
+                let new_values = self.is_match_array(dict_array.values())?;
                 // convert to conventional (not dictionary-encoded) array
                 let result = take(&new_values, dict_array.keys(), None)?;
                 Ok(ColumnarValue::Array(result))
             }
             ColumnarValue::Array(array) => {
-                let inputs = array
-                    .as_any()
-                    .downcast_ref::<StringArray>()
-                    .expect("string array");
-                let array = self.is_match(inputs);
-                Ok(ColumnarValue::Array(Arc::new(array)))
+                let result = self.is_match_array(&array)?;
+                Ok(ColumnarValue::Array(Arc::new(result)))
             }
             ColumnarValue::Scalar(scalar) => {
                 if scalar.is_null() {
@@ -180,7 +196,22 @@ impl PhysicalExpr for RLike {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use datafusion::physical_expr::expressions::Literal;
+    use arrow::array::{LargeStringArray, StringArray, StringViewArray};
+    use arrow::datatypes::Field;
+    use datafusion::physical_expr::expressions::{Column, Literal};
+
+    fn assert_bool_array_rose_null(result: ColumnarValue) {
+        let ColumnarValue::Array(arr) = result else {
+            panic!("expected array result");
+        };
+        let bools = arr
+            .as_any()
+            .downcast_ref::<BooleanArray>()
+            .expect("boolean array");
+        assert_eq!(bools.len(), 2);
+        assert!(bools.value(0));
+        assert!(bools.is_null(1));
+    }
 
     #[test]
     fn test_rlike_scalar_string_variants() {
@@ -224,5 +255,57 @@ mod tests {
 
         let result = expr.evaluate(&RecordBatch::new_empty(Arc::new(Schema::empty())));
         assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_rlike_utf8_array() {
+        let schema = Arc::new(Schema::new(vec![Field::new("s", DataType::Utf8, true)]));
+        let batch = RecordBatch::try_new(
+            Arc::clone(&schema),
+            vec![Arc::new(StringArray::from(vec![Some("Rose"), None]))],
+        )
+        .unwrap();
+
+        let expr = RLike::try_new(Arc::new(Column::new("s", 0)), "R[a-z]+").unwrap();
+        assert_bool_array_rose_null(expr.evaluate(&batch).unwrap());
+    }
+
+    #[test]
+    fn test_rlike_large_utf8_array() {
+        let schema = Arc::new(Schema::new(vec![Field::new("s", DataType::LargeUtf8, true)]));
+        let batch = RecordBatch::try_new(
+            Arc::clone(&schema),
+            vec![Arc::new(LargeStringArray::from(vec![Some("Rose"), None]))],
+        )
+        .unwrap();
+
+        let expr = RLike::try_new(Arc::new(Column::new("s", 0)), "R[a-z]+").unwrap();
+        assert_bool_array_rose_null(expr.evaluate(&batch).unwrap());
+    }
+
+    #[test]
+    fn test_rlike_utf8_view_array() {
+        let schema = Arc::new(Schema::new(vec![Field::new("s", DataType::Utf8View, true)]));
+        let batch = RecordBatch::try_new(
+            Arc::clone(&schema),
+            vec![Arc::new(StringViewArray::from(vec![Some("Rose"), None]))],
+        )
+        .unwrap();
+
+        let expr = RLike::try_new(Arc::new(Column::new("s", 0)), "R[a-z]+").unwrap();
+        assert_bool_array_rose_null(expr.evaluate(&batch).unwrap());
+    }
+
+    #[test]
+    fn test_rlike_array_non_string_error() {
+        let schema = Arc::new(Schema::new(vec![Field::new("b", DataType::Boolean, true)]));
+        let batch = RecordBatch::try_new(
+            Arc::clone(&schema),
+            vec![Arc::new(BooleanArray::from(vec![Some(true), None]))],
+        )
+        .unwrap();
+
+        let expr = RLike::try_new(Arc::new(Column::new("b", 0)), "R[a-z]+").unwrap();
+        assert!(expr.evaluate(&batch).is_err());
     }
 }
--- a/native/spark-expr/src/predicate_funcs/rlike.rs
+++ b/native/spark-expr/src/predicate_funcs/rlike.rs
@@ -18,9 +18,7 @@
 use crate::SparkError;
 use arrow::array::builder::BooleanBuilder;
 use arrow::array::types::Int32Type;
-use arrow::array::{
-    Array, ArrayAccessor, ArrayRef, BooleanArray, RecordBatch, StringArrayType,
-};
+use arrow::array::{Array, ArrayAccessor, ArrayRef, BooleanArray, RecordBatch, StringArrayType};
 use arrow::compute::take;
 use arrow::datatypes::{DataType, Schema};
 use datafusion::common::cast::{
@@ -272,7 +270,11 @@ mod tests {
 
     #[test]
     fn test_rlike_large_utf8_array() {
-        let schema = Arc::new(Schema::new(vec![Field::new("s", DataType::LargeUtf8, true)]));
+        let schema = Arc::new(Schema::new(vec![Field::new(
+            "s",
+            DataType::LargeUtf8,
+            true,
+        )]));
         let batch = RecordBatch::try_new(
             Arc::clone(&schema),
             vec![Arc::new(LargeStringArray::from(vec![Some("Rose"), None]))],
--- a/native/spark-expr/src/predicate_funcs/rlike.rs
+++ b/native/spark-expr/src/predicate_funcs/rlike.rs
@@ -36,7 +36,6 @@ use std::sync::Arc;
 /// regular expression engine, which are documented at:
 ///
 /// https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html
-///
 #[derive(Debug)]
 pub struct RLike {
     child: Arc<dyn PhysicalExpr>,
@@ -304,6 +303,9 @@ mod tests {
 
         let utf8_values: ArrayRef = Arc::new(StringArray::from(vec!["Rose", "Daisy"]));
         let utf8_view_values: ArrayRef = Arc::new(StringViewArray::from(vec!["Rose", "Daisy"]));
+        // Null in dictionary values (keys all valid): is_match emits null, take carries it.
+        let utf8_values_with_null: ArrayRef =
+            Arc::new(StringArray::from(vec![Some("Rose"), None, Some("Daisy")]));
 
         let cases: Vec<(DataType, ArrayRef)> = vec![
             (
@@ -325,6 +327,13 @@ mod tests {
                 Arc::new(DictionaryArray::<Int8Type>::new(
                     Int8Array::from(vec![Some(0), None, Some(1)]),
                     Arc::clone(&utf8_values),
+                )),
+            ),
+            (
+                DataType::Dictionary(Box::new(DataType::Int32), Box::new(DataType::Utf8)),
+                Arc::new(DictionaryArray::<Int32Type>::new(
+                    Int32Array::from(vec![Some(0), Some(1), Some(2)]),
+                    utf8_values_with_null,
                 )),
             ),
         ];
--- a/native/spark-expr/src/predicate_funcs/rlike.rs
+++ b/native/spark-expr/src/predicate_funcs/rlike.rs
@@ -16,15 +16,11 @@
 // under the License.
 
 use crate::SparkError;
-use arrow::array::builder::BooleanBuilder;
-use arrow::array::types::Int32Type;
-use arrow::array::{Array, ArrayAccessor, ArrayRef, BooleanArray, RecordBatch, StringArrayType};
+use arrow::array::{Array, ArrayRef, AsArray, BooleanArray, RecordBatch, StringArrayType};
 use arrow::compute::take;
 use arrow::datatypes::{DataType, Schema};
-use datafusion::common::cast::{
-    as_dictionary_array, as_large_string_array, as_string_array, as_string_view_array,
-};
-use datafusion::common::{exec_err, internal_err, Result, ScalarValue};
+use datafusion::common::cast::{as_large_string_array, as_string_array, as_string_view_array};
+use datafusion::common::{internal_err, Result, ScalarValue};
 use datafusion::physical_expr::PhysicalExpr;
 use datafusion::physical_plan::ColumnarValue;
 use regex::Regex;
@@ -41,10 +37,6 @@ use std::sync::Arc;
 ///
 /// https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html
 ///
-/// Array matching keeps the plan-time compiled [`Regex`] and loops over Utf8 /
-/// LargeUtf8 / Utf8View inputs. Arrow's `regexp_is_match(_scalar)` was considered
-/// but recompiles the pattern per batch; criterion benches showed regressions on
-/// common patterns such as character classes (see issue #5102).
 #[derive(Debug)]
 pub struct RLike {
     child: Arc<dyn PhysicalExpr>,
@@ -83,25 +75,14 @@ impl RLike {
     ///
     /// Keeps the plan-time compiled [`Regex`] rather than calling Arrow's
     /// `regexp_is_match(_scalar)`, which recompiles the pattern on every batch.
-    fn is_match<'a, S>(&'a self, inputs: &'a S) -> BooleanArray
+    fn is_match<'a, S>(&self, inputs: &'a S) -> BooleanArray
     where
         &'a S: StringArrayType<'a>,
     {
-        let mut builder = BooleanBuilder::with_capacity(inputs.len());
-        if inputs.is_nullable() {
-            for i in 0..inputs.len() {
-                if inputs.is_null(i) {
-                    builder.append_null();
-                } else {
-                    builder.append_value(self.pattern.is_match(inputs.value(i)));
-                }
-            }
-        } else {
-            for i in 0..inputs.len() {
-                builder.append_value(self.pattern.is_match(inputs.value(i)));
-            }
-        }
-        builder.finish()
+        inputs
+            .iter()
+            .map(|v| v.map(|s| self.pattern.is_match(s)))
+            .collect()
     }
 
     fn is_match_array(&self, array: &ArrayRef) -> Result<BooleanArray> {
@@ -109,7 +90,9 @@ impl RLike {
             DataType::Utf8 => Ok(self.is_match(as_string_array(array)?)),
             DataType::LargeUtf8 => Ok(self.is_match(as_large_string_array(array)?)),
             DataType::Utf8View => Ok(self.is_match(as_string_view_array(array)?)),
-            other => exec_err!("RLike requires string type for input, got {other:?}"),
+            other => {
+                internal_err!("RLike requires string type for input, got {other:?}")
+            }
         }
     }
 }
@@ -138,7 +121,7 @@ impl PhysicalExpr for RLike {
             ColumnarValue::Array(array)
                 if matches!(array.data_type(), DataType::Dictionary(_, _)) =>
             {
-                let dict_array = as_dictionary_array::<Int32Type>(&array)?;
+                let dict_array = array.as_any_dictionary();
                 // evaluate the regexp pattern against the dictionary values
                 let new_values = self.is_match_array(dict_array.values())?;
                 // convert to conventional (not dictionary-encoded) array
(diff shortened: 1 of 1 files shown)


native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -111,29 +135,19 @@ impl PhysicalExpr for RLike {
...
-                    .expect("dict array");
-                let dict_values = dict_array
-                    .values()
-                    .as_any()
-                    .downcast_ref::<StringArray>()
-                    .expect("strings");
+            ColumnarValue::Array(array)
+                if matches!(array.data_type(), DataType::Dictionary(_, _)) =>
+            {
+                let dict_array = as_dictionary_array::<Int32Type>(&array)?;
```
andygrove: The guard above now matches `DataType::Dictionary(_, _)` for any key type, but the body still calls `as_dictionary_array::<Int32Type>`, so a non-`Int32`-keyed dictionary trades the old panic for an `Internal` error rather than actually being handled. I checked this on your branch and a `Dictionary(Int8, Utf8)` input gives:

```
Err(Internal("could not cast array of type Dictionary(Int8, Utf8) to ... DictionaryArray<Int32Type>"))
```

I tried `array.as_any_dictionary()` instead and it handles every key type. `arrow::compute::take` already accepts `indices: &dyn Array` in arrow 58, so `take(&new_values, dict_array.keys(), None)` works unchanged, and the `Int32Type` and `as_dictionary_array` imports become unnecessary. With that swap the `Int8`-keyed case returns the correct `[true, null, false]`. Would that work for you here?

Worth noting the good news too: `Dictionary(Int32, Utf8View)` and `Dictionary(Int32, LargeUtf8)` do work correctly with your change.

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -88,6 +103,15 @@ impl RLike {
         }
         builder.finish()
     }
+
+    fn is_match_array(&self, array: &ArrayRef) -> Result<BooleanArray> {
+        match array.data_type() {
+            DataType::Utf8 => Ok(self.is_match(as_string_array(array)?)),
+            DataType::LargeUtf8 => Ok(self.is_match(as_large_string_array(array)?)),
+            DataType::Utf8View => Ok(self.is_match(as_string_view_array(array)?)),
+            other => exec_err!("RLike requires string type for input, got {other:?}"),
```
andygrove: Small consistency question. This uses `exec_err!` while the scalar branch below uses `internal_err!` for the same "not a string" condition. Since a non-string child on `RLike` means the planner produced something invalid rather than the user doing something wrong, would `internal_err!` be the better fit for both?

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -37,6 +40,11 @@ use std::sync::Arc;
 /// regular expression engine, which are documented at:
 ///
 /// https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html
+///
+/// Array matching keeps the plan-time compiled [`Regex`] and loops over Utf8 /
```
andygrove: This mentions criterion benches showing regressions on patterns like character classes and points to #5102, but the issue only says per-batch compilation "is a measurable regression" without any numbers, and I do not see a bench added under `native/spark-expr/benches/`. Do you have the numbers handy? Posting them in the PR description, or adding the bench alongside the existing ones like `regexp_extract.rs`, would make this a much stronger reference for the next person who wonders why the kernel was not used.

The same rationale also appears twice, here and again on `is_match`. This struct-level rustdoc is where the Spark-compatibility caveat lives, so an explanation of a rejected implementation alternative reads a little out of place. Keeping the single copy on `is_match` would be cleaner.

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -225,4 +254,60 @@ mod tests {
         let result = expr.evaluate(&RecordBatch::new_empty(Arc::new(Schema::empty())));
         assert!(result.is_err());
     }
+
+    #[test]
+    fn test_rlike_utf8_array() {
```
andygrove: Thanks for adding these. A few gaps worth closing.

The dictionary path is the biggest change in the diff and it does not have a test yet. Could you add coverage there? A `Dictionary(Int32, Utf8)` case plus a `Dictionary(Int32, Utf8View)` case would exercise both the layout dispatch and the `take`, and a non-`Int32` key case would pin down whatever behavior you settle on for the comment above.

`Array::is_nullable()` is `logical_null_count() != 0` in arrow 58, so the `else` branch of `is_match` only runs when the array has no nulls. All three of these tests include a null, so that branch never gets hit. Adding one all-non-null array would close it.

These three tests are also near-identical. `test_rlike_scalar_string_variants` right above loops over the layouts, so it might be nice to follow that shape here too.

andygrove (commented):
Thanks for taking this on. Keeping the plan-time compiled `Regex` and just fixing the downcast robustness is the right call out of the two options in #5102, and the layout dispatch reads well.

I ran the native checks locally on this branch. `cargo test -p datafusion-comet-spark-expr --lib rlike` passes and `cargo clippy -p datafusion-comet-spark-expr --lib --tests -- -D warnings` is clean. CI itself has not run yet, all three workflows are sitting at the first-time-contributor approval gate, so I will get those approved.

I have left some comments inline. The main one is that the dictionary arm is still restricted to `Int32` keys, so that shape trades a panic for an error rather than being handled.

One thing that would help in the PR description: which of these layouts is reachable today? `planner.rs:3387` casts `Utf8View` results back to `Utf8` with the comment "Comet does not yet support view types", and I could not find a plan that feeds a view array into `RLike` right now. The hardening is still worth doing and the issue asked for it, but if there is a known reachable case it would be good to name it. If one of these layouts is reachable through a config, an end-to-end query added to `spark/src/test/resources/sql-tests/expressions/string/rlike_rust.sql` would be a nice complement to the Rust unit tests.

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -71,7 +79,14 @@ impl RLike {
...
 
-    fn is_match(&self, inputs: &StringArray) -> BooleanArray {
+    /// Match the pre-compiled pattern against a string array of any Arrow string layout.
+    ///
+    /// Keeps the plan-time compiled [`Regex`] rather than calling Arrow's
+    /// `regexp_is_match(_scalar)`, which recompiles the pattern on every batch.
+    fn is_match<'a, S>(&'a self, inputs: &'a S) -> BooleanArray
+    where
+        &'a S: StringArrayType<'a>,
+    {
```
0lai0, line 80: Optional and non-blocking
This follows on from the comment by @andygrove below about `is_nullable()`

Since `is_nullable()` is `logical_null_count() != 0`, the else branch only runs on an array with no nulls, and `StringArrayType` gives us `iter()`. `BooleanArray` implements `FromIterator<Option<bool>>`, so both branches collapse into:

```
fn is_match<'a, S>(&self, inputs: &'a S) -> BooleanArray
where
    &'a S: StringArrayType<'a>,
{
    inputs.iter().map(|v| v.map(|s| self.pattern.is_match(s))).collect()
}
```

The uncovered branch stops existing rather than needing a test, and null handling no longer depends on `is_nullable()`. 
`process_parse_url `in `url_funcs/parse_url.rs` already uses the same `StringArrayType` bound and the same iter/collect shape.

Worth noting `ArrayIter`'s docs call interleaved null-mask handling suboptimal, but relative to `Regex::is_match` I would expect that to be noise 
ref: https://docs.rs/arrow/latest/arrow/array/struct.ArrayIter.html

Also, `&'a self` ties the borrow of `self` to the input lifetime and nothing is borrowed out of it, so plain `&self` would do

0lai0 (commented):
Thanks @sam-1112 for picking this up! Congrats on your first PR.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Good catch — thanks for checking the non-Int32 key case.

You're right: widening the guard to `Dictionary(_, _)` while still casting with `as_dictionary_array::<Int32Type>` only turned the old panic into an `Internal` error for other key types. I've switched to `array.as_any_dictionary()` and `take(&new_values, dict_array.keys(), None)` as you suggested, and added a `Dictionary(Int8, Utf8)` test that expects `[true, null, false]`.

Glad to hear `Dictionary(Int32, Utf8View)` / `LargeUtf8` already looked good — those stay covered as well.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Good point — agreed. A non-string child here means an invalid plan rather than a user-facing execution error, so I've switched the array path to `internal_err!` as well to match the scalar branch.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Agreed on the docs cleanup. I don't have the criterion numbers saved anymore — I'll remove the strong "benches showed regressions" claim from the struct docs (or soften it) and keep the compile-per-batch rationale only on `is_match`. Happy to add a proper bench in a follow-up if useful.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Thanks for the review — agreed on all three points.

I've added dictionary coverage for `Dictionary(Int32, Utf8)`, `Dictionary(Int32, Utf8View)`, and a non-Int32 key case (`Dictionary(Int8, Utf8)`) so the layout dispatch, `take`, and key-type handling are all exercised. I also collapsed the three near-identical layout tests into a single looped test (same shape as `test_rlike_scalar_string_variants`), and added an all-non-null array case.

One related note: `is_match` no longer has the `is_nullable()` if/else — it now uses `iter().map(...).collect()` — so the no-nulls test is mainly a regression check rather than covering a separate branch. Happy to adjust further if you'd like anything else pinned down.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Good suggestion — I've collapsed `is_match` to the `iter`/`map`/`collect` form (and switched to `&self`), matching `process_parse_url`. That removes the `is_nullable()` branch entirely, so the uncovered else path no longer exists.

sam-1112:
Thanks for the review and for kicking CI past the first-time-contributor gate.

On the dictionary arm: agreed — I've switched to `as_any_dictionary()` so non-`Int32` keys are handled, and added unit coverage for `Dictionary(Int32, Utf8)`, `Dictionary(Int32, Utf8View)`, and `Dictionary(Int8, Utf8)`.

On reachability: I don't currently have a known plan that feeds `Utf8View` (or `LargeUtf8`) into native `RLike`. As you noted, `planner.rs` still casts `Utf8View` results back to `Utf8` because Comet does not yet support view types, and the existing `rlike_rust.sql` path uses plain Spark `string` / Parquet, which shows up as `Utf8`. So this change is mainly hardening for the layouts #5102 called out, plus dictionary key-type correctness, rather than fixing a query I can reproduce end-to-end today. If we later find a reachable config/path, I'm happy to add an e2e case to `rlike_rust.sql` in a follow-up.

I've also updated the PR for the other inline comments (`internal_err!`, docs cleanup, test consolidation, and the `iter`/`collect` simplification).

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -37,6 +36,7 @@ use std::sync::Arc;
 /// regular expression engine, which are documented at:
 ///
 /// https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html
+///
```
andygrove: Looks like removing the perf paragraph left a bare `///` behind just above `#[derive(Debug)]`. Could you drop that line? It is the only noise left in an otherwise tidy diff.

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -225,4 +246,107 @@ mod tests {
...
+
+    #[test]
+    fn test_rlike_dictionary_arrays() {
+        let pattern = "R[a-z]+";
+        let expected = [Some(true), None, Some(false)];
+
+        let utf8_values: ArrayRef = Arc::new(StringArray::from(vec!["Rose", "Daisy"]));
+        let utf8_view_values: ArrayRef = Arc::new(StringViewArray::from(vec!["Rose", "Daisy"]));
+
+        let cases: Vec<(DataType, ArrayRef)> = vec![
```
andygrove, line 310: All three dictionary cases put the null in the keys. A null in the dictionary *values* takes a different route to the same place: `is_match` emits it and `take` carries it through, rather than `take` producing it from a null key. I tried it on your branch and it gives the right answer, so this is about pinning the behavior rather than a bug. Since the dictionary arm is the bulk of this diff, would you mind adding a case with values like `[Some("Rose"), None, Some("Daisy")]` and all-valid keys?

andygrove (commented):
Thanks for the quick turnaround. All four of my earlier comments and 0lai0's `iter`/`collect` suggestion are addressed, and I verified each one on your branch rather than just reading the replies.

`Dictionary(Int8, Utf8)` now returns the correct `[true, null, false]` instead of the `Internal` cast error. The `internal_err!` switch is consistent across both paths. And the one remaining rationale on `is_match` is accurate. I checked `arrow-string-58.3.0/src/regexp.rs:193` and `regexp_is_match_scalar` does call `Regex::new` inside the function body, so it really does recompile per call.

I also probed seven edge cases beyond your tests: sliced string arrays, sliced dictionaries, zero-length arrays, nulls inside dictionary values, nested `Dictionary(Int32, Dictionary(...))`, a dictionary scalar, and the null buffer on the result of a non-nullable input. Every one of them either gives the right answer or a clean `internal_err`. I could not find a surviving panic. One thing I had expected to be a problem turned out not to be: I thought `collect()` into `BooleanArray` would always materialize a null buffer, making the non-nullable case heavier than the old builder produced, but `nulls()` comes back `None`, so the rewrite is a clean win with no change to the output shape.

On CI, note that your latest push reset the first-time-contributor gate, so all three workflows are sitting at `action_required` again and nothing has actually run yet. I ran the native checks locally on `4874f48` in the meantime: `cargo test -p datafusion-comet-spark-expr --lib rlike` passes all 6, `cargo clippy -p datafusion-comet-spark-expr --lib --tests -- -D warnings` is clean, and `cargo fmt --check` is clean.

Two small things inline, plus one item below.

**On the benchmark.** Thanks for softening the docs claim. The gap now is that `is_match` documents why we hand-roll the loop while nothing measures it, and this PR rewrote that loop. I do not think that blocks the change. I would expect the null-check removal to be lost in the noise next to `Regex::is_match`. But I do not want it to just evaporate, so could you file an issue for adding `native/spark-expr/benches/rlike.rs` alongside the existing `regexp_extract.rs` and link it here?

**On reachability.** Following up on my own question from last round, I went looking and I agree with you. `schema_force_view_types` is deliberately left out of the allowlist in `native/core/src/parquet/parquet_exec.rs`, Comet always passes its own Spark-derived schema where strings are `Utf8`, and `planner.rs:3387` casts `Utf8View` back to `Utf8`. So there is no plan today that feeds a view array into `RLike`, and I am dropping my request for an end-to-end case in `rlike_rust.sql`. The reachable dictionary shape is `Dictionary(Int32, Utf8)` and that is unit-tested here. Recording it so the next person does not have to re-derive it.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Done — added a dictionary case with values `[Some("Rose"), None, Some("Daisy")]` and all-valid keys `[0, 1, 2]` in `test_rlike_dictionary_arrays` (commit c799e5bd).

sam-1112:
Thanks for digging into those edge cases, it's really helpful.

I have pushed the last bits: (1) removed the stray `///`; (2) added a dictionary case with nulls in the values (all-valid keys). Also opened #5246 for the RLike Criterion bench.

I also agree on the Utf8View reachability point; I'll skip the e2e case in `rlike_rust.sql` for now.

andygrove:

This is good hardening. Replacing the hard `downcast_ref::<StringArray>().expect(...)` with a match that returns `internal_err!`, and switching the dictionary path from `DictionaryArray<Int32Type>` to `as_any_dictionary()` so any key type works, both remove real panic paths. Keeping the plan-time compiled `Regex` rather than moving to Arrow's kernel is the right call and the rustdoc explaining why is worth having.

Two things.

**`collect()` now always produces a null buffer**

The old code had two branches, and the non-nullable one used `append_value` throughout, so `BooleanBuilder::finish` produced an array with no null buffer at all. The new `inputs.iter().map(|v| v.map(...)).collect()` goes through `FromIterator<Option<bool>>`, which as far as I can tell allocates a validity buffer regardless.

For an all-valid input that is an extra allocation of `len / 8` bytes per batch, and more importantly some downstream Arrow kernels take a faster path when `nulls()` is `None` rather than when `null_count() == 0`. Since `RLike` output feeds directly into filters, that could matter.

Could `test_rlike_string_array_no_nulls` assert `result.nulls().is_none()` rather than just checking the values? If it turns out a buffer is allocated, keeping the two-branch form for the non-nullable case would preserve the old behavior while still getting the layout generality from `StringArrayType`.

**Reachability**

The description is honest that `LargeUtf8` and `Utf8View` may not be reachable end to end today. That is fine for hardening, but it does mean the new code paths are exercised only by the Rust unit tests. Is it worth adding a `#[test]` that runs a `Utf8View` input through the dictionary path as well, since that is the combination with the most moving parts (any key type, view values, `take` over the result)? The current dictionary tests use `Int32Type` and `Int8Type` keys but I did not see one with view values.

sam-1112:
Thanks for the follow-up.

On the null-buffer point: as you found in your earlier review of `4874f48`, collecting into `BooleanArray` leaves `nulls()` as `None` for an all-valid input. Therefore, this rewrite does not introduce the extra validity allocation or affect the downstream null fast path. I’ve now added an explicit `assert!(arr.nulls().is_none())` to `test_rlike_string_array_no_nulls` to preserve that property as a regression check.

On the dictionary `Utf8View` path: this is already covered in `test_rlike_dictionary_arrays` by the `Dictionary(Int32, Utf8View)` case with `StringViewArray` values, which exercises `as_any_dictionary()` → `is_match_array(Utf8View)` → `take`. The separate `Dictionary(Int8, Utf8)` case covers non-`Int32` keys, so the two dimensions are independently covered. I’m happy to add an `Int8 + Utf8View` cross-product case as well if you would prefer explicit coverage of that combination.

sam-1112:
Hi @andygrove, just a gentle ping when you get a chance. I’ve added the null-buffer assertion, and the existing dictionary test already covers the `Dictionary(Int32, Utf8View)` path. I think that addresses both points from your latest review. Thanks!

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -225,4 +245,122 @@ mod tests {
...
+                ])),
+            ),
+            (
+                DataType::Utf8View,
+                Arc::new(StringViewArray::from(vec![
+                    Some("Rose"),
+                    None,
+                    Some("Daisy"),
+                ])),
+            ),
```
andygrove, line 273: All the `Utf8View` values in these tests are `"Rose"` and `"Daisy"`, which are both 12 bytes or fewer, so they live inline in the view struct. The representation that actually differs from `Utf8` is the one for strings longer than 12 bytes, where the view holds a `(len, prefix, buffer_index, offset)` pointer into a separate data buffer, and that path is never constructed here.

Since `Utf8View` support is the headline of this PR and there is no end-to-end coverage behind it, could you make one of these values longer than 12 bytes? I checked that a 42-byte value passes on your branch and panics on `main`, so it is in scope. The `Dictionary(Int32, Utf8View)` case below would benefit from the same thing.

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -225,4 +245,122 @@ mod tests {
...
+                    Arc::clone(&utf8_view_values),
+                )),
+            ),
+            (
+                DataType::Dictionary(Box::new(DataType::Int8), Box::new(DataType::Utf8)),
+                Arc::new(DictionaryArray::<Int8Type>::new(
+                    Int8Array::from(vec![Some(0), None, Some(1)]),
+                    Arc::clone(&utf8_values),
+                )),
+            ),
```
andygrove, line 344: Two more rows here that would each be one line. `Dictionary(UInt64, Utf8)`, since `as_any_dictionary()` covers the unsigned key types too and both of the current cases are signed. And a sliced dictionary, since that is the shape that turns up after a filter or a limit rather than the contiguous one. I confirmed both panic on `main` and pass on your branch.

native/spark-expr/src/predicate_funcs/rlike.rs
```diff
@@ -225,4 +245,122 @@ mod tests {
...
+            Arc::clone(&schema),
+            vec![Arc::new(StringArray::from(vec!["Rose", "Daisy"]))],
+        )
+        .unwrap();
+
+        let expr = RLike::try_new(Arc::new(Column::new("s", 0)), "R[a-z]+").unwrap();
+        let ColumnarValue::Array(arr) = expr.evaluate(&batch).unwrap() else {
+            panic!("expected array result");
+        };
+        // All-valid input must not allocate a null buffer (filter fast path).
```
andygrove: The assertion is worth keeping, but the reason in the comment is mine from the last round and it was wrong. `FilterBuilder::new_with_count` in arrow-select 59.3.0 dispatches on `filter.null_count()` rather than on `nulls().is_some()`, so a present-but-all-valid null buffer would not have hit `prep_null_mask_filter` either way.

What the assertion really pins down is the output shape and the `len / 8` bytes per batch, and that `BooleanBuilder`'s lazy null buffer keeps behaving this way across arrow upgrades. Could you reword it to say that instead of "filter fast path"?

andygrove (commented):
Thanks for the ping, and sorry for the wait.

I rebuilt this from scratch rather than re-reading the thread, since my last round was LLM-generated and got the null-buffer point wrong. Everything I can check locally looks good.

I appended 14 probe cases to the test module and ran the identical module against both your branch and `main`. All 14 pass on your branch and 6 of them panic on `main`: sliced `Utf8View`, `Dictionary(Int32, LargeUtf8)`, `Dictionary(UInt64, Utf8)`, nested `Dictionary(Int32, Dictionary(Int8, Utf8))`, `RunEndEncoded(Int32, Utf8)`, and `Utf8View` with values longer than 12 bytes. Each of those either returns the right answer or a clean `internal_err` on your branch. I could not find a case where `main` returns a result and your branch errors, which is the thing I most wanted to rule out, and it follows from the new dispatch being strictly more permissive than the old downcasts.

Two other things worth recording. Your branch is 200 commits behind `main` now, but `rlike.rs` has not been touched on `main` since your branch point, the merge is clean, and your `rlike.rs` compiles unmodified against current `main`'s DataFusion 55 even though your branch is on 54. Separately, I built your native code on top of current `main`'s JVM and ran the existing `rlike_rust.sql` fixture through `CometSqlFileTestSuite`, and it passes.

On the null-buffer question from my last round, you were right and I was wrong. arrow-array 59.3.0's `FromIterator for BooleanArray` routes through `BooleanBuilder`, whose `NullBufferBuilder` is lazy, so `nulls()` really is `None` for all-valid input. Worth noting that this is version-dependent, older arrow allocated the buffer unconditionally, which is what makes the assertion you added genuinely useful rather than redundant.

I also re-checked the rustdoc claim on `is_match` against the pinned arrow rather than the version I looked at last time. `regexp_is_match_scalar` still calls `Regex::new` inside the function body at `arrow-string-59.3.0/src/regexp.rs:193`, and the `HashMap` cache there is call-local, so it really does recompile per batch. The comment is accurate and should stay.

**On the benchmark.** `native/spark-expr/benches/rlike.rs` landed on `main` from #5246 while this PR was open, so the bench you filed for now exists and this PR can carry numbers. I ran it, since this PR rewrote the loop it measures. Building both binaries from the same tree with only `rlike.rs` swapped and running them interleaved, at 65536 rows: `no_nulls` 2.820 ms to 2.924 ms (+3.7%), `sparse` 2.640 ms to 2.687 ms (+1.8%), `all_null` 142.4 us to 74.2 us (1.9x faster). The gradient tracks null density, which is what you would expect from `ArrayIter` doing a per-row `is_null` that the old hoisted `is_nullable()` branch skipped.

I am not asking you to add the branch back. Under 5% on an expression that is off by default, against a 1.9x win on null-heavy data and the robustness this buys, reads like a good trade to me. Adding the numbers to the PR description would be worth it though. One caveat if you reproduce this: criterion's sequential `--save-baseline`/`--baseline` gave me plus or minus 17% "p < 0.05" swings for `main` measured against itself, so it cannot resolve an effect this small. Interleaving two pre-built binaries got the confidence intervals to about 0.1%.

**On CI.** `gh pr checks` still reports no checks at all on this branch. Each push resets the first-time-contributor gate, so nothing has ever actually run here. I will get those approved again. Everything above is local and on Spark 4.1 / Scala 2.13 only, so the cross-platform and Spark 3.x matrices are still unverified.

Three small things inline.

sam-1112:
Thanks for the thorough re-review and for checking the additional edge cases against both this branch and main. The compatibility checks against current main and the benchmark measurements were especially helpful.

I’ve addressed the three inline comments:
- Updated both the direct and dictionary Utf8View cases to use "Rhododendrons" (13 bytes), so they exercise the out-of-line view representation.
- Added coverage for Dictionary(UInt64, Utf8) and a sliced dictionary with a non-zero offset.
- Reworded the null-buffer assertion comment to describe preserving the null-buffer-free output shape and avoiding the validity allocation, without referring to the filter fast path.

The benchmark results and methodology are now included in the PR description. The targeted local checks still pass:
- cargo fmt --check
- cargo test -p datafusion-comet-spark-expr --lib rlike
- cargo clippy -p datafusion-comet-spark-expr --lib --tests -- -D warnings
Thanks again for the detailed validation and for confirming that the iterator-based implementation is a reasonable performance trade-off.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Updated both the direct and dictionary Utf8View cases to use "Rhododendrons" (13 bytes), so they exercise the out-of-line view representation rather than the inline representation used for values of 12 bytes or fewer.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Added coverage for Dictionary(UInt64, Utf8) and a dictionary sliced with a non-zero offset. Both cases use the shared [Some(true), None, Some(false)] expected result.

native/spark-expr/src/predicate_funcs/rlike.rs
sam-1112: Agreed. I reworded the comment to describe preserving the null-buffer-free output shape and avoiding the unnecessary validity allocation, without referring to the filter fast path.

sam-1112:
Hi @andygrove, just a gentle ping when you get a chance. I think those inline comments mentioned above have been solved from your latest review. Thanks!

andygrove (approved):
LGTM. Thanks @sam-1112

sunchao (approved):
LGTM. Reviewed head `8c0bc9a4cea257998f974a49a9d99c0ff30267dd` against base `b7f35b6acb447d167d2a77eae6847f7b5fbb59b1` and found no actionable issues.

I checked native evaluation, Spark routing, ownership and error handling, compatibility, and edge cases. The layout dispatch and dictionary handling preserve nulls and reuse the compiled regex correctly.

- Local validation: all 10 tests passed, comprising the six existing tests and four additional tests covering 1,620 layout, slice, and dictionary combinations, plus empty, all-null, all-valid, and invalid-type cases. All eight dictionary key types were exercised across Utf8, LargeUtf8, and Utf8View. Formatting also passed.
- CI: 23 successful checks and 14 skipped. I verified that the tested merge commit has the same source tree as this head. The native job passed 1,581 tests, including all six RLike tests. The Spark 4.1 expression job passed 1,517 tests, including the RLike SQL fixtures and regex parity tests. [CI run](https://github.com/apache/datafusion-comet/actions/runs/35440529324)
- Earlier feedback: the requested implementation and test changes are present, although eight review threads remain unresolved.

Local testing compiled the exact expression source against cached dependencies matching the lockfile. I did not perform a full local Spark/JNI rebuild or repeat the benchmarks. The native CI job skipped five tests, and the Spark expression job had one canceled and 12 ignored tests. Other Spark runtime profiles, Spark's upstream SQL suites, and macOS remain outside this CI validation.

Issue #4554, opened 2026-06-01
try_make_timestamp returns wrong values for invalid inputs instead of NULL


andygrove:
## Bug Report

When `try_make_timestamp` is called with column arguments containing invalid date/time components (e.g., month=13, day=32, hour=25), Comet runs the query natively (CometProject) but returns an incorrect timestamp value instead of NULL.

Spark's `try_make_timestamp` rewrites to `MakeTimestamp(failOnError=false)`, which should return NULL instead of throwing an exception for invalid inputs.

## Reproduction

```sql
CREATE TABLE test_try_make_ts(y int, mo int, d int, h int, mi int, s decimal(8,6)) USING parquet;
INSERT INTO test_try_make_ts VALUES (2024, 13, 1, 0, 0, 0.0), (2024, 6, 32, 0, 0, 0.0), (2024, 6, 15, 25, 0, 0.0);
SELECT try_make_timestamp(y, mo, d, h, mi, s) FROM test_try_make_ts;
```

## Expected

All three rows should return NULL (invalid components).

## Actual (Comet)

Returns garbage timestamp values (e.g., `1969-12-31 15:59:59.999999`) for the invalid rows.

## Notes

- Literal invalid inputs (e.g., `try_make_timestamp(2024, 13, 1, 0, 0, 0.0)`) may also be affected.
- The related feature request is #3117.

PR #5219, opened 2026-08-02
fix: codegen dispatcher null short-circuit swallowed ANSI errors, plus two latent TIME-type gaps

### LINKED ISSUE #5218: Codegen dispatcher: whole-tree NullIntolerant short-circuit suppresses ANSI errors, plus TIME type gaps between canHandle and the runtime dispatcher
## Describe the bug

Three related defects in the JVM codegen dispatcher (`spark/src/main/scala/org/apache/comet/codegen/`). The first produces wrong results today; the other two are latent gaps in the same code where the plan-time gate accepts a type the runtime dispatcher rejects.

Context: the dispatcher is not only used for `ScalaUDF`. Roughly 70 built-in expressions route through `CometCodegenDispatch` (math, arrays, maps, strings, json, csv, xpath, datetime), so the blast radius is wide.

### Finding 1 (confirmed): whole-tree null short-circuit suppresses runtime errors that Spark raises

`CometBatchKernelCodegen.defaultBody` applies a "any input null implies null output" short-circuit when every node in the bound tree is `NullIntolerant`. It emits a pre-check on the **union of all input ordinals**:

```java
if (this.col0.isNull(i) || this.col1.isNullAt(i)) { output.setNull(i); } else { ev.code; write }
```

Spark's own null handling is **per-node and left-to-right**: `BinaryExpression.nullSafeCodeGen` emits the left child's code unconditionally, then tests the left child's null, then the right child's. So when a left subtree raises for a given row and a *different* input is null on that same row, Spark raises but Comet silently returns NULL.

This is an error-semantics divergence only; values are unaffected. A `NullIntolerant` node that ignored one of its inputs would already be a Spark bug (checked `nanvl` specifically, which ignores its second argument when the first is non-NaN, and it is correctly *not* `NullIntolerant`).

The divergence needs ANSI mode to be observable in most cases, and ANSI is on by default in Spark 4. Many dispatch roots are `NullIntolerant` and have error paths: `AddMonths`, `MonthsBetween`, `MakeTimestamp`, `GetTimestamp`, `MakeDTInterval`, `ToNumber`, `Conv`, `WidthBucket`, `Pmod`, and others.

### Finding 2 (latent): plan-time gate accepts TIME, runtime dispatcher rejects it

`CometBatchKernelCodegen.isSupportedDataType` accepts time types, `primitiveArrowClasses` includes `TimeNanoVector`, and `emitTypedGetters` emits a `getLong` case for it, but `CometScalaUDFCodegen.specFor` omits `TimeNanoVector` and so falls into `case other => throw new UnsupportedOperationException`.

Because `canHandle` greenlights the expression at plan time, there is no fallback: the query fails at execute time. Not reachable today because Spark 4.1.3 rejects TIME columns in file-based data sources (`UNSUPPORTED_TIME_TYPE` from `DataSourceUtils.verifySchema`), so this is a latent trap that goes live as soon as Spark supports TIME in Parquet or a native operator produces a TIME column.

### Finding 3 (latent): same TIME hole in the generic getter dispatch

`CometSpecializedGettersDispatch.get` has no time-type branch, so `get(ordinal, TimeType)` throws. `emitSpecializedGetterExpr` and `elementGetterCall` both handle time types, so the hole is only in the generic path, which is exactly the path `CodegenFallback.eval(row)` and `SafeProjection` (for struct/array arguments) use.

## Steps to reproduce

For Finding 1, with the default profile (Spark 4.1) and `spark.sql.ansi.enabled=true`:

```sql
CREATE TABLE t (s STRING, i INT) USING parquet;
INSERT INTO t VALUES ('notadate', NULL);
SELECT add_months(CAST(s AS DATE), i) FROM t;
```

Result:

- Spark (`spark.comet.enabled=false`): throws `SparkDateTimeException` `[CAST_INVALID_INPUT]`
- Comet: returns `NULL`

The plan confirms the dispatcher runs:

```
*(1) CometColumnarToRow
+- CometProject [add_months(CAST(s AS DATE), i)#21], [add_months(cast(s#8 as date), i#9) AS ...]
   +- CometNativeScan parquet spark_catalog.default.t[s#8,i#9] ...
```

`add_months` is the witness because it is a plain `BinaryExpression` whose `doGenCode` goes through `defineCodeGen`/`nullSafeCodeGen`. Note that `pmod` and the `div`/`mod` family are *not* witnesses: `DivModLike` deliberately evaluates its right child first so it can skip the left when the divisor is zero, so Spark also returns NULL there.

Also reproducible at the kernel level without a query plan, by compiling `Add(Cast(BoundReference(0, StringType), IntegerType, EvalMode.ANSI), BoundReference(1, IntegerType))` and running one batch of `('abc', NULL)`: the kernel returns NULL while `expr.eval(row)` throws `SparkNumberFormatException`.

For Findings 2 and 3, no end-to-end reproducer exists yet (see above); they are visible by inspection of the type surface accepted by `canHandle` versus the one handled by `specFor` and `CometSpecializedGettersDispatch.get`.

## Expected behavior

Finding 1: Comet should raise the same error Spark raises. Proposed fix is to restrict the short-circuit to trees with a **single distinct** `BoundReference` ordinal. With one input, Spark also evaluates nothing when that input is null, so the optimization stays exact, and the common unary shapes (`upper`, `length`, `date_format`) keep the fast path. Multi-input trees fall back to the plain `ev.code` plus `ev.isNull` body.

Findings 2 and 3: `specFor` should recognize `TimeNanoVector` and `CometSpecializedGettersDispatch.get` should have a time-type branch, so the runtime type surface matches what `canHandle` promises.

## Additional context

Found during an audit sweep of the codegen area. Existing coverage is otherwise healthy: `CometCodegenSuite`, `CometCodegenSourceSuite`, and `CometCodegenHOFSuite` pass 135/135 on the default profile, so these are gaps rather than regressions.

Two further items were found in the same sweep and are *not* included here, since they need separate discussion:

- Output decimal writes in `CometBatchKernelCodegenOutput` do not normalize scale before writing, whereas Spark's `UnsafeRowWriter.write(ordinal, Decimal, precision, scale)` calls `changePrecision` first and nulls out on failure. No live reproducer was found because `CatalystTypeConverters.DecimalConverter` already normalizes for `ScalaUDF`.
- The per-row body is inlined in `process`'s loop and never split (already noted as `TODO(method-size)` in the source), so a deep tree can exceed Janino's 64KB method limit at execute time. `canHandle` gates on `spark.sql.codegen.maxFields`, which bounds field count rather than code size.

andygrove:

Closes #5218.

## Rationale for this change

#5218 bundles three independent defects in the `CometCodegenDispatch` code path. This PR fixes all three. They share no code path with each other, so each section below is independently testable and revertable.

### 1. The null short-circuit swallowed errors Spark raises

`CometBatchKernelCodegen.defaultBody` applied an "any input null implies null output" short-circuit on the **union of input ordinals** whenever every node in the bound tree was `NullIntolerant`:

```java
if (this.col0.isNull(i) || this.col1.isNullAt(i)) { output.setNull(i); } else { ev.code; write }
```

Spark's null handling is **per-node and left-to-right**. `BinaryExpression.nullSafeCodeGen` emits the left child's code unconditionally, then tests the left child's null, then the right child's. A short-circuit over the union of ordinals therefore skips a subtree Spark would have evaluated, and with it any error that subtree raises.

Under ANSI, `add_months(CAST(s AS DATE), i)` on the row `('notadate', NULL)` raises `CAST_INVALID_INPUT` in Spark. Comet returned `NULL`. ANSI is on by default in Spark 4, and roughly 70 built-in expressions route through `CometCodegenDispatch` (math, arrays, maps, strings, json, csv, xpath, datetime), many of them `NullIntolerant` with error paths: `AddMonths`, `MonthsBetween`, `MakeTimestamp`, `GetTimestamp`, `MakeDTInterval`, `ToNumber`, `Conv`, `WidthBucket`, `Pmod`.

Values were never wrong, only errors were lost: a `NullIntolerant` node that ignored one of its inputs would already be a Spark bug. I checked `nanvl` specifically, since it ignores its second argument when the first is non-NaN, and it is correctly *not* `NullIntolerant`.

The defect needs a subtree between the root and a leaf that can raise independently of nullness, like the `Cast` above. A `BoundReference` or `Literal` read cannot raise on its own, so a one-level-deep tree is still safe to short-circuit; see the fix below.

### 2 and 3. Two TIME-type gaps in the same dispatcher

In both cases the plan-time gate accepted a type the runtime rejected. Because `canHandle` greenlights the expression before the plan commits, both were execute-time failures with no fallback.

Both are latent today: Spark 4.1.3 rejects TIME columns in file-based data sources (`UNSUPPORTED_TIME_TYPE`), so no SQL query can produce a TIME input to the dispatcher yet. They become live as soon as Spark supports TIME in Parquet or a native operator emits a TIME column.

**1. The short-circuit fix.** On top of the existing `NullIntolerant` root and `allNullIntolerant` tree conditions, `canShortCircuitNulls` now requires the tree to read **either exactly one input ordinal, or have only leaf children on the root**. Both shapes make the short-circuit exact:

- *One ordinal*: there is nothing left for Spark to evaluate ahead of that ordinal's own null check. Keeps the fast path for the common unary shapes (`upper`, `length`, `date_format`).
- *Leaf-only children* (every direct child of the root is a `BoundReference` or `Literal`): the tree is one level deep, so the only code Spark runs ahead of its null checks is leaf reads, which cannot raise. Any error comes from the root's own logic, which runs only once every input is known non-null — in Spark's shape and in ours alike. Keeps the fast path for the common no-cast multi-argument case (`pmod(a, b)`, `conv(a, b, c)`, `make_timestamp(y, m, d, h, mi, s)`).

Anything else — a real subtree between the root and a leaf, which is what makes an error reachable ahead of a null check — falls back to the plain `ev.code` plus `ev.isNull` body, which is what Spark's own generated code does. `defaultBody` now emits the disjunction over every ordinal the tree reads rather than only the first.

The leaf-only rule leans on a `NullIntolerant` root not raising before it has checked every input for null. I verified that against Spark's generated code for the one family that reorders its children: `DivModLike.doGenCode`, and `Pmod.doGenCode` which carries its own copy of the same shape, evaluate the divisor first but still test the dividend's null *before* the divide/remainder-by-zero throw, so `pmod(NULL, 0)` returns NULL rather than raising. A root that raised ahead of its null checks would contradict `NullIntolerant`.

**2. `CometScalaUDFCodegen.specFor` missing `TimeNanoVector`.** It fell into `case other => throw new UnsupportedOperationException`, even though `isSupportedDataType` accepts time types, `primitiveArrowClasses` includes the vector class, and `emitTypedGetters` emits a `getLong` case for it.

**3. `CometSpecializedGettersDispatch.get` missing a time-type branch,** so `get(ordinal, TimeType)` threw. That is the path `SafeProjection` (for ScalaUDF struct arguments) and every `CodegenFallback.eval(row)` use, including higher-order functions. `emitSpecializedGetterExpr` and `elementGetterCall` already routed time types to `getLong`, so only the generic dispatch was missing.

Doc comments were added on `specFor` and the dispatch object noting that their type surface has to stay in step with `isSupportedDataType`.

Four new tests, each verified to fail against the pre-fix behaviour and pass after:

- `CometCodegenSuite`: "multi-input NullIntolerant tree does not swallow an ANSI error (#5218)" runs the `add_months(CAST(s AS DATE), i)` query through `checkSparkAnswerMaybeThrows` under ANSI and asserts Comet raises the same `CAST_INVALID_INPUT` Spark does, plus asserts the dispatcher actually ran for the query. Pre-fix this returned `cometErr=None`.
- `CometCodegenSuite`: "single-input NullIntolerant tree still short-circuits nulls" guards against over-correcting.
- `CometCodegenSuite`: "multi-input leaf-only NullIntolerant tree short-circuits nulls correctly (#5218)" runs `pmod(a, b)` and `a + b` over every null combination across two ordinals, confirming the restored fast path matches Spark row for row.
- `CometCodegenSuite`: "leaf-only short-circuit preserves ANSI remainder-by-zero behaviour (#5218)" pins the `Pmod` reordering case from both sides — `(NULL, 0)` must return NULL, `(7, 0)` must raise `REMAINDER_BY_ZERO`.
- `CometCodegenSourceSuite`: "NullIntolerant short-circuit skipped when a subtree sits between root and leaf (#5218)" and "NullIntolerant short-circuit kept for a multi-input leaf-only tree (#5218)" pin both emitted source shapes.
- `CometCodegenSuite`: "TIME input column routes through the dispatcher (#5218)" drives `CometScalaUDFCodegen.evaluate` with a `TimeNanoVector` directly, since no SQL query can reach it. Gated on `isSpark41Plus`.
- New `spark/src/test/scala/org/apache/comet/codegen/CometSpecializedGettersDispatchSuite.scala` covers `get` for `TimeType`, the null slot, and the neighbouring long/int-backed temporal types so a future reordering that shadows a case is caught.

The #4554 regression test ("nullable NullIntolerant root keeps post-eval isNull guard inside short-circuit") keeps its original two-`setNull`-site assertion: its `MakeTimestamp` over six `BoundReference`s is a leaf-only tree, so it still gets the short-circuit. It now additionally asserts the short-circuit tests all six ordinals, not just the first. A new sibling test, "multi-input tree with a non-leaf child keeps only the post-eval guard (#5218)", covers the same root with a `Cast` under it and pins the single-`setNull` shape, so the post-eval guard is verified to survive independently of the short-circuit.

I confirmed the new source-level assertions discriminate: reverting `canShortCircuitNulls` to the single-ordinal-only rule fails exactly the two tests that pin the restored fast path, and nothing else.

Suite runs on the default profile (Spark 4.1 / Scala 2.13, JDK 17):

- `CometCodegenSuite`, `CometCodegenSourceSuite`, `CometCodegenHOFSuite`, `CometCodegenFuzzSuite`, `CometSpecializedGettersDispatchSuite`: 175 tests, 0 failures.
- Because the change touches the shared code path for every `CometCodegenDispatch` expression, also ran `CometTemporalExpressionSuite`, `CometStringExpressionSuite`, `CometMathExpressionSuite`, `CometArrayExpressionSuite`, `CometMapExpressionSuite`, `CometJsonExpressionSuite`, `CometCsvExpressionSuite`: 137 tests, 0 failures.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
+++ b/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
@@ -337,9 +337,10 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
   }
 
   /**
-   * Per-row body. For `NullIntolerant` expressions over a single input ordinal, prepends a
-   * short-circuit on that ordinal so the whole `ev.code` cost is skipped on null rows. Otherwise
-   * the standard shape: run `ev.code`, then `setNull` or write based on `ev.isNull`.
+   * Per-row body. For `NullIntolerant` expressions whose input nulls fully determine a null
+   * output (see [[canShortCircuitNulls]]), prepends a short-circuit on those ordinals so the
+   * whole `ev.code` cost is skipped on null rows. Otherwise the standard shape: run `ev.code`,
+   * then `setNull` or write based on `ev.isNull`.
    *
    * `subExprsCode` is the CSE helper-invocation block. It must run before `ev.code`. Inside the
    * short-circuit it lives in the else branch so null rows skip CSE too.
@@ -356,9 +357,13 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
       // exposes `isNullAt(int)` rather than the raw Arrow `isNull(int)`. Pick the right method
       // for the ordinal so the short-circuit compiles for timestamp / int / float columns too,
       // not just VarChar / Decimal vectors that stay as raw Arrow types.
-      val ord = inputOrdinals.head
-      val nullCheck =
-        s"this.col$ord.${CometBatchKernelCodegenInput.nullCheckMethod(inputSchema(ord))}(i)"
+      //
+      // Multi-ordinal trees test the disjunction of their ordinals. That is only reachable for
+      // leaf-only-children roots, where it is exact; see [[canShortCircuitNulls]].
+      val nullCheck = inputOrdinals
+        .map(ord =>
+          s"this.col$ord.${CometBatchKernelCodegenInput.nullCheckMethod(inputSchema(ord))}(i)")
+        .mkString(" || ")
       // `NullIntolerant` only constrains "any input null -> output null"; it does NOT promise
       // that non-null inputs always produce non-null output. `MakeTimestamp(failOnError=false)`
       // is `NullIntolerant=true` but its `doGenCode` catches `DateTimeException` for invalid
@@ -414,24 +419,57 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
   }
 
   /**
-   * Gates the [[defaultBody]] null short-circuit. Three conditions, all necessary:
+   * Gates the [[defaultBody]] null short-circuit ("any input null implies null output", tested as
+   * the disjunction of the tree's input ordinals before `ev.code` runs).
    *
-   *   - The tree reads exactly one input ordinal. Spark's null handling is per-node and
-   *     left-to-right (`BinaryExpression.nullSafeCodeGen` emits the left child's code
-   *     unconditionally, then tests the left child's null, then the right child's), so a
-   *     short-circuit on the union of several ordinals is not equivalent: it skips a subtree that
-   *     Spark would have evaluated, and with it any error that subtree raises. Under ANSI,
-   *     `add_months(cast(s as date), i)` on `('notadate', NULL)` raises `CAST_INVALID_INPUT` in
-   *     Spark, so returning null here would be wrong. With a single ordinal there is nothing left
-   *     for Spark to evaluate ahead of that ordinal's own null check, so the short-circuit is
-   *     exact. (A literal-only subtree that raises would be the one exception, but Catalyst's
-   *     `ConstantFolding` evaluates those at analysis time and never leaves one in the tree.)
+   * The short-circuit is only equivalent to Spark when no subtree that Spark would have evaluated
+   * ahead of a null check gets skipped. Spark's null handling is per-node and left-to-right:
+   * `BinaryExpression.nullSafeCodeGen` emits the left child's code unconditionally, then tests
+   * the left child's null, then the right child's. So for `add_months(cast(s as date), i)` on
+   * `('notadate', NULL)` Spark evaluates the cast and, under ANSI, raises `CAST_INVALID_INPUT`; a
+   * short-circuit on the union of ordinals would skip the cast and return null, swallowing the
+   * error (#5218).
+   *
+   * Conditions, all necessary:
+   *
+   *   - The tree reads at least one input ordinal. A literal-only tree has nothing to test.
+   *     (Catalyst's `ConstantFolding` evaluates those at analysis time, so this is defensive.)
    *   - The root is `NullIntolerant`, so a null input really does mean a null result.
    *   - Every node in the tree is null-propagating ([[allNullIntolerant]]); a `Coalesce` / `If` /
    *     `CaseWhen` anywhere would break the chain.
+   *   - Either the tree reads exactly one ordinal, or every direct child of the root is a leaf
+   *     ([[rootChildrenAreLeaves]]). Both shapes make the short-circuit exact:
+   *     - One ordinal: there is nothing left for Spark to evaluate ahead of that ordinal's own
+   *       null check.
+   *     - Leaf-only children: the tree is one level deep, so the only code Spark runs ahead of
+   *       its null checks is `BoundReference` / `Literal` reads, which cannot raise. Any error
+   *       comes from the root's own logic, which runs only once every input is known non-null --
+   *       in both Spark's shape and ours. This keeps the fast path for the common no-cast
+   *       multi-argument case (`pmod(a, b)`, `conv(a, b, c)`, `make_timestamp(y, m, d, h, mi,
+   *       s)`) across the ~70 expressions that route through this dispatcher.
+   *
+   * The leaf-only rule relies on a `NullIntolerant` root not raising before it has checked every
+   * input for null. Spark's own generated code honors that even where it reorders children: the
+   * divide/remainder family (`DivModLike.doGenCode`, and `Pmod.doGenCode`, which carries its own
+   * copy of the same shape) evaluates the divisor first, but still tests the dividend's null
+   * *before* its divide/remainder-by-zero throw, so `pmod(NULL, 0)` returns null rather than
+   * raising. A root that raised ahead of its null checks would contradict `NullIntolerant`.
    */
   private def canShortCircuitNulls(expr: Expression, inputOrdinals: Seq[Int]): Boolean =
-    inputOrdinals.size == 1 && isNullIntolerant(expr) && allNullIntolerant(expr)
+    inputOrdinals.nonEmpty && isNullIntolerant(expr) && allNullIntolerant(expr) &&
+      (inputOrdinals.size == 1 || rootChildrenAreLeaves(expr))
+
+  /**
+   * True iff every direct child of the root is a leaf (`BoundReference` or `Literal`), i.e. the
+   * tree is one level deep with no subtree between the root and its inputs. One of the conditions
+   * on [[canShortCircuitNulls]]: it is what makes a multi-ordinal short-circuit exact, because a
+   * leaf read cannot raise an error that Spark's left-to-right evaluation would have surfaced.
+   */
+  private def rootChildrenAreLeaves(expr: Expression): Boolean =
+    expr.children.forall {
+      case _: BoundReference | _: Literal => true
+      case _ => false
+    }
 
   /**
    * True iff every node in the tree propagates nulls (`NullIntolerant`, `BoundReference`, or
--- a/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
+++ b/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
@@ -337,10 +337,9 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
   }
 
   /**
-   * Per-row body. For `NullIntolerant` expressions where the entire tree propagates nulls,
-   * prepends a short-circuit on the union of input ordinals so the whole `ev.code` cost is
-   * skipped on null rows. Otherwise the standard shape: run `ev.code`, then `setNull` or write
-   * based on `ev.isNull`.
+   * Per-row body. For `NullIntolerant` expressions over a single input ordinal, prepends a
+   * short-circuit on that ordinal so the whole `ev.code` cost is skipped on null rows. Otherwise
+   * the standard shape: run `ev.code`, then `setNull` or write based on `ev.isNull`.
    *
    * `subExprsCode` is the CSE helper-invocation block. It must run before `ev.code`. Inside the
    * short-circuit it lives in the else branch so null rows skip CSE too.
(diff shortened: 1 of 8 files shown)


spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
```diff
@@ -351,82 +350,94 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
...
+   *     Spark, so returning null here would be wrong. With a single ordinal there is nothing left
+   *     for Spark to evaluate ahead of that ordinal's own null check, so the short-circuit is
+   *     exact. (A literal-only subtree that raises would be the one exception, but Catalyst's
+   *     `ConstantFolding` evaluates those at analysis time and never leaves one in the tree.)
+   *   - The root is `NullIntolerant`, so a null input really does mean a null result.
+   *   - Every node in the tree is null-propagating ([[allNullIntolerant]]); a `Coalesce` / `If` /
+   *     `CaseWhen` anywhere would break the chain.
+   */
+  private def canShortCircuitNulls(expr: Expression, inputOrdinals: Seq[Int]): Boolean =
+    inputOrdinals.size == 1 && isNullIntolerant(expr) && allNullIntolerant(expr)
```
mbutrovich: This drops the short-circuit for every multi-argument `NullIntolerant` tree, not only the ones that actually had the bug.

The bug needs a subtree between the root and a leaf that can raise an error independent of nullness, like the `Cast` in `add_months(CAST(s AS DATE), i)`. A `BoundReference` or `Literal` read can't raise anything on its own. So for a tree where every direct child of the root is a `BoundReference` or `Literal`, such as `Pmod(colA, colB)` or `Conv(colA, colB, colC)` with no cast on any argument, the union-of-ordinals short-circuit is still exact: the only place an error can occur is the root's own `eval`, and that only runs in the `else` branch either way.

`allNullIntolerant` already walks the tree to reject `Coalesce` / `If` / `CaseWhen`. Could `canShortCircuitNulls` check "every direct child of the root is a `BoundReference` or `Literal`" and keep the short-circuit for that shape, falling back to the current single-ordinal rule otherwise? That would restore the fast path for the common no-cast case across the ~70 expressions that route through this dispatcher, without reintroducing the error-swallowing bug.

mbutrovich (commented):
First pass, thanks @andygrove! Also one non-inline comment:

The title and commit message name only the null short-circuit bug, but the PR also fixes two unrelated latent type-coverage gaps: `CometScalaUDFCodegen.specFor` missing `TimeNanoVector`, and `CometSpecializedGettersDispatch.get` missing a time-type branch. Issue #5218 bundles all three as one issue, and the PR body does disclose all three under "What changes are included in this PR," so this isn't hidden. But the three fixes touch different files with no shared code path between them, and each is independently testable and revertable. Worth splitting into a separate PR against a separate issue, or if not, updating the title to say the PR does more than fix the short-circuit.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
andygrove: Done in dd6ae3225 — `canShortCircuitNulls` now accepts either exactly one input ordinal or a leaf-only-children root, and `defaultBody` emits the disjunction over every ordinal the tree reads instead of just the first. That restores the fast path for the no-cast multi-argument shapes.

One wrinkle worth recording, because it makes the rule correct for a slightly different reason than stated above. The argument was that with leaf-only children "the only place an error can occur is the root's own `eval`, and that only runs in the `else` branch either way." That isn't quite true for `pmod`: it is one of the expressions that deliberately reorders its children, so its error check is *not* simply gated behind all-inputs-non-null. `Pmod.doGenCode` (it extends `BinaryArithmetic` directly and carries its own copy of the `DivModLike` shape) evaluates the divisor first and throws `REMAINDER_BY_ZERO` under ANSI:

```java
${eval2.code}                              // divisor
if (${eval2.isNull}) {
  ${ev.isNull} = true;
} else {
  ${eval1.code}                            // dividend
  if (${eval1.isNull}) {
    ${ev.isNull} = true;                   // <-- reached before the throw
  } else {
    if ($isZero) throw ...remainderByZeroError(...);
    $result
  }
}
```

The rule survives, but because the *dividend's null check precedes the throw*, not because the throw is gated on all inputs being non-null. So `pmod(NULL, 0)` returns NULL in Spark, and the union short-circuit stays exact for it. Had the throw been hoisted above the `eval1.isNull` test, `pmod(a, b)` with no cast on either argument would have been a swallowed-error case under the leaf-only rule.

Since that is load-bearing and invisible from Comet's side, I documented it on `canShortCircuitNulls` and pinned it from both directions in `CometCodegenSuite`: `(NULL, 0)` must return NULL, `(7, 0)` must raise `REMAINDER_BY_ZERO`. If a future Spark version reorders that throw, the second assertion fails rather than the behaviour silently regressing.

Also added: source-shape tests for both the kept (leaf-only) and skipped (subtree-under-root) shapes, and a runtime test over every null combination across two ordinals. The #4554 test goes back to its original two-`setNull` assertion, since its `MakeTimestamp` over six `BoundReference`s is leaf-only and keeps the short-circuit; it now also asserts all six ordinals get tested. A new sibling test puts a `Cast` under the same root to pin the single-`setNull` shape, so the #4554 post-eval guard is still verified independently of the short-circuit.

I confirmed the new assertions actually discriminate: reverting the gate to the single-ordinal-only rule fails exactly the two tests that pin the restored fast path and nothing else. 175 tests across the 5 codegen suites and 137 across the dispatcher-backed expression suites, 0 failures.

mbutrovich (approved):
LGTM, thanks @andygrove!

PR #5225, opened 2026-08-02
fix: avoid duplicate CheckOverflow evaluation for decimal division

### LINKED ISSUE #5190: Avoid duplicate CheckOverflow evaluation for decimal division

Decimal `Divide` is currently serialized with two consecutive, equivalent `CheckOverflow` expressions.

After #5171:

- [`DecimalPrecision.promote`](https://github.com/apache/datafusion-comet/blob/f3050be3f19f274cc226b27a460ab2e842e2b9f7/spark/src/main/scala/org/apache/spark/sql/comet/DecimalPrecision.scala#L60-L62) adds an outer wrapper with `datatype = Divide.dataType` and `failOnError = (Divide.evalMode == ANSI)`.
- [`CometDivide.convert`](https://github.com/apache/datafusion-comet/blob/f3050be3f19f274cc226b27a460ab2e842e2b9f7/spark/src/main/scala/org/apache/comet/serde/arithmetic.scala#L276-L286) adds an inner wrapper with the same datatype and overflow mode.

The resulting shape is:

```text
CheckOverflow(type=T, fail=F)
  CheckOverflow(type=T, fail=F)
    Divide(...)
```

The native planner only removes a `CheckOverflow` whose child is a matching `WideDecimalBinaryExpr`, or fuses one whose child is a decimal `Cast`; it does not collapse nested `CheckOverflow` expressions ([`planner.rs`](https://github.com/apache/datafusion-comet/blob/f3050be3f19f274cc226b27a460ab2e842e2b9f7/native/core/src/execution/planner.rs#L515-L563)).

On successful array-valued decimal-division batches, both wrappers therefore perform the precision scan in [`CheckOverflow.evaluate`](https://github.com/apache/datafusion-comet/blob/f3050be3f19f274cc226b27a460ab2e842e2b9f7/native/spark-expr/src/math_funcs/internal/checkoverflow.rs#L105-L135).

Stop `DecimalPrecision.promote` from wrapping decimal `Divide`, while retaining the specialized wrapper in `CometDivide.convert`. That wrapper is required to handle Comet's native decimal-division overflow sentinel correctly.

Acceptance criteria:

- Decimal `Divide` serializes to exactly one `CheckOverflow`.
- Its datatype equals `Divide.dataType`.
- `failOnError` is true only for ANSI mode.
- LEGACY and TRY overflow behavior remains null-producing.
- ANSI overflow behavior remains error-producing.
- Existing divide-by-zero behavior is unchanged.
- Add a proto-level regression covering LEGACY, TRY, and ANSI modes.

### Additional context

Raised in [review of #5171](https://github.com/apache/datafusion-comet/pull/5171#pullrequestreview-4830862439), while fixing #5075.

peterxcli:

Closes #5190.

## Rationale for this change

Decimal overflow wrapping should have a single owner. Some serializers, such as
`ArrayContains`, recursively call the public `exprToProto` API on children that
`DecimalPrecision.promote` has already visited, so promotion must be idempotent.

https://github.com/apache/datafusion-comet/issues/5190 proposed removing the
promotion wrapper and keeping the specialized wrapper in `CometDivide.convert`. This PR
intentionally takes the opposite approach: it removes the serde wrapper and keeps
`DecimalPrecision.promote` as the single owner. That also makes the fix apply consistently to
decimal Add, Subtract, Multiply, Divide, and Remainder expressions.

https://github.com/apache/datafusion-comet/issues/5248 tracks removing unnecessary
recursive promotion from serializers.

- Make `DecimalPrecision.promote` collapse adjacent equivalent `CheckOverflow` wrappers around decimal binary arithmetic while preserving wrappers with different result types or overflow modes.
- Remove the specialized decimal `CheckOverflow` creation from `CometDivide.convert`, keeping overflow wrapping centralized in decimal promotion.
- Add a real-SQL proto regression through `array_contains` covering decimal Add, Subtract, Multiply, Divide, and Remainder, including LEGACY, TRY, and ANSI divide modes.
- Update the existing per-expression eval-mode regression to expect a single decimal Divide wrapper.

## Performance

Compared the PR base (`2cba02f66`) with the PR head (`e27a9f645`) using Spark 4.2,
release builds, the same optimized native library, and TPC-DS SF1 `store_sales`
(2,879,789 rows) on an Apple M4. The temporary decimal-division-heavy microbenchmark query was:

```sql
SELECT ss_net_profit / CAST(3.00 AS DECIMAL(7, 2))
FROM store_sales
```

Five alternating fresh-JVM runs produced these results:

| Engine | Revision | Best times (ms) | Median (ms) | Change |
| --- | --- | ---: | ---: | ---: |
| Comet | Base | 58, 59, 58, 60, 63 | 59 | — |
| Comet | PR | 60, 54, 53, 56, 55 | 55 | -6.8% |
| Spark control | Base | — | 509 | — |
| Spark control | PR | — | 513 | +0.8% |

The Spark control moved by less than 1%, so the Comet improvement was larger than the
run-to-run drift.

- `/mvnw test -Pspark-4.1 -Dtest=none -Dsuites="org.apache.spark.sql.comet.CometDecimalArithmeticViewSuite"`
- `./mvnw test -Pspark-4.2 -Dtest=none -Dsuites="org.apache.spark.sql.comet.CometDecimalArithmeticViewSuite"`
- `./mvnw test -Pspark-3.4 -Dtest=none -Dsuites="org.apache.comet.CometSqlFileTestSuite decimal_div"`
- `./mvnw test -Pspark-4.2 -Dtest=none -Dsuites="org.apache.comet.CometSqlFileTestSuite decimal_div"`
- `git diff --check`


spark/src/main/scala/org/apache/comet/serde/arithmetic.scala
```diff
@@ -263,7 +263,7 @@ object CometDivide extends CometExpressionSerde[Divide] with MathBase {
     // For now, use NullIf to swap zeros with nulls.
     val rightExpr =
       if (expr.evalMode != EvalMode.ANSI) nullIfWhenPrimitive(expr.right) else expr.right
-    val divideExpr = createMathExpression(
+    createMathExpression(
```
andygrove, line 271: One thing I want to check here. The guard you removed fired on `expr.dataType.isInstanceOf[DecimalType]` alone. The `promote` rule requires both operands to match `DecimalExpression` as well, and `getSupportLevel` only looks at `expr.left.dataType`. So if a decimal-typed `Divide` with a non-decimal operand ever reaches this serde, there is no `CheckOverflow` anywhere and the native `i128::MAX` sentinel comes back as a real value.

I could not construct that shape through Spark's type coercion, so I believe it is unreachable today. Since the failure mode is silent wrong data rather than a fallback, would you be up for a defensive guard? Returning `Unsupported` from `getSupportLevel` when `expr.dataType` is decimal but the operands would not match `promote`'s pattern would keep the invariant checkable next to the code that depends on it.

spark/src/main/scala/org/apache/spark/sql/comet/DecimalPrecision.scala
```diff
@@ -45,6 +45,15 @@ object DecimalPrecision {
       // happen if the Spark version is < 3.4
       case e: BinaryArithmetic if e.left.prettyName == "promote_precision" => e
 
+      // Recursive exprToProto calls can re-promote every decimal binary operator below. Collapse
+      // only equivalent wrappers so the shared promotion rule remains idempotent.
+      case outer @ CheckOverflow(
```
andygrove, line 46: This is a good fix, and the bottom-up collapse converges correctly at any depth.

The re-promotion it works around comes from serdes calling the public `exprToProto` on children that `promote` has already visited. There are 64 such call sites in the serde package. The aggregate ones genuinely need the promoting entry point, but the ones in `arrays.scala` and `bitwise.scala` are re-walking an already-promoted tree, and each walk drags `liftCoverageTags` along with it. That is separate work from this PR, but could you file a tracking issue for it and link it here? Otherwise it will not get picked up.

spark/src/test/spark-4.1+/org/apache/spark/sql/comet/CometDecimalArithmeticViewSuite.scala
```diff
@@ -119,4 +104,38 @@ class CometDecimalArithmeticViewSuite extends CometTestBase {
       }
     }
   }
+
+  test("issue #5190: recursive serialization does not duplicate decimal CheckOverflow") {
```
andygrove: This regression does not depend on anything 4.1-specific, but sitting under `spark-4.1+` means 3.4, 3.5, and 4.0 CI never runs it, and the duplication it guards against exists on all of them. Could it move to a version-agnostic suite under `spark/src/test/scala/org/apache/spark/sql/comet/`? A new `CometDecimalPromotionSuite` would work. The other two tests in this file can stay where they are since they need `NumericEvalContext`.

While it is moving, two additions would be worth it. ANSI is only exercised for `divide`, but the collapse guard compares `nullOnOverflow`, so ANSI is exactly the case that would catch a wrong guard for add, subtract, multiply, and remainder. And a direct assertion that `promote(promote(e)) == promote(e)` would state the idempotency property outright instead of leaving it to be inferred from the `array_contains` proto shape.

spark/src/test/spark-4.1+/org/apache/spark/sql/comet/CometDecimalArithmeticViewSuite.scala
```diff
@@ -119,4 +104,38 @@ class CometDecimalArithmeticViewSuite extends CometTestBase {
...
+    val left = "CAST(id AS DECIMAL(10, 0))"
+    val right = "CAST(id + 1 AS DECIMAL(10, 0))"
+    val operations: Seq[(String, Boolean, String, ExprOuterClass.Expr => Boolean, Boolean)] = Seq(
```
andygrove: This tuple has two unlabeled booleans in it, and `("divide TRY", true, ..., false)` is hard to read without counting positions back to the destructuring on the next block. A small case class with named fields, or moving `ansiEnabled` and `failOnError` next to their names, would make this easier to extend when someone adds a case later.

andygrove (commented):
Thanks for picking this up. I think the direction here is better than what the issue proposed. Centralizing the wrapper in `DecimalPrecision.promote` also fixes the same duplication for `Add`, `Subtract`, `Multiply`, and `Remainder`, which #5190 never mentioned.

I checked the native side to make sure removing the serde wrapper is safe. Decimal `Divide` plans to a `decimal_div` `ScalarFunctionExpr` rather than a `WideDecimalBinaryExpr`, so the surviving `CheckOverflow` is never eliminated by the fast path in `planner.rs`. The `i128::MAX` sentinel handling is intact. I also worked through the collapse rule against `transformUp` and it converges at any nesting depth, since each level collapses on the way up.

A few things I would like to see addressed before this merges. I left them inline, plus two notes on the PR description below.

The rationale is that the duplicate wrapper does the precision scan twice on every successful decimal division batch. Could you add a rough before and after number to confirm the win is real? Something from `tpcds-micro-benchmarks/` over a decimal-division-heavy query would be enough.

It would also be worth noting in the description that you took the opposite approach from the one proposed in #5190, and why. I think your choice is the better one, but a reader comparing the issue to the diff will be confused without that note.

CI is fully green across 3.4, 3.5, 4.0, 4.1, 4.2, TPC-DS, TPC-H, and rust-test.

spark/src/main/scala/org/apache/spark/sql/comet/DecimalPrecision.scala
peterxcli: filed. https://github.com/apache/datafusion-comet/issues/5248

spark/src/test/spark-4.1+/org/apache/spark/sql/comet/CometDecimalArithmeticViewSuite.scala
peterxcli: - moved to `CometDecimalPromotionSuite`. 
- make ANSI also exercise for `add`, `subtract`, `multiply`, and `remainder`
- addded `promote(promote(e)) == promote(e)`

spark/src/test/spark-4.1+/org/apache/spark/sql/comet/CometDecimalArithmeticViewSuite.scala
peterxcli: refactored.

spark/src/main/scala/org/apache/comet/serde/arithmetic.scala
peterxcli: added a sanity check.

spark/src/test/scala/org/apache/spark/sql/comet/CometDecimalPromotionSuite.scala
```diff
@@ -0,0 +1,104 @@
...
+      divide,
+      Operation(name = "remainder", symbol = "%", hasProto = _.hasRemainder))
+    val legacy = TestMode(name = "LEGACY", ansiEnabled = false, failOnError = false)
+    val ansi = TestMode(name = "ANSI", ansiEnabled = true, failOnError = true)
+
+    def check(operation: Operation, mode: TestMode, arithmetic: String): Unit = {
+      val name = s"${operation.name} ${mode.name}"
+      withSQLConf(SQLConf.ANSI_ENABLED.key -> mode.ansiEnabled.toString) {
+        val plan = spark
+          .sql(s"SELECT array_contains(array($arithmetic), $arithmetic) FROM range(1, 4)")
```
andygrove, line 55: One thing worth guarding against here. The re-promotion this test exercises only happens because `CometArrayContains` serializes its children through the public `exprToProto`. That is exactly what #5248 proposes to change. Once someone switches `arrays.scala` to `exprToProtoInternal`, the `array_contains` half of this test stops covering the duplicate-wrapper path and keeps passing silently.

Could you add a short comment naming that dependency and linking #5248, so whoever picks up that work knows this regression needs re-pointing? The `promote(promote(e)) == promote(e)` assertion below is independent and will still hold, so it is only the proto-shape half that is at risk.

andygrove (commented):
Thanks for working through all four of the earlier comments. The guard in `CometDivide.getSupportLevel` mirrors `promote`'s pattern exactly, since `DecimalExpression.unapply` returns `Some` only when the dataType is decimal, so the guard covers precisely the shapes promotion would skip. The suite move is confirmed by CI, it now runs in the `[exec]` group on 3.4 through 4.2. The ANSI coverage across all five operators is the coverage I wanted, since the collapse guard compares `nullOnOverflow`. And the explicit `promote(promote(e)) == promote(e)` assertion states the property outright.

I also traced the removal of the serde wrapper for regressions and did not find one. No serde constructs a decimal `Divide`, so every decimal `Divide` reaching `CometDivide.convert` arrived through `exprToProto` and has been promoted. The end-to-end behavior is covered by the existing SQL file tests, `decimal_div.sql` has legacy overflow to null and `decimal_div_ansi.sql` has `expect_error(NUMERIC_VALUE_OUT_OF_RANGE)` for both `/` and `div` plus `try_divide` overflow to null. Those are the paths that would break if the `i128::MAX` sentinel leaked, and they pass.

#5248 is a good writeup. It correctly notes that not all 64 call sites should change, and that some serialize independent or synthesized roots.

CI is green apart from `Spark SQL Tests (Spark 4.1) / spark-sql-sql_hive-2`, which is `HiveExternalCatalogVersionsSuite` aborting on the mirror download. That is unrelated to this PR.

Two things from my earlier review are still open on the PR description.

First, the note about approach. This PR does the opposite of what #5190 proposed. It removes the serde wrapper and keeps promotion as the single owner, rather than removing the promotion wrapper. I think your choice is the better one, and it is what lets the same fix cover add, subtract, multiply, and remainder. But a reader comparing the issue to the diff will be confused without a sentence saying so.

Second, the numbers. I asked for a rough before and after over a decimal-division-heavy query from `tpcds-micro-benchmarks/`. If you have run one, please add it. If the difference turns out to be too small to measure reliably, that is a fine answer too, but then I would rather the description lead with the idempotency and single-ownership argument and drop the framing that the duplicate scan is a meaningful cost. Either resolution works, I just do not want the perf claim sitting there unbacked.

It would also help to link #5248 in the description, not only in the review thread, so it stays visible from the merged commit.

peterxcli:
@andygrove thanks for another round of review! addressed all of your review.

> First, the note about approach. This PR does the opposite of what https://github.com/apache/datafusion-comet/issues/5190 proposed. It removes the serde wrapper and keeps promotion as the single owner, rather than removing the promotion wrapper. I think your choice is the better one, and it is what lets the same fix cover add, subtract, multiply, and remainder. But a reader comparing the issue to the diff will be confused without a sentence saying so.

added some sentence regard the difference if #5190 (original issue) and this pr's change.

> Second, the numbers. I asked for a rough before and after over a decimal-division-heavy query from tpcds-micro-benchmarks/. If you have run one, please add it. If the difference turns out to be too small to measure reliably, that is a fine answer too, but then I would rather the description lead with the idempotency and single-ownership argument and drop the framing that the duplicate scan is a meaningful cost. Either resolution works, I just do not want the perf claim sitting there unbacked.

Ran benchmark with TPC-DS SF1 `store_sales` (2,879,789 rows) on an Apple M4, it shows about 6.8% improvement in time. Also added in PR description as table.

> It would also help to link https://github.com/apache/datafusion-comet/issues/5248 in the description, not only in the review thread, so it stays visible from the merged commit.

yes, added.

> Could you add a short comment naming that dependency and linking https://github.com/apache/datafusion-comet/issues/5248, so whoever picks up that work knows this regression needs re-pointing? The promote(promote(e)) == promote(e) assertion below is independent and will still hold, so it is only the proto-shape half that is at risk.

thanks for the heads-up. added.

spark/src/main/scala/org/apache/spark/sql/comet/DecimalPrecision.scala
```diff
@@ -45,6 +45,15 @@ object DecimalPrecision {
       // happen if the Spark version is < 3.4
       case e: BinaryArithmetic if e.left.prettyName == "promote_precision" => e
```
andygrove: Now that the serde no longer adds its own wrapper, `promote` is the only thing standing between a decimal `Divide` and the raw `i128::MAX` sentinel. That makes this branch worth a second look, because it is the one arm that returns early without wrapping.

The comment says it can only fire on Spark versions below 3.4, and `QueryPlanSerde.scala:911` says the same thing about 3.3. `PromotePrecision` was removed by SPARK-39316 in 3.4.0, and `pom.xml` puts the floor at 3.4.3, so this cannot match on any version Comet builds against. It is also the one remaining shape where a decimal `Divide` reaches `CometDivide.convert` with both operands decimal and no `CheckOverflow` anywhere, which is exactly the silent-garbage case the new `getSupportLevel` guard was added to rule out. That guard cannot catch it, since the operands are decimal there.

Could this branch just be deleted? It is three lines, and removing it makes the invariant total rather than nearly total. If you would rather not touch an unrelated arm of the rule in this PR, a tracking issue linked here works too. I would just not leave it as-is, now that it is load-bearing in a way it was not before.

andygrove (commented):
Thanks for another round. All four of the earlier inline comments are genuinely addressed, and the description now carries both the approach note and the numbers.

I traced the removal of the serde wrapper once more and it holds up. `DecimalPrecision.promote` is called from exactly one place, `QueryPlanSerde.exprToProto:793`. Every `exprToProtoInternal` call site lives inside a `CometExpressionSerde.convert` or a shim, all reached from an `exprToProto` root, and `aggregates.scala` uses the public `exprToProto` at every child call site. The only two places outside the analyzer that touch a `Divide` are `CometExecRule.normalize:583`, which rebuilds it inside a Project or Filter that is serialized through `exprToProto` anyway and only rewrites float and double children, and `CometWindowExec:108`, which matches a `Divide` in order to strip it rather than serialize it. So no synthesized decimal `Divide` reaches `CometDivide.convert` unpromoted.

The new `getSupportLevel` guard lines up with `promote`'s pattern exactly, since `DecimalExpression.unapply` returns `Some` only for a decimal `dataType`. It also closes a hole the old code had, because the old wrapper was additionally conditional on `serializeDataType(expr.dataType).isDefined` and silently produced an unwrapped divide when that returned `None`. Now the `CheckOverflow` serde handles that case and falls back instead.

The collapse rule is also semantically free, not just convergent. The native `CheckOverflow` in `checkoverflow.rs` validates precision and relabels without rescaling, so applying it twice with the same target type is identical to applying it once.

On the benchmark, four of the five PR runs beat all five base runs and only the first one overlaps, which reads like warmup. Holding the native library and the revision of the Spark control fixed is the right way to show the drift floor. That is enough to back the claim.

CI is green, 61 passing and nothing failing.

One thing left, inline.

spark/src/main/scala/org/apache/spark/sql/comet/DecimalPrecision.scala
peterxcli: Thanks, addressed in https://github.com/apache/datafusion-comet/pull/5225/commits/d6625e134127b8606e7e9d9073da3edc565280af

andygrove (approved):
LGTM pending CI. Thanks @peterxcli

PR #2716, opened 2025-11-06
fix: Fix missing SortOrder fallback reason in range partitioning

andygrove:
## Which issue does this PR close?

Closes https://github.com/apache/datafusion-comet/issues/2651

PR #2323, opened 2025-09-05
fix: TakeOrderedAndProjectExec is not reporting all fallback reasons

kazuyukitanimura:
## Which issue does this PR close?

Closes #2311 

## Rationale for this change

Adding missing reasons for falling back

## What changes are included in this PR?

Added reasons

## How are these changes tested?

Existing tests

PR #5236, opened 2026-08-03
refactor: replace hand-coded rollup of expression fallback reasons onto operators

### LINKED ISSUE #5230: Replace hand-coded rollup of expression fallback reasons onto operators

Comet's serde code records fallback reasons in a `TreeNodeTag` side channel (`CometExplainInfo.FALLBACK_REASONS`). `withFallbackReasons` in `CometSparkSessionExtensions.scala` does two jobs at once: it records the node's own reasons, and it *eagerly copies* the tag values off the child nodes passed in `exprs`.

That copy is necessary because extended explain output only walks plan nodes — `ExtendedExplainInfo.sortup` follows `children`/`innerChildren`, never `expressions`. An expression-level reason is therefore invisible unless something lifts it onto the enclosing operator. Today that lifting is hand-written at roughly 200 call sites, and about 118 of them pass no message at all and exist purely to roll up:

```scala
withFallbackReason(op, op.projectList: _*)          // CometProjectExec
withFallbackReason(op, op.condition, op.child)      // CometFilterExec
withFallbackReason(expr, left, right)               // QueryPlanSerde.createBinaryExpr
```

This is repetitive and easy to get wrong. Concrete problems:

1. **Nothing forces it.** `convert` returns `Option`, so returning `None` without tagging a reason compiles fine. This has shipped as a bug at least twice, both fixed by adding the forgotten roll-up argument:
   - PR #2323 — "TakeOrderedAndProjectExec is not reporting all fallback reasons"
   - PR #2716 — "missing SortOrder fallback reason in range partitioning"

2. **The catch-all masks it.** `CometExecRule` tags a generic `"${op.nodeName} is not supported"` when an unconverted operator has no reason. A missing roll-up therefore produces a plausible-looking but useless message rather than a visible hole, which is why these slip through review.

3. **Snapshot, not reference.** Reasons are copied at call time, so anything tagged on a child *after* the parent rolled up is silently dropped. The conventional shape (convert children, then roll up in the `else` branch) is safe, but nothing marks that ordering as load-bearing.

4. **The signature gives no type safety.** `withFallbackReason[T <: TreeNode[_]](node: T, exprs: T*)` unifies `T` to `TreeNode[_]` as soon as kinds are mixed, so `withFallbackReason(op, op.condition, op.child)` typechecks with an `Expression` and a `SparkPlan` in the same varargs. There is no check that the passed nodes are related to `node` at all.

5. **It distorts serde code.** `CometExpandExec` carries a `var allProjExprs`, accumulated inside a `flatMap`, purely so it has something to hand to the roll-up.

6. **The tag is dual-purpose.** `hasFallbackReason` is a control signal, not only explain output (`CometNativeScan`, `CometShuffleExchangeExec`, `CometExecRule`). So this cannot be fixed by simply rolling up more aggressively — over-rolling changes planning decisions.

7. **The codebase already contains the better design, for the other tag.** `CometExecRule.rollUpInfoMessages` handles `EXTENSION_INFO` and `CODEGEN_DISPATCH_EXPRS` in one central place by walking `op.expressions.flatMap(_.collect { case e: Expression => e })`. Its doc comment even explains why ("explain only traverses plan nodes, not expressions"). The fallback path does the identical job 200 times by hand.

Proposed in order of increasing cost. The first three are complementary; the last two are alternatives to consider.

### Step 1 — Strict mode (cheap, independent, do first)

Replace the masking behaviour in `CometExecRule` with a test-only strictness config: when an operator is left unconverted and neither it nor any of its expressions carries a fallback reason, throw. Enable it in `CometTestBase` for all suites.

This turns the whole bug class into test failures and is a prerequisite for trusting the refactors below.

### Step 2 — Central traversal instead of per-site roll-up (the structural fix)

Mirror `rollUpInfoMessages`: add `rollUpFallbackReasons(op)` that collects `FALLBACK_REASONS` from `op.expressions.flatMap(_.collect { ... })`, and call it at the single point where `CometExecRule` decides to keep the Spark operator. Every pure roll-up call site then deletes, and serdes only write real reasons.

Two things to get right:

- Keep `hasFallbackReason` reading only the node's own tag. It is the control signal and must not observe the traversal.
- Verify the shared-instance question: `AttributeReference`s and DPP subquery expressions are shared across operators, so a reason tagged on one could surface under several. Scoping the roll-up to the operator that actually failed conversion — which is what the current code effectively does — contains this.

### Step 3 — A combinator, so the roll-up stops being a separate step

Most operator serdes have the same shape. Add a helper to `CometOperatorSerde`:

```scala
protected def serializeExprs[E <: Expression](
    op: SparkPlan, exprs: Seq[E], inputs: Seq[Attribute])(
    f: Seq[Expr] => Option[Operator]): Option[Operator]
```

It converts the expressions and, on any failure, rolls up exactly the exprs it was handed and returns `None`. `CometProjectExec`, `CometFilterExec`, `CometExpandExec` and the join serdes lose their `else` branches entirely, and the `var allProjExprs` goes away. `QueryPlanSerde.optExprWithFallbackReason` is already a partial version of this for expressions; it is just used inconsistently.

After steps 2 and 3, the only thing a serde author has to remember is to state a real reason — and step 1 enforces that.

## Additional context

Two further options, recorded for completeness:

**Return the reason instead of tagging it** — change serde conversion to `Either[Seq[FallbackReason], Expr]`. Failure then carries its reason, the framework composes child reasons into the parent's, and the compiler makes silent failure impossible; tags become a write-only rendering detail at the explain boundary. This is the right end state, but it touches both serde traits and roughly 35 files, so it is better bundled with a serde API break than done on its own.

**Ambient collector scope** — have `exprToProtoInternal` append failures to a conversion-scoped collector opened by `convertToComet`, attaching everything collected in scope to the operator on failure. Same outcome as step 2 without touching serde signatures, but it introduces hidden mutable state and needs care around reentrancy and nested AQE plans. Step 2 is preferable because it derives the same result from data already present in the tree.

andygrove:

Closes #5230 (steps 1 and 2).

## Rationale for this change

Comet records fallback reasons in a `TreeNodeTag` side channel. Extended explain output only walks plan nodes (`ExtendedExplainInfo.sortup` follows `children`/`innerChildren`, never `expressions`), so an expression-level reason is invisible unless something lifts it onto the enclosing operator.

That lifting was hand-written at roughly 200 call sites, and nothing forced it. Forgetting the roll-up argument produced a plausible-looking generic `"<operator> is not supported"` message instead of the real reason, which is why the bug slipped through review twice before (#2323, #2716). The codebase already contained the better design for the *other* tag: `CometExecRule.rollUpInfoMessages` does the identical job once, centrally.

**Step 1 — strict mode.** `CometExecRule.reportUnexplainedFallback` replaces the masking behaviour. When Comet declines an operator whose children are all already native, and neither the operator nor any of its expressions carries a reason, `spark.comet.explain.fallback.strict.enabled` (internal, default off) throws instead of tagging the generic message. `CometTestBase` enables it, so the whole test corpus now enforces it.

The check is deliberately scoped to operators a serde actually attempted. An operator with no registered handler was never attempted, so demanding a specific reason there would be wrong — it keeps the generic message. Same for an operator whose children are not native: Comet had no opportunity, so there is nothing to explain.

**Step 2 — central traversal.** `CometExecRule.rollUpFallbackReasons` collects `FALLBACK_REASONS` from `op.expressions.flatMap(_.collect { ... })` and tags them on the operator, at the single point where `convertToComet` returns `None`. Two details from the issue:

- `hasFallbackReason` still reads **only** the node's own tag. It is a planning control signal (`CometNativeScan`, `CometShuffleExchangeExec`, `CometExecRule`), not explain output, and must not observe the traversal.
- The roll-up runs only for the operator that actually failed conversion. That containment matters because `AttributeReference`s and DPP subquery expressions are shared across operators, so an unscoped roll-up could surface one expression's reason under several unrelated operators.

**API change that falls out of step 2.** With the central traversal in place the roll-up parameters are dead, so they are removed rather than left as a trap:

- `withFallbackReason[T](node: T, info: String, exprs: T*)` → `withFallbackReason[T](node: T, info: String)`
- `withFallbackReasons[T](node: T, info: Set[String], exprs: T*)` → `withFallbackReasons[T](node: T, info: Set[String])`
- the pure-roll-up overload `withFallbackReason[T](node: T, exprs: T*)` is deleted
- `QueryPlanSerde.optExprWithFallbackReason` is deleted (77 call sites; it becomes the identity once the roll-up is gone)

The compiler now rejects any attempt to reintroduce a hand-rolled roll-up, which also removes the type-safety hole in issue point 4 (the old signature unified `T` to `TreeNode[_]`, so `withFallbackReason(op, op.condition, op.child)` typechecked with an `Expression` and a `SparkPlan` in the same varargs).

Also drops the `var allProjExprs` accumulator in `CometExpandExec` (issue point 5), which existed only to have something to hand to the roll-up.

New tests in `CometExecRuleSuite`:
- an expression-level reason is rolled up onto the falling-back operator, and the generic message does *not* also appear
- the strict config is off by default and on for Comet suites
- strict mode does not fire for operators Comet never attempted

Existing suites, all passing with strict mode on: `CometExecSuite`, `CometExpressionSuite`, `CometAggregateSuite`, `CometJoinSuite`, `CometWindowExecSuite`, `CometGenerateExecSuite`, `CometExecRuleSuite`, `CometScanRuleSuite`, `CometSparkSessionExtensionsSuite`, `CometFuzzTestSuite`, `CometFuzzAggregateSuite`, `CometCastSuite`, `CometArrayExpressionSuite`, `CometStringExpressionSuite`, `CometShuffleSuite`, `CometNativeShuffleSuite`, `CometShuffleFallbackStickinessSuite`, `CometDppFallbackRepro3949Suite`, `CometCodegenSuite` — 1000 tests, 0 failures.

Compiles clean (main + test) on `spark-3.4`, `spark-3.5`, `spark-4.0` and `spark-4.1`.

Not run locally: the TPC-DS plan stability suites need `SPARK_HOME`, which isn't available in my environment. Worth watching in CI, since they assert on the rendered `[COMET: ...]` segments — the per-node attribution of a rolled-up reason can differ from the old snapshot-at-call-time behaviour.

Step 3 of #5230 (the `serializeExprs` combinator on `CometOperatorSerde`) is not included; it is independent and easier to review separately.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
@@ -382,7 +382,11 @@ case class CometExecRule(session: SparkSession)
             // when COMET_EXPLAIN_FALLBACK_LOG_ENABLED=true) even when the write is fully native.
             op
           case _ =>
-            // The operator was not converted to a Comet plan. Possible reasons for this happening:
+            // The operator was not converted to a Comet plan and no serde handler claimed it, so
+            // Comet simply has no support for it. (Operators that do have a handler are reported
+            // by `reportUnexplainedFallback` inside `convertToComet`, which is also where the
+            // strict check lives - it would be wrong to demand a specific reason here, because
+            // nothing ever attempted this operator.) Possible reasons for reaching this point:
             // 1. Comet does not support this operator.
             // 2. The operator could not be supported based on query context and current
             //    configs. In this case, it should have already been tagged with fallback
@@ -698,6 +702,20 @@ case class CometExecRule(session: SparkSession)
 
   /** Convert a Spark plan to a Comet plan using the specified serde handler */
   private def convertToComet(op: SparkPlan, handler: CometOperatorSerde[_]): Option[SparkPlan] = {
+    val converted = tryConvertToComet(op, handler)
+    if (converted.isEmpty) {
+      // Comet looked at this operator and declined it, so it stays in the Spark plan. Lift any
+      // reasons recorded on its expressions onto the operator itself - see
+      // `rollUpFallbackReasons` for why this is needed - and make sure something was recorded.
+      rollUpFallbackReasons(op)
+      reportUnexplainedFallback(op)
+    }
+    converted
+  }
+
+  private def tryConvertToComet(
+      op: SparkPlan,
+      handler: CometOperatorSerde[_]): Option[SparkPlan] = {
     val serde = handler.asInstanceOf[CometOperatorSerde[SparkPlan]]
     if (isOperatorEnabled(serde, op)) {
       // For operators that require native children (like writes), check if all data-producing
@@ -739,6 +757,57 @@ case class CometExecRule(session: SparkSession)
       }
     }
     None
+  }
+
+  /**
+   * Lift fallback reasons recorded on `op`'s expression trees onto `op` itself.
+   *
+   * Extended explain output only walks plan nodes (`ExtendedExplainInfo.sortup` follows
+   * `children` / `innerChildren`, never `expressions`), so a reason tagged on an expression is
+   * invisible unless something lifts it onto the enclosing operator. This mirrors what
+   * [[rollUpInfoMessages]] already does for the informational tags, and replaces the roll-up that
+   * used to be hand-written at every serde call site (see
+   * https://github.com/apache/datafusion-comet/issues/5230).
+   *
+   * Only child *expressions* are collected, not child operators: reasons on a child operator are
+   * already reachable by the explain traversal via `children`.
+   *
+   * Called only when `op` was left in the Spark plan, which scopes the roll-up to the operator
+   * that actually failed conversion. That matters because some expression instances
+   * (`AttributeReference`s, DPP subquery expressions) are shared across operators, so an unscoped
+   * roll-up could surface one expression's reason under several unrelated operators.
+   */
+  private def rollUpFallbackReasons(op: SparkPlan): Unit = {
+    val reasons = op.expressions
+      .flatMap(_.collect { case e: Expression => e })
+      .flatMap(_.getTagValue(CometExplainInfo.FALLBACK_REASONS))
+      .flatten
+      .toSet
+    if (reasons.nonEmpty) {
+      withFallbackReasons(op, reasons)
+    }
+  }
+
+  /**
+   * Handle an operator that Comet declined without stating why.
+   *
+   * When every child is already native, Comet had a real opportunity to convert `op`, so the
+   * absence of any reason - on `op` or anywhere in its expression trees - means a serde returned
+   * `None` and forgot to record one. Under `COMET_STRICT_FALLBACK_REASONS` (enabled for Comet's
+   * own test suites) that is a hard failure; otherwise fall back to a generic message so users
+   * still see something. The generic message is what used to mask this whole class of bug, which
+   * is why the strict check exists.
+   */
+  private def reportUnexplainedFallback(op: SparkPlan): Unit = {
+    if (op.children.forall(_.isInstanceOf[CometNativeExec]) && !hasFallbackReason(op)) {
+      if (CometConf.COMET_STRICT_FALLBACK_REASONS.get(op.conf)) {
+        throw new IllegalStateException(
+          s"Comet did not convert ${op.nodeName} but recorded no fallback reason on the " +
+            "operator or any of its expressions. Add a withFallbackReason call stating why " +
+            s"conversion failed. Operator:\n$op")
+      }
+      withFallbackReason(op, s"${op.nodeName} is not supported")
+    }
   }
 
   /**
--- a/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
@@ -706,7 +706,10 @@ case class CometExecRule(session: SparkSession)
     if (converted.isEmpty) {
       // Comet looked at this operator and declined it, so it stays in the Spark plan. Lift any
       // reasons recorded on its expressions onto the operator itself - see
-      // `rollUpFallbackReasons` for why this is needed - and make sure something was recorded.
+      // `rollUpFallbackReasons` for why this is needed - and then make sure something was
+      // recorded. The order is required, not incidental: `reportUnexplainedFallback` inspects only
+      // the operator's own tag, so a reason still sitting on an expression would look like no
+      // reason at all and trip the strict check.
       rollUpFallbackReasons(op)
       reportUnexplainedFallback(op)
     }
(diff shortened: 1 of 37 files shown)


spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -741,6 +759,57 @@ case class CometExecRule(session: SparkSession)
...
+   * roll-up could surface one expression's reason under several unrelated operators.
+   */
+  private def rollUpFallbackReasons(op: SparkPlan): Unit = {
+    val reasons = op.expressions
+      .flatMap(_.collect { case e: Expression => e })
+      .flatMap(_.getTagValue(CometExplainInfo.FALLBACK_REASONS))
+      .flatten
+      .toSet
+    if (reasons.nonEmpty) {
+      withFallbackReasons(op, reasons)
```
andygrove, line 792: We roll up the expression fallback reasons to the operator once in the framework instead of hand-coding it for every single operator

andygrove:
@parthchandra could you review this one, since you are familiar with this code?

spark/src/test/scala/org/apache/comet/rules/CometExecRuleSuite.scala
```diff
@@ -77,6 +77,64 @@ class CometExecRuleSuite extends CometTestBase {
     }.sum
   }
 
+  test("expression-level fallback reasons are rolled up onto the operator that falls back") {
```
parthchandra, line 80: Should we add a unit test that forces a handled operator to return None with no reason and asserts that an  exception is thrown in strict mode?

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -698,6 +702,20 @@ case class CometExecRule(session: SparkSession)
 
   /** Convert a Spark plan to a Comet plan using the specified serde handler */
   private def convertToComet(op: SparkPlan, handler: CometOperatorSerde[_]): Option[SparkPlan] = {
+    val converted = tryConvertToComet(op, handler)
+    if (converted.isEmpty) {
+      // Comet looked at this operator and declined it, so it stays in the Spark plan. Lift any
+      // reasons recorded on its expressions onto the operator itself - see
+      // `rollUpFallbackReasons` for why this is needed - and make sure something was recorded.
+      rollUpFallbackReasons(op)
+      reportUnexplainedFallback(op)
```
parthchandra, line 714: Is it correct to say that `reportUnexplainedFallback` relies on `rollUpFallbackReasons` to have rolled up the expression tags first? Should we add a comment to make sure a future change  from separating them?

parthchandra (approved):
lgtm with some minor comments.

andygrove:
Thanks @parthchandra. Good feedback. I addressed it.

PR #5243, opened 2026-08-03
docs: [branch-1.0] add 1.0.0 changelog

andygrove:

Part of #4082.

## Rationale for this change

The 1.0.0 release needs a change log covering everything that landed since 0.17.1. Per the release process, the change log is generated late in the cycle and added to the release branch first, then brought into `main` after the release is tagged.

- Adds `docs/source/changelog/1.0.0.md`, generated with `dev/release/generate-changelog.py 0.17.1 HEAD 1.0.0` against `branch-1.0` at `af534e0f`. It covers 244 commits from 23 contributors.
- Adds the new file to the change log `toctree` in `docs/source/changelog/index.md`.

If more changes land on `branch-1.0` before the release candidate is tagged, the change log will be regenerated and this PR updated.

Documentation only. The generated markdown was formatted with prettier so the docs style check passes, and the new page is wired into the change log index so the Sphinx build picks it up.

### DIFF
--- a/docs/source/changelog/1.0.0.md
+++ b/docs/source/changelog/1.0.0.md
@@ -1,0 +1,313 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


comphead (approved):
Thanks @andygrove

PR #4658, opened 2026-06-15
feat: Optionally split the Iceberg V2 write operator into distinct writer and committer operations

jordepic:
## Which issue does this PR close?

(Partially, first of three) Closes #4322.

## Rationale for this change

Iceberg spark writes are V2 operators and contain the functionality for writing data files, metadata files, and committing to the catalog. Ultimately, Comet is only well-positioned to just accelerate data file writing (assuming they're parquet files).  It is also crucial to ensure that the actual data file writing piece of the spark plan for iceberg writing is included within the AQE block of a spark plan, thereby ensuring that we re-plan writes in response to runtime decisions regarding its upstream operators.

Our split is fairly simple - we write the data files like normal in the "writer" operator, serialize its output, and pass it back to the "committer" operator. In the future, we'll target just the "writer" operator for speedup with iceberg-rust.

## What changes are included in this PR?

This PR contains 5 commits.

1) Docs outlining the WHOLE iceberg-write acceleration feature, not just these changes (I'm happy to modify/remove as needed).
2) Planning rules to move iceberg append and overwrite operations to our "split operator" design.
3) Planning rules to move iceberg delete, update, and merge operations to our "split operator" design.
4) Tests for part 2
5) Tests for part 3

## How are these changes tested?

We have unit tests for each operator that we're replacing that ensures that the plan shape is correct, we commit to our iceberg table the proper number of times, and our iceberg table end state is correct when we scan it after a write operation. I've been running with these changes locally now and they're all performing as expected as well.

Issue #4322, opened 2026-05-14
Writes to Apache Iceberg Tables


jordepic:
Our goal is to support writes to iceberg so that we can execute an entire ETL job using native code (falling back where we can't provide equivalent functionality to iceberg java)!

PR #5255, opened 2026-08-04
test: audit CI failures from enabling the Iceberg write split-operator plan by default

andygrove:

Relates to #4322. Follow-up experiment on top of #4658.

## Rationale for this change

**This is a CI audit sweep, not a merge candidate.** It is intentionally left red.

#4658 adds Comet's split-operator plan for Iceberg V2 writes behind
`spark.comet.write.iceberg.splitOperator.enabled`, defaulting to `false`. With the feature off,
no existing test exercises it, so we have no visibility into what breaks once it is on.

This PR rebases #4658 onto current `main` and flips that default to `true`, so the full CI
matrix — Comet's own suites, the Spark SQL suites, and the Iceberg Spark suites — runs against
the split-operator plan. The point is to enumerate every failure the feature causes so we can
scope the follow-on work needed before it can ship enabled.

Commits 1-5 are #4658 rebased onto `main` (only conflict was an additive one in
`IcebergReflection.ClassNames`).

Commit 6 is the actual change here:

- `spark.comet.write.iceberg.splitOperator.enabled` default `false` -> `true`
- `docs/source/user-guide/latest/iceberg-writes.md` updated to describe the new default

No test expectations were adjusted — the whole point is to let them fail and be catalogued.

## Known failures so far

Reproduced locally (Spark 3.5, Scala 2.13, JDK 17), all four from a single root cause:

`org.apache.comet.CometIcebergRewriteActionSuite` — 4 of 5 tests fail:

- `binPack rewrite reads each file group via CometIcebergNativeScan`
- `sort rewrite runs scan, exchange, and sort natively in Comet`
- `single-column zOrder rewrite runs scan, native exchange, and sort natively in Comet`
- `binPack rewrite applies positional and equality deletes during compaction (MOR)`

Each fails with `Expected at least one captured plan with AppendData but got none`. Iceberg's
`rewrite_data_files` action writes its compacted output through `AppendData`, which the split
operator now replaces with `IcebergWrite` under `IcebergCommit`. The rewrite itself is correct
(the row-preservation and file-count assertions pass); only the plan-shape expectation is stale.
The same suite passes 5/5 with the flag back at `false`, confirming this PR is the cause.

Likely fix for follow-on work: teach the suite to accept either write shape, rather than
matching `AppendData` alone.

By CI, deliberately. `CometIcebergWriteActionSuite` (19 tests, added by #4658) passes locally
with the new default. Iceberg 1.11 / Spark 4.1 runs on every PR, so the Iceberg Spark suites
will report against the split-operator plan without needing the `run-iceberg-tests` label; add
that label if we want 1.8 / 1.9 / 1.10 coverage too.

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -314,6 +314,7 @@ jobs:
               org.apache.comet.CometIcebergNativeSuite
               org.apache.comet.CometIcebergEncryptionSuite
               org.apache.comet.CometIcebergRewriteActionSuite
+              org.apache.comet.CometIcebergWriteActionSuite
               org.apache.comet.iceberg.IcebergReflectionSuite
               org.apache.comet.csv.CometCsvNativeReadSuite
               org.apache.comet.CometFuzzTestSuite
--- a/.github/workflows/pr_build_macos.yml
+++ b/.github/workflows/pr_build_macos.yml
@@ -130,6 +130,7 @@ jobs:
               org.apache.comet.CometIcebergNativeSuite
               org.apache.comet.CometIcebergEncryptionSuite
               org.apache.comet.CometIcebergRewriteActionSuite
+              org.apache.comet.CometIcebergWriteActionSuite
               org.apache.comet.iceberg.IcebergReflectionSuite
               org.apache.comet.csv.CometCsvNativeReadSuite
               org.apache.comet.CometFuzzTestSuite
--- a/docs/source/user-guide/latest/iceberg-writes.md
+++ b/docs/source/user-guide/latest/iceberg-writes.md
@@ -19,8 +19,9 @@
 
 # Iceberg Writes: Comet's Split-Operator Plan (Experimental)
 
-**This feature is experimental and disabled by default.** Enable it only after validating it
-against your own workloads.
+**This feature is experimental and enabled by default.** Set
+`spark.comet.write.iceberg.splitOperator.enabled=false` to restore Spark's stock combined
+write operator.
 
 ## Overview
 
@@ -45,7 +46,8 @@ planned follow-up in which Comet writes the data files natively via
 
 ## Configuration
 
-Standard Comet + Iceberg setup (see [`iceberg.md`](iceberg.md)) plus the write-side toggle:
+Standard Comet + Iceberg setup (see [`iceberg.md`](iceberg.md)) is all that is required; the
+split-operator plan is applied automatically. To turn it off:
 
 ```
 # Standard Comet / Iceberg wiring
@@ -55,8 +57,8 @@ spark.sql.catalog.<name>=org.apache.iceberg.spark.SparkCatalog
 spark.sql.catalog.<name>.type=hadoop                          # or hive / glue / rest / ...
 spark.sql.catalog.<name>.warehouse=...
 
-# Split-operator plan (experimental, off by default)
-spark.comet.write.iceberg.splitOperator.enabled=true
+# Split-operator plan (experimental, on by default); set to false to opt out
+spark.comet.write.iceberg.splitOperator.enabled=false
 ```
 
 ## Supported operations
@@ -78,7 +80,7 @@ operations is the same either way.
 
 The rewrite is skipped — and the write runs through Spark's stock combined operator — when:
 
-- `spark.comet.write.iceberg.splitOperator.enabled` is `false` (the default);
+- `spark.comet.write.iceberg.splitOperator.enabled` is `false`;
 - the write is not an Iceberg `SparkWrite` (any other V2 data source);
 - the table uses merge-on-read: delta writes (Iceberg `WriteDelta`) are not intercepted;
 - the write requires Spark's commit coordinator, which Comet's per-task commit protocol does
--- a/docs/source/user-guide/latest/iceberg-writes.md
+++ b/docs/source/user-guide/latest/iceberg-writes.md
@@ -1,0 +1,90 @@
+<!---
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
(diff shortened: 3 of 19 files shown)


andygrove:
cc @jordepic - let's see what breaks and we can file an epic for follow on work

jordepic:
Thanks @andygrove ! I'm following along :)

andygrove:
I have catalogued the CI failures from this audit run in #5259.

Run: https://github.com/apache/datafusion-comet/actions/runs/30920419488

8 failing jobs, 4 distinct root causes:

1. `CometIcebergRewriteActionSuite` (4 tests, Spark 3.4/3.5/4.0/4.1): the `AppendData` plan filter is stale, as noted in the PR description.
2. `ClassCastException: IcebergCommitExec cannot be cast to V2TableWriteExec` in Iceberg's own `TestSystemFunctionPushDownInRowLevelOperations` (7 failures on Iceberg 1.8, 12 each on 1.9/1.10/1.11).
3. 36 `TestCopyOnWriteMergeMetrics` failures on Iceberg 1.11: `spark.merge-into.num-target-rows-copied` is missing from the snapshot summary. The merge-on-read equivalents pass.
4. `TestCachedTableRefresh.testCachedTableWithSessionSchemaChangeAddColumn` on Iceberg 1.11 (2 parameterizations).

Buckets 3 and 4 are behaviour gaps rather than stale test expectations, so they need fixing before the flag can default to `true`.

One caveat on reading these results: `PR Build (Linux) / Spark 4.2, JDK 17 [scans]` passing is not a signal. Iceberg is not on the 4.2 classpath, so every Iceberg suite cancels there.

jordepic:
Gonna see if I can get all 4 of these patched up. Thanks for the help guys!

andygrove:
## Audit sweep results — CI run [30920419488](https://github.com/apache/datafusion-comet/actions/runs/30920419488)

Full matrix completed on `9100a84` with `run-iceberg-tests` applied, so all four Iceberg versions
ran. **9 of 76 jobs failed**, in two distinct groups.

### Group 1 — Iceberg's own `iceberg-spark-extensions` suite: fails on every version

| Iceberg | Spark | Scala | Java | `iceberg-spark` | `iceberg-spark-extensions` | `iceberg-spark-runtime` |
| ------- | ----- | ----- | ---- | --------------- | -------------------------- | ----------------------- |
| 1.8.1   | 3.4.3 | 2.13  | 11   | pass            | **FAIL**                   | pass                    |
| 1.9.1   | 3.5.9 | 2.13  | 17   | pass            | **FAIL**                   | pass                    |
| 1.10.0  | 3.5.9 | 2.13  | 17   | pass            | **FAIL**                   | pass                    |
| 1.11.0  | 4.1.3 | 2.13  | 17   | pass            | **FAIL**                   | pass                    |

The failure is version-independent, and it is confined to `extensions`. That is a useful narrowing:
`iceberg-spark` (core read/write paths, DDL/DML, pushdown) and `iceberg-spark-runtime` (shaded-JAR
smoke test) are **green on all four versions** — so plain appends and overwrites survive the split
operator. `iceberg-spark-extensions` is where the row-level copy-on-write `UPDATE`/`DELETE`/`MERGE`
tests and the stored procedures (`rewrite_data_files`, `rewrite_manifests`, `expire_snapshots`, ...)
live, which is exactly the surface `ReplaceData` interception and procedure-driven writes touch.

I do not yet have the per-test breakdown for these four jobs — see "Blocked on" below.

### Group 2 — Comet's own `scans` bucket: fails on every profile except Spark 4.2

| Runner | Profile                     | Result   |
| ------ | --------------------------- | -------- |
| Linux  | Spark 3.4, JDK 11, Scala 2.12 | **FAIL** |
| Linux  | Spark 3.5, JDK 17, Scala 2.13 | **FAIL** |
| Linux  | Spark 4.0, JDK 21             | **FAIL** |
| Linux  | Spark 4.1, JDK 17             | **FAIL** |
| Linux  | Spark 4.2, JDK 17             | pass     |
| macOS  | Spark 4.0, JDK 17, Scala 2.13 | **FAIL** |

Root cause confirmed locally on Spark 3.5 / Scala 2.13 / JDK 17 —
`org.apache.comet.CometIcebergRewriteActionSuite`, 4 of 5 tests:

- `binPack rewrite reads each file group via CometIcebergNativeScan`
- `sort rewrite runs scan, exchange, and sort natively in Comet`
- `single-column zOrder rewrite runs scan, native exchange, and sort natively in Comet`
- `binPack rewrite applies positional and equality deletes during compaction (MOR)`

All four fail on `Expected at least one captured plan with AppendData but got none`. Iceberg's
`rewrite_data_files` action writes its compacted output through `AppendData`; the split operator
replaces that with `IcebergWrite` under `IcebergCommit`, so the `hasNode("AppendData")` filter at
`CometIcebergRewriteActionSuite.scala:207` and `:269` matches nothing. The rewrite itself is
correct — the row-preservation, file-count, and delete-application assertions all pass; only the
plan-shape expectation is stale. Control run: the same suite is **5/5 green** with the flag back at
`false`, so this PR is the cause.

Spark 4.2 passing is a red herring rather than a signal that 4.2 is unaffected: that profile pulls
`iceberg-spark-runtime-4.0` 1.10.0 (`spark/pom.xml:318-324`) because no 4.2 runtime is published, so
the rewrite-action tests most likely cancel there instead of running. Worth confirming rather than
assuming.

### Everything else is green

All other `pr_build` buckets (`exec`, `shuffle`, `expressions`) on all six profiles, `Spark SQL
Tests (Spark 3.5)`, `Spark SQL Tests (Spark 4.1)`, the Rust tests, and every lint job passed.
`CometIcebergWriteActionSuite` (19 tests, added by #4658) passes locally with the new default.

### Proposed follow-on work

1. **Teach `CometIcebergRewriteActionSuite` to accept either write shape.** Replace the bare
   `hasNode("AppendData")` filter with one that also matches `IcebergWrite`, so the suite is
   correct with the feature on or off. Small and well understood — this clears all of Group 2.
2. **Triage `iceberg-spark-extensions`.** The real work. Needs the per-test failure list before it
   can be scoped; it may be more stale plan-shape expectations, or genuine gaps in `ReplaceData`
   / procedure handling.
3. **Confirm the Spark 4.2 pass** is test-cancellation rather than real coverage.
4. **Reconsider the config category.** The entry is `CATEGORY_TESTING`; if it ships on by default
   it probably belongs under `CATEGORY_EXEC` so it appears in the user-facing config table rather
   than under "Development & Testing Settings".

### Blocked on

The Iceberg reusable workflow uploads only the native library — no test reports — so the
`iceberg-spark-extensions` failure detail exists only in the job logs, and those are not reachable
from my environment (the run's log/artifact blob host is not on my egress allowlist). If someone can
paste the failing test names from any one of the four `iceberg-spark-extensions` jobs, that is enough
to scope item 2. Separately, it may be worth having
`iceberg_spark_test_reusable.yml` upload the Gradle test reports on failure, so this is
self-service next time.

jordepic:
@andygrove I'm running iceberg-spark-extensions suite locally now.  Would you mind kicking off the suite in the meantime? Thanks for your help on the detailed audit here.

PR #5260, opened 2026-08-04
feat: support make_interval via codegen dispatch

### LINKED ISSUE #3099: [Feature] Support Spark expression: make_interval

andygrove:

Closes #3099. Part of the interval support epic #5061.

Split out of #5030, which originally carried this alongside `timestampadd` and `timestampdiff`. Those two are unrelated to `make_interval` beyond sharing the codegen-dispatch mechanism, so they now travel separately.

Note there is an alternative native implementation in #5039, which wires `make_interval` to DataFusion's `SparkMakeInterval`. The two approaches are mutually exclusive; see the trade-off below.

## Rationale for this change

`MakeInterval` has no Comet handler today, so any query using `make_interval` falls the entire operator back to Spark. It is a regular expression, not `RuntimeReplaceable`, so it reaches serde directly.

This PR routes it through the JVM codegen dispatcher rather than adding a native implementation. The dispatcher runs Spark's own generated code inside the native Comet pipeline, which keeps the operator native while guaranteeing bit-for-bit Spark compatibility. For `make_interval` specifically this covers three behaviors that are easy to diverge on natively:

- the `seconds` argument is a `Decimal(18, 6)` scaled to microseconds, with its own overflow check;
- `failOnError` (defaulting to `SQLConf.ansiEnabled`) chooses between raising `ARITHMETIC_OVERFLOW` and returning NULL, and it must be the same exception Spark raises;
- the `years * 12` and `weeks * 7` products overflow `int` independently of one another.

`make_interval` produces `CalendarIntervalType`, which Comet's columnar layer gained support for in #4898. With that in place the dispatcher can carry its output.

The trade-off against #5039: codegen dispatch gives exact Spark semantics for free but runs JVM code per batch, so it is slower than a native kernel. A native implementation is the better end state if it can match Spark on the decimal scaling and both ANSI paths. This PR is the low-risk option and can be superseded.

- `CometMakeInterval` codegen-dispatch serde in `datetime.scala`, registered in `QueryPlanSerde`'s `temporalExpressions` map.
- `make_interval.sql` and `make_interval_ansi.sql` Comet SQL file tests.
- The `make_interval` row in the Supported Spark Expressions guide.

New Comet SQL file tests run each query through both Spark and Comet, verify the results match, and verify Comet executes the expression through the dispatcher rather than falling back.

`make_interval.sql` (non-ANSI, so `failOnError` is false) covers:

- all seven arguments as columns, and each of the shorter arities filled in by `MakeInterval`'s auxiliary constructors;
- mixed literal/column arguments and all-literal arguments (the suite disables constant folding, so these still reach Comet);
- negative and zero components, and values that carry across units (25 months, 80 minutes, 299.889987 seconds);
- NULL propagation from nullable columns. It has to come from columns rather than NULL literals: `MakeInterval` is `NullIntolerant`, so `NullPropagation` rewrites any call with a literal NULL argument into a null interval literal that never reaches Comet. That literal currently fails natively, tracked separately by #5058.
- overflow returning NULL when `years * 12` exceeds the int range.

`make_interval_ansi.sql` pins `spark.sql.ansi.enabled=true` so `failOnError` is true, and asserts the overflow raises the same error Spark does for both the `years * 12` and `weeks * 7` products. Each error query sits alongside valid-input queries in the same file so the case cannot pass vacuously through a fallback.

Verified green on Spark 3.5 and 4.1 locally, along with the pre-existing `calendar_interval.sql` fixture and the `explain comet` test in `CometExpressionSuite`, both of which use `make_interval`.

### DIFF
--- a/spark/src/test/resources/sql-tests/expressions/datetime/make_interval.sql
+++ b/spark/src/test/resources/sql-tests/expressions/datetime/make_interval.sql
@@ -1,0 +1,83 @@
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file
+-- distributed with this work for additional information
+-- regarding copyright ownership.  The ASF licenses this file
+-- to you under the Apache License, Version 2.0 (the
+-- "License"); you may not use this file except in compliance
+-- with the License.  You may obtain a copy of the License at
+--
+--   http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+-- make_interval runs through the codegen dispatcher and produces CalendarIntervalType.
+-- The suite disables ANSI mode, so MakeInterval.failOnError is false here and overflow
+-- yields NULL. See make_interval_ansi.sql for the throwing path.
+-- Config: spark.comet.exec.scalaUDF.codegen.enabled=true
+
+statement
+CREATE TABLE test_make_interval(y int, mo int, w int, d int, h int, mi int, s decimal(18,6)) USING parquet
+
+statement
+INSERT INTO test_make_interval VALUES
+  (1, 2, 3, 4, 5, 6, 7.008009),
+  (30, 25, 0, -100, 40, 80, 299.889987),
+  (0, -1, 0, 1, 0, 0, -1.000000),
+  (-1, -2, -3, -4, -5, -6, -7.500000),
+  (0, 0, 0, 0, 0, 0, 0.000000),
+  (NULL, 2, 3, 4, 5, 6, 7.008009),
+  (1, 2, 3, 4, 5, 6, NULL)
+
+-- all seven arguments as columns
+query
+SELECT make_interval(y, mo, w, d, h, mi, s) FROM test_make_interval
+
+-- the shorter arities filled in by MakeInterval's auxiliary constructors
+query
+SELECT
+  make_interval(y),
+  make_interval(y, mo),
+  make_interval(y, mo, w),
+  make_interval(y, mo, w, d),
+  make_interval(y, mo, w, d, h),
+  make_interval(y, mo, w, d, h, mi)
+FROM test_make_interval
+
+-- mixed literal and column arguments
+query
+SELECT
+  make_interval(1, mo, 3, d, 5, mi, 7.008009),
+  make_interval(y, 2, w, 4, h, 6, s)
+FROM test_make_interval
+
+-- literal arguments (constant folding is disabled by the test suite)
+query
+SELECT
+  make_interval(1, 2, 3, 4, 5, 6, 7.008009),
+  make_interval(0, 0, 0, 0, 0, 0, 0),
+  make_interval(-1, -2, -3, -4, -5, -6, -7.5),
+  make_interval(0, 13, 0, 0, 25, 61, 61.5)
+
+-- NULL arguments propagate. These have to come from nullable columns rather than NULL
+-- literals: MakeInterval is NullIntolerant, so NullPropagation rewrites any call with a
+-- literal NULL argument to a null interval literal and the expression never reaches Comet.
+-- Such a literal currently fails natively rather than falling back, tracked by #5058.
+query
+SELECT
+  make_interval(y, 2, 3, 4, 5, 6, 7.008009),
+  make_interval(1, 2, 3, 4, 5, 6, s),
+  make_interval(y)
+FROM test_make_interval
+
+-- years * 12 exceeds the int range. Outside ANSI mode MakeInterval swallows the
+-- ArithmeticException and returns NULL.
+query
+SELECT
+  make_interval(200000000, 0, 0, 0, 0, 0, 0),
+  make_interval(y + 200000000, mo, w, d, h, mi, s)
+FROM test_make_interval
--- a/spark/src/test/resources/sql-tests/expressions/datetime/make_interval_ansi.sql
+++ b/spark/src/test/resources/sql-tests/expressions/datetime/make_interval_ansi.sql
@@ -1,0 +1,49 @@
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file
+-- distributed with this work for additional information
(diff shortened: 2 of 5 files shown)


peterxcli:
> The trade-off against #5039: ... . A native implementation is the better end state .... . This PR is the low-risk option and can be superseded.

Just to clarify — is this planned for 1.0.0 so we can deliver the native implementation without schedule pressure?

spark/src/test/resources/sql-tests/expressions/datetime/make_interval.sql
peterxcli, line 1: we could also add test cases which is from the review of native `make_interval` impl: https://github.com/apache/datafusion-comet/pull/5039#discussion_r3659501474

> Spark's own sql-tests/inputs/interval.sql exercises exactly this range:
> ```sql
> select make_interval(1, 2, 3, 4, 0, 0, 123456789012.123456);
> ```

peterxcli: other reference: https://github.com/apache/datafusion-comet/pull/5039#discussion_r3659505430

> Spark's own IntervalExpressionsSuite / interval.sql exercise that aren't covered yet:
> 
> - Microsecond-precision seconds like Spark's docstring example make_interval(0, 1, 0, 1, 0, 0, 100.000001) asserted directly (it's currently only exercised via the column path where it can be hard to spot a per-row precision drift).
> - Nulls in components other than years in the column path (currently only the years=NULL row is tested).
> - Large-second cases from Spark's interval.sql: make_interval(1, 2, 3, 4, 0, 0, 123456789012.123456) and make_interval(0, 0, 0, 0, 0, 0, 1234567890123456789). If either is a known divergence (see the nanos-overflow comment on the Rust file), wrapping them in query ignore(<tracking issue>) would at least pin the behavior for future readers.
> - Int.MinValue for a signed-overflow smoke test on the years column.

spark/src/test/resources/sql-tests/expressions/datetime/make_interval_ansi.sql
peterxcli, line 1: ditto as `make_interval.sql.
ref: https://github.com/apache/datafusion-comet/pull/5039#discussion_r3659506420

> Spark's IntervalExpressionsSuite ANSI mode block covers weeks = Int.MaxValue, and per-row overflow via hours/mins/seconds interactions. Something like:
> ```sql
> query expect_error(overflow)
> SELECT make_interval(0, 0, 2147483647)
> ```

andygrove:
> > The trade-off against #5039: ... . A native implementation is the better end state .... . This PR is the low-risk option and can be superseded.
> 
> Just to clarify — is this planned for 1.0.0 so we can deliver the native implementation without schedule pressure?

Hi @peterxcli This work was part of a PR that was created before your PR. We can review both and see what makes sense (maybe even hybrid). Let me know when your PR is ready for another review.

andygrove:
superseded by https://github.com/apache/datafusion-comet/pull/5039

PR #4730, opened 2026-06-25
perf: unwrap identity casts in schema adapter to enable Parquet stats pruning

mbutrovich:
## Which issue does this PR close?

Closes #.

## Rationale for this change

DataFusion's default `PhysicalExprAdapter` inserts a `CastExpr` around every Column reference whenever the logical and physical Arrow Fields differ in any attribute, including metadata-only or nullability-only mismatches. DataFusion itself absorbs this because its `PruningPredicate` analyzer recognizes its own `CastExpr` and peels it to resolve the column against parquet statistics.

Comet's `SparkPhysicalExprAdapter::replace_with_spark_cast` then swaps that `CastExpr` for a `datafusion_comet_spark_expr::Cast` because Spark cast semantics diverge from arrow-cast for overflow, null handling, ANSI mode, etc. The Spark `Cast` is a different `PhysicalExpr` type that DataFusion's pruning analyzer does not understand, so `build_pruning_predicates` returns `None` at file open time and no row groups are pruned. With Spark range-derived schemas (non-nullable logical) read from Parquet (nullable physical), this fires on every column reference, silently disabling row-group and page-index stats pruning.

For identical source and target data types there is no Spark-specific cast semantics to preserve, so the swap costs us pruning for no benefit.

## What changes are included in this PR?

`SparkPhysicalExprAdapter::replace_with_spark_cast` now skips the Spark `Cast` wrap when the physical and target data types are equal. Unwrapping is safe because a `Cast` with equal source and target types is a value-level identity (it does not null-strip or enforce non-null), and Arrow field nullability and metadata are informational, not computational.

## How are these changes tested?

`CometNativeReaderSuite` adds a regression test that writes a 1000-row Parquet file with `parquet.block.size=1024`, asserts more than one row group via `ParquetFileReader`, then runs `SELECT ... WHERE c1 > 500` and asserts the scan's `numOutputRows` is strictly less than 1000. Without the fix the scan reads all rows; with the fix row groups whose max is at most 500 are pruned.

PR #5262, opened 2026-08-04
deps: bump DataFusion 55.0 and Arrow/Parquet 59.2

### LINKED ISSUE #4859: Comet native Parquet scan reads full nested columns despite pruned Spark ReadSchema
### Describe the bug

Comet's native Parquet scan appears to ignore nested schema pruning for complex columns. In a production Spark 3.5.6 comparison job, plain Spark and Comet produced the same Spark physical-plan `ReadSchema`, same input row count, same selected files, and same downstream shuffle output, but Comet read/decompressed far more data during the scan.

Spark 3.5.6, Comet 0.17.0

The query reads Parquet from HDFS and writes Parquet back to HDFS. Column names below are anonymized. The relevant scan reads four top-level fields:

```text
is_flagged
source_type
events
event_hour
```

The Spark UI shows the same pruned nested schema under `events` for both plans:

```text
ReadSchema: struct<
  is_flagged:boolean,
  source_type:string,
  events:array<struct<
    is_available:boolean,
    event_time_ms:bigint,
    is_active:boolean,
    items:array<struct<
      group_id:bigint,
      entity_id:bigint,
      has_amount:boolean,
      item_type:string,
      metric_value:double,
      actor_id:bigint,
      is_skipped:boolean
    >>
  >>,
  event_hour:timestamp
>
```

The physical input schema is much wider than the requested schema. A representative anonymized shape is:

```text
InputSchema: struct<
  is_flagged:boolean,
  source_type:string,
  event_hour:timestamp,
  dimension_id:bigint,
  region_code:string,
  client_type:string,
  ingestion_time:timestamp,
  events:array<struct<
    is_available:boolean,
    event_time_ms:bigint,
    is_active:boolean,
    event_token:string,
    source_name:string,
    secondary_id:bigint,
    debug_flags:array<string>,
    latency_parts:struct<
      queue_time_ms:bigint,
      fetch_time_ms:bigint,
      render_time_ms:bigint,
      retry_count:int
    >,
    items:array<struct<
      group_id:bigint,
      entity_id:bigint,
      has_amount:boolean,
      item_type:string,
      metric_value:double,
      actor_id:bigint,
      is_skipped:boolean,
      auxiliary_id:bigint,
      class_code:string,
      mode:string,
      raw_score:double,
      normalized_score:double,
      reason_codes:array<string>,
      audit_blob:binary,
      feature_map:map<string,double>,
      diagnostics:struct<
        module_id:string,
        module_version:string,
        trace_id:string,
        extra_payload:binary
      >
    >>
  >>,
  raw_payload:binary,
  metadata_json:string
>
```

The expected behavior is to read only the requested nested leaves, not all child fields under `events` / `items`.

However, the observed scan metrics were very different:

| Metric | Plain Spark scan | Comet native scan |
| --- | ---: | ---: |
| Spark version | 3.5.6 | 3.5.6 |
| SQL duration | 120,642 ms | 1,688,780 ms |
| Scan tasks | 20,012 | 20,012 |
| Files selected | 10,300 | 10,300 |
| Input records | 310,897,758 | 310,897,758 |
| Stage input bytes | 30,945,027,031 | 1,351,791,550,588 |
| Shuffle records written | 157,270,190 | 157,270,190 |
| Shuffle bytes written | 3,216,674,929 | 3,216,674,929 |
| Comet `bytes_scanned` metric | n/a | 1259.0 GiB |

`size of files read = 2.2 TiB` appeared in both UIs, but that is the selected file length metric, not physical bytes fetched. The physical read evidence is Spark stage `inputBytes` and Comet's native `bytes_scanned` metric.

This strongly suggests the logical Spark-side pruning is present, but Comet's native reader still reads a much wider nested Parquet schema for the projected complex column.

### Steps to reproduce

1. Use Spark 3.5.x with Comet native Parquet scans enabled.
2. Read a Parquet dataset with a wide nested column, for example an `array<struct<...>>` with many child fields.
3. Run a query that only needs a subset of nested fields from that complex column.
4. Compare against plain Spark using the same query and selected files.
5. In Spark UI, verify both plans show the same pruned `ReadSchema`.
6. Compare stage input bytes and Comet native scan `Number of bytes scanned`.

Minimal shape of the reproducer:

```scala
val df = spark.read.schema(prunedSchema).parquet(path)

df
  .filter(!$"is_flagged")
  .selectExpr(
    "event_hour",
    "posexplode_outer(events) as (event_idx, event)"
  )
  .filter("event.is_available and (event.is_active is null or event.is_active)")
  .selectExpr(
    "event_hour",
    "posexplode_outer(event.items) as (item_idx, item)",
    "event.event_time_ms"
  )
  .filter("item.entity_id is not null")
  .groupBy(
    $"event_hour",
    $"item.actor_id",
    $"item.group_id",
    $"item.entity_id"
  )
  .count()
  .write.mode("overwrite").parquet(outputPath)
```

The important part is that the query projects only selected child fields from a nested Parquet column while the original physical Parquet files contain many more children under that same top-level column.

### Expected behavior

Comet native Parquet scan should honor Spark's pruned nested `ReadSchema` and avoid reading/decompressing unrequested nested child fields, matching Spark's Parquet scan behavior as closely as possible.

For this workload, Comet should not read ~1.35 TB of stage input when plain Spark reads ~30.9 GB for the same files, same rows, and same displayed `ReadSchema`.

### Additional context

Local code inspection points to the native V1 Parquet scan path using full `data_schema` plus a top-level projection vector, instead of giving DataFusion a nested-pruned Parquet read schema.

Relevant code path:

- Spark writes the pruned `requiredSchema` into Parquet read config:
  - `sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala`
  - `ParquetReadSupport.SPARK_ROW_REQUESTED_SCHEMA`
- Spark clips nested Parquet schemas in:
  - `sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetReadSupport.scala`
  - `clipParquetSchema`, `clipParquetType`, `clipParquetListType`, `clipParquetMapType`
- Comet serializes both the pruned schema and the full data schema:
  - `spark/src/main/scala/org/apache/comet/serde/operator/CometNativeScan.scala`
  - `requiredSchema = schema2Proto(scan.requiredSchema.fields)`
  - `dataSchema = schema2Proto(scan.relation.dataSchema.fields)`
  - `projectionVector` is built with `scan.relation.dataSchema.fieldIndex(field.name)`, so it is top-level only.
- Native planner passes both schemas into the native scan:
  - `native/core/src/execution/planner.rs`
  - `init_datasource_exec(required_schema, Some(data_schema), ..., Some(projection_vector), ...)`
- Native Parquet setup chooses the full `data_schema` as the base schema when present:
  - `native/core/src/parquet/parquet_exec.rs`
  - `(Some(schema), Some(proj)) => (Arc::clone(schema), Some(proj.clone()))`
  - then applies `with_projection_indices(Some(projection))`

That can prune top-level columns, but it does not appear to prune nested leaves inside a projected complex column such as `events`.

Possible fixes / workarounds:

1. Workaround: set `spark.comet.scan.enabled=false` for this workload. That keeps Spark's Parquet scan path and avoids the native scan behavior.
2. Optional workaround to test: combine `spark.comet.scan.enabled=false` with `spark.comet.convert.parquet.enabled=true`, so Spark performs the Parquet read and Comet can still convert the resulting batches after the scan.
3. Code fix direction: for native V1 Parquet scan, initialize DataFusion's Parquet reader with the nested-pruned `required_schema` when Spark has already produced the read schema, instead of using full `data_schema` with only a top-level projection vector.
4. Add a regression test with `array<struct<...many fields...>>`, select only a subset of nested fields, verify `CometNativeScanExec` is used, and assert the native scan does not read unrequested nested leaves.

### LINKED ISSUE #4865: chore: DataFusion 55.0.0

DataFusion tracker:
https://github.com/apache/datafusion/issues/22393

54.0.0 migration #3978 

## Internal dependencies 
- [ ] #4680 
- [x] #4131 
- [ ] #5040 
- [ ] https://github.com/apache/datafusion-comet/pull/4587
- [ ] https://github.com/apache/datafusion-comet/issues/5210
- [ ] https://github.com/apache/datafusion-comet/pull/4957 maybe

_No response_

### Additional context

_No response_

mbutrovich:
DataFusion 55.0.0 is released, so this branch now pins the published crates rather than a git commit and is no longer blocked. Along the way it doubled as a testbed for DataFusion and upstream-Comet fixes that needed validation against the new version before it was released, most notably apache/datafusion#24090 below.

Closes #4865. Closes #4859.

## Rationale for this change

Bumping DataFusion this far ahead of a release requires fixing API breakage and validating that fixes landing upstream in the meantime, like apache/datafusion#24090 (nested Parquet leaf pruning, for #4859), actually work end to end against Comet.

- Bump DataFusion to 55.0.0, Arrow/Parquet to 59.2.0, and iceberg-rust/opendal to compatible revisions (native/Cargo.toml, native/Cargo.lock, native/core/Cargo.toml).
- API-compat fixes for the new arrow-rs/DataFusion versions: `FixedSizeBinaryArray::from` to `try_from`, `ShufflePartitioner: Send + Sync` to `Send`, `GroupsAccumulator` trait changes (`merge_batch` drops `_opt_filter`, new required `convert_to_state`), the now required `ExecutionPlan::apply_expressions`, arrow's `CompressionContext` to `IpcWriteContext`, `UnnestOptions::preserve_nulls` becoming a method, and the iceberg-rust builder/getter APIs for `FileScanTask`. No behavior change.
- Track spilled bytes in `BufBatchWriter` itself rather than via `writer_stream_position()`: DataFusion 55's spill writer is a `Box<dyn SpillWriter>`, which is not `Seek`.
- Add `spark_sqrt`, a Comet-only scalar UDF (native/spark-expr/src/math_funcs/sqrt.rs, wired in comet_scalar_funcs.rs and spark/.../serde/math.scala). DataFusion 55 added `validate_sqrt_input`, so `sqrt(-1)` now errors where DataFusion 54 returned NaN, while Spark's `Sqrt` returns NaN. This is required by the bump, not incidental to it.
- #4859: Comet's `SparkPhysicalExprAdapter` swaps DataFusion's `CastExpr` for Comet's own `CometCastColumnExpr` whenever a cast touches a Struct, List, or Map, to apply Spark-specific semantics via `spark_parquet_convert`. #24090's leaf-pruning in `build_projection_read_plan` only recognizes the literal `CastExpr` type, so once Comet swaps it, pruning never fires and the whole column gets read regardless of the DataFusion bump. Same failure shape as #4730, which hit an analogous problem with row-group pruning.

  Fix: add `is_pure_structural_narrowing` (native/core/src/parquet/schema_adapter.rs). DataFusion's own `datafusion_common::nested_struct::cast_column`, which `ColumnarValue::cast_to` runs at execution time, already matches struct fields by name, drops extra source fields, and null-fills missing target fields, exactly the shape a pruned nested ReadSchema needs. For a pure structural narrowing (fields dropped only, no leaf-type change, no Map, no Parquet field-id matching, no case-insensitive-only match), that's identical to what Comet's own path computes. When the predicate holds, `replace_with_spark_cast` leaves DataFusion's `CastExpr` in place instead of swapping in `CometCastColumnExpr`, so #24090's pruning sees it.

  Built as an allow list (only Struct/List/LargeList recurse, every leaf must be an exact type match), not a deny list, since a deny list fails open: a future Comet special case not mirrored there would silently produce wrong results instead of just missing an optimization. Map narrowing is still uncovered (`nested_struct::cast_column` has no Map arm), needs a Comet-side accommodation or an upstream extension.

- Hoist the two regression tests from #4866 (array of struct and array of struct of struct leaf pruning) into CometNativeReaderSuite.
- Iceberg: derive a task's partition spec, partition type, partition values and required source columns from one list of live partition fields, so they cannot drift apart. Presence of a required field id is now tested with `Schema.findField`, which descends into structs, so a partition source nested inside a struct is no longer appended a second time as a top-level column. A partition field whose source column was dropped resolves to Iceberg's `unknown` type and is now filtered out of the serialized spec and the required source ids too, not just out of the partition type and values. This fixes 10 failures across the two Iceberg CI jobs (`Multiple entries with same key: 4=nested.b and 4=b` in `TestPartitionValues`, and `Invalid schema: multiple fields for name category: 4 and 2` in `TestAlterTablePartitionFields`).
- Update the Celeborn nullable-map-key assertion for Arrow 59.2 (#5697, thanks @pingzh).
- Skip the signed-zero cases in the `array_distinct` / `array_union` / `array_except` / `array_intersect` SQL fixtures. DataFusion collapses `-0.0` and `0.0`, which SPARK-54918 makes the correct answer, but that fix is only in Spark 4.2.0, 4.1.4 and 4.0.5. Tracked as #5701.
- Ignore two `CometAggregateSuite` metric tests whose `peak_mem_used` expectations changed under apache/datafusion#22710. Tracked as #5703.
- CometTPCHQuerySuite: raise MEMORY_OFFHEAP_SIZE to 4g, CI tuning found necessary while running this branch's suites. Whether this reflects a real memory requirement or the same accounting change is tracked as #5704.

Review findings that are deliberately not addressed here, so this PR stays scoped to the bump, are collected under #5709.

- 17 new Rust unit tests in schema_adapter.rs covering `is_pure_structural_narrowing` (allowed: struct and list-of-struct-of-struct narrowing; denied: case-insensitive-only match, missing target field, zero field overlap, Parquet field-id matching at top and nested levels, Map, Dictionary, timestamp NTZ to LTZ relabeling, leaf type change), plus two end-to-end tests through the real `PhysicalExprAdapter::rewrite` pipeline and two tests comparing Comet's `spark_parquet_convert` against the actual DataFusion functions this relies on for byte-identical output.
- The two `#4859` regression tests in CometNativeReaderSuite pass; full CometNativeReaderSuite passes too, to catch any regression in other array/struct/map shapes from swapping fewer casts.
- Three new `CometIcebergNativeSuite` regression tests for the partition metadata fix: a table partitioned by a nested source column, a v1 table with a live and a dropped partition field, and a v1 table where a dropped partition source column is re-added under the same name. Each one fails with the original error when the fix is reverted.
- Existing test suites cover the API-compat fixes and spark_sqrt.


andygrove:
I had a similar effort here - https://github.com/apache/datafusion-comet/pull/4888

Happy to close mine though

sunchao:
Hi @mbutrovich @andygrove is this work still active? 

It looks like the latest DF comes with several improvements related to nested Parquet projection and selective decoding (some are in 55 and some are in main), which will be quite useful for Comet when processing Parquet nested data.

comphead:
I'll resolve conflicts

comphead:
2 metrics tests failed. looks like they come from #5423 

@sunchao are you ok to ignore those tests for now and adapt to DF 55 in follow up PR.

the `peak_mem_used` changed in apache/datafusion#22710

sunchao:
> are you ok to ignore those tests for now and adapt to DF 55 in follow up PR.

Yes it's OK for me. I can open a follow-up PR to fix this once the DF 55 PR is merged.

comphead:
bunch of array tests failed likely after added in https://github.com/apache/datafusion-comet/pull/5393

failed looks like some regression was made, I think before latest upmerge the test passed

Also might be related to https://github.com/apache/datafusion-comet/pull/5403

comphead:
Comet works fine in terms of literals, however for the same input in parquet the Spark has different result

```
scala>   spark.sql("SELECT array_distinct(arr) FROM t_distinct").show(false)
+-------------------+
|array_distinct(arr)|
+-------------------+
|[0.0, -0.0, 1.0]   |
+-------------------+


scala> spark.sql("select array_distinct(array(-0.0, 0.0, 1.0))").show(false)
+------------------------------------+
|array_distinct(array(0.0, 0.0, 1.0))|
+------------------------------------+
|[0.0, 1.0]                          |
+------------------------------------+

```

@andygrove @sunchao wondering if its expected Spark behavior? 


UPD:
This is a Spark bug tracked in https://issues.apache.org/jira/browse/SPARK-54918 and should be addressed in Spark 4.2
The same was done in DF earlier https://github.com/apache/datafusion/pull/22835

I'm planning to move those +0, -0 tests separately with `query fallback`

comphead:
@andygrove @sunchao folks may I ask you to wait other PR merges until this one merged?
Keep resolving conflicts and tests

andygrove:
The `-0.0` handling worries me more than the skipped tests suggest. `array_intersect` and `array_except` already report `Incompatible` from `getSupportLevel`, so those only reach the native path under `allowIncompatible`. But `ArrayDistinct` goes through a plain `CometScalarFunction` and `ArrayUnion` through `CometArrayUnion`, and neither overrides `getSupportLevel`, so both run natively by default and now return a different answer from Spark whenever a float or double array contains `-0.0`. I checked the branch out and ran it against Spark 4.1:

```
SELECT array_distinct(array(0.0, double('-0.0'), 1.0))
Spark:  [0.0, -0.0, 1.0]
Comet:  [0.0, 1.0]
```

The plan is `CometProject` with no fallback. SPARK-54918 fixes this in Spark 4.2, but we support 3.4 through 4.1 where nothing will. Could we add a `getSupportLevel` override on the `ArrayDistinct` and `ArrayUnion` serdes returning `Incompatible` when the element type is `FloatType` or `DoubleType`, in the same shape `CometArrayExcept` already uses for its unsupported element types? That keeps the fast path for every other element type and makes the divergence opt-in rather than silent. I'd also like a Comet issue that the `ignore(...)` markers can point at, so the skips are findable from our side and not only via the Spark JIRA.

Related to that: the comment on the literal case says Spark and Comet agree there, but the query above *is* the literal case and it fails. `NormalizeFloatingNumbers` only rewrites grouping keys, join keys, window partition specs and `Distinct`, not a plain `SELECT`, and the optimized plan keeps the `-0.0` literal intact. The same "literals are fine, only parquet columns diverge" framing appears in the `array_except`, `array_intersect` and `array_union` files too. Worth correcting so nobody re-enables these on the strength of the comment.

Could we also file an issue for the two ignored `CometAggregateSuite` metric tests and reference it in the TODO? A prose TODO is easy to lose track of, and `peak_mem_used` changing under apache/datafusion#22710 seems worth tracking on our side.

On the off-heap bump in `CometTPCHQuerySuite`, is 2g to 4g because DF 55 genuinely needs more memory for these queries, or is it a consequence of that same `peak_mem_used` accounting change? If it is the former then it seems worth a changelog note, since anyone who sized their off-heap against TPC-H would be under-provisioned after upgrading.

Two smaller things in the shuffle and copy paths. `SpillFile::path()` returning `Option` now folds two different situations into one `None` in `SpillWriter::path()`: nothing was spilled, and the backend has no local path. `local_partition_writer.rs` treats both as "skip the copy", so a pathless spill backend would silently drop spilled data while the recorded offsets keep advancing. It cannot happen today since `create_tmp_file` always hands back a `RefCountedTempFile` and that impl always returns `Some`, but `TempFileFactory` is pluggable. Would it be worth keeping the two cases apart so a future backend fails loudly instead? Separately, `copy_array` is the one place the `try_extend` migration goes back to panicking, via `.expect`. Both callers are already in `Result` context, so making it return `Result<ArrayRef>` looks cheap and would turn an offset overflow into a task failure rather than a JVM abort, which is what arrow deprecated `extend` for.

One correction for the description: `spark_sqrt` is not unrelated to the bump. DF 55 added `validate_sqrt_input`, so `sqrt(-1)` now errors where DF 54 returned NaN. That is precisely why it has to be in this PR, so it is worth saying.

native/Cargo.toml
```diff
@@ -35,18 +35,18 @@ license = "Apache-2.0"
...
+arrow-select = { version = "59.2.0" }
 async-trait = { version = "0.1" }
 bytes = { version = "1.11.1" }
-parquet = { version = "58.4.0", default-features = false, features = ["experimental"] }
-datafusion = { version = "54.1.0", default-features = false, features = ["unicode_expressions", "crypto_expressions", "nested_expressions", "parquet"] }
-datafusion-datasource = { version = "54.1.0" }
-datafusion-physical-expr-adapter = { version = "54.1.0" }
-datafusion-spark = { version = "54.1.0", features = ["core"] }
+parquet = { version = "59.2.0", default-features = false, features = ["experimental"] }
+datafusion = { version = "55.0.0", default-features = false, features = ["unicode_expressions", "crypto_expressions", "nested_expressions", "parquet"] }
```
ziting-openai, line 46: [P2] Finish migrating the remaining DF55/Arrow consumers

The native library no longer compiles with this version: `ExplodeExec` and `IcebergWriteExec` still omit the now-required `ExecutionPlan::apply_expressions`, and `explode.rs:898` still accesses `UnnestOptions::preserve_nulls` as a field. Both operators are compiled unconditionally. The [current-head Rust CI job](https://github.com/apache/datafusion-comet/actions/runs/33820438498/job/100862299933) confirms E0046/E0615, plus stale `preserve_nulls` test initializers and `CompressionContext` in the shuffle-scan test. Please port these remaining consumers and test helpers to the new APIs; updating the other execution plans is not enough to make this dependency bump build.

native/Cargo.toml
```diff
@@ -62,8 +62,8 @@ object_store = { version = "0.13.2", features = ["gcp", "azure", "aws", "http"]
 url = "2.2"
 aws-config = "1.8.18"
 aws-credential-types = "1.2.13"
-iceberg = { git = "https://github.com/apache/iceberg-rust", rev = "8adaa872f31549dd5ad8255848715758228038bc" }
-iceberg-storage-opendal = { git = "https://github.com/apache/iceberg-rust", rev = "8adaa872f31549dd5ad8255848715758228038bc", features = ["opendal-memory", "opendal-fs", "opendal-s3", "opendal-gcs", "opendal-oss", "opendal-azdls"] }
+iceberg = { git = "https://github.com/apache/iceberg-rust", rev = "665c64e48e8d33797ecb1a421f327edd9b024879" }
```
ziting-openai, line 65: [P2] Adapt Comet to the newly pinned Iceberg task API

At this revision, `FileScanTask` fields are private and `FileScanTaskDeleteFile` requires `file_format`, but Comet still reads/mutates `task.deletes` in `iceberg_scan.rs:270,323`, constructs `FileScanTask` with a struct literal in `planner.rs`, and omits `file_format` from the delete-task initializer. Iceberg is a nonoptional dependency and these modules are unconditional, so this blocks native compilation even when no Iceberg query is used. The [current-head Rust CI job](https://github.com/apache/datafusion-comet/actions/runs/33820438498/job/100862299933) reports E0616/E0063 for these call sites. Please migrate task construction/access to the pinned builder/getter APIs and supply the delete-file format, including the affected tests.

ziting-openai (commented):
Reviewed head `96e88b6a0fa2e73aae39c0432bc6645e1975b73f`. Three new significant P2 findings are attached. The previously reported [signed-zero P1](https://github.com/apache/datafusion-comet/pull/5262#issuecomment-5457402136) also remains, so approval is held. Source review and current-head CI logs were checked; no local build, test suite, or benchmark was run.

native/core/src/parquet/schema_adapter.rs
```diff
@@ -77,6 +78,91 @@ fn schema_has_field_ids(schema: &SchemaRef) -> bool {
...
+                // additionally matches case-insensitively when `case_sensitive` is false,
+                // which could resolve a field DataFusion would instead treat as missing (and
+                // null-fill). Requiring an exact match sidesteps that divergence regardless
+                // of the `case_sensitive` setting, and also sidesteps the missing-field
+                // nullability divergence: DataFusion errors when a non-nullable target field
+                // is missing from the source, whereas Comet null-fills unconditionally.
+                source_fields
+                    .iter()
+                    .find(|f| f.name() == target_field.name())
+                    .is_some_and(|source_field| {
```
ziting-openai: [P2] Reject ambiguous nested names before retaining the generic cast

With `spark.sql.caseSensitive=false`, a Parquet column `s:struct<ID:bigint,id:bigint>` read through the explicit schema `s:struct<id:bigint>` passes this exact-name lookup. I traced the adapter retaining `CastExpr`, which can prune `ID` and bypass the duplicate-field check in `spark_parquet_convert`; Spark and the previous Comet path reject this ambiguous match instead of returning one field. The existing `CAFÉ`/`café` native-reader regression requests `Café`, so it misses the case where one sibling matches exactly. Please require uniqueness under the configured case-insensitive resolver at each nested struct level before allowing this cast, and add the same native-scan regression requesting `café` (or `id`) exactly.

native/Cargo.toml
ziting-openai: Update for `3ee7fa65b2f5e7018b5c60f4088a57eba2b9934f`: adding top-level partition source fields addresses the reported projected-task-schema failure, including the ordinary metadata scan used by partitioned `DELETE`. The merged-partition-type fixture also now avoids the invalid spec-without-values task while preserving its pool-order assertions. These fixes are verified at source level; current-head CI was still running when checked.

The mixed live/dropped v1 partition case remains. A v1 table partitioned by `[identity(region), identity(category)]` can remove the trailing category partition, write under its retained `void` spec, then drop the category column. `serializePartitionData` still filters the now-UNKNOWN category value from `task.spec().partitionType()` but serializes the full real spec because region remains. The new task-schema union does not rebind or change that Java task spec or the serialized values. The builder therefore still rejects one value versus two spec fields. Please keep the serialized spec and values consistent for this supported evolution path, preserving field order/IDs and the spec ID.

This remains the existing Iceberg task-migration P2. The dropped-field case is source-traced; no local execution was performed. The earlier [scan CI](https://github.com/apache/datafusion-comet/actions/runs/33824701130/job/100880184873), [DELETE CI](https://github.com/apache/datafusion-comet/actions/runs/33824701130/job/100880832892), and [Rust fixture failure](https://github.com/apache/datafusion-comet/actions/runs/33824701130/job/100875289278) were from prior head `598f5e979bdaf82b929596dfcae6ec116bae265e` through synthetic merge `27cfb46bbf6309823fcfa3d90a56140fe15c898f`, not executions of the new head.

native/Cargo.toml
ziting-openai: Update for `3ee7fa65b2f5e7018b5c60f4088a57eba2b9934f`: the compiler migrations and obsolete `collect_set` drift sentinel are fixed. [Rust CI](https://github.com/apache/datafusion-comet/actions/runs/33891332915/job/101084453021) passed all 1,131 executed tests, including the updated collect-aggregate and Iceberg fixtures (four skipped). The revised aggregate test retains the `collect_list` negative check and positive coerced-state checks for both aggregates.

The two nullable-map-key diagnostic assertions still fail at `CometCelebornShuffleReaderSuite.scala:1343`, now reproduced by [Linux Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/33891332915/job/101093022179), [Linux Spark 4.2](https://github.com/apache/datafusion-comet/actions/runs/33891332915/job/101093022178), and [macOS Spark 4.0](https://github.com/apache/datafusion-comet/actions/runs/33891332915/job/101094487398). Each job's only two failures are the JVM-reader and native-ShuffleScan nullable-key cases; the preceding empty-row assertion succeeds, but the expected `type mismatch` substring is absent. The source trace remains that Arrow 59.3 rejects the nullable key field during IPC decoding with `Map key field must not be nullable` before Comet's later type check. Please adapt the diagnostic expectation while preserving no-output and fetch-failure checks.

This remains one DF/Arrow test-migration P2, not a new production correctness finding. These jobs tested synthetic merge `282cc8faa482c4254a3173cfd7087e1130128712`; the relevant test, decoder and dependency blobs match the reviewed head. The Rust fixtures/helper/parser function bodies also match. No local tests were run.

comphead:
ooooph, now Iceberg is broken

native/core/src/parquet/schema_adapter.rs
comphead: it might be a good follow up ticket

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
```diff
@@ -1082,6 +1082,19 @@ object CometIcebergNativeScan extends CometOperatorSerde[CometBatchScanExec] wit
...
+                // the task schema, so a task carrying a partition spec needs that spec's source
+                // columns present even when the query projects them out (e.g. selecting only
+                // _spec_id / _partition). Union them in. project_field_ids still drives the read,
+                // so these columns are not materialized into the output.
+                val schema: AnyRef =
+                  IcebergReflection
+                    .schemaWithRequiredFields(
+                      baseSchema,
+                      metadata.table,
+                      IcebergReflection.partitionSourceFieldIds(task, fileScanTaskClass))
```
ziting-openai, line 1060: **Withdrawn by Codex:** This feedback was posted by a misconfigured AI agent and should be disregarded. Sorry for the unsolicited review. Non-OpenAI repositories are now excluded from the agent’s review scope.

<details>

[P2] Preserve valid schemas when adding partition source fields

This new call can construct an invalid schema in two supported cases. First, for `s:struct<region:string>` partitioned by `identity(s.region)`, the full table schema already contains the nested source ID, but `schemaWithRequiredFields` checks only top-level IDs through `buildFieldIdMapping`. It appends the same field at the top level, so Iceberg's `Schema` constructor rejects the duplicate ID. Even an ordinary `SELECT id` without deletes reaches this path. The [Iceberg Spark CI job](https://github.com/apache/datafusion-comet/actions/runs/33891332915/job/101092878196) now reproduces this in nine parameterized cases across `testNestedPartitionValues`, `testReadPartitionColumn`, and `testPartitionedByNestedString`, including `Multiple entries with same key: 2=struct.string and 2=string`.

Second, [current Iceberg 1.11 CI](https://github.com/apache/datafusion-comet/actions/runs/33891332915/job/101092878262) fails `testReaddColumnAfterIdentityPartitionDrop` at this call with `Invalid schema: multiple fields for name category: 4 and 2`. The v1 test drops the category partition field and column, re-adds category under a new ID, inserts a row, then selects it. The original task spec still references the old void source ID 2; this union restores its historical field alongside current category ID 4. The partition serializer already omits the UNKNOWN value and sends an empty spec for this case, so requiring that discarded source introduces the collision.

Please derive required sources from the partition representation actually sent to native, and merge fields with recursive ID checks while preserving valid names and nesting. These failures propagate from execution-time `serializePartitions`, without fallback. Both classes of collision were observed on synthetic merge `282cc8faa482c4254a3173cfd7087e1130128712`, whose relevant Comet files match this head. No local tests were run.

</details>

ziting-openai (commented):
Re-reviewed head `3ee7fa65b2f5e7018b5c60f4088a57eba2b9934f`. One new significant P2 is attached for nested partition-source schema construction. One existing P1 and three existing significant P2s also remain, so approval is held. The two migration follow-ups are updated to distinguish fixed issues from remaining work. This was source review; current CI was still running, and no local tests, builds or benchmarks were run.

comphead:
Celeborn tests failing

comphead:
@pingzh appreciate if you can take a quick look

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
ajsquared: The nested partition-source concern remains active at `3ee7fa65b2f5e7018b5c60f4088a57eba2b9934f`. The new `schemaWithRequiredFields` call runs even without deletes, using the full table schema. Its `buildFieldIdMapping` checks only `schema.columns()`, so an existing nested source such as `s.region` is treated as missing; `findFieldObject` retrieves that same nested field and the helper appends it at the top level, duplicating its ID. Please check IDs recursively and preserve nesting when adding missing sources. This is source-confirmed; I did not run tests or inspect CI.

ziting-openai:

I'd like to close the threads the agent opened, but GitHub does not give this account permission to resolve them. Could a maintainer please resolve [this thread](https://github.com/apache/datafusion-comet/pull/5262#discussion_r3929822782), [this thread](https://github.com/apache/datafusion-comet/pull/5262#discussion_r3929822788), [this thread](https://github.com/apache/datafusion-comet/pull/5262#discussion_r3929822797), [this thread](https://github.com/apache/datafusion-comet/pull/5262#discussion_r3935733976)? Apologies for the confusion and extra work.

pingzh:
hi @comphead , the linked pr (https://github.com/apache/datafusion-comet/pull/5697) should help to fix the CI related to celeborn in https://github.com/apache/datafusion-comet/actions/runs/33891332915/job/101093022201#step:7:1282

sunchao:
I found **four correctness issues still present** at `f536ca7f`, reviewed against `a089669d`. These overlap existing discussions; fresh tests confirm they remain.

1. **[P1] Array functions silently change results on supported Spark versions.**  
   [The signed-zero tests are skipped](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/spark/src/test/resources/sql-tests/expressions/array/array_distinct.sql#L139), but `array_distinct` and `array_union` remain enabled by default. Spark 4.1.3 returns `[0.0, -0.0, 1.0]` for the distinct example; DataFusion 55 returns `[0.0, 1.0]`. I reproduced the difference for both functions. Add compatible evaluation or type/version-aware fallback, then restore the tests. The comment claiming literal inputs are normalized is also incorrect for Spark 4.1.

2. **[P2] Nested Parquet pruning bypasses Spark’s ambiguity check.**  
   [schema_adapter.rs:132](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/native/core/src/parquet/schema_adapter.rs#L132) accepts an exact `id` match within `struct<id:bigint,ID:bigint>` when case sensitivity is disabled. The retained DataFusion cast returns `id`; Spark and Comet’s existing converter reject the ambiguity. Reproduced through the actual adapter for both structs and lists of structs. Require uniqueness under Spark’s configured name resolver before retaining the cast.

3. **[P2] Iceberg partition-source augmentation creates invalid schemas.**  
   [CometIcebergNativeScan.scala:1092](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala#L1092) calls a helper that checks only top-level IDs. A partition source already present as `s.region` gets appended again, causing a duplicate-ID exception. Dropping and re-adding `category` can instead produce conflicting same-name fields with different IDs. Both failures reproduced using the exact Scala helper and Iceberg 1.11. Resolve IDs recursively and preserve nesting; exclude partition sources that are omitted from the serialized representation.

4. **[P2] Mixed live/dropped v1 partitions fail the new task validation.**  
   [planner.rs:4540](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/native/core/src/execution/planner.rs#L4540) rejects a serialized `[identity(region), void(category)]` spec containing only the surviving region value. I passed the exact Scala-emitted protobufs into the native parser and reproduced “partition has 1 fields but partition spec has 2 fields.” Preserving a null placeholder makes the control pass. Keep partition specs and values consistent through evolution.

For **performance**, the new unconditional Iceberg helper allocated approximately **79 KB per file for a 256-column schema**, even with no required fields. Checking only required IDs avoids this work. I also reproduced an **8 MiB allocation for two visible array values** because DataFusion normalizes the entire backing buffer before slicing; slicing first reduced the largest allocation to 416 bytes. These are component allocation measurements, not query-speed results.

For **design and abstraction**, reusing DataFusion’s cast and pruning machinery is sensible, and `spark_sqrt` is a simple, appropriate compatibility implementation. The main simplification is to derive Iceberg’s effective spec, values, and required schema fields together. I found no additional actionable issue in the aggregate or shuffle API migrations.

**Validation:** 213 existing native tests passed: 103 shuffle, 61 aggregate, 46 schema-adapter, and 3 square-root tests. The added Parquet regression failed as expected. Core tests excluded optional HDFS support; full Comet/Spark integration and throughput benchmarks were not run. GitHub reports no checks for the current head. Nothing was posted, and the checkout is clean.

comphead:
> [The signed-zero tests are skipped](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/spark/src/test/resources/sql-tests/expressions/array/array_distinct.sql#L139), but array_distinct and array_union remain enabled by default. Spark 4.1.3 returns [0.0, -0.0, 1.0] for the distinct example; DataFusion 55 returns [0.0, 1.0]. I reproduced the difference for both functions. Add compatible evaluation or type/version-aware fallback, then restore the tests. The comment claiming literal inputs are normalized is also incorrect for Spark 4.1.

This should not be the case anymore as `[0.0, 1.0]` is the correct answer. Spark should report the same once  https://issues.apache.org/jira/browse/SPARK-54918 backported to 4.0

sunchao:
Ah good point! my Codex is out-dated on this

sunchao (approved):
We can probably address these remaining issues in follow-ups

andygrove:
> We can probably address these remaining issues in follow-ups

Funny timing. I just filed https://github.com/apache/datafusion-comet/issues/5709 for the follow on work. I'd like to get this merged to unblock other work waiting on the upgrade. Thanks for the approval!

andygrove:
I have merged the latest `main` into this branch and fixed the two Iceberg failures, so the remaining review items are now tracked rather than outstanding. Summary of where things stand.

**Upmerge.** `main` at 93c3234 merged in as 12540de. Only two files conflicted, both in the shuffle write path, and both were additive on each side:

- `buf_batch_writer.rs`: #5568 added the caller-owned scratch buffer and its debug-only `scratch_addr` identity check, while this branch added `total_bytes_written` because DataFusion 55's `Box<dyn SpillWriter>` is not `Seek`. Both fields kept.
- `spill.rs`: `recycled_buffer` is threaded through `write`/`flush`, and the spilled-bytes count comes from `bytes_written()` instead of the `writer_stream_position()` delta DataFusion 55 can no longer take.

`shared_buffer_spill_metrics_do_not_depend_on_input_batching` from #5628 covers exactly that accounting and passes, along with the rest of the 118 native shuffle tests.

**Iceberg (2d1e7bf).** The 10 failures across the two Iceberg jobs had one root cause: partition metadata was derived in three places that could disagree.

`schemaWithRequiredFields` tested whether a required field id was already present using `buildFieldIdMapping`, which walks only `Schema.columns()`. A partition source nested inside a struct therefore looked missing, and `findField`, which does descend, resolved it and appended it a second time as a top-level column. That is the `Multiple entries with same key: 4=nested.b and 4=b` failure in `TestPartitionValues`. Presence is now tested with `findField`.

Separately, a partition field whose source column was dropped resolves to Iceberg's `unknown` type. Those were filtered out of the serialized partition type and values, but not out of the spec, which came straight from `PartitionSpecParser.toJson`, nor out of the source ids unioned into the task schema. iceberg-rust rejects a task whose partition value count differs from its spec's field count, and a stale source id resolved out of schema history could reintroduce a historical column that collides by name with a re-added one, which is `testReaddColumnAfterIdentityPartitionDrop` failing with `Invalid schema: multiple fields for name category: 4 and 2`.

The spec, partition type, partition values and required source ids now all come from one `IcebergReflection.livePartitionFields` list. The spec is filtered positionally with its spec id preserved, which also subsumes the previous special case for value-less tasks: with no live fields the filter yields the same empty-fields spec that case built by hand.

@sunchao, this also resolves the allocation you measured on the unconditional Iceberg helper. The old code built a field-id map for every task even when nothing was required; presence is now tested per required id, so an empty required set does no work.

Three regression tests in `CometIcebergNativeSuite` cover it: a table partitioned by a nested source column, a v1 table with a live and a dropped partition field, and a v1 table where a dropped partition source column is re-added under the same name. Each fails with the original error when the fix is reverted. One note in case it saves someone time: Iceberg cannot write new files under a spec whose partition source column is gone, since it has no accessor for it, so the mixed live/dropped case only ever arises when reading files written before the drop, and the test is built that way.

**Follow-ups.** Everything else raised in review is filed under #5709 so it is not lost when this merges:

- #5701 `array_distinct` / `array_union` and `-0.0`
- #5707 nested Parquet cast retention accepting an ambiguous case-insensitive match
- #5703 the two ignored `CometAggregateSuite` metric tests
- #5704 whether the TPC-H off-heap bump is a real requirement
- #5705 `SpillWriter::path()` conflating two cases
- #5706 `copy_array` panicking on offset overflow
- #5708 slicing arrays before normalizing
- #5702 the misleading signed-zero fixture comments

Two things on #5701 that change the shape of that fix, since the discussion above assumed a clean version split. SPARK-54918's fix versions are 4.2.0, **4.1.4 and 4.0.5**, so it was backported to maintenance releases. The build pins 4.0.4 and 4.1.3 but Comet runs against any patch in those lines, which means a compile-time shim keyed on `shims.minorVerSrc` cannot express the split and a runtime check is needed. Also, `CometSparkSessionExtensions.isSparkNNPlus` compares `SPARK_VERSION` as a string, so `>= "4.0.5"` would silently break at 4.0.10. Both are written up in the issue.

I have also refreshed the PR description, which still described the branch as blocked on unpublished crates and called `spark_sqrt` unrelated to the bump. It is required by it: DataFusion 55 added `validate_sqrt_input`, so `sqrt(-1)` now errors where 54 returned NaN.

@ziting-openai's threads were withdrawn as the output of a misconfigured agent. Three are resolved. The fourth, on `CometIcebergNativeScan.scala`, described a real problem that @ajsquared independently confirmed, and is fixed by 2d1e7bf, so it can be resolved now too.

CI is running on 2d1e7bf.

comphead:
I already addressed some of them.

comphead:
I tried to address Chao original comments
```
[P2] Nested Parquet pruning bypasses Spark’s ambiguity check.
[schema_adapter.rs:132](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/native/core/src/parquet/schema_adapter.rs#L132) accepts an exact id match within struct<id:bigint,ID:bigint> when case sensitivity is disabled. The retained DataFusion cast returns id; Spark and Comet’s existing converter reject the ambiguity. Reproduced through the actual adapter for both structs and lists of structs. Require uniqueness under Spark’s configured name resolver before retaining the cast.

[P2] Iceberg partition-source augmentation creates invalid schemas.
[CometIcebergNativeScan.scala:1092](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala#L1092) calls a helper that checks only top-level IDs. A partition source already present as s.region gets appended again, causing a duplicate-ID exception. Dropping and re-adding category can instead produce conflicting same-name fields with different IDs. Both failures reproduced using the exact Scala helper and Iceberg 1.11. Resolve IDs recursively and preserve nesting; exclude partition sources that are omitted from the serialized representation.

[P2] Mixed live/dropped v1 partitions fail the new task validation.
[planner.rs:4540](https://github.com/apache/datafusion-comet/blob/f536ca7f9bcd49b1012f2017cdcf67555b38410d/native/core/src/execution/planner.rs#L4540) rejects a serialized [identity(region), void(category)] spec containing only the surviving region value. I passed the exact Scala-emitted protobufs into the native parser and reproduced “partition has 1 fields but partition spec has 2 fields.” Preserving a null placeholder makes the control pass. Keep partition specs and values consistent through evolution.
```

now addressing conflicts, I'm not sure btw how those reported issues matches to GH issues created by Andy

andygrove:
@pingzh there are still some Celeborn failures in CI

comphead:
> @pingzh there are still some Celeborn failures in CI

Those are related to formatting, I'm creating an issue to `make format` find those issues earlier

UPD: Filed apache/datafusion-comet#5711

andygrove (approved):

andygrove:
All six `[shuffle]` job failures on the last run (Spark 3.4/3.5/4.0/4.1/4.2 on Linux, plus macOS 4.0) were the same two tests in `CometCelebornShuffleReaderSuite`, failing identically everywhere:

```
- JVM reader rejects nullable map keys before cast or import *** FAILED ***
- native ShuffleScan rejects nullable map keys before cast or import *** FAILED ***
  failure.exists(... contains "type mismatch")) was false (CometCelebornShuffleReaderSuite.scala:1393)
```

This is exactly what @pingzh already fixed in #5697: Arrow 59.2 rejects a nullable map-key field during IPC validation with `Map key field must not be nullable`, which now fires before Comet's own `validate_remote_schema` check that produces `type mismatch`. The reader still rejects the malformed frame correctly and reports the fetch failure once; only the expected message changed.

#5697 was squash-merged into `df55` on 2026-09-04 as f536ca7f9bcd49b1012f2017cdcf67555b38410d, on top of 3ee7fa65b. The branch was then rewritten: 3ee7fa65b came back as cb23c2598 with the same commit message and author date but a new hash, and the squash-merge commit was not carried over. So the fix was silently dropped from the branch even though the description still lists it.

I have cherry-picked it back as 2e0a757d976645201819c1ab0b5d9f3d0729603b (`git cherry-pick -x`, authorship preserved) and pushed to `df55`.

Worth watching on the next run: several jobs were still pending when the failures were triaged (Iceberg extensions, the Spark SQL `sql_core`/`sql_hive` shards, macOS `[expressions]`), so this may not have been the complete failure set.

Issue #4764, opened 2026-06-29
[EPIC] Consistent handling of invalid UTF-8 in native StringType (ingress policy)


andygrove:
## Background

Spark's `StringType` can hold arbitrary bytes, including sequences that are not valid UTF-8 (Spark stores them verbatim via `UTF8String`). Arrow's `Utf8`/`LargeUtf8` types require valid UTF-8, and arrow-rs `StringArray::value` hands out a `&str` via an *unchecked* decode that trusts the array's UTF-8 invariant. As a result, **any native string kernel that reads an Arrow string array containing invalid UTF-8 (via `value()`) materializes an invalid `&str`, which is undefined behaviour** (iterating chars, slicing on char boundaries, etc. can misbehave, panic, or be miscompiled).

Comet currently handles invalid-UTF-8 strings **inconsistently** across sites. This EPIC tracks giving Comet a single, coherent policy.

## Current state

**String-*producing* sites (resolved -- decode JVM-compatibly to valid UTF-8):**

| Site | Behavior |
|------|----------|
| Native shuffle `get_string` | Decodes via `decode_utf8_spark_lossy` (matches the JVM's `new String(bytes, UTF_8)` byte-for-byte, including surrogate-range cases). #4521 |
| `CAST(binary AS string)` | Same `decode_utf8_spark_lossy` (shared with shuffle). #4488 / PR #4763 |

Both produce valid UTF-8 that matches Spark's rendered output, so neither feeds the downstream-kernel UB.

**String-*ingress* sites (the gaps):**

| Site | Behavior | Problem |
|------|----------|---------|
| Native (Rust) Parquet reader | **Rejects** invalid UTF-8 (`encountered non UTF-8 data`) | Comet errors where Spark succeeds (#4121) |
| JVM -> native Arrow FFI import (`jni_api.rs` `from_ffi`) | **Passes bytes through unchecked** | Latent UB: invalid UTF-8 enters the native pipeline and downstream `value()` is unsound |

The two ingress paths are simultaneously **too strict** (reject) and **too loose** (unchecked) for the *same* input.

### Gap A -- native scan rejects invalid UTF-8 (#4121)

The arrow-rs Parquet reader validates UTF-8 and errors on non-UTF-8 `STRING` columns. Spark reads them fine. Surfaces in Spark 4.1.1's `hll.sql`; currently worked around by disabling Comet for that file.

### Gap B -- FFI import does not validate (confirmed)

[...]

PR #4763, opened 2026-06-29
fix: decode CAST(binary AS string) JVM-compatibly instead of unsafe reinterpret

andygrove:
## Which issue does this PR close?

Closes #4488.

## Rationale for this change

Native `CAST(<binary> AS string)` ran through `cast_binary_formatter`, which used `unsafe { String::from_utf8_unchecked(value.to_vec()) }` to turn non-UTF-8 bytes into a Rust `String`. `String` carries a documented invariant that its bytes are valid UTF-8, so constructing one from arbitrary bytes is undefined behaviour, and an Arrow string array that holds invalid UTF-8 is unsound for the downstream native string kernels that read it via `StringArray::value`.

Spark's `StringType` tolerates arbitrary bytes while Arrow's `Utf8` type requires valid UTF-8, so Comet cannot store the raw bytes natively. Instead, decode them the same way the JVM does.

## What changes are included in this PR?

- **Native (`cast.rs`):** `cast_binary_formatter` now decodes with `decode_utf8_spark_lossy`, which replaces each ill-formed sequence with `U+FFFD` exactly as the JVM's `new String(bytes, UTF_8)` does — including the surrogate-range cases (`ED A0..BF`) where Rust's `str::from_utf8_lossy` emits a different number of `U+FFFD`. The output is always valid UTF-8 and matches Spark's rendered result byte-for-byte, so the cast stays `Compatible` (native by default, no opt-in or gating). The `ToPrettyString` formatting path is unchanged.
- **Shared decoder:** `decode_utf8_spark_lossy` already existed in the native shuffle (#4521). Move it into the shared `spark-expr` `utils` module so the native shuffle and `CAST(binary AS string)` use a single implementation; the shuffle now imports it from there.
- **Docs:** add a "Strings with non-UTF-8 bytes" section to the compatibility guide, a "Producing strings from arbitrary bytes" note to the contributor guide (`adding_a_new_expression.md`), and update the cast expression-audit note.

The only behavior that differs from Spark is byte-preserving round-trips (e.g. `CAST(CAST(X'FF' AS STRING) AS BINARY)`): Spark keeps the original bytes, Comet has the `U+FFFD` encoding. This is consistent with the native shuffle and is documented.

## How are these changes tested?

- Native: `decode_utf8_spark_lossy` JVM-parity tests (moved alongside the function) and a `cast_binary_to_string` test that exercises the surrogate-range case; `cargo fmt` and clippy clean.
[...]

PR #5267, opened 2026-08-05
fix: preserve concat binary string bytes

manuzhang:

Relates to https://github.com/apache/datafusion-comet/issues/4180.

## Rationale for this change

`spark.sql.function.concatBinaryAsString=true` makes Spark insert `BinaryType`-to-`StringType` casts before `concat`. Spark preserves malformed bytes in those casts, while Comet's native Arrow string path normalizes them to UTF-8 replacement bytes. For example, Spark keeps `FFFE41`, but Comet produced `EFBFBDEFBFBD41`.

- Route `concat` trees containing `BinaryType`-to-`StringType` casts through Spark's generated code inside the Comet pipeline.
- Add a two-value SQL config matrix covering result bytes, malformed UTF-8, empty values, and nulls.
- Add a Scala assertion for result types and direct proof that both config values use the JVM codegen dispatcher without operator fallback.
- Record the supported behavior in the Spark configuration contributor guide.

- `make core`
- Focused `CometSqlFileTestSuite concat_binary_as_string` on Spark 3.4.3, 3.5.8, 4.0.2, and 4.1.2
- Focused `CometCodegenSuite concatBinaryAsString` on the same versions
- `./mvnw spotless:check -Pspark-4.1 -Pjdk17 -DskipTests -Dscalastyle.skip=true`
- Dynamic documentation generation for all four Spark profiles

### DIFF
--- a/spark/src/main/scala/org/apache/comet/serde/strings.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/strings.scala
@@ -258,6 +258,8 @@ object CometConcat
     with CometTypeShim
     with CodegenDispatchFallback {
   private val unsupportedReason = "CONCAT supports only string input parameters"
+  private val binaryToStringCastReason =
+    "CONCAT over BinaryType-to-StringType casts requires Spark's byte-preserving semantics"
 
   // Spark 4.0 widens Concat to accept collated strings and preserves the collation in the merged
   // result type. The native concat UDF always produces UTF8 (UTF8_BINARY semantics), so a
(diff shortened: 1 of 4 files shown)


andygrove:
Hi @manuzhang. Could you take a look at the epic https://github.com/apache/datafusion-comet/issues/4764? I think the approach in this PR goes in a different direction. Would be good to get your thoughts on the bigger picture.

spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -258,6 +258,8 @@ object CometConcat
     with CometTypeShim
     with CodegenDispatchFallback {
   private val unsupportedReason = "CONCAT supports only string input parameters"
+  private val binaryToStringCastReason =
+    "CONCAT over BinaryType-to-StringType casts requires Spark's byte-preserving semantics"
```
andygrove, line 262: Shouldn't this logic apply to all cast-binary-to-string, not just when called from concat?

andygrove (commented):
Thanks for digging into this. The diagnosis is right, and comparing bytes with `hex()` in the test file is exactly the right way to pin the behavior down.

My main question is about where the fix sits. The divergence comes from `Cast(binary AS string)`, not from `concat`. `CometCast` already mixes in `CodegenDispatchFallback` (`CometCast.scala:34`), so the equivalent change there is returning `Unsupported` for `BinaryType -> StringType` at `CometCast.scala:322` instead of `Compatible()`. That version also covers `hex(cast(b AS string))` on its own, `octet_length(cast(b AS string))`, the `CAST(CAST(X'FF' AS STRING) AS BINARY)` round trip that the compatibility guide uses as its worked example, and nested shapes like `concat(coalesce(cast(b AS string), ''), y)` that the `expr.children.exists` check does not see. As the PR stands, `hex(concat(cast(b AS string), x))` will match Spark while `hex(cast(b AS string))` still will not, which is a confusing place for users to land.

I am not saying the answer is obviously "move it to the cast", because there is a real tradeoff there. #4763 chose the native path plus `decode_utf8_spark_lossy` deliberately, so that the cast stays native for the valid-UTF-8 case, which is nearly all real data. Marking binary-to-string `Unsupported` would send every one of those casts to the JVM dispatcher. Could you say which side of that you want to take? Both are defensible. I would just like the PR to make the choice explicitly rather than land on `concat` because that is where the original report happened to come from.

Two other things.

The blast radius is wider than the description suggests. Spark's `ConcatTypeCoercion` inserts the binary-to-string casts when `concatBinaryAsString` is true, or when the children are not all binary. So `concat(binary_col, 'suffix')` picks up a `Cast(binary, StringType)` child under the default `false` and hits the new branch too, which moves a default-config query off the native path. Worth calling out in the description and the docs section, and worth a test case.

`docs/source/user-guide/latest/compatibility/index.md` will need updating either way. The "Strings with non-UTF-8 bytes" section currently tells users that this whole class of expression normalizes to `U+FFFD`, and after this change `concat` no longer does.

One note for later rather than something to fix here. The preserved bytes come back into native through `from_ffi` at `native/spark-expr/src/jvm_udf/mod.rs:239`, which is the JVM-UDF import site listed under Gap B of #4764. If Gap B ever decodes at that boundary unconditionally, this fix silently reverts and the new SQL test starts failing. I will leave a note on the EPIC so that whoever picks up Gap B knows that site now has a consumer depending on it not decoding.

PR #5268, opened 2026-08-05
test: rename CometCastSuite to CometNativeCastSuite

andygrove:

N/A — test/docs clarity.

## Rationale for this change

`CometCast` mixes in `CodegenDispatchFallback`, so a cast that `CometCast.isSupported` rejects does *not* fall back to Spark — it still runs inside the Comet operator by way of the Arrow-direct codegen dispatcher, evaluating Spark's own `Cast` against Arrow vectors.

`CometCastSuite` only covers the first half of that: native cast parity, plus the `isSupported` matrix that decides which casts reach the native kernel. The name didn't convey the split, and reading a `(from, to)` pair asserted as `Unsupported` in that suite is easy to mistake for "this cast isn't tested" rather than "this cast is tested through the dispatcher instead".

- Rename `CometCastSuite` → `CometNativeCastSuite`.
- Add a class-level scaladoc stating the scope explicitly: Spark-parity for the native `Cast` implementation and the `isSupported` matrix that gates it; codegen dispatch semantics are out of scope and belong to `CometCodegenSuite` and friends. It also notes that adding a native cast means moving a pair out of the `Unsupported` assertions and into the parity matrix.
- Annotate `"cast ArrayType(DateType) to unsupported ArrayType routes through codegen dispatch"` to say it is a boundary case — it only pins that `Unsupported` keeps the enclosing operator native — not dispatch coverage.
- Update every cross-reference to the old name: the `expressions` suite group in `pr_build_linux.yml` / `pr_build_macos.yml`, the `-Dsuites=` examples in `development.md`, a note in `spark_configs_support.md`, two SQL fixture headers under `sql-tests/expressions/cast/`, the `CometCast.scala` comment on the shared reason string, and two test comments in `native/spark-expr/src/conversion_funcs/string.rs`.

The historical entry in `docs/source/changelog/0.1.0.md` is left alone.

No test logic changed — this is a rename plus comments.

`./mvnw test -Dtest=none -Dsuites="org.apache.comet.CometNativeCastSuite valid"` — suite is discovered under the new name, 3 tests run, all pass. `./mvnw test-compile`, `spotless:apply`, `scalastyle`, `cargo fmt --check`, and `prettier` are all clean.

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -381,7 +381,7 @@ jobs:
               org.apache.comet.CometHashExpressionSuite
               org.apache.comet.CometTemporalExpressionSuite
               org.apache.comet.CometArrayExpressionSuite
-              org.apache.comet.CometCastSuite
+              org.apache.comet.CometNativeCastSuite
               org.apache.comet.CometDateTimeUtilsSuite
               org.apache.comet.CometMathExpressionSuite
               org.apache.comet.CometStringExpressionSuite
--- a/.github/workflows/pr_build_macos.yml
+++ b/.github/workflows/pr_build_macos.yml
@@ -197,7 +197,7 @@ jobs:
               org.apache.comet.CometHashExpressionSuite
               org.apache.comet.CometTemporalExpressionSuite
               org.apache.comet.CometArrayExpressionSuite
-              org.apache.comet.CometCastSuite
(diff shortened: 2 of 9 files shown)


andygrove:
@comphead hopefully this helps re your question in https://github.com/apache/datafusion-comet/pull/5185#discussion_r3693851681

comphead (approved):
Thanks @andygrove makes sense

PR #5276, opened 2026-08-05
fix: align wide-decimal ANSI overflow value with Spark (#5211)

### LINKED ISSUE #5211: Wide-decimal overflow reports a different value than Spark
### Describe the bug

In ANSI mode, Comet's wide-decimal arithmetic overflow reports a different `value` parameter from Spark. PR #5169 fixes the exception type and propagates the error class, SQLSTATE, and query context, but Comet still formats the rescaled i256 result while Spark formats its pre-`toPrecision` Decimal.

This is a follow-up to #5072.

### Steps to reproduce

Using Spark 4.1.2, run the query once with Comet disabled and once with Comet enabled:

```sql
SET spark.sql.ansi.enabled = true;

CREATE TABLE tbl (_1 DECIMAL(20, 0)) USING PARQUET;
INSERT INTO tbl VALUES (11000000000000000000);

SELECT _1 * _1 FROM tbl;
```

Spark reports:

```text
[NUMERIC_VALUE_OUT_OF_RANGE.WITH_SUGGESTION] 121000000000000000000000000000000000000 cannot be represented as Decimal(38, 6). If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error, and return NULL instead. SQLSTATE: 22003
```

Comet reports:

```text
[NUMERIC_VALUE_OUT_OF_RANGE.WITH_SUGGESTION] 121000000000000000000000000000000000000.000000 cannot be represented as Decimal(38, 6). If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error, and return NULL instead. SQLSTATE: 22003
```

### Expected behavior

Comet should match Spark's `value` parameter and report `121000000000000000000000000000000000000` without the output-scale suffix.

### Additional context

The regression in `spark/src/test/scala/org/apache/comet/CometExpressionSuite.scala` added by #5169 documents the current difference: Spark formats its pre-`toPrecision` Decimal, while Comet formats the rescaled i256 value. The test intentionally compares the structured error fields and query context without comparing this differing message parameter.

Matching Spark is not only a scale-formatting change. Spark's `Decimal.$times` multiplies with `MATH_CONTEXT`, which is `MathContext(38, HALF_UP)`, so the value in Spark's error has already been rounded to 38 significant digits before `toPrecision` is attempted. The reproduction above rounds exactly because `1.21e38` has trailing zeros; products with non-zero low digits can differ from Comet's exact i256 even after natural-scale formatting. A fix must reproduce this significant-digit rounding as well as the scale formatting.

sam-1112:

Closes #5211.

## Rationale for this change

This is a follow-up to #5072 and #5169.

Under ANSI mode, wide-decimal overflow raises `NUMERIC_VALUE_OUT_OF_RANGE`, but the `value` reported by Comet did not match Spark.

Spark reports the pre-`toPrecision` intermediate value using `toPlainString`. Comet instead formatted the already-rescaled `i256` value at the output scale, which introduced a spurious all zero fractional suffix, such as `.000000`.

For multiplication, Spark also applies `MathContext(39, DOWN)` before formatting the reported value. Without this step, values with high storage scales, such as `Decimal(38, 18)`, retained a long `.000…` suffix that Spark does not emit.

* Format ANSI overflow values using the pre-rescale intermediate value at its natural scale, rather than the rescaled output-scale value.
* For multiplication, apply Spark's `MathContext(39, DOWN)` when constructing the reported `value`.
* Parse `NumericValueOutOfRange` parameters using `java.math.BigDecimal` shims so that Scala `BigDecimal`'s default `DECIMAL128` `MathContext` does not rewrite large integer values.
* Strengthen `CometExpressionSuite` to verify that the `value` parameter matches Spark and does not contain a spurious `.000000` suffix.

> **Note:** This branch is based on #5169. If #5169 has not been merged yet, please review and merge it first, or update this PR's base branch accordingly.

* Native unit tests in `wide_decimal_binary_expr.rs` covering overflow-value formatting for addition and multiplication, including the `Decimal(38, 18)` multiplication case.
* Scalar decimal overflow tests in `CometExpressionSuite` for both legacy and ANSI modes.

### DIFF
--- a/native/core/src/execution/expressions/arithmetic.rs
+++ b/native/core/src/execution/expressions/arithmetic.rs
@@ -45,6 +45,10 @@ impl CheckedBinaryExpr {
             query_context,
         }
     }
+
+    pub(crate) fn child(&self) -> &Arc<dyn PhysicalExpr> {
+        &self.child
+    }
 }
 
 impl Display for CheckedBinaryExpr {
--- a/native/core/src/execution/planner.rs
+++ b/native/core/src/execution/planner.rs
@@ -101,6 +101,7 @@ use datafusion::physical_expr::expressions::{Literal, StatsType};
 use datafusion::physical_expr::window::WindowExpr;
 use datafusion::physical_expr::LexOrdering;
 
+use crate::execution::expressions::arithmetic::CheckedBinaryExpr;
 use crate::parquet::parquet_exec::init_datasource_exec;
 use arrow::array::{
     new_empty_array, Array, ArrayRef, BinaryBuilder, BooleanArray, Date32Array, Decimal128Array,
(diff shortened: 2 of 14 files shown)


peterxcli:
should we change this to draft? as https://github.com/apache/datafusion-comet/pull/5169 hasnt been merged.

sam-1112:
ok

PR #5280, opened 2026-08-06
refactor: delegate ANSI integer arithmetic to arrow checked kernels

### LINKED ISSUE #5092: checked_arithmetic: delegate ANSI integer arithmetic to arrow checked kernels

In `native/spark-expr/src/math_funcs/checked_arithmetic.rs`, the ANSI-mode integer branches of `checked_binary` re-implement arrow's checked numeric kernels: they call `arity::try_binary` with `add_checked`/`sub_checked`/`mul_checked`/`div_checked`, which is exactly what `arrow::compute::kernels::numeric::{add, sub, mul, div}` do for primitive integers. In addition, `checked_arithmetic_internal` materializes scalar operands with `to_array_of_size`, which the arrow kernels' `Datum` path makes unnecessary.

For integer types in ANSI mode, call the arrow kernels directly with `Datum` operands and keep the existing error remap (`ArrowError::DivideByZero` to Spark DIVIDE_BY_ZERO, other errors to Spark ARITHMETIC_OVERFLOW).

Not replaceable, must stay as is:
- Float `checked_div`: `div_checked` on floats errors on a zero divisor (pinned by `test_checked_div_by_zero`), while arrow's `div` kernel on floats wraps to inf/NaN.
- The Try-mode branch (null on overflow with a manual `BooleanBufferBuilder`): arrow has no null-on-overflow kernel.

### Additional context

Cleanup opportunity spotted in the same file: inside the Try-mode loop, the `else if` branch guarded by `is_ansi_mode` is unreachable because ANSI mode returns earlier in the function.

kazantsev-maksim:

Closes https://github.com/apache/datafusion-comet/issues/5092

## Rationale for this change

This PR refactors the module to delegate integer arithmetic in ANSI mode directly to the native arrow-rs numeric kernels, eliminating code duplication and improving maintainability while preserving the expected Spark error behavior (DIVIDE_BY_ZERO and ARITHMETIC_OVERFLOW).

1. Added ansi_arithmetic_kernel to dispatch operations directly to `arrow::compute::kernels::numeric::{add, sub, mul, div}` for integer types when eval_mode is ANSI.
2. Introduced a strongly typed `MathOp enum (Add, Sub, Mul, Div)` to replace string literals across arithmetic functions.
3. Removed unreachable dead code (the redundant else if is_ansi_mode branch) and the obsolete is_div parameter inside `checked_binary` / `try_arithmetic_kernel`.
4. Streamlined checked_arithmetic_internal using match guards `(DataType::Int... if is_ansi_mode)` to cleanly separate ANSI and Try code paths without nested conditionals.

Tested with existing unit tests

image


native/spark-expr/src/math_funcs/checked_arithmetic.rs
```diff
@@ -29,40 +30,66 @@ use datafusion::common::DataFusionError;
...
-            op
-        ))),
+        MathOp::Add => checked_binary(left, right, is_ansi_mode, |l, r| l.add_checked(r)),
+        MathOp::Sub => checked_binary(left, right, is_ansi_mode, |l, r| l.sub_checked(r)),
+        MathOp::Mul => checked_binary(left, right, is_ansi_mode, |l, r| l.mul_checked(r)),
+        MathOp::Div => checked_binary(left, right, is_ansi_mode, |l, r| l.div_checked(r)),
     }
 }
 
+fn ansi_arithmetic_kernel<T>(
```
andygrove: `ArrayRef` already implements `Datum` via `impl<T: Array> Datum for T` in `arrow-array/src/scalar.rs`, so this function does not need a type parameter at all. If you change the signature to take `&dyn Datum`, the four integer match arms below collapse into a single one:

```rust
DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64
    if is_ansi_mode =>
{
    ansi_arithmetic_kernel(&left_arr, &right_arr, op)
}
```

I tried this locally and it compiles clean with all six tests passing. It drops about 30 lines, which feels worth it given that the goal here is removing duplication. It also sets up the `Datum` change from the issue nicely, since `&dyn Datum` is what you would pass a `Scalar` through.

native/spark-expr/src/math_funcs/checked_arithmetic.rs
```diff
@@ -29,40 +30,66 @@ use datafusion::common::DataFusionError;
...
+
+    result_array.map_err(|e| match e {
+        ArrowError::DivideByZero => divide_by_zero_error().into(),
+        _ => DataFusionError::from(SparkError::ArithmeticOverflow {
+            from_type: String::from("integer"),
+        }),
+    })
+}
+
 fn checked_binary<T, F>(
```
andygrove, line 122: After this change, is the `is_ansi_mode` branch in here only reachable from the three float division arms? If so, we now have two ANSI implementations in the file, and the name does not really hint that this one is float-division-only. Would it be clearer to pull that branch out into a small dedicated helper, or at least note the float-only reachability in a comment?

native/spark-expr/src/math_funcs/checked_arithmetic.rs
```diff
@@ -189,50 +206,68 @@ fn checked_arithmetic_internal(
         (ColumnarValue::Scalar(l), ColumnarValue::Scalar(r)) => (l.to_array()?, r.to_array()?),
     };
 
-    // Rust only supports checked_arithmetic on numeric types
     let result_array = match data_type {
-        DataType::Int8 => try_arithmetic_kernel::<Int8Type>(
+        DataType::Int8 if is_ansi_mode => ansi_arithmetic_kernel(
```
andygrove: These four arms are near-identical and only differ in the primitive type used for the downcast. See my note on `ansi_arithmetic_kernel` above. Taking `&dyn Datum` lets this become one arm and removes the downcasts entirely.

native/spark-expr/src/math_funcs/checked_arithmetic.rs
```diff
@@ -189,50 +206,68 @@ fn checked_arithmetic_internal(
...
+        ),
+        DataType::Int64 => try_arithmetic_kernel::<Int64Type>(
+            left_arr.as_primitive::<Int64Type>(),
+            right_arr.as_primitive::<Int64Type>(),
             is_ansi_mode,
+            op,
         ),
-        // Spark always casts division operands to floats
-        DataType::Float16 if (op == "checked_div") => try_arithmetic_kernel::<Float16Type>(
+        DataType::Float16 if op == MathOp::Div => try_arithmetic_kernel::<Float16Type>(
```
andygrove: Could we keep the `// Spark always casts division operands to floats` comment? It is the thing that explains why these three arms are `Div`-only.

It would also help to spell out why the float arms deliberately do *not* go through `ansi_arithmetic_kernel`, because it is quite subtle. Arrow's `float_op` handles `Op::Div` with the infallible `op!` macro using `div_wrapping`, so `numeric::div` on floats returns inf instead of raising, whereas `div_checked` on floats does error on a zero divisor. Sending floats through the numeric kernel would silently break ANSI DIVIDE_BY_ZERO for double division. Right now the only thing guarding against that is one assertion in `test_checked_div_by_zero`, and I think the next person tidying this file would reasonably try to finish the refactor and hit it.

andygrove (commented):
Thanks for picking this up. The `MathOp` enum, removing the unreachable `else if is_ansi_mode` branch, and dropping the `is_div` parameter are all nice cleanups.

I checked the equivalence of the integer ANSI path locally and it holds up. In `arrow-arith`, `numeric::add` on two same-typed integer arrays goes `arithmetic_op` -> `integer_op::<T>` -> `try_op!` -> `arity::try_binary(add_checked)`, which is exactly what the old code did. `try_binary` uses `try_for_each_valid_idx`, so null slots are skipped in both versions, and the `DivideByZero` remap lines up. I applied the diff and all six unit tests in the file pass.

A few things I would like to work through before this goes in.

### The `Datum` half of #5092 is not implemented

The issue asked for two changes, and this PR does the first. The issue also asked to pass `Datum` operands so that `checked_arithmetic_internal` can stop materializing scalars with `to_array_of_size`. Right now scalar operands are still expanded to full-length arrays and then downcast, so arrow takes the array-array `try_binary` branch and never reaches the `try_unary` scalar fast path. That is the part of the issue with an actual runtime win, so as written the refactor is roughly performance-neutral.

Would you be up for handling the scalar case here too? If you would rather keep this PR focused, could you open a follow-up issue and link it from the description? Otherwise it will get lost when this PR closes #5092.

### Test coverage for the newly routed paths

This moves all four integer widths onto a new kernel, and the existing tests only exercise Int32 and Float64 with array-array inputs. It might be worth extending them a little:

- An ANSI integer divide-by-zero case. `test_checked_div_by_zero` is Float64 only, so the integer `MathOp::Div` arm, which is the newly routed path, has no divide-by-zero coverage. That is exactly where the `ArrowError::DivideByZero` remap has to be right.
- A case with a `ColumnarValue::Scalar` operand, which is the branch the `Datum` change above would touch.
- An ANSI Int64 overflow case, since only Int32 is covered today.

The benchmark in `native/spark-expr/benches/checked_arithmetic.rs` is array-array only as well, so it would not be able to show a scalar-path improvement if the `Datum` change lands.

### CI

No CI has run on this yet. Head sha `68eb53df` has zero check runs, so the workflows will need triggering before we can evaluate it.

native/spark-expr/src/math_funcs/checked_arithmetic.rs
kazantsev-maksim: Great catch! You're completely right. To address this and make the separation crystal clear, I've pulled that float-specific ANSI path out into a dedicated helper function called ansi_float_div.

kazantsev-maksim:
Thanks @andygrove for very much for the thorough review and great suggestions! I've tried to address all your feedback in the latest update.

Smallfu666:
I was looking at this while working on the `Abs` side of #5071, so this may already be on your radar.

`ansi_arithmetic_kernel` maps every non-`DivideByZero` error to `from_type: "integer"`, so Int64 overflow still comes out as `integer overflow` instead of Spark's `long overflow`. `test_ansi_int64_overflow` only checks `is_err()`, so it stays green either way.

Might be simpler to handle here than as a follow-up, but either way works.

kazantsev-maksim:
Thanks @Smallfu666, I tried to fix it.

sunchao (commented):
Thanks for the updates. I reviewed `90bb646eb9a00476a57fba616822a3f248bfc966`, including the changes since the earlier review, and found no new P1/P2 issues.

The ANSI integer path now uses `Datum` for scalar operands without batch-sized materialization. Null masking and operand order are preserved. `ansi_float_div` retains checked division-by-zero behavior instead of using Arrow's IEEE floating-point division, and Int64 overflow now reports `long overflow`.

Local validation:

- All 10 checked-arithmetic Rust unit tests passed.
- All 16 checked-arithmetic benchmark smoke cases passed. These were smoke tests, not performance measurements.
- A separate scratch harness compiled the exact head and base arithmetic modules against the pinned dependencies and passed 14,720 edge-case scenarios against both versions. These covered all four signed integer widths, ANSI/TRY modes, scalar/array combinations, null masking, slices, empty arrays, signed zero, NaN, and infinity.
- Formatting and diff checks passed.

I also checked the relevant Spark 3.5/4.0 implementations and the pinned Arrow 58.4.0/DataFusion 54.1.0 sources. I did not run the full native/JVM suites or measure speedups.

CI and CodeQL have now started and were still in progress at the final check.

sunchao (approved):
Re-reviewed `df530c78ff88dafa7eb455a250de165fdd6a0762` against `367ca64be4e62dd10dc4c932de9ca738355a45fa` and found no new P1/P2 issues. Both changed files are unchanged since the previous reviewed head, and I checked their integration with the current base.

The ANSI integer `Datum` path preserves operand order and null masking. Floating-point ANSI division retains checked divide-by-zero handling, and Int64 overflow reports `long overflow`.

Validation at this head:

- All 10 checked-arithmetic Rust unit tests passed.
- All 16 benchmark smoke cases passed. These were not performance measurements.
- A scratch harness compiled the exact head and base arithmetic modules against the pinned dependencies and passed 14,840 edge-case scenarios against both versions. Coverage included all four signed integer widths, ANSI/TRY modes, scalar/array combinations, overflow, null masking, slices, bitmap-word boundaries, empty arrays, signed zero, NaN, and infinity.
- Rust formatting and diff checks passed.

CI is complete with 63 successful checks and 8 skipped, with no failing or pending checks. The CI merge tree matches this head. I did not run the full native/JVM suites locally or measure speedups.

sunchao:
Merged, thanks @kazantsev-maksim for the contribution!

PR #5282, opened 2026-08-06
fix: propagate the Spark task ClassLoader to JVM UDF calls

### LINKED ISSUE #5281: ScalaUDF in a user jar fails with ClassCastException on SerializedLambda when dispatched from a Tokio worker
### Describe the bug

When a `ScalaUDF` (or Java UDF) is dispatched through the codegen dispatcher (`spark.comet.exec.scalaUDF.codegen.enabled=true`) and the class that captured the UDF closure comes from a user jar (`--jars` / `spark.jars`), the query fails at execution time:

```
java.lang.ClassCastException: cannot assign instance of java.lang.invoke.SerializedLambda to field
  org.apache.spark.sql.catalyst.expressions.ScalaUDF.f of type scala.Function1
  in instance of org.apache.spark.sql.catalyst.expressions.ScalaUDF
	at java.base/java.io.ObjectStreamClass$FieldReflector.setObjFieldValues(ObjectStreamClass.java:2096)
	at java.base/java.io.ObjectStreamClass$FieldReflector.checkObjectFieldValueTypes(ObjectStreamClass.java:2060)
	at java.base/java.io.ObjectStreamClass.checkObjFieldValueTypes(ObjectStreamClass.java:1347)
	at java.base/java.io.ObjectInputStream$FieldValues.defaultCheckFieldValues(ObjectInputStream.java:2679)
	...
	at org.apache.comet.udf.codegen.CometScalaUDFCodegen.lookupOrCompile(CometScalaUDFCodegen.scala:160)
	at org.apache.comet.udf.codegen.CometScalaUDFCodegen.evaluate(CometScalaUDFCodegen.scala:123)
	at org.apache.comet.udf.CometUdfBridge.evaluateInternal(CometUdfBridge.java:203)
	at org.apache.comet.udf.CometUdfBridge.evaluate(CometUdfBridge.java:122)
```

### Root cause

The `ClassCastException` is a **masked `ClassNotFoundException`**. When the classloader used for deserialization cannot resolve a lambda's *capturing class*, `ObjectInputStream` records the CNFE against the object handle, therefore skips `SerializedLambda.readResolve()`, and the raw `SerializedLambda` then fails the field-type check in `defaultCheckFieldValues`. The CNFE never surfaces.

The loader is wrong because of where the deserialization runs. `CometScalaUDFCodegen.lookupOrCompile` deserializes the bound expression with the calling thread's context classloader:

```scala
val loader = Option(Thread.currentThread().getContextClassLoader)
  .getOrElse(classOf[Expression].getClassLoader)
```

`CometUdfBridge.evaluate` is invoked from a Tokio worker thread, which attaches to the JVM through JNI. **An attached thread has no context classloader**, so this falls back to `classOf[Expression].getClassLoader` — the loader that loaded spark-catalyst and Comet, which never contains user jars. Spark's executor classloader (`MutableURLClassLoader`, which does contain them) is only installed on Spark task threads.

Measured, by returning thread identity from inside a UDF body:

| stage leaf | thread running the UDF | context classloader |
| --- | --- | --- |
| `CometNativeScan` | `Thread-28..32` (Tokio workers) | **null** |
| shuffle read / join build side | `Executor task launch worker …` | `MutableURLClassLoader` |
| native scan disabled | `Executor task launch worker …` | `MutableURLClassLoader` |

So any plan whose stage leaf is a native scan (`CometNativeScan`, `CometIcebergNativeScan`) evaluates UDFs on a Tokio worker and hits this. `CometUdfBridge` already propagates `TaskContext` across the JNI boundary for exactly this reason, but the classloader is not propagated.

This is invisible to Comet's own test suites because they register UDFs from test classes that are already on the application classpath.

`CometUdfBridge.java:180` has the same latent problem: a user-supplied `CometUDF` implementation shipped in a user jar would fail the same way at `Class.forName`.

### Steps to reproduce

Requires two conditions together:

1. The class capturing the UDF lambda is reachable only from Spark's executor classloader (a `--jars` submission), not from Comet's own classloader.
2. The UDF is evaluated on a Tokio worker, i.e. the stage's leaf is a native scan.

A self-contained failing test compiles a class holding a serializable `scala.Function1` lambda into a jar at test time and wires it in via `spark.executor.extraClassPath` (the local-mode equivalent of `--jars`: `LocalSchedulerBackend` feeds it into the executor's `MutableURLClassLoader`), then runs `SELECT hiddenUdf(s) FROM t`. It reproduces the trace above exactly. The same query with the native scan disabled passes, which isolates the failure to classloader propagation rather than class availability.

Note that `sparkContext.addJar` is *not* sufficient to set this up in local mode — the jar does not reach the local executor's classloader, and the job then fails earlier, in Spark's own task deserialization.

### Expected behavior

A UDF whose closure class lives in a user jar should execute regardless of which thread the native side dispatches from. The Spark task thread's context classloader should be captured at plan-creation time and propagated to the Tokio worker, alongside the `TaskContext` that is already propagated.

andygrove:

Closes #5281.

## Rationale for this change

A `ScalaUDF` (or Java UDF) whose closure was captured by a class from a user jar (`--jars` / `spark.jars`) fails at execution time:

```
java.lang.ClassCastException: cannot assign instance of java.lang.invoke.SerializedLambda to field
  org.apache.spark.sql.catalyst.expressions.ScalaUDF.f of type scala.Function1
	...
	at org.apache.comet.udf.codegen.CometScalaUDFCodegen.lookupOrCompile(CometScalaUDFCodegen.scala:160)
	at org.apache.comet.udf.CometUdfBridge.evaluateInternal(CometUdfBridge.java:203)
```

That `ClassCastException` is a **masked `ClassNotFoundException`**. When the deserializing classloader cannot resolve a lambda's *capturing class*, `ObjectInputStream` records the CNFE against the object handle, therefore skips `SerializedLambda.readResolve()`, and the raw `SerializedLambda` then fails the field-type check in `defaultCheckFieldValues`. The CNFE never surfaces.

The classloader is wrong because of which thread runs the deserialization. `CometUdfBridge.evaluate` is invoked from a Tokio worker, which attaches to the JVM through JNI, and **an attached thread has no context classloader**. `lookupOrCompile` therefore fell back to `classOf[Expression].getClassLoader` — the loader that loaded spark-catalyst and Comet, which never contains user jars. Spark installs the executor's `MutableURLClassLoader` on task threads only.

Measured by returning thread identity from inside a UDF body:

| stage leaf | thread running the UDF | context classloader |
| --- | --- | --- |
| `CometNativeScan` | `Thread-28..32` (Tokio workers) | **null** |
| shuffle read / join build side | `Executor task launch worker …` | `MutableURLClassLoader` |
| native scan disabled | `Executor task launch worker …` | `MutableURLClassLoader` |

So any plan whose stage leaf is a native scan (`CometNativeScan`, `CometIcebergNativeScan`) hits this on every UDF call. `CometUdfBridge` already propagates `TaskContext` across the JNI boundary for exactly this reason; the classloader was not propagated.

This is invisible to Comet's existing test suites because they register UDFs from test classes already on the application classpath.

The task thread's context classloader is now propagated to the Tokio worker, following the same path as `TaskContext`:

- `CometExecIterator` captures `Thread.currentThread().getContextClassLoader` (it is constructed on the Spark task thread) and passes it to `createPlan`.
- `Native.createPlan` gains a `classLoader` parameter; `jni_api.rs` holds it as a JNI global ref on `ExecutionContext` and hands it to the planner via `with_class_loader`.
- `PhysicalPlanner` clones it into every `JvmScalarUdfExpr`, under the same `debug_assert` invariant already used for `task_context`.
- `JvmScalarUdfExpr` passes it through the bridge call; the JNI signature gains `Ljava/lang/ClassLoader;`.
- `CometUdfBridge.evaluate` installs it as the calling thread's context classloader for the duration of the call, restoring the prior value in `finally` — the same save-and-restore shape used for `TaskContext`.

Installing it on the thread rather than passing it down to the deserializer fixes both affected lookups at once: the closure deserialization in `CometScalaUDFCodegen`, and the `Class.forName(udfClassName)` in `CometUdfBridge`, which had the same latent bug for a user-supplied `CometUDF` implementation shipped in a user jar. It also means user code inside a UDF that relies on the context classloader behaves as it does under Spark's own execution.

One line was added to the "Behavior" section of `docs/source/user-guide/latest/scala_java_udfs.md`, alongside the existing `TaskContext.get()` guarantee.

New suite `CometScalaUDFClassLoaderSuite`, which compiles a class holding a serializable `scala.Function1` lambda into a jar at test time and wires it in via `spark.executor.extraClassPath` — the local-mode equivalent of a `--jars` submission, since `LocalSchedulerBackend` feeds it into the executor's `MutableURLClassLoader`. Four tests:

1. sanity — the class is loadable from task threads and raises `ClassNotFoundException` from `classOf[Expression].getClassLoader` (the old fallback), so the fixture reproduces the production classloader topology.
1. the failing query: `SELECT hiddenUdf(s) FROM t` with a `CometNativeScan` leaf.
1. the propagation itself: the UDF body reports whether the classloader installed on whatever thread invoked it can reach the user jar.
1. control: the same UDF with the native scan disabled, which runs on the task thread and passed even before this change — isolating the failure to classloader propagation rather than class availability.

The two new behavioral tests were confirmed to fail without the fix (reverting only the loader install), reporting `MISSING|Thread-53, MISSING|Thread-28, …` and naming the Tokio workers.

Existing coverage run locally on Spark 4.1 / Scala 2.13: `CometCodegenSuite`, `CometCodegenHOFSuite`, `CometCodegenSourceSuite`, `CometCodegenFuzzSuite` plus the new suite (180 tests), and `CometExecSuite` (143 tests), which exercises the changed `createPlan` signature broadly. `cargo fmt --check` and `cargo clippy --all-targets -D warnings` are clean.

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -401,6 +401,7 @@ jobs:
               org.apache.comet.CometCodegenHOFSuite
               org.apache.comet.CometFuzzMathSuite
               org.apache.comet.CometCodegenFuzzSuite
+              org.apache.comet.CometScalaUDFClassLoaderSuite
               org.apache.comet.codegen.CometSpecializedGettersDispatchSuite
               org.apache.comet.CometStringDecodeSuite
               org.apache.comet.CometWidthBucketSuite
--- a/.github/workflows/pr_build_macos.yml
+++ b/.github/workflows/pr_build_macos.yml
@@ -217,6 +217,7 @@ jobs:
               org.apache.comet.CometCodegenHOFSuite
               org.apache.comet.CometFuzzMathSuite
               org.apache.comet.CometCodegenFuzzSuite
+              org.apache.comet.CometScalaUDFClassLoaderSuite
               org.apache.comet.codegen.CometSpecializedGettersDispatchSuite
               org.apache.comet.CometStringDecodeSuite
               org.apache.comet.CometWidthBucketSuite
--- a/docs/source/user-guide/latest/scala_java_udfs.md
+++ b/docs/source/user-guide/latest/scala_java_udfs.md
@@ -54,6 +54,7 @@ When a UDF is rejected, the reason surfaces through Comet's standard fallback di
 
 - Non-deterministic expressions referenced from the argument tree (`rand`, `uuid`, `monotonically_increasing_id`) produce per-partition sequences consistent with Spark.
 - `TaskContext.get()` inside the user function returns the driving Spark task's context.
+- The Spark task thread's context ClassLoader is propagated to the thread that runs the user function, so functions defined in jars supplied with `--jars` / `spark.jars` resolve the same way they do under Spark's own execution.
 - The user function must be closure-serializable; the same function that works with Spark's executor execution works here.
 
 ## Known limitations
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -352,14 +352,10 @@ struct ExecutionContext {
     /// cheap to clone; the underlying `Global<JObject>` releases its JNI global ref on drop
     /// via `jni`'s `Drop` impl.
     pub task_context: Option<Arc<Global<JObject<'static>>>>,
-    /// Context `ClassLoader` of the driving Spark task thread, captured at `createPlan` time.
-    /// Threaded into every JVM scalar UDF the planner builds so the JNI bridge can install it on
-    /// the Tokio worker running the UDF. Tokio workers attach through JNI and an attached thread
-    /// has no context `ClassLoader`, so without this the bridge cannot see classes from user jars
-    /// (`--jars` / `spark.jars`): deserializing a `ScalaUDF` closure captured by a user class
-    /// fails, as does resolving a user-supplied `CometUDF` implementation. `None` when no driving
-    /// Spark task is present (unit tests, direct native driver runs). The `Arc` is cheap to clone;
-    /// the underlying `Global<JObject>` releases its JNI global ref on drop.
+    /// Context `ClassLoader` of the driving Spark task thread, captured at `createPlan` time and
+    /// threaded into every JVM scalar UDF the planner builds; see `CometUdfBridge.evaluate` for why
+    /// it has to travel with the plan. `None` when no driving Spark task is present (unit tests,
+    /// direct native driver runs). Lifetime is as for `task_context` above.
     pub class_loader: Option<Arc<Global<JObject<'static>>>>,
 }
 
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -352,6 +352,15 @@ struct ExecutionContext {
     /// cheap to clone; the underlying `Global<JObject>` releases its JNI global ref on drop
     /// via `jni`'s `Drop` impl.
     pub task_context: Option<Arc<Global<JObject<'static>>>>,
+    /// Context `ClassLoader` of the driving Spark task thread, captured at `createPlan` time.
+    /// Threaded into every JVM scalar UDF the planner builds so the JNI bridge can install it on
+    /// the Tokio worker running the UDF. Tokio workers attach through JNI and an attached thread
+    /// has no context `ClassLoader`, so without this the bridge cannot see classes from user jars
+    /// (`--jars` / `spark.jars`): deserializing a `ScalaUDF` closure captured by a user class
+    /// fails, as does resolving a user-supplied `CometUDF` implementation. `None` when no driving
+    /// Spark task is present (unit tests, direct native driver runs). The `Arc` is cheap to clone;
+    /// the underlying `Global<JObject>` releases its JNI global ref on drop.
+    pub class_loader: Option<Arc<Global<JObject<'static>>>>,
 }
 
 /// Accept serialized query plan and return the address of the native query plan.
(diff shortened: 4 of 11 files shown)


mbutrovich (approved):
Thanks @andygrove! Test suggestion but I don't consider it blocking:

The PR description credits this fix with also closing a second latent bug: the `Class.forName` resolution in `CometUdfBridge.java` for a user-supplied `CometUDF` implementation shipped in a user jar. The new suite only exercises the `ScalaUDF` / `CometScalaUDFCodegen` path (`spark.udf.register` plus the codegen dispatcher). Is a `CometUDF` ever loaded from a class name outside Comet's own classpath today? `CometScalaUDFCodegen` looks like the only class in the codebase ever passed as `JvmScalarUdf.class_name` (`CometScalaUDF.scala:138` is the only `setClassName` call site), so as far as I can tell nothing currently routes a user-jar class name through that second path. If that's right, would it be worth a comment noting the second fix is defensive for a currently-unreachable path, or is there a way to reach it that a test should pin down now?

hsiang-c (approved):
Thanks Andy

comphead (commented):
Thanks @andygrove I'm checking this, the PR LGTM overall I just wondering if this class loader tricks can somehow impact if user specify `userClassPathFirst`

parthchandra (approved):
lgtm

comphead (approved):
Thanks @andygrove lets give it a try

PR #5285, opened 2026-08-07
test: cover the narrowing direction of cast_and_stamp_schema

andygrove:

Follow-up to review feedback on #5138: https://github.com/apache/datafusion-comet/pull/5138#discussion_r3732929900

## Rationale for this change

The doc comment on `cast_and_stamp_schema` claims that reconciliation follows the declared schema in both directions, and that narrowing a nullable nested child to non-null is not silently lossy because arrow's `StructArray::try_new` rejects unmasked nulls under a non-nullable field. That claim was not covered by a test, so nothing would catch it if the narrowing direction ever started producing an array whose nullability flag misreports its own contents.

One test in `native/common/src/schema.rs`, `narrowing_a_child_errors_only_when_it_holds_a_null`. It stamps a `List(Struct(nullable Boolean))` array against a schema declaring the child non-null, twice:

- null-free data narrows successfully and comes back with the declared schema
- the same shape carrying a real null in the child errors, and the message names the operator, the column index, and the dotted path `c0.element.flag`

The two cases together pin that the child's `nullable` flag alone does not decide the outcome: the data does.

No production code changes.

`cargo test -p datafusion-comet-common schema::` (10 passed).

### DIFF
--- a/native/common/src/schema.rs
+++ b/native/common/src/schema.rs
@@ -340,6 +340,57 @@ mod tests {
         ))
     }
 
+    /// The narrowing direction of the same reconciliation. Rewriting a nullable child as non-null
+    /// is only safe when the child holds no nulls, so the flag alone must not decide the outcome:
+    /// null-free data narrows, and data with a real null errors instead of producing an array whose
+    /// nullability misreports its own contents.
+    #[test]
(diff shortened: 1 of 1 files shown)


parthchandra (approved):
thanks for adding the test @andygrove

PR #4704, opened 2026-06-22
docs: restore 0.13 user guide dropped in 0.17 release

andygrove:
## Which issue does this PR close?

N/A

## Rationale for this change

The documentation update for the 0.17.0 release (#4698) rolled the published user guide version window forward and inadvertently dropped the 0.13 user guide. The 0.13 docs were still being published prior to that change, so they should be retained for reference.

## What changes are included in this PR?

Restores 0.13 in the three places the 0.17 release PR removed it:

- `docs/generate-versions.py`: add `0.13.0` back to `previous_versions`
- `docs/build.sh`: restore the `rm -rf comet-0.13` cleanup
- `docs/source/user-guide/older-versions.md`: re-add the `0.13.x` entry to the older versions toctree

## How are these changes tested?

Documentation-only change. The docs build is exercised by CI.

PR #4698, opened 2026-06-20
chore: update documentation links for 0.17.0 release

andygrove:
## Which issue does this PR close?

N/A. This is a routine documentation update following the 0.17.0 release.

## Rationale for this change

Now that Comet 0.17.0 has been published, the documentation site needs to reference 0.17.x as the current stable release, move 0.16.x into the list of older versions, and publish the 0.17.0 changelog.

## What changes are included in this PR?

Documentation links:
- `docs/generate-versions.py`: set `latest_released_version` to `0.17.0` and slide the `previous_versions` window to `["0.14.0", "0.15.0", "0.16.0"]`
- `docs/build.sh`: update the cleanup of locally cloned release branches to `comet-0.14`, `comet-0.15`, `comet-0.16`
- `docs/source/user-guide/index.md`: present 0.17.x as the current stable release
- `docs/source/user-guide/older-versions.md`: add 0.16.x and drop 0.13.x from the older versions list, and point the stable-release link at 0.17

This keeps the published guides at the current release plus the three prior versions, matching the convention used for previous releases.

Changelog:
- `docs/source/changelog/0.17.0.md`: add the 0.17.0 changelog
- `docs/source/changelog/index.md`: register 0.17.0 in the changelog index

## How are these changes tested?

Documentation-only change. Verified by building the docs locally per `docs/README.md`.

PR #4314, opened 2026-05-13
chore: update documentation links for 0.16.0 release

andygrove:
## Which issue does this PR close?

N/A. Routine docs prep for the 0.16.0 release, mirroring #4029 for 0.15.0.

## Rationale for this change

The 0.16.0 release is being prepared and the main-branch docs need to point at it as the latest released version, while bumping the snapshot to the next development cycle (0.17.0).

## What changes are included in this PR?

- `docs/generate-versions.py`: set `latest_released_version` to `0.16.0` and shift `previous_versions` to `["0.13.0", "0.14.0", "0.15.0"]`.
- `docs/build.sh`: drop the `comet-0.12` clean-up line and add one for `comet-0.15`.
- `docs/source/user-guide/index.md`: bump the snapshot entry to `0.17.0-SNAPSHOT`, add `0.16.x`, and drop `0.12.x`.

## How are these changes tested?

Covered by the documentation publishing workflow, which regenerates and deploys docs on merge to main.

PR #5290, opened 2026-08-07
chore: update documentation links for 1.0.0 release

andygrove:

N/A — release chore for 1.0.0.

## Rationale for this change

Publishes the 1.0.0 user guide and makes it the default (current stable) version in the
documentation, following the "Publishing Documentation" step of the
[release process](https://github.com/apache/datafusion-comet/blob/main/docs/source/contributor-guide/release_process.md).
Mirrors the pattern from #4698 (0.17.0) and #4314 (0.16.0).

- `docs/generate-versions.py`: `latest_released_version = "1.0.0"`, and `0.17.0` moves into `previous_versions`
  (older guides are kept, per #4704).
- `docs/build.sh`: clean up the locally cloned `comet-0.17` and `comet-1.0` branch dirs before regenerating.
  `comet-0.17` was missing because the rm list historically only tracked `previous_versions`; without a line per
  cloned version a rebuild silently reuses the stale clone from the previous run.
- `docs/source/user-guide/index.md`: current stable release is now 1.0.x, sidebar entry points at `1.0/index`.
- `docs/source/user-guide/older-versions.md`: point the "current stable release" link at 1.0 and add 0.17.x to the list.
- `docs/source/contributor-guide/release_process.md`: refresh the "Publishing Documentation" instructions, which
  still referred to `docs/source/index.rst` and the `_toc.user-guide-links-versioned` section — neither exists since
  the sidebar overhaul. The section now lists the four files actually involved.

Not included (separate follow-ups per the release process): bringing `docs/source/changelog/1.0.0.md` into `main`,
which happens after the vote passes.

`branch-1.0` exists and has generated (frozen) user guide content under `docs/source/user-guide/latest/`, including
`index.rst` and the per-Spark-version compatibility pages, so the versioned guide will render. The docs build itself
is exercised by the documentation CI workflow.

### DIFF
--- a/docs/build.sh
+++ b/docs/build.sh
@@ -29,6 +29,8 @@ rm -rf comet-0.13
 rm -rf comet-0.14
 rm -rf comet-0.15
 rm -rf comet-0.16
+rm -rf comet-0.17
+rm -rf comet-1.0
 python3 generate-versions.py
 
 # Generate dynamic content (configs, compatibility matrices) for latest docs
--- a/docs/generate-versions.py
+++ b/docs/generate-versions.py
@@ -154,6 +154,6 @@ This is **out-of-date** documentation. The latest Comet release is version {late
 if __name__ == "__main__":
     print("Generating versioned user guide docs...")
     snapshot_version = get_version_from_pom()
-    latest_released_version = "0.17.0"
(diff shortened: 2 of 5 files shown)


mbutrovich (approved):
Thanks @andygrove!

PR #5292, opened 2026-08-07
feat: full native make_interval with 2.6x faster than spark and 1.1x faster than codegen dispatch

### LINKED ISSUE #5131: [Bug] Native make_interval overflows valid time components or loses seconds precision
## Describe the bug

Native `make_interval` delegates to DataFusion `SparkMakeInterval`. Its seconds argument is coerced from Spark `Decimal(18,6)` to `Float64`, and its time components are accumulated as i64 nanoseconds. This creates two Spark compatibility gaps:

- `Float64` coercion can lose microsecond precision.
- Nanosecond accumulation gives hours, minutes, and seconds a 1000x smaller range than Spark's microsecond accumulation, so valid Spark intervals can overflow natively.

## Steps to reproduce

```sql
SELECT make_interval(0, 0, 0, 0, 2562048);
SELECT make_interval(0, 0, 0, 0, 0, 0, 999999999.000001);
SELECT make_interval(1, 2, 3, 4, 0, 0, 123456789012.123456);
```

The hours query is valid in Spark but overflows the native nanosecond calculation. The second query loses one microsecond natively. The third returns NULL in non-ANSI mode or throws in ANSI mode natively, while Spark returns a valid interval.

The native overflow cutoffs include hours >= 2,562,048 and minutes >= 153,722,868; large seconds values have the same root cause.

## Expected behavior

Native `make_interval` should preserve `Decimal(18,6)` microsecond precision and support every hours, minutes, and seconds value Spark can represent.

## Additional context

Found while reviewing #5039. Until this is fixed, `CometMakeInterval` is marked Incompatible: the compatible default uses JVM codegen dispatch, and the native path requires explicit opt-in.

Review context: https://github.com/apache/datafusion-comet/pull/5039#discussion_r3659501474

### LINKED ISSUE #5279: [Bug] Preserve CalendarInterval microseconds across Comet boundaries and native kernels
## Describe the bug

Comet maps Spark's `CalendarIntervalType` representation (`months: i32`, `days: i32`, `microseconds: i64`) to Arrow `IntervalMonthDayNano` (`months: i32`, `days: i32`, `nanoseconds: i64`). The JVM codegen-dispatch output path therefore multiplies `CalendarInterval.microseconds` by 1,000 before writing it to an `IntervalMonthDayNanoVector`.

That conversion limits otherwise valid Spark elapsed-time components to approximately 292 years. This limit applies to the microseconds component expressed through hours, minutes, or seconds, not to calendar years stored separately as months. For larger elapsed-time values, the default dispatch path aborts with an uncaught `ArithmeticException: long overflow`, even when ANSI mode is disabled.

The native path has a second, independent compatibility problem. Comet currently wraps `datafusion-spark`'s `SparkMakeInterval`, which:

- hardcodes its return type as Arrow `Interval(MonthDayNano)`;
- computes hours, minutes, and seconds in nanoseconds, imposing the same approximately 292-year elapsed-time limit; and
- accepts seconds as `Float64`, so it cannot preserve Spark's exact `Decimal(18,6)` microsecond semantics.

Introducing a lossless boundary representation alone therefore will not make native `make_interval` compatible. The native kernel must also be replaced or extended to produce the new representation using exact microsecond arithmetic.

## Steps to reproduce

Run without `spark.comet.expression.MakeInterval.allowIncompatible` so `MakeInterval` uses the default JVM codegen-dispatch path:

```sql
CREATE TABLE test_make_interval_dispatch(hours INT) USING parquet;
INSERT INTO test_make_interval_dispatch VALUES (2562048);
SELECT make_interval(0, 0, 0, 0, hours) FROM test_make_interval_dispatch;
```

Spark returns a valid `CalendarInterval`. Comet aborts while writing the result to `IntervalMonthDayNanoVector` because `Math.multiplyExact(interval.microseconds, 1000L)` overflows.

## Expected behavior

The complete fix has two parts.

### 1. Represent `CalendarIntervalType` losslessly across Comet boundaries

Use the Spark-compatible components:

```text
months: i32
days: i32
microseconds: i64
```

A tagged Arrow struct is one possible representation. The implementation must keep calendar days separate from elapsed microseconds because a calendar day can be 23, 24, or 25 hours across DST transitions.

Update the representation consistently across the protobuf type mapping, FFI, serde, JVM and native writers/readers, codegen dispatch, native shuffle, and any native expressions that consume or produce `CalendarIntervalType`.

### 2. Make the native interval kernel use the lossless representation

Replace or extend the imported `datafusion-spark` `SparkMakeInterval` kernel so it:

- returns the new Spark-compatible calendar interval representation instead of `IntervalMonthDayNano`;
- computes elapsed time directly in microseconds without an intermediate nanosecond conversion;
- preserves `Decimal(18,6)` seconds exactly without coercing to `Float64`;
- implements Spark-compatible overflow and NULL behavior for ANSI, non-ANSI, and `try_make_interval`; and
- supports the full Spark range for months, days, and elapsed microseconds.

Only after both parts land should Comet remove `spark.comet.expression.MakeInterval.allowIncompatible` and the related ignored tests.

Coverage should include boundary round trips, JVM dispatch, native execution, native shuffle, NULL inputs, negative values, microsecond precision, extreme elapsed-time values, ANSI overflow behavior, and `try_make_interval`.

## Additional context

- PR: #5039
- Default dispatch range-limit review: https://github.com/apache/datafusion-comet/pull/5039#discussion_r3722969753
- Native precision/range issue: #5131
- Spark precedent: https://github.com/apache/spark/commit/5ca6b1062887664b16b11f2bfcc015c9e616dc49

peterxcli:

Closes #5279.
Closes #5131.

## Rationale for this change

Comet represented Spark `CalendarIntervalType` as Arrow `IntervalMonthDayNano`. Converting Spark's microseconds to nanoseconds reduced the valid elapsed-time range by 1,000x, while the native `datafusion-spark` kernel also coerced `Decimal(18,6)` seconds to `Float64`, losing microsecond precision.

Spark represents calendar intervals losslessly as separate months, days, and microseconds. Comet needs the same representation across JVM/native boundaries and exact microsecond arithmetic in the native kernel.

- Represent `CalendarIntervalType` as a Spark-tagged Arrow struct containing `months: Int32`, `days: Int32`, and `microseconds: Int64`.
- Preserve that logical type through Arrow conversion, protobuf serde, FFI, JVM readers/writers, codegen input/output, Scala UDF codegen, and native execution.
- Replace the `datafusion-spark` `make_interval` wrapper with an exact `Decimal(18,6)` microsecond kernel with Spark-compatible NULL and ANSI/TRY overflow behavior.
- Remove the obsolete incompatibility gate, ignored regressions, and unused `datafusion-spark` dependency.
- Port Spark 4.2.0 `make_interval` boundary and arity cases with source permalinks.
- Remove the obsolete duplicate codegen-dispatch benchmark case now that `make_interval` is fully native.
- **Why not fix upstream `datafusion-spark` instead:** the upstream `make_interval` hard-codes `Interval(MonthDayNano)` as its return type, which is itself the 1000x range bug; the lossless fix requires the Comet-specific metadata-tagged struct that only Comet's Arrow/FFI layers can map back to `CalendarIntervalType`, plus Comet's ANSI `fail_on_error` semantics and `Decimal(18,6)` seconds. The `Float64` precision half is separately fixable upstream.
- **Codegen-dispatch path removed:** `MakeInterval` moves from `CometCodegenDispatch` to fully-native serde, so no dispatch route remains for it; fallback is directly to Spark. The `_dispatch` fixtures are deleted accordingly.
- **Wire-format note:** the Arrow representation change is not a cross-version concern — the Comet jar is fixed per Spark application and shuffle blocks are application-scoped, so old and new representations never mix within a supported deployment.

- `cargo test --manifest-path native/Cargo.toml -p datafusion-comet-spark-expr preserves_spark_microsecond_range_and_overflow`
- `cargo check --manifest-path native/Cargo.toml -p datafusion-comet-spark-expr`
- `cargo check --manifest-path native/Cargo.toml -p datafusion-comet`
- `make core`
- Focused `CometArrowStreamSuite` CalendarInterval round-trip test
- Focused `CometCodegenSuite` CalendarInterval codegen test
- `CometSqlFileTestSuite make_interval`: 6/6 passed, 0 ignored
- `CometSqlFileTestSuite calendar_interval`: 1/1 passed
- Spotless, Scalastyle, and `git diff --check`

### Benchmark

`CometDatetimeExpressionBenchmark`, 1,048,576 rows, Apple M4, JDK 17. The codegen-dispatch result is from parent commit `268849c0c`; the full-native and Spark results are from this PR at `72997e9e8`. Both Comet revisions used optimized native builds and the same query and input.

| Execution path | Best time | Average time | Throughput | Per row | Relative to dispatch |
| --- | ---: | ---: | ---: | ---: | ---: |
| Comet codegen dispatch | 31 ms | 33 ms | 33.8 M rows/s | 29.6 ns | 1.00x |
| Comet full native | 28 ms | 30 ms | 37.3 M rows/s | 26.8 ns | 1.11x |
| Spark | 74 ms | 81 ms | 14.1 M rows/s | 70.7 ns | 0.42x |

Full native is about 10% faster than codegen dispatch and 2.6x faster than Spark by best time.


andygrove (commented):
Thanks for the significant cleanup here — moving `make_interval` fully native and fixing the microsecond-precision loss is a real quality improvement, and the benchmark numbers show it. A few things worth addressing before merge.

### ANSI overflow message diverges from Spark

Spark's `MakeInterval.nullSafeEval` catches the `ArithmeticException` from `IntervalUtils.makeInterval` and rethrows via `arithmeticOverflowError(e.getMessage)`, where `e.getMessage` is Java's `Math.addExact` / `multiplyExact` string — literally `"integer overflow"` or `"long overflow"`. Under ANSI, Spark reports:

`[ARITHMETIC_OVERFLOW] integer overflow. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error.`

Comet's kernel always calls `arithmetic_overflow_error(""interval"")` at `native/spark-expr/src/datetime_funcs/make_interval.rs:150`, producing `interval overflow` instead. I confirmed this with a targeted `expect_error(integer overflow)` on `make_interval(2147483647)` — Spark's message matches, Comet's doesn't. The fixtures in this PR use `expect_error(overflow. If necessary set)`, which matches both and hides the divergence.

Could the free `make_interval` function return the specific label (`""integer""` for the year/month and week/day paths, `""long""` for the microsecond path) so `arithmetic_overflow_error` gets the same string Spark produces? A tighter `expect_error` in `make_interval_ansi.sql` would then pin the parity going forward.

### `Utils.toArrowType` `CalendarIntervalType` branch is now unreachable

With the new `case CalendarIntervalType =>` branch in `toArrowField` at `spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala:212`, the corresponding arm of `toArrowType` at line 172 is unreachable. If a future caller ever hits it directly, they get an unnamed, untagged `ArrowType.Struct.INSTANCE` with no children, and `isCalendarIntervalStructField` would fail to identify it on the round-trip. Would it be safer to make that arm throw with a message pointing callers at `toArrowField`, since the tagged struct can't be produced from an `ArrowType` alone?

### The `_dispatch` SQL fixtures no longer exercise a distinct path

With `MakeInterval` unconditionally native, `make_interval_dispatch.sql` and `make_interval_dispatch_ansi.sql` run through the same kernel as `make_interval.sql` / `make_interval_ansi.sql`. The `dispatch` name is now misleading and the queries mostly duplicate coverage. Could these either be removed, or repurposed with `-- Config: spark.comet.exec.enabled=false` at the top so they cover the pure-Spark fallback path instead?

### Argument downcasts `.unwrap()` where sibling kernels return errors

At `native/spark-expr/src/datetime_funcs/make_interval.rs:109-118`, the six `Int32Array` and one `Decimal128Array` downcasts unconditionally `.unwrap()`. This is unreachable today because the planner inserts a `CastExpr` for any input whose type differs from the `Signature::exact`, but `SparkMakeDate` in the same directory uses `.ok_or_else(|| DataFusionError::Execution(...))?` for the same pattern. Matching that style would give a diagnostic instead of a panic if a future serde change ever bypasses the cast.

### Boundary test coverage: only positive extreme

The `Int.MaxValue` boundary row in `make_interval.sql` covers the positive side of the checked arithmetic. It might be worth adding a mirror row that pushes at least one of year/month/week/day toward `Int.MinValue` with a negative `secs` at the boundary, so the negative side of `checked_add` / `checked_mul` is validated against Spark too. Spark's own `IntervalExpressionsSuite` only tests the positive case, but with the native kernel the two sides are separate branches of `i32::checked_mul` and it's cheap to cover both.

### End-to-end test for chained native consumption

`calendar_interval.sql` covers the shuffle case for `array<interval>`, but there's no test where a native `make_interval` result flows directly into another native step. If FFI (or any other stage on the interval column's path) ever drops the `SPARK::calendarInterval::struct` metadata on the `months` child, `isCalendarIntervalStructField` would silently degrade the type to an anonymous `StructType(months, days, microseconds)` and downstream `CalendarInterval` consumers would break with no clear error. A query like `SELECT date '2020-01-01' + make_interval(1, 2, 3, 4, 5, 6, 7.123456) FROM t` or a similar chained native consumer would pin the metadata-preservation invariant.

---

I ran `CometSqlFileTestSuite make_interval` locally on this branch — all 6 files pass. I also probed several extreme non-ANSI boundary cases (both signs of `i32::MAX` years/hours/minutes with `-999999999999.999999` and `+999999999999.999999` secs) and results match Spark. The change looks solid; the items above are all in-band for this PR.

peterxcli:
@andygrove Thanks for the detailed review. I addressed each point:

> Could `make_interval` return `"integer"` or `"long"` so `arithmetic_overflow_error` gets the same string Spark produces?

Changed the helper to return `"integer"` for month/day arithmetic overflow and `"long"` for microsecond arithmetic overflow. ANSI execution forwards that label to `arithmetic_overflow_error`, and native unit tests pin both values.

I kept the SQL assertion as `overflow. If necessary set` because [Spark 4.1.3 preserves `integer overflow`](https://github.com/apache/spark/blob/v4.1.3/sql/core/src/test/scala/org/apache/spark/sql/errors/QueryExecutionErrorsSuite.scala#L704-L713), while [SPARK-55714 canonicalizes it to `overflow` in Spark 4.2](https://github.com/apache/spark/blob/v4.2.0/sql/api/src/main/scala/org/apache/spark/sql/errors/ExecutionErrors.scala#L118-L144). This is documented beside the assertion; the native unit test provides the tighter check on Comet’s returned type.

> Would it be safer to make the `CalendarIntervalType` arm throw with a message pointing callers at `toArrowField`?

Changed `toArrowType` to throw an `UnsupportedOperationException` directing callers to `toArrowField`, and added unit-test coverage.

> Could the `_dispatch` SQL fixtures either be removed or repurposed?

Removed both dispatch fixtures. With `MakeInterval` fully native, they no longer exercised a distinct path.

> Matching `SparkMakeDate`’s downcast style would give a diagnostic instead of a panic.

Replaced all seven argument-downcast `.unwrap()` calls with descriptive `DataFusionError::Execution` errors.

> It might be worth adding a mirror row covering the negative boundaries.

Added a row reaching `i32::MIN` exactly for both computed months and days while also exercising negative hours, minutes, and seconds near the microsecond boundary.

> A chained native consumer would pin the metadata-preservation invariant.

Added an end-to-end query that constructs `ARRAY<INTERVAL>` using native `make_interval` below a native shuffle and consumes `intervals[0]` using native `GetArrayItem` above it. This pins the `CalendarInterval` metadata across both native-expression and shuffle boundaries.

peterxcli:

andygrove:

Fixing the 1000x range loss and the `Decimal(18,6)` to `Float64` precision loss in one go is a real improvement, and representing the interval the way Spark does, as separate months, days, and microseconds, is clearly the right model. Good to see `getSupportLevel` go from `Incompatible` to `Compatible` as a result. The metadata check in `isCalendarIntervalStructField` does require the marker key rather than just the field shape, which I was worried about, so a user's own `struct<months:int, days:int, microseconds:bigint>` will not be misread. Worth saying that explicitly in the doc comment, since it is the property the whole design rests on.

Four things.

**Should the fix go upstream to `datafusion-spark` instead?**

This replaces the `datafusion-spark` `make_interval` wrapper with a Comet-local kernel and drops the dependency. The bugs being fixed, `Float64` seconds and nanosecond storage, exist in the upstream implementation too, so every other `datafusion-spark` consumer still has them. Comet has been moving expressions the other direction, upstreaming native implementations into `datafusion-spark`. Is there a reason this one has to fork rather than fix upstream and pin? If the answer is "the tagged-struct representation is Comet-specific and cannot go upstream", saying so in the description would settle it.

**`toArrowType(CalendarIntervalType)` now throws**

Turning a working case into an `UnsupportedOperationException` means every caller has to have been converted to `toArrowField`. Did you audit them all, including callers outside `Utils.scala` and in test code? A runtime `UnsupportedOperationException` from a type conversion is not a friendly failure mode, and it will only show up on a path that happens to be exercised. If some callers legitimately cannot produce the metadata, the exception message should say what to call instead in more actionable terms than "requires toArrowField".

**Cross-version compatibility of the wire format**

The Arrow representation of `CalendarIntervalType` changes from `Interval(MonthDayNano)` to a struct. That affects native shuffle output and anything else that persists or transfers an Arrow schema. Within a single job everything is consistent. What about a rolling upgrade, where some executors run the old Comet and some the new one, or a shuffle service holding blocks written by the previous version? Is that a scenario Comet supports? If it is, this needs a note in the migration guide. If it is not, saying so in the description would close the question.

**Scope**

23 files touching serde, FFI, Arrow readers and writers, codegen input and output, Scala UDF codegen, and the native kernel. That is a lot of surface for one review pass, and the two `make_interval_dispatch*.sql` files being deleted means the dispatch path loses its dedicated coverage at the same time as the native path gains it. Is the dispatch path still reachable for `make_interval`? If it is, it still needs a fixture. If it is not, the description should say the dispatch path is gone rather than leaving the deletion unexplained.

peterxcli:
@andygrove Thanks for the second pass — replies inline, plus one doc improvement pushed.

**Upstream vs. fork:** The tagged-struct representation is Comet-specific, so this can't be an upstream fix. In `datafusion-spark`, `SparkMakeInterval::return_type` is hard-coded to `Interval(MonthDayNano)` — the 1000x range loss *is* the return type, not an implementation detail behind it. Fixing it means returning a struct whose "this is a CalendarInterval" meaning lives in field metadata that only Comet's Arrow conversion and FFI layers know to map back to Spark's `CalendarIntervalType`; DataFusion has no such logical type, and other upstream consumers expect the Arrow-native interval. The kernel also needs Comet's per-expression ANSI `fail_on_error` and exact `Decimal(18,6)` seconds. The `Float64`-seconds precision loss *is* separately fixable upstream, and I'm happy to file a `datafusion-spark` issue for that half. I've added this rationale to the PR description.

**`toArrowType` callers:** Audited repo-wide. The only production caller is the fallthrough inside `toArrowField` itself (`Utils.scala:235`), which can't reach the `CalendarIntervalType` arm because `toArrowField` matches that type first (line 214). The only other reference is the `UtilsSuite` test that pins the exception. So no caller can hit the throw today; it exists purely as a guard for future direct callers, per your earlier review's suggestion, and its message names `toArrowField` as the fix.

**Cross-version wire format:** Not a supported scenario, because it can't arise: the Comet jar is fixed per Spark application, so all executors in one app run the same Comet build, and shuffle blocks (external shuffle service included) are application-scoped — nothing written by one app's Comet is ever read by a different version. Comet persists no Arrow-format state beyond shuffle lifetime; durable storage is Parquet. I've added a note to the description rather than the migration guide.

**Dispatch path:** It's intentionally gone for `make_interval` — `CometMakeInterval` moved from `CometCodegenDispatch` to a fully-native `CometExpressionSerde`, so fallback is now straight to Spark with no codegen-dispatch route. The fixture deletion was the "removed" option from your first review. The pure-Spark path is still exercised by every remaining query, since the SQL-file harness computes each expected answer with Comet disabled. I've stated the dispatch-path removal explicitly in the description.

**Doc comment:** Good call on `isCalendarIntervalStructField` — added a Scaladoc stating that shape alone is insufficient and the `SPARK::calendarInterval::struct` marker is what prevents a user's own `struct<months:int, days:int, microseconds:bigint>` from being misread.

peterxcli:
@sunchao would you like to take a look at this and see if this can be merged? thanks!

PR #4487, opened 2026-05-28
feat: Native Parquet Iceberg Data File Writes In Comet

jordepic:
## Which issue does this PR close?

Closes #4322.

## Rationale for this change

Comet, up until this point, has mainly been focused with accelerating reads from iceberg tables.  However, a significant resources are being spent across various companies in order to rewrite iceberg data.  Large tables need to be compacted to maintain their sort/Z order, and general data pipelines may write significant amounts of iceberg data.  Having to do a large transpose between columnar and row-wise data is inefficient, and we'd much prefer to go directly from arrow-based column batches to parquet on disk.

## What changes are included in this PR?

This change is split into three parts.

1) Splitting the existing iceberg V2 spark write command into two: a "writer" spark operator and a "committer" spark operator.
- This allows the iceberg data file write to be treated identically to a spark V1 parquet writing command, thereby allowing using similar code to handle the operator
- Without it, both the data file writing and committing would live outside of the scope of spark adaptive query execution operator wrappers - we instead want the write operator to be within AQE so that as the data feeds into it gets re-planned we can determine whether the upstream data of the write is columnar

2) Determining which operators should be converted to native code
- This follows a simple philosophy: only convert writes to native code if they produce an "identical" outcome as the Java path
- Disclaimer: nothing truly produces an identical result because parquet row group flushing is different
- Besides that though, we can generally convert "normal" writes (parquet, default settings, some flexibility to change other settings as outlined in `iceberg-writes.md`, no delete files since iceberg-rust doesn't support positional deletes/DVs)

3) Native iceberg write operator
- The JVM is responsible for computing all iceberg writing settings and using a protobuf object to pass them to rust
- Rust uses iceberg-rs in order to write the file and return avro-encoded dummy manifest bytes back to the JVM

## How are these changes tested?

This change is tested extensively.
1) Tests to ensure that iceberg writes are replaced by our "two-operator" structure
2) Tests to ensure that the comet JVM properly serializes relevant data to protocol buffer form to go to the native layer
3) Tests to ensure that native writes are only performed under very specific iceberg table properties
4) Tests to ensure that native writes actually function as expected
5) Tests to ensure that compaction/sorting/z-ordering can now be fully accelerated with native writing

PR #5298, opened 2026-08-07
feat: detect Iceberg V2 writes and emit fall-back reasons

### LINKED ISSUE #4322: Writes to Apache Iceberg Tables

jordepic:

(Partially, second of three) Closes #4322. Part one was #4658 (merged); part three — the native writer itself — is tracked in #5308.

## Rationale for this change

The first part of this work (#4658) split the Iceberg V2 write into a "writer" operator (inside AQE) and a "committer" operator, with the writer still running through iceberg-java. 

This PR's goal is to detect WHICH iceberg writes we should attempt to complete, falling back otherwise. Iceberg-rust is less mature than iceberg-java when it comes to data file writing, so we establish whitelist criteria for which we are willing to write data files (I initially had blacklist critieria, but it turns out there's a lot more of those). Reasoning is listed in the markdown file attached to this commit.

1) A CometNativeIcebergWrite operator which will never be selected (reserved for part 3).
2) A detailed gate which checks whether we should accelerate a write. I believe that a vast majority of iceberg tables are just using parquet files on S3 with default settings (pareto principle), so we aim to hit those common use cases, while safely falling back for the less common ones.

There are certain metadata properties (mainly to do with column stats) that iceberg-rust doesn't currently support. We're still able to provide complete parity here by virtue of comitting metadata in the JVM (will happen in part 3, slight divergence from #4487 ). Other aspects of data file writes simply can't be equal between java and rust without a lot of forking on the rust side (one example would be checking the flushed row group sizes at a granularity of x bytes). All of these conditions are listed in the associated docs. 

A new `CometIcebergWriteDetectionSuite` (32 tests) pins every rule: one negative test per fall-back trigger asserting the reason string, positive tests for the boundary cases (defaults, explicit metrics modes, session-conf codec overrides, s3 data locations, partitioned tables), a smoke test that every reflection accessor resolves against the Iceberg runtime on the classpath, and a test that the planner registration records a fall-back reason on the write exec. The suite passes on the spark-3.4 (Iceberg 1.5.2, Scala 2.12) and spark-3.5 (Iceberg 1.8.1) profiles.

### DIFF
--- a/docs/source/user-guide/latest/iceberg-writes.md
+++ b/docs/source/user-guide/latest/iceberg-writes.md
@@ -41,7 +41,8 @@ writes into two operators:
 Data files are still written by iceberg-java; only the plan shape changes. The split makes the
 write's input visible to AQE and to Comet's columnar rules, and it is the groundwork for a
 planned follow-up in which Comet writes the data files natively via
-[iceberg-rust](https://github.com/apache/iceberg-rust).
+[iceberg-rust](https://github.com/apache/iceberg-rust), tracked in
+[#5308](https://github.com/apache/datafusion-comet/issues/5308).
 
 ## Configuration
 
@@ -57,6 +58,9 @@ spark.sql.catalog.<name>.warehouse=...
 
 # Split-operator plan (experimental, off by default)
 spark.comet.write.iceberg.splitOperator.enabled=true
+
+# Native-write eligibility detection (experimental, off by default; requires the split plan)
+spark.comet.iceberg.write.enabled=true
 ```
 
 ## Supported operations
@@ -95,3 +99,102 @@ The rewrite is skipped — and the write runs through Spark's stock combined ope
 
 In every fallback case the write is planned as if Comet were absent; there is no correctness
 trade-off, only no plan change.
+
+## Native Parquet write eligibility
+
+A planned follow-up ([#5308](https://github.com/apache/datafusion-comet/issues/5308)) replaces
+the `IcebergWrite` operator's per-task Parquet write with
+[iceberg-rust](https://github.com/apache/iceberg-rust). The native writer must produce the same
+outcome as iceberg-java — the same Parquet features, statistics, and manifest metadata — so a
+write is only eligible when every table property it depends on is one the native path reproduces
+exactly. `spark.comet.iceberg.write.enabled` enables this eligibility check;
+with the current release the native writer itself is not yet wired in, so every write still runs
+through iceberg-java and the check's outcome is reported as a fall-back reason in Comet's
+extended EXPLAIN output.
+
+**Most Iceberg write settings are not supported.** Detection is an allowlist: a write is
+eligible only when its entire effective configuration matches the table below, and anything
+else — any other write-affecting property, any key added by a future Iceberg version, any
+value outside the supported set, any reflection failure while inspecting the write — falls
+back to iceberg-java with a reason reported in extended EXPLAIN. Checks run on the effective
+configuration: table properties overlaid with `SparkWrite.writeProperties`, which is where
+iceberg-java resolves per-write options and `spark.sql.iceberg.*` session overrides.
+
+A write is eligible only when ALL of the following hold:
+
+| Setting                                                                                                                                     | Supported values                                                                  |
+| ------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------- |
+| resolved write format (`write-format` option overlaid on `write.format.default`)                                                            | `parquet`                                                                         |
+| `format-version`                                                                                                                            | `1` or `2`                                                                        |
+| `write.parquet.compression-codec` / `compression-level` / `row-group-size-bytes` / `page-size-bytes` / `page-row-limit` / `dict-size-bytes` | any value (translated to the native writer)                                       |
+| `write.parquet.row-group-check-min-record-count`                                                                                            | unset or `100` (the default)                                                      |
+| `write.parquet.row-group-check-max-record-count`                                                                                            | unset or `10000` (the default)                                                    |
+| `write.parquet.page-version`                                                                                                                | unset or `v1`                                                                     |
+| `write.parquet.shred-variants`                                                                                                              | unset or `false` (Spark 4.x / Iceberg 1.11 resolve this into every parquet write) |
+| `write.parquet.variant-inference-buffer-size`                                                                                               | any value (only meaningful when shredding, which is gated)                        |
+| `write.parquet.bloom-filter-enabled.column.<col>`                                                                                           | unset or `false`                                                                  |
+| `write.metadata.metrics.default`                                                                                                            | unset, `truncate(N)`, or `full`                                                   |
+| `write.metadata.metrics.column.<col>`                                                                                                       | unset, `truncate(N)`, or `full`                                                   |
+| `write.spark.fanout.enabled`                                                                                                                | any value (the native writer implements both clustered and fanout modes)          |
+| `write.target-file-size-bytes`                                                                                                              | any value (file rolling cadence differs; see accepted divergences)                |
+| data location URI scheme                                                                                                                    | `file`, `memory`, `s3`, `s3a`, `gs`, `oss`                                        |
+| partition spec                                                                                                                              | any (but see partition paths under accepted divergences)                          |
+
+Within the namespaces that shape data-file bytes — `write.parquet.*` and `parquet.*` —
+everything not listed above must be absent: unvetted `write.parquet.*` keys (e.g.
+`bloom-filter-max-bytes`, `stats-enabled.column.*`, keys added by future Iceberg versions),
+metrics modes outside the supported set (`counts`, `none`, or unparseable values), any
+`parquet.*` table property (including `parquet.enable.dictionary`), and any `parquet.*` key in
+the session Hadoop configuration (with `HadoopFileIO`-backed output those reach iceberg-java's
+writer but not the native one). Also gated explicitly: any `encryption.*` key,
+`write.object-storage.enabled=true`, `write.location-provider.impl`, and `io-impl`.
+
+Other `write.*` properties are intentionally not gated because they cannot make the native
+writer produce different data files: distribution and ordering settings shape the Spark plan
+identically on both paths, WAP / branch / snapshot properties act on the JVM committer,
+`write.avro.*` / `write.orc.*` apply only to formats already excluded, and merge-on-read
+settings route the write through `WriteDelta`, which the split plan never intercepts. Every
+rule is pinned by `CometIcebergWriteDetectionSuite`.
+
+Manifest `DataFile` metrics will be assembled on the JVM at commit time using Iceberg's own
+`MetricsConfig` logic, so iceberg-java's metadata decisions — metrics modes, the
+inferred-column cap (`write.metadata.metrics.max-inferred-column-defaults`), bound truncation,
+and list/map bounds suppression — are respected exactly regardless of what the native writer
+reports. The `counts`/`none` restrictions above remain only until that assembly lands.
+
+## Accepted divergences behind the toggle
+
+Some differences between parquet-mr and the pinned parquet-rs / iceberg-rust are unconditional —
+they apply to every native write and cannot be configured away. Enabling
+`spark.comet.iceberg.write.enabled` accepts them:
+
+- Footer key-value metadata differs: native files carry an `ARROW:schema` entry and no
+  `iceberg.schema` entry; iceberg-java files are the opposite.
+- The Parquet root schema element is named `arrow_schema` (iceberg-java: `table`).
+- `created_by` identifies parquet-rs, not parquet-mr.
+- No page CRC checksums and no page-header statistics (parquet-mr writes both by default).
+- Dictionary-encoded pages are labeled `RLE_DICTIONARY` (parquet-mr v1 files: `PLAIN_DICTIONARY`).
+- Fixed-length binary columns (`uuid`, `fixed`, decimals with precision > 18) are not
+  dictionary-encoded (parquet-mr dictionary-encodes them).
+- Row-group boundaries: parquet-mr flushes by byte size at a record-count check cadence,
+  parquet-rs buffers by row count. File rolling and file naming follow the same cadence-style
+  differences (iceberg-java checks the target file size every 1000 rows and names files
+  `<partition>-<task>-<operation>-<count>`; iceberg-rust checks per batch and uses a
+  process-local counter).
+- Partition paths are not URL-escaped: iceberg-java percent-encodes partition directory names
+  and values (`region=a%2Fb`), iceberg-rust writes them raw (`region=a/b`). Readers resolve
+  files through manifest metadata, not paths, so query results are unaffected — but the
+  directory layout differs from iceberg-java's, and partition values containing characters
+  that are invalid in a URI (`:`, `#`, newline) may produce paths that `HadoopFileIO`-based
+  readers cannot open.
+- Compressed page bytes are implementation-defined: the codec and any explicit level are
+  translated, but parquet-rs and parquet-mr embed different encoder implementations and
+  defaults (zstd default levels, LZ4 framing), so byte-identical output is not achievable even
+  for a default `zstd` table. The decompressed data is identical. For the same reason,
+  codec-level side channels (`zlib.compress.level`, `compression.brotli.quality`,
+  `io.compression.codec.zstd.level` — the last is present in every Hadoop configuration by
+  default) are not gated: they can only shift compressed bytes, which are already accepted as
+  divergent.
+
+All content not listed above — the logical data, encodings for non-FLBA columns, statistics
+values, and manifest metadata — must match iceberg-java exactly, or the write falls back.
--- a/spark/src/main/scala/org/apache/comet/CometConf.scala
+++ b/spark/src/main/scala/org/apache/comet/CometConf.scala
@@ -131,6 +131,16 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(false)
 
+  val COMET_ICEBERG_NATIVE_WRITE_ENABLED: ConfigEntry[Boolean] =
+    conf("spark.comet.iceberg.write.enabled")
+      .category(CATEGORY_TESTING)
+      .doc(
+        "Whether to delegate the executor-side Parquet write to Comet's native (iceberg-rust) " +
+          "writer when the table's properties allow it. Requires " +
+          "`spark.comet.write.iceberg.splitOperator.enabled = true`. Off by default.")
+      .booleanConf
+      .createWithDefault(false)
+
   val COMET_ICEBERG_DATA_FILE_CONCURRENCY_LIMIT: ConfigEntry[Int] =
     conf("spark.comet.scan.icebergNative.dataFileConcurrencyLimit")
       .category(CATEGORY_SCAN)
--- a/spark/src/test/scala/org/apache/comet/CometIcebergWriteDetectionSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometIcebergWriteDetectionSuite.scala
@@ -1,0 +1,656 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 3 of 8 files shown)


jordepic:
Hey @andygrove , @comphead , @mbutrovich , @parthchandra ! Here's commit 2/3 in accelerating iceberg writes. I've had to deal with a lot of ambiguity in this one when deciding what we do and don't want to accelerate, but I think it's a pretty comprehensive list at this point and covers the edge cases. Feel free to reply here, iceberg's slack, or the datafusion discord if you want to discuss more or have a call. Thanks!

jordepic:
@andygrove would you mind kicking CI off on this one? Sorry, I fixed a formatting issue

andygrove:
> Hey @andygrove , @comphead , @mbutrovich , @parthchandra ! Here's commit 2/3 in accelerating iceberg writes. I've had to deal with a lot of ambiguity in this one when deciding what we do and don't want to accelerate, but I think it's a pretty comprehensive list at this point and covers the edge cases. Feel free to reply here, iceberg's slack, or the datafusion discord if you want to discuss more or have a call. Thanks!

Thanks @jordepic. I appreciate the offer of the direct communications channels, but we prefer to keep discussions in issues/PRs so that the whole community can follow along. All issue/PR discussions get copied to a mailing list and archived as well. This is an important part of the Apache process.

jordepic:
@andygrove sounds good on the communications front, good point!

andygrove:
> @andygrove sounds good on the communications front, good point!

It also helps the LLMs follow along!

andygrove (commented):
Thanks for this @jordepic. The staging makes sense, and a couple of things I checked and liked: the `SparkWrite$` prefix test in `isIcebergBatchWrite` cleanly excludes merge-on-read deltas, since those arrive as `SparkPositionDeltaWrite$PositionDeltaBatchWrite` and friends. And the storage scheme set (`file`, `memory`, `s3`, `s3a`, `gs`, `oss`) matches what `iceberg_scan.rs:256-270` actually supports, including the same `contains("://")` parsing, so the two sides will not drift apart in confusing ways.

A handful of things I would like to work through before this goes in.

### 1. `PropertyKeys` can turn an intended fallback into a planning crash

`CometIcebergNativeWrite.scala:317` and `:323` are strict `private val`s whose initializers read `PropertyKeys.ParquetRowGroupCheckMinRecordCount` and the matching `...Default`. Because they are strict vals in an `object` body, this runs during `CometIcebergNativeWrite$` initialization, which is triggered by the new `case op: IcebergWriteExec` arm in `CometExecRule.scala:301`, outside `getSupportLevel`'s try. `CometExecRule.apply` has no safety net either. Could those be `lazy val`?

The broader version of the problem: `getSupportLevel` catches `Exception`, but any failure inside `PropertyKeys` surfaces as `ExceptionInInitializerError` on first access and `NoClassDefFoundError` after that, and both are `Error`. So the reflection-failure fallback this whole design rests on would not actually catch a missing `TableProperties` field. Would you mind widening the catch to `NonFatal`?

Reachability today is narrow, since an `IcebergWriteExec` in the plan implies `IcebergWriteStrategy` ran, which implies Iceberg is on the classpath. The realistic trigger is a future or forked Iceberg renaming one of these fields, which is exactly the case the docs promise falls back.

### 2. The doc describes an allowlist, the code is a denylist outside `write.parquet.*`

`iceberg-writes.md` says "Detection is an allowlist: a write is eligible only when its entire effective configuration matches the table below, and anything else, any other write-affecting property, any key added by a future Iceberg version ... falls back", then "Everything else must be absent."

That is stronger than what the rules enforce. `requireOnlyVettedParquetWriteProperties` only covers keys starting with `write.parquet.`, and `requireNoUnvettedParquetMrProperties` only covers `parquet.`. Everything outside those prefixes has to be named explicitly. I checked `TableProperties` in the Iceberg 1.11 runtime jar and these all pass the gate today:

- `write.spark.fanout.enabled`
- `write.upsert.enabled`
- `write.distribution-mode` (and the `write.delete`/`merge`/`update` variants)
- `write.target-file-size-bytes`
- `write.object-storage.path`
- `write.avro.compression-codec`
- `write.wap.enabled`

Most of those look harmless for data-file bytes, and `write.target-file-size-bytes` is arguably covered by the file-rolling divergence you already document. `write.spark.fanout.enabled=true` is the one I would want gated explicitly, because it changes the writer structure rather than a parameter. A fanout writer keeps writers open across many partitions on unsorted input, so a clustered-only native writer would produce wrong output rather than merely different bytes. Does part 3 plan to support fanout? If not, could that get its own rule?

Either way, could the "Everything else must be absent" line be reworded to describe what is actually checked?

### 3. The metrics-mode rules are also a denylist

The doc table says `write.metadata.metrics.default` and `write.metadata.metrics.column.<col>` support "unset, `truncate(N)`, or `full`". `requireMetricsModeIsNotCounts` and `requireMetricsModeIsNotNone` only reject the literal strings `counts` and `none`, so a typo like `truncat(16)` or any mode a future Iceberg adds would pass.

Would it be worth inverting these into a single allowlist rule that matches the doc? That would also let `requireNoPerColumnStatsEnabled` and `requirePropertyAbsent(ParquetBloomFilterMaxBytes)` go away, since `requireOnlyVettedParquetWriteProperties` already subsumes both. That rule table is going to keep growing, so trimming the redundancy now seems worthwhile.

### 4. `requireFormatVersionAtMostTwo` fails open

```scala
IcebergReflection.getFormatVersion(ctx.table) match {
  case Some(v) if v >= MinUnsupportedFormatVersion => Some(s"format-version=$v unsupported")
  case _ => None
}
```

`getFormatVersion` returns `None` on reflection failure after only a `logError`, and `case _ => None` reads that as eligible. So the one case where Comet does not know the format version gets treated as V1/V2. Should the `None` case return a reason instead?

### 5. `getOuterSparkWrite` fails open on a missing `this$0`

`IcebergReflection.scala:174-175` returns `Some(batchWrite)` when `this$0` is absent. Downstream, `getSparkWriteField(batchWrite, "table")` then calls `SparkWrite.class.getDeclaredField("table").get(batchWrite)` on an object that is not a `SparkWrite`. That raises `IllegalArgumentException`, gets caught, logged at ERROR, and reported as `"SparkWrite.table is null"`, which is not what happened. Would returning `None` be better here, so `checkTriggers` reports "could not unwrap SparkWrite"?

### 6. Compression divergence is missing from the accepted-divergences list

The doc table says `write.parquet.compression-codec` and `compression-level` accept "any value (translated to the native writer)", and the closing paragraph says compression "must match iceberg-java exactly, or the write falls back." I think those are in tension for the default case.

From the 1.11 `TableProperties`, `PARQUET_COMPRESSION_DEFAULT_SINCE_1_4_0` is `zstd` and `PARQUET_COMPRESSION_LEVEL_DEFAULT` is unset. `SparkWriteConf.dataWriteProperties()` only forwards the level when it is non-null, so in the default configuration neither writer gets an explicit level and parquet-mr and parquet-rs pick different zstd defaults. Same question for `lz4` and the `LZ4` versus `LZ4_RAW` codec id.

Nothing incorrect falls out of that, since the data reads back the same and `file_size_in_bytes` is measured from the real file. But it does mean byte-identical output is not achievable for the most common table setting, so it seems like it belongs in the divergences list rather than under "must match exactly".

### 7. The `CometExecRule` registration is dead code with a visible side effect

Since `convert` always returns `None` and `createExec` throws, the arm at `CometExecRule.scala:301` cannot change a plan. What it does do is run `isOperatorEnabled`, which with the config off (the default) tags every `IcebergWriteExec` with "Native support for operator IcebergWriteExec is disabled. Set `spark.comet.write.iceberg.nativeAcceleration.enabled=true` to enable it."

`IcebergWriteExec` is Comet's own operator, inserted by `IcebergWriteStrategy`. So anyone on the split-operator path now sees extended EXPLAIN report a fall-back on a node Comet itself added, when nothing is falling back.

Would it be cleaner to hold the registration and the throwing `createExec` for part 3, and keep this PR to the detection logic plus `CometIcebergWriteDetectionSuite`? The suite already calls `getSupportLevel` directly, so it does not need the rule wired up. If you would rather keep the wiring so the explain plumbing gets exercised now, that is reasonable, but the spurious reason deserves a note in the docs.

### 8. `captureWriteExec` swallows every `Throwable`

In `CometIcebergWriteDetectionSuite.scala`:

```scala
try trigger
catch { case _: Throwable => () }
```

Combined with capturing from `onFailure`, this is what makes the `s3://nonexistent-bucket` and `hdfs://nonexistent.invalid` tests work, which makes sense. The side effect is that the positive tests cannot distinguish a successful write from one that blew up for an unrelated reason, so `assertSupportLevelIs[Compatible]("ok")` would keep passing if `INSERT INTO` regressed on some profile.

Could the tests that write to the real local warehouse assert the write actually succeeded, and only the deliberately-unreachable-location tests tolerate a failure?

### 9. Test coverage

A few cases worth pinning down, mostly around the doc-versus-code question above:

- `write.spark.fanout.enabled=true` and `write.target-file-size-bytes`, whichever way you decide they should go. Right now their behavior is a side effect of which prefixes the allowlist covers rather than a stated decision.
- The non-Iceberg `batchWrite` branch and the "could not unwrap SparkWrite" branch, which are the paths item 5 touches.
- Positive scheme tests for `gs`, `oss`, `memory`, and an explicit `file://` data path. Only `s3` is covered, and plain `file` is reached by the fall-through rather than by matching the set.
- The reason reported when `nativeAcceleration.enabled=false`.

Also, the description says the suite was verified on spark-3.4 (Iceberg 1.5.2) and spark-3.5 (Iceberg 1.8.1), but the workflow change puts it in the `scans` bucket, which runs on all five profiles up through Spark 4.2. Iceberg 1.9 through 1.11 add `write.parquet.shred-variants` and the `write.parquet.bloom-filter-fpp.column.` / `bloom-filter-ndv.column.` prefixes. The allowlist should handle all three correctly, but worth confirming the newer profiles are green rather than assuming.

docs/source/user-guide/latest/iceberg-writes.md
```diff
@@ -95,3 +98,84 @@ The rewrite is skipped — and the write runs through Spark's stock combined ope
 
 In every fallback case the write is planned as if Comet were absent; there is no correctness
 trade-off, only no plan change.
+
+## Native Parquet write eligibility
+
+A planned follow-up replaces the `IcebergWrite` operator's per-task Parquet write with
```
andygrove: Talking about a planned follow-up here is confusing to me. Is this PR the planned follow-up, or should this link to an issue?

jordepic: 3/3 is the planned follow up - how would you prefer I handle it?

andygrove: Perhaps you could link to an issue describing the work in 3/3?

jordepic: See #5308

jordepic:
@andygrove would you mind another CI kick? Also, if there's a way to grant me permissions for these, I promise not to mine crypto on the CI runners :)

docs/source/user-guide/latest/iceberg-writes.md
andygrove: Sorry, I meant could you include the link in the docs.

jordepic: My apologies, I knew that and forgot, all done, may need to kick CI again

jordepic:
@andygrove Every point is addressed in the current push. Would you mind kicking CI again? Point by point:

1. **PropertyKeys init / catch width**: the two strict rule vals are now `lazy`, so no `PropertyKeys` access happens during object initialization — every reflective constant lookup now occurs inside `getSupportLevel`'s try, and Scala lazy vals retry on failure so nothing gets poisoned. The catch is widened to `NonFatal`.
2. **Allowlist scope / fanout**: part 3 implements both clustered and fanout writer modes (mirroring `SparkWriteConf.useFanoutWriter`), so `write.spark.fanout.enabled` is declared supported rather than gated, and `write.target-file-size-bytes` rides the documented file-rolling divergence — both now have explicit rows in the doc table and tests. The "everything else must be absent" wording is reworded to state exactly which namespaces the allowlist enforces (`write.parquet.*`, `parquet.*`) and why other `write.*` keys are out of scope (plan-shaping, committer-side, non-parquet formats, or merge-on-read which the split plan never intercepts).
3. **Metrics modes**: inverted into a single allowlist rule — `full` and `truncate(N)` parse, anything else (including `truncat(16)` typos and future modes) falls back, with a test. The redundant `bloom-filter-max-bytes` and `stats-enabled.column.*` rules are gone; the property allowlist subsumes them.
4. **format-version**: the `None` case now falls back with "could not determine the table format-version".
5. **getOuterSparkWrite**: a missing `this$0` now returns `None`, so the reported reason is "could not unwrap SparkWrite".
6. **Compression**: moved to the accepted-divergences list (zstd default levels, LZ4 framing — byte-identical output isn't achievable even for a default table), and the "must match exactly" closing no longer claims compression.
7. **Registration**: the `CometExecRule` arm is now guarded by the config flag, so the default configuration tags nothing at all — no spurious fall-back reason on Comet's own operator. A test pins that no reason is recorded when the flag is off. I kept the wiring (rather than deferring to part 3) so the explain plumbing is exercised when users opt in.
8. **Test writes**: `captureWriteExec` is strict by default and fails the test if the INSERT throws; only the fixtures that deliberately write to unreachable locations or set broken classes (hdfs/s3/gs/oss/memory schemes, encryption, custom location provider / FileIO) tolerate failure.
9. **Coverage**: added tests for fanout, target-file-size, gs/oss/memory and explicit `file://` schemes, the disabled-flag behavior, a stubbed non-Iceberg `BatchWrite`, the metrics typo, and `write.parquet.shred-variants`. The "could not unwrap SparkWrite" branch isn't directly constructible (it needs a real `SparkWrite$` inner class without `this$0`), so it's covered only by the code path being fail-closed.

Your version-drift concern was on the money: running the suite on all four profiles locally caught two real issues the 3.4/3.5-only verification missed, and both are what CI was failing on. Iceberg 1.11 on Spark 4.x unconditionally resolves `write.parquet.shred-variants=false` into every parquet write's `writeProperties` (now vetted, with `true` gated since shredding changes the parquet schema), and Hadoop 3.4's `core-default.xml` ships `io.compression.codec.zstd.level`, which made gating codec-level keys via the Hadoop conf a false positive on every Spark 4 write — those gates are removed and folded into the accepted compressed-bytes divergence (they can only shift compressed bytes). The suite (40 tests) is green locally on spark-3.4/1.5.2, 3.5/1.8.1, 4.0/1.10, and 4.1/1.11.

andygrove (commented):
Thanks for the thorough round, @jordepic. I went through the new revision and can confirm each item lands.

1. Two rule vals are `lazy`, the catch is `NonFatal`, and `Try(raw.trim.toInt)` in `requireIntPropertyAtDefault` sits inside that same catch so overflow parses fall through cleanly.
2. Fanout has its own row, `target-file-size-bytes` rides the file-rolling divergence, and the wording now describes the actual scope (`write.parquet.*` and `parquet.*`) with a nice explanation of why other `write.*` families are out of scope. The merge-on-read note lines up with the `SparkWrite$` prefix filter in `isIcebergBatchWrite`, so the two claims stay consistent.
3. `requireSupportedMetricsModes` collapses the four rules to one and matches the doc. The `truncate(N)` regex catches typos, and the previously-redundant `bloom-filter-max-bytes` and `stats-enabled.column.*` rules are gone. `write.parquet.stats-enabled.column.<col>` is still caught by `requireOnlyVettedParquetWriteProperties`, so the existing test at line 315 still fires.
4. `requireFormatVersionAtMostTwo` no longer fails open.
5. `getOuterSparkWrite` returns `None` on missing `this$0`.
6. Compression divergence and the codec-level side channels are both in the accepted-divergences list, and the "must match exactly" closing was trimmed to not claim compression. Removing `requireNoCodecLevelProperties` and the codec-level check inside `requireNoParquetHadoopConfOverrides` also explains the CI signal you mentioned.
7. The `case op: IcebergWriteExec` arm is now guarded by `COMET_ICEBERG_NATIVE_WRITE_ENABLED`, so the default configuration never tags the operator. The new `no fall-back reason is recorded when nativeAcceleration is disabled` test pins this.
8. `captureWriteExec` is strict by default, `allowWriteFailure` is opt-in per test, and the fixtures that use it are exactly the ones that need to (unreachable schemes, custom `FileIO`, encryption stub).
9. Added coverage for `gs`/`oss`/`memory`/explicit `file://`, `write.spark.fanout.enabled`, `write.target-file-size-bytes`, `write.parquet.shred-variants`, the metrics typo, the disabled-flag case, and the non-Iceberg `BatchWrite` branch via a `SparkPositionDeltaWrite`-style stub.

Also good catch on `write.parquet.shred-variants` and `io.compression.codec.zstd.level` — that is the kind of thing that only surfaces on real profiles.

LGTM once CI comes back green. The four Iceberg-profile jobs are the ones that would still surprise us, so I'd wait on those before merging.

jordepic:
@andygrove they passed! thank you!

andygrove (approved):
All 61 checks green, 9 skips, nothing failing. That includes the four Iceberg profile jobs (1.5.2 on Spark 3.4, 1.8.1 on Spark 3.5, 1.10 on Spark 4.0, 1.11 on Spark 4.1) that were the remaining risk after round 2. `CometIcebergTestBase.icebergAvailable` gates Spark 4.2 out because there's no matching Iceberg runtime published yet, so that profile skips the suite as expected.

Head SHA is unchanged from round 2 — no new commits to look at. I did a fresh pass over the reflection additions in `IcebergReflection.scala` (`isIcebergBatchWrite`, `getOuterSparkWrite`, `getSparkWriteField`, `getDataLocation`, the `TableProperties` constant readers) with new eyes and don't find anything to raise. The `getDataLocation` scheme parsing is robust to both `file:/warehouse/...` (Hadoop single-slash form) and `file:///warehouse/...` because the else branch defaults to `file`, and every reflection helper is either wrapped in a caller-side try or returns `Option` that surfaces as a fallback reason.

Approving. Nice work on this one.

jordepic:
@andygrove Thank you for your quick reviews here! I cannot merge myself, but when this gets merged I'll get to work putting up patchset 3! Getting close :)

andygrove:
@mbutrovich @comphead @parthchandra any objection to merging this one?

parthchandra:
I think we can merge this.

spark/src/test/scala/org/apache/comet/CometIcebergWriteDetectionSuite.scala
```diff
@@ -0,0 +1,685 @@
...
+  test("clean parquet V2 table planned as AppendData yields Compatible") {
+    assume(icebergAvailable, "Iceberg not available in classpath")
+    withDetectionCatalog { dir =>
+      createTable(dir, "ok", partitionSpec = "")
+      assertSupportLevelIs[Compatible]("ok")
+    }
+  }
+
+  test("registration tags a fall-back reason on the write exec") {
+    assume(icebergAvailable, "Iceberg not available in classpath")
```
comphead: maybe we can override `test` and include `assume(icebergAvailable, "Iceberg not available in classpath")` there to avoid calling it every test?

spark/src/main/scala/org/apache/comet/CometConf.scala
```diff
@@ -131,6 +131,16 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(false)
 
+  val COMET_ICEBERG_NATIVE_WRITE_ENABLED: ConfigEntry[Boolean] =
+    conf("spark.comet.write.iceberg.nativeAcceleration.enabled")
```
comphead: reg to param name would be nice to be in sync with existing writer param name

```
    conf("spark.comet.parquet.write.enabled")
```

spark/src/test/scala/org/apache/comet/CometIcebergWriteDetectionSuite.scala
jordepic: Good call — done. The suite now overrides `test` with the `assume` (same pattern as `CometFuzzIcebergBase`), and the per-test calls are gone.

spark/src/main/scala/org/apache/comet/CometConf.scala
jordepic: Done — renamed to `spark.comet.iceberg.write.enabled` to mirror `spark.comet.parquet.write.enabled`.

jordepic:
@andygrove would you mind kicking CI once more? I addressed @comphead 's nits. Would love to merge after that though! Seems you, Parth and Oleks are generally okay with it!

jordepic:
@andygrove I believe the failing test is transient

Issue #4521, opened 2026-05-29
native shuffle: get_string should not panic on non-UTF-8 bytes (use lossy decode)


schenksj:
### Background (standalone)

Spark's `UnsafeRow.getUTF8String` wraps bytes with no UTF-8 validation, and `cast(BinaryType -> StringType)` is a zero-copy reinterpret — so a `StringType` column can legitimately hold arbitrary, non-UTF-8 bytes. Comet's native shuffle `get_string` decoded with `from_utf8(...).unwrap()`, which **panics** on such rows even though Spark itself treats them as opaque.

### Proposal

Decode with `from_utf8_lossy`: it returns the original borrow for valid UTF-8 (zero-cost) and a `String` with `U+FFFD` replacements for invalid bytes — defined behavior, no UB. This avoids `from_utf8_unchecked`, which would construct a `&str` from arbitrary bytes (UB per the Rust reference, and would propagate into downstream Arrow ops that internally call `str::from_utf8_unchecked` on the buffer). Changes `get_string` to return `Cow<str>`; callers use it via `AsRef<str>`.

### Relationship to the Delta integration

Standalone — any `binary -> string` cast routed through shuffle hits this, no Delta knowledge needed to review. **Required by the in-progress `contrib/delta` integration**: Delta Z-Order `OPTIMIZE` emits `interleave_bits(...).cast(StringType)`, which produces non-UTF-8 bytes and panics shuffle today. Please prioritize accordingly.

### Tests

- Shuffle round-trip of a `StringType` column containing invalid UTF-8 bytes (e.g. from a `binary -> string` cast) returns without panicking and matches Spark's lenient semantics.

PR #5310, opened 2026-08-08
fix: decode invalid UTF-8 at the JVM to native FFI import boundary

manuzhang:
This PR continues the work from #4945.

Part of #4764 (EPIC: consistent handling of invalid UTF-8 in native StringType). This PR implements **Gap B** only, decoding invalid UTF-8 at the JVM to native Arrow FFI import boundary. It does not close the EPIC, which still tracks Gap A (the native scan rejecting invalid UTF-8).

## Rationale for this change

Spark's `StringType` (`UTF8String`) can hold arbitrary bytes, including sequences that are not valid UTF-8. When a JVM side source hands string columns to native code over the Arrow C Data Interface, arrow-rs imports them with `from_ffi` / `from_ffi_and_data_type`, which build the array via `ArrayData::new_unchecked` and do not validate UTF-8. The imported Arrow `Utf8` / `LargeUtf8` array then lies about its validity, and any downstream native string kernel that reads `&str` through arrow-rs's unchecked `StringArray::value()` (`from_utf8_unchecked`) exercises undefined behaviour: iterating chars, slicing on char boundaries, and similar operations can misbehave, panic, or be miscompiled. This is a latent, default configuration soundness hazard.

The string producing sites already decode invalid bytes the way Spark renders them: `CAST(binary AS string)` (#4763) and native shuffle `get_string` (#4521) both use `decode_utf8_spark_lossy`. This PR applies the same policy to the string ingress side, so the whole native pipeline agrees on a single invariant: native string data is always valid UTF-8.

- A new `decode_string_arrays` walker in `datafusion-comet-common`, beside the existing `decode_utf8_spark_lossy`. It ensures every `Utf8` / `LargeUtf8` array reachable from an imported column holds valid UTF-8, decoding invalid bytes to Spark's rendered form. It is zero copy for the valid common case (one validation pass plus an O(number of strings) boundary check, returning the same `Arc`), and rebuilds element by element only when bytes are genuinely invalid. It recurses through `Dictionary`, `Struct`, `List`, `LargeList`, `FixedSizeList`, and `Map`.
- There are four `from_ffi` call sites. Three carry string data and decode directly:
  - `ScanExec::pull_next`, which handles all native query input (the JVM reader, Spark columnar handoff, shuffle reads, mapInArrow). Decoding runs before dictionary unpack so compact dictionary values are validated rather than the expanded ones.
  - `columnarToRow` conversion.
  - JVM UDF result.

  The fourth, `batch_from_ffi` in `aligned_stream_reader.rs`, is consumed only by `ScanExec`, which passes every column through `import_column` immediately after the batch is materialised. Decoding there as well would validate the same buffers twice, so its doc comment records that any future consumer must keep that step.
- Anything string-bearing that the walker does not handle fails closed rather than passing unchecked data through. `data_type_contains_string` is an exhaustive match, so an Arrow upgrade that adds a data type forces this decision to be revisited, and types such as `Utf8View`, the view lists and `RunEndEncoded` return `NotYetImplemented` if they ever reach this boundary.
- The fast path validates the used byte range and also confirms no element boundary splits a codepoint. This second check is required for soundness: a whole buffer valid `"é"` (bytes `C3 A9`) split across two offsets would otherwise hand each element an invalid slice, and `value()` decodes those unchecked.
- Validation uses `simdutf8` rather than `std::str::from_utf8` (see Performance below). `simdutf8` is already in the dependency graph through `arrow-json`, so this adds no new third-party crate.
- Benchmarks: a criterion benchmark for the valid fast path, and `CometStringFfiImportBenchmark`, which measures both import sites end to end.

No configuration flag gates this. It fixes undefined behaviour in the default configuration and matches the always on behaviour of the sibling cast and shuffle fixes.

The only observable divergence from Spark is the previously documented one: decoding rather than preserving raw bytes differs only under byte level round trips (for example `CAST(CAST(X'FF' AS STRING) AS BINARY)`), already noted in the compatibility guide.

## Performance

Validation costs one pass over each imported string buffer, so the cost scales with the bytes crossing the boundary, not with the number of rows.

`CometStringFfiImportBenchmark` (added here) measures both import sites over 1M rows of roughly 180 byte strings, in ASCII and in multibyte text. Each imported case is paired with a control that does the same work without crossing an import boundary, and every case checks its executed plan before it is timed. Best time on an M2:

| case | before | after |
| --- | --- | --- |
| ASCII, JVM Parquet scan into native (imported) | 108 ms | 115 ms (+6%) |
| ASCII, Comet native Parquet scan (control) | 63 ms | 64 ms |
| ASCII, native columnar to row (imported) | 56 ms | 62 ms (+11%) |
| ASCII, JVM columnar to row (control) | 42 ms | 42 ms |
| multibyte, JVM Parquet scan into native (imported) | 109 ms | 234 ms (+115%) |
| multibyte, Comet native Parquet scan (control) | 76 ms | 77 ms |
| multibyte, native columnar to row (imported) | 59 ms | 194 ms (+229%) |
| multibyte, JVM columnar to row (control) | 58 ms | 66 ms |

The controls stay flat, which is what places the cost in validation rather than anywhere else.

Those numbers come from validating with `std::str::from_utf8`. It has a fast ASCII path but degrades to a byte at a time loop on non-ASCII text. Timing the validator alone on the same string shapes, 8192 rows per batch:

| strings | `std::str::from_utf8` | `simdutf8::basic` |
| --- | --- | --- |
| ASCII, 156 bytes/row | 4.8 ns/row (32 GB/s) | 1.7 ns/row (92 GB/s) |
| multibyte, 184 bytes/row | 111 ns/row (1.65 GB/s) | 15.7 ns/row (11.7 GB/s) |

111 ns/row accounts for the whole multibyte regression above, and the boundary check costs under 1 ns/row. The last commit therefore switches the fast path to `simdutf8`, which is 7x faster on multibyte data and 3x on ASCII. Substituting 15.7 ns/row back into the cases above projects roughly +15% for the multibyte converted scan and +29% for multibyte native columnar to row, with ASCII near noise.

**The `simdutf8` numbers are measured at the validator, not end to end.** Repeated attempts to re-time the full benchmark against the `simdutf8` build on this machine were spoiled by background load (the same case varied up to 4x between runs), so the end to end table above is still the `std` build. A quiet machine or the EC2 micro benchmark runner (`benchmarks/micro/run.py`) should confirm the projection before this is taken as final.

Skipping validation where the producer is known to be native is not available as a shortcut: native columnar to row can receive strings built on the JVM, for example through a union, the in-memory cache, or broadcast.

- Rust unit tests for the walker covering: valid input returned zero copy (pointer identical buffers); invalid bytes decoding to U+FFFD matching the JVM; the split codepoint boundary case; nested `Dictionary` / `Struct` / `List` / `FixedSizeList` / `Map` (decode and zero copy paths); null preservation; sliced arrays with a non zero starting offset; and trailing empty strings.
- A Rust test at the `ScanExec` import site proving an invalid UTF-8 column is decoded through the production per column path.
- Both suites pass against the `simdutf8` fast path.
- A criterion benchmark for the valid fast path, and `CometStringFfiImportBenchmark` for the two import sites end to end.

### DIFF
--- a/native/common/src/utf8.rs
+++ b/native/common/src/utf8.rs
@@ -1,0 +1,444 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::decode_utf8_spark_lossy;
+use arrow::array::{
+    downcast_dictionary_array, Array, ArrayRef, FixedSizeListArray, GenericListArray,
+    GenericStringArray, GenericStringBuilder, MapArray, OffsetSizeTrait, StructArray,
+};
+use arrow::datatypes::DataType;
+use arrow::error::ArrowError;
+use std::sync::Arc;
+
+/// Ensure every `Utf8`/`LargeUtf8` array reachable from `array` holds valid UTF-8, decoding invalid
+/// bytes the way Spark renders `StringType`. Returns the same `Arc` (zero-copy) when nothing needed
+/// decoding. Used at the JVM->native FFI import boundary, where arrow's `from_ffi` builds string
+/// arrays via `new_unchecked` and does not validate UTF-8.
+pub fn decode_string_arrays(array: &ArrayRef) -> Result<ArrayRef, ArrowError> {
+    match array.data_type() {
+        DataType::Utf8 => decode_generic_string::<i32>(array),
+        DataType::LargeUtf8 => decode_generic_string::<i64>(array),
+        DataType::Dictionary(_, value_type)
+            if matches!(value_type.as_ref(), DataType::Utf8 | DataType::LargeUtf8) =>
+        {
+            // Capture the original Arc before `downcast_dictionary_array!` shadows `array`, so the
+            // unchanged branch returns it verbatim, preserving the zero-copy contract that the
+            // Struct/List/Map arms rely on via `Arc::ptr_eq`.
+            let original = Arc::clone(array);
+            downcast_dictionary_array!(
+                array => {
+                    let values = array.values();
+                    let decoded = decode_string_arrays(values)?;
+                    if Arc::ptr_eq(&decoded, values) {
+                        Ok(original)
+                    } else {
+                        Ok(Arc::new(array.with_values(decoded)))
+                    }
+                }
+                t => unreachable!("dictionary type checked by guard: {t}"),
+            )
+        }
+        DataType::Struct(fields) => {
+            let s = array
+                .as_any()
+                .downcast_ref::<StructArray>()
+                .expect("data type checked by caller");
+            let mut changed = false;
+            let mut columns = Vec::with_capacity(s.num_columns());
+            for col in s.columns() {
+                let decoded = decode_string_arrays(col)?;
+                changed |= !Arc::ptr_eq(&decoded, col);
+                columns.push(decoded);
+            }
+            if !changed {
+                return Ok(Arc::clone(array));
+            }
+            Ok(Arc::new(StructArray::new(
+                fields.clone(),
+                columns,
+                s.nulls().cloned(),
+            )))
+        }
+        DataType::List(_) => decode_list::<i32>(array),
+        DataType::LargeList(_) => decode_list::<i64>(array),
+        DataType::FixedSizeList(field, size) => {
+            let list = array
+                .as_any()
+                .downcast_ref::<FixedSizeListArray>()
+                .expect("data type checked by caller");
+            let values = list.values();
+            let decoded = decode_string_arrays(values)?;
+            if Arc::ptr_eq(&decoded, values) {
+                return Ok(Arc::clone(array));
+            }
+            Ok(Arc::new(FixedSizeListArray::try_new(
+                Arc::clone(field),
+                *size,
+                decoded,
+                list.nulls().cloned(),
+            )?))
+        }
+        DataType::Map(field, ordered) => {
+            let map = array
+                .as_any()
+                .downcast_ref::<MapArray>()
+                .expect("data type checked by caller");
+            let entries: ArrayRef = Arc::new(map.entries().clone());
+            let decoded = decode_string_arrays(&entries)?;
+            if Arc::ptr_eq(&decoded, &entries) {
+                return Ok(Arc::clone(array));
+            }
+            let decoded_struct = decoded
+                .as_any()
+                .downcast_ref::<StructArray>()
+                .expect("entries are a struct")
+                .clone();
+            Ok(Arc::new(MapArray::try_new(
+                Arc::clone(field),
+                map.offsets().clone(),
+                decoded_struct,
+                map.nulls().cloned(),
+                *ordered,
+            )?))
+        }
+        // Note: string *view* types (Utf8View, BinaryView, etc.) fall through here undecoded.
+        // Spark's JVM->native FFI only ever emits StringType as Utf8/LargeUtf8 (plus dictionaries
+        // and nesting thereof), so this is safe today. If a future producer starts handing a
+        // view-typed string column across FFI, a decode arm must be added above, or the
+        // from_utf8_unchecked UB this function exists to prevent will silently return.
+        _ => Ok(Arc::clone(array)),
+    }
+}
+
+fn decode_generic_string<O: OffsetSizeTrait>(array: &ArrayRef) -> Result<ArrayRef, ArrowError> {
+    let arr = array
+        .as_any()
+        .downcast_ref::<GenericStringArray<O>>()
+        .expect("data type checked by caller");
+    let len = arr.len();
+    if len == 0 {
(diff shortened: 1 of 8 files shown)


native/core/src/execution/jni_api.rs
```diff
@@ -1315,7 +1316,10 @@ pub unsafe extern "system" fn Java_org_apache_comet_Native_columnarToRowConvert(
             let array_data = from_ffi(ffi_array, &ffi_schema)
                 .map_err(|e| CometError::Internal(format!("Failed to import array: {}", e)))?;
 
-            arrays.push(arrow::array::make_array(array_data));
+            let imported = arrow::array::make_array(array_data);
+            arrays.push(decode_string_arrays(&imported).map_err(|e| {
+                CometError::Internal(format!("Failed to decode imported string array: {}", e))
```
0lai0: nit: Could we map this to `CometError::Arrow` instead of `CometError::Internal`?

`decode_string_arrays` is declared as `Result<ArrayRef, ArrowError>`, and the other two call sites (`scan.rs`, `jvm_udf`) already surface that as `CometError::Arrow` (via `?` or an explicit map). Using `Internal` here misclassifies an Arrow/data failure as a Comet internal error.

`arrays.push(decode_string_arrays(&imported)?)` should be enough, since `CometError` implements `From<ArrowError>`.

native/common/src/utf8.rs
```diff
@@ -0,0 +1,444 @@
...
+                decoded,
+                list.nulls().cloned(),
+            )?))
+        }
+        DataType::Map(field, ordered) => {
+            let map = array
+                .as_any()
+                .downcast_ref::<MapArray>()
+                .expect("data type checked by caller");
+            let entries: ArrayRef = Arc::new(map.entries().clone());
```
0lai0: on the Map arm, `Arc::new(map.entries().clone())` runs even when nothing needs decoding
Suggestion : Could we inline the struct walk over `map.entries()` columns (like the Struct arm) so the valid/zero-copy path doesn't allocate that Arc?

0lai0 (commented):
Thanks @manuzhang for picking this up. left some comments

andygrove:
Thanks for picking this up @manuzhang. I probably can't review in detail until next week. How does this relate to the work in https://github.com/apache/datafusion-comet/pull/5267? Will that PR remain open?

manuzhang:
@andygrove #5267 has been closed in favor of this approach.

andygrove:

This is worth doing. `from_ffi` building string arrays through `new_unchecked` and then arrow-rs handing out `&str` via `from_utf8_unchecked` really is undefined behavior in the default configuration, and closing it at the import boundary is the right place. The zero-copy `Arc::ptr_eq` contract through the nested arms is nicely done, and decoding before `copy_or_unpack_array` so a dictionary decodes its compact values rather than the expanded ones is a good catch.

Three things.

**`aligned_stream_reader.rs` is a fourth import site and is not covered**

`batch_from_ffi` in `native/core/src/execution/operators/aligned_stream_reader.rs` calls `from_ffi_and_data_type`, and its own doc comment says the function "returns the producer's buffers untouched (via `new_unchecked`)". If that stream ever carries a JVM-produced string column, it has exactly the hole this PR exists to close, and it would be the one place left after this merges.

Is that path reachable with string data? If it is, it needs `decode_string_arrays` too. If it is not, the description should list all four `from_ffi` sites and say why that one is exempt, so the next person auditing this does not have to rediscover it.

**The default arm silently passes through**

```rust
_ => Ok(Arc::clone(array)),
```

The comment is honest about the risk and says a view-typed column would silently return the UB. But a comment is not a mechanism. In a function whose entire purpose is to prevent unsoundness, an unknown type should not take the quiet path.

Could the default arm return an error for the types that could plausibly contain strings, specifically `Utf8View`, `ListView`, `LargeListView`, and `RunEndEncoded`, and keep the silent clone only for types that provably cannot? A hard error on an unexpected string-bearing type turns a latent memory-safety bug into a clear failure, and if `Utf8View` ever does arrive over FFI, that is much better than what the comment describes.

**The performance cost needs an end-to-end number**

Every string column crossing the JVM to native boundary now pays a full UTF-8 validation pass per batch. The benchmark measures `decode_string_arrays` in isolation on an 8192-row valid batch, which tells us the function is fast but not what it costs a query.

Could you add a before-and-after on something string-heavy end to end, say TPC-H Q1 or a filter over a wide string column with `spark.comet.scan.impl` set to the JVM path? This is a soundness fix so it should land regardless of the number, but users should be told what it costs, and if it turns out to be material there may be a case for validating once per column rather than once per batch, or for trusting a producer that guarantees validity.

**One smaller note**

`scans.md` is updated, which is right. Does it now say clearly which import boundaries are covered and that Gap A (the native scan) is still open under #4764? A user reading that page should be able to tell whether their configuration is affected.

manuzhang:
@andygrove could you please take another look?

kazuyukitanimura:
Hi @manuzhang 
Thank you for this PR. Would you mind fixing the conflict please?

andygrove (approved):
All three points from my last review are addressed, and I checked the soundness-critical part rather than taking it on trust.

Validating `values[start..end]` and then rejecting any interior offset that lands on a continuation byte is sufficient. In a valid buffer a non-continuation position is a codepoint boundary, and splitting valid UTF-8 at codepoint boundaries leaves every element valid. `offsets[0]` and `offsets[len]` do not need their own check, because a leading or trailing split would already fail the range validation. The `o < values.len()` guard can only cost a trip through the slow path on a sliced array, never a missed split.

Listing all four `from_ffi` sites and recording why `batch_from_ffi` is exempt is what I was after. Making `data_type_contains_string` exhaustive and returning `NotYetImplemented` for the view types turns the comment into a mechanism, which was the point.

One thing is still open, and it is the one you flagged yourself. The end-to-end table is the `std::str::from_utf8` build and the `simdutf8` figures are a projection from the validator microbenchmark. That is honest and I do not think it should hold the fix, since this closes undefined behaviour in the default configuration. Could you re-run `CometStringFfiImportBenchmark` on the EC2 micro-benchmark runner after this lands and update the table, so the numbers users see are measured rather than extrapolated?

Approving.

PR #5314, opened 2026-08-08
feat: support S3 compliant filesystems

### LINKED ISSUE #5313: Support to read S3-compliant object stores
### Motivation

  Comet today accepts only `s3://` and `s3a://` URIs, which effectively couples native execution to AWS S3 and the Hadoop `s3a` connector. In practice, most production Spark deployments live on S3-compliant
  storage that speaks the S3 protocol against a non-AWS endpoint. Users on these systems either lose Comet acceleration entirely or have to rewrite their table locations to fit the `s3a` mold.

  Common S3-compliant systems Spark users route through today:

  - **Google Cloud Storage** via its S3-compatible interoperability endpoint (`storage.googleapis.com`).
  - **Cloudflare R2**, **Backblaze B2**, **Wasabi**, **DigitalOcean Spaces**, **Linode Object Storage**.
  - **MinIO** (self-hosted and as the storage layer for many Kubernetes / on-prem data platforms).
  - **Ceph RADOS Gateway**, **SeaweedFS**, **OpenStack Swift** (with the S3 middleware).
  - **IBM Cloud Object Storage**, **Oracle Cloud Object Storage** (S3 compatibility API).
  - **Dell ECS**, **NetApp StorageGRID**, **Pure FlashBlade** and other enterprise on-prem object stores.
  - Internal / vendor-specific S3-compliant blob services that many organizations expose through a custom `blob://` (or similarly aliased) Hadoop `FileSystem`.

  These systems share a common shape: S3 protocol, custom endpoint, usually path-style addressing, frequently no meaningful region, and often a vendor-specific config namespace instead of `fs.s3a.*`. The
  current gate treats any of these as "unsupported filesystem" and falls back to Spark, which defeats the purpose of Comet on precisely the deployments that would benefit most.

  ### Goal

  Allow Comet to natively read (and Iceberg-scan) tables backed by any S3-compliant object store, regardless of the URI scheme the user's Hadoop `FileSystem` exposes or which vendor namespace they configure
  credentials under. Behavior on AWS S3 and the standard `s3a` connector must remain unchanged.

  ### Out of scope

  - Native writes to S3-compliant stores (there is no S3 write path today).
  - Non-S3 protocols (GCS native HMAC / OAuth, Azure Blob / ADLS Gen2). Those already have their own paths and are not affected.

  ### Compatibility

  Additive. Users on AWS S3 or `s3a` see no change. Users on S3-compliant storage go from "falls back to Spark" to "runs natively in Comet".

_No response_

### Additional context

_No response_

comphead:

Closes #5313

## Rationale for this change

Some environments front an S3-compatible service (MinIO, Ceph RGW, Cloudflare R2,
Wasabi, and similar) with a vendor-branded Hadoop filesystem client that registers
its own URL scheme, for example `blob://`, instead of `s3://` or `s3a://`. Today
Comet does not recognize these schemes, so the native Parquet and Iceberg scans
fall back to Spark. This PR lets Comet treat such schemes as aliases for `s3://`
and read them directly, without the caller rewriting URLs.

- New opt-in config `fs.comet.s3Compliant.schemes` (Hadoop key, set via
  `spark.hadoop.*`/SparkConf/core-site.xml at session start). It is a
  comma-separated, case-insensitive list of schemes to treat as S3 aliases. Empty
  by default, so no alias is claimed unless opted in. Short names like `blob` are
  not unique to S3, so the empty default avoids misrouting (for example Azure Blob).
- Native Parquet scan: `normalize_object_store_url` rewrites alias schemes and
  `s3a` to `s3://bucket/key` so the existing `scheme == "s3"` dispatch fires.
- Native Iceberg scan: routes alias schemes to the S3 backend by scheme instead of
  rewriting the URL, preserving the raw `file_path` bytes that iceberg-rust uses for
  exact-match delete-file resolution.
- Vendor-style per-authority Hadoop keys `fs.<scheme>.<authority>.<property>` are
  translated to the `fs.s3a.*` surface (access/secret/session, endpoint, region,
  path-style). Path-style access defaults to enabled when an endpoint is set.
- `CometScanRule` gates alias schemes for both Parquet and Iceberg and emits clear
  fallback messages listing the supported schemes.
- `CometFileKeyUnwrapper` canonicalizes alias schemes so the put and get sides
  agree on one cache key for Parquet encryption key unwrapping.
- Docs: new S3-compliant filesystem scheme section in the datasources user guide.

- Rust unit tests in `native/core/src/parquet/objectstore/s3_blob_fs_support.rs`
  for URL normalization and alias-scheme detection.
- Scala/Java suites: `NativeConfigSuite`, `CometScanSchemeFallbackSuite`,
  `TestCometFileKeyUnwrapper`, and `CometIcebergNativeScanSuite`.
- End-to-end `blob://` reads against MinIO in `ParquetReadFromS3Suite` and
  `IcebergReadFromS3Suite`, using a `BlobSchemeFileSystem` test shim over
  `S3AFileSystem`

### DIFF
--- a/docs/source/user-guide/latest/datasources.md
+++ b/docs/source/user-guide/latest/datasources.md
@@ -216,6 +216,68 @@ Beyond credential providers, Comet's Parquet scan supports additional S3 configu
 
 All configuration options support bucket-specific overrides using the pattern `fs.s3a.bucket.{bucket-name}.{option}`.
 
+### S3-Compliant Filesystem Schemes
+
+Some environments front an S3-compatible service (MinIO, Ceph RGW, Cloudflare R2, Wasabi, and
+similar) with a vendor-branded Hadoop filesystem client that registers its own URL scheme, for
+example `blob://`, instead of `s3://` or `s3a://`. Comet can treat such schemes as aliases for
+`s3://` so the native Parquet and Iceberg scans read them directly, without the caller rewriting
+URLs.
+
+This is opt-in and disabled by default. Enable it by listing the schemes to treat as S3-compliant
+aliases in `spark.hadoop.fs.comet.s3Compliant.schemes` (Hadoop key
+`fs.comet.s3Compliant.schemes`), a comma-separated, case-insensitive list. This mirrors the
+existing `fs.comet.libhdfs.schemes` config.
+
+```shell
+--conf spark.hadoop.fs.comet.s3Compliant.schemes=blob
+```
+
+Multiple schemes can be listed together:
+
+```shell
+--conf spark.hadoop.fs.comet.s3Compliant.schemes=blob,minio,r2
+```
+
+With no configuration, Comet claims none of these aliases, so for example a `blob://` path falls
+back to Spark unchanged. The empty default is deliberate: short scheme names like `blob` are not
+unique to S3-compatible storage. Azure Blob Storage is the clearest example, so claiming `blob://`
+unconditionally would risk misrouting paths that were never meant for Comet's S3 client. Only add a
+scheme here if you intend Comet to treat it as S3-compatible.
+
+For each scheme `<s>` listed in `fs.comet.s3Compliant.schemes`, Comet also reads vendor-style,
+per-authority Hadoop keys of the form `fs.<s>.<authority>.<property>` (the authority is typically
+the bucket or account name from the URL) and translates them into the `fs.s3a.*` surface described
+above. The recognized vendor-style properties and their `fs.s3a.*` targets are:
+
+| Vendor property (`fs.<s>.<authority>.<property>`) | Translated `fs.s3a.*` suffix |
+| ------------------------------------------------- | ---------------------------- |
+| `awsAccessKeyId`                                  | `access.key`                 |
+| `awsSecretAccessKey`                              | `secret.key`                 |
+| `awsSessionToken`                                 | `session.token`              |
+| `endpoint`                                        | `endpoint`                   |
+| `region`                                          | `endpoint.region`            |
+| `pathStyleAccess`                                 | `path.style.access`          |
+
+Unrecognized `fs.<s>.<authority>.*` properties are ignored.
+
+A URL with no authority, such as the triple-slash `blob:///bucket/key` form, reports its authority
+as the literal string `default`. For that case, list the keys under `fs.<s>.default.<property>`.
+Comet resolves `default` to the bucket taken from the URL path (`bucket` here) and applies the
+translated settings at that bucket's scope, exactly as if they had been written under
+`fs.<s>.bucket.<property>`.
+
+When `fs.<s>.<authority>.endpoint` is set, Comet defaults path-style access to enabled for that
+bucket, since most non-AWS S3-compatible services require it. This is only a default: set
+`fs.s3a.bucket.<bucket>.path.style.access` (or the equivalent per-scheme key) explicitly to
+override it.
+
+Once translated, these values are applied at the same `fs.s3a.bucket.{bucket-name}.*` scope
+described in [Additional S3 Configuration Options](#additional-s3-configuration-options), so the
+credential providers and options documented above also apply to alias-scheme URLs. The same
+translation feeds the native Iceberg scan; see
+[Object store configuration (S3)](iceberg.md#object-store-configuration-s3) in the Iceberg guide.
+
 ### Examples
 
 The following examples demonstrate how to configure S3 access using different authentication methods.
--- a/native/core/src/execution/operators/iceberg_scan.rs
+++ b/native/core/src/execution/operators/iceberg_scan.rs
@@ -255,7 +255,7 @@ impl IcebergScanExec {
         };
         match scheme {
             "file" => Ok(Arc::new(OpenDalStorageFactory::Fs)),
-            "s3" | "s3a" => {
+            "s3" | "s3a" | "blob" => {
                 let customized_credential_load =
                     build_s3_credential_loader(path, catalog_properties, catalog_name);
                 Ok(Arc::new(OpenDalStorageFactory::S3 {
@@ -366,6 +366,22 @@ impl IcebergScanExec {
             if STORAGE_PROPERTY_PREFIXES.iter().any(|p| key.starts_with(p)) {
                 file_io_builder = file_io_builder.with_prop(key, value);
             }
+        }
+
+        // Object-store's AmazonS3Builder defaults the SigV4 region to `us-east-1` when unset;
+        // iceberg-storage-opendal's S3 factory instead errors with `region is missing. Please
+        // find it by S3::detect_region() or set them in env.` Non-AWS S3-compliant storage
+        // services accept any region in the credential, so default to `us-east-1` when the
+        // catalog didn't ship one. Both key spellings iceberg-rust reads (`client.region`
+        // wins over `s3.region`).
+        let region_forwarded = catalog_properties.contains_key("s3.region")
+            || catalog_properties.contains_key("client.region");
+        let is_s3_family = matches!(
+            metadata_location.split_once("://"),
+            Some(("s3" | "s3a" | "blob", _))
+        );
+        if !region_forwarded && is_s3_family {
+            file_io_builder = file_io_builder.with_prop("s3.region", "us-east-1");
         }
 
         Ok(file_io_builder.build())
--- a/native/core/src/execution/planner.rs
+++ b/native/core/src/execution/planner.rs
@@ -98,6 +98,7 @@ use iceberg::expr::Bind;
 use crate::execution::operators::ExecutionError::GeneralError;
 use crate::execution::shuffle::{CometPartitioning, CompressionCodec};
 use crate::execution::spark_plan::SparkPlan;
+use crate::parquet::objectstore::s3_blob_fs_support::normalize_object_store_url;
 use crate::parquet::parquet_support::prepare_object_store_with_configs;
 use datafusion::common::scalar::ScalarStructBuilder;
 use datafusion::common::{
@@ -158,7 +159,6 @@ use num::{BigInt, ToPrimitive};
 use object_store::path::Path;
 use std::cmp::max;
 use std::{collections::HashMap, sync::Arc};
-use url::Url;
 
 // For clippy error on type_complexity.
 type PhyAggResult = Result<Vec<AggregateFunctionExpr>, ExecutionError>;
@@ -450,6 +450,7 @@ impl PhysicalPlanner {
     fn get_partitioned_files(
         &self,
         partition: &SparkFilePartition,
+        object_store_options: &HashMap<String, String>,
     ) -> Result<Vec<PartitionedFile>, ExecutionError> {
         let mut files = Vec::with_capacity(partition.partitioned_file.len());
         partition.partitioned_file.iter().try_for_each(|file| {
@@ -462,10 +463,11 @@ impl PhysicalPlanner {
                 file.start + file.length,
             );
 
-            // Spark sends the path over as URL-encoded, parse that first.
-            let url =
-                Url::parse(file.file_path.as_ref()).map_err(|e| GeneralError(e.to_string()))?;
-            // Convert that to a Path object to use in the PartitionedFile.
+            // Apply the same alias/s3a scheme rewrite the object store construction uses, so the
+            // object-store key we hand DataFusion is stripped of the bucket prefix. Skipping this
+            // would leave `bucket/key` as the object key, and path-style S3 GETs would double the
+            // bucket (`<endpoint>/bucket/bucket/key`).
+            let url = normalize_object_store_url(&file.file_path, object_store_options)?;
             let path = Path::from_url_path(url.path()).map_err(|e| GeneralError(e.to_string()))?;
             partitioned_file.object_meta.location = path;
 
@@ -1714,7 +1716,7 @@ impl PhysicalPlanner {
                 )?;
 
                 // Get files for this partition
-                let files = self.get_partitioned_files(partition_files)?;
+                let files = self.get_partitioned_files(partition_files, &object_store_options)?;
                 let file_groups: Vec<Vec<PartitionedFile>> = vec![files];
 
                 let scan = init_datasource_exec(
@@ -1768,8 +1770,10 @@ impl PhysicalPlanner {
                     one_file,
                     &object_store_options,
                 )?;
-                let files =
-                    self.get_partitioned_files(&scan.file_partitions[self.partition as usize])?;
+                let files = self.get_partitioned_files(
+                    &scan.file_partitions[self.partition as usize],
+                    &object_store_options,
+                )?;
                 let file_groups: Vec<Vec<PartitionedFile>> = vec![files];
                 let scan = init_csv_datasource_exec(
                     object_store_url,
@@ -1841,6 +1845,11 @@ impl PhysicalPlanner {
                     .iter()
                     .map(|(k, v)| (k.clone(), v.clone()))
                     .collect();
+                // Passed RAW to the native scan. `storage_factory_for` routes an opted-in alias
+                // (e.g. blob://, carried in catalog_properties) to the S3 backend by scheme, and
+                // iceberg-storage-opendal's S3 operator is scheme-agnostic, so no URL rewrite is
+                // needed here. Not rewriting keeps data-file and delete-file paths byte-identical,
+                // which iceberg-rust requires to match deletes.
                 let metadata_location = common.metadata_location.clone();
                 let catalog_name = common.catalog_name.clone();
                 let tasks = parse_file_scan_tasks_from_common(common, &scan.file_scan_tasks)?;
@@ -4309,6 +4318,8 @@ fn parse_file_scan_tasks_from_common(
             };
 
             Ok(iceberg::scan::FileScanTaskDeleteFile {
+                // Passed RAW, like `data_file_path` below (same exact-string delete-matching
+                // constraint -- see there).
                 file_path: del.file_path.clone(),
                 file_type,
                 // Not serialized; filled in by IcebergScanExec::fill_delete_file_sizes.
@@ -4530,6 +4541,11 @@ fn parse_file_scan_tasks_from_common(
 
             Ok(iceberg::scan::FileScanTask {
                 file_size_in_bytes: proto_task.file_size_in_bytes,
+                // RAW data-file path -- do NOT rewrite the alias to s3://. iceberg-rust matches
+                // positional deletes by comparing this against the path recorded inside the delete
+                // file, so changing the scheme drops deletes. The S3 backend opens a raw alias path
+                // fine (bucket from host, key sliced verbatim), and `metadata_location` above
+                // already selected the storage factory.
                 data_file_path: proto_task.data_file_path.clone(),
                 start: proto_task.start,
                 length: proto_task.length,
--- a/native/core/src/execution/planner.rs
+++ b/native/core/src/execution/planner.rs
@@ -85,6 +85,9 @@ use iceberg::expr::Bind;
 use crate::execution::operators::ExecutionError::GeneralError;
 use crate::execution::shuffle::{CometPartitioning, CompressionCodec};
 use crate::execution::spark_plan::SparkPlan;
+use crate::parquet::objectstore::blob_alias::{
+    normalize_object_store_url, normalize_object_store_url_string,
+};
 use crate::parquet::parquet_support::prepare_object_store_with_configs;
 use datafusion::common::scalar::ScalarStructBuilder;
 use datafusion::common::{
@@ -143,7 +146,6 @@ use num::{BigInt, ToPrimitive};
 use object_store::path::Path;
 use std::cmp::max;
 use std::{collections::HashMap, sync::Arc};
-use url::Url;
 
 // For clippy error on type_complexity.
 type PhyAggResult = Result<Vec<AggregateFunctionExpr>, ExecutionError>;
@@ -388,6 +390,10 @@ impl PhysicalPlanner {
         partition: &SparkFilePartition,
     ) -> Result<Vec<PartitionedFile>, ExecutionError> {
         let mut files = Vec::with_capacity(partition.partitioned_file.len());
+        // Empty object-store configs are fine here: the only thing `normalize_object_store_url`
+        // consults them for is `is_hdfs_scheme`, and blob/s3a rewriting is the only shape
+        // we care about for the object-store key. HDFS-routed schemes are handled elsewhere.
+        let empty_configs: HashMap<String, String> = HashMap::new();
         partition.partitioned_file.iter().try_for_each(|file| {
             assert!(file.start + file.length <= file.file_size);
 
@@ -398,10 +404,12 @@ impl PhysicalPlanner {
                 file.start + file.length,
             );
 
-            // Spark sends the path over as URL-encoded, parse that first.
-            let url =
-                Url::parse(file.file_path.as_ref()).map_err(|e| GeneralError(e.to_string()))?;
-            // Convert that to a Path object to use in the PartitionedFile.
+            // Apply the same blob/s3a scheme rewrite and three-slash normalization the object
+            // store construction uses, so the object-store key we hand DataFusion is stripped of
+            // the bucket prefix. Skipping this here means paths like `blob:///bucket/key` end up
+            // with `bucket/key` as the object key, and path-style S3 GETs double the bucket
+            // (`<endpoint>/bucket/bucket/key`).
+            let url = normalize_object_store_url(&file.file_path, &empty_configs)?;
             let path = Path::from_url_path(url.path()).map_err(|e| GeneralError(e.to_string()))?;
             partitioned_file.object_meta.location = path;
 
@@ -1752,7 +1760,13 @@ impl PhysicalPlanner {
                     .iter()
                     .map(|(k, v)| (k.clone(), v.clone()))
                     .collect();
-                let metadata_location = common.metadata_location.clone();
+                // Normalize blob/s3a aliases (including Java's `blob:/bucket/key` single-slash
+                // form that Iceberg manifests store) to canonical `s3://bucket/key`. Without
+                // this, storage_factory_for's `contains("://")` check falls back to LocalFs and
+                // iceberg-storage-opendal's fs branch strips one char with `&path[1..]`,
+                // producing `lob:/...` errors.
+                let metadata_location =
+                    normalize_object_store_url_string(&common.metadata_location)?;
                 let catalog_name = common.catalog_name.clone();
                 let tasks = parse_file_scan_tasks_from_common(common, &scan.file_scan_tasks)?;
                 let data_file_concurrency_limit = common.data_file_concurrency_limit as usize;
@@ -3919,7 +3933,8 @@ fn parse_file_scan_tasks_from_common(
             };
 
             Ok(iceberg::scan::FileScanTaskDeleteFile {
-                file_path: del.file_path.clone(),
+                // Normalize blob/s3a aliases so iceberg-rust routes through S3, not LocalFs.
+                file_path: normalize_object_store_url_string(&del.file_path)?,
                 file_type,
                 // Not serialized; filled in by IcebergScanExec::fill_delete_file_sizes.
                 file_size_in_bytes: 0,
@@ -4140,7 +4155,8 @@ fn parse_file_scan_tasks_from_common(
 
             Ok(iceberg::scan::FileScanTask {
                 file_size_in_bytes: proto_task.file_size_in_bytes,
-                data_file_path: proto_task.data_file_path.clone(),
+                // Normalize blob/s3a aliases so iceberg-rust routes through S3, not LocalFs.
+                data_file_path: normalize_object_store_url_string(&proto_task.data_file_path)?,
                 start: proto_task.start,
                 length: proto_task.length,
                 record_count: proto_task.record_count,
--- a/native/core/src/execution/planner.rs
+++ b/native/core/src/execution/planner.rs
@@ -1845,11 +1845,9 @@ impl PhysicalPlanner {
                     .iter()
                     .map(|(k, v)| (k.clone(), v.clone()))
                     .collect();
-                // Passed RAW to the native scan. `storage_factory_for` routes an opted-in alias
-                // (e.g. blob://, carried in catalog_properties) to the S3 backend by scheme, and
-                // iceberg-storage-opendal's S3 operator is scheme-agnostic, so no URL rewrite is
-                // needed here. Not rewriting keeps data-file and delete-file paths byte-identical,
-                // which iceberg-rust requires to match deletes.
+                // Passed RAW, like every other Iceberg location (see `data_file_path` in
+                // `parse_file_scan_tasks_from_common`). `storage_factory_for` routes an opted-in
+                // alias to the S3 backend by scheme, so no URL rewrite is needed here.
                 let metadata_location = common.metadata_location.clone();
                 let catalog_name = common.catalog_name.clone();
                 let tasks = parse_file_scan_tasks_from_common(common, &scan.file_scan_tasks)?;
--- a/native/core/src/lib.rs
+++ b/native/core/src/lib.rs
@@ -160,15 +160,16 @@ pub extern "system" fn Java_org_apache_comet_NativeBase_isFeatureEnabled(
     })
 }
 
-/// JNI method: does object_store recognize this URL's scheme?
+/// JNI: does object_store recognize this URL's scheme?
 ///
-/// This is the source of truth for the JVM planner's "can Comet's native reader handle this
-/// filesystem?" check. Comet's `prepare_object_store_with_configs` dispatches non-hdfs/non-s3
-/// schemes to object_store's `parse_url`, which is driven by `ObjectStoreScheme::parse`; an
-/// unrecognized scheme (e.g. a custom Hadoop FileSystem) fails there at execution time. By
-/// answering from `ObjectStoreScheme::parse` here, the planner can decline early without
-/// hardcoding -- and drifting from -- the object_store-supported scheme set. (hdfs / libhdfs
-/// schemes are handled separately on the JVM side via the user's libhdfs scheme config.)
+/// Source of truth for the JVM planner's "can the native reader handle this filesystem?" check.
+/// `prepare_object_store_with_configs` dispatches non-hdfs/non-s3 schemes to object_store's
+/// `parse_url` (driven by `ObjectStoreScheme::parse`), so answering from the same parser lets the
+/// planner decline early without hardcoding the supported set. (hdfs/libhdfs are handled JVM-side.)
+///
+/// Alias schemes (e.g. `blob`) are NOT answered here: this gate has no config. Their opt-in lives
+/// in the JVM `CometScanRule` via `fs.comet.s3Compliant.schemes`, and opted-in schemes are
+/// rewritten to `s3://` before reaching `prepare_object_store_with_configs`.
 #[no_mangle]
 pub extern "system" fn Java_org_apache_comet_NativeBase_isObjectStoreSchemeSupported(
     env: EnvUnowned,
--- a/native/core/src/lib.rs
+++ b/native/core/src/lib.rs
@@ -160,16 +160,22 @@ pub extern "system" fn Java_org_apache_comet_NativeBase_isFeatureEnabled(
     })
 }
 
-/// JNI: does object_store recognize this URL's scheme?
+/// JNI: can object_store build a store AND an object key for this URL?
 ///
 /// Source of truth for the JVM planner's "can the native reader handle this filesystem?" check.
 /// `prepare_object_store_with_configs` dispatches non-hdfs/non-s3 schemes to object_store's
 /// `parse_url` (driven by `ObjectStoreScheme::parse`), so answering from the same parser lets the
 /// planner decline early without hardcoding the supported set. (hdfs/libhdfs are handled JVM-side.)
 ///
+/// `ObjectStoreScheme::parse` validates the PATH as well as the scheme -- it ends in
+/// `Path::from_url_path` -- so a recognized scheme carrying a key object_store forbids (e.g. a
+/// directory name with a newline) answers false. `CometScanRule` relies on both halves: it probes
+/// a synthetic scheme-only URL for the cached scheme gate and the real URL for the path gate.
+///
 /// Alias schemes (e.g. `blob`) are NOT answered here: this gate has no config. Their opt-in lives
-/// in the JVM `CometScanRule` via `fs.comet.s3Compliant.schemes`, and opted-in schemes are
-/// rewritten to `s3://` before reaching `prepare_object_store_with_configs`.
+/// in the JVM `CometScanRule` via `fs.comet.s3Compliant.schemes`; natively they are rewritten to
+/// `s3://` by `normalize_object_store_url`, the first step of
+/// `prepare_object_store_with_configs`.
 #[no_mangle]
 pub extern "system" fn Java_org_apache_comet_NativeBase_isObjectStoreSchemeSupported(
     env: EnvUnowned,
--- a/native/core/src/lib.rs
+++ b/native/core/src/lib.rs
@@ -169,6 +169,9 @@ pub extern "system" fn Java_org_apache_comet_NativeBase_isFeatureEnabled(
 /// answering from `ObjectStoreScheme::parse` here, the planner can decline early without
 /// hardcoding -- and drifting from -- the object_store-supported scheme set. (hdfs / libhdfs
 /// schemes are handled separately on the JVM side via the user's libhdfs scheme config.)
+///
+/// `blob` is not recognized by `ObjectStoreScheme::parse`, but Comet treats it as a synonym
+/// for `s3` (rewritten in `prepare_object_store_with_configs`), so report it as supported.
 #[no_mangle]
 pub extern "system" fn Java_org_apache_comet_NativeBase_isObjectStoreSchemeSupported(
     env: EnvUnowned,
@@ -179,7 +182,7 @@ pub extern "system" fn Java_org_apache_comet_NativeBase_isObjectStoreSchemeSuppo
         let url_str: String = url.try_to_string(env)?;
         let supported = url::Url::parse(&url_str)
             .ok()
-            .map(|u| object_store::ObjectStoreScheme::parse(&u).is_ok())
+            .map(|u| u.scheme() == "blob" || object_store::ObjectStoreScheme::parse(&u).is_ok())
             .unwrap_or(false);
         Ok(supported)
     })
--- a/native/core/src/parquet/objectstore/blob_alias.rs
+++ b/native/core/src/parquet/objectstore/blob_alias.rs
@@ -1,0 +1,130 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! S3-compatible filesystem alias handling.
+//!
+//! Comet treats `blob://` and `s3a://` as aliases for `s3://` -- they route through the same
+//! `object_store::AmazonS3` / `iceberg-storage-opendal::S3` code paths but the aliased forms are
+//! not directly recognized by those crates. This module rewrites the URL shape (scheme +
+//! `blob:/bucket/key` single-slash form that Java's `URI.toString()` produces from
+//! `blob:///bucket/key` and Iceberg manifests store) so downstream code can treat everything as
+//! canonical `s3://bucket/key`.
+//!
+//! The actual S3 store construction lives in [`super::s3`]; this module is only about URL shape
+//! so all the alias code has one home instead of drifting across `parquet_support.rs`,
+//! `execution/planner.rs`, and `execution/operators/iceberg_scan.rs`.
+
+use std::collections::HashMap;
+
+use url::Url;
+
+use crate::execution::operators::ExecutionError;
+use crate::parquet::parquet_support::is_hdfs_scheme;
+
+/// Parses `url_str` and rewrites `blob`/`s3a` schemes (and the awkward three-slash
+/// `blob:///bucket/key` form that `ObjectStoreScheme::parse` rejects because host=None) to the
+/// canonical `s3://bucket/key`. Non-alias schemes are returned unchanged.
+///
+/// `object_store_configs` is consulted only via `is_hdfs_scheme`: if the user routed `s3a`
+/// through libhdfs via `fs.comet.libhdfs.schemes`, we must NOT rewrite it to `s3` -- HDFS
+/// handling takes over.
+pub(crate) fn normalize_object_store_url(
+    url_str: &str,
+    object_store_configs: &HashMap<String, String>,
+) -> Result<Url, ExecutionError> {
+    let mut url = Url::parse(url_str)
+        .map_err(|e| ExecutionError::GeneralError(format!("Error parsing URL {url_str}: {e}")))?;
+    if is_hdfs_scheme(&url, object_store_configs) {
+        return Ok(url);
+    }
+    let scheme = url.scheme();
+    if scheme != "s3a" && scheme != "blob" {
+        return Ok(url);
+    }
+    let original = scheme.to_string();
+    let needs_host_promotion = url.host_str().is_none();
+    url.set_scheme("s3").map_err(|_| {
+        ExecutionError::GeneralError(format!("Could not convert scheme from {original} to s3"))
+    })?;
+    if needs_host_promotion {
+        // Some deployments emit `blob:///bucket/key` (three slashes, empty authority) or Java
+        // collapses that to `blob:/bucket/key` (opaque form) in Iceberg manifests. In both,
+        // `url::Url` reports host=None and path=`/bucket/key`, but `ObjectStoreScheme::parse`
+        // requires a non-empty host. Lift the first path segment into the host.
+        let trimmed = url.path().trim_start_matches('/').to_string();
+        let (bucket, key) = match trimmed.split_once('/') {
+            Some((b, k)) => (b.to_string(), k.to_string()),
+            None => (trimmed, String::new()),
+        };
+        if bucket.is_empty() {
+            return Err(ExecutionError::GeneralError(format!(
+                "{original}:// URL is missing bucket name: {url}"
+            )));
+        }
+        url = Url::parse(&format!("s3://{bucket}/{key}")).map_err(|e| {
+            ExecutionError::GeneralError(format!("Could not normalize {original}:// URL: {e}"))
+        })?;
+    }
+    Ok(url)
+}
+
+/// String-returning wrapper: iceberg-rust stores paths as owned strings on
+/// `FileScanTask`/`FileScanTaskDeleteFile`, and iceberg-storage-opendal's `storage_factory_for`
+/// picks the LocalFs backend for any path without `://` -- so a single-slash `blob:/...` path
+/// lands in the fs backend and gets its first character stripped (`&path[1..]`), producing
+/// errors like `path: lob:/...`. Rewrite before handing paths to iceberg-rust.
+pub(crate) fn normalize_object_store_url_string(path: &str) -> Result<String, ExecutionError> {
+    Ok(normalize_object_store_url(path, &HashMap::new())?.to_string())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_normalize_object_store_url_string_blob_single_slash_form() {
+        // Java's URI.toString() collapses `blob:///bucket/key` to `blob:/bucket/key` (opaque form
+        // with a leading slash on the path), and that's exactly what Iceberg manifests store.
+        // Iceberg-storage-opendal's `storage_factory_for` uses `path.contains("://")` to detect
+        // the scheme, so a `blob:/...` path routes to the LocalFs backend and its `&path[1..]`
+        // fallback strips the first char, producing `lob:/...` errors. Guard: this string helper
+        // must produce the canonical `s3://bucket/key` before the path reaches iceberg-rust.
+        let out = normalize_object_store_url_string(
+            "blob:/mybucket/tmp/warehouse/db/test_table/data/part-0.parquet",
+        )
+        .expect("single-slash blob URL should normalize");
+        assert_eq!(
+            out,
+            "s3://mybucket/tmp/warehouse/db/test_table/data/part-0.parquet"
+        );
+
+        // Two-slash form (authority present) also normalizes to s3://.
+        let out = normalize_object_store_url_string("blob://bucket/key.parquet").unwrap();
+        assert_eq!(out, "s3://bucket/key.parquet");
+
+        // s3a shares the alias path and gets rewritten too.
+        let out = normalize_object_store_url_string("s3a://bucket/key.parquet").unwrap();
+        assert_eq!(out, "s3://bucket/key.parquet");
+
+        // Non-alias schemes pass through unchanged so file:// / memory:/ Iceberg paths keep
+        // working with iceberg-storage-opendal's Fs / Memory backends.
+        let out = normalize_object_store_url_string("file:///tmp/warehouse/db/t").unwrap();
+        assert_eq!(out, "file:///tmp/warehouse/db/t");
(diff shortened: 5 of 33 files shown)


xiaoh1024:
OSS may also be an S3-compatible alias, so it might be worth adding support for it here.

comphead:
I was experimenting with only one internal S3 compliant system, so not sure if the approach is generic enough for others, but at least it would be a good start

comphead:
Not sure where so many conflicts came from 🤔

xiaoh1024:
> OSS may also be an S3-compatible alias, so it might be worth adding support for it here.

Maybe I can try it for you later.

spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala
```diff
@@ -30,9 +30,11 @@ import org.apache.comet.CometConf.COMET_LIBHDFS_SCHEMES_KEY
 object NativeConfig {
 
   private val objectStoreConfigPrefixes = Map(
-    // Amazon S3 configurations
+    // Amazon S3 configurations. `blob` is a Comet-recognized synonym for `s3` and shares the
+    // same Hadoop `fs.s3a.*` credential surface (see `prepare_object_store_with_configs`).
     "s3" -> Seq("fs.s3a."),
     "s3a" -> Seq("fs.s3a."),
+    "blob" -> Seq("fs.s3a."),
```
parthchandra: `blob` is not a standard scheme — object_store doesn't know it, Hadoop has no blob FileSystem, and the issue itself calls it a "custom / vendor-specific" alias.  Since the goal is "any scheme the user's FileSystem exposes," can we make the alias set a config instead?  For example - `extractObjectStoreOptions — fs.comet.libhdfs.schemes` lets users declare schemes and threads them to native the same way we'd need here.

native/core/src/lib.rs
```diff
@@ -179,7 +182,7 @@ pub extern "system" fn Java_org_apache_comet_NativeBase_isObjectStoreSchemeSuppo
         let url_str: String = url.try_to_string(env)?;
         let supported = url::Url::parse(&url_str)
             .ok()
-            .map(|u| object_store::ObjectStoreScheme::parse(&u).is_ok())
+            .map(|u| u.scheme() == "blob" || object_store::ObjectStoreScheme::parse(&u).is_ok())
```
parthchandra: See comment in `NativeConfig.scala` on making external object store schemes configurable. 
Also, there is an Azure Blob and this can be confusing.

native/core/src/parquet/objectstore/blob_alias.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/// s3-compliant alias schemes Comet rewrites to canonical `s3://` on the Iceberg path. Currently
+/// only `blob`. Kept deliberately narrow: iceberg-rust matches positional/equality deletes by
+/// comparing the delete file's recorded `file_path` (raw, as Iceberg wrote it) against the
+/// `data_file_path` Comet supplies (`delete_filter.rs::get_delete_vector` looks up a map keyed by
+/// the delete file's `file_path` column in `caching_delete_file_loader.rs`). Rewriting a scheme
+/// the delete files were NOT written with desyncs those two and silently drops deletes, so only
+/// schemes that genuinely need an `s3://` shape to reach an object store belong here.
+const S3_COMPLIANT_ALIAS_SCHEMES: &[&str] = &["blob"];
```
parthchandra: See also other places about making `blob' generic. Good this this is already a list since this can have more than one alias.

comphead:
I'm thinking to separate implementations to provide more flexibility

native/core/src/execution/planner.rs
```diff
@@ -4266,7 +4281,8 @@ fn parse_file_scan_tasks_from_common(
 
             Ok(iceberg::scan::FileScanTask {
                 file_size_in_bytes: proto_task.file_size_in_bytes,
-                data_file_path: proto_task.data_file_path.clone(),
+                // Normalize blob/s3a aliases so iceberg-rust routes through S3, not LocalFs.
+                data_file_path: normalize_object_store_url_string(&proto_task.data_file_path)?,
```
sunchao: **[P1] Preserve blob file identity for positional deletes**

Position-delete records retain the original `blob:` data-file path, but this changes the task's identity to `s3:`. iceberg-rust then looks up a different delete-vector key and silently returns deleted rows. With real Parquet and the pinned ArrowReader over memory FileIO, `[10,20,30]` with position 1 deleted returned `[10,30]` before normalization and `[10,20,30]` afterward for all three blob slash forms; s3a/file controls passed. Preserve the original task identity and normalize only storage access. This finding concerns positional deletes, not equality deletes.

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -368,6 +368,22 @@ impl IcebergScanExec {
...
+        // catalog didn't ship one. Both key spellings iceberg-rust reads (`client.region`
+        // wins over `s3.region`).
+        let region_forwarded = catalog_properties.contains_key("s3.region")
+            || catalog_properties.contains_key("client.region");
+        let is_s3_family = matches!(
+            metadata_location.split_once("://"),
+            Some(("s3" | "s3a" | "blob", _))
+        );
+        if !region_forwarded && is_s3_family {
+            file_io_builder = file_io_builder.with_prop("s3.region", "us-east-1");
```
sunchao: **[P2] Honor the environment region before supplying a default**

When catalog properties omit a region, OpenDAL previously honored `AWS_REGION`/`AWS_DEFAULT_REGION`. This unconditional property overrides that behavior for existing s3/s3a tables too. A local HTTP probe using the exact base/head `load_file_io` functions and dummy credentials signed for `us-west-2` before this change and `us-east-1` afterward with `AWS_REGION=us-west-2`. AWS buckets outside east-1 can therefore fail authentication. Honor the existing region sources before defaulting, or restrict the fallback to the intended non-AWS case.

spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala
```diff
@@ -42,6 +44,55 @@ object NativeConfig {
...
+          val s3aSuffix = property match {
+            case "endpoint" => Some("endpoint")
+            case "awsAccessKeyId" => Some("access.key")
+            case "awsSecretAccessKey" => Some("secret.key")
+            case _ => None
+          }
+          s3aSuffix.foreach { suffix =>
+            val scope =
+              if (authority == blobDefaultAuthority) "fs.s3a"
+              else s"fs.s3a.bucket.$authority"
```
sunchao: **[P2] Resolve authority-specific blob settings before constructing Iceberg FileIO**

For `blob://mybucket/...` configured through `fs.blob.mybucket.*`, these bucket-scoped keys become `s3.bucket.mybucket.*` in `hadoopToIcebergS3Properties`. The pinned Iceberg S3 parser only consumes global properties, so the endpoint, credentials and path style are ignored. A local OpenDAL request probe selected the unrelated global endpoint/identity; flattening the selected bucket properties fixed it. This parser limitation already existed for s3a, but the new blob support now depends on it. Resolve the relevant bucket settings to the properties FileIO actually consumes.

spark/src/main/java/org/apache/comet/parquet/CometFileKeyUnwrapper.java
```diff
@@ -103,21 +103,27 @@ public class CometFileKeyUnwrapper {
...
+    // This handles the case where ObjectStoreUrl uses s3:// but Spark uses s3a:// or blob://
     String s3Prefix = "s3://";
     String s3nPrefix = "s3n://";
+    String blobPrefix = "blob://";
     if (filePath.startsWith(s3Prefix)) {
       return "s3a://" + filePath.substring(s3Prefix.length());
     } else if (filePath.startsWith(s3nPrefix)) {
       return "s3a://" + filePath.substring(s3nPrefix.length());
+    } else if (filePath.startsWith(blobPrefix)) {
+      return "s3a://" + filePath.substring(blobPrefix.length());
```
sunchao: **[P2] Canonicalize single/triple-slash blob paths on both cache sides**

When Spark's input paths retain `blob:/bucket/key` or `blob:///bucket/key`, native normalization promotes the bucket and calls back with `s3://bucket/key`. This method leaves the single-slash path unchanged and maps the triple-slash path to `s3a:///bucket/key`, so neither cache entry matches the callback's `s3a://bucket/key`. Production Java cache methods with Spark 4.1.3 path conversion reproduced `Failed to find DecryptionKeyRetriever` for both forms; double-slash blob and s3a controls passed. Apply the same authority normalization when storing and retrieving keys.

native/core/src/parquet/objectstore/blob_alias.rs
```diff
@@ -0,0 +1,235 @@
...
+    if !S3_COMPLIANT_ALIAS_SCHEMES.contains(&url.scheme()) {
+        // file://, s3://, s3a://, ... -> return the ORIGINAL string byte-for-byte. Re-serializing
+        // through `url` (e.g. `String::from(url)`) rewrites the single-slash `file:/x` to the
+        // three-slash `file:///x`, which no longer matches the delete file's stored path.
+        return Ok(path.to_string());
+    }
+    // A recognized s3-compliant alias (blob) -> canonical `s3://bucket/key`, promoting the bucket
+    // out of the path for the single/three-slash forms. `String::from(Url)` moves the Url's
+    // serialization buffer out rather than re-serializing through `Display`.
+    Ok(String::from(rewrite_alias_to_s3(url)?))
```
sunchao: **[P2] Preserve raw Iceberg object keys during alias normalization**

A valid raw location such as `blob://bucket/warehouse/café/part.parquet` becomes `s3://bucket/warehouse/caf%C3%A9/part.parquet` here. Iceberg's OpenDAL adapter strips the prefix without decoding the key, and the S3 backend encodes it again. A local HTTP capture showed `caf%25C3%25A9` in the request instead of `caf%C3%A9`, so the scan targets a different object even without deletes. Preserve the raw key while changing the scheme/authority; the unchanged s3a path passed the same probe.

sunchao (commented):
Reviewed head `3db38b91a719d39be9bd0dade35abdb1580527be`.

Focused validation passed: 11 Rust tests, 11 Scala tests on Spark 4.1.3/JDK 17, and 2 JUnit tests. The inline findings were reproduced with component and local transport probes; these were not full Spark/blob-service end-to-end runs. Native validation used `--no-default-features`.

spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala
```diff
@@ -42,6 +44,55 @@ object NativeConfig {
     "abfs" -> Seq("fs.azure.", "fs.abfs."),
     "abfss" -> Seq("fs.azure.", "fs.abfss.", "fs.abfs."))
 
+  private val blobKeyPattern = "^fs\\.blob\\.([^.]+)\\.(.+)$".r
```
sunchao: **[P2] Keep dots inside blob authorities when translating settings**

For a valid dotted bucket such as `blob://my.bucket/data.parquet`, `fs.blob.my.bucket.endpoint` is parsed as authority `my` and property `bucket.endpoint`, which is discarded. The access-key and secret-key settings are lost the same way. Compiling and calling the actual `NativeConfig` returned an empty map for this case, while the `mybucket` control retained all settings. Match the supported property suffixes and preserve the full authority, otherwise native reads use missing or unrelated defaults.

spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala
```diff
@@ -42,6 +44,55 @@ object NativeConfig {
...
+        case blobKeyPattern(authority, property) =>
+          val s3aSuffix = property match {
+            case "endpoint" => Some("endpoint")
+            case "awsAccessKeyId" => Some("access.key")
+            case "awsSecretAccessKey" => Some("secret.key")
+            case _ => None
+          }
+          s3aSuffix.foreach { suffix =>
+            val scope =
+              if (authority == blobDefaultAuthority) "fs.s3a"
```
sunchao: **[P2] Preserve default-authority precedence after bucket promotion**

For `blob:///mybucket/...`, translating `fs.blob.default.*` only to global `fs.s3a.*` does not override existing `fs.s3a.bucket.mybucket.*`: native `get_config` checks the bucket scope first. An exact Scala/native configuration probe selected an unrelated S3 endpoint and identity despite the configured blob defaults. Translating a separate `fs.blob.mybucket.*` also overrides the default authority for this authorityless URI. Resolve settings using the original blob authority and apply them at the scope actually consulted after normalization.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -1054,6 +1054,26 @@ object CometScanRule extends Logging {
...
+   * you add there. Currently Aliyun OSS via `OpenDalStorageFactory::Oss`.
+   */
+  private val icebergExtraSchemes: Set[String] = Set("oss")
+
+  /**
+   * Scheme gate for the Iceberg scan path. Accepts anything the Parquet scan gate accepts, plus
+   * schemes iceberg-rust reads via OpenDAL that object_store's parser doesn't recognize.
+   */
+  private[rules] def isIcebergReadableScheme(uri: URI): Boolean = {
+    if (isNativelyReadableScheme(uri)) return true
```
sunchao: **[P2] Match Iceberg admission to its actual storage factory**

`object_store` accepts http/https, but `IcebergScanExec.storage_factory_for` rejects those schemes. An S3FileIO table whose metadata/effective location and data locations use `https://bucket/...` therefore changes from JVM fallback to `Unsupported storage scheme: https`. Unmodified Iceberg Java S3FileIO and its Parquet reader read `[10,20,30]` through this alias against a local S3 endpoint; the actual JNI gate accepted it and the exact native factory rejected it. Keep the gate aligned with Iceberg's backends. HTTP data paths alone are not the trigger when metadata remains s3.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -1095,12 +1112,14 @@ object CometScanRule extends Logging {
...
-      // Filesystem scheme check for data file
+      // Filesystem scheme check for data file. Delegated to the native gate
+      // (`isNativelyReadableScheme` -> `NativeBase.isObjectStoreSchemeSupported`) so the iceberg
+      // and Parquet-scan paths share a single source of truth.
       try {
         val filePath = pathMethod.invoke(dataFile).toString
         val uri = new URI(filePath)
         val scheme = uri.getScheme
-        if (scheme != null && !supportedSchemes.contains(scheme)) {
+        if (scheme != null && !isIcebergReadableScheme(uri)) {
```
sunchao: **[P2] Avoid caching raw Iceberg path failures as unsupported schemes**

Iceberg formats a newline partition value as the literal directory `p=%0A`, which its native reader can read. Passing that raw path to `object_store` decodes `%0A` into a control character and returns false. If this is the first `file` check, `schemeSupportCache` stores false for the whole scheme, forcing subsequent ordinary Iceberg and Parquet scans to fall back until JVM restart. The actual reader returned `[10,20,30]` from this path; JNI accepted an ordinary file URL, but the Scala gate rejected it after the first lookup. Use a scheme-only probe rather than caching individual file-parse failures.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
parthchandra: ` ObjectStoreScheme::parse` accepts — including `abfs, abfss, az, azure, http, https, memory`.  Iceberg backend is `storage_factory_for `(iceberg_scan.rs:256), which only handles` file, s3/s3a/blob, gs, oss`

comphead:
Thanks addressed in https://github.com/apache/datafusion-comet/pull/5314/commits/a4fa4ab392983dc037b7f09f80a9b612f37e517f

andygrove (commented):
Thanks for working on this. Before writing anything up I checked the trickiest claims against the pinned iceberg-rust revision (`3d84c81`) rather than taking the comments at face value, and two of them hold up well.

`OpenDalStorage::S3::create_operator` derives its prefix from `url.scheme()` at call time, so a raw `blob://bucket/key` data-file path really does work through the S3 backend. And `DeleteFilter::get_delete_vector` really does key `delete_vectors` by the string in the delete file's `file_path` column and look it up by `file_scan_task.data_file_path()`, so leaving `data_file_path` un-rewritten is the right call. That reasoning is correct and non-obvious, and I am glad it is captured in the code.

CI is fully green.

The rest is what I would like to see addressed. The first item matters more to me than any individual bug.

### 1. The `blob` scheme and the `fs.blob.*` key names are one vendor's conventions, hardcoded

`spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala:36`, `native/core/src/parquet/objectstore/s3_blob_fs_support.rs`

#5313 asks for generic S3-compliant support and lists MinIO, Ceph RGW, R2, Wasabi, GCS interop, and "internal / vendor-specific S3-compliant blob services that many organizations expose through a custom `blob://` (or similarly aliased) Hadoop `FileSystem`". This PR handles exactly one of those, the literal string `blob`, and exactly one vendor's property spelling (`awsAccessKeyId` / `awsSecretAccessKey`). Someone whose FileSystem is registered as `minio://` or `r2://` gets nothing, which is also what @xiaoh1024 was getting at with the `oss` question.

Comet already has the shape for this. `spark.comet.libhdfs.schemes` lets a user declare which schemes route through libhdfs. Could the alias set be a config in the same style, say `spark.comet.s3Compliant.schemes`, threaded through the proto to `is_s3_compliant_alias_scheme` the way `fs.comet.libhdfs.schemes` already reaches `is_hdfs_scheme`? That would cover every store the issue lists instead of one, and it would make the behavior opt-in.

Opt-in matters here for a second reason. Today a `blob://` path falls back cleanly to Spark. After this PR, Comet claims it unconditionally and reads it as S3. Anyone using `blob://` for something that is not S3-compatible (Azure Blob is the obvious collision, since `wasb` literally stands for Windows Azure Storage Blob) goes from a working Spark job to a broken native one, with no way to turn it off. A config default of empty avoids that entirely.

You mentioned upthread that you were thinking about separating the implementations for more flexibility. I think that instinct is right and this is the shape I would suggest.

### 2. `path.style.access=true` is forced on and cannot be turned off

`spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala:82` and `:181`

When a `fs.blob.<authority>.endpoint` is present, `translateBlobKeys` emits `path.style.access=true`, and `extractObjectStoreOptions` applies the blob translations after the `fs.s3a.*` pass with a plain `options(k) = v`. The `getOrElseUpdate` only guards against another blob key, not against the user. So an explicit `fs.s3a.bucket.<bucket>.path.style.access=false` is silently discarded for any `blob://` URI.

Path-style is a good default for these stores, but it is not universal. The GCS interoperability endpoint named in the issue is virtual-hosted, and several of the others accept both. Could the emitted `path.style.access` be skipped when the user has already set it at that scope, so there is an escape hatch?

Related: the translation allowlist is only `endpoint`, `awsAccessKeyId`, and `awsSecretAccessKey`. A user on a blob-scheme bucket cannot express a session token, a region, or path-style at all through the `fs.blob.*` namespace. Is that deliberate, or worth extending to match `hadoopS3aSuffixToIcebergGlobalKey`?

### 3. The single-slash `blob:/bucket/key` form will fail as a data-file path

`native/core/src/execution/planner.rs:4288`

The PR is explicit and repeated about this shape. The `metadata_location` comment says "including Java's `blob:/bucket/key` single-slash form that Iceberg manifests store", and `test_normalize_object_store_url_string_blob_single_slash_form` says "that's exactly what Iceberg manifests store". Delete-file paths get normalized for the same reason.

But `data_file_path` is passed through raw, and it comes straight from `DataFile.location()` in the manifest. In the pinned iceberg-rust, `s3_config_build` does `Url::parse(path)` then `url.host_str().ok_or_else(...)`, and `blob:/bucket/key` parses with a null host. So a manifest that stores the single-slash form would fail the data-file read with `Invalid s3 url: blob:/bucket/key, missing bucket`, even though the metadata and delete files resolved fine.

Which form do the manifests actually store in your deployment? If it really is single-slash, this is worse than an error, because normalizing the data path to `blob://bucket/key` would then desync it from the delete file's recorded `blob:/bucket/key` and drop positional deletes silently. If it is always the two-slash form, then the comments on `metadata_location` and on the tests are overclaiming and should be corrected, since they are the reason a reader would expect `data_file_path` to need the same treatment.

Either way this seems worth pinning down before merge, since one branch of the answer is a silent-wrong-results path on MOR tables.

### 4. `get_partitioned_files` normalizes with an empty config map

`native/core/src/execution/planner.rs:404`

The comment says empty configs are fine because the only thing consulted is `is_hdfs_scheme`. But `is_hdfs_scheme` reads `fs.comet.libhdfs.schemes` out of that map, so with an empty map it falls back to `scheme == "hdfs"`. Meanwhile `prepare_object_store_with_configs` at lines 1644 and 1696 gets the real `object_store_options`. If a user lists `blob` (or `s3a`) in `fs.comet.libhdfs.schemes`, the store is built from the un-rewritten URL while the object key here is computed from the rewritten one, and the two disagree.

Both call sites already have `object_store_options` in scope a few lines above the `get_partitioned_files` call. Could it just be threaded through as a parameter? That removes the divergence rather than reasoning about when it is harmless.

### 5. Two comments now contradict the code they describe

`spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala:1133` says the data-file scheme check is "Delegated to the native gate (`isNativelyReadableScheme` -> `NativeBase.isObjectStoreSchemeSupported`) so the iceberg and Parquet-scan paths share a single source of truth". The code on the next line calls `isIcebergReadableScheme`, the hardcoded allowlist, and the whole point of this PR's change is that the two paths deliberately do not share a source of truth. Looks like a leftover from an earlier revision.

The fallback message at line 674 has the same problem. It now reads "filesystem schemes not recognized by Comet's native object_store", but the Iceberg gate rejects `http`, `https`, `abfs`, `abfss`, `wasb`, `wasbs`, and `memory` precisely because object_store does recognize them and iceberg-rust cannot build them. The old message at least told the user which schemes were supported. Since `icebergReadableSchemes` is a static set, could the message list it?

More generally, the block comments in this PR are very long and lean heavily on the internals of a pinned upstream revision. That is genuinely useful for the delete-matching invariant, which is subtle and easy to break. It is less useful where it restates what the adjacent line does, and item 5 is what happens when it drifts. Worth a pass to keep the invariants and trim the narration.

### 6. Scheme-support cache is keyed by scheme but probed with an authority

`spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala:1039`

The `p=%0A` fix is a good catch and the right shape. One gap remains: `schemeSupportCache` is keyed on the scheme alone, but the probe is built from whichever URI happened to arrive first, `s"$lower:///"` when that URI has no authority. `ObjectStoreScheme::parse` needs a host for `s3`, so a single authorityless URL arriving first would memoize `false` for the whole scheme until JVM restart, which is the same failure mode this change is fixing. Could the probe use a fixed synthetic authority, or the cache key be `(scheme, hasAuthority)`?

### 7. No end-to-end test for the feature

Everything here is unit-tested at the string level. The tests are thorough for what they cover, but none of them exercise an actual read, so nothing verifies that a `blob://` table is really readable end to end.

`CometS3TestBase` already gives you MinIO via testcontainers, and `IcebergReadFromS3Suite` already has a `MOR table with deletes in S3` test. A `blob://` variant of that one and of `ParquetReadFromS3Suite` would cover the parts that unit tests cannot: the raw `data_file_path` plus normalized delete-path combination from item 3, the `path.style.access` behavior from item 2, and the encryption cache-key agreement that `TestCometFileKeyUnwrapper` can only assert about strings. Those suites are excluded from CI as manual, so this is about running them locally and reporting the result.

### 8. `hadoopToIcebergS3Properties` behavior change has no test

`spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala:572`

The rewrite drops `s3.bucket.*` emission entirely and promotes only the target bucket's keys to global. The reasoning is sound, but it is a real behavior change to an untested method, and the `targetBucket: Option[String] = None` default silently restores the old lossy behavior for any future caller that forgets to pass it. There is only one call site, so is the default needed at all? A unit test covering target-bucket promotion, non-target-bucket keys being dropped, and global-key override ordering would lock this in.

### 9. PR description is the unfilled template

Rationale, what changed, and how tested are all still the template comments. For a change this size, with the delete-matching invariant and the deliberate asymmetry between `data_file_path` and delete paths, the reasoning needs to live in the PR rather than only in code comments. It is also where the Iceberg gate change belongs: `abfs`, `abfss`, `wasb`, `wasbs`, and `gcs` are no longer admitted to the native Iceberg scan. I think that is a fix rather than a regression, since `storage_factory_for` would have errored on them anyway, but it is user-visible and currently undocumented anywhere.

### 10. No user documentation

`docs/source/user-guide/latest/datasources.md` is hand-edited and already documents the `fs.s3a.*` surface, per-bucket overrides, and credential providers. A new scheme and a new `fs.blob.<authority>.*` config namespace should be described there, including the path-style default and which properties are translated. `iceberg.md:147` also states that S3 config reaches the native reader "through one of those two channels", which is no longer the complete picture.

spark/src/main/java/org/apache/comet/parquet/CometFileKeyUnwrapper.java
```diff
@@ -92,6 +93,12 @@
  */
 public class CometFileKeyUnwrapper {
 
+  // S3-family schemes that all address the same logical filesystem and must fold to a single
+  // canonical cache key. Kept in sync with the native alias handling
+  // (native/core/src/parquet/objectstore/s3_blob_fs_support.rs and the s3a/s3n aliases object_store
+  // recognizes). The `:` disambiguates them, so order is irrelevant.
+  private static final String[] S3_ALIAS_SCHEMES = {"s3", "s3n", "s3a", "blob"};
```
andygrove: This list is hardcoded to `blob`, but the scheme set is configurable and the docs suggest `blob,minio,r2`. If someone sets `fs.comet.s3Compliant.schemes=minio` and reads encrypted Parquet, I think the read fails.

The put side caches under `minio://bucket/key`, since `CometExecIterator` passes `relation.inputFiles` straight through and nothing here matches. The get side arrives as `s3://bucket/key`, because `parquet_exec.rs` sets `uri_base` from the already-normalized `ObjectStoreUrl`, and that folds to `s3a://bucket/key`. The two keys never meet, so it comes out as `Failed to find DecryptionKeyRetriever`.

`storeDecryptionKeyRetriever` already receives the `Configuration` and already caches it in the `conf` field, so the configured set is available at exactly the right moment. Could the alias list be read from `fs.comet.s3Compliant.schemes` there and cached for the get side, instead of the static array? `TestCometFileKeyUnwrapper` only covers `blob` today, so a `minio://` case alongside it would keep this from drifting again.

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -255,6 +255,9 @@ impl IcebergScanExec {
         };
         match scheme {
             "file" => Ok(Arc::new(OpenDalStorageFactory::Fs)),
+            // blob:// never reaches here: the planner normalizes it to s3:// via
```
andygrove: Does the Iceberg path need the rewrite at all?

The S3 backend derives both the bucket and the key prefix from the path's own scheme: `s3_config_build` takes the bucket from `Url::host_str` without checking the scheme, and `create_operator` builds `format!("{}://{}/", url.scheme(), op_info.name())`. That is exactly why a raw `blob://` data file path works, as the comment on `data_file_path` says. Which means the only reason `metadata_location` gets normalized upstream is to land in this match arm. As far as I can tell it is never opened, only used here, in `build_s3_credential_loader` for the host and path, and in the `is_s3_family` region check below.

If this arm matched configured aliases directly, metadata, delete files and data files could all stay raw, and `normalize_object_store_url_string` could go away entirely, taking both of its invariants with it, along with the asymmetry where the data file path must stay raw but the delete file's own location must not. That asymmetry is the part I would most like to see shrink, because it is exactly the kind of thing a future change breaks silently: get it wrong and deleted rows come back, with no error.

The one thing you would give up is the single-slash `blob:/bucket/key` form, which the tests already describe as defensive rather than a shape Iceberg emits. `is_s3_family` below would need to consult the alias set too. Am I missing a reason the rewrite has to happen upstream in the planner?

spark/src/main/scala/org/apache/comet/CometConf.scala
```diff
@@ -878,6 +878,19 @@ object CometConf extends ShimCometConf {
       .stringConf
       .createOptional
 
+  val COMET_S3_COMPLIANT_SCHEMES_KEY = "fs.comet.s3Compliant.schemes"
+
+  val COMET_S3_COMPLIANT_SCHEMES: OptionalConfigEntry[String] =
```
andygrove, line 948: This entry is never read anywhere. Every consumer goes through `hadoopConf.get(COMET_S3_COMPLIANT_SCHEMES_KEY)` instead, which is different from `COMET_LIBHDFS_SCHEMES`, where `CometScanRule` reads the entry with `.get()`.

The practical consequence is that `spark.conf.set("spark.hadoop.fs.comet.s3Compliant.schemes", ...)` after the session has started is silently ignored: `SessionState.newHadoopConf` copies SQLConf keys with the `spark.hadoop.` prefix still attached, so `fs.comet.s3Compliant.schemes` is never set in the Hadoop conf. Only SparkConf at startup or core-site.xml works, which is fine but not what a user would guess from seeing it in configs.md.

Could this either read the entry the way the libhdfs gate does, or say in the doc string that it has to be set before the session starts?

spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala
```diff
@@ -42,43 +44,144 @@ object NativeConfig {
...
+  // or be an ambient cluster default. Per-bucket synth wins over global in native anyway.
+  private def userSetPathStyle(hadoopConf: Configuration, bucket: Option[String]): Boolean =
+    bucket.exists(b => hadoopConf.get(s"fs.s3a.bucket.$b.path.style.access") != null)
+
+  /**
+   * The S3 bucket a URI addresses: its authority, or -- for the authorityless
+   * `blob:///bucket/key` form -- the first path segment (matching the native rewrite that
+   * promotes it into the host). None when neither is present (e.g. a schemeless root path).
+   */
+  def bucketForUri(uri: URI): Option[String] = {
```
andygrove: For a local Hadoop catalog this returns something odd: `file:///tmp/warehouse/db/t/metadata/v1.metadata.json` has no authority, so the first path segment wins and the bucket comes back as `Some("tmp")`. Every local-catalog Iceberg table then calls `hadoopToIcebergS3Properties` with a target bucket of `tmp`.

Harmless today, since nobody has `fs.s3a.bucket.tmp.*` configured, but it is a surprising value to hand to a function whose entire job is choosing which bucket's credentials get promoted to global. Worth returning `None` when the scheme is not S3-family?

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -654,24 +667,36 @@ case class CometScanRule(session: SparkSession)
               false
           }
 
+        // Opt-in S3-compliant alias schemes (e.g. `blob`) from `fs.comet.s3Compliant.schemes`,
+        // resolved once for this scan (used by the file scheme gate below and its fallback
+        // message).
+        val icebergS3CompliantSchemes =
+          CometScanRule.resolveS3CompliantSchemes(session.sessionState.newHadoopConf())
```
andygrove: Minor: this builds a fresh `Configuration` per Iceberg scan node just to read one key, and `transformV2Scan` already builds another one at line 464 inside the `metadataOpt` block. Could one be hoisted to the top of the method and shared?

docs/source/user-guide/latest/datasources.md
```diff
@@ -216,6 +216,62 @@ Beyond credential providers, Comet's Parquet scan supports additional S3 configu
...
+| Vendor property (`fs.<s>.<authority>.<property>`) | Translated `fs.s3a.*` suffix |
+| ------------------------------------------------- | ---------------------------- |
+| `awsAccessKeyId`                                  | `access.key`                 |
+| `awsSecretAccessKey`                              | `secret.key`                 |
+| `awsSessionToken`                                 | `session.token`              |
+| `endpoint`                                        | `endpoint`                   |
+| `region`                                          | `endpoint.region`            |
+| `pathStyleAccess`                                 | `path.style.access`          |
+
+Unrecognized `fs.<s>.<authority>.*` properties are ignored.
```
andygrove, line 262: The `fs.<s>.default.<property>` case is implemented and tested, where an authority of literally `default` resolves to the bucket taken from the URL path, but it is not mentioned here. Since that is the form an authorityless `blob:///bucket/key` URI produces, it seems worth a sentence so users know the key exists.

andygrove (commented):
I went through this again at 032c2a2d, and rather than take the comments at their word I checked the two claims the design rests on against the pinned iceberg-rust (3d84c81). Both hold. `OpenDalStorage::S3::create_operator` builds its prefix from `url.scheme()` of the path itself and `s3_config_build` takes the bucket from `Url::host_str` without ever looking at the scheme, so the S3 backend really does open a raw `blob://` data path. And `delete_filter.rs` keys delete vectors by exact string match on `data_file_path`, so the "leave the data file path raw" rule is genuinely load-bearing rather than defensive. The comments explaining that are the best part of this PR.

Narrowing the Iceberg allowlist from ten schemes to `{file, s3, s3a, gs, oss}` is also a real fix, not just tidying: `storage_factory_for` on main has no azdls arm, so `abfss://` tables were being admitted by the planner only to die at execution.

One thing looks like an actual bug, and I have a question about whether the Iceberg half needs to rewrite paths at all. Both inline.

The other thing I'd like settled before this merges is test coverage. `dev/ci/check-suites.py` ignores both `ParquetReadFromS3Suite` and `IcebergReadFromS3Suite` as manual suites, so the `blob://` MinIO read, the encrypted `blob://` read, and the `blob://` MOR delete test never run in CI. Those are the three tests that actually exercise what this PR does, and everything green above is unit tests. Could you run them locally and note the result in the description?

spark/src/main/scala/org/apache/comet/objectstore/NativeConfig.scala
```diff
@@ -42,43 +44,157 @@ object NativeConfig {
...
+    val vendorKeyPattern = ("^fs\\." + Pattern.quote(scheme) + "\\.(.+)\\.([^.]+)$").r
+    val out = scala.collection.mutable.Map[String, String]()
+    hadoopConf.iterator().asScala.foreach { entry =>
+      entry.getKey match {
+        case vendorKeyPattern(authority, property) =>
+          vendorPropertyToS3aSuffix.get(property).foreach { suffix =>
+            val bucket =
+              if (authority == vendorDefaultAuthority) defaultBucket else Some(authority)
+            val scope = bucket.map(b => s"fs.s3a.bucket.$b").getOrElse("fs.s3a")
+            out(s"$scope.$suffix") = entry.getValue
```
andygrove: There is no precedence rule here between an authority of `default` and an explicit authority that resolves to the same bucket. Both write `fs.s3a.bucket.<bucket>.<suffix>` through this plain assignment, so whichever one `hadoopConf.iterator()` happens to yield last wins.

With `fs.blob.default.endpoint=https://DEFAULT.example` and `fs.blob.<bucket>.endpoint=https://EXPLICIT.example` both set, I get:

```
blob://mybucket/data/part-0.parquet   -> fs.s3a.bucket.mybucket.endpoint  = https://EXPLICIT.example
blob://warehouse/data/part-0.parquet  -> fs.s3a.bucket.warehouse.endpoint = https://DEFAULT.example
blob://lake/data/part-0.parquet       -> fs.s3a.bucket.lake.endpoint      = https://DEFAULT.example
blob://data/data/part-0.parquet       -> fs.s3a.bucket.data.endpoint      = https://EXPLICIT.example
```

Same config, same code path, different answer depending on where the bucket name lands in the `Configuration` hash order. Across 40 synthetic bucket names it split 30 explicit to 10 default.

This is reachable from the shape the new docs recommend. `fs.<s>.default.*` is the documented spelling for the authorityless form, so anyone with both URI forms in play, or with `fs.blob.default.*` shipped in `core-site.xml` and `fs.blob.<bucket>.*` set at the job level, gets a coin flip on which endpoint and credentials are used. The symptom would be a 403 or a read against the wrong endpoint, with nothing in the config to explain why one bucket works and another does not.

Could the two be ordered, so `default`-authority keys are folded in first and an explicit authority overwrites them? Worth a `NativeConfigSuite` case pinning explicit-beats-default too. The existing `blob:// default authority overrides per-bucket key` test passes only because `mybucket` happens to land on the winning side, so it would not catch a regression here.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -490,7 +499,9 @@ case class CometScanRule(session: SparkSession)
             val hadoopS3Options = NativeConfig.extractObjectStoreOptions(hadoopConf, effectiveUri)
 
             val hadoopDerivedProperties =
-              CometIcebergNativeScan.hadoopToIcebergS3Properties(hadoopS3Options)
+              CometIcebergNativeScan.hadoopToIcebergS3Properties(
+                hadoopS3Options,
+                NativeConfig.bucketForUri(effectiveUri, s3CompliantSchemes))
```
andygrove: `effectiveUri` is the only string `storage_factory_for` and `load_file_io` dispatch on, and it never goes through a scheme gate. It is used here and at line 499, then handed to `CometIcebergNativeScanMetadata.extract` unchecked. `isIcebergReadableScheme` is only applied to data-file and delete-file paths inside `validateIcebergFileScanTasks`.

The comment at line 1094 explains why that gate exists: object_store recognizes schemes iceberg-rust's storage factory cannot build, and admitting them "turns a clean JVM fallback into a native runtime `Unsupported storage scheme` error". That reasoning applies to the metadata location at least as much as to the data files, since it is the value the factory actually matches on. A table whose data files pass the gate but whose metadata location uses a scheme outside `{file, s3, s3a, gs, oss}` plus the configured aliases is claimed by the planner and then dies at execution, which is the exact outcome the narrowing was meant to prevent. It is also the half of the earlier `Match Iceberg admission to its actual storage factory` review point that narrowing the task-path allowlist did not cover.

Would adding `isIcebergReadableScheme(effectiveUri, s3CompliantSchemes)` alongside the existing checks, with a fallback reason, close it? A schemeless local-catalog path already returns true there, so the local Hadoop-catalog case keeps working.

spark/src/test/scala/org/apache/comet/IcebergReadFromS3Suite.scala
```diff
@@ -37,6 +37,22 @@ class IcebergReadFromS3Suite extends CometS3TestBase with RESTCatalogHelper {
     conf.set("spark.sql.catalog.s3_catalog.warehouse", s"s3a://$testBucketName/warehouse")
     applyS3CatalogProps(conf, "s3_catalog")
 
+    // blob:// variant: `blob` is an opt-in S3-compliant alias, backed by an S3A-derived FileSystem
+    // so Iceberg reads/writes blob://<bucket>/... against the same MinIO. Exercises the
+    // delete-matching invariant (metadata/delete paths normalized to s3://, data_file_path raw).
```
andygrove: This comment and the one at line 246 both describe the design the head commit removed. Here it says "metadata/delete paths normalized to s3://, data_file_path raw", and line 246 says "the delete file's own location is normalized to s3://. If either half is wrong, deletes leak". After `413b47f2` nothing on the Iceberg path is normalized, and removing that asymmetry was the whole point of the change. A reader landing here later will go looking for a normalizer that no longer exists.

I would not usually push on wording, but this is the same drift I flagged last round and it recurred inside the commit that fixed the previous instance. It also matters more here than elsewhere: `dev/ci/check-suites.py` ignores this suite, so no CI run will ever surface the next one.

Two smaller ones in the same family while you are in here. `is_s3_family_scheme` at `native/core/src/execution/operators/iceberg_scan.rs:418` says it "Mirrors the Scala `NativeConfig.isS3FamilyScheme` gate", but the Scala one also matches `s3n` and the Rust one does not. And `CometScanRule.scala:517` says the alias key "never leaks" because `load_file_io` filters `catalog_properties`, which is true of `FileIO` but not of `CometS3CredentialBridge::new`, which is handed the unfiltered map and forwards it to the user's credential provider.

andygrove (commented):
I went through this again at `413b47f2`, built it locally, and ran the parts CI covers plus a few probes of my own.

The big change since my last pass is the right one. Dropping `normalize_object_store_url_string` and letting the Iceberg path carry raw paths end to end collapses three invariants into one, and the delete-desync failure mode is now gone by construction rather than by comment. I checked the two upstream facts it rests on against the pinned `3d84c81` rather than taking the comments at face value, and both hold: `OpenDalStorage::create_operator` builds its S3 prefix as `format!("{}://{}/", url.scheme(), op_info.name())` at `crates/storage/opendal/src/lib.rs:302`, and `s3_config_build` takes the bucket from `Url::host_str` without ever reading the scheme. I also confirmed the region change is sound, since opendal 0.58.1 really does fall back to `AWS_REGION` / `AWS_DEFAULT_REGION` when `config.region` is unset (`opendal-service-s3-0.58.1/src/backend.rs:813-822`), and the non-empty check in `env_region_present` is stricter than opendal's own, which is the safer direction.

What I ran locally on Spark 4.1 / JDK 17: the six `s3_blob_fs_support` Rust tests, `NativeConfigSuite` (17), `CometIcebergNativeScanSuite` (10), `CometScanSchemeFallbackSuite` (7), and `TestCometFileKeyUnwrapper` (5). All green. I also probed the percent-encoding round trip through `rewrite_alias_to_s3` including the host-promotion branch, with `café`, `%C3%A9`, `%20` and `%2F` keys, and it is clean. The double-encoding concern from the earlier review does not reproduce on this revision.

Two things I found by running the code rather than reading it, plus some comment drift, are inline.

The remaining thing I would like before this merges is the same one as last time. `dev/ci/check-suites.py` ignores both `ParquetReadFromS3Suite` and `IcebergReadFromS3Suite` as manual suites, so the three tests that actually exercise `blob://` never run anywhere. I do not have Docker on this machine so I could not run them either, which means that as of `413b47f2` nobody has demonstrated that a `blob://` table reads. The tests themselves look right and cover exactly what the unit tests cannot, so this is just about running them and pasting the result.

The description is also still the unfilled template. For a change this size that matters more than usual, because the narrowing of the Iceberg allowlist is user visible and recorded nowhere else: `abfs`, `abfss`, `wasb`, `wasbs` and `gcs` are no longer admitted to the native Iceberg scan. I think that is a fix rather than a regression, since `storage_factory_for` would have errored on them anyway, but it belongs in the description.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -262,13 +262,18 @@ case class CometScanRule(session: SparkSession)
         s.split(",").map(_.trim.toLowerCase(Locale.ROOT)).filter(_.nonEmpty).toSet
```
andygrove: This PR adds `NativeConfig.parseSchemeSet` and documents it as the shared spelling so the JVM gate and the native rewrite parse identically, but `libhdfsSchemes` four lines above still inlines the byte-identical `split(",").map(_.trim.toLowerCase(Locale.ROOT)).filter(_.nonEmpty).toSet`. Could it call the helper too?

```scala
val libhdfsSchemes: Set[String] = COMET_LIBHDFS_SCHEMES.get() match {
  case Some(s) => NativeConfig.parseSchemeSet(s)
  case None => Set("hdfs")
}
```

Keeps the `hdfs` default and leaves one parse to maintain instead of two adjacent ones.

native/core/src/parquet/objectstore/s3_blob_fs_support.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+/// Rewrites a parsed alias URL (`s3a` or a configured alias) to `s3://bucket/key`. When it has no
+/// authority (empty-authority `blob:///bucket/key` or opaque `blob:/bucket/key`, both host=None)
+/// the first path segment is promoted into the host. Defensive: object-store URLs normally have one.
+fn rewrite_alias_to_s3(mut url: Url) -> Result<Url, ExecutionError> {
+    let original = url.scheme().to_string();
+    let needs_host_promotion = url.host_str().is_none();
+    url.set_scheme("s3").map_err(|_| {
```
andygrove: `Url::set_scheme` refuses transitions between the URL spec's special schemes (`file`, `http`, `https`, `ftp`, `ws`, `wss`) and non-special ones, so listing any of those in `fs.comet.s3Compliant.schemes` does not just fail to help. It makes every URL of that scheme return `Err` from `normalize_object_store_url`, and since both `prepare_object_store_with_configs` and `get_partitioned_files` propagate it, the query fails instead of falling back. I confirmed it for all six:

```
fs.comet.s3Compliant.schemes = file
normalize_object_store_url("file:///tmp/x") -> Err("Could not convert scheme from file to s3")
```

A stray `file` in the list would take down every local Parquet scan in the session, and `https` is the one someone might plausibly reach for, since these services are usually described to users by their HTTPS endpoint.

Since the list is user-typed and the failure is a hard error rather than a fallback, would it be worth skipping those schemes in `is_s3_compliant_alias_scheme` the way `s3a` is already excluded, or validating the list once in `CometConf` with a message that says why?

andygrove (approved):
Approving this. I went through the head at `6d94114` and re-checked the claims the design rests on against the pinned dependencies rather than the code comments, and they hold.

The one I most wanted to confirm was the `hadoopToIcebergS3Properties` rewrite, since it drops `s3.bucket.*` emission entirely and that would be a silent credential loss if I had it wrong. In the pinned iceberg-rust (`3d84c81`), `S3Config::try_from` at `crates/iceberg/src/io/storage/config/s3.rs:150` reads only flat global `s3.*` keys — there is no bucket-scoped lookup anywhere in it. So the old `s3.bucket.*` emission was dead config, and promoting the target bucket's keys to global is a genuine fix for existing s3a Iceberg users rather than a behavior change to worry about. I also grepped the native side and nothing reads `s3.bucket.*`, and there are no callers of either changed signature besides `CometScanRule`.

The region default checks out too. `opendal-service-s3-0.58.1/src/backend.rs:813` hard-errors with `region is missing` when `config.region` is `None` and neither `AWS_REGION` nor `AWS_DEFAULT_REGION` is set, so the `us-east-1` fallback can only fire where the read previously failed outright — it cannot misroute a working AWS scan. `env_region_present` mirrors that lookup exactly and its non-empty check is stricter than opendal's, which is the safer direction.

Narrowing the Iceberg allowlist to `{file, s3, s3a, gs, oss}` is a fix rather than a regression, since `storage_factory_for` really does have only Fs/S3/Gcs/Oss arms and `gcs`, `abfs`, `abfss`, `wasb` and `wasbs` were being claimed by the planner only to die at execution.

Two things I chased because they looked like hazards and turned out not to be. `new java.net.URI(metadata.metadataLocation)` at `CometScanRule.scala:547` sits outside a try, but it is the same `effectiveLocation` string already parsed inside the try at line 494, so a malformed location falls back before it ever reaches the new gate. And the normalization added to `get_partitioned_files` is a no-op for every non-alias scheme: `s3a://bucket/key` and `s3://bucket/key` both yield `/key`, hdfs returns early, file is untouched. The registration key and the per-file object key now come from the same rewrite, which is the actual bug being fixed.

I also re-ran @sunchao's percent-encoding finding rather than assuming the redesign killed it, since reading the wrong object silently is the worst failure mode here. Through `rewrite_alias_to_s3`, `%C3%A9` stays `%C3%A9` with no double-encode, `%2F` is not decoded into a separator, `%20` survives, a raw `café` is encoded exactly once, and the host-promotion branch for `blob:/...` and `blob:///...` round-trips identically to the two-slash form. It does not reproduce on this revision.

Locally I built native and ran the eight `s3_blob_fs_support` tests green. CI is green against this exact SHA.

What I could not close is the same thing as last time. `dev/ci/check-suites.py` ignores `ParquetReadFromS3Suite` and `IcebergReadFromS3Suite`, so the `blob://` MinIO read, the encrypted `blob://` read and the `blob://` MOR delete test run nowhere, and I still have no Docker on this machine. I am approving anyway because the feature is opt-in with an empty default and every change on the default path moves in the fallback direction, so a mistake in the alias handling can only reach someone who set `fs.comet.s3Compliant.schemes`. If you can run those three locally before you merge and paste the result, that closes the last gap.

@comphead one housekeeping request: @sunchao's nine review threads are all addressed in the head, but none of them has a reply or has been resolved, so the review state reads worse than the code is. Worth replying with the commit that fixed each and resolving them, so a re-request lands on a clean page.

Two small things that are not blockers. `COMET_S3_COMPLIANT_SCHEMES` is still never read — every consumer goes through `hadoopConf.get` — so `spark.conf.set` after the session starts silently does nothing. The doc string now says so, which I think is fine. And the synthetic probe in `isNativelyReadableScheme` uses `file://comet-probe-host/` for any `file` URI that has an authority, and `object_store-0.13.2/src/parse.rs:109` only has a `("file", None)` arm, so a `file://localhost/...` root path now declines to Spark where it previously went native. Correct results, just a fallback, and an odd path shape.

The description is still the unfilled template. Please fill it in before merging — the Iceberg allowlist narrowing is user-visible and this is the only place it gets recorded.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -1018,30 +1060,48 @@ case class CometScanTypeChecker() extends DataTypeSupport with CometTypeShim {
...
+   * early-fallback optimization and such a build can't run the native scan anyway.
    */
-  private[rules] def isNativelyReadableScheme(uri: URI): Boolean = {
+  private[rules] def isNativelyReadableScheme(
+      uri: URI,
+      s3CompliantSchemes: Set[String]): Boolean = {
     val scheme = uri.getScheme
     if (scheme == null) return true
+    val lower = scheme.toLowerCase(Locale.ROOT)
+    if (s3CompliantSchemes.contains(lower)) return true
```
sunchao, line 1143: **[P1] Preserve bucket identity when admitting multi-bucket alias scans**

With `fs.comet.s3Compliant.schemes=blob`, this now admits a Parquet read spanning two buckets, but native planning still selects the first file's object store for the entire `FilePartition` and strips the authority from every file's object key.

I reproduced this against MinIO with two distinct, equal-size Parquet files at the same key, asserted one native scan and one partition containing both input URLs:

```text
blob://comet-review-first/review-multibucket/same-key.parquet  -> 111
blob://comet-review-second/review-multibucket/same-key.parquet -> 222
Spark: [111, 222]
Comet: [111, 111]
```

Restoring the previous alias-admission behavior on this head makes the same query fall back and return `[111, 222]`. The single-store limitation already exists for plain S3, but this PR newly exposes alias reads that previously fell back. Preserve resolved storage/bucket identity when grouping files, or decline mixed-store alias scans.

sunchao (commented):
Reviewed head `6d9411451470f7ef3499292176a13136c2154f6f` against base `5baa6b035270a31faaa61cb49b047f62a6a3f29e`.

The overall adapter design is reasonable: explicit opt-in, reuse of the existing S3 clients, and preserving Iceberg file identity for delete matching. The remaining gap is admission and storage scope: recognizing a scheme does not establish that every file can use the selected reader, bucket, and configuration.

I found one P1 and three P2 issues, detailed inline. The P1 is a silent wrong-result case: a native alias scan over two buckets returned `[111, 111]` instead of `[111, 222]`. I think these issues should be addressed before merging.

Validation on this head:

- Built native with `--locked --offline --no-default-features`; 11 focused Rust tests passed (8 alias tests and 3 object-store tests).
- Root-reactor testing on Spark 4.1.3 / JDK 17: `NativeConfigSuite`, `CometScanSchemeFallbackSuite`, and `CometIcebergNativeScanSuite` passed all 35 Scala tests; `TestCometFileKeyUnwrapper` passed all 5 Java tests.
- Ran the manual `ParquetReadFromS3Suite` and `IcebergReadFromS3Suite` against MinIO: all 15 tests passed, including the new `blob://` read, encrypted read, and merge-on-read delete case. This closes the earlier execution gap for those fixtures.
- Two additional Parquet regression probes reproduced the wrong-bucket results and the actual-newline path failure. Both passed with Spark fallback after restoring the previous admission decisions on this head. That was a focused causal-control overlay, not a full exact-base query run.
- The two Iceberg findings were reproduced with configuration/FileIO component probes against the pinned dependencies, including actual local HTTP requests; they were not full Spark/Iceberg query reproductions.

Current CI at publication: 65 passing checks and 9 skipped checks. Local validation used the configuration above; it does not cover every Spark version, vendor filesystem client, or the native HDFS feature. The review covered native storage, planner/configuration, encryption/lifetimes, compatibility/tests, and an independent challenge pass.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -490,7 +498,9 @@ case class CometScanRule(session: SparkSession)
             val hadoopS3Options = NativeConfig.extractObjectStoreOptions(hadoopConf, effectiveUri)
 
             val hadoopDerivedProperties =
-              CometIcebergNativeScan.hadoopToIcebergS3Properties(hadoopS3Options)
+              CometIcebergNativeScan.hadoopToIcebergS3Properties(
+                hadoopS3Options,
+                NativeConfig.bucketForUri(effectiveUri, s3CompliantSchemes))
```
sunchao: **[P2] Do not apply metadata-bucket settings to every Iceberg file**

`effectiveUri` selects the metadata bucket, so this promotes that bucket's overrides into the global properties used by one native `FileIO` for all data/delete files. Iceberg supports a separate data location through `write.data.path`.

For a `HadoopFileIO` table with metadata under `s3://metadata/...`, data under `s3://data/...`, working global `fs.s3a.*` data-store settings, and `fs.s3a.bucket.metadata.*` overrides, the base translation sends a data request to the data endpoint with the data credentials. This revision instead sends the same request to the metadata endpoint with metadata credentials. An actual HTTP probe through the pinned FileIO returned success before the change and 404 afterward.

This trigger applies when FileIO properties do not supply overriding `s3.*` settings, and also affects ordinary S3 scans with aliases disabled. Evidence is compiled exact base/head configuration helpers plus FileIO requests, not a full Spark query. Resolve settings for the actual data/delete locations, or fall back when the scan cannot share one configuration.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -1018,30 +1060,48 @@ case class CometScanTypeChecker() extends DataTypeSupport with CometTypeShim {
...
-        _ =>
-          try JBoolean.valueOf(NativeBase.isObjectStoreSchemeSupported(uri.toString))
+        (lower, hasAuthority),
+        _ => {
+          // Probe a scheme(+fixed dummy host) URL, never the caller's authority/path: object_store
+          // keys on (scheme, host-presence) so we cache/probe by that pair. A fixed host stops an
+          // authorityless URL poisoning the authority-bearing form; dropping the real path avoids a
+          // spurious `false` from chars object_store rejects in a `Path` (e.g. Iceberg's `p=%0A`).
+          val probe = if (hasAuthority) s"$lower://comet-probe-host/" else s"$lower:///"
+          try JBoolean.valueOf(NativeBase.isObjectStoreSchemeSupported(probe))
```
sunchao: **[P2] Preserve actual-path validation before selecting native Parquet**

The dummy probe loses a path-specific rejection that native execution still enforces. A local Parquet directory whose name contains an actual newline is readable by Spark; its URI contains `%0A`. The exact native probe returns `false` for that URI and `true` for `file:///`.

In a full Spark reproduction, Spark read `[0, 1, 2]`, while this head selected `CometNativeScanExec` and failed with `Generic URL error` / `Encountered illegal character sequence` when opening the real path. Restoring the previous admission decisions on this head, in a fresh test JVM, restored successful Spark fallback.

This is an actual newline in a Parquet directory name, distinct from Iceberg's literal `p=%0A` path that motivated the earlier cache fix. Keep scheme-capability caching path-independent, but validate the actual Parquet path separately before claiming native execution.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -1054,13 +1114,52 @@ object CometScanRule extends Logging {
...
+   * (`icebergReadableSchemes`) or opt-in S3-compliant aliases (`fs.comet.s3Compliant.schemes`),
+   * plus schemeless local paths (which route to its LocalFs backend).
+   */
+  private[rules] def isIcebergReadableScheme(
+      uri: URI,
+      s3CompliantSchemes: Set[String]): Boolean = {
+    val scheme = uri.getScheme
+    if (scheme == null) return true
+    val lower = scheme.toLowerCase(Locale.ROOT)
+    icebergReadableSchemes.contains(lower) || s3CompliantSchemes.contains(lower)
```
sunchao, line 1252: **[P2] Reject authorityless Iceberg alias file paths**

The scheme-only gate admits an actual data/delete location such as `blob:///bucket/key.parquet`, but Iceberg preserves that raw string and the selected S3 backend requires a URL host. Promoting the bucket while extracting Scala configuration does not change the transport path.

Using the pinned Iceberg/OpenDAL FileIO and a local HTTP success control, `new_input(path).metadata()` succeeds for `blob://bucket/key.parquet` and returns `DataInvalid: ... missing bucket` for `blob:///bucket/key.parquet`. This is a FileIO component reproduction; an authorityless metadata location alone is not sufficient, since it selects the factory without necessarily being opened.

Parquet explicitly promotes the authorityless form, so its support cannot be assumed for the raw Iceberg route. Decline these data/delete paths before native admission, or adapt only their transport location while preserving the original logical identity used for delete matching.

PR #5315, opened 2026-08-08
docs: correct Spark 4.2 version and CI test status in installation guide

andygrove:

N/A — docs correction spotted while reading the installation guide.

## Rationale for this change

The experimental-versions table in the installation guide has two stale claims about Spark 4.2:

1. It lists the version as `4.2.0-preview4`, but the `spark-4.2` Maven profile in `pom.xml` targets the `4.2.0` final release (`<spark.version>4.2.0</spark.version>`).
2. It says "Comet Tests in CI: No", but `.github/workflows/pr_build_linux.yml` has a `"Spark 4.2, JDK 17"` entry in the `linux-test` matrix, so Comet tests do run for 4.2.

- `installation.md`: experimental table row becomes `4.2.0` with "Comet Tests in CI: Yes".
- `compatibility/spark-versions.md`: drop the `-preview4` suffix and reword the warning, which claimed 4.2 support "targets a preview release of Spark".
- `adding_a_new_spark_version.md`: the two `4.2.0-preview4` examples cited a version Comet no longer targets; switched to a hypothetical future release so they keep illustrating the release-qualifier point.

Spark SQL tests still do not run for 4.2 — there is no `spark_4_2` job in `ci.yml` — so that column stays "No" and 4.2 remains in the experimental tier. Per the promotion criteria in `adding_a_new_spark_version.md`, promotion to "Supported" requires both CI columns to be "Yes", so this PR intentionally does not promote it.

Docs only. `npx prettier` reports no changes for all three files.

### DIFF
--- a/docs/source/contributor-guide/adding_a_new_spark_version.md
+++ b/docs/source/contributor-guide/adding_a_new_spark_version.md
@@ -62,7 +62,8 @@ existing profile (for example `spark-4.1`) and update the version
 properties:
 
 - `spark.version`: the full upstream version, including any qualifier
-  (for example `4.2.0-preview4`).
+  (for example `4.3.0` for a final release, or `4.3.0-preview1` while only a
+  preview has been published).
 - `spark.version.short`: the major.minor (for example `4.2`).
 - `parquet.version`, `slf4j.version`, `scala.version`,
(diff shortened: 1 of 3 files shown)


mbutrovich (approved):
Thanks @andygrove!

PR #5322, opened 2026-08-09
perf: optimize string contains function

kazantsev-maksim:

- N/A

## Rationale for this change

Optimize existing expression.

**Performance Optimization for Scalar Haystack (`Scalar vs Array`)**:
   - Replaced the heavy heap allocation `haystack_scalar.to_array_of_size(needle_array.len())` with Arrow's native `StringArray::new_scalar(...)`.
   - Avoids materializing a full $O(N)$ array in memory, shifting to an efficient $O(1)$ broadcasting mechanism (mirroring the optimization already used for scalar needles).

Existing tests.

Benchmark (criterion):

image


native/spark-expr/src/string_funcs/contains.rs
```diff
@@ -134,6 +138,21 @@ fn contains_with_arrow_scalar(
...
+    needle_array: &ArrayRef,
+) -> Result<ArrayRef> {
+    if haystack_scalar.is_null() {
+        return Ok(Arc::new(BooleanArray::new_null(needle_array.len())));
+    }
+
+    let haystack_str = get_string_scalar_value(haystack_scalar, "haystack")?;
+    let haystack_scalar_array = StringArray::new_scalar(haystack_str);
+
+    let result = arrow_contains(&haystack_scalar_array, needle_array)?;
```
sunchao: [P2] Preserve the haystack's Arrow type

Could this use `Scalar::new(haystack_scalar.to_array()?)` instead of always constructing a `StringArray`? Arrow 58.4 requires matching string representations. For example, `LargeUtf8("abc")` with a `LargeStringArray` containing `["a", "bc", null, "", "d"]` returns `[true, true, null, true, false]` before this change, but now fails with `Utf8 CONTAINS LargeUtf8`. The equivalent `Utf8View` inputs also fail. A non-null dictionary-string scalar previously worked too, but `get_string_scalar_value` now rejects it.

I reproduced these against the exact before/after UDFs. I have not established an ordinary Spark query that produces those scalar representations, so this is a native-UDF compatibility finding. Keeping the scalar's type preserves the allocation benefit. Could you add regression tests for these representations?

sunchao (commented):
The scalar-left allocation reduction makes sense. I checked the before/after UDFs against Arrow 58.4.0 and DataFusion 54.1.0 and found one native-UDF type-preservation regression. Details and a minimal fix are inline.

native/spark-expr/src/string_funcs/contains.rs
kazantsev-maksim: Thanks @sunchao! I tried addressed your coment

andygrove:

Using `Scalar::new` for the broadcast side instead of materializing an N-element array is the right change, and the extracted `get_string_scalar_value` helper removes three copies of the same match.

Three things.

**"Existing tests" is thin for the path that changed**

The scalar-haystack, array-needle branch is described in the code as "less common", and it is the one this PR rewrites. Are there existing tests that cover it? Specifically:

- a non-null scalar haystack against an array needle, comparing against the old behavior
- a null scalar haystack, which now takes a new early-return path producing `BooleanArray::new_null(needle_array.len())` instead of going through `arrow_contains`
- an empty needle array

If those are not covered today, this change is unverified. Since the whole point is that this path is rare, it is exactly the one where a regression would sit unnoticed.

**Does `arrow_contains` broadcast a scalar on the left?**

The previous code always passed two arrays. The new code passes `Scalar::new(...)` as the first argument. Some Arrow comparison kernels only broadcast one side. Presumably this works or the build would fail, but a test result rather than an inference would be reassuring, which loops back to the point above.

**Error message consistency**

`contains_array_scalar` and `contains_scalar_scalar` both go through `get_string_scalar_value` and produce "contains function requires string type for haystack/needle, got ...". `contains_scalar_array` does not validate at all and lets Arrow produce whatever message it produces for a non-string datum. Since the helper now exists, would you run the haystack scalar through it there too, so all three paths report the same thing?

**One process note**

The benchmark is an image with no numbers in the text. Images do not survive well in commit history or in a changelog, and they cannot be searched. Could you paste the Criterion medians as a table in the description as well, the way #5174 does? It also makes it much easier for a reviewer to sanity-check the magnitude of the claim.

native/spark-expr/src/string_funcs/contains.rs
rich7420: Thanks for the update! Chao’s dictionary-string scalar case still fails at `get_string_scalar_value` on `79efafd8`, though it passes before this change. Could you preserve support for that case and add a regression test?

Issue #4813, opened 2026-07-03
Distinct-aggregate rewrite can split an incompatible-buffer aggregate across Comet and Spark, causing crash or wrong results


andygrove:
## Describe the bug

An aggregate whose intermediate buffer format is incompatible between Spark and Comet (`CometAggregateExpressionSerde.supportsMixedPartialFinal == false`) can be **split across Comet and Spark within a distinct-aggregate rewrite**, so a Comet-encoded partial buffer is handed to a Spark aggregate (or vice versa). Depending on the aggregate this produces a hard crash or silently wrong results.

This is a pre-existing gap in the mixed-execution guards added for #1389. Those guards only handle a direct single-stage `Partial → Final` pair:

- `CometExecRule.tagUnsafePartialAggregates` / `findPartialAggInPlan` only looked for a `Partial` aggregate feeding *directly* into a non-convertible `Final`, and **stopped at intermediate aggregate stages**.
- `CometBaseAggregate.doConvert` only guarded the `Final` mode (`sparkFinalMode`), not intermediate `PartialMerge` stages.

Spark's single-distinct rewrite (`AggUtils.planAggregateWithOneDistinct`) produces a **four-stage** plan where the non-distinct aggregate's `Partial` is separated from its `Final` by intermediate `PartialMerge` stages:

```
Final(pa, count)            <- consumes buffers
  PartialMerge(pa) + Partial(count distinct)
    PartialMerge(pa)        <- consumes buffers, keyed by the distinct column
      Partial(pa)           <- produces the incompatible buffer
```

If any stage of this chain falls back to Spark while another converts to Comet (e.g. because the distinct aggregate or its grouping key is unsupported by Comet, or via the `comet.exec.*HashAggregate.enabled` test configs), the incompatible buffer crosses the engine boundary.

## Steps to reproduce

Any incompatible-buffer aggregate combined with a distinct aggregate, forced to split. For example with a `TypedImperativeAggregate`:

```sql
SELECT collect_set(v), count(DISTINCT k) FROM t GROUP BY g
```

Forcing the split with `spark.comet.exec.finalHashAggregate.enabled=false` (Comet does the partial/merge, Spark does the final) crashes, because Spark's aggregate tries to deserialize a Comet-encoded buffer.

## Expected behavior

When an incompatible-buffer aggregate cannot run entirely in one engine across the whole distinct-rewrite chain, the entire chain for that aggregate must fall back to Spark, as already happens for the simple `Partial → Final` case.

[...]

PR #5329, opened 2026-08-11
test: add expression fallback-invariance suite

### LINKED ISSUE #5328: Add a fallback-invariance suite: query outcome should not change when a compatible expression is forced from native to Spark

`QueryPlanSerde` honours `spark.comet.expression.<Name>.enabled=false` as forced fallback, and there are 293 expressions registered to it:

```bash
grep -oE "classOf\[[A-Za-z0-9_]+\] ->" spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala | sort -u | wc -l
```

That gives a free invariant, with no product change needed: **for any expression Comet rates compatible, a query's outcome — its rows, or the error it raises — must be identical whether the expression is evaluated natively or forced back to Spark.**

Four distinct tests exercise that lever today, each pinning one hand-picked expression: `disable expression using dynamic config` (`Add`) in `CometExpressionSuite`, plus the `ToPrettyString`, `WidthBucket` and `StringDecode` suites. All four *do* check the answer, via `checkSparkAnswerAndOperator` on the native leg and `checkSparkAnswerAndFallbackReason` on the forced leg. What is missing is a mechanical sweep across the registry, and any systematic check that the two legs agree on **which error** they raise.

### The evidence, and its limits

To test whether an invariance sweep of this shape has teeth, I reapplied the pre-fix condition from #5218 — dropping the `inputOrdinals.size == 1 || rootChildrenAreLeaves(expr)` narrowing from `CometBatchKernelCodegen.canShortCircuitNulls`, so the null short-circuit again applies to a whole tree — and ran a set of detectors against that one probe on one build:

| Detector | Result |
|---|---|
| `CometFuzzMathSuite` | 30/30 pass — miss |
| `CometCodegenFuzzSuite` | 28/28 pass — miss |
| `CometCodegenHOFSuite` | 5/5 pass — miss |
| `org.apache.comet.rules.CometExecRuleSuite` | 29/29 pass — miss |
| `CometCodegenSuite` | 1 failure — the `(#5218)` regression guard |
| `CometCodegenSourceSuite` | 2 failures — both `(#5218)` guards |
| proposed suite | fails: `AddMonths default=VALUE(2024-02-01) forced=THREW(CAST_INVALID_INPUT)` |

**Two honest caveats on that table, stated up front.** The three suites that catch the probe are #5218's own regression tests, shipped with its fix — their failure shows the probe is genuinely behaviour-changing, and nothing more. And my corpus contains `AddMonths` *because* #5218 named it, so the last row is a validation replay, not an independent rediscovery: `CometCodegenSuite`'s guard already covers that same query end-to-end. What the table does support is narrower — four generic mechanical detectors miss this class, including a fuzzer aimed squarely at codegen, while a fallback-invariance comparison surfaces it with a named witness. The comparator is not specific to `add_months`; it would flag this shape for any expression in the corpus.

The oracle also differs from the existing guard's. `CometCodegenSuite`'s test compares Comet against Spark directly. This suite compares Comet-native against Comet-with-that-expression-forced-to-Spark, which is the config lever no current test sweeps.

Same build with the probe reverted: **24 pass / 0 fail / 1 skipped-vacuous** — the single delta is the witness above.

### Describe the solution

A deterministic scalatest suite, no product code, over the existing `CometFuzzTestBase` fixture. Per expression it runs one fixed query twice — default config, then `.enabled=false` — and:

- **gates every comparison on evidence the flip actually moved execution.** The default leg must show Comet operators and no Spark `ProjectExec`; the forced leg must show a Spark `ProjectExec`. A plan that shows neither proves nothing and is reported `SKIPPED-VACUOUS`, never counted as a pass.
- **compares three-valued outcomes** (`Rows` vs `Threw(errorClass)`), so a value on one leg against an exception on the other is a failure with a witness rather than a harness error. Differing error classes are also a failure.
- treats a divergence on an incompatible-rated expression as `EXCUSED` — logged, never a pass.
- canonicalises NaN, `-0.0` and NULL to distinct tokens, and row order by sorting.

25 expressions × the base suite's 3 shuffle/C2R variants = 75 comparisons in **~10 s**, so it is per-PR viable. Compiles and passes under `spark-3.4`, `spark-3.5` and `spark-4.1`.

Current scope is deliberately 25 of 293 expressions. Widening is mechanical, and the bind gate is what keeps the coverage claim honest as it grows — without it a sweep like this silently overstates itself, which happened three times while this one was being built.

### Related work I checked first

- **#4654** (dispatcher vs incompatible native impls) is the reason the sweep reports one skip rather than a pass: `StringTranslate` is incompatible-rated, so by default it never runs natively and there is no native leg to compare against. Whether an expression is reachable natively at all is a dispatcher question, and the bind gate makes that visible instead of silently counting it as verified.
- **#4825 / #4827** (partial project fallback, JVM expression detour) would change the plan shape this suite reads: it keeps an operator native by routing an unsupported subtree through the JVM, so a forced-fallback leg might no longer surface as a Spark `ProjectExec`. That work is default-off today, so there is no conflict now — but the bind gate is the part that would need updating alongside it, and it would report SKIPPED-VACUOUS rather than passing silently if the shape changed underneath it.
- I did not find an existing invariance sweep over this lever; if one exists and I missed it, that is the fastest way to close this issue.

### Additional context

This came out of a systematic, AI-assisted audit of Comet's verification machinery, human-verified at each step and mutation-tested with positive controls. Given #4085's note that LLM-assisted audits have outperformed fuzzing here, the suite is built to that lesson: fast, deterministic, named witnesses. Happy to share the full evidence trail, including the negative results — the audit also produced two candidate findings about the config lever that dissolved on verification, and those are not being filed.

PR attached.

4ktLuffy:

Closes #5328.

## Rationale for this change

For every expression Comet rates compatible, forcing it back to Spark with `spark.comet.expression.<Name>.enabled=false` must not change a query's outcome — neither its rows nor whether and what it throws. The invariant is free: the config already ships, and `QueryPlanSerde` already honours it for all 293 registered expressions.

It is not currently swept. Four tests exercise the lever, each on one hand-picked expression; all four check the answer, but none checks a range of expressions mechanically and none checks that the two legs raise the *same* error.

Replaying #5218's pre-fix logic shows the shape has teeth: the defect is invisible to `CometFuzzMathSuite` (30/30), `CometCodegenFuzzSuite` (28/28), `CometCodegenHOFSuite` (5/5) and `CometExecRuleSuite` (29/29), while this suite names it:

```
FAIL-OUTCOME [errors] AddMonths default=VALUE(2024-02-01) forced=THREW(CAST_INVALID_INPUT)
```

To be clear about what that is and isn't: `AddMonths` is in the corpus *because* #5218 named it, so this is a validation replay rather than an independent find, and `CometCodegenSuite` already guards that exact query end-to-end. The generic-detector misses are the load-bearing part, and the comparator is not specific to `add_months` — it flags this shape for any corpus expression.

One new suite, `CometFallbackInvarianceSuite`, extending `CometFuzzTestBase`. No product code.

- **Fixed query corpus** — 25 expressions over the existing seeded fuzz fixture, no new fixtures. A value-parity section with ANSI off, and an error-parity section with ANSI set explicitly on so it behaves identically on every Spark profile.
- **Bind gate, mandatory.** Before any comparison, the executed plan must prove the flip moved execution: default leg shows Comet operators and zero Spark `ProjectExec`; forced leg shows a Spark `ProjectExec`. A plan showing neither is `SKIPPED-VACUOUS` and reported — never a pass. Plan presence is captured whether or not `collect()` succeeded, so a leg that throws is gated too.
- **Three-valued outcome comparison** (`Rows` vs `Threw(errorClass)`): value-vs-throw is a failure, differing error classes are a failure, row-count mismatch is a failure. Error parity is where #5218 lives, so these comparator rules are load-bearing rather than incidental.
- **Verdicts** — compatible-rated divergence `FAIL` · value-vs-throw `FAIL` · differing error class `FAIL` · length mismatch `FAIL` · incompatible-rated divergence `EXCUSED`, logged but never certified · NaN / `-0.0` / NULL as distinct tokens · row order canonicalised by sorting.

The suite is the test. On a clean tree:

```
FALLBACK-INVARIANCE-RESULT pass=24 fail=0 excused=0 vacuous=1
FALLBACK-INVARIANCE SKIPPED-VACUOUS [values] StringTranslate not-native-by-default
```

24 pass, 0 fail, 1 skipped-vacuous per variant × 3 variants, ~10 s. The one skip is honest and expected: `StringTranslate` is incompatible-rated, so it already runs on Spark by default and there is no native leg to compare against. Verified under `spark-3.4`, `spark-3.5` and `spark-4.1`.

Validated by injected divergence twice: a deliberately broken `spark_decimal_div` as a positive control (also caught by `CometFuzzMathSuite`, as it should be), and the #5218 pre-fix replay. The probe run reports `pass=23 fail=1` — one delta from clean, and it is the witness quoted above.

**On the bind gate.** It earned its place three times over while this suite was being written. An early version reported 22 passes where the truth was 19 passes and 3 vacuous. A second pinned ANSI off and mapped throw-vs-value to a harness error, making it structurally blind to the very bug class it exists for. A third used `ORDER BY` for determinism, which introduces a shuffle and roots the executed plan at `AdaptiveSparkPlanExec` — whose `children` is empty, so plan inspection silently read zero for every query in that section and the gate stopped working entirely. Comet's own `spark.comet.expression.*` tests sidestep all of this by using simple queries with no shuffle; this suite now does the same, avoiding `ORDER BY` and canonicalising row order in the comparator instead.

**Scope, stated plainly.** 25 of 293 expressions. The suite owns one historical bug (#5218) and is not claimed to cover operator-level mixed execution, fallback-*decision* logic, or scan-metadata mismatches — #4813, #4051, #4789 and #2720 are different mechanisms and this invariant would not have caught any of them.

Provenance note: produced during an AI-assisted audit of Comet's verification machinery, human-verified at each gate; mutation experiments were run with positive controls and the tree restored clean afterward.

### DIFF
--- a/spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
@@ -53,17 +53,23 @@ import org.apache.spark.sql.internal.SQLConf
 class CometFallbackInvarianceSuite extends CometFuzzTestBase {
 
   /**
-   * Expressions Comet documents as not guaranteed to match Spark exactly. A divergence here is
-   * reported as EXCUSED: logged for the record, never counted as a pass.
+   * Whether a value divergence for `expr` is excusable under the configuration this suite is
+   * actually running with.
+   *
+   * Derived rather than listed. A static set of "incompatible" names goes stale silently and in
+   * the dangerous direction: it converts real failures into EXCUSED. `RLike` was the concrete
+   * case -- `CometRLike.getSupportLevel` returns `Compatible` on both branches, so the default
+   * path routes through Spark's own codegen and is required to match, yet a hardcoded exemption
+   * still waved opposite Boolean results through as green.
+   *
+   * Comet only evaluates an expression in a mode it documents as possibly-divergent when
+   * `spark.comet.expression.<Name>.allowIncompatible` is set. With it unset, the expression is
+   * either rated `Compatible` and must match, or rated `Incompatible` and never runs natively at
+   * all -- in both cases a divergence is a defect, not an exemption. Reading the live config is
+   * therefore the whole rule, and it self-updates as support levels change.
    */
-  private val incompatibleRated: Set[String] =
-    Set(
-      "DateFormatClass",
-      "FromUTCTimestamp",
-      "GetJsonObject",
-      "RLike",
-      "StringTranslate",
-      "TruncTimestamp")
+  private def divergenceExcusable(expr: String): Boolean =
+    CometConf.isExprAllowIncompat(expr)
 
   private sealed trait Outcome
   private case class Rows(rows: Seq[String]) extends Outcome
@@ -169,7 +175,7 @@ class CometFallbackInvarianceSuite extends CometFuzzTestBase {
             bump("pass")
           } else {
             val i = a.zip(b).indexWhere { case (x, y) => x != y }
-            if (incompatibleRated.contains(expr)) {
+            if (divergenceExcusable(expr)) {
               bump("excused")
               findings += s"EXCUSED [$label] $expr row=$i default=[${a(i)}] forced=[${b(i)}]"
             } else {
--- a/spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
@@ -1,0 +1,305 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.comet
+
+import scala.collection.mutable
+import scala.util.{Failure, Success, Try}
+
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.execution.{ProjectExec, SparkPlan}
+import org.apache.spark.sql.internal.SQLConf
+
+/**
+ * Checks that forcing an expression from Comet's native path back to Spark does not change the
+ * outcome of a query.
+ *
+ * Comet exposes `spark.comet.expression.<Name>.enabled` for every registered expression, and
+ * `QueryPlanSerde` honours it as forced fallback. For any expression Comet rates compatible, that
+ * gives a free invariant: the outcome of a query -- its rows, or the error it raises -- must be
+ * identical whether the expression is evaluated natively or by Spark.
+ *
+ * Two properties of this suite are load-bearing:
+ *
+ *   1. "Outcome" is three-valued. A leg either produces rows or throws. A value on one leg and an
+ *      exception on the other is a failure with a named witness, not an error in the harness --
+ *      exception parity is where several historical divergences have lived.
+ *
+ * 2. Every comparison is gated on evidence that the config flip actually moved execution. A query
+ * whose executed plan proves nothing is reported SKIPPED-VACUOUS and never counted as a pass.
+ * Without that gate an invariance sweep silently overstates its own coverage.
+ *
+ * Queries deliberately avoid `ORDER BY`: a shuffle roots the executed plan at
+ * `AdaptiveSparkPlanExec`, whose `children` is empty, so plan inspection would read zero for
+ * everything and the gate above would be inoperative. Row order is canonicalised in the
+ * comparator instead.
+ */
+class CometFallbackInvarianceSuite extends CometFuzzTestBase {
+
+  /**
+   * Expressions Comet documents as not guaranteed to match Spark exactly. A divergence here is
+   * reported as EXCUSED: logged for the record, never counted as a pass.
+   */
+  private val incompatibleRated: Set[String] =
+    Set(
+      "DateFormatClass",
+      "FromUTCTimestamp",
+      "GetJsonObject",
+      "RLike",
+      "StringTranslate",
+      "TruncTimestamp")
+
+  private sealed trait Outcome
+  private case class Rows(rows: Seq[String]) extends Outcome
+  private case class Threw(errorClass: String) extends Outcome
+
+  /**
+   * What the executed plan proves about where the projection ran. Captured whether or not
+   * execution succeeded, so a leg that throws is still gated rather than waved through.
+   *
+   * `opaque` is the guard that matters: a plan showing neither a Comet operator nor a Spark
+   * `ProjectExec` proves nothing at all, and must never be read as "ran natively".
+   */
+  private case class Presence(cometOps: Int, cometProjects: Int, sparkProjects: Int) {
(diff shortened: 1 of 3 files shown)


spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
```diff
@@ -0,0 +1,305 @@
...
+ * 2. Every comparison is gated on evidence that the config flip actually moved execution. A query
+ * whose executed plan proves nothing is reported SKIPPED-VACUOUS and never counted as a pass.
+ * Without that gate an invariance sweep silently overstates its own coverage.
+ *
+ * Queries deliberately avoid `ORDER BY`: a shuffle roots the executed plan at
+ * `AdaptiveSparkPlanExec`, whose `children` is empty, so plan inspection would read zero for
+ * everything and the gate above would be inoperative. Row order is canonicalised in the
+ * comparator instead.
+ */
+class CometFallbackInvarianceSuite extends CometFuzzTestBase {
```
sunchao, line 53: [P2] Register the suite in both CI workflows

Could you add `org.apache.comet.CometFallbackInvarianceSuite` to the suite matrices in both `.github/workflows/pr_build_linux.yml` and `.github/workflows/pr_build_macos.yml`? They use explicit suite lists. `python3 dev/ci/check-suites.py` passes on the base but exits 255 on this head, and the [Preflight job](https://github.com/apache/datafusion-comet/actions/runs/31517356712/job/96231037450) fails on this class before the downstream tests can run.

spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
```diff
@@ -0,0 +1,305 @@
...
+  /**
+   * Expressions Comet documents as not guaranteed to match Spark exactly. A divergence here is
+   * reported as EXCUSED: logged for the record, never counted as a pass.
+   */
+  private val incompatibleRated: Set[String] =
+    Set(
+      "DateFormatClass",
+      "FromUTCTimestamp",
+      "GetJsonObject",
+      "RLike",
```
sunchao: [P2] Do not excuse compatible RLike mismatches

Could you remove this exemption or derive it from the active implementation and configuration? At this head, [`CometRLike.getSupportLevel`](https://github.com/apache/datafusion-comet/blob/e3c6fea350e44558194090945f3a0b0bd604e2af/spark/src/main/scala/org/apache/comet/serde/strings.scala#L378-L399) rates the default JVM-codegen path compatible. The incompatible Rust regex implementation requires `spark.comet.expression.RLike.allowIncompatible=true`, which this suite does not enable. With valid execution-plan evidence, opposite Boolean results are nevertheless recorded as `EXCUSED`, leaving the test green. I reproduced that with the unchanged comparator on Scala 2.12.18 and 2.13.17. The same mismatch for `EqualTo` correctly produces `FAIL-VALUE`.

sunchao (commented):
The execution-plan gate and value-versus-error comparison are useful. I found two issues to address: registering the suite in CI and removing the stale exemption for the compatible default RLike path. I verified the missing-suite failure against the base and head, reproduced the RLike comparator false negative on Scala 2.12.18 and 2.13.17, and checked the comparator's outcome branches. I did not run the full native Spark suite.

4ktLuffy:
Both addressed — registered in both workflow matrices (check-suites.py now exits 0),
and the exemption is derived from the live config rather than listed:

    private def divergenceExcusable(expr: String): Boolean =
      CometConf.isExprAllowIncompat(expr)

Reproduced your experiment: with the static set, an injected mismatch on RLike
records EXCUSED (suite green) while the same mismatch on EqualTo gives FAIL-VALUE.
With the derived predicate, RLike gives FAIL-VALUE. All six exempted names returned
allowIncompatible=false under this suite's config, so RLike wasn't the only stale one.

Deriving does leave the EXCUSED branch unreachable here, since nothing in this suite
sets allowIncompatible — happy to delete it instead if you'd rather not carry dead code.

Rebased on latest main.

sunchao (approved):

sunchao:
Merged, thanks for the contribution @4ktLuffy !

Issue #5323, opened 2026-08-10
Sort-aware Iceberg reads in Comet via a per-partition streaming merge


parthchandra:
## What are we trying to do?

When an Iceberg table is already sorted, we want Comet to *know* that and stop re-sorting data
that is already in order. More precisely, when Comet's native scan reads a sorted Iceberg table, we
want each Spark partition to come out already sorted, and we want to tell Spark's optimiser that
it is sorted — so redundant `Sort` and shuffle stages in sort-merge joins, grouped aggregates, and
windows over that data simply disappear (and an `ORDER BY ... LIMIT` becomes a cheap bounded take).

In short - sort the data once when you write it, and never pay to sort it again on the way
out.

## How is it done today, and what are the limits?

Upstream Iceberg is building exactly this pattern:

- [apache/iceberg#14948](https://github.com/apache/iceberg/pull/14948) implements the Spark DSv2
  `SupportsReportOrdering` API and a k-way merge reader, so a partition that holds several
  already-sorted files still hands Spark one globally sorted stream. Spark can then drop the sort.
- [apache/iceberg#16305](https://github.com/apache/iceberg/pull/16305) adds a `k-way-merge`
  `RewriteDataFiles` strategy that keeps files sorted over time cheaply — no shuffle, no spill —
  instead of re-running a full sort compaction.

The two reinforce each other: #16305 keeps the table cheaply sorted, and #14948 makes every read
benefit from it.

Comet's native Iceberg scan does not take part in any of this yet. Today it hands *all* of a
partition's files to iceberg-rust in one call that reads them **unordered** and concatenates the
result, and it hard-codes the scan's reported ordering to "none". So the moment Comet takes over a
scan from Spark, the ordering Iceberg worked out is thrown away, and Spark has to re-sort — even
though the data on disk was already in order. We get the correctness, but we pay for a sort we
should not need.

## What is new in our approach, and why do we think it will work?

We will replicate Iceberg's read-side behaviour natively, using a building block DataFusion already has:
its order-preserving k-way merge (`SortPreservingMergeExec`, a loser-tree merge over already-sorted
streams).

So for each Spark partition:

- read each sorted file in the partition as its own stream (instead of one unordered lump);
- feed those streams into `SortPreservingMergeExec`, which merges them into a single sorted stream;
- advertise that ordering out of the scan so Spark eliminates the sort above it.

Why we are confident:

- We are not inventing a merge — we reuse DataFusion's tested loser-tree merge, its memory
[...]

PR #5331, opened 2026-08-11
feat: Sort-aware Iceberg reads in Comet via a per-partition streaming merge

### LINKED ISSUE #5337: Sort-aware Iceberg reads: per-partition streaming merge + report ordering and grouping (v1)

  The native Iceberg scan now preserves the sort order instead of throwing it away. Within each Spark partition it reads each already-sorted file as its own stream and k-way-merges them into one sorted stream, then tells Spark the output is sorted so redundant sorts can be dropped. It also reports Iceberg's key-grouped partitioning so
  storage-partitioned joins can skip the shuffle.

  Scope of this first version:
  - Ordering reporting is limited to plain-column (identity) sort keys that are in the projection. Anything else falls back to an unordered read, so it's always correct.
  - Two flags: spark.comet.scan.icebergNative.sortMerge.enabled (on by default) and spark.comet.scan.icebergNative.reportPartitioning.enabled (off by default).
  - No changes to iceberg-rust.

_No response_

### Additional context

_No response_

parthchandra:

Closes #5337

## Rationale for this change

When Comet reads a sorted Iceberg table, the native scan throws away the ordering to get
parallelism. Spark then can't tell the data is already sorted and re-sorts on every read — in
joins, aggregates, windows, and order-by queries — even though the sort was done at write time.

This PR makes the native scan preserve and report that ordering so Spark can drop the redundant
sorts. It builds on Iceberg's `SupportsReportOrdering` (apache/iceberg#14948): when Iceberg reports
a sort order, Comet merges the sorted files per partition and tells Spark the result is sorted.

For each Spark partition, the scan reads every sorted file as its own stream and k-way-merges them
into one sorted stream with DataFusion's `SortPreservingMergeExec`.

Summary of changes:
- **Proto**: `IcebergScanCommon` carries the reported sort order (`table_sort_orders`) and the
  merge cap (`max_files_per_partition`) to the native side.
- **Scala (rule + serde + scan exec)**: report the sort order to Spark, gated to the safe v1 case —
  identity-transform sort fields on top-level columns that are in the projection. Anything else
  (a transform sort key, a UUID sort key whose Iceberg byte order differs from Spark's string order,
  or a schema that can't be read) reports nothing and stays on Spark, so the result is always
  correct. The gate is evaluated and bound to proto once at planning time, so a binding failure
  falls back cleanly instead of erroring at task start.
- **Native (Rust)**: `IcebergScanExec` becomes multi-partition when an ordering is present (one
  sorted stream per file), and the planner wraps it in `SortPreservingMergeExec`. Above
  `maxFilesPerPartition` files in a partition it instead reads unordered and wraps a spillable
  `SortExec`, which bounds concurrently-open readers. Both paths produce sorted output. No changes
  to iceberg-rust.

Config flags:
- `spark.comet.scan.icebergNative.sortMerge.enabled` (default **on**) — do the per-partition merge.
  When off, the scan stays native and still honours the reported order via a spillable sort instead
  of the k-way merge (equivalent to `maxFilesPerPartition = 0`). Only does anything when Iceberg's
  `spark.sql.iceberg.planning.preserve-data-ordering` is on (off by default).
- `spark.comet.scan.icebergNative.sortMerge.maxFilesPerPartition` (default **64**) — the most files
  in one partition Comet will k-way merge before falling back to the spillable sort.

Note: a global `ORDER BY` keeps its final sort — a per-partition merge isn't a cluster-wide order —
so that case is unchanged.

- **Native planner tests** in `planner.rs`: merge-vs-sort at the `maxFilesPerPartition` boundary, and
  that the cap is read from the proto (a low cap flips a 3-file scan to the sort path).
- **End-to-end suite** `CometIcebergSortMergeReadSuite` over real Iceberg tables (local Hadoop
  catalog, sort order set via the Iceberg Java API, one file per insert), plus unit tests of the
  reportability gate. Correctness (`checkSparkAnswer`) runs on any Iceberg build; the
  sort/shuffle-elimination assertions self-cancel on the published Iceberg that doesn't report an
  ordering.

Key edits: dropped the key-grouped-partitioning paragraph and the reportPartitioning.enabled flag (both gone with the outputPartitioning override), added maxFilesPerPartition and the enabled=false semantics, fixed the malformed Closes # link, and refreshed the tests section. Double-check #5337 is the right issue — the old body had a malformed link; I preserved the number.

### DIFF
--- a/native/core/src/execution/operators/iceberg_scan.rs
+++ b/native/core/src/execution/operators/iceberg_scan.rs
@@ -28,7 +28,7 @@ use arrow::datatypes::SchemaRef;
 use datafusion::common::{DataFusionError, Result as DFResult};
 use datafusion::execution::{RecordBatchStream, SendableRecordBatchStream, TaskContext};
 use datafusion::physical_expr::expressions::Column;
-use datafusion::physical_expr::{EquivalenceProperties, PhysicalExpr};
+use datafusion::physical_expr::{EquivalenceProperties, LexOrdering, PhysicalExpr};
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
 use datafusion::physical_plan::metrics::{
     BaselineMetrics, Count, ExecutionPlanMetricsSet, MetricBuilder, MetricsSet,
@@ -86,6 +86,24 @@ pub struct IcebergScanExec {
     tasks: Vec<FileScanTask>,
     /// Number of data files to read concurrently
     data_file_concurrency_limit: usize,
+    /// FileIO (and, for S3, the JVM credential bridge behind it) built once at plan time and shared
+    /// across partitions. FileIO is cheap to clone (Arc-backed), so each `execute` clones this
+    /// rather than rebuilding the storage factory + credential bridge. This matters in the ordered
+    /// path, where the scan is one partition per file and `execute` is called once per file.
+    file_io: FileIO,
+    /// Table sort order Iceberg reported, translated against `output_schema`. `Some` makes this a
+    /// multi-partition scan: one sorted stream per task, which a SortPreservingMergeExec above
+    /// merges back into one sorted partition. It is also advertised in `plan_properties`. `None`
+    /// keeps the old single-partition unordered read (all tasks streamed together).
+    ///
+    /// Concurrency note: in the ordered path each partition reads exactly one task, so
+    /// `data_file_concurrency_limit` no longer bounds cross-file concurrency; instead the wrapping
+    /// SortPreservingMergeExec drives one reader per file to merge them. That fan-out (files per
+    /// Spark partition) is intrinsic to a k-way merge of per-file sorted streams -- the files must
+    /// be read as separate streams to stay individually sorted -- and is the natural granularity
+    /// for a sorted Iceberg table. `data_file_concurrency_limit` still bounds delete-file stats and
+    /// the unordered path.
+    ordering: Option<LexOrdering>,
     /// Metrics
     metrics: ExecutionPlanMetricsSet,
 }
@@ -98,11 +116,30 @@ impl IcebergScanExec {
         catalog_name: String,
         tasks: Vec<FileScanTask>,
         data_file_concurrency_limit: usize,
+        ordering: Option<LexOrdering>,
     ) -> Result<Self, ExecutionError> {
         let output_schema = schema;
-        let plan_properties = Self::compute_properties(Arc::clone(&output_schema), 1);
+        // With an ordering, read each task as its own sorted stream on its own partition, so the
+        // SortPreservingMergeExec above can merge them. Without one, keep the single partition that
+        // reads every task. Comet only drives execute(0), so a multi-partition leaf with no merge
+        // above it would read only the first task.
+        let num_partitions = if ordering.is_some() {
+            tasks.len().max(1)
+        } else {
+            1
+        };
+        let plan_properties = Self::compute_properties(
+            Arc::clone(&output_schema),
+            num_partitions,
+            ordering.as_ref(),
+        );
 
         let metrics = ExecutionPlanMetricsSet::new();
+
+        // Build FileIO (and the S3 credential bridge) once here rather than per `execute`. In the
+        // ordered path `execute` is called once per file, so rebuilding it there would repeat the
+        // JNI/reflection credential-bridge construction for every file in the partition.
+        let file_io = Self::load_file_io(&catalog_properties, &metadata_location, &catalog_name)?;
 
         Ok(Self {
             metadata_location,
@@ -112,13 +149,26 @@ impl IcebergScanExec {
             catalog_name,
             tasks,
             data_file_concurrency_limit,
+            file_io,
+            ordering,
             metrics,
         })
     }
 
-    fn compute_properties(schema: SchemaRef, num_partitions: usize) -> Arc<PlanProperties> {
+    fn compute_properties(
+        schema: SchemaRef,
+        num_partitions: usize,
+        ordering: Option<&LexOrdering>,
+    ) -> Arc<PlanProperties> {
+        let eq_properties = match ordering {
+            Some(lex) => EquivalenceProperties::new_with_orderings(
+                Arc::clone(&schema),
+                std::iter::once(lex.iter().cloned()),
+            ),
+            None => EquivalenceProperties::new(Arc::clone(&schema)),
+        };
         Arc::new(PlanProperties::new(
-            EquivalenceProperties::new(schema),
+            eq_properties,
             Partitioning::UnknownPartitioning(num_partitions),
             EmissionType::Incremental,
             Boundedness::Bounded,
@@ -152,10 +202,18 @@ impl ExecutionPlan for IcebergScanExec {
 
     fn execute(
         &self,
-        _partition: usize,
+        partition: usize,
         context: Arc<TaskContext>,
     ) -> DFResult<SendableRecordBatchStream> {
-        self.execute_with_tasks(self.tasks.clone(), context)
+        // With a reported ordering this is a multi-partition operator: partition `i` reads only
+        // task `i` as its own sorted stream, and the SortPreservingMergeExec above merges them.
+        // Without one, the single partition reads every task together (legacy unordered path).
+        let tasks = if self.ordering.is_some() {
+            self.tasks.get(partition).cloned().into_iter().collect()
+        } else {
+            self.tasks.clone()
+        };
+        self.execute_with_tasks(tasks, context)
     }
 
     fn metrics(&self) -> Option<MetricsSet> {
@@ -172,11 +230,7 @@ impl IcebergScanExec {
         context: Arc<TaskContext>,
     ) -> DFResult<SendableRecordBatchStream> {
         let output_schema = Arc::clone(&self.output_schema);
-        let file_io = Self::load_file_io(
-            &self.catalog_properties,
-            &self.metadata_location,
-            &self.catalog_name,
-        )?;
+        let file_io = self.file_io.clone();
         let batch_size = context.session_config().batch_size();
 
         let metrics = IcebergScanMetrics::new(&self.metrics);
@@ -640,6 +694,7 @@ mod tests {
     use iceberg_storage_opendal::OpenDalStorageFactory;
 
     use super::IcebergScanExec;
+    use datafusion::physical_plan::ExecutionPlan;
 
     fn fs_file_io() -> FileIO {
         FileIOBuilder::new(Arc::new(OpenDalStorageFactory::Fs)).build()
@@ -739,6 +794,59 @@ mod tests {
         IcebergScanExec::fill_delete_file_sizes(&mut tasks, &fs_file_io(), 4)
             .await
             .unwrap();
+    }
+
+    fn int_schema() -> arrow::datatypes::SchemaRef {
+        use arrow::datatypes::{DataType, Field, Schema};
+        Arc::new(Schema::new(vec![Field::new("a", DataType::Int32, false)]))
+    }
+
+    fn single_col_ordering() -> Option<datafusion::physical_expr::LexOrdering> {
+        use arrow::compute::SortOptions;
+        use datafusion::physical_expr::expressions::Column;
+        use datafusion::physical_expr::{LexOrdering, PhysicalSortExpr};
+        LexOrdering::new(vec![PhysicalSortExpr {
+            expr: Arc::new(Column::new("a", 0)),
+            options: SortOptions::default(),
+        }])
+    }
+
+    // Builds a scan over three empty-delete tasks with the given reported ordering.
+    fn exec_with_ordering(
+        ordering: Option<datafusion::physical_expr::LexOrdering>,
+    ) -> IcebergScanExec {
+        use std::collections::HashMap;
+        let tasks = vec![
+            task_with_deletes(vec![]),
+            task_with_deletes(vec![]),
+            task_with_deletes(vec![]),
+        ];
+        IcebergScanExec::new(
+            "metadata.json".to_string(),
+            int_schema(),
+            HashMap::new(),
+            "cat".to_string(),
+            tasks,
+            1,
+            ordering,
+        )
+        .unwrap()
+    }
+
+    // A reported ordering turns the scan into a multi-partition operator (one partition per task)
+    // so a SortPreservingMergeExec above can k-way merge the per-file sorted streams.
+    #[test]
+    fn reported_ordering_makes_scan_multi_partition() {
+        let exec = exec_with_ordering(single_col_ordering());
+        assert_eq!(exec.properties().partitioning.partition_count(), 3);
+    }
+
+    // Without a reported ordering the scan stays single-partition (Comet drives only execute(0),
+    // which must read every task), preserving the legacy unordered behaviour.
+    #[test]
+    fn no_ordering_keeps_single_partition() {
+        let exec = exec_with_ordering(None);
+        assert_eq!(exec.properties().partitioning.partition_count(), 1);
     }
 
     fn from_hex(s: &str) -> Vec<u8> {
--- a/native/core/src/execution/operators/iceberg_scan.rs
+++ b/native/core/src/execution/operators/iceberg_scan.rs
@@ -213,7 +213,7 @@ impl ExecutionPlan for IcebergScanExec {
         } else {
             self.tasks.clone()
         };
-        self.execute_with_tasks(tasks, context)
+        self.execute_with_tasks(tasks, partition, context)
     }
 
     fn metrics(&self) -> Option<MetricsSet> {
@@ -227,13 +227,14 @@ impl IcebergScanExec {
     fn execute_with_tasks(
         &self,
         tasks: Vec<FileScanTask>,
+        partition: usize,
         context: Arc<TaskContext>,
     ) -> DFResult<SendableRecordBatchStream> {
         let output_schema = Arc::clone(&self.output_schema);
         let file_io = self.file_io.clone();
         let batch_size = context.session_config().batch_size();
 
-        let metrics = IcebergScanMetrics::new(&self.metrics);
+        let metrics = IcebergScanMetrics::new(&self.metrics, partition);
         metrics.num_splits.add(tasks.len());
 
         // Fill delete-file sizes as the first step of the task stream so the stats run on the
@@ -479,11 +480,11 @@ struct IcebergScanMetrics {
 }
 
 impl IcebergScanMetrics {
-    fn new(metrics: &ExecutionPlanMetricsSet) -> Self {
+    fn new(metrics: &ExecutionPlanMetricsSet, partition: usize) -> Self {
         Self {
-            baseline: BaselineMetrics::new(metrics, 0),
-            num_splits: MetricBuilder::new(metrics).counter("num_splits", 0),
-            bytes_scanned: MetricBuilder::new(metrics).counter("bytes_scanned", 0),
+            baseline: BaselineMetrics::new(metrics, partition),
+            num_splits: MetricBuilder::new(metrics).counter("num_splits", partition),
+            bytes_scanned: MetricBuilder::new(metrics).counter("bytes_scanned", partition),
         }
     }
 }
@@ -847,6 +848,88 @@ mod tests {
     fn no_ordering_keeps_single_partition() {
         let exec = exec_with_ordering(None);
         assert_eq!(exec.properties().partitioning.partition_count(), 1);
+    }
+
+    // The ordered scan reads each file as its own sorted partition and relies on
+    // SortPreservingMergeExec to k-way merge them into one globally sorted stream. This feeds known
+    // sorted partitions (with duplicate keys across partitions, and both asc and desc) into that
+    // merge with the same kind of LexOrdering the planner builds, and checks the output is globally
+    // sorted and complete. It is deterministic coverage of the merge that does not depend on an
+    // ordering-reporting Iceberg build (which is why the end-to-end suite's merge assertions cancel
+    // on the published Iceberg used in CI).
+    async fn merge_ints(input: Vec<Vec<i32>>, descending: bool) -> Vec<i32> {
+        use arrow::array::Int32Array;
+        use arrow::array::RecordBatch;
+        use arrow::compute::SortOptions;
+        use arrow::datatypes::{DataType, Field, Schema};
+        use datafusion::datasource::memory::MemorySourceConfig;
+        use datafusion::physical_expr::expressions::Column;
+        use datafusion::physical_expr::{LexOrdering, PhysicalSortExpr};
+        use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
+        use datafusion::prelude::SessionContext;
+        use futures::StreamExt;
+
+        let schema = Arc::new(Schema::new(vec![Field::new("id", DataType::Int32, false)]));
+        let partitions: Vec<Vec<RecordBatch>> = input
+            .into_iter()
+            .map(|vals| {
+                vec![RecordBatch::try_new(
+                    Arc::clone(&schema),
+                    vec![Arc::new(Int32Array::from(vals))],
+                )
+                .unwrap()]
+            })
+            .collect();
+        let source =
+            MemorySourceConfig::try_new_exec(&partitions, Arc::clone(&schema), None).unwrap();
+
+        let ordering = LexOrdering::new(vec![PhysicalSortExpr {
+            expr: Arc::new(Column::new("id", 0)),
+            options: SortOptions {
+                descending,
+                nulls_first: false,
+            },
+        }])
+        .unwrap();
+        let spm = SortPreservingMergeExec::new(ordering, source);
+        assert_eq!(spm.properties().partitioning.partition_count(), 1);
+
+        let ctx = SessionContext::new();
+        let mut stream = spm.execute(0, ctx.task_ctx()).unwrap();
+        let mut got = Vec::new();
+        while let Some(batch) = stream.next().await {
+            let batch = batch.unwrap();
+            let col = batch
+                .column(0)
+                .as_any()
+                .downcast_ref::<Int32Array>()
+                .unwrap();
+            (0..col.len()).for_each(|i| got.push(col.value(i)));
+        }
+        got
+    }
+
+    #[tokio::test]
+    async fn spm_merges_ascending_sorted_partitions() {
+        // Two individually-sorted partitions with duplicate keys across them.
+        let got = merge_ints(vec![vec![1, 3, 3, 5, 8], vec![2, 3, 6, 7]], false).await;
+        let mut expected = vec![1, 3, 3, 5, 8, 2, 3, 6, 7];
+        expected.sort_unstable();
+        assert_eq!(
+            got, expected,
+            "ascending merge must be globally sorted and complete"
+        );
+    }
+
+    #[tokio::test]
+    async fn spm_merges_descending_sorted_partitions() {
+        let got = merge_ints(vec![vec![8, 5, 3, 3, 1], vec![7, 6, 3, 2]], true).await;
+        let mut expected = vec![8, 5, 3, 3, 1, 7, 6, 3, 2];
+        expected.sort_unstable_by(|a, b| b.cmp(a));
+        assert_eq!(
+            got, expected,
+            "descending merge must be globally sorted and complete"
+        );
     }
 
     fn from_hex(s: &str) -> Vec<u8> {
--- a/native/core/src/execution/spark_config.rs
+++ b/native/core/src/execution/spark_config.rs
@@ -26,6 +26,21 @@ pub(crate) const COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED: &str =
     "spark.comet.parquet.rowFilterPushdown.enabled";
 pub(crate) const SPARK_EXECUTOR_CORES: &str = "spark.executor.cores";
 
+/// Above this many files in a single Spark partition, the native Iceberg scan does not attempt a
+/// per-file-stream k-way merge (which would open one reader per file at once); it reads the
+/// partition unordered and sorts with a spillable SortExec instead. Read from the config below and
+/// carried to the physical planner as an [`IcebergSortMergeConfig`] session extension.
+pub(crate) const COMET_ICEBERG_SORT_MERGE_MAX_FILES_PER_PARTITION: &str =
+    "spark.comet.scan.icebergNative.sortMerge.maxFilesPerPartition";
+pub(crate) const DEFAULT_ICEBERG_SORT_MERGE_MAX_FILES_PER_PARTITION: usize = 64;
+
+/// Comet planner settings carried on the DataFusion `SessionConfig` as a type-keyed extension, so
+/// the physical planner can read Comet scan configs that are not DataFusion options.
+#[derive(Debug)]
+pub(crate) struct IcebergSortMergeConfig {
+    pub max_files_per_partition: usize,
+}
+
 pub(crate) trait SparkConfig {
     fn get_bool(&self, name: &str) -> bool;
     fn get_u64(&self, name: &str, default_value: u64) -> u64;
--- a/spark/src/main/scala/org/apache/comet/CometConf.scala
+++ b/spark/src/main/scala/org/apache/comet/CometConf.scala
@@ -133,18 +133,6 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(true)
 
-  val COMET_ICEBERG_REPORT_PARTITIONING_ENABLED: ConfigEntry[Boolean] =
-    conf("spark.comet.scan.icebergNative.reportPartitioning.enabled")
-      .category(CATEGORY_SCAN)
-      .doc("Whether the native Iceberg scan reports Iceberg's key-grouped partitioning (requires " +
-        "spark.sql.sources.v2.bucketing.enabled) so storage-partitioned joins avoid a shuffle, " +
-        "which is what lets a reported sort order eliminate a join's sort. Independent of " +
-        "sortMerge.enabled. Defaults to disabled: the AQE + push-down-partition-values path is " +
-        "not yet verified, so reporting is off until it is exercised. When disabled, " +
-        "partitioning is reported as unknown, as before.")
-      .booleanConf
-      .createWithDefault(false)
-
   val COMET_ICEBERG_WRITE_SPLIT_OPERATOR_ENABLED: ConfigEntry[Boolean] =
     conf("spark.comet.write.iceberg.splitOperator.enabled")
       .category(CATEGORY_TESTING)
--- a/spark/src/main/scala/org/apache/comet/CometConf.scala
+++ b/spark/src/main/scala/org/apache/comet/CometConf.scala
@@ -133,6 +133,18 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(true)
 
+  val COMET_ICEBERG_SORT_MERGE_MAX_FILES_PER_PARTITION: ConfigEntry[Int] =
+    conf("spark.comet.scan.icebergNative.sortMerge.maxFilesPerPartition")
+      .category(CATEGORY_SCAN)
+      .doc(
+        "The maximum number of files in a single partition that the native Iceberg scan will " +
+          "k-way merge to preserve sort order. A merge opens one reader per file at once, so " +
+          "above this many files the scan instead reads the partition unordered and sorts it " +
+          "with a spillable sort, bounding concurrently-open readers. Both paths produce sorted " +
+          "output; this only trades merge for a full sort on partitions with many files.")
+      .intConf
+      .createWithDefault(64)
+
   val COMET_ICEBERG_WRITE_SPLIT_OPERATOR_ENABLED: ConfigEntry[Boolean] =
     conf("spark.comet.write.iceberg.splitOperator.enabled")
       .category(CATEGORY_TESTING)
--- a/spark/src/main/scala/org/apache/comet/CometConf.scala
+++ b/spark/src/main/scala/org/apache/comet/CometConf.scala
@@ -121,6 +121,30 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(true)
 
+  val COMET_ICEBERG_SORT_MERGE_ENABLED: ConfigEntry[Boolean] =
+    conf("spark.comet.scan.icebergNative.sortMerge.enabled")
+      .category(CATEGORY_SCAN)
+      .doc("Whether the native Iceberg scan reports the table sort order and performs a " +
+        "per-partition streaming merge of already-sorted files. When enabled and Iceberg reports " +
+        "an ordering (requires Iceberg's spark.sql.iceberg.planning.preserve-data-ordering), " +
+        "each Spark partition reads its files as separate sorted streams merged into one sorted " +
+        "output, and the ordering is surfaced to Spark so redundant sorts are eliminated. When " +
+        "disabled, files are read unordered as before.")
+      .booleanConf
+      .createWithDefault(true)
+
+  val COMET_ICEBERG_REPORT_PARTITIONING_ENABLED: ConfigEntry[Boolean] =
+    conf("spark.comet.scan.icebergNative.reportPartitioning.enabled")
+      .category(CATEGORY_SCAN)
+      .doc("Whether the native Iceberg scan reports Iceberg's key-grouped partitioning (requires " +
+        "spark.sql.sources.v2.bucketing.enabled) so storage-partitioned joins avoid a shuffle, " +
+        "which is what lets a reported sort order eliminate a join's sort. Independent of " +
+        "sortMerge.enabled. Defaults to disabled: the AQE + push-down-partition-values path is " +
+        "not yet verified, so reporting is off until it is exercised. When disabled, " +
+        "partitioning is reported as unknown, as before.")
+      .booleanConf
+      .createWithDefault(false)
+
   val COMET_ICEBERG_WRITE_SPLIT_OPERATOR_ENABLED: ConfigEntry[Boolean] =
     conf("spark.comet.write.iceberg.splitOperator.enabled")
       .category(CATEGORY_TESTING)
--- a/spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
+++ b/spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
@@ -936,6 +936,32 @@ object IcebergReflection extends Logging {
   }
 
   /**
+   * Names of top-level columns whose Iceberg sort order can differ from Spark's comparison of the
+   * Spark type they map to, so a native k-way merge keyed on the Spark value could mis-order
+   * rows. Today that is UUID: Iceberg maps UUID to Spark StringType (see TypeToSparkType) but
+   * sorts by its own UUID comparator, not by the canonical string, so the file order and a string
+   * comparison can disagree. reportableOrdering refuses a sort key in this set. v1 only reports
+   * identity, top-level sort keys, so top-level columns are enough.
+   */
+  def orderingUnsafeColumns(schema: Any): Set[String] = {
+    import scala.jdk.CollectionConverters._
+    try {
+      val columns = getMethod(schema.getClass, "columns")
+        .invoke(schema)
+        .asInstanceOf[java.util.List[_]]
+      columns.asScala.flatMap { column =>
+        val name = getMethod(column.getClass, "name").invoke(column).asInstanceOf[String]
+        val typeStr = getMethod(column.getClass, "type").invoke(column).toString
+        if (typeStr == "uuid") Some(name) else None
+      }.toSet
+    } catch {
+      case e: Exception =>
+        logWarning(s"Failed to inspect schema for ordering-unsafe columns: ${e.getMessage}")
+        Set.empty[String]
+    }
+  }
+
+  /**
    * Validates partition column types for compatibility with iceberg-rust.
    *
    * iceberg-rust's Literal::try_from_json() has incomplete type support: - Binary/fixed types:
--- a/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
@@ -835,12 +835,38 @@ case class CometScanRule(session: SparkSession)
           }
         }
 
+        // If Iceberg reports an ordering, EnsureRequirements may have already dropped the Sort
+        // above this scan (it decides that on the vanilla BatchScanExec, before Comet converts the
+        // scan). If the native scan cannot guarantee that ordering, reading unordered here would
+        // silently return wrong results, so stay on Spark -- its Iceberg reader produces the sorted
+        // output it promised. reportableOrdering is the same gate the native scan/serde use, so the
+        // decision here cannot diverge from what the native path would do.
+        val orderingHonored: Boolean = {
+          val reported = scanExec.ordering.getOrElse(Nil)
+          if (reported.isEmpty) {
+            true
+          } else {
+            val unsafe = IcebergReflection.orderingUnsafeColumns(metadata.tableSchema)
+            val honorable =
+              CometIcebergNativeScan.reportableOrdering(
+                scanExec.ordering,
+                scanExec.output,
+                unsafe)
+            if (honorable.isEmpty) {
+              fallbackReasons += "Iceberg reports a sort order the native scan cannot guarantee " +
+                "(sort-merge disabled, or a transform / unsafe-type sort key); staying on Spark " +
+                "so the reported ordering is preserved"
+            }
+            honorable.nonEmpty
+          }
+        }
+
         if (schemaSupported && fileIOCompatible && formatVersionSupported &&
           defaultValuesSupported && schemaTypesSupported && encryptionKeyLengthSupported &&
           taskValidation.allParquet && allSupportedFilesystems && partitionTypesSupported &&
           unifiedPartitionTypeSupported &&
           complexTypePredicatesSupported && transformFunctionsSupported &&
-          deleteFileTypesSupported && dppSubqueriesSupported) {
+          deleteFileTypesSupported && dppSubqueriesSupported && orderingHonored) {
           CometBatchScanExec(
             scanExec.clone().asInstanceOf[BatchScanExec],
             runtimeFilters = scanExec.runtimeFilters,
--- a/spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
@@ -877,15 +877,24 @@ object CometIcebergNativeScan extends CometOperatorSerde[CometBatchScanExec] wit
    *
    * We read scanExec.ordering (the raw reported order), not scanExec.outputOrdering. Spark blanks
    * outputOrdering when a partition holds more than one file -- the case this merge handles.
+   *
+   * `unsafeColumns` are top-level columns whose Iceberg sort order can differ from Spark's
+   * comparison of the mapped Spark type (today: UUID, which Iceberg maps to StringType but sorts
+   * by its own comparator -- see IcebergReflection.orderingUnsafeColumns). A sort key on such a
+   * column is refused, because the files are sorted by an order the native string merge would not
+   * reproduce. This cannot be detected from the Spark type alone (UUID looks like a plain
+   * string), so the caller passes the names in.
    */
   def reportableOrdering(
       ordering: Option[Seq[SortOrder]],
-      output: Seq[Attribute]): Seq[SortOrder] = {
+      output: Seq[Attribute],
+      unsafeColumns: Set[String] = Set.empty): Seq[SortOrder] = {
     if (!CometConf.COMET_ICEBERG_SORT_MERGE_ENABLED.get()) {
       Nil
     } else {
       ordering match {
-        case Some(orders) if orders.nonEmpty && orders.forall(isReportable(_, output)) =>
+        case Some(orders)
+            if orders.nonEmpty && orders.forall(isReportable(_, output, unsafeColumns)) =>
           orders
         case _ =>
           Nil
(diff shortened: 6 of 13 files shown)


parthchandra:
This PR has some followups - 

 -  **Sort keys not in the projection.** Today, if a sort column isn't selected, we skip the merge and read unordered. Todo: read the sort columns anyway (even when not selected), merge on them, then drop them before returning.
  - **Transform sort orders (bucket/truncate/etc.).** We only handle plain-column sort orders for now. todo: also handle transformed sort keys, matching Iceberg's per-type comparison exactly.
  - **Read concurrency.** The merge reads files more lazily than today's unordered read. As a followup benchmark, and if many-small-file reads regress, read files ahead into small buffers.
  - **Single-file partitions.** A partition with one file still gets wrapped in the merge (a no-op). Skip the wrapper in that case.
  - **Storage-partitioned join confidence tests, then default `reportPartitioning` on.** Add tests for: (a) a heavily skewed key so the replicate path runs, (b) different bucket counts per side (e.g. bucket(8) vs bucket(4)) so a partition reducer kicks in, (c) three-way and self joins on the same key. Once green, flip the flag on by default.

parthchandra:
@anuragmantri, @peter-toth , you might be interested in looking at this. k-way merge to maintain the ordered property of  sorted Iceberg tables. Feedback, especially about test coverage, would be highly appreciated.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
```diff
@@ -859,6 +859,49 @@ object CometIcebergNativeScan extends CometOperatorSerde[CometBatchScanExec] wit
...
+          orders
+        case _ =>
+          Nil
+      }
+    }
+  }
+
+  private def isReportable(order: SortOrder, output: Seq[Attribute]): Boolean =
+    isIdentityProjected(order, output) && exprToProto(order, output).isDefined
+
```
anuragmantri: As identified in the [Iceberg PR](https://github.com/apache/iceberg/pull/14948#pullrequestreview-4932308014) and the design doc https://github.com/apache/datafusion-comet/issues/5323, UUID orders differently in Iceberg than in Spark's comparator, and the identity case for UUID needs to follow Iceberg's byte ordering specifically. This gate doesn't check the sort column's type at all. I believe we could rely on upstream https://github.com/apache/iceberg/pull/16750 to not report ordering on UUID. Is my understanding correct?

parthchandra:
Added more followup issues in parent issue - https://github.com/apache/datafusion-comet/issues/5323

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
```diff
@@ -859,6 +859,49 @@ object CometIcebergNativeScan extends CometOperatorSerde[CometBatchScanExec] wit
...
+   * SortOrder child must be an AttributeReference in `output`, and must serialize to proto.
+   * Transform sort fields (bucket/truncate/...) are not AttributeReferences, so they fall through
+   * to Nil and we read unordered. Checking exprToProto here, not just in the proto path, keeps
+   * the two callers in step: outputOrdering never advertises an order the proto path would drop.
+   *
+   * We trust Iceberg on file-level sortedness. If it reports an ordering, SortOrderAnalyzer has
+   * already checked each file's sort_order_id matches the table order, so every file is sorted.
+   *
+   * We read scanExec.ordering (the raw reported order), not scanExec.outputOrdering. Spark blanks
+   * outputOrdering when a partition holds more than one file -- the case this merge handles.
```
anuragmantri, line 862: I believe using `scanExec.ordering` is to bypass the check in Spark 3.4 and 3.5 which was fixed in Spark 4.2 by [SPARK-55715](https://github.com/apache/spark/pull/55116) and this is needed to support all these Spark versions. @peter-toth, would you please also take a look at this to see if this is safe?

native/core/src/execution/operators/iceberg_scan.rs
anuragmantri: Claude flagged this: 

`BaselineMetrics::new(metrics, 0)` hardcodes partition 0, but this operator is now multi-partition in the ordered path. Should this thread through the actual partition index from `execute()`?

spark/src/test/scala/org/apache/comet/CometIcebergSortMergeReadSuite.scala
```diff
@@ -0,0 +1,638 @@
...
+ *
+ * These cannot be SQL-file fixtures: setting an Iceberg sort order needs the Iceberg Java API
+ * (the Comet test session registers no Iceberg SQL extensions, so `WRITE ORDERED BY` will not
+ * parse), and the plan-shape assertions need access to the executed SparkPlan.
+ *
+ * Each test gets its own catalog name and temp warehouse (the Hadoop `SparkCatalog` instance is
+ * cached per catalog name, so a shared name would bind every test to the first warehouse), and
+ * its tables are dropped in a `finally` so a failing test cannot leak a table into a later one.
+ */
+class CometIcebergSortMergeReadSuite
```
anuragmantri, line 80: I reviewed the tests, very nice coverage already. I would also a test that deletes some rows from one file in a multi-file sorted partition, to cover the merge alongside MOR deletes.

anuragmantri (commented):
Thanks for pinging here @parthchandra. It is great to see there is interest in this optimization in the comet project. 

I reviewed the general implementation and the test coverage in this PR. I don't have any major comments, most are clarifications and minor suggestions. That said, I'm new to Comet codebase and would like someone more familiar to also take a look.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
peter-toth: Seems good to me, `scanExec.ordering` is the order reported by the source and that's what we need to keep with k-way merge if possible, and return as `scanExec.outputOrdering` so as Spark can use it to elimiate sorts.

peter-toth: So as we discussed offline `CometIcebergNativeScanExec.scala.outputOrdering` and this whole logic might not be needed if `BatchScanExec` -> `CometIcebergNativeScanExec` translation happend after `EnsureRequirements` runs (i.e. we eliminated the `SortExec`s based on `BatchScanExec.outputOrdering`).

And same for `CometIcebergNativeScanExec.outputPartitioning`...

andygrove:
One thing I'd like to dig into more before this lands: the reader fan-out in the ordered path.

`SortPreservingMergeExec` calls `execute()` on all of its input partitions up front, and `IcebergScanExec::execute` eagerly builds the `ArrowReader` and kicks off the read, so we end up with a live Parquet reader per data file in the Spark partition plus a buffered batch per stream inside the merge. `data_file_concurrency_limit` doesn't bound any of that anymore.

The comment in `iceberg_scan.rs` describes the fan-out as intrinsic to a k-way merge. That's true of the number of *streams*, but not of how many readers we have decoding concurrently or how much we buffer — a cascaded merge, or just falling back to the unordered read when a partition has too many files, would bound both.

The tables in the tests are fine, but the workload this targets is a sorted table, and a sorted table that has accumulated a lot of small commits is exactly where you get hundreds of files in one task. Since the merge reserves against Comet's memory pool, hitting the limit there is a query failure rather than a slowdown, which is a worse failure mode than the extra sort we're trying to avoid.

Could we add a config for the maximum files per partition we're willing to merge, and fall back to the unordered read above it? That keeps the default safe and lets people opt into deeper merges once we have a better sense of the memory profile.

andygrove:
Following up on test coverage, because I want to make sure I'm reading this right.

As far as I can tell `SupportsReportOrdering` is what populates `BatchScanExec.ordering`, and that field is the only trigger for everything in this PR. Iceberg's implementation of it is apache/iceberg#16750, which is still open and targeted at Spark 4.1, so on the published Iceberg versions in our poms (1.5.2, 1.8.1, 1.10.0, 1.11.0) `ordering` is always `None`. That means `reportableOrdering` returns `Nil`, `table_sort_orders` goes over empty, native takes `ordering = None`, and we get the old single-partition unordered scan with no `SortPreservingMergeExec`.

If that's right, the effect in CI is bigger than the canceled plan assertions. The k-way merge itself never runs — every `checkSparkAnswer` in the new suite is exercising the read path we already had, so the only CI coverage of the new native code is the two Rust tests asserting `partition_count()`.

There's a related wrinkle even on a reporting build: a global `ORDER BY` keeps its final sort, which would repair any mis-ordering the merge produced, so those tests catch dropped or duplicated rows but not wrong order. The order-sensitive assertions are all in the `assume`-gated SMJ/aggregate/window tests.

Am I understanding the situation correctly, or is there something in the CI setup I'm missing that does exercise the merge?

andygrove:
One more, on the gate in `serializePartitions`:

```scala
val reportable = reportableOrdering(scanExec.ordering, output)
if (reportable.nonEmpty) {
  val protoOrders = reportable.map(exprToProto(_, output))
  if (protoOrders.forall(_.isDefined)) {
    commonBuilder.addAllTableSortOrders(protoOrders.map(_.get).asJava)
  }
}
```

The comment above it argues the `forall` always holds, and I agree that's true today. My worry is what happens if it ever stops being true, because the guard makes the unsafe direction the silent one. If we merge without telling Spark, we've wasted some work and nothing else. If we tell Spark and then don't merge, Spark drops the Sort and we return wrong results with no error.

That's not purely hypothetical, because the gate gets evaluated twice against two different points in time. `outputOrdering` is a lazy val forced during planning, this runs later off `serializedPartitionData`, and `reportableOrdering` reads `COMET_ICEBERG_SORT_MERGE_ENABLED` out of the active `SQLConf` each time. Two dynamic conf reads on separate threads at separate stages is the kind of thing that holds until it doesn't.

`serializePartitions` only has the one caller, so could we just pass `outputOrdering` in from the exec and have a single evaluation? Failing that I'd rather the mismatch threw than silently produced a proto that disagrees with what we told Spark.

native/core/src/execution/operators/iceberg_scan.rs
parthchandra: Good catch Claude! Updated this.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
parthchandra: @anuragmantri  Correct. `DataSourceV2ScanExecBase.outputOrdering` blanks the ordering when a partition has more than one file on Spark 3.4–4.1 (relaxed in 4.2 by SPARK-55715). We need to keep this as long as we support versions before 4.2.

parthchandra: > So as we discussed offline `CometIcebergNativeScanExec.scala.outputOrdering` and this whole logic might not be needed if `BatchScanExec` -> `CometIcebergNativeScanExec` translation happend after `EnsureRequirements` runs (i.e. we eliminated the `SortExec`s based on `BatchScanExec.outputOrdering`).
> 
> And same for `CometIcebergNativeScanExec.outputPartitioning`...

@peter-toth  I dug in deeper and you are correct. `EnsureRequirements` does run before Comet converts the scan (confirmed in `AdaptiveSparkPlanExec.queryStagePreparationRules` and the non-AQE order), so it eliminates the shuffle on the `BatchScanExec`. An experiment confirmed it: with partitioning reporting off, the SMJ still has no `Exchange`. So I removed the `outputPartitioning` override and the `reportPartitioning` flag entirely.
`outputOrdering` is still needed because pre 4.2 Spark still has the check that blanks out  `BatchScanExec.outputOrdering` for multi-file  partitions. So we need to include the check to eliminate sort.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
parthchandra: I was planning to address this in a follow up. But now I've added a type check rather than relying only on upstream. `IcebergReflection.orderingUnsafeColumns` returns UUID column  names, and `reportableOrdering` now refuses any sort key in that set

spark/src/test/scala/org/apache/comet/CometIcebergSortMergeReadSuite.scala
parthchandra: added new test for this case

parthchandra:
> 
> That's not purely hypothetical, because the gate gets evaluated twice against two different points in time. `outputOrdering` is a lazy val forced during planning, this runs later off `serializedPartitionData`, and `reportableOrdering` reads `COMET_ICEBERG_SORT_MERGE_ENABLED` out of the active `SQLConf` each time. Two dynamic conf reads on separate threads at separate stages is the kind of thing that holds until it doesn't.
> 

Fixed. This is evaluated once now. 

> `serializePartitions` only has the one caller, so could we just pass `outputOrdering` in from the exec and have a single evaluation? Failing that I'd rather the mismatch threw than silently produced a proto that disagrees with what we told Spark.

Fixed. This throws if a reported order can't be serialized, instead of the old silent `if (forall(_.isDefined))` drop

parthchandra:
> Following up on test coverage, because I want to make sure I'm reading this right.
> 
> As far as I can tell `SupportsReportOrdering` is what populates `BatchScanExec.ordering`, and that field is the only trigger for everything in this PR. Iceberg's implementation of it is [apache/iceberg#16750](https://github.com/apache/iceberg/pull/16750), which is still open and targeted at Spark 4.1, so on the published Iceberg versions in our poms (1.5.2, 1.8.1, 1.10.0, 1.11.0) `ordering` is always `None`. That means `reportableOrdering` returns `Nil`, `table_sort_orders` goes over empty, native takes `ordering = None`, and we get the old single-partition unordered scan with no `SortPreservingMergeExec`.
> 
> If that's right, the effect in CI is bigger than the canceled plan assertions. The k-way merge itself never runs — every `checkSparkAnswer` in the new suite is exercising the read path we already had, so the only CI coverage of the new native code is the two Rust tests asserting `partition_count()`.
> 
> There's a related wrinkle even on a reporting build: a global `ORDER BY` keeps its final sort, which would repair any mis-ordering the merge produced, so those tests catch dropped or duplicated rows but not wrong order. The order-sensitive assertions are all in the `assume`-gated SMJ/aggregate/window tests.
> 
> Am I understanding the situation correctly, or is there something in the CI setup I'm missing that does exercise the merge?

You're reading this right. I cannot think of a way to mock Iceberg without the actual implementation (which is still in review). The tests cover the fallback, and are somewhat forward looking.  In a way this feature is itself forward looking - once this feature is released in Iceberg and Spark starts to eliminate sorts based on what iceberg-java reports, Comet will produce absolutely garbage results unless we have this implementation in place.

parthchandra:
> One thing I'd like to dig into more before this lands: the reader fan-out in the ordered path.
> 
> `SortPreservingMergeExec` calls `execute()` on all of its input partitions up front, and `IcebergScanExec::execute` eagerly builds the `ArrowReader` and kicks off the read, so we end up with a live Parquet reader per data file in the Spark partition plus a buffered batch per stream inside the merge. `data_file_concurrency_limit` doesn't bound any of that anymore.
> 
> The comment in `iceberg_scan.rs` describes the fan-out as intrinsic to a k-way merge. That's true of the number of _streams_, but not of how many readers we have decoding concurrently or how much we buffer — a cascaded merge, or just falling back to the unordered read when a partition has too many files, would bound both.
> 
> The tables in the tests are fine, but the workload this targets is a sorted table, and a sorted table that has accumulated a lot of small commits is exactly where you get hundreds of files in one task. Since the merge reserves against Comet's memory pool, hitting the limit there is a query failure rather than a slowdown, which is a worse failure mode than the extra sort we're trying to avoid.
> 
> Could we add a config for the maximum files per partition we're willing to merge, and fall back to the unordered read above it? That keeps the default safe and lets people opt into deeper merges once we have a better sense of the memory profile.

I was planning to address this properly in a followup (https://github.com/apache/datafusion-comet/issues/5343). But I can implement that in this PR if you think I should address it right away.

spark/src/test/scala/org/apache/comet/CometIcebergSortMergeReadSuite.scala
```diff
@@ -0,0 +1,641 @@
...
+    withSortedTables(orderedReadConf)("t") { cat =>
+      spark.sql(s"CREATE TABLE $cat.db.t (id INT, data STRING) USING iceberg")
+      replaceSortOrder(cat, "db", "t", "id" -> true)
+      insertBatches(cat, "t", "(1,'a'),(2,'b')")
+
+      checkSparkAnswer(s"SELECT id, data FROM $cat.db.t ORDER BY id")
+    }
+  }
+
+  test("partitioned table with several files per partition") {
```
mbutrovich, line 511: Every test in this suite inserts 2 to 4 files per partition. Given the memory concern in `iceberg_scan.rs` is specifically about a partition with a large number of files (a live Parquet reader plus a buffered batch per file, all opened at once), would it be worth adding one test that inserts something like 50 to 100 single-row files into one partition? It would give the merge path a correctness check under the actual shape of the workload this feature targets, not just its happy-path shape, and would be a natural place to assert on memory pool usage or peak concurrent readers once #5343 lands.

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -86,6 +86,24 @@ pub struct IcebergScanExec {
...
+    file_io: FileIO,
+    /// Table sort order Iceberg reported, translated against `output_schema`. `Some` makes this a
+    /// multi-partition scan: one sorted stream per task, which a SortPreservingMergeExec above
+    /// merges back into one sorted partition. It is also advertised in `plan_properties`. `None`
+    /// keeps the old single-partition unordered read (all tasks streamed together).
+    ///
+    /// Concurrency note: in the ordered path each partition reads exactly one task, so
+    /// `data_file_concurrency_limit` no longer bounds cross-file concurrency; instead the wrapping
+    /// SortPreservingMergeExec drives one reader per file to merge them. That fan-out (files per
+    /// Spark partition) is intrinsic to a k-way merge of per-file sorted streams -- the files must
```
mbutrovich: Following up on the fan-out discussion: I don't think this needs a wholly new algorithm, but it does need more than `IcebergScanExec` alone, since the operator that decides when to poll each partition is `SortPreservingMergeExec`, not this one. That operator has no notion of a bound and primes every input stream up front because polling is the only signal it has for "what's in this stream." Making `IcebergScanExec::execute` lazy internally doesn't avoid that, since `SortPreservingMergeExec` still calls `poll_next` on every partition almost immediately to seed its loser tree.

What would actually bound this is a custom merge operator that takes each file's min/max bound on the sort key alongside its `FileScanTask`, keeps a small active set merged the way `SortPreservingMergeExec` does today, and holds a priority queue of unopened files ordered by min bound, only calling `execute`/`poll_next` on the next pending file once its min bound could produce the next output value. That bounds concurrently-open readers by the overlap width of the file ranges at a given point in the merge rather than by total file count, which is the number that actually blows up on a sorted table with a lot of small commits.

The blocker today is data, not algorithm: `FileScanTask` in iceberg-rust doesn't carry column-level bounds. They exist upstream on the manifest-entry `DataFile` (`lower_bounds`/`upper_bounds`, already used for predicate pushdown), just not threaded down into the task struct the native scan gets. So this needs plumbing in iceberg-rust or an extra field fetched at scan-planning time on the JVM side, either way before a bound-driven admission scheme can work.

Given the risk, could we land @andygrove's simpler mitigation in this PR, a config for the max files to merge per partition that falls back to the unordered read above it, and keep the bound-driven design above as the concrete plan for #5343? Sort-merge ships on by default, and the workload it targets (a sorted table with accumulated small commits) is exactly where a partition ends up with hundreds of files, so I'd rather the default be safe now and the deeper optimization follow once the stats plumbing exists.

mbutrovich (commented):
First pass, looks to be in pretty good shape.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
parthchandra: Made more one change. Added a check to fall back to spark if iceberg reports ordering and native cannot support it (without that we will silently get wrong results because Spark would have removed the sort already).

native/core/src/execution/operators/iceberg_scan.rs
parthchandra: Agreed, and added the fix recommended by @andygrove to limit max files per partition with a change. We cannot fallback to unordered because by this time Spark planning would have removed the downstream sort and not producing ordered results will produce wrong results. So now, beyond the threshold, we will fall back to a spillable `SortExec` so we always produce correct results

parthchandra:
@andygrove verified this with a local build of iceberg patched with support for `SupportsReportOrdering` and all the tests added in this PR pass.  One test is skipped/canceled because it operates on an unpartitioned table and Iceberg reports (correctly) that the data may not be ordered.
Also added tests to check we fallback correctly to Spark for cases where we do not support this.
cc @anuragmantri - fyi.

spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
```diff
@@ -935,6 +935,32 @@ object IcebergReflection extends Logging {
...
+        .invoke(schema)
+        .asInstanceOf[java.util.List[_]]
+      columns.asScala.flatMap { column =>
+        val name = getMethod(column.getClass, "name").invoke(column).asInstanceOf[String]
+        val typeStr = getMethod(column.getClass, "type").invoke(column).toString
+        if (typeStr == "uuid") Some(name) else None
+      }.toSet
+    } catch {
+      case e: Exception =>
+        logWarning(s"Failed to inspect schema for ordering-unsafe columns: ${e.getMessage}")
```
andygrove: This returns `Set.empty` when the reflection fails, which the callers read as "no unsafe columns", so they go on to report the ordering. That's the fail-open direction on a decision where being wrong means silently mis-ordered rows, because Spark has already dropped the `Sort` by the time we get here.

The premise of this method is that a UUID sort key is invisible at the Spark type level. If we couldn't read the schema, we can't rule one out. Could the failure path make the ordering unreportable instead? Returning something like `None` and having `reportableOrdering` treat that as "refuse" would keep the safe direction the default.

The `nativeIcebergScanMetadata == null` branch in `CometIcebergNativeScanExec.outputOrdering` has the same shape.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -835,12 +835,38 @@ case class CometScanRule(session: SparkSession)
           }
         }
 
+        // If Iceberg reports an ordering, EnsureRequirements may have already dropped the Sort
+        // above this scan (it decides that on the vanilla BatchScanExec, before Comet converts the
+        // scan). If the native scan cannot guarantee that ordering, reading unordered here would
+        // silently return wrong results, so stay on Spark -- its Iceberg reader produces the sorted
+        // output it promised. reportableOrdering is the same gate the native scan/serde use, so the
+        // decision here cannot diverge from what the native path would do.
+        val orderingHonored: Boolean = {
```
andygrove: Thanks for collapsing the proto and `outputOrdering` down to one evaluation. I think this guard reintroduces the same hazard one level up, though.

This is a second evaluation of `reportableOrdering`, and it's the one that decides whether Comet converts the scan at all. It reads `COMET_ICEBERG_SORT_MERGE_ENABLED` out of the active `SQLConf` and runs `orderingUnsafeColumns` reflection independently of the exec's lazy val. If this one says honorable, Comet converts and Spark drops the `Sort`. If the later one returns `Nil`, native reads unordered and we return wrong results with no error. The reverse mismatch just wastes some work.

You already have `honorable` right here. Could we stash it on `nativeIcebergScanMetadata` and have `CometIcebergNativeScanExec.outputOrdering` read that instead of re-running the gate? Then there's one evaluation for the whole decision, and no conf read after the point where Spark has committed to dropping the sort.

spark/src/main/scala/org/apache/comet/CometConf.scala
```diff
@@ -121,6 +121,30 @@ object CometConf extends ShimCometConf {
...
 
+  val COMET_ICEBERG_SORT_MERGE_ENABLED: ConfigEntry[Boolean] =
+    conf("spark.comet.scan.icebergNative.sortMerge.enabled")
+      .category(CATEGORY_SCAN)
+      .doc("Whether the native Iceberg scan reports the table sort order and performs a " +
+        "per-partition streaming merge of already-sorted files. When enabled and Iceberg reports " +
+        "an ordering (requires Iceberg's spark.sql.iceberg.planning.preserve-data-ordering), " +
+        "each Spark partition reads its files as separate sorted streams merged into one sorted " +
+        "output, and the ordering is surfaced to Spark so redundant sorts are eliminated. When " +
+        "disabled, files are read unordered as before.")
```
andygrove: The doc says "When disabled, files are read unordered as before", but with the new `orderingHonored` guard that isn't what happens. Disabling this makes `reportableOrdering` return `Nil`, which makes `orderingHonored` false, which keeps the scan on Spark entirely. Your own two `fallback: sort-merge disabled ...` tests assert exactly that. So the escape hatch for a merge bug also costs all Comet acceleration on every sorted Iceberg table, which is a steep price for the knob people would reach for when something looks wrong.

I don't think it has to work that way. `maxFilesPerPartition = 0` already gives us "report the ordering, never merge, honor it with the native `SortExec`", which is correct and keeps Comet on the scan. Could `sortMerge.enabled=false` mean that instead? Either way the doc needs to describe the actual behavior.

native/core/src/execution/spark_config.rs
```diff
@@ -26,6 +26,21 @@ pub(crate) const COMET_PARQUET_ROW_FILTER_PUSHDOWN_ENABLED: &str =
     "spark.comet.parquet.rowFilterPushdown.enabled";
 pub(crate) const SPARK_EXECUTOR_CORES: &str = "spark.executor.cores";
 
+/// Above this many files in a single Spark partition, the native Iceberg scan does not attempt a
+/// per-file-stream k-way merge (which would open one reader per file at once); it reads the
+/// partition unordered and sorts with a spillable SortExec instead. Read from the config below and
+/// carried to the physical planner as an [`IcebergSortMergeConfig`] session extension.
+pub(crate) const COMET_ICEBERG_SORT_MERGE_MAX_FILES_PER_PARTITION: &str =
+    "spark.comet.scan.icebergNative.sortMerge.maxFilesPerPartition";
+pub(crate) const DEFAULT_ICEBERG_SORT_MERGE_MAX_FILES_PER_PARTITION: usize = 64;
```
andygrove: `data_file_concurrency_limit` is the same kind of setting and it rides on `IcebergScanCommon`, set from `CometConf` in the serde, so there's exactly one default. This one goes through the Spark config map plus a `SessionConfig` extension with `64` written on both sides of JNI, and nothing catches it if the two drift.

Is there a reason not to put `max_files_per_partition` on `IcebergScanCommon` next to `data_file_concurrency_limit`? That removes the duplicated default and the extension plumbing in one go.

Related: both `iceberg_scan_merges_when_files_within_limit` and `iceberg_scan_falls_back_to_sort_above_file_limit` use `PhysicalPlanner::default()`, so they only ever test the compiled-in constant. If the extension weren't reaching the planner both would still pass, and the Scala test that sets the conf to `1000` would quietly exercise the sort fallback instead of the 70-way merge it's written to cover. Worth one test that builds the planner from a `SessionConfig` carrying `IcebergSortMergeConfig { max_files_per_partition: 2 }` and asserts a 3-file scan takes the `SortExec` path.

andygrove (commented):
Thanks for working through the earlier rounds. The fan-out cap, the single-evaluation of the proto gate, the `require` instead of the silent drop, and dropping the `outputPartitioning` override after peter-toth's point all landed well, and the new tests for MOR deletes, the 70-file merge, the 70-file sort fallback, and the three Spark-fallback cases close most of what I was worried about on coverage. I also liked that you pushed back on my "fall back to the unordered read" phrasing. You're right that by that point Spark has dropped the sort, so it has to be a native sort.

A few things I checked and am happy with, so nobody else has to redo them. The fallback `SortExec` uses the same `SortExec::new` plus `create_sort_expr` as `OpStruct::Sort`, so it sorts identically to the sort it replaces. `DiskManagerMode::Directories` is configured, so that sort can spill rather than fail. Moving the metrics off partition 0 is safe because `update_comet_metric` uses `aggregate_by_name()`. And on types beyond UUID, I worked through the other Iceberg-to-Spark mappings: `binary` and `fixed` compare unsigned on both sides, and `float`/`double` are safe in the direction that matters, since Iceberg's total order refines Spark's `nanSafeCompare` where `-0.0 == 0.0`, and a refinement of a weaker order is still non-decreasing under it. UUID does look like the only real hazard today.

I have six things I'd like addressed. Five are inline. The first two are correctness, and the rest are about keeping this maintainable.

The sixth has no diff line to attach to: `docs/source/user-guide/latest/iceberg.md`. The "Current limitations" list is where people look to find out why a scan fell back, and this PR adds three new entries to it: a transform sort key, a UUID sort key, and sort-merge disabled while Iceberg still reports an ordering. Could we add those? The `### Tuning` section above already documents `dataFileConcurrencyLimit` and looks like the right home for `sortMerge.enabled` and `sortMerge.maxFilesPerPartition`, including the note that this only does anything with Iceberg's `spark.sql.iceberg.planning.preserve-data-ordering` turned on. `configs.md` is generated so that one is fine.

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -86,6 +86,24 @@ pub struct IcebergScanExec {
...
+    /// be read as separate streams to stay individually sorted -- and is the natural granularity
+    /// for a sorted Iceberg table. `data_file_concurrency_limit` still bounds delete-file stats and
```
andygrove: This is the paragraph @mbutrovich and I were arguing with, and the code has moved on. It still says the fan-out "is intrinsic to a k-way merge" and "is the natural granularity for a sorted Iceberg table", with no mention that the planner now caps it at `sortMerge.maxFilesPerPartition` and drops to a spillable `SortExec` above that. Could you point it at the cap, and at #5343 for the bound-driven admission scheme? Otherwise the next person to read this concludes the fan-out is unbounded by design.

spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
parthchandra: Agreed. `orderingUnsafeColumns` now returns `None` when the schema can't be read, and the caller treats `None` as "refuse the ordering" — a reflection failure makes it un-reportable instead of assuming no UUID. The `nativeIcebergScanMetadata == null `branch is gone too; see the single-evaluation change below.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
parthchandra: Good point.  Changes so we evaluate the gate once and save the result on `nativeIcebergScanMetadata`; `outputOrdering` and the proto serde both read that field.

spark/src/main/scala/org/apache/comet/CometConf.scala
parthchandra: You're right, doing what you suggest: `enabled=false` now behaves like `maxFilesPerPartition=0` — the scan stays native and still honors the reported order via the spillable `SortExec`, it just skips the k-way merge. The two disabled-fallback tests are now "stays native" tests.

native/core/src/execution/spark_config.rs
parthchandra: Moved `max_files_per_partition` onto `IcebergScanCommon` next to `data_file_concurrency_limit`, so there's one default (in CometConf) and removed the SessionConfig extension. Added `iceberg_scan_honors_max_files_per_partition_from_proto`, which sets the proto field to 2 and asserts a 3-file scan takes SortExec — so the value is actually read from the plan  now, not the compiled-in constant

native/core/src/execution/operators/iceberg_scan.rs
parthchandra: Comment updated

parthchandra:
> `ocs/source/user-guide/latest/iceberg.md`. The "Current limitations" list is where people look to find out why a scan fell back, and this PR adds three new entries to it: a transform sort key, a UUID sort key, and sort-merge disabled while Iceberg still reports an ordering. Could we add those? The `### Tuning` section above already documents `dataFileConcurrencyLimit` and looks like the right home for `sortMerge.enabled` and `sortMerge.maxFilesPerPartition`, including the note that this only does anything with Iceberg's `spark.sql.iceberg.planning.preserve-data-ordering` turned on. `configs.md` is generated so that one is fine.

Added the transform-sort-key and UUID-sort-key fallbacks to "Current limitations", and documented `sortMerge.enabled + sortMerge.maxFilesPerPartition` under Tuning, including the `preserve-data-ordering `pre-requisite

andygrove:
A few more from another pass over the current head:

On the `require(protoOrders.forall(_.isDefined), ...)` in `serializePartitions` — thanks for closing the silent-drop hole, but the failure mode is now an `IllegalArgumentException` at every executor task rather than a planning-time fallback to Spark. If the `output` binding ever drifts between `CometScanRule.reportableOrdering` (where the gate ran and stashed `metadata.reportedOrdering`) and this call — an aliasing pass, an exprId rewrite, something along those lines — a query that would previously have executed on Spark hard-fails at exec startup. Could this raise an unchecked exception higher up so the whole scan falls back cleanly, rather than `require` at task start?

Also, `COMET_ICEBERG_SORT_MERGE_MAX_FILES_PER_PARTITION` has no `checkValue` bound, and the proto field is `uint32`. A user passing `-1` (or any negative int) casts to `4294967295` on the wire and the cap is effectively gone — a partition with tens of thousands of files opens all of them concurrently, which is exactly what the cap was added to prevent. The sibling `COMET_ICEBERG_DATA_FILE_CONCURRENCY_LIMIT` has `.checkValue(v => v > 0, ...)`; want to add the same here?

Small one: the doc string for `COMET_ICEBERG_SORT_MERGE_ENABLED` reads "When disabled, the scan native and the ordering is still reported and honoured" — missing a verb after "scan". That description gets surfaced in config listings and generated docs.

parthchandra:
Thanks Andy. Addressed all three - 

> 
> On the `require(protoOrders.forall(_.isDefined), ...)` in `serializePartitions` — thanks for closing the silent-drop hole, but the failure mode is now an `IllegalArgumentException` at every executor task rather than a planning-time fallback to Spark. If the `output` binding ever drifts between `CometScanRule.reportableOrdering` (where the gate ran and stashed `metadata.reportedOrdering`) and this call — an aliasing pass, an exprId rewrite, something along those lines — a query that would previously have executed on Spark hard-fails at exec startup. Could this raise an unchecked exception higher up so the whole scan falls back cleanly, rather than `require` at task start?


good point — moved the binding to planning time. CometScanRule now binds the reported ordering to proto against the scan's output up front (serializeReportedOrdering), and if that fails it's just another reason orderingHonored is false, so the scan falls back to Spark during planning instead of hard-failing at exec startup. The serde now writes the already-bound protos, so the require is gone.


> 
> Also, `COMET_ICEBERG_SORT_MERGE_MAX_FILES_PER_PARTITION` has no `checkValue` bound, and the proto field is `uint32`. A user passing `-1` (or any negative int) casts to `4294967295` on the wire and the cap is effectively gone — a partition with tens of thousands of files opens all of them concurrently, which is exactly what the cap was added to prevent. The sibling `COMET_ICEBERG_DATA_FILE_CONCURRENCY_LIMIT` has `.checkValue(v => v > 0, ...)`; want to add the same here?

added .checkValue(v => v >= 0, ...), matching the sibling. 0 stays meaningful ("never merge, honor via sort"); negatives are now rejected instead of wrapping to 4294967295 and disabling the cap.  

> 
> Small one: the doc string for `COMET_ICEBERG_SORT_MERGE_ENABLED` reads "When disabled, the scan native and the ordering is still reported and honoured" — missing a verb after "scan". That description gets surfaced in config listings and generated docs.

fixed — "the scan stays native"

andygrove (commented):
A few things from a pass over the current head. I ran the new suite locally against the published Iceberg (26 passed, 6 canceled, 0 failed) and spent most of the time on the native side, since that is where CI cannot reach today.

Three are inline. The fourth has no diff line to attach to: the PR description still describes the key-grouped partitioning reporting and `spark.comet.scan.icebergNative.reportPartitioning.enabled`, both of which went away when the `outputPartitioning` override was dropped, and it does not mention `sortMerge.maxFilesPerPartition`. Since this becomes the squash commit message, could you refresh it? The "Summary of changes" bullets are otherwise still accurate.

One thing I want to record as a positive, so nobody re-derives it: I probed the null ordering round-tripping through the proto in all four direction and null-ordering combinations, on both the merge path and the `maxFilesPerPartition` sort-fallback path, and it is correct in every one. Multi-column and descending orderings are fine too.

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -164,18 +224,14 @@ impl IcebergScanExec {
...
         context: Arc<TaskContext>,
     ) -> DFResult<SendableRecordBatchStream> {
         let output_schema = Arc::clone(&self.output_schema);
-        let file_io = load_file_io(
-            &self.catalog_properties,
-            &self.metadata_location,
-            &self.catalog_name,
-            AccessMode::Read,
-        )?;
+        let file_io = self.file_io.clone();
```
andygrove, line 245: `execute_with_tasks` builds a new `ArrowReaderBuilder(...).build()` on every call, a few lines below this one, and that is where iceberg-rust constructs the `CachingDeleteFileLoader`. Its `delete_filter` is the shared state the crate documents as existing "to allow caching loaded deletes across multiple calls to `load_deletes` (e.g., across multiple file scan tasks)". In the ordered path `execute` runs once per file, so that cache is now per-file, and a delete file shared across the partition gets downloaded and parsed once per data file instead of once for the whole partition. `fill_delete_file_sizes` dedups within a single call too, so it also issues one HEAD per data file for the same delete file now. At the default cap that is up to 64x on a merge-on-read table, and equality deletes are the worst case since they are always partition-scoped and the expensive ones to parse.

I measured it with two data files sharing one positional delete file. `bytes_scanned` goes from 2717 on the unordered path to 4256 on the ordered path, and the 1539-byte delta is exactly the delete file being read a second time. Rows are correct either way, so this is purely IO and CPU rather than a correctness problem.

The fix looks like the same shape as the `file_io` hoist on this line: hold one `ArrowReader` on `IcebergScanExec` and clone it per partition. It needs `batch_size` off the `TaskContext`, so it has to be a `OnceLock` filled on first `execute` rather than built in `new`, but `ArrowReader` is `Clone`, and `read()` calls `ScanMetrics::new()` per invocation and re-bases the loader's metrics through `with_scan_metrics`, so per-partition metrics stay separate while the delete cache is shared. With that prototype the ordered path drops back to 2717 with identical rows. Would you rather do it here or fold it into #5343?

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -647,6 +704,141 @@ mod tests {
...
+    }
+
+    // The ordered scan reads each file as its own sorted partition and relies on
+    // SortPreservingMergeExec to k-way merge them into one globally sorted stream. This feeds known
+    // sorted partitions (with duplicate keys across partitions, and both asc and desc) into that
+    // merge with the same kind of LexOrdering the planner builds, and checks the output is globally
+    // sorted and complete. It is deterministic coverage of the merge that does not depend on an
+    // ordering-reporting Iceberg build (which is why the end-to-end suite's merge assertions cancel
+    // on the published Iceberg used in CI).
+    async fn merge_ints(input: Vec<Vec<i32>>, descending: bool) -> Vec<i32> {
```
andygrove, line 824: `merge_ints` and the two `spm_merges_*` tests below build a `MemorySourceConfig` and merge it, so they never touch `IcebergScanExec`, the planner or the proto. Every symbol they use predates this PR, which means they compile and pass verbatim on `main`. I checked with a mutation: changing `IcebergScanExec::execute` to `let tasks = self.tasks.clone()` so every partition reads every file leaves both of them green.

The useful part is that the gap looks closable without waiting on Iceberg. I tried writing probes that write real Parquet into a tempdir and drive the ordered path through `IcebergScanExec` and through `create_plan`, and the whole set runs in about 50ms. Merging three real files ascending, descending, and with duplicate keys across them gives tests that the mutation above does kill. Going through `create_plan` and asserting the merge path and the `maxFilesPerPartition` sort-fallback path produce identical output on the same data covers the proto to `create_sort_expr` to `LexOrdering` translation, which nothing exercises today. Round-tripping the null ordering through the proto in all four direction and null-ordering combinations covers ASC NULLS LAST and DESC NULLS FIRST, which are not Iceberg's defaults but are legal `SortField` values.

All of those pass on your branch, so this is a coverage gap rather than a bug report. Happy to hand you the patch if it saves you time.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -914,16 +915,69 @@ case class CometScanRule(session: SparkSession)
...
+                CometIcebergNativeScan
+                  .reportableOrdering(scanExec.ordering, scanExec.output, unsafe)
+              case None => Nil
+            }
+          }
+        }
+        // Bind the reported ordering to proto now, against the same output the gate used, so the
+        // executor-side serde writes it directly. None means the binding failed -- the gate below
+        // then keeps the scan on Spark instead of converting and hard-failing at task start.
+        val reportedOrderingProto: Option[Seq[Expr]] =
```
andygrove, line 1049: `serializeReportedOrdering` opens with the same `if (reportedOrdering.isEmpty) Some(Nil)` check that this branch does, so one of the two is unreachable. It has a single caller, so dropping the guard here reads a little better since the callee is the one that documents the contract.

native/core/src/execution/operators/iceberg_scan.rs
parthchandra: Nice catch. Let me take care of this in #5343

native/core/src/execution/operators/iceberg_scan.rs
parthchandra: Can you point me to your patch, I'll merge it in. Thank you!

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
parthchandra: Done.

parthchandra:
> the PR description still describes the key-grouped partitioning reporting and `spark.comet.scan.icebergNative.reportPartitioning.enabled`, both of which went away when the `outputPartitioning` override was dropped, and it does not mention `sortMerge.maxFilesPerPartition`. Since this becomes the squash commit message, could you refresh it? The "Summary of changes" bullets are otherwise still accurate.

Updated the PR description

parthchandra:
@andygrove any more comments?

andygrove:
I built Iceberg from apache/iceberg#16750 plus apache/iceberg#14948 and ran this suite against it with a probe at the native merge-or-sort decision. Every test on an unpartitioned table runs the old unordered read, because `isOrderingEnabled` in the Iceberg PR needs a non-empty grouping key type and Iceberg's own `testNoMergeReaderForUnpartitionedSortedTable` asserts exactly that. That covers "merges multiple sorted files", the duplicate-key, MOR-delete, descending, multi-column, single-file, on-versus-off and order-by-limit tests, and both 70-file tests. The above-the-cap sort fallback therefore has no end-to-end coverage at all.

I also ran two mutations. Making every ordered partition read every file was caught by the NULLS FIRST, NULLS LAST, partitioned-table, SMJ and group-by tests only. Flipping the merge direction so rows come out complete but mis-ordered was caught by the SMJ test alone, since every test with a global ORDER BY has its order repaired by Spark's final sort.

Could the merge tests move to the shape the NULLS tests already use, a table partitioned by a constant column with preserve-data-grouping on, and add an `assumeOrderingReported` after the correctness check so a reporting build proves the merge ran? For the 70-file tests the same change would make the cap boundary real. And for order sensitivity, a join, a window, or a `SortAggregate` over the merged input is what actually verifies the order, since ORDER BY cannot.

One thing to be aware of on the Iceberg side: with those two PRs as they stand, Spark's own answer is nondeterministically wrong. The window test failed for me with Spark returning `row_number` 2 for the first row, and a run with Comet's sort path untouched had Spark return 4 join rows where 6 is correct. Which revisions did you validate with?

parthchandra:
Thanks for thanks for building the fork and measuring this.  I verified this PR against against a personal build of  Spark 3.4.3 + Iceberg build whose isOrderingEnabled is correct, which is why I didn't hit the nondeterminism you saw (your window/join getting wrong Spark answers is an Iceberg-side issue, not this PR).
You're also right that the merge tests were on unpartitioned tables. I've moved all of them to the partitioned-by-constant pattern with assumeOrderingReported after the correctness check, so a build that reports ordering proves the merge ran. 
The above-the-cap sort fallback is now partitioned too. And the ascending/descending merge tests now verify order with a ROW_NUMBER() window (no ORDER BY to repair it), matching the existing window test. 
PTAL.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -1022,16 +1023,66 @@ case class CometScanRule(session: SparkSession)
...
         }
 
+        // If Iceberg reports an ordering, EnsureRequirements may have already dropped the Sort
+        // above this scan (it decides that on the vanilla BatchScanExec, before Comet converts the
+        // scan). If the native scan cannot guarantee that ordering, reading unordered here would
+        // silently return wrong results, so stay on Spark -- its Iceberg reader produces the sorted
+        // output it promised. Evaluate the gate exactly once here and stash the result on the
+        // metadata; CometIcebergNativeScanExec.outputOrdering and the proto serde both read that
+        // stashed value, so the reported order cannot diverge from what native advertises.
+        val icebergReportsOrdering: Boolean = scanExec.ordering.exists(_.nonEmpty)
```
andygrove, line 1033: I went and read the Iceberg side to work out when `orderingHonored` actually fires, and I think it costs us the whole native scan in a case where it does not have to.

In apache/iceberg#16750, `SparkPartitioningAwareScan.outputOrdering()` returns `Spark3Util.toOrdering(table().sortOrder())` whenever `isOrderingEnabled()` holds. `SortOrderAnalyzer.canReportOrdering` checks four things: the table has a sort order, the sort order has no nested fields, each partition key maps to exactly one task group, and every file carries the current sort order id. None of them looks at the projection. On the Spark side, `V2ScanPartitioningAndOrdering.ordering` resolves those references against `relation`, which is the unpruned relation, and unlike the `partitioning` branch directly above it there is no `references.subsetOf(d.outputSet)` guard.

So for `SELECT data FROM sorted_table` where the sort key is `id`, `scanExec.ordering` comes back non-empty carrying `id`'s exprId, which is not in `scanExec.output`. `isIdentityProjected` fails, `reportedOrdering` is `Nil`, `orderingHonored` is false, and the whole Iceberg scan goes back to Spark.

I do not think that buys any safety. A required ordering can only reference attributes in the scan's output, and `SortOrder.orderingSatisfies` matches position by position starting at index 0, so an ordering whose first field is an unprojected attribute satisfies nothing above the scan. Spark has dropped no `Sort`, and reading unordered would be correct.

Could we keep the scan native when the first reported sort field references an attribute outside `scanExec.output`, and report no ordering in that case? The general version is to honour the longest prefix of identity-projected fields, since that is exactly as far as any satisfiable required ordering can reach. Declining is still the right answer when the blocking field is a projected UUID, because a required ordering can reach into it, so the UUID path would not change.

This matters because it is the common shape. Once `preserve-data-ordering` is on, any query on a sorted table that does not select every sort key loses the native scan. The follow-up bullet in the PR description still describes this case as "we skip the merge and read unordered", which was true before the `orderingHonored` gate went in.

spark/src/test/scala/org/apache/comet/CometIcebergSortMergeReadSuite.scala
```diff
@@ -0,0 +1,885 @@
...
+        "(2,'b','P1'),(4,'d','P1')",
+        "(5,'e','P2'),(7,'g','P2')",
+        "(6,'f','P2'),(8,'h','P2')")
+
+      checkSparkAnswer(s"SELECT c1, c2 FROM $cat.db.t WHERE c3 = 'P1' ORDER BY c1")
+      checkSparkAnswer(s"SELECT c1, c2 FROM $cat.db.t WHERE c3 = 'P2' ORDER BY c1")
+    }
+  }
+
+  test("sort key absent from the projection: falls back to an unordered read but stays correct") {
```
andygrove, line 574: Following on from the comment on `CometScanRule`, I do not think this test pins the behaviour it is named for.

```scala
val (_, plan) = checkSparkAnswer(s"SELECT c2 FROM $cat.db.t WHERE c3 = 'P1'")
nativeScans(plan).foreach { scan =>
  assert(scan.outputOrdering.isEmpty, ...)
}
```

On a reporting build this query falls back to Spark, so `nativeScans(plan)` is empty and the `foreach` asserts nothing. On the published Iceberg nothing was reported in the first place, so the assertion is trivially true there too. Either way it passes, and the title says "falls back to an unordered read" while the behaviour is a fallback to Spark.

Could this assert the outcome directly, whichever one we settle on above? If the scan should stay native with no ordering, `assert(nativeScans(plan).nonEmpty)` first and then the `outputOrdering.isEmpty` check. If it should stay a Spark fallback, `assertFellBackToSpark` already exists and says so.

docs/source/user-guide/latest/iceberg.md
```diff
@@ -180,6 +193,12 @@ The following scenarios will fall back to the JVM Iceberg reader:
 - Scans with residual filters using `truncate`, `bucket`, `year`, `month`, `day`, or `hour`
   transform functions (partition pruning still works, but row-level filtering of these
   transforms falls back)
+- Sorted tables where the sort key is a transform (e.g. `bucket`, `truncate`) rather than a plain
+  column: Comet reports only identity sort keys today, so the scan falls back to Spark to preserve
+  the reported ordering
+- Sorted tables whose sort key is a `UUID` column: Iceberg sorts UUID by its own byte comparator
```
andygrove, line 199: The transform sort key and the UUID sort key landing here is exactly what I was after last round, thank you.

The unprojected sort key belongs in this list too, since as things stand today that is also a fallback to Spark and it will be the most common one people hit. Whatever we settle on in the `CometScanRule` comment, the list should describe it.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeScan.scala
```diff
@@ -893,6 +894,82 @@ object CometIcebergNativeScan extends CometOperatorSerde[CometBatchScanExec] wit
...
+   * v1 accepts only identity sort fields on top-level columns that are in the projection. Each
+   * SortOrder child must be an AttributeReference in `output`, and must serialize to proto.
+   * Transform sort fields (bucket/truncate/...) are not AttributeReferences, so they fall through
+   * to Nil and we read unordered. Checking exprToProto here, not just in the proto path, keeps
+   * the two callers in step: outputOrdering never advertises an order the proto path would drop.
+   *
+   * We trust Iceberg on file-level sortedness. If it reports an ordering, SortOrderAnalyzer has
+   * already checked each file's sort_order_id matches the table order, so every file is sorted.
+   *
+   * We read scanExec.ordering (the raw reported order), not scanExec.outputOrdering. Spark blanks
```
andygrove, line 913: This docstring explains the choice of `scanExec.ordering` over `scanExec.outputOrdering` with "Spark blanks outputOrdering when a partition holds more than one file", and I think that has the mechanism backwards in a way worth fixing, because this is the safety-critical read.

Spark blanks it when multiple `InputPartition`s share a partition key, not when a partition holds multiple files. `DataSourceV2ScanExecBase.outputOrdering` is `ordering.filter(_ => groupedPartitions.forall(_.groupedParts.forall(_.parts.length <= 1)))`. Iceberg's `SortOrderAnalyzer.hasUniquePartitionKeys` refuses to report in exactly that same case, with a comment giving the same reason. So on the Iceberg path the two values are always equal, and the many-files-in-one-partition case this merge targets is one where Spark does not blank at all.

That means reading `outputOrdering` instead would be behaviour-preserving today, and it would stop us depending on an Iceberg-side invariant to stay safe. If a future Iceberg relaxes `hasUniquePartitionKeys`, or another source reports an ordering Spark then blanks, reading `ordering` has us pay a k-way merge, or a full spillable `SortExec` above the cap, underneath a `Sort` that Spark kept anyway. Would you rather switch the read, or keep `ordering` and have the comment name the Iceberg check it is leaning on?

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -707,6 +768,141 @@ mod tests {
...
```
andygrove, line 831: Picking this thread back up. `merge_ints` and the two `spm_merges_*` tests below it still build a `MemorySourceConfig` and merge it, so they never touch `IcebergScanExec`, the planner, or the proto, and they still compile and pass verbatim on `main`. The mutation I described holds: changing `IcebergScanExec::execute` to `let tasks = self.tasks.clone()` so every partition reads every file leaves both green.

The three planner tests added alongside them are real coverage of the merge-versus-sort decision and of the cap arriving over the proto, so the gap is narrower than it was. What is still uncovered is the data path, reading actual sorted files through `IcebergScanExec` and checking what comes out.

You asked me for the patch and I dropped it, sorry about that. The shape is three real Parquet files written into a tempdir and driven through `IcebergScanExec`, ascending, descending, and with duplicate keys across files, then the same data through `create_plan` asserting the merge path and the `maxFilesPerPartition` sort path agree, plus the null ordering round-tripped through the proto in all four direction and null-ordering combinations. It runs in about 50ms and the mutation above kills it. Say the word and I will push it to your branch.

native/core/src/execution/operators/iceberg_scan.rs
```diff
@@ -175,18 +239,14 @@ impl IcebergScanExec {
...
         context: Arc<TaskContext>,
     ) -> DFResult<SendableRecordBatchStream> {
         let output_schema = Arc::clone(&self.output_schema);
-        let file_io = load_file_io(
-            &self.catalog_properties,
-            &self.metadata_location,
-            &self.catalog_name,
-            AccessMode::Read,
-        )?;
+        let file_io = self.file_io.clone();
```
andygrove, line 246: On the per-file `CachingDeleteFileLoader`, you said you would fold this into #5343. Reading that issue, its body is entirely about bounding how many files the merge opens at once, and it does not mention the delete loader being rebuilt per `execute`, or the measurement: `bytes_scanned` going from 2717 to 4256 with two data files sharing one positional delete file, and up to 64x on a merge-on-read table at the default cap.

Could you add it to #5343, or file it separately? Once this lands, the shape of the regression is not recoverable from the code without redoing the measurement.

andygrove (commented):
Another pass over the current head. Most of what I raised earlier has landed and reads well: the single evaluation of the gate, the planning-time proto binding, the `checkValue` bound, the `maxFilesPerPartition` cap, `enabled=false` meaning "sort instead of merge" rather than "give up the scan", and moving the merge tests onto partitioned tables with `assumeOrderingReported` plus a `ROW_NUMBER()` order check. Thanks for working through all of it.

Six things left. Five are inline. The first of those is new and is the one I care most about. Two are threads from earlier passes that never quite closed.

The sixth has no diff line to attach to. The branch is `CONFLICTING`. It is based on `4c2ab9686`, from before #5933, and `CometScanRule.scala` has taken three commits on `main` since then. The conflict is with #5732, which removed `complexTypePredicatesSupported` from the same eligibility conjunction that this PR adds `orderingHonored` to, so the resolution is mechanical. #5732 also reworked `IcebergReflection`, though not the method this PR refactors, so that file merges clean. Worth re-running the Iceberg matrices after the merge rather than trusting the current green, since it was measured against a tree three weeks behind `main`.

PR #5332, opened 2026-08-11
bug: Revert "chore(deps): bump object_store_opendal from 0.57.0 to 0.58.0 in /native"

coderfender:
Reverts apache/datafusion-comet#5289


comphead (approved):
Thanks @coderfender looks like the opendal version changed and now it is a dependency conflict

Issue #5347, opened 2026-08-13
Implement unbase64 natively instead of JVM codegen dispatch


andygrove:
## Summary

`unbase64` is the last remaining codegen-dispatched member of Comet's base64 family: `base64` has a
native Rust kernel (`spark_base64`, `native/spark-expr/src/string_funcs/base64.rs`) with a criterion
benchmark, while the decode direction still runs Spark's generated JVM code inside the Comet
pipeline. Closing this asymmetry removes 5 JVM heap allocations per non-null row and a double scan of
every input value.

Assessed with the `suggest-native-expression` skill (survey mode over the 58 `Codegen dispatch` rows
in the expression reference). `unbase64` topped the ranking: **compatibility confidence High**,
**native upside High**.

Note on prior art: #423 ("Support spark unbase64 function") is closed, but it was about *support*,
which was delivered via codegen dispatch. This issue is about the native implementation, which does
not exist.

## How it runs today

`object CometUnBase64 extends CometCodegenDispatch[UnBase64]`
(`spark/src/main/scala/org/apache/comet/serde/strings.scala`). A plain `CometCodegenDispatch` means
the codegen dispatcher is the only path: no native path exists and there is no `NativeOptInAvailable`
opt-in. `getSupportLevel` is `Compatible()`, so behavior matches Spark exactly, at the cost of one
JNI round trip per batch plus per-row JVM evaluation.

`StringType` input and `BinaryType` output are both inside `CometBatchKernelCodegen.isSupportedDataType`,
so there are **no** input shapes that fall back to Spark entirely. The only fallback trigger is
`spark.comet.exec.scalaUDF.codegen.enabled=false`. The upside here is purely the dispatch cost, not a
recovered whole-operator fallback.

## Native upside: High

Spark's `doGenCode` for `UnBase64` (unchanged across 3.4.3, 3.5.8, 4.0.1, 4.1.1) emits:

```java
${ev.value} = java.util.Base64.getMimeDecoder().decode($child.toString());
```

`getMimeDecoder()` returns a cached singleton, so that part allocates nothing. Everything else does.
Per non-null row, counted from the Spark and JDK 17 sources rather than estimated:

1. `UTF8String.toString()` calls `getBytes()`. In the dispatcher, strings are read zero-copy through
   `UTF8String.fromAddress(null, addr, len)`
   (`CometBatchKernelCodegenInput.scala:275`), so `base` is off-heap and `getBytes()` takes the
[...]

Issue #4654, opened 2026-06-14
Discussion: remove incompatible native implementations that have a codegen dispatch fallback?


andygrove:
## Background

Comet has a `CodegenDispatchFallback` mechanism for expression serdes that have a native implementation which is `Incompatible` with Spark for some (or all) inputs. When such an expression reports `Incompatible` and the user has not enabled that expression's `allowIncompatible`, it routes through the JVM codegen dispatcher (Spark's own `doGenCode` compiled into an Arrow-direct batch kernel) instead of falling the operator back to Spark. The dispatcher is on by default (`spark.comet.exec.scalaUDF.codegen.enabled=true`).

A consequence of this design: for these expressions the incompatible native code is **only reachable when `allowIncompatible=true`**. By default everyone already runs through the (correct) codegen dispatcher. So for the "always incompatible" expressions, the native impl exists purely to give opt-in users a faster-but-wrong result instead of a correct dispatcher result that is already near-native speed.

This issue is to discuss whether we should remove some of these incompatible native implementations entirely and rely on codegen dispatch.

## Survey

There are currently 14 serdes mixing in `CodegenDispatchFallback`. They split into two groups.

### Group A: always `Incompatible` (native path is opt-in / wrong-by-default)

`getSupportLevel` returns `Incompatible` for every non-error input. Default users always route through codegen dispatch; the native impl only runs under `allowIncompatible`.

| Expression | File | Reason for incompatibility |
|---|---|---|
| `CometFromUTCTimestamp` | `serde/datetime.scala` | DataFusion timezone-string parsing diverges from Spark (e.g. `GMT+1`, `PST` throw a native parse error) — #2013 |
| `CometToUTCTimestamp` | `serde/datetime.scala` | same timezone-parse divergence — #2013 |
| `CometConvertTimezone` | `serde/datetime.scala` | same timezone-parse divergence — #2013 |
| `CometFromUnixTime` | `serde/unixtime.scala` | only supports the default pattern; DataFusion's valid timestamp range differs from Spark — apache/datafusion#16594 |
| `CometArrayExcept` | `serde/arrays.scala` | null handling and ordering may differ from Spark |
| `CometArrayJoin` | `serde/arrays.scala` | null handling may differ from Spark |
| `CometArrayIntersect` | `serde/arrays.scala` | result element order may differ when the right array is longer (DataFusion probes the longer side) |

### Group B: conditionally `Incompatible` (native path is `Compatible` for the common case)

[...]

Issue #4506, opened 2026-05-28
[EPIC] Provide JVM/codegen-dispatch implementations for Incompatible expressions so they never fall back by default


andygrove:
## Background

A Comet expression marked `Incompatible(Some(reason))` in `getSupportLevel` is the native Rust implementation telling the dispatcher "I diverge from Spark in this case." Today the dispatcher behaviour is binary:

- `spark.comet.expr.allowIncompatible=false` (default): Comet falls back to Spark for the whole projection, losing the native pipeline for every other expression in that projection.
- `spark.comet.expr.allowIncompatible=true`: Comet runs the divergent native impl and the user opts in to the wrong-answer risk.

Neither default is great. The fallback path silently regresses performance for users who never opted in to the trade-off, and `allowIncompatible=true` changes correctness for every Incompatible expression at once.

## Proposal

For each `Incompatible` branch, provide a JVM-side implementation routed through the existing codegen-dispatch path (the same mechanism that wires `CometScalaUDF` to Janino-codegen Scala functions). The dispatcher then has a third option that is always correct: stay in Comet, evaluate the divergent expression in the JVM via codegen, and keep the rest of the projection native.

This lets us flip the default for Incompatible expressions to "Comet handles it correctly without falling back," and leaves `allowIncompatible=true` purely as a perf knob for users who accept the native-path divergence.

## Scope by category

Each item below is an `Incompatible` branch with a file:line anchor in `spark/src/main/scala/org/apache/comet/serde/`. Conditional Incompatibles (e.g. only for `TimestampNTZ` or only with `COMET_EXEC_STRICT_FLOATING_POINT=true`) only need the JVM dispatch for the divergent branch; the rest stays on the native path.

### String

- `CometInitCap` (`initcap`) - `strings.scala:98` - unconditional Incompatible (Unicode case mapping divergence).
- `CometConcat` (`concat`) - `strings.scala:236` - Incompatible for `ArrayType` inputs.
- `CometRegExpReplace` (`regexp_replace`) - `strings.scala:384` - Java/Rust regex engine differences.
- `CometStringSplit` (`split`) - `strings.scala:424` - Java/Rust regex engine differences.
- `CometGetJsonObject` (`get_json_object`) - `strings.scala:451` - single-quoted JSON / control-character handling diverges from Spark.

### Date/time

- `CometHour` (`hour`) - `datetime.scala:191` - Incompatible for `TimestampNTZType` (#3180).
[...]

PR #5348, opened 2026-08-13
docs: add suggest-native-expression skill for assessing native expression candidates

andygrove:

N/A. This adds contributor tooling rather than fixing a filed issue. It supports the roadmap item

## Rationale for this change

58 Spark expressions currently run through the JVM codegen dispatcher. Deciding which of them is
worth a native Rust implementation is a recurring judgment call, and it is easy to get wrong in
both directions: proposing a native kernel that cannot match Spark, or spending a week on a
rewrite whose only saving is the per-batch JNI round trip.

The two questions that decide it are always the same:

1. Can a native path be **100% Spark-compatible**? The dispatcher it would replace is byte-exact by
   construction, so replacing it with an approximate path is a regression, not an optimization.
2. Does it deliver a real **throughput or allocation** win over codegen dispatch? The dispatcher
   already runs in-pipeline, reads Arrow vectors directly, reads strings zero-copy, and compiles its
   kernel once per `(expression, schema)` pair. "Native is faster than JVM" is not an argument on its
   own.

Today those questions get re-answered from scratch every time, and the answers are not written down,
so a candidate that was already ruled out can be proposed again. This skill makes the assessment
repeatable and, more importantly, **cumulative**: every verdict lands in the per-expression audit log,
including the disqualifications, so the pool shrinks over time instead of being re-walked.

- **`.claude/skills/suggest-native-expression/SKILL.md`**, a new skill that:
  - builds the candidate pool from the `Codegen dispatch` rows of the expression reference, excluding
    families the project has ruled out, expressions already assessed in the audit log, and
    expressions with an existing issue;
  - scores **compatibility confidence** against a 12-row hazard checklist (JVM formatting APIs,
    `java.util.regex` features Rust's `regex` lacks, Spark 4 collation, `BigDecimal`/`MathContext`,
    tzdb and hybrid-calendar rebase, ANSI error-class parity, `Double.toString`, invalid-UTF-8 byte
    semantics, lambda bodies, JVM session state, nondeterminism, cross-version divergence), with a
    `Low` rating as a hard disqualifier;
  - scores **native upside** from what the dispatcher actually costs, and spells out what it does
    *not* cost so the upside cannot be overstated. The rating has to rest on per-row heap allocation,
    hoistable per-row work, a named technique from `optimizing_expressions.md`, nested-type wrapper
    allocation, a recovered whole-operator Spark fallback, or workload presence;
  - requires an empirical calibration step when either axis lands on `Medium`, since that is where a
    static read is least trustworthy;
  - maps the two ratings to `Recommended` / `Deferred` / `Disqualified`, and requires a specific,
    re-checkable blocker for anything not recommended;
  - files exactly one issue per run and records the verdict in the audit log for **every** verdict,
    including disqualifications.
- **`docs/source/contributor-guide/expression-audits/index.md`**: documents the new
  `Native candidate (assessed ...)` line as a third kind of audit entry, alongside correctness and
  performance audits.

The audit-log lines the skill produced when I ran it were originally in this PR. Per
[@comphead's review](https://github.com/apache/datafusion-comet/pull/5348#discussion_r3778885329)
they are now split into **#5350** (draft), so this PR is the skill and its output format only.

## Example of what the skill produces

Running the skill in survey mode over the pool produced
**[#5347: Implement `unbase64` natively instead of JVM codegen dispatch](https://github.com/apache/datafusion-comet/issues/5347)**,
which is a fair sample of the output quality this is aiming for:

- **Ranking.** 58 candidates. `unbase64` topped it because `base64` already has a native kernel
  (`spark_base64`) with a criterion benchmark while the decode direction does not, so the family is
  asymmetric for no stated reason.
- **Upside, counted rather than asserted.** 5 JVM heap allocations per non-null row, traced through
  `UTF8String.getBytes()` (which copies, because the dispatcher hands it an off-heap
  `UTF8String.fromAddress`), `new String(bytes, UTF_8)`, `decode(String)`'s `getBytes(ISO_8859_1)`,
  and the output `byte[]`. About 41k allocations per 8192-row batch, plus a double scan of every value
  because MIME's `decodedOutLength` pre-scans to size the output. All removable with the
  `spark_unhex` pattern.
- **Compatibility, checked across versions.** Semantics identical on 3.4.3, 3.5.8, 4.0.1, and 4.1.1,
  so no shims. Collation and invalid-UTF-8 are explicitly cleared **with reasoning** rather than
  waved off: `unbase64` never routes through `CollationSupport`, and Spark itself decodes through
  `toString()`, so Comet's ingress replacement produces the same character stream.
- **The finding that makes the issue worth filing.** `datafusion-spark::SparkUnBase64` exists, so the
  obvious move is to wire it up. Reading it shows it lowers to `decode(bin, 'base64pad')`, whose
  strict engine errors on the non-alphabet bytes Spark's MIME decoder skips. Since Spark's `base64`
  CRLF-chunks at 76 characters by default, wiring the upstream function would **break round-tripping
  Comet's own `base64` output**. The skill's "read it and diff the semantics before crediting it" rule
  is what caught that.
- **Honest limits.** The allocation counts are labelled as derived from source, not measured, and
  workload presence is labelled unmeasured (zero hits in `benchmarks/tpc/queries/`).

The corresponding audit-log line records the same verdict in one scannable bullet, so the next run
sees `unbase64` is taken without re-reading any of it. Those lines are in #5350.

A second run produced **[#5349](https://github.com/apache/datafusion-comet/issues/5349)** for
`sequence`, which exercised the parts the first run did not: the already-assessed exclusion, the
required empirical calibration on a `Medium` rating (dumping the real dispatcher kernel and counting
allocation sites in the emitted Java), and a scoped rather than blanket recommendation. See the
comment thread below.

Docs and skill definition only, so there is nothing executable to test. Validation was to run the
skill end to end, which is what produced #5347 and the `unbase64` audit-log entry. Every command in
the skill was executed during that run, and the four issues it cites (#4496, #4506, #4654, #4764)
were verified open. `npx prettier` was run over the changed markdown.

### DIFF
--- a/.claude/skills/suggest-native-expression/SKILL.md
+++ b/.claude/skills/suggest-native-expression/SKILL.md
@@ -1,0 +1,497 @@
+---
+name: suggest-native-expression
+description: Use when picking the next Spark expression to implement natively in Comet, or when asked whether a specific codegen-dispatched expression is worth a native Rust implementation. Scores the candidate on how likely a native path can be 100% Spark-compatible and how much throughput or allocation it would save versus the JVM codegen dispatcher, records the verdict in the per-expression audit log so disqualified candidates are not re-litigated, and files a GitHub issue for the recommendation.
+argument-hint: [expression-name]
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+Assess whether a Spark expression that Comet runs through the JVM codegen dispatcher is worth
+implementing natively in Rust, and propose the best candidate as a GitHub issue.
+
+This skill does **not** implement anything. Its deliverables are exactly three:
+
+1. A verdict on two axes: **compatibility confidence** (can a native path be 100%
+   Spark-compatible?) and **native upside** (how much throughput or allocation does the native
+   path save versus codegen dispatch?).
+2. A dated `Native candidate` line in the expression's page under
+   `docs/source/contributor-guide/expression-audits/`, so a future run can see that this
+   expression was already assessed and why it was or was not taken.
+3. One filed GitHub issue for the recommended (or deferred-with-a-named-blocker) candidate.
+
+`$ARGUMENTS` may name a single expression to assess. If it is empty, run in survey mode (Step 1)
+and pick the best candidate from the whole pool.
+
+## Background reading
+
+Read these before scoring anything. Do not score from memory: the routing model and the
+cost of the dispatcher are both easy to misremember, and every judgment in this skill depends
+on them.
+
+- `docs/source/contributor-guide/roadmap.md`, section "Native Coverage for Codegen-Dispatched
+  Expressions". This is the project-level statement of why this work matters and it is the
+  section the filed issue should reference.
+- `docs/source/user-guide/latest/compatibility/index.md`, section "Native and codegen-dispatch
+  implementations". Defines the two paths and which one is the default.
+- `docs/source/user-guide/latest/expressions.md`, the Implementation legend and tables. The
+  `Codegen dispatch` rows are the candidate pool.
+- `docs/source/contributor-guide/optimizing_expressions.md`, the "Proven techniques" table. If
+  no technique in that table applies to the candidate, the native upside claim is weak.
+- `spark/src/main/scala/org/apache/comet/serde/CometExpressionSerde.scala` for the
+  `CometCodegenDispatch` / `CodegenDispatchFallback` / `NativeOptInAvailable` contracts.
+- `spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala` for what the
+  dispatcher can and cannot compile (`isSupportedDataType`, `canHandle`).
+
+## What the dispatcher actually costs
+
+Getting this right is the whole point of the skill. Overstating the dispatcher's cost produces
+issues that waste a contributor's week on a 2% win.
+
+What codegen dispatch **does** cost:
+
+- One JNI round trip per batch, out of native execution and back.
+- Per-row evaluation of Spark's generated Java. No vectorization, no SIMD, no dictionary
+  awareness, no bulk buffer operations.
+- Per-row JVM heap allocation for every non-primitive intermediate and result: `UTF8String`,
+  `Decimal`/`BigDecimal`, `GenericArrayData`, `ArrayBasedMapData`, boxed values, plus whatever
+  Spark's own `doGenCode` allocates (formatters, `StringBuilder`, temporary arrays).
+- That allocation lands on the JVM heap, outside Comet's native memory pool, so it is invisible
+  to Comet's memory accounting and it feeds GC pressure that shows up as task-level jitter
+  rather than as an expression cost.
+- Per-row reads of nested Arrow data go through `CometArrayData` / `CometMapData` /
+  `CometSpecializedGettersDispatch` wrappers, which allocate per row per nesting level.
+
+What codegen dispatch **does not** cost, and therefore is not available as an argument for going
+native:
+
+- It does **not** convert the whole batch to rows and back. The kernel reads Arrow vectors
+  directly.
+- String reads are zero-copy (`UTF8String.fromAddress`).
+- The Janino kernel is compiled once per `(expression, schema)` pair, not per batch.
+- A `NullIntolerant` root short-circuits null rows without evaluating the body.
+- It is byte-exact with Spark by construction, so it needs no compatibility work at all. A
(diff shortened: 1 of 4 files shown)


.claude/skills/suggest-native-expression/SKILL.md
```diff
@@ -0,0 +1,497 @@
...
+
+## Step 3: Score compatibility confidence
+
+The question is narrow: **can a native Rust implementation match Spark bit-for-bit, for every
+input, on every supported Spark version?** "Close enough" is not a passing answer, because the
+dispatcher it would replace is exact.
+
+### Read the Spark implementation across versions
+
+Reuse the clone recipe from `audit-comet-expression` (Spark 3.4.3, 3.5.8, 4.0.1, 4.1.1) and read
```
comphead: not sure if we need hardcode versions? should it be revised when support dropped/added?

docs/source/contributor-guide/expression-audits/string_funcs.md
```diff
@@ -245,6 +245,10 @@
...
 - Spark 4.0.1 (audited 2026-05-27): unchanged alias of `Upper`.
 
+## unbase64
+
+- Native candidate (assessed 2026-08-13): Recommended ([#5347](https://github.com/apache/datafusion-comet/issues/5347)). Compatibility: High (semantics identical across 3.4.3, 3.5.8, 4.0.1, and 4.1.1; no collation, timezone, locale, or decimal dependence; the `java.util.Base64.getMimeDecoder()` skip rule and its four `IllegalArgumentException` conditions are fully enumerable from the JDK source, and Comet already ports the mirror-image MIME encoder in `spark_base64`). Upside: High (5 JVM heap allocations per non-null row, roughly 41k per 8192-row batch, plus a double scan from the MIME `decodedOutLength` pre-scan, all removable with the `spark_unhex` pattern of a 256-entry table plus a preallocated `BinaryBuilder`; counts derived from source, not measured, and workload presence is unmeasured). Upstream `datafusion-spark::SparkUnBase64` is not usable: it lowers to `decode(bin, 'base64pad')`, whose strict engine errors on the non-alphabet bytes Spark skips, so it cannot round-trip Comet's own CRLF-chunked `base64` output.
```
comphead: is it part of PR?

comphead (approved):
Thanks @andygrove some nonblocking nits

andygrove:
Second run of the skill, added to this PR as another worked example: **[#5349: Implement `sequence` natively for integral types instead of JVM codegen dispatch](https://github.com/apache/datafusion-comet/issues/5349)**.

This one exercises parts of the skill that the `unbase64` run did not:

- **The exclusion check did its job.** `unbase64` was skipped because the audit log already records it as assessed, so the run went to the next-ranked candidate instead of re-deriving the same verdict.
- **A `Medium` rating triggered the required empirical calibration.** Rather than reasoning from Spark's source, the run dumped the actual dispatcher kernel through `CometBatchKernelCodegen.generateSource` and counted allocation sites in the emitted Java: `new long[n]`, plus the internal `new long[...]` and `new UnsafeArrayData()` inside `UnsafeArrayData.fromPrimitiveArray`, with two of the three scaling with sequence length, and three separate passes over every element (fill loop, `copyMemory`, output-side `getLong`/`setSafe`). Roughly 48 MB of transient `long[]` per 8192-row batch for a 365-element date spine.
- **A hypothesis got tested and came back negative.** The generated error branch calls `Map<String,String>.put(String, long)`, which looked like it might fail the Janino compile and therefore mean `sequence` silently falls back to Spark today (which would have raised the upside further). Calling `CometBatchKernelCodegen.compile` directly showed it compiles fine, so the issue says there is no hidden fallback rather than claiming one.
- **A `Medium` compatibility rating produced a scoped recommendation instead of a blanket one.** The timezone hazard is real but confined to the temporal element types, and `Sequence.impl` selects on `dataType.elementType`, so the guard is knowable at plan time. The proposal is integral types natively, date and timestamp staying on `CodegenDispatchFallback`.

The scratch test used for the kernel dump was deleted after the run, per the skill.

.claude/skills/suggest-native-expression/SKILL.md
andygrove: Good catch, and it was worse than a style issue: those versions were already stale. I copied them from `audit-comet-expression`, but `pom.xml` pins 3.4.3, 3.5.9, 4.0.4, 4.1.3 and carries a `spark-4.2` profile at 4.2.0, so the skill was sending readers to 3.5.8 and 4.0.1 sources for a build that uses 3.5.9 and 4.0.4.

Fixed in 1e2ce22 by deriving the list from the build instead:

```bash
grep -oE "<spark.version>[0-9.]+</spark.version>" pom.xml | sed -E 's/<[^>]+>//g' | sort -u
```

The hazard-table row now says "the Spark versions pinned in `pom.xml`" rather than naming tags. I also swapped the clone recipe for a `curl` of the single source file, since fetching four full Spark clones to read one class was overkill.

I re-verified both assessments against the pinned versions: the `UnBase64` and `Sequence` class bodies are byte-identical across 3.4.3, 3.5.9, 4.0.4, 4.1.3 and 4.2.0, so neither verdict changes. Worth noting `audit-comet-expression` has the same hardcoded list and is presumably worth the same fix, but I left it alone rather than widen this PR.

docs/source/contributor-guide/expression-audits/string_funcs.md
andygrove: No, and you were right to question it. These lines are output the skill produced when I ran it, not part of the skill itself, so they were muddying the review. Split out into #5350 (draft) and removed here in 1e2ce22.

This PR now contains only the skill plus the `expression-audits/index.md` change that documents the new `Native candidate (assessed ...)` line kind, which does belong with the skill since it defines the output format.

andygrove:
Thanks for the review @comphead. Let's merge and keep iterating

Issue #5349, opened 2026-08-13
Implement sequence natively for integral types instead of JVM codegen dispatch


andygrove:
## Summary

`sequence` is the highest-upside remaining codegen-dispatched expression. The dispatcher kernel makes
**3 heap allocations per non-null row and passes over every generated element three times**, and two
of those allocations scale with the sequence length rather than with the row count. For a 365-element
date spine that is roughly **48 MB of transient `long[]` garbage per 8192-row batch**.

Assessed with the `suggest-native-expression` skill: **compatibility confidence Medium** (clean
plan-time guard, see scope), **native upside High**, and the upside is **measured from the emitted
kernel**, not inferred from Spark's source.

Recommendation is scoped: implement the integral element type natively, keep the date and timestamp
variants on the dispatcher.

## How it runs today

`object CometSequence extends CometCodegenDispatch[Sequence]`
(`spark/src/main/scala/org/apache/comet/serde/arrays.scala:867`). A plain `CometCodegenDispatch` means
the dispatcher is the only path and `getSupportLevel` is `Compatible()`.

`ArrayType(IntegralType)` is inside `CometBatchKernelCodegen.isSupportedDataType`, and I confirmed the
kernel really does compile (`CometBatchKernelCodegen.compile` with two nullable `BigIntVector` inputs
returns without throwing), so there is **no** hidden whole-operator Spark fallback here. The only
fallback trigger is `spark.comet.exec.scalaUDF.codegen.enabled=false`.

## Native upside: High (measured)

Dumped the real kernel via `CometBatchKernelCodegen.generateSource` for
`Sequence(BoundReference(0, LongType), BoundReference(1, LongType), None, Some("UTC"))` over two
nullable `BigIntVector` inputs. The relevant emitted Java, verbatim:

```java
arr_0 = new long[i_1];
while (i_1 > 0) {
  i_1--;
  arr_0[i_1] = (long) (value_1 + (value_1 <= value_2 ? 1L : -1L) * i_1);
}
value_0 = UnsafeArrayData.fromPrimitiveArray(arr_0);
```

and on the output side:

```java
org.apache.spark.sql.catalyst.util.ArrayData arr_1 = value_5;
int n_0 = arr_1.numElements();
int cidx_0 = output.startNewValue(i);
for (int j_0 = 0; j_0 < n_0; j_0++) {
  outListChild_0.setSafe(cidx_0 + j_0, arr_1.getLong(j_0));
}
output.endValue(i, n_0);
[...]

Issue #4825, opened 2026-07-04
Partial project fallback: keep a projection/filter native by evaluating an unsupported subexpression in the JVM


schenksj:
_Disclosure: this issue was drafted with the help of an AI assistant._

## Is your feature request related to a problem?

Today, when a single expression in a `ProjectExec` or `FilterExec` has no native translation, Comet abandons the **whole operator**. `CometProjectExec.convert` requires every `projectList` entry to convert, and `CometFilterExec.convert` requires the predicate to convert; one `None` falls the operator back to Spark. That also breaks the native island below it: a `ColumnarToRow` transition materializes the scan output and the *supported* sibling expressions run row-wise in Spark WSCG. One unsupported expression sitting above a native (Parquet/Iceberg/Delta) scan discards the native-scan investment for the whole pipeline. This is one of the most common real-world reasons native coverage collapses.

## Describe the solution you'd like

Evaluate **only** the unsupported subexpression in the JVM and keep the operator — and the pipeline — native. This is the pattern Gluten ships as `ColumnarPartialProjectExec` and Blaze/Auron ship over Arrow FFI.

Comet is well positioned because the JVM-callback machinery already exists on `main`:
- `JvmScalarUdfExpr` — a DataFusion `PhysicalExpr` that exports its argument arrays over the Arrow C Data Interface and calls back into the JVM.
- `CometUdfBridge` + `CometScalaUDFCodegen` / `CometBatchKernelCodegen`, which Janino-compile Spark's own `doGenCode` into an Arrow-direct batch kernel.

What is missing is a *splitter*: a last-resort hook that routes an otherwise-unsupported projection/filter subexpression to that existing detour. Proposed shape for a first version:

- A last-resort hook in expression serde that retries any unsupported node via the existing codegen dispatcher, at expression granularity (fires at the outermost unsupported node; supported ancestors stay native and reference its output).
- Scoped to `ProjectExec` and `FilterExec` only in v1. Join keys, sort keys, aggregate/group expressions, and partitioning keep today's all-or-nothing behavior.
- Plan-time eligibility via the dispatcher's existing `canHandle` (rejects aggregates, generators, `Unevaluable`, unsupported types, and oversized trees); subquery-bearing trees fall back cleanly rather than failing in the kernel.
- Behind a new config, **default off**, while it stabilizes.
- **No proto or native changes** — reuses `JvmScalarUdfExpr`.

[...]

Issue #4497, opened 2026-05-28
[Bug] replace returns wrong result for empty-string search


andygrove:
## Describe the bug

`replace(str, search, replacement)` diverges from Spark when `search` is the empty literal string. Spark returns `str` unchanged (short-circuit on `search.numBytes == 0` in `StringReplace.eval`). Comet delegates to DataFusion's `replace`, which inserts `replacement` between every character and at both boundaries.

## How to Reproduce

```sql
SELECT replace('hello', '', 'x');
```

| Engine | Result |
| --- | --- |
| Spark | `hello` |
| Comet (DataFusion) | `xhxexlxlxox` |

## Additional context

Surfaced by the string-expressions audit (#4461) follow-up. Issue #3344 covered the same divergence but the body had the expected/actual values swapped, leading to it being closed as already-fixed. This issue restates the divergence with the correct direction.

Workaround: the `audit-comet-expression` follow-up marks `replace` as `Incompatible(Some(reason))` only when `searchExpr` is a literal empty string, so the dispatcher falls back to Spark for that specific case unless the user opts in with `spark.comet.expression.StringReplace.allowIncompatible=true`.

Issue #4310, opened 2026-05-13
[DISCUSS] Simplify regex engine + incompatibility config model


andygrove:
## What is the problem the feature request solves?

Comet has two regex engines (Rust DataFusion regex and a JVM-backed UDF engine), and the current config model exposes two orthogonal axes for selecting which one runs:

- `spark.comet.exec.regexp.engine` ∈ `{rust, java}` — picks the engine.
- `spark.comet.expression.<Expr>.allowIncompatible=true` (per regex expression) and `spark.comet.expression.regexp.allowIncompatible=true` (pattern-level) — opt in to running known-incompatible cases.

The axes interact non-orthogonally and the result is confusing:

- The Rust regex engine **can never be fully Spark/Java-regex compatible** because of fundamental differences in regex semantics (`\d`, `\b`, backreferences, lookaround, embedded flags, etc.). So opting into the Rust engine is, in effect, opting into incompatibilities.
- Despite that, users currently still have to set per-expression `allowIncompatible` flags on top of choosing `rust` to actually run incompatible cases natively. The two opt-ins encode the same intent twice.
- Some regex expressions (`regexp_extract`, `regexp_extract_all`, `regexp_instr`) have no Rust implementation at all. Today `engine=rust` causes those to fall back to Spark, regardless of any `allowIncompatible` setting. That's a third, implicit behavior layered on top of the two configs.

## Describe the potential solution

Treat the engine choice as the regex-wide incompatibility opt-in. Concretely:

1. Drop the per-expression `*.allowIncompatible=true` flags for regex expressions (they become redundant).
2. Extend the existing `SupportLevel` ADT so an `Incompatible` result can name the config that already opted the user in:

   ```scala
   case class Incompatible(
       notes: Option[String] = None,
       optedInBy: Option[String] = None
   ) extends SupportLevel
   ```

3. The gating logic in `QueryPlanSerde.exprToProtoInternal` treats `Incompatible(_, Some(key))` as opted-in when that config is set, in addition to the existing per-expression `allowIncompatible` check.

4. Regex serdes read the engine config inside `getSupportLevel`:

   ```scala
   // RegExpReplace, StringSplit, RLike, etc.: rust supported but incompatible
   override def getSupportLevel(expr: RegExpReplace): SupportLevel =
     CometConf.regexpEngine.get match {
       case "java" => Compatible(None)
       case "rust" => Incompatible(
         Some("Rust regex engine differs from Java in subtle ways"),
[...]

PR #5350, opened 2026-08-13
docs: record native candidate assessments for unbase64 and sequence

andygrove:

N/A. This records assessment results in the audit log. The assessments themselves are tracked by the
issues they produced, #5347 and #5349.

## Rationale for this change

Split out of #5348 in response to [@comphead's review comment](https://github.com/apache/datafusion-comet/pull/5348#discussion_r3778885329) asking whether the audit-log lines belonged in that PR. They did not: #5348 adds the
`suggest-native-expression` skill, and these two lines are output the skill produced when it was run.
Keeping them separate means the skill can be reviewed on its own merits and these lines can be
reviewed on theirs.

Draft because the lines are only worth landing if reviewers agree the assessments they record are
sound. The issues they point at (#5347, #5349) carry the full reasoning.

The first two `Native candidate (assessed ...)` lines in the per-expression audit logs:

- **`string_funcs.md`, `unbase64`**: Recommended (#5347). Compatibility High, upside High. 5 JVM heap
  allocations per non-null row, and a note that upstream `datafusion-spark::SparkUnBase64` is not
  usable because it lowers to a strict base64 engine that cannot round-trip Comet's own CRLF-chunked
  `base64` output.
- **`array_funcs.md`, `sequence`**: Recommended for integral element types only (#5349).
  Compatibility Medium, upside High. 3 heap allocations per non-null row and three passes over every
  element, measured from the emitted dispatcher kernel.

The point of these lines is to make the assessment cumulative. A later run of the skill greps for
`Native candidate (assessed` and skips anything already covered, so a candidate that was examined and
rejected does not get re-proposed. That is why the lines record the blocker and the ratings rather
than just linking the issue.

Both expressions were re-verified against the Spark versions pinned in `pom.xml` (3.4.3, 3.5.9,
4.0.4, 4.1.3, 4.2.0). The `UnBase64` and `Sequence` class bodies are byte-identical across all five,
so neither line needs a version caveat.

The new line kind is documented in `expression-audits/index.md`, which is part of #5348.

Documentation only. `npx prettier` was run over the changed files.


andygrove:
Third run of the skill, added here: **[#5351: Run `rlike` natively by default for patterns that are provably Java-regex equivalent](https://github.com/apache/datafusion-comet/issues/5351)**, recorded under `## rlike` in `predicate_funcs.md`.

This is the first run to reach the `Hybrid` sub-pool, which only became visible after the pool-widening fix in #5348, and it turned out to be the strongest candidate assessed so far:

- **The kernel already exists.** `rlike` runs on the dispatcher not because there is no native path but because `CometRLike.nativeApplicable` checks only *whether the pattern is a literal*, never what the pattern contains. So `^abc[0-9]+$`, which both engines agree on, is treated exactly like `(?<=foo)bar`, which the Rust engine cannot compile at all. The proposal is a plan-time pattern oracle mirroring `CometCast`, so the provably-equivalent subset becomes the default.
- **The divergence list is already written down.** `compatibility/regex.md` enumerates every construct that differs, and each one is detectable by inspecting a literal pattern. The issue turns that prose into a reject list.
- **Measured, from the emitted kernel.** 7 allocation sites per non-null row, 4 of them inside a single `Pattern.matcher()` call: the `Matcher` plus its `groups` (`int[20]` minimum), `locals`, and `localsPos` arrays, on top of `UTF8String.toString()`'s `getBytes` copy, `String` header, and internal `byte[]`. I also checked what is *not* a cost: the `Pattern` is compiled once into mutable state, and the input read is zero-copy, so neither can be claimed as upside.
- **It engages with the prior decision rather than ignoring it.** #4310 concluded the Rust *engine* can never be fully Java-compatible, which is true. This is a per-pattern decision, an axis that discussion did not consider. Saying so in the issue seemed better than letting a reviewer discover the overlap.
- **Multiplier.** The same oracle unlocks `regexp_replace`, `split`, `regexp_extract`, and `regexp_extract_all`, which are all `NativeOptInAvailable` for the same reason.

One thing I deliberately did not measure: the end-to-end dispatcher-versus-native A/B, which needs a release build of the native library. Since both paths already exist it is cheap for whoever implements this (`CometRegExpBenchmark` with and without `allowIncompatible=true`), and the issue puts it in the acceptance criteria rather than guessing at a number.

This PR now carries three assessments (`unbase64`, `sequence`, `rlike`). Still draft, since they are only worth landing if the assessments they record look sound.

andygrove:
Fourth run, and the first **Deferred** verdict: **[#5353: `upper`/`lower` deferred on the lack of a per-batch defer-to-dispatcher mechanism](https://github.com/apache/datafusion-comet/issues/5353)**. Recorded under both `## upper` and `## lower` in `string_funcs.md`.

`upper`/`lower` look like #5351 all over again: existing native kernel, blanket `Incompatible` reason, gated off by `spark.comet.caseConversion.enabled=false`. They are the most common string expressions in the pool, `ilike` rewrites to `Lower`, and the emitted kernel allocates 2 objects per non-null row unconditionally (`toUpperCaseAscii` is `convertAscii`, so a fresh `byte[]` plus a `UTF8String` wrapper, ~16k per batch). High upside on every axis.

It still doesn't work, for two reasons that only showed up by dumping the kernel and reading `CollationSupport`:

- **The kernel calls `execBinaryICU`, not `execBinary`.** `spark.sql.icu.caseMappings.enabled` defaults to `true` from Spark 4.0, so on 4.x the non-ASCII path is **ICU root-locale** mapping, while on 3.x (and on 4.x with that config off) it is `toString().toUpperCase()` under the **JVM default locale**. That is three behaviours across the pinned matrix, one of which is host-dependent: the same query on a JVM started with `-Duser.language=tr` produces different output, and nothing in the serialised plan captures that.
- **The obvious guard is not expressible.** Both engines agree on all-ASCII input, and Spark itself branches on exactly that property (`UTF8String.isFullAscii()`, checked first inside `CollationAwareUTF8String.toUpperCase`). But that is a per-batch runtime property, and Comet fixes the native-versus-dispatcher choice at plan time during serde. A native kernel that finds non-ASCII bytes mid-batch has nowhere to hand them.

So the issue records the blocker and two concrete unblocking conditions rather than proposing work that cannot land. The reason it is worth filing at all is that the blocker is **shared**: `initcap` and the JSON family (`get_json_object`, `from_json`, `to_json`, `length_of_json_array`) are all gated on data-dependent divergences too. A per-batch defer-to-dispatcher mechanism, adjacent to #4825 but per batch, would unlock the whole class at once — likely a better investment than picking them off individually.

Two asides from this run:

- I ranked `replace` second and dropped it, but it is probably the best *next* candidate: its entire divergence is "empty search string", which is plan-time detectable when the search argument is a literal. What stopped me recommending it is the upside, not the compatibility — `UTF8String.replace` returns `this` with zero allocation when there is no match, so unlike the other candidates the allocation saving only materialises on matching rows. That needs a match-density measurement before anyone commits to it.
- `CometInitCap.getIncompatibleReasons()` cites #1052, which is **closed**. That string is user-facing (it lands in the Compatibility Guide and in EXPLAIN output), so it is a stale link in shipped docs. Not touching it here, but worth a small follow-up.

This PR now carries four assessments across three category pages.

andygrove:
Fifth run: **[#5354: Run `replace` natively by default when the search string is a non-empty literal](https://github.com/apache/datafusion-comet/issues/5354)**, recorded under `## replace` in `string_funcs.md`.

**Correction to what I wrote last run.** In the [previous comment](https://github.com/apache/datafusion-comet/pull/5350#issuecomment-5286725412) I said `replace`'s allocation saving "only materialises on matching rows", because `UTF8String.replace` returns `this` on a non-match. That is wrong, and measuring it is what caught it. The dispatcher hands the expression an **off-heap** `UTF8String.fromAddress`, so on a non-match the returned string's base is `null` rather than a `byte[]`, and the kernel's output write takes the `getBytes()` copy branch:

```java
Object utfBase_0 = value_0.getBaseObject();
if (utfBase_0 instanceof byte[]) {
  output.setSafe(i, (byte[]) utfBase_0, ..., utfLen_0);
} else {
  byte[] utfArr_0 = value_0.getBytes();     // allocates, even for a pass-through row
  output.setSafe(i, utfArr_0, 0, utfArr_0.length);
}
```

Allocated bytes per row, via `getThreadAllocatedBytes` over 8192 off-heap-backed 64-byte values, including that output branch:

| match density | bytes/row | ns/row |
| --- | --- | --- |
| 0% | 80.0 | 171.0 |
| 10% | 85.6 | 195.8 |
| 50% | 108.0 | 170.5 |
| 100% | 136.0 | 182.4 |

The same measurement with heap-backed inputs reports **0.0 bytes/row at 0% density**, which is the number my earlier reasoning implicitly assumed. That gap is the whole correction: allocation is unconditional, 80 to 136 bytes/row, roughly 650 KB to 1.1 MB per batch.

The rest of the assessment:

- **Compatibility High**, and the narrowest gap in the pool: the entire divergence is "empty search string" (Spark short-circuits and returns the input, DataFusion inserts the replacement between every character). Decidable at plan time when the search argument is a `Literal`. Semantics identical across all five pinned versions for `UTF8_BINARY`, since Spark 4.x's `CollationSupport.StringReplace.execBinary` is the same `src.replace(...)` call as 3.x.
- **The guard is an expression property, not a data property**, so unlike `upper`/`lower` in #5353 it needs no per-batch fallback mechanism and is expressible today.
- **No native code.** The kernel exists; this is a `getSupportLevel` change plus tests.
- **What I did not prove:** the wall-clock win. Time per row is flat at 170 to 195 ns and barely moves with density, so this path is scan-bound rather than allocation-bound. Whether Rust's substring search beats `UTF8String.find`'s byte loop needs a release build of the native lib, so it is in the acceptance criteria with an explicit instruction to report the number even if it is unflattering.

Also worth noting for anyone reading the history: #3344, one of the two closed issues on this divergence, **states it backwards** (it claims Spark produces `xhxexlxlxox`). #4497 has it right. I flagged that in the issue so nobody re-derives the semantics from the wrong one.

Five assessments now, across three category pages: three Recommended, one Recommended-and-scoped, one Deferred.

PR #5357, opened 2026-08-14
fix: use Spark type names in ANSI abs overflow errors

### LINKED ISSUE #5356: ANSI `abs` integer overflow errors use Arrow-style type names instead of Spark's
### Describe the bug

Under ANSI mode, `abs` on an integer minimum raises an overflow error whose type name comes from
Arrow rather than from Spark. Comet reports `Int64 overflow` where Spark reports `long overflow`.

`native/spark-expr/src/math_funcs/abs.rs` passes Arrow-style type names (`Int8`, `Int16`, `Int32`,
and `Int64`) as the overflow `from_type`, so the rendered message diverges from Spark's `byte`,
`short`, `integer`, and `long`. The names are hard-coded at each arm, four on the array path through
`ansi_compute_op!` and four on the scalar path.

This is user-visible: the message is what a Spark user sees in the exception, and it is what a test
asserting on the error class parameters would compare against.

### Steps to reproduce

```sql
SELECT abs(col) FROM (SELECT CAST(-9223372036854775808 AS BIGINT) AS col);
```

Spark raises `[ARITHMETIC_OVERFLOW] long overflow.`
Comet raises the same error class with `Int64 overflow`.

The same divergence applies to `INT` (`integer` vs `Int32`).

### Expected behavior

The error carries Spark's type name, so `long overflow` and `integer overflow`.

### Additional context

This is item 5 ("wrong type names" in `Abs`) from #5071. That issue is now closed; #5162 addressed
item 4, while this divergence remains on `main`. I asked on #5071 whether the remaining items should
reopen it or be tracked separately and did not want to leave it sitting, so this issue tracks the
`Abs` mismatch as a self-contained fix. Happy to move it back under #5071 if you would rather track
it there.

The mapping is not uniform across supported Spark versions, and that constrains the fix:

- `integer` and `long` match Spark on 3.4, 3.5 and 4.x.
- `byte` and `short` match 4.x only. On 3.4 and 3.5, `Abs` routes those two widths to
  `QueryExecutionErrors.unaryMinusCauseOverflowError`, which raises `_LEGACY_ERROR_TEMP_2043` with
  `- <sqlValue> caused overflow.` rather than `ARITHMETIC_OVERFLOW`, so no single string satisfies
  every version. 4.0 sends all four widths through `MathUtils.negateExact`.

I verified this against the 3.4.3, 3.5.8, 4.0.2, and 4.1.2 jars.

Worth noting that the existing `abs_ansi.sql` fixture cannot catch this: every assertion is
`expect_error(overflow)`, which `Int64 overflow` satisfies just as happily as `long overflow`.

Decimal types are left alone here. `Decimal128` and `Decimal256` overflow goes through a different
Spark path and needs its own analysis.

I have a fix and tests ready and can send a PR.

Smallfu666:

Closes #5356.

## Rationale for this change

Under ANSI mode, `abs` on an integer minimum raises `ARITHMETIC_OVERFLOW`. Comet's
Spark error shim turns the native `fromType` field into Spark's arithmetic-overflow
message by passing `fromType + " overflow"` to
`QueryExecutionErrors.arithmeticOverflowError`. `native/spark-expr/src/math_funcs/abs.rs`
was passing Arrow-style type names as `fromType`, hard-coded at each arm, so Comet
reported `Int64 overflow` where Spark reports `long overflow`, and `Int32 overflow`
where Spark reports `integer overflow`.

```sql
SELECT abs(v) FROM t;   -- v BIGINT, single row -9223372036854775808
```

Spark: `[ARITHMETIC_OVERFLOW] long overflow.`
Comet before this PR: same error class, `Int64 overflow`.

The message determines the user-visible Spark exception message.

The eight `from_type` strings in `abs.rs`, four on the array path through `ansi_compute_op!` and
four on the scalar path, now carry Spark's name instead of Arrow's:

| Arrow | was | now |
| --- | --- | --- |
| `Int8` | `"Int8"` | `"byte"` |
| `Int16` | `"Int16"` | `"short"` |
| `Int32` | `"Int32"` | `"integer"` |
| `Int64` | `"Int64"` | `"long"` |

The mapping is not uniform across the supported Spark versions. Spark's `Abs` negates
through `MathUtils.negateExact`. For int and long that surfaces the JDK
`ArithmeticException` text, which is `integer overflow` and `long overflow` on 3.4, 3.5
and 4.x alike. Byte and short match 4.x only: 3.4 and 3.5 route those two widths to
`QueryExecutionErrors.unaryMinusCauseOverflowError`, raising
`_LEGACY_ERROR_TEMP_2043` (`- <sqlValue> caused overflow.`), a different error class,
so no single string can satisfy every version. 4.0 sends all four widths through
`MathUtils.negateExact`. I originally reproduced this against 3.4.3, 3.5.8, 4.0.2 and
4.1.2 jars; the same code paths remain in the currently supported 3.5.9 and 4.1.3
sources.

`Decimal128` and `Decimal256` are deliberately left alone. Those guards fire only at `i128::MIN` and
`i256::MIN`, which no Spark decimal reaches at its maximum precision of 38, and decimal overflow
goes through a different Spark error class. That belongs in its own change.

**`abs_ansi.sql`.** The fixture already ran `abs(v)` over a column of each width at its minimum, but
every assertion was `expect_error(overflow)`, which `"Int64 overflow"` satisfies exactly as happily
as `"long overflow"`. It could not fail on this bug. The int and long cases now assert the full
message, `expect_error(integer overflow)` and `expect_error(long overflow)`.

Byte and short stay on the loose pattern, and the fixture says why in a comment: `ExpectError`
asserts the pattern against Spark's message as well as Comet's, and on 3.4 and 3.5 Spark genuinely
words those two differently.

`ExpectError` runs Spark and Comet separately and requires both errors to contain the
pattern. It does not by itself prove native coverage. Native reachability for the INT
column case is established by the negative control below: reverting only the production
change while keeping the strengthened fixture makes the Comet run fail on
`integer overflow`, while Spark is unchanged.

Negative control at the Spark level, production reverted and the fixture kept:
`org.apache.comet.CometSqlFileTestSuite` on Spark 3.5 / Scala 2.12 goes from **444 run, 444
succeeded** to **444 run, 443 succeeded, 1 failed**, the failure being
`expressions/math/abs_ansi.sql:57`, `SELECT abs(v) FROM ansi_test_abs_int`, `does not contain
'integer overflow'`. Spark is identical between the two runs, so that failure is the Comet side of
the assertion.

**Rust.** The ANSI branches had no error assertions at all, so this adds two tests:

- `test_ansi_abs_min_uses_spark_type_names` downcasts to `SparkError::ArithmeticOverflow` and
  asserts `from_type` for all four widths on both the array and the scalar path. Asserting the
  variant and the field rather than a substring of the rendered string.
- `test_ansi_abs_just_inside_boundary_succeeds` checks `MIN + 1` for each width still returns `MAX`,
  so the guard is not over-broad.

Negative control on the unpatched production code with these tests kept: **18 passed, 1 failed**,
the failure being `assertion left == right failed, left: "Int8", right: "byte"`. Note it stops at
the first case in the table, so what is proven red on `main` is the byte case in Rust and the int
case at the Spark level.

Also run on this branch: the full `datafusion-comet-spark-expr` lib suite, `cargo fmt --all --
--check`, and `cargo clippy -p datafusion-comet-spark-expr --all-targets -- -D warnings`.

## Note for reviewers

TINYINT and SMALLINT have no Spark-level regression test here, only the Rust one, for the version
reason above. If you would rather see them covered, the fixture would need to gate the assertion on
the Spark version, which felt like more machinery than the case is worth.

### DIFF
--- a/native/spark-expr/src/math_funcs/abs.rs
+++ b/native/spark-expr/src/math_funcs/abs.rs
@@ -38,6 +38,7 @@ macro_rules! legacy_compute_op {
     }};
 }
 
+/// `$FROM_TYPE` is the Spark SQL type name carried in the `ArithmeticOverflow` payload.
 macro_rules! ansi_compute_op {
     ($ARRAY:expr, $FUNC:ident, $TYPE:ident, $RESULT:ident, $NATIVE:ident, $FROM_TYPE:expr) => {{
         let n = $ARRAY.as_any().downcast_ref::<$TYPE>();
@@ -95,7 +96,7 @@ pub fn abs(args: &[ColumnarValue]) -> Result<ColumnarValue, DataFusionError> {
                     let result = legacy_compute_op!(array, wrapping_abs, Int8Array, Int8Array);
                     Ok(ColumnarValue::Array(Arc::new(result?)))
                 } else {
-                    ansi_compute_op!(array, abs, Int8Array, Int8Type, i8, "Int8")
+                    ansi_compute_op!(array, abs, Int8Array, Int8Type, i8, "byte")
                 }
             }
             DataType::Int16 => {
@@ -103,7 +104,7 @@ pub fn abs(args: &[ColumnarValue]) -> Result<ColumnarValue, DataFusionError> {
                     let result = legacy_compute_op!(array, wrapping_abs, Int16Array, Int16Array);
                     Ok(ColumnarValue::Array(Arc::new(result?)))
                 } else {
-                    ansi_compute_op!(array, abs, Int16Array, Int16Type, i16, "Int16")
+                    ansi_compute_op!(array, abs, Int16Array, Int16Type, i16, "short")
                 }
             }
             DataType::Int32 => {
@@ -111,7 +112,7 @@ pub fn abs(args: &[ColumnarValue]) -> Result<ColumnarValue, DataFusionError> {
                     let result = legacy_compute_op!(array, wrapping_abs, Int32Array, Int32Array);
                     Ok(ColumnarValue::Array(Arc::new(result?)))
                 } else {
-                    ansi_compute_op!(array, abs, Int32Array, Int32Type, i32, "Int32")
+                    ansi_compute_op!(array, abs, Int32Array, Int32Type, i32, "integer")
                 }
             }
             DataType::Int64 => {
@@ -119,7 +120,7 @@ pub fn abs(args: &[ColumnarValue]) -> Result<ColumnarValue, DataFusionError> {
                     let result = legacy_compute_op!(array, wrapping_abs, Int64Array, Int64Array);
                     Ok(ColumnarValue::Array(Arc::new(result?)))
                 } else {
-                    ansi_compute_op!(array, abs, Int64Array, Int64Type, i64, "Int64")
+                    ansi_compute_op!(array, abs, Int64Array, Int64Type, i64, "long")
                 }
             }
             DataType::Float32 => {
@@ -207,7 +208,7 @@ pub fn abs(args: &[ColumnarValue]) -> Result<ColumnarValue, DataFusionError> {
                             // return the original value
                             Ok(ColumnarValue::Scalar(ScalarValue::Int8(Some(*v))))
                         } else {
-                            Err(arithmetic_overflow_error("Int8").into())
+                            Err(arithmetic_overflow_error("byte").into())
                         }
                     }
                 },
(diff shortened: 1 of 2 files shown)


sunchao (approved):
Reviewed the ANSI abs overflow type names against the canonical Spark 3.5 and 4.0 error paths. I found no P1/P2 issues introduced by this change. A Spark 4.0.4 baseline confirmed the byte, short, integer, and long overflow names.

CI has now started. The [rust-test job](https://github.com/apache/datafusion-comet/actions/runs/31780646699/job/97138515104) fails during Clippy 1.98 on `chunks_exact_to_as_chunks` in `spark_bit_array.rs` and `manual_isolate_lowest_one` in `mersenne.rs`, before tests run. I verified both files are unchanged from this PR's merge base. The full CI matrix has not passed. I did not run the full Comet abs test suite locally.

Smallfu666:
Thanks for the review, and for checking those two files against the merge base.

That clippy failure came from #5400 , which landed after this branch was cut, so I rebased onto current main to pick it up. Workspace clippy is clean locally now.

sunchao (approved):

sunchao:
Merged, thanks @Smallfu666 for your contribution!

Issue #4837, opened 2026-07-06
Support WindowGroupLimitExec (window-based limit pushdown)


andygrove:
Spark inserts `WindowGroupLimitExec` to push a rank/row-number limit below a window (for example `WHERE rn <= 10` over a ranking window). Comet does not support this operator today, so plans containing it fall back to Spark.

It is currently marked as planned in the operator support reference under Window.

comphead:
related to #2551

comphead:
This operator is a new TopK operator introduced in Spark 3.5

Before Spark 3.5

```
 Catalyst rewrote this to something like Limit(3, Window(Sort(t))), which combined with TakeOrderedAndProjectExec gave you the classic heap-based top-K over a single stream. But this only worked when there
  was no PARTITION BY — the moment you added one, you fell back to the naive full-sort-full-window plan.
```

comphead:
Prob we can use `PartitionedTopKExec` DF operator to handle this

comphead:
we currently addressing rank in Comet but we can migrate to https://github.com/apache/datafusion/pull/22885
and dense_rank https://github.com/apache/datafusion/pull/23869

PR #4547, opened 2026-05-31
feat: implement native S3 write support

kazantsev-maksim:
## Which issue does this PR close?

Part of: https://github.com/apache/datafusion-comet/issues/1625

## Rationale for this change

Currently, when Comet executes ETL queries that read from Parquet, perform a transformation, and then write back to Parquet (or S3-backed Parquet), a columnar-to-row conversion is required before the write step, because the write path falls back to the JVM Spark writer. This conversion adds unnecessary overhead and negates the performance benefits of native execution.

## What changes are included in this PR?

- **native/core/src/execution/operators/parquet_writer.rs** — Extended the native Parquet writer to support writing to S3-compatible object storage. Added S3 object store registration and wired it into the DataFusion execution context so that output paths with s3:// / s3a:// schemes are handled natively via the object_store crate.

- **spark/src/main/scala/org/apache/comet/serde/operator/CometDataWritingCommand.scala** — Updated the Scala-side CometDataWritingCommand to detect S3/S3A output paths and route them through the new native write code path instead of delegating to the JVM Spark writer. Passes the necessary S3 credentials and configuration from Hadoop/Spark config to the native layer.

## How are these changes tested?

local testing

PR #5358, opened 2026-08-14
chore: drop Spark 3.4 support

### LINKED ISSUE #4329: Drop support for Spark 3.4

This issue is to discuss the timing for dropping support for Spark 3.4

See also the proposed [versioning policy](https://github.com/apache/datafusion-comet/issues/4317).

_No response_

### Additional context

_No response_

manuzhang:

Closes #4329.

## Rationale for this change

Spark 3.4 was deprecated in Comet 1.0.0 and is scheduled for removal in 1.1.0. Dropping it reduces the supported build/test matrix and removes compatibility paths that are no longer needed now that Spark 3.5 is the oldest supported release.

- Remove the Spark 3.4 Maven profile, source/test shims, Spark SQL patch, CI jobs, release artifacts, generated compatibility page, and documentation references.
- Simplify shared code and tests to assume Spark 3.5 or newer, including removing the Spark 3.4 AQE dynamic partition-pruning fallback.
- Promote the Spark 3.5 plan-stability overrides into the base golden directories and keep Spark 4.x fallback chains intact.
- Retarget the Iceberg 1.8 compatibility job to Spark 3.5.9 and JDK 17.

- `make format PROFILES=-Pspark-3.5`
- `./mvnw clean test-compile -DskipTests -Pspark-4.0`
- `./mvnw clean test-compile -DskipTests -Pspark-4.1`
- `./mvnw clean test-compile -DskipTests -Pspark-4.2`
- Focused Spark 3.5 runtime coverage for AQE DPP and hash joins
- `CometTPCDSV1_4_PlanStabilitySuite` on Spark 3.5 (97 tests)
- `CometTPCDSV2_7_PlanStabilitySuite` on Spark 3.5 (32 tests)
- YAML, shell syntax, Python bytecode, active-reference, and `git diff --check` validation


.github/workflows/pr_build_linux.yml
```diff
@@ -336,7 +329,7 @@ jobs:
           - name: "exec"
             value: |
               org.apache.comet.exec.CometAggregateSuite
-              org.apache.comet.exec.CometExec3_4PlusSuite
+              org.apache.comet.exec.CometExecCompatibilitySuite
```
kazuyukitanimura, line 389: Is this used?

manuzhang: Yes, `CometExecCompatibilitySuite` is renamed from `CometExec3_4PlusSuite` and used by all supported Spark versions.

manuzhang:
@andygrove @coderfender @comphead I'd appreciate your review here.

coderfender:
Sure! Thank you for the pr @manuzhang

comphead (commented):
Thanks @manuzhang 

this is def on the table, however lets keep this PR opened for now. 

There are some pending use cases still using 3.4.3, I would propose doing this in parts. 
Currently it would make sense to: 
- remove 3.4 from CI
- Add a warn message saying 3.4 support would be dropped in 1.1.0
- update roadmap

manuzhang:
@comphead It's already announced in [1.0.0 release blog](https://datafusion.apache.org/blog/output/2026/08/07/datafusion-comet-1.0.0/).

> Apache Spark 3.4 is deprecated and scheduled for removal in Comet 1.1.0.

Where else do we need to add a warn message and update?

comphead:
> @comphead It's already announced in [1.0.0 release blog](https://datafusion.apache.org/blog/output/2026/08/07/datafusion-comet-1.0.0/).
> 
> > Apache Spark 3.4 is deprecated and scheduled for removal in Comet 1.1.0.
> 
> Where else do we need to add a warn message and update?

I was thinking to add a warn runtime message that Spark 3.4 support is sunsetting. Not sure if it makes sense though.

But still I would propose doing a soft drop:
- disable CI first to save resources 

Keep the code for couple of more weeks and drop it closer to the dedicated release date

comphead (changes requested):

parthchandra:
What does the community think about retaining the code for 3.4 (but maybe relaxed ci and backporting of fixes)? We have a ton of users (and even more production jobs) still on 3.4.

spark/src/main/spark-3.x/org/apache/comet/shims/CometCollectShim.scala
```diff
@@ -22,7 +22,7 @@ package org.apache.comet.shims
 import org.apache.spark.sql.catalyst.expressions.aggregate.{CollectList, CollectSet}
 
 /**
- * Shim for the `ignoreNulls` flag on `CollectList` / `CollectSet`. Spark 3.4 through 4.1 have no
+ * Shim for the `ignoreNulls` flag on `CollectList` / `CollectSet`. Spark 3.5 through 4.1 have no
```
kazuyukitanimura, line 25: Does it make sense to move `spark-3.x` to `spark-3.5`?

hsiang-c:
I'd like to have extended Spark 3.4 support for 2 quarters b/c lots of users are still with Spark 3.4.

manuzhang:
> I'd like to have extended Spark 3.4 support for 2 quarters b/c lots of users are still with Spark 3.4.

3.4 is still on [branch-1.0](https://github.com/apache/datafusion-comet/tree/branch-1.0) and bug fixes can be released on 1.0.x.

> Keep the code for couple of more weeks and drop it closer to the dedicated release date

This doesn't make sense on the main branch since codes committed will never be shipped.

comphead:
> This doesn't make sense on the main branch since codes committed will never be shipped.

Thanks @manuzhang It's more than that, we focusing on stopping Spark 3.4 support soon, however grace period may take some time. Comet evolving fast and there are still bunch of scenarios still using 3.4. I'm afraid back porting major features to 3.4 would take much more efforts than just supporting it for now. 

Do you see any risks in having 3.4 during following weeks? 

We can keep this PR opened and get back to it as soon as we can, WDYT?

manuzhang:
@comphead 
> I'm afraid back porting major features to 3.4 would take much more efforts than just supporting it for now.

Which major features do we need to land on 3.4? Any open issues and pull requests? 

> We can keep this PR opened and get back to it as soon as we can, WDYT?

No problem, but if we decide to add more features to 3.4, then we shouldn't drop the support now (for 1.1.0). We can't have both.

comphead:
I would expect to see #4837 #4547 #5314 and some other things currently in progress like shuffle writes are slow for nested schemas, etc

spark/src/main/spark-3.x/org/apache/comet/shims/CometCollectShim.scala
manuzhang: I don't think it's necessary. Just moving files around doesn't add much value.

kazuyukitanimura: I would say this is a great opportunity to clean up majorVerSrc in pom, not only moving files

andygrove (commented):
I went through the mechanical side of this closely: the `spark-3.4` Maven profile and shim directories, `dev/diffs/3.4.3.diff`, the CI matrices in `ci.yml`, `pr_build_linux.yml`, and `pr_build_macos.yml`, `dev/ci/compute-changes.py`, and the user-facing docs including `installation.md`, `spark-versions.md`, and `versioning_policy.md`. It all holds together. `isSpark35Plus` is deleted rather than left behind as an always-true guard, `CometSpark34AqeDppFallbackRule` is gone along with every call site and tag field that referenced it, and the now-unreachable `classOf[Base64] -> CometBase64` mapping in `QueryPlanSerde` is cleaned up since no supported Spark version constructs a plain `Base64` node anymore. Only `3.4.3.diff` was touched under `dev/diffs/`, none of the surviving diffs were hand-edited. I didn't find anything removed here that Spark 3.5, 4.0, 4.1, or 4.2 still depends on.

On the thread with @comphead: the three items cited as reasons to hold off, #4837, #4547, and #5314, are all closed or merged now. Does that change things, or is the concern more about the general rate of feature work still landing on main before 3.4 support closes out? That review is still sitting as changes requested, and I'd want to see it explicitly cleared before this merges rather than left open.

The branch is also showing as conflicting with main right now, so it will need a rebase before it can land regardless of how the above gets resolved.

manuzhang:
@andygrove I don't think we can drop Spark 3.4 support now, since those efforts will be otherwise wasted.

andygrove:
> @andygrove I don't think we can drop Spark 3.4 support now, since those efforts will be otherwise wasted.

Sorry, that was an autoamted review. I agree that we can't drop the support yet. We should update the docs to remove the claim that support will be dropped in 1.1.0 and instead say that it will be dropped in a future release

andygrove:
Thanks for the rebase; the branch is on current `main` now and merges cleanly. But your own reply is the thing to settle before any of the mechanical review matters.

> I don't think we can drop Spark 3.4 support now, since those efforts will be otherwise wasted.

If that is where you have landed, I would rather this sat as a draft, or closed with an issue tracking the removal, than stayed open as a rebase-and-conflict treadmill. It touches the Maven profiles, the shim directories, `dev/diffs/3.4.3.diff`, three CI matrices and `compute-changes.py`, so it will conflict with almost anything that lands, and someone will keep having to rebase it for a decision that has not been made. Which would you prefer? If the answer is "hold", I am happy to convert it to a draft myself.

On the substance, since it is worth recording either way: the mechanical work is sound. I went through it in detail and found nothing removed that 3.5, 4.0, 4.1 or 4.2 still needs. `isSpark35Plus` is deleted rather than left as an always-true guard, `CometSpark34AqeDppFallbackRule` is gone with all its call sites and tag fields, the unreachable `classOf[Base64] -> CometBase64` mapping is cleaned up, and only `3.4.3.diff` was touched under `dev/diffs/` with no hand-editing of the surviving diffs. If the project decides to drop 3.4 later, this is the change to pick up rather than redo.

The three items @comphead cited as reasons to hold are all resolved now (#4837 closed, #4547 closed, #5314 merged), and their `CHANGES_REQUESTED` from 19 August is still the blocking review. @comphead, whichever way the timing question goes, could you clear or restate that? At the moment the PR is blocked on a review whose stated reasons no longer apply, which makes it hard to tell whether the remaining objection is those specific items or the broader "not yet".

My own view on the timing, for the record: I do not think the argument that in-flight 3.4 work would be wasted is a reason to keep 3.4 indefinitely, but it is a reason not to drop it in the middle of that work. So a tracked deferral with a target version seems right, rather than either merging now or closing without a plan.

PR #5359, opened 2026-08-14
fix: guard against silent fail_on_error loss in scalar wiring (#5074)

### LINKED ISSUE #5074: Guard against silent fail_on_error loss in scalar function wiring

The scalar-function wiring makes it easy to silently lose ANSI semantics, and it has already happened once (`MakeDecimal`). Two independent mechanisms both fail "open" toward non-ANSI behavior:

1. `CometScalarFunction.convert` uses the plain `scalarFunctionExprToProto` helper, which hardcodes `builder.setFailOnError(false)` (`spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala:1070-1075`). Any ANSI-sensitive Spark expression wired via plain `CometScalarFunction` drops the flag at serialization time.
2. On the native side, `create_comet_physical_fun` threads `fail_on_error` only into the hand-written match arms; the catch-all `_ => registry.udf(fun_name)` (`native/spark-expr/src/comet_scalar_funcs.rs:240`) drops it for every registry-resolved UDF, which includes all functions wired in from the datafusion-spark crate. `create_comet_physical_fun_with_eval_mode` also defaults `fail_on_error.unwrap_or(false)`.

Today no wired datafusion-spark function is ANSI-sensitive (verified against Spark 4.1.1: date_add, date_sub, last_day, from_utc_timestamp, to_utc_timestamp, factorial, chr, substring, space, luhn_check, url_encode, hex, rint, etc. never consult ANSI), so there is no live gap besides `MakeDecimal` (tracked separately). But the failure mode is silent and will recur as more expressions are wired.

Options, not mutually exclusive:

- Make `scalarFunctionExprToProtoWithReturnType` (the only helper that can set `fail_on_error`) reject or assert when the native registration cannot receive the flag, or conversely make the native side return an error when `fail_on_error=Some(true)` reaches the registry catch-all instead of silently ignoring it.
- Add a debug assertion or serde-level check: expressions whose Spark class has a `failOnError`/`evalMode` field must not be registered via plain `CometScalarFunction`.
- Document the constraint in the contributor guide for wiring datafusion-spark functions.

### Additional context

Found by an ANSI-mode audit of all native expressions against Spark 4.1.1. The audit relied on name-based variant selection (`parse_url` vs `try_parse_url`, `url_decode` vs `try_url_decode`) as the sound pattern for ANSI-sensitive scalar functions; a guard would enforce that pattern.

sam-1112:

Closes #5074

## Rationale for this change

Registry-resolved UDFs (including datafusion-spark) previously ignored `fail_on_error=true`, which could silently change ANSI semantics. Plain `CometScalarFunction` also always serialized `fail_on_error=false`, so ANSI-sensitive Spark expressions could be miswired without failing at planning time.

- **Native**: `create_comet_physical_fun_with_eval_mode` fails closed when `fail_on_error=true` would fall through to registry lookup.
- **Scala serde**: `CometScalarFunction.convert` rejects expressions with `failOnError`, `evalMode`, or `nullOnOverflow` fields; adds
  `isAnsiSensitive` helpers for registration audits.
- **Docs**: document ANSI / `fail_on_error` wiring constraints in the contributor guide.
- **Tests**: Rust integration tests in `test_udf_registration`; Scala `CometScalarFunctionSuite` for serde guards and audit.

- `cargo test -p datafusion-comet-spark-expr --test test_udf_registration -- --nocapture`
- `./mvnw test -Dtest=none -Dsuites="org.apache.comet.serde.CometScalarFunctionSuite" -Pspark-4.1`

### DIFF
--- a/native/spark-expr/src/comet_scalar_funcs.rs
+++ b/native/spark-expr/src/comet_scalar_funcs.rs
@@ -260,6 +260,13 @@ pub fn create_comet_physical_fun_with_eval_mode(
             let func = Arc::new(crate::string_funcs::spark_levenshtein);
             make_comet_scalar_udf!("levenshtein", func, without data_type)
         }
+        // Registry UDFs (including datafusion-spark) cannot receive fail_on_error.
+        _ if fail_on_error => Err(DataFusionError::Execution(format!(
+            "Function '{fun_name}' is resolved from the UDF registry and cannot \
+             honor fail_on_error=true. Use a name-based ANSI/try variant \
+             (e.g. parse_url / try_parse_url) or a dedicated match arm that \
+             consumes the flag."
+        ))),
         _ => registry.udf(fun_name).map_err(|e| {
             DataFusionError::Execution(format!(
                 "Function {fun_name} not found in the registry: {e}",
--- a/native/spark-expr/src/comet_scalar_funcs.rs
+++ b/native/spark-expr/src/comet_scalar_funcs.rs
@@ -260,6 +260,10 @@ pub fn create_comet_physical_fun_with_eval_mode(
             let func = Arc::new(crate::string_funcs::spark_levenshtein);
             make_comet_scalar_udf!("levenshtein", func, without data_type)
         }
+        // Spark 4.1+ serde always sets fail_on_error=true (always-throw semantics).
+        // SparkMakeTime already throws on invalid input, so accept the flag here rather
+        // than falling through to the registry fail-closed path.
+        "make_time" => Ok(Arc::new(ScalarUDF::new_from_impl(SparkMakeTime::new()))),
         // Registry UDFs (including datafusion-spark) cannot receive fail_on_error.
         _ if fail_on_error => Err(DataFusionError::Execution(format!(
             "Function '{fun_name}' is resolved from the UDF registry and cannot \
--- a/spark/src/main/scala/org/apache/comet/serde/CometScalarFunction.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/CometScalarFunction.scala
@@ -31,8 +31,8 @@ case class CometScalarFunction[T <: Expression](name: String) extends CometExpre
     if (CometScalarFunction.isAnsiSensitive(expr)) {
       withFallbackReason(
         expr,
-        s"${expr.nodeName} carries failOnError/evalMode/nullOnOverflow and cannot use " +
-          s"CometScalarFunction('$name'). Prefer name-based ANSI/try variants " +
+        s"${expr.nodeName} carries failOnError/evalMode/nullOnOverflow/ansiEnabled/evalContext " +
+          s"and cannot use CometScalarFunction('$name'). Prefer name-based ANSI/try variants " +
           "(e.g. parse_url / try_parse_url), or a custom serde with " +
           "scalarFunctionExprToProtoWithReturnType plus a native match arm that " +
           "consumes fail_on_error.")
@@ -46,21 +46,21 @@ case class CometScalarFunction[T <: Expression](name: String) extends CometExpre
 
 object CometScalarFunction {
 
-  /** Product field names that indicate ANSI / eval-mode sensitive Spark expressions. */
+  /**
+   * Product / Java field names that indicate ANSI / eval-mode sensitive Spark expressions.
+   *
+   * Detection uses Java reflection (not `Product.productElementNames`) so the check compiles on
+   * Scala 2.12 used by Spark 3.4 / 3.5.
+   */
   private val AnsiSensitiveFields: Set[String] =
-    Set("failOnError", "evalMode", "nullOnOverflow")
+    Set("failOnError", "evalMode", "nullOnOverflow", "ansiEnabled", "evalContext")
 
   /**
    * True when the Spark expression case class declares an ANSI-related constructor field. Used to
    * reject miswiring via plain [[CometScalarFunction]].
    */
   private[serde] def isAnsiSensitive(expr: Expression): Boolean = {
-    expr match {
-      case p: Product =>
-        p.productElementNames.exists(AnsiSensitiveFields.contains)
-      case _ =>
-        isAnsiSensitive(expr.getClass)
-    }
+    isAnsiSensitive(expr.getClass)
   }
 
   /**
@@ -68,6 +68,13 @@ object CometScalarFunction {
    * ANSI-related field regardless of any particular instance's flag value.
    */
   private[serde] def isAnsiSensitive(clazz: Class[_]): Boolean = {
-    clazz.getDeclaredFields.exists(f => AnsiSensitiveFields.contains(f.getName))
+    var current: Class[_] = clazz
+    while (current != null && current != classOf[Object]) {
+      if (current.getDeclaredFields.exists(f => AnsiSensitiveFields.contains(f.getName))) {
+        return true
+      }
+      current = current.getSuperclass
+    }
+    false
   }
 }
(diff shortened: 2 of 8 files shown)


spark/src/test/scala/org/apache/comet/serde/CometScalarFunctionSuite.scala
```diff
@@ -0,0 +1,166 @@
...
+    with Unevaluable {
+  override def children: Seq[Expression] = Seq(child)
+  override def nullable: Boolean = child.nullable
+  override def dataType: DataType = IntegerType
+  override protected def withNewChildrenInternal(
+      newChildren: IndexedSeq[Expression]): Expression =
+    copy(child = newChildren.head)
+}
+
+class CometScalarFunctionSuite extends CometTestBase {
```
sunchao, line 88: [P1] Register the new suite in both CI workflow matrices

`CometScalarFunctionSuite` is not listed in either `.github/workflows/pr_build_linux.yml` or `.github/workflows/pr_build_macos.yml`. The mandatory preflight runs `python3 dev/ci/check-suites.py`, which requires every `*Suite.scala` in both workflows; on this head it exits 255 with `Suite not found in workflow .github/workflows/pr_build_linux.yml: org.apache.comet.serde.CometScalarFunctionSuite`. Please add the suite to the `expressions` bucket in both workflows; otherwise all downstream CI jobs are blocked.

spark/src/main/scala/org/apache/comet/serde/CometScalarFunction.scala
```diff
@@ -21,14 +21,53 @@ package org.apache.comet.serde
...
+    Set("failOnError", "evalMode", "nullOnOverflow")
+
+  /**
+   * True when the Spark expression case class declares an ANSI-related constructor field. Used to
+   * reject miswiring via plain [[CometScalarFunction]].
+   */
+  private[serde] def isAnsiSensitive(expr: Expression): Boolean = {
+    expr match {
+      case p: Product =>
+        p.productElementNames.exists(AnsiSensitiveFields.contains)
```
sunchao: [P1] Keep ANSI-field detection compatible with Scala 2.12

`scala.Product.productElementNames` exists in Scala 2.13 but not Scala 2.12. The supported Spark 3.4 and 3.5 profiles include Scala 2.12, so their lint/build matrix entries cannot compile this source: `value productElementNames is not a member of Product`. I reproduced the failure with Scala 2.12.18 and confirmed the same expression compiles with Scala 2.13.16. Please use a Scala-2.12-compatible way to inspect these fields, or introduce version-specific shims.

sunchao (commented):
Three verified P1 issues: mandatory CI suite registration, Scala 2.12 source compatibility, and existing Spark 4.1+ `make_time` native planning.

native/spark-expr/src/comet_scalar_funcs.rs
```diff
@@ -260,6 +260,13 @@ pub fn create_comet_physical_fun_with_eval_mode(
             let func = Arc::new(crate::string_funcs::spark_levenshtein);
             make_comet_scalar_udf!("levenshtein", func, without data_type)
         }
+        // Registry UDFs (including datafusion-spark) cannot receive fail_on_error.
+        _ if fail_on_error => Err(DataFusionError::Execution(format!(
```
sunchao, line 268: [P1] Preserve existing `make_time` registry dispatch before rejecting `fail_on_error=true`

The Spark 4.1+ shim already calls `scalarFunctionExprToProtoWithReturnType("make_time", s.dataType, true, ...)`, but `SparkMakeTime` is registered only in `all_scalar_functions()` and has no dedicated match arm. This branch therefore rejects every nonconstant `make_time` query during native planning, including valid inputs. `SparkMakeTime` already implements Spark's always-throw semantics correctly, so please add an explicit `"make_time"` match arm or a narrowly safe exemption, and cover `create_comet_physical_fun("make_time", ..., Some(true))` in the regression test.

spark/src/main/scala/org/apache/comet/serde/CometScalarFunction.scala
```diff
@@ -21,14 +21,53 @@ package org.apache.comet.serde
...
     val optExpr = scalarFunctionExprToProto(name, childExpr: _*)
     optExpr
   }
 }
+
+object CometScalarFunction {
+
+  /** Product field names that indicate ANSI / eval-mode sensitive Spark expressions. */
+  private val AnsiSensitiveFields: Set[String] =
+    Set("failOnError", "evalMode", "nullOnOverflow")
```
sunchao: [P2] Recognize `ansiEnabled` and Spark 4.1+ `evalContext`

Could we include `ansiEnabled` and `evalContext` in `AnsiSensitiveFields`? Spark `Round`, `BRound`, and `Conv` store ANSI behavior in `ansiEnabled`, while Spark 4.1+ `Add`, `Subtract`, `Multiply`, `Divide`, and related arithmetic expressions store `NumericEvalContext` as `evalContext` and expose `evalMode` only through an inherited method. Neither marker is present in this set, so both `isAnsiSensitive(expr)` and `isAnsiSensitive(clazz)` return false. A plain `CometScalarFunction` registration can therefore silently serialize `fail_on_error=false` and still pass the registration audit. The current handlers for these expressions are specialized, so this is a gap in the proposed safeguard rather than a regression in existing registrations. Could we recognize both fields and add regression coverage using real Spark `Round` and Spark 4.1+ `Add` expressions?

spark/src/test/scala/org/apache/comet/serde/CometScalarFunctionSuite.scala
sam-1112: Fixed. I registered `org.apache.comet.serde.CometScalarFunctionSuite` in the expressions bucket of both `.github/workflows/pr_build_linux.yml` and `.github/workflows/pr_build_macos.yml`.

Verified locally with `python3.12 dev/ci/check-suites.py`; the suite is now found in both workflow matrices.

spark/src/main/scala/org/apache/comet/serde/CometScalarFunction.scala
sam-1112: Fixed. I removed `Product.productElementNames` and now detect ANSI-sensitive fields with Java reflection (`Class.getDeclaredFields`, walking superclasses). That compiles on both Scala 2.12 and 2.13. Verified with `./mvnw -pl spark -am compile -DskipTests -Pspark-3.4` using JDK 11.

native/spark-expr/src/comet_scalar_funcs.rs
sam-1112: Fixed. I added an explicit `"make_time" `match arm before the registry fail-closed branch. `SparkMakeTime` already implements always-throw semantics, so the arm accepts `fail_on_error=true` without taking the flag as a constructor argument.

I also added regression coverage for `create_comet_physical_fun("make_time", ..., Some(true))`. Verified with `cargo test -p datafusion-comet-spark-expr --test test_udf_registration`; all 5 tests pass.

spark/src/main/scala/org/apache/comet/serde/CometScalarFunction.scala
sam-1112: Yeah, good catch. Both fields are in `AnsiSensitiveFields` now. `Round` / `BRound` / `Conv` keep ANSI in `ansiEnabled`, and Spark 4.1+ arithmetic uses `evalContext` instead of an `evalMode` field, so the old set would miss them and the audit would still pass. Coverage is a real Spark `Round` in `CometScalarFunctionSuite`, and a real Spark 4.1+ `Add(..., NumericEvalContext(...))` in `CometDecimalArithmeticViewSuite` (that constructor only exists on 4.1+).

sunchao (approved):
LGTM, cc @andygrove @mbutrovich

PR #5361, opened 2026-08-14
feat: implement native Iceberg V2 writer via iceberg-rust

### LINKED ISSUE #4322: Writes to Apache Iceberg Tables

### LINKED ISSUE #5308: feat: Write Iceberg Parquet data files natively via iceberg-rust (part 3 of 3)

This is the third and final part of #4322 (accelerating Iceberg V2 write operations), which was split into three deliverables:

1. **Split-operator plan** — #4658 (merged). Iceberg's single V2 write command is rewritten into a writer operator (inside AQE) and a committer operator, with file writing still running through iceberg-java.
2. **Native-write eligibility detection** — #5298. An allowlist over the write's effective configuration decides whether a native write would produce the same outcome as iceberg-java, and reports a fall-back reason via extended EXPLAIN when it would not. The supported configuration and the accepted divergences are documented in `iceberg-writes.md`.

This issue tracks part 3: actually performing the per-task Parquet data-file write via [iceberg-rust](https://github.com/apache/iceberg-rust) when detection accepts the write.

High-level approach:

- **Conversion**: `CometIcebergNativeWrite.convert` (currently a stub) produces a native write operator. The JVM resolves all write settings and passes them to the native side via protobuf; the vetted `write.parquet.*` configuration (compression codec/level, row-group/page/dictionary sizes, page row limit) is translated into parquet-rs `WriterProperties`. Footer statistics are written untruncated to match parquet-mr.
- **Writer modes**: unpartitioned, clustered, and fanout writers, mirroring iceberg-java's writer selection.
- **Transport**: the native side serialises the resulting `DataFile`s as an in-memory Iceberg manifest via iceberg-rust's `ManifestWriter`; the JVM decodes the bytes with Iceberg's own `ManifestFiles.read`. Both ends use their library's spec codec, so no hand-rolled Avro.
- **Commit**: unchanged from part 1 — the recovered `DataFile[]` goes into `SparkWrite$TaskCommit` and `BatchWrite.commit`, preserving Iceberg's overwrite semantics, copy-on-write validation, WAP, and snapshot summaries.
- **Manifest metrics**: assembled on the JVM at commit time using Iceberg's own `MetricsConfig` / `ParquetUtil` code against the written file's footer (plus NaN counts from the native writer, which the footer cannot provide). Java makes every metadata decision — metrics modes, the inferred-column cap, bound truncation, list/map bounds suppression — so the `counts` / `none` metrics restrictions from part 2 can be lifted.
- **Testing**: parity tests comparing native output against iceberg-java per data type (including a targeted float/double NaN metrics test), plus end-to-end tests showing compaction / sort / z-order rewrites running fully native.

The guiding rule from part 2 carries over: if the native write cannot reproduce iceberg-java's outcome for the effective configuration, the write falls back — the documented accepted divergences (row-group/file-rolling cadence, compressed page bytes, footer key-value metadata, partition-path escaping) are the only exceptions.

## Additional context

Detection (#5298) is deliberately conservative so this part can land incrementally: individual restrictions can be relaxed as the native writer proves each behaviour byte- or semantics-identical.

jordepic:

(Third of three) Closes #4322 and closes #5308. Part one was #4658 (split-operator plan, merged); part two was #5298 (detection + fall-back reasons, merged).

## Rationale for this change

We want to speed up writing iceberg data for ETL jobs and table maintenance!

With detection in place (#5298), this PR wires in the native writer itself: when `spark.comet.iceberg.write.enabled=true` and a write passes the eligibility gate, the `IcebergWrite` operator's per-task Parquet write is delegated to iceberg-rust through Comet's native execution pipeline. Metadata writing (this diverges from my original approach where I had rust responsible for writing it all) and commit semantics are unchanged and go through the JVM, ensuring identical semantics to Java.

## Performance

`INSERT INTO ... SELECT * FROM parquet` end-to-end (read + write), local[*], Spark 4.0.1 / Iceberg 1.10.0, 2M rows x 100 mixed-type columns (~1.5 GB parquet source, 16 files), median of 3 iterations. Scenario A is stock Spark, B adds the Comet scan only, C adds the native Iceberg write (A→B isolates the scan speedup; B→C isolates the writer):

| table | A: Spark | B: Comet read | C: Comet read + native write | B→C |
| unpartitioned | 10.4 s | 9.5 s | 3.3 s | **2.9x** |
| partitioned `bucket(8, int)`, fanout | 22.9 s | 24.1 s | 4.4 s | **5.5x** |

The partitioned run uses `write.distribution-mode=none` + `write.spark.fanout.enabled=true` so all three scenarios share one plan shape. With the default hash distribution, the exchange feeding the write partitions by Iceberg's `BucketFunction` SQL function, which Comet does not execute natively — the shuffle stays on the JVM and the eligibility gate declines the native write (the plan under the write must be fully Comet-native), so today's default-distribution bucketed append falls back with a reported reason rather than accelerating.

The harness (data generator + spark-submit runner) is in the `bench/` directory of my fork if anyone wants to reproduce.

High level design:
- Java uses reflection to unpack writer settings for iceberg
- If those settings are achievable in iceberg-rust, it accelerates the write
- Pass a protobuf to rust with relevant writer settings, object store settings, table path, etc
- Iceberg-rs does the data file writing, passes back an "in-memory" iceberg manifest file via JNI (iceberg rust has similar writer logic to Java, allowing it to provide similar behavior across writer settings like when to split files)
- Java takes the in-memory manifest, uses its own writer settings to convert the information (useful for column stats), uses existing iceberg spark logic for committing

Where iceberg-java and iceberg-rust can't be byte identical, `iceberg-writes.md` now classifies every divergence by blast radius: differences confined to the physical bytes of a data file (row-group cadence, footer metadata, compressed bytes — cosmetic, no reader decision is based on them) vs differences visible in manifest metadata that later readers prune on. The manifest-visible class is deliberately almost empty because `DataFile` metrics are re-derived on the JVM from each written file's parquet footer through iceberg-java's own `ParquetUtil.footerMetrics` + `MetricsConfig.forTable` — the two residual footer-vs-writer-tracked differences (zero-sign bound normalization, Iceberg 1.10+ float/double counts under a nullable struct) are each analyzed safe for pruning in the doc.

The doc also gained a "Failure handling" section: eligibility (including eager resolution of the whole executor-side reflection surface) is decided at plan time; a mid-write task failure retries through the native writer with attempt-unique file names; partial results are never committed; files from failed attempts are reclaimed by `remove_orphan_files`; commit failures reuse `SparkWrite.abort` unchanged.

- `CometIcebergWriteActionSuite` gains a "native acceleration" section (~20 new tests): appends (unpartitioned/partitioned/fanout with unsorted input), overwrite static/dynamic, CoW DELETE/UPDATE natively and a pin that CoW MERGE intentionally falls back (`MergeRowsExec` is not Comet-native; the pin references #5318, which is adding it), CTAS/RTAS on Spark 3.5+ (pinning the empty-metadata-location path), complex types with nested field IDs, multi-file rolling via `write.target-file-size-bytes`, empty appends (one commit, zero files), and `sort_order_id` parity with the JVM writer (0 through Iceberg 1.10; the resolved order id on 1.11+).
- Manifest-metadata parity tests write identical data through the native and JVM paths and compare the committed manifest metrics: NaN float/double counts and bounds (incl. the all-NaN bounds-drop), string bound truncation under the default `truncate(16)`, metrics mode `none` with a per-column `full` override, and mode `counts` with `WRITE ORDERED BY` (pinning the sorted-column promotion inside `MetricsConfig.forTable`).
- Rewrite actions (`CometIcebergRewriteActionSuite`) now assert compaction data files were physically written by iceberg-rust via the parquet footer `created_by` signature.
- Rust unit + integration tests for the writer translation, manifest round-trip, partitioning writers (including a regression test for a partition spanning a batch boundary under the clustered writer), and file naming.
- `IcebergReflectionSuite` pins that the executor-side reflection surface (manifest decode, footer-metrics rebuild, `TaskCommit` construction) resolves eagerly against the linked Iceberg, so a version bump that moves any of it fails CI loudly instead of silently falling back.
- All suites pass locally on all four profiles: spark-3.4 (Iceberg 1.5.2, Scala 2.12), spark-3.5 (1.8.1), spark-4.0 (1.10.0), spark-4.1 (1.11.0) — the version spread matters because the metrics rebuild reflects version-specific Iceberg internals (e.g. Iceberg 1.10+ has two 6-arg `FieldMetrics` constructors, and 1.11 changed what `sort_order_id` the Spark writer stamps).


jordepic:
@andygrove here is 3/3! I'd greatly appreciate a CI kick. cc @parthchandra , @comphead , @mbutrovich .

@unikdahal - I have addressed your comments regarding CTAS and RTAS in this one, thank you for helping to validate this feature!!

jordepic:
Sigh..sorry @andygrove I needed to cargo fmt, feel free to kick off the CI whenever

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
```diff
@@ -163,12 +159,12 @@ object CometIcebergNativeWrite extends CometOperatorSerde[IcebergWriteExec] with
       "custom location provider unsupported"),
     requireFormatVersionAtMostTwo,
     requireNoEncryptionPrefix,
-    requireSupportedMetricsModes,
     requireNoBloomFilterColumnsEnabled,
     requireRowGroupCheckMinRecordCountAtDefault,
     requireRowGroupCheckMaxRecordCountAtDefault,
     requireParquetPageVersionDefault,
     requireShredVariantsDisabled,
+    requireParseableCompressionLevel,
```
andygrove: I think a table with a `uuid` column would be judged eligible here and then fail the task rather than falling back.

Iceberg's `TypeToSparkType` maps `uuid` to `StringType`, so Comet hands the native writer a Utf8 column, while `schema_to_arrow_schema` makes the target `FixedSizeBinary(16)`. Since `decorate_batch_with_field_ids` casts with `safe: false`, a 36-character UUID string has nowhere to go. The gate is entirely table-property based with no schema-type check, so detection would return `Compatible` first and the failure would surface at execution time.

Have you tried a table with a `uuid` column? If it does fail, a trigger rule that declines on column types the native writer cannot reproduce (plus a detection test pinning it) would keep this a fallback rather than a task error. `fixed(N)` should be fine via `Binary -> FixedSizeBinary` as long as every value is exactly N bytes, but it would be good to have that pinned too.

spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
```diff
@@ -569,6 +569,569 @@ class CometIcebergWriteActionSuite
...
+          |       SELECT 3 AS id, 'eu' AS region, 30.0 AS amount) s
+          |ON t.id = s.id
+          |WHEN MATCHED THEN UPDATE SET t.amount = s.amount
+          |WHEN NOT MATCHED THEN INSERT (id, region, amount) VALUES (s.id, s.region, s.amount)
+          |""".stripMargin)
+      }
+    }
+  }
+
+  test("native acceleration: complex types (struct, array, map) round-trip with field IDs") {
```
andygrove, line 756: The native-acceleration tests all write `(id INT, region STRING, amount DOUBLE)`, and this complex-types test adds STRING and INT leaves, so the type surface actually exercised through the native writer is fairly narrow.

Would you mind adding a round-trip parity test over the wider primitive set: `DATE`, `TIMESTAMP`, `TIMESTAMP_NTZ`, `DECIMAL` at a couple of precisions, `BINARY`, `BOOLEAN`, `BIGINT`, `FLOAT`? The one I would most like pinned is timestamps under a non-UTC `spark.sql.session.timeZone`, with both `timestamp` and `timestamptz` columns. That is exactly where a silent value shift would hide, and none of the current tests would notice it. Your `manifestMetricsParity` helper looks like it would adapt nicely to compare row values as well as metrics.

native/core/src/execution/operators/iceberg_write.rs
```diff
@@ -0,0 +1,1111 @@
...
+            }
+            InnerWriter::Clustered(w) => {
+                let mut parts = splitter
+                    .expect("partition splitter must be Some for partitioned writes")
+                    .split(&batch)
+                    .map_err(iceberg_err)?;
+                let (calculator, partition_type) = clustered_order
+                    .expect("clustered order helper must be Some for clustered writes");
+                let order = partition_first_occurrence_order(&batch, calculator, partition_type)?;
+                parts.sort_by_key(|(key, _)| order.get(key.data()).copied().unwrap_or(usize::MAX));
```
andygrove: This computes the partition values a second time. `partition_first_occurrence_order` runs `PartitionValueCalculator::calculate` over the whole batch, but `RecordBatchPartitionSplitter::split` (two lines up) has already evaluated the same partition transforms internally. Clustered is the default for partitioned tables, so this doubles the partition-transform cost on the common path.

Is there a way to recover the ordering without the second pass? The input is partition-sorted by construction, so anything that gives each part's first row index would do. Alternatively, a single `calculate` whose result feeds both the split and the ordering, if iceberg-rust exposes that shape.

spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
```diff
@@ -1124,6 +1280,464 @@ object IcebergReflection extends Logging {
...
+  }
+
+  /**
+   * Stamp `sortOrderId` on every `DataFile` in `dataFiles`. iceberg-rust's writer doesn't expose
+   * the field (it's `pub(crate)` on `DataFile`), so the manifest comes back with sort_order_id
+   * unset. iceberg-java's `BaseFile` declares a private `sortOrderId: Integer` field that the
+   * normal `SparkWrite` path populates from the table's `outputSortOrderId`; we mirror that here
+   * by writing the same value via reflection before handing files to the committer.
+   */
+  def stampSortOrderId(dataFiles: java.util.List[_], sortOrderId: Int): Unit = {
```
andygrove: This is a step beyond the rest of the reflection in this file, which reads fields and calls package-private constructors but does not mutate iceberg-java state. An uncaught `NoSuchFieldException` here would also be a task failure rather than a fallback, unlike `newDataManifestFile` which soft-fails on exactly that.

`DataFiles.Builder.withSortOrder(SortOrder)` is public, and `rebuildDataFilesWithJavaMetrics` already builds through that builder immediately afterwards. Could the sort order be applied there instead, by resolving the `SortOrder` on the driver (it is `Serializable`) and shipping it into the task closure alongside `metricsConfig`? That would drop the private-field write entirely. If you would rather keep the current shape, could it at least soft-fail on `NoSuchFieldException` for consistency?

native/core/src/execution/operators/iceberg_write.rs
```diff
@@ -0,0 +1,1111 @@
...
+            common: Arc::new(common),
+            iceberg_schema,
+            partition_spec,
+            writer_mode,
+            writer_properties: Arc::new(writer_properties),
+            partition_id,
+            task_attempt_id,
+            output_schema,
+            plan_properties,
+            metrics: ExecutionPlanMetricsSet::new(),
```
andygrove, line 131: I do not see anything writing to this `ExecutionPlanMetricsSet`, so `metrics()` always returns an empty set, and `CometIcebergWriteExec` passes `CometMetricNode(metrics, Nil)` with no child nodes either. That leaves no native-side visibility at all for the writer.

Some counters would be valuable here: time spent inside the iceberg-rust writer stack, rows and bytes handed to parquet-rs, files rolled. If you would rather defer that, dropping the field for now would be clearer than keeping a metrics set that never reports anything.

spark/src/main/scala/org/apache/spark/sql/comet/CometMetricNode.scala
```diff
@@ -354,10 +354,21 @@ object CometMetricNode {
...
+   * final val`, which can be `null` if `SparkSession.getActiveSession` returned `None` at the
+   * moment AQE's stage-finalisation rules built it. Forcing such a node's `metrics` lazy val NPEs
+   * in `SQLMetrics.createMetric(sparkContext, ...)`. We don't own those metrics anyway -- the
+   * native side only sources updates against operators it actually planned, and JVM-side
+   * AQE-stage nodes belong to a different stage that's already been materialised independently.
    */
-  def fromCometPlan(cometPlan: SparkPlan): CometMetricNode = {
-    val children = cometPlan.children.map(fromCometPlan)
-    CometMetricNode(cometPlan.metrics, children)
+  def fromCometPlan(cometPlan: SparkPlan): CometMetricNode = cometPlan match {
```
andygrove: The AQE reasoning in the doc comment makes sense, but this changes behaviour for every caller, not just the Iceberg write: `CometNativeScanExec`, `CometShuffleExchangeExec`, `CometNativeWriteExec` and `operators.scala` all come through here, and a whole subtree now silently reports `Map.empty` where it previously would have contributed metrics.

I went looking and could not find a non-`CometPlan` node living inside a native block today, so I think it is safe in practice. Could you either narrow the guard to the node type that actually triggers this (or guard the `metrics` access itself), or add a test pinning that no existing operator's metrics regress? A silent metric drop is the kind of thing that goes unnoticed for a long time.

native/core/src/execution/operators/iceberg_write.rs
```diff
@@ -0,0 +1,1111 @@
...
+    partition_spec: PartitionSpecRef,
+    partition_id: Option<i32>,
+    task_attempt_id: Option<i64>,
+    operation_id: &str,
+) -> DFResult<Vec<u8>> {
+    // The manifest is assembled entirely in-process via the `memory` scheme, so the credential
+    // dispatch key / access mode are inert here.
+    let memory_io = load_file_io(
+        &std::collections::HashMap::new(),
+        "memory:///",
```
andygrove, line 599: Small thing to confirm: does this `memory://` `FileIO` own its backing store, so it is dropped when the function returns, or is opendal's memory service process-global? If the store is shared, every task's manifest bytes would stay resident for the lifetime of the executor, since nothing deletes the entry after the `read()` below.

andygrove (commented):
Thanks for pushing this all the way through, and for splitting it into three reviewable parts. The JVM-side metrics rebuild is a really nice idea: re-deriving each `DataFile`'s metrics through the version-matched `ParquetUtil.footerMetrics` + `MetricsConfig` makes manifest metadata iceberg-java's decision by construction rather than something we have to imitate, and it let you drop the `counts`/`none` gates entirely. The comments throughout are genuinely helpful, especially the ones recording *why* a piece of reflection is shaped the way it is.

A few things I verified while reading, recorded here so nobody re-does the work:

- `DataFiles.Builder.copy` does carry `sortOrderId` and `splitOffsets` across (`DataFiles.java:203` on 1.8.1), so the stamp-then-rebuild ordering works.
- `ParquetUtil.footerMetrics` falls back to footer stats for fields absent from `fieldMetrics`, and sources `nanValueCounts` only from `fieldMetrics`, which matches what `buildFloatFieldMetrics` synthesises.
- `ctx.properties` is `tableProperties ++ writeProperties`, so `requireParseableCompressionLevel` does see per-write option overrides.
- The `arrow` cast in `decorate_batch_with_field_ids` short-circuits when types already match, so it is not a per-batch copy.

## CI

Lint is red on both Linux and macOS. I reproduced it locally with `rustfmt --check`: it is `cargo fmt` on `native/core/src/execution/operators/iceberg_write.rs` in three spots.

- around line 406, the `parts.sort_by_key(...)` closure fits on one line
- around line 551, the `ManifestWriterBuilder::new(...)` call wants collapsing
- around line 773, a double blank line before the `integration` module comment

`make format` should clear all three. `iceberg_common.rs` and `iceberg-writes.md` are both already clean.

The rest of my comments are inline.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
jordepic: You're right — a `uuid` column passed detection and then failed the task inside `decorate_batch_with_field_ids` (Arrow has no `Utf8 -> FixedSizeBinary(16)` cast). Added a type-based trigger rule: detection now walks the write schema (nested fields included) and declines any `uuid` column with a fall-back reason. Pinned by a new detection test that evolves a `uuid` column onto a table through the Iceberg API (Spark DDL can't declare one) and asserts the write falls back and still succeeds through iceberg-java.

`fixed(N)` is pinned too: a new round-trip test writes exact-N-byte values through the native path and compares row values and `readable_metrics` bounds against a JVM-written twin — `Binary -> FixedSizeBinary(4)` casts fine as long as lengths match, and a wrong-length value fails the task the same way the JVM writer rejects it. I audited the remaining Iceberg types: `uuid` is the only Spark-writable type with a cast gap (`time` / `timestamp_ns` can't be written through Spark's Iceberg integration at all, and the V3-only types are behind the format-version gate).

spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
jordepic: Added `native acceleration: wide primitive types keep JVM-parity values and manifest metrics`: `BIGINT, BOOLEAN, FLOAT, DECIMAL(9,2), DECIMAL(38,10), DATE, TIMESTAMP, TIMESTAMP_NTZ, BINARY` written under `spark.sql.session.timeZone=America/New_York`, comparing both the full row sets and the aggregated `readable_metrics` bounds/counts against a JVM-written twin. On top of the twin parity there's an absolute check: the `timestamptz` value must read back as the same zoned instant it was written as and the ntz value must pass through untouched, so a shift that happened to hit both paths equally would still fail. The `fixed(N)` test from the other thread rides the same helper shape. Green on all four Spark/Iceberg profiles.

native/core/src/execution/operators/iceberg_write.rs
jordepic: Fixed — the clustered path no longer uses `RecordBatchPartitionSplitter` at all. A new `ClusteredBatchSplitter` runs `PartitionValueCalculator::calculate` once per batch, walks the literals for contiguous-run boundaries (preserving batch order by construction, so the sort and the first-occurrence map are gone too), and emits one part per run. Input that isn't actually clustered produces multiple runs with the same key and surfaces the same `ClusteredWriter` error as before.

One deliberate non-optimisation, recorded in a comment on `materialize_run`: the runs are materialised with `filter_record_batch` (what the splitter produced before) rather than zero-copy `RecordBatch::slice`. iceberg-rust's NaN-count visitor reads list/map children via `list_array.values()`, which ignores a slice's offset window, so sliced list-of-float columns would over-count NaNs in the manifest. The transform double-compute is gone; the copy the splitter was already paying stays.

spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
jordepic: Done — `stampSortOrderId` (and its `setAccessible` field write) is gone. The driver resolves the write's `SortOrder` from `Table.sortOrders()` (falling back to `SortOrder.unsorted()` for id 0, which isn't always in the map), ships it in the task closure (`SortOrder` is `Serializable`), and `rebuildDataFilesWithJavaMetrics` applies it through the public `DataFiles.Builder.withSortOrder` it was already building through. Verified `withSortOrder(SortOrder)` is identical on 1.5.2 / 1.8.1 / 1.10.0 / 1.11.0. The existing sort_order_id JVM-parity test still passes on all four profiles.

native/core/src/execution/operators/iceberg_write.rs
jordepic: Wired up. The native operator now reports `write_time` — time inside the iceberg-rust writer stack (`write` + `close`), excluding waiting on upstream — which flows through the root `CometMetricNode` by name into a new nano-timing SQL metric ("time in native Iceberg writer") on `CometIcebergWriteExec`. Rows / bytes / files-rolled deliberately stay JVM-derived: they're already surfaced as `numFiles` / `numOutputRows` / `numOutputBytes` from the decoded manifest (the committed values, which is what the stock Spark UI row shows), so native counters with the same meaning would just double-report.

spark/src/main/scala/org/apache/spark/sql/comet/CometMetricNode.scala
jordepic: Narrowed to the metrics access itself. `fromCometPlan` now recurses through every node exactly as upstream does; the only remaining change is that a node whose `session` is `null` (the actual NPE trigger — `metrics` is a lazy val that calls `SQLMetrics.createMetric(sparkContext, ...)`) contributes an empty metric map while its subtree is still walked. Any node with a live session — Comet or not — reports exactly as before, so no existing operator's metrics can regress, and the null-session node's children are no longer pruned either (they were under the old guard).

native/core/src/execution/operators/iceberg_write.rs
jordepic: It owns its store. opendal's memory service creates a fresh `MemoryCore { data: Arc<Mutex<BTreeMap>> }` inside `Builder::build()` — there's no process-global registry — and `load_file_io` builds a new `FileIO` per call, so the manifest bytes are freed when the function returns and the `FileIO`/operator drops. Added a sentence to the comment at the call site so the next reader doesn't have to re-derive it.

jordepic:
@andygrove comments all addressed again, happy to do another round of CI! Thanks for iterating so quickly here!

spark/src/main/scala/org/apache/spark/sql/comet/CometMetricNode.scala
jordepic: Correction to the above after the first full CI run: pre-checking `session != null` was wrong. `fromCometPlan` also runs inside task closures on executors, where the `@transient session` is always null after deserialisation — but `metrics` is safe there because the lazy val was materialised on the driver and shipped with the plan. The pre-check therefore blanked every native operator's metrics on executors, which `CometTaskMetricsSuite`'s "native parquet write reports task-level output metrics" caught on all five [exec] shards.

Now the guard wraps the `metrics` access itself and only swallows the NPE from forcing an unmaterialised lazy val with a null session (the original AQE stage-finalisation case). Every reachable metrics map — driver or executor, Comet node or not — reports exactly as upstream did; `CometTaskMetricsSuite` passes locally again.

Separately, the macos-14 [scans] job died in a hotspot-level crash ("error occurred during error reporting, SIGTRAP") right after an unrelated timestamp suite — that one just needs a re-run.

jordepic:
@andygrove all seven review comments are addressed — each thread above has a reply describing the change. The first full CI run also caught a real bug in my `fromCometPlan` rework (the session pre-check blanked native metrics on executors, failing `CometTaskMetricsSuite` on every [exec] shard); that's fixed in the current head, details on the metrics thread.

Local verification on `810000baf`:

- `CometIcebergWriteDetectionSuite` + `CometIcebergWriteActionSuite`: green on all four profiles (spark-3.4/scala-2.12, 3.5, 4.0, 4.1), including the new uuid-gate, fixed(N), and wide-primitive/non-UTC-timezone parity tests
- `CometTaskMetricsSuite` (the CI failure): 6/6 on all four profiles
- native `iceberg_write` unit tests 17/17; clippy, rustfmt, spotless, scalastyle, scalafix (spark-4.0), prettier, check-suites all clean

Could you approve a CI run when you get a chance? One caveat from the last run: the macos-14 [scans] job died in a hotspot-level runner crash (`SIGTRAP` during error reporting, right after an unrelated timestamp suite) — same class of macos-14 flake as the [shuffle] one, so it may just need a re-run if it recurs.

native/core/src/execution/operators/iceberg_write.rs
```diff
@@ -0,0 +1,1161 @@
...
+    }
+}
+
+/// Copy `len` rows starting at `start` out of `batch`. A zero-copy `RecordBatch::slice` would be
+/// cheaper, but the parquet writer's NaN-count visitor reads list/map children via
+/// `list_array.values()`, which ignores a slice's offset window -- sliced list-of-float columns
+/// would over-count NaNs. `filter_record_batch` materialises compacted children (exactly what
+/// `RecordBatchPartitionSplitter` produces), keeping those counts correct.
+fn materialize_run(batch: &RecordBatch, start: usize, len: usize) -> DFResult<RecordBatch> {
+    let mut mask = vec![false; batch.num_rows()];
```
andygrove: The run-boundary split is a much better shape than what it replaced, and the comment on why a plain `slice` is unsafe here is a good catch. But I think this reintroduces the cost problem from the other direction.

`materialize_run` allocates a `Vec<bool>` the length of the whole batch, bit-packs it into a `BooleanArray`, and then `filter_record_batch` has to scan that full-length predicate to build its filter strategy. That is O(batch rows) per run rather than O(run length), so a batch with R runs costs O(R x batch_rows) across every column. For a low-cardinality partition that is nothing, but a clustered write on something like `bucket(1000, id)` or an hourly partition over a wide range can easily produce hundreds of runs in one 8192-row batch, and then this is a good deal more work than the two transform evaluations I originally asked you to remove.

Would building one sequential `UInt32Array` of `0..num_rows` per batch and then `take`-ing `indices.slice(start, len)` per run work here? `take` on a list or map column gathers the referenced children into fresh compacted arrays, same as `filter` does, so the NaN-visitor concern in your comment should still be satisfied, and each run becomes O(run length) with the total back to O(batch rows).

Separately, `PartitionKey::new(self.partition_spec.as_ref().clone(), ...)` clones the whole `PartitionSpec` once per run. Small next to the above, but it scales the same way, and it would be nice not to have both.

spark/src/main/scala/org/apache/spark/sql/comet/CometMetricNode.scala
```diff
@@ -354,10 +354,22 @@ object CometMetricNode {
...
+   * after deserialisation but `metrics` was materialised on the driver and shipped with the plan.
+   * A node whose metrics are unreachable contributes an empty map, but its subtree is still
+   * walked so every other operator keeps reporting.
    */
   def fromCometPlan(cometPlan: SparkPlan): CometMetricNode = {
-    val children = cometPlan.children.map(fromCometPlan)
-    CometMetricNode(cometPlan.metrics, children)
+    val nodeMetrics =
+      try cometPlan.metrics
+      catch { case _: NullPointerException => Map.empty[String, SQLMetric] }
```
andygrove: This is much better than pruning the subtree, and the note about why the `session` pre-check does not work is worth having. My remaining worry is that the catch is wider than the cause you diagnosed. It will also swallow an NPE thrown from inside any operator's own `metrics` definition, so a genuine bug in a Comet operator's metric construction turns into that operator silently reporting nothing instead of failing loudly.

Since the failure you are guarding is specifically a foreign JVM node built off the planning thread, would gating on that narrow the blast radius enough? Something like only swallowing when the node is not a `CometPlan`. Comet operators reach here either on the driver with a live session or on executors with `metrics` already materialised, so neither case should need the guard, and a real NPE inside a Comet operator's metrics would still surface.

spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
```diff
@@ -569,6 +569,675 @@ class CometIcebergWriteActionSuite
...
+      val fileRows = spark
+        .sql("SELECT record_count FROM cat.db.native_roll.data_files")
+        .collect()
+        .map(_.getLong(0))
+      assert(fileRows.length >= 2, s"expected a multi-file roll, got ${fileRows.length} file(s)")
+      assert(fileRows.sum == 300L, s"rows across rolled files must sum to 300, got $fileRows")
+    }
+  }
+
+  test("native acceleration: empty append commits exactly once with zero data files") {
```
andygrove, line 975: This pins the empty append on an unpartitioned table, so it exercises `UnpartitionedWriter::close()` with nothing written. The partitioned equivalent goes through `ClusteredWriter::close()` or `FanoutWriter::close()` having never seen a batch, and that path is not covered anywhere. An `INSERT INTO ... SELECT` whose filter matches nothing is an entirely ordinary thing to do against a partitioned table, so if either of those writers errors or panics with no current writer it would be a hard failure on a common case.

Could you add a partitioned variant here? It should be a couple of lines given the helper already exists, and it would also cover the `ClusteredBatchSplitter` never being invoked at all.

andygrove (commented):
Thanks for turning all of that around so quickly. Every one of the seven points is addressed, and a few of the answers taught me something: I had not realised opendal's memory service has no global registry, and the correction on `session != null` not being a usable pre-check (because `fromCometPlan` also runs in task closures where `session` is always null) is exactly the kind of thing that only a full CI run surfaces. Thanks for writing that reasoning into the doc comment rather than just fixing it.

I checked the claim in your new `requireNoUuidColumns` comment that `uuid` is the only Spark-writable Iceberg type with an Arrow mismatch, since that is the sort of thing that quietly stops being true. It holds: `TypeToSparkType` still throws `UnsupportedOperationException` for `TIME` as of Iceberg 1.11, so a `time` column cannot be planned by Spark at all, and everything else new in 1.11 (`variant`, `geometry`, `geography`, the nanosecond timestamps, `unknown`) is V3 and already excluded by the format-version gate. So the gate is complete for V1 and V2, and the comment is accurate.

I also confirmed the `write_time` plumbing works the way you describe. `subset_time("write_time", partition)` matches the existing convention in `scan.rs` and `parquet_writer.rs`, and `CometMetricNode.set_all` zips the native children against the JVM children, so the extra native `Projection` and `Scan` nodes under a `Nil`-children root are silently ignored rather than mismatched. Unknown metric names are a debug no-op, so nothing else is disturbed.

On formatting: `cargo fmt --all -- --check` now passes on this head. I re-ran it against the previous head to be sure the original report was real, and it was. Worth noting that CI has not run at all on `810000ba` yet, so none of this is machine-verified beyond the format check.

Three things left, all on the new code.

native/core/src/execution/operators/iceberg_write.rs
jordepic: Done, exactly as you sketched: one sequential `UInt32Array` of `0..num_rows` per batch, and each run gathers through `take_record_batch(batch, indices.slice(start, len))` — O(run length) per run, O(batch rows) per batch total, and `take` compacts list/map children the same way `filter` did, so the NaN-visitor constraint holds. Single-run batches (the common case) short-circuit to a zero-copy `batch.clone()` — arrays keep their original zero offsets there, so the slice hazard doesn't apply.

On the `PartitionSpec` clone: I couldn't remove it — `PartitionKey::new` takes its spec by value, and `copy_with_data` clones internally too, so one spec clone per run is the floor with the current iceberg-rust API. Left a comment saying so; it's now the only per-run cost that isn't O(run length), and specs are a handful of fields.

spark/src/main/scala/org/apache/spark/sql/comet/CometMetricNode.scala
jordepic: Narrowed as suggested: `CometPlan` nodes read `metrics` unguarded, so a genuine NPE inside a Comet operator's metric construction fails loudly; only foreign nodes get the catch. Doc comment updated with the reasoning (Comet operators reach here on the driver with a live session or on executors with `metrics` already materialised, so neither case needs it). `CometTaskMetricsSuite` still passes on all four profiles.

spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
jordepic: Added `native acceleration: empty append to a partitioned table commits with zero data files`, covering both partitioned writers: an identity-partitioned table on the default clustered path and a fanout twin (`write.spark.fanout.enabled=true`), each fed an `INSERT INTO ... SELECT` matching nothing. Both commit exactly once with zero data files through `CometIcebergWriteExec`, with the splitter never invoked. I also read both writers' `close()` at the pinned rev — `ClusteredWriter` takes the (absent) current writer and `FanoutWriter` iterates an empty map, so both return empty output rather than erroring — and the test now pins that. Green on all four profiles.

jordepic:
@andygrove good for a CI kick

jordepic:
@andygrove failure seems transient

jordepic:
Mind if we kick off CI again this morning @andygrove ? Thank you!

jordepic:
Huh this thing ballooned a bit. Let me know if you'd prefer I split up this patchset again. Either way!

andygrove:

> Scope note: 4,200 lines across a JVM reflection layer, a new native operator, and a new proto surface. I have reviewed this structurally rather than line by line.

The split of responsibilities is the right one, and I am glad you moved away from having Rust write the metadata too. Data files in iceberg-rust, an in-memory manifest back over JNI, and commit semantics staying entirely in iceberg-java means the parts where divergence would be catastrophic (commit, snapshot isolation, manifest correctness) are untouched, and the part where divergence is merely a file-layout difference is the part that moved. Documenting the places where iceberg-java and iceberg-rust cannot be byte-identical in `iceberg-writes.md` rather than pretending they can be is exactly right.

Five things.

**What are the actual divergences, and can any of them be correctness issues?**

`iceberg-writes.md` gains 66 lines describing where the two writers differ. Could the description summarize the list, and more importantly classify each one? A different row-group boundary or file split point is cosmetic. A different column-stats rounding, a different null-count convention, or a different lower/upper bound truncation would affect *pruning correctness for later readers*, which is a much bigger deal because the damage outlives the write.

Since the Java side converts the in-memory manifest using its own writer settings for column stats, I suspect the stats are safe. Confirming that explicitly would be valuable.

**673 lines of new reflection**

`IcebergReflection.scala` grows substantially to unpack writer settings. Which Iceberg versions is this tested against, and what happens on a version where a field or method moved: a clean decline through the eligibility gate, or a `NoSuchMethodError` during planning or, worse, during a write? A failure mid-write on a table is much worse than a failure at plan time, so it would be good to know the reflection is all resolved up front.

The project convention of applying the `run-iceberg-tests` label for reflection changes applies here.

**No performance numbers**

The rationale is "we want to speed up writing iceberg data for ETL jobs and table maintenance". The description has no numbers. What does a native append cost versus the Java writer, on an unpartitioned table and a partitioned one? That is the entire justification and it should be in the description before merge.

**Eligibility gate and failure mode**

If the native writer starts a write and then fails partway through (an unsupported setting discovered at runtime, an object-store error), what happens? Does the task retry through the Java writer, or does the write fail? Partial data files from an aborted native write need to not be committed, and the answer should be stated even if it is "Spark's task retry handles it, and uncommitted files are cleaned up by the normal Iceberg orphan-file process".

**Test coverage of the fallback pins**

The suite pins that CoW MERGE intentionally falls back because `MergeRowsExec` is not Comet-native. #5318 is adding native `MergeRowsExec`. When that lands, this pin becomes wrong. Could the test reference #5318 so whoever lands it knows to revisit, rather than just seeing a failing assertion?

jordepic:
Thanks Andy — all five points addressed. Point by point:

**1. Divergence classification + stats safety**

Restructured the divergence section of `iceberg-writes.md` into three explicitly classified groups: *physical file layout only* (footer metadata, schema element naming, `created_by`, page CRCs/header stats, dictionary labeling, row-group/rolling cadence, compressed bytes — no reader bases a decision on any of them), *manifest metadata visible to later readers*, and one *operational caveat* (unescaped partition paths).

Your suspicion about the stats is right, and the doc now says so explicitly: the manifest-visible class is deliberately almost empty because `DataFile` metrics are not taken from the native writer at all — they are re-derived on the JVM from each written file's parquet footer through iceberg-java's own `ParquetUtil.footerMetrics` + `MetricsConfig.forTable`, so bound truncation, null-count conventions, and metrics modes are iceberg-java's code making iceberg-java's decisions (and the parity suite compares committed manifests against JVM-written ones). Exactly two footer-derived values can differ from what iceberg-java's *writer-tracked* state would have recorded, and both are analyzed in the doc: zero-sign bound normalization (a strictly conservative widening, cannot change pruning) and the Iceberg 1.10+ float/double-under-nullable-struct counts (value and null counts inflate by the same amount, so null-based pruning is unchanged).

**2. Reflection resolution**

Tested against Iceberg 1.5.2 (spark-3.4 profile), 1.8.1 (spark-3.5), 1.10.0 (spark-4.0), and 1.11.0 (spark-4.1) — the version-sensitive spots (two ambiguous 6-arg `FieldMetrics` constructors on 1.10+, the `GenericManifestFile` constructor arity change at 1.6, `sort_order_id` stamping change at 1.11) are each handled and pinned by tests.

Your worry was legitimate, though: the plan-time gate resolved everything *it* used, but the commit-message assembly that runs on executors (manifest decode, the footer-metrics rebuild, `TaskCommit` construction) resolved its reflection lazily on first use — a method moved there would have been a task failure after the data files were already written. Fixed in this pass: the eligibility gate now eagerly probes the entire executor-side reflection surface (every class, method, and constructor) on the driver at plan time, so an Iceberg release that moves any of it becomes a clean fallback with a reported reason. A new test in `IcebergReflectionSuite` pins that the probe resolves on every CI profile, so an Iceberg version bump that breaks the surface fails loudly in CI rather than silently falling every write back.

I don't have permission to apply the `run-iceberg-tests` label — could you add it?

**3. Performance numbers**

Added to the description. Summary (2M rows x 100 mixed-type columns, ~1.5 GB, Spark 4.0.1 / Iceberg 1.10.0, local[*], medians of 3): an unpartitioned `INSERT INTO ... SELECT` goes 9.5 s → 3.3 s when the native writer replaces the Java writer under an identical Comet scan (2.9x on the write side); a `bucket(8, int)` partitioned append with fanout goes 24.1 s → 4.4 s (5.5x). One honest caveat is in the description: with Iceberg's default hash distribution, the exchange partitions by Iceberg's `BucketFunction`, which Comet doesn't run natively, so a default-distribution bucketed append currently declines (fully-Comet-native-input rule) — the partitioned numbers use `write.distribution-mode=none` + fanout.

**4. Failure mode**

Now stated in a "Failure handling" section in `iceberg-writes.md`. The short version is close to your suggested answer, with one honest caveat:

- Eligibility is decided entirely at plan time; the physical plan is fixed, so a failed task retries through the native writer again (never the Java writer). Retries cannot collide — the task attempt id is embedded in file names.
- Partial results are never committed: the commit set is exactly the commit messages of successful tasks, and a job failure aborts before commit.
- The caveat: data files finalized by a failed attempt are *not* deleted by that task — iceberg-java's `DataWriter.abort` deletes them, and the native path has no abort hook yet. They're invisible to all readers (files resolve through committed manifests) and are reclaimed by the normal `remove_orphan_files` maintenance. A task-abort cleanup hook is a reasonable follow-up.
- Driver-side commit failure behaves exactly as stock: the commit messages are genuine `SparkWrite$TaskCommit`s, so `SparkWrite.abort`'s cleanup applies unchanged.

**5. #5318 cross-reference**

Done — the CoW MERGE fallback pin now points at #5318 and says what to flip when it lands.

jordepic:
Sorry for the delay here, @andygrove ! I was on vacation. If you want to rerun the suites that'd be great!

docs/source/user-guide/latest/iceberg-writes.md
```diff
@@ -156,23 +170,64 @@ identically on both paths, WAP / branch / snapshot properties act on the JVM com
...
+failure), the error propagates as an ordinary Spark task failure and Spark's task retry
+re-executes it — through the native writer again. Retries cannot collide: each attempt's task
+attempt id is embedded in its data file names.
+
+Partial results are never committed. The commit set is exactly the commit messages returned by
+successful tasks — a failed task contributes none — and if the job fails, the driver-side
+commit operator aborts without committing anything. Data files already finalized by a failed
+task attempt are not deleted by that task (iceberg-java's writer abort deletes them; the
+native path has no abort hook yet): they are invisible to every reader, since readers resolve
+files through committed manifests only, and are reclaimed by Iceberg's normal
```
andygrove: This is honest documentation of a real gap, and I appreciate that it's called out rather than glossed. My worry is that it will not turn into work unless it gets a tracking issue. In practice this is a per-task-failure orphan cost that iceberg-java's `SparkPartitionedDataWriter` cleans up synchronously via its abort hook. On a well-behaved cluster it is invisible; on spot-heavy or preemption-prone environments it accumulates until someone remembers to schedule `remove_orphan_files`, and users on the native path may not know they've opted in to that trade until they hit it.

Could you file a tracking issue for the abort hook (Spark's `TaskContext.addTaskFailureListener` is the natural spot, mirroring what `SparkPartitionedDataWriter` does) and link it from this sentence? A pointer here makes the operational cost visible to users reading the compat page, and makes it easy for the follow-up to close it out.

spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
```diff
@@ -1110,6 +1268,607 @@ object IcebergReflection extends Logging {
...
+    ctor.newInstance(array.asInstanceOf[AnyRef]).asInstanceOf[AnyRef]
+  }
+
+  /**
+   * Eagerly resolves every class, method, and constructor the executor-side commit-message
+   * assembly reflects on (`decodeManifestToDataFiles`, `rebuildDataFilesWithJavaMetrics`,
+   * `sumDataFileMetrics`, `buildTaskCommit`). That code runs after iceberg-rust has already
+   * written the task's data files, so a reflection miss there is a task failure; probing the full
+   * surface from the eligibility gate turns an Iceberg release that moves any of it into a
+   * plan-time fallback instead. Memoized: the linked Iceberg cannot change within a JVM, and
```
andygrove: Small thing on the memoization scope. `lazy val` on a Scala object caches per JVM, not per Spark session. The comment says "the linked Iceberg cannot change within a JVM," and for a long-running driver that is *almost* always true — but `SparkSession.sql("ADD JAR /path/to/iceberg-1.9.jar")` and `spark-shell --jars`-style dynamic classpath changes can shift what `loadClass(...)` resolves to mid-run, and a memoized `Some("...")` from the first probe would then persist even after the classpath was fixed. The reverse is worse: a first probe that resolved successfully against an old Iceberg gets cached as `None`, and a later `ADD JAR` to a moved release would then let ineligible writes through.

Either the comment should soften the claim ("unless dynamic classpath changes intervene") or the memoization could be keyed by the resolved Iceberg class-loader identity, whichever you prefer. This is not a real hazard in the common deployment shape, but the current comment is more absolute than the reality.

andygrove (commented):
The three points from my last pass are cleanly resolved. `ClusteredBatchSplitter` now uses one sequential `UInt32Array` per batch and `take_record_batch` per run, with a nice zero-copy fast path for the single-run common case, and the residual `PartitionSpec` clone is honestly acknowledged as an iceberg-rust API constraint rather than papered over. `fromCometPlan` is narrowed to swallow only on foreign nodes, so a genuine NPE in a Comet operator's `metrics` still fails loudly. The partitioned empty-append test covers both `ClusteredWriter::close()` and `FanoutWriter::close()` with the splitter never invoked.

While checking the newly added `executorReflectionUnresolved` probe I wanted to flag one thing that ended up being an unexpected improvement: the same commit removes the blanket `catch (_: Exception)` from `extractFileLocation`, `getFileFormat`, and `getEqualityFieldIds`. Under the old shape, if `.invoke(file)` threw at runtime (e.g. an unexpected concrete type), a delete-file path lookup or an `equalityFieldIds()` call silently returned `None`/empty, and the scan-side callers in `CometIcebergNativeScan` (lines 287, 1024, 1064) would then either misreport the file as position-only or fail with a misleading "neither location() nor path() is declared" message. The new narrow shape propagates the real exception instead, which is what those call sites want. That is a scan-side correctness improvement that happens to ride along in this write PR.

Two things worth a look, one small.

docs/source/user-guide/latest/iceberg-writes.md
jordepic: Filed https://github.com/apache/datafusion-comet/issues/5618 and linked it from this sentence. The issue spells out the shape you suggested (`TaskContext.addTaskFailureListener`, mirroring `SparkCleanupUtil.deleteTaskFiles`) and splits it into the easy half — failures after the native writer returns, where the decoded `DataFile` paths and the `FileIO` are already in the task closure — and the harder half, failures inside the native write, where closing the gap needs the native operator to report finalized paths alongside the error since iceberg-rust's writers have no abort/Drop cleanup of their own.

spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
jordepic: You're right — and the second direction (a cached `None` letting ineligible writes through after an `ADD JAR` to a moved release) convinced me to fix the memoization rather than soften the comment. The probe result is now keyed by the class loader of the resolved `ContentFile` class: `executorReflectionUnresolved` re-probes whenever `loadClass` resolves Iceberg from a different loader, and the doc comment now states the `ADD JAR` rationale explicitly.

Chasing that through also surfaced that the three parquet-mr-bridging lookups (`ParquetUtil.footerMetrics`, `ParquetIO.file`, and the shading-derived `ParquetFileReader.open`) were JVM-lifetime lazy vals, so a re-probe against a new loader could have been satisfied by `Method`s memoized against the old one. They're now resolved together per call (`resolveParquetFooterReflection()`), once per task rather than per file, so both the probe and the executor-side rebuild always reflect the currently-loadable Iceberg.

jordepic:
@andygrove all of those issues are addressed! Would you mind kicking CI again?

jordepic:
@andygrove I think the failure is a transient maven one

jordepic:
@andygrove CI is green across the whole matrix except `Verify TPC-DS Results`, which hit a Maven Central HTTP 429 downloading Maven itself before any project code ran — could you re-run that one job when you get a chance? (Same infra flake hit the lint job on the previous head.)

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
```diff
@@ -172,21 +169,23 @@ object CometIcebergNativeWrite extends CometOperatorSerde[IcebergWriteExec] with
...
     requireOnlyVettedParquetWriteProperties,
     requirePropertyAbsent(
       PropertyKeys.ParquetEnableDictionary,
       "dictionary override unsupported"),
     requireNoUnvettedParquetMrProperties,
     requirePropertyAbsent(PropertyKeys.FileIOImpl, "custom FileIO unsupported"),
```
unikdahal, line 189: This only detects io-impl when it is present in the table/write properties. Iceberg can also configure the FileIO at the catalog level, so table.io() may be custom while this check still passes, causing the native path to bypass it.Could we validate the actual instantiated table.io() here (similar to the scan-side compatibility check) ?

spark/src/main/scala/org/apache/comet/serde/operator/IcebergWriteProtoTranslation.scala
```diff
@@ -0,0 +1,193 @@
...
+    explicit.orElse(parquetMrDefaultLevel(codec))
+  }
+
+  private def parquetMrDefaultLevel(codec: CompressionCodec): Option[Int] = codec match {
+    case CompressionCodec.Zstd => Some(3)
+    case _ => None
+  }
+
+  private def parseLong(props: Map[String, String], key: String, default: Long): Long =
+    props.get(key).flatMap(s => Try(s.trim.toLong).toOption).getOrElse(default)
```
unikdahal: These helpers silently fall back to the default on parse failure, which differs from iceberg-java's validation behavior. The corresponding Parquet settings are parsed as positive Ints there, so values like garbage, 0, -1, or an Int overflow can behave differently on the native path. Could we preserve the iceberg-java parsing/range semantics here, or fall back before selecting the native writer, with a few boundary tests?

native/core/src/execution/operators/iceberg_common.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::cloud::s3::credential_bridge::{AccessMode, CometS3CredentialBridge};
+
+/// Activation key for the `CometS3CredentialProvider` SPI, read from a catalog's `s3.*` property
+/// bag.
+const ICEBERG_PROVIDER_CLASS_PROPERTY: &str = "s3.comet.credential.provider.class";
+
+/// Key prefixes forwarded to iceberg-rust's `FileIO`. The full unfiltered catalog bag (catalog
+/// URI, OAuth tokens, credentials.uri, tenant-id, etc.) is kept upstream so
+/// `CometS3CredentialBridge` can read whatever the vendor needs.
+const STORAGE_PROPERTY_PREFIXES: &[&str] = &["s3.", "gcs.", "adls.", "client."];
```
unikdahal, line 36: oss is supported by storage_factory_for, but oss.* is missing from the property prefixes forwarded into iceberg-rust's FileIO. That means OSS-specific endpoint/credential configuration can be dropped even though the write is considered supported.Could we either forward oss.* and add a functional OSS write test, or leave OSS out of the supported schemes until that path is covered?

unikdahal (commented):
Thanks @jordepic for this PR. I've been experimenting quite a bit with the native write path since your earlier PRs, so I took a closer look at this one as well.

I did an AI-assisted review and manually revalidated the findings below against the current implementation and upstream Iceberg behavior. I found a few compatibility gaps that seem worth addressing.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
```diff
@@ -216,29 +215,35 @@ object CometIcebergNativeWrite extends CometOperatorSerde[IcebergWriteExec] with
...
+          .findFieldWithTypeIds(schema, UnsupportedWriteTypeIds)
+          .map { case (name, typeId) =>
+            s"column $name has Iceberg type ${typeId.toLowerCase(Locale.ROOT)}, " +
+              "which the native writer cannot reproduce"
+          }
+    }
+
   private val requireNoEncryptionPrefix: TriggerRule = ctx =>
     ctx.properties.keys
       .find(_.startsWith(EncryptionPropertyPrefix))
```
unikdahal, line 247: This infers plaintext behavior from the absence of encryption.* properties, but Iceberg's output-file contract is based on table.encryption(), which is independent of table.properties(). Could we fail closed based on the actual EncryptionManager rather than the property prefix? That would also cover custom TableOperations implementations.

native/core/src/execution/operators/iceberg_common.rs
```diff
@@ -0,0 +1,142 @@
...
+        url.path(),
+        access_mode,
+        catalog_properties,
+    );
+    match bridge {
+        Ok(b) => Some(CustomAwsCredentialLoader::new(b)),
+        Err(e) => {
+            log::warn!(
+                "Failed to initialize CometS3CredentialBridge for {provider_class}: {e}; \
+                 falling back to default opendal credential chain"
```
unikdahal: If an S3 credential provider is explicitly configured but bridge initialization fails, this continues with OpenDAL's default credential chain. For AccessMode::Write, that can change which credentials perform the write after the configured provider failed. Should an explicitly configured provider fail closed for write mode instead?

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
jordepic: Good catch — the property rule genuinely cannot see a catalog-level `io-impl` (or a custom catalog installing its own FileIO). Added `requireRecognizedTableFileIO`: the gate now inspects the instantiated `table.io()` against the same class-hierarchy allowlist the scan side uses (`COMPATIBLE_FILE_IO_CLASSES`), with one deliberate difference — the `EncryptingFileIO` family, which the scan accepts, is rejected on the write side because the native writer produces plaintext data files. The property rule stays as a cheap early signal. New detection test registers a catalog with a catalog-level `io-impl` pointing at a FileIO that delegates to `HadoopFileIO` by composition (a subclass would pass the hierarchy check by design) and pins the fallback reason.

spark/src/main/scala/org/apache/comet/serde/operator/IcebergWriteProtoTranslation.scala
jordepic: Right on all counts — iceberg-java reads these through `PropertyUtil.propertyAsInt` (`Integer.parseInt` semantics: no trimming, overflow throws) and parquet-mr rejects non-positive values at write time, while the native translation was silently substituting defaults. Fixed in both layers: a new gate rule (`requirePositiveIntParquetSizes`) declines any of the four size/limit properties that is not a positive Java int, so those values fail on the stock path exactly as they would without Comet; and the translation helpers now use strict `Integer.parseInt` with no silent fallback (a parse failure there would be a gate/translation mismatch, and throws loudly). The compression-level rule was aligned to the same no-trim semantics. Boundary tests cover `garbage`, `0`, `-1`, `2147483648`, and `" 1024"` per key in the detection suite, plus strict-parse pins in `IcebergWriteProtoTranslationSuite`.

native/core/src/execution/operators/iceberg_common.rs
jordepic: You are right that this was aspirational: `oss.*` was never in the forwarded property prefixes and nothing tests the path. Took your second option — `oss` is removed from the supported schemes (gate, docs, and the rust factory arm, which nothing else could reach since the scan side never admits an OSS FileIO), with a detection test pinning the fallback. Forwarding `oss.*` plus a functional test can bring it back as a follow-up if someone has an OSS environment to validate against.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
jordepic: Agreed — the property prefix is a proxy, and a custom `TableOperations` can install an `EncryptionManager` without any table property. Added `requirePlaintextEncryptionManager`: the gate now reflects `table.encryption()` and declines anything whose class is not `org.apache.iceberg.encryption.PlaintextEncryptionManager`, failing closed when the accessor cannot be resolved at all. The `encryption.*` property rule stays as an early signal. (Same-commit related change: the new `table.io()` check also rejects the `EncryptingFileIO` family on the write side, so encrypted tables are declined from both directions.) Unit tests in `IcebergReflectionSuite` cover the plaintext, custom-manager, and unresolvable cases.

native/core/src/execution/operators/iceberg_common.rs
jordepic: Agreed — silently switching which credentials perform a write after the configured provider failed is not acceptable. `build_s3_credential_loader` now returns `Result<Option<...>>`: with a provider explicitly configured, `AccessMode::Write` fails closed with an error naming the provider class, while `AccessMode::Read` keeps the warn-and-fall-back behavior (a wrong-credential read fails on permissions rather than corrupting anything, and changing scan behavior felt out of scope here). No provider configured still means `Ok(None)` → default chain, unchanged.

native/core/src/execution/operators/iceberg_common.rs
unikdahal: Thanks. I think removing OSS from the shared factory introduces a regression on the existing read path, though.

storage_factory_for is now shared by reads and writes. Before this refactor, iceberg_scan.rs had an OpenDalStorageFactory::Oss arm, and CometScanRule still explicitly treats oss as a supported Iceberg scan scheme. HadoopFileIO is also accepted by the scan-side FileIO compatibility check, so an oss:// table can reach native scan even without a dedicated OSS Java FileIO.

That means such a scan can still pass JVM eligibility but now fail here with Unsupported storage scheme: oss.

Could we preserve OpenDalStorageFactory::Oss for AccessMode::Read and reject it only for AccessMode::Write until OSS write configuration/testing exists? A read regression test would also pin that this write-safety change doesn't remove an existing scan capability.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
unikdahal: Thanks, validating the instantiated table.io() fixes the custom-FileIO case I raised, but I think there is still a configuration-equivalence gap for a recognized HadoopFileIO.

The write proto currently builds catalogProperties only from getFileIOProperties(table). properties() is not part of the FileIO interface and HadoopFileIO normally carries its S3A configuration through Hadoop Configuration instead.

The existing native Iceberg scan path already handles this explicitly: it calls NativeConfig.extractObjectStoreOptions(...) and translates fs.s3a.* settings into the s3.* properties iceberg-rust/OpenDAL consumes. The write path doesn't do that.

So a HadoopFileIO table at s3a://... can pass this new compatibility gate while the JVM writer sees credentials/custom endpoint/path-style settings from spark.hadoop.fs.s3a.* that the native writer never receives.

Could we reuse the scan-side Hadoop-property extraction/translation here and merge it with the FileIO properties, with table/vended FileIO properties taking precedence?

jordepic: Right — `getFileIOProperties` covers vended/table FileIO properties, but `HadoopFileIO` carries its S3A configuration in the Hadoop Configuration, so the JVM writer would honour settings the native writer never received (the custom-endpoint case is the genuinely scary one). Done exactly as you suggested: the write proto assembly now extracts the object-store options for the data location from the session Hadoop configuration (`NativeConfig.extractObjectStoreOptions`), translates them with the scan side's `hadoopToIcebergS3Properties`, and merges them UNDER the FileIO properties (`hadoopDerived ++ fileIO`), so table/vended FileIO properties take precedence as you asked. A new detection test sets `fs.s3a.endpoint` / `fs.s3a.access.key` / `fs.s3a.path.style.access` on the session, converts an s3a-location insert, and asserts the built proto's `catalog_properties` carry the translated `s3.*` keys.

native/core/src/execution/operators/iceberg_common.rs
jordepic: You are right — I over-rotated. The factory is shared, and the scan side does admit `oss://` (HadoopFileIO passes the scan FileIO check and `CometScanRule` lists oss as a supported scheme), so dropping the arm regressed an existing read capability. Restored exactly as you propose: `OpenDalStorageFactory::Oss` for `AccessMode::Read`, an explicit error for `AccessMode::Write` (a native-side backstop behind the JVM write gate, which already declines oss locations). Added rust unit tests on `storage_factory_for` pinning oss read-ok / write-err, the common schemes in both modes, and unknown-scheme rejection.

spark/src/main/scala/org/apache/spark/sql/comet/CometIcebergWriteExec.scala
```diff
@@ -0,0 +1,280 @@
...
+    val bytes = new Array[Byte](size)
+    val codedOutput = CodedOutputStream.newInstance(bytes)
+    nativeOp.writeTo(codedOutput)
+    codedOutput.checkNoSpaceLeft()
+    SerializedPlan(Some(bytes))
+  }
+
+  override def withNewChildInternal(newChild: SparkPlan): SparkPlan = copy(child = newChild)
+
+  override def nodeName: String = "CometIcebergWrite"
```
andygrove, line 110: Now that `catalog_properties` picks up the translated `fs.s3a.*` settings, could this override `stringArgs` the way `CometIcebergNativeScanExec` does?

`nativeOp` is the first field of the case class and there is no `stringArgs` override, so `TreeNode.argString` falls through to its `case other => other :: Nil` branch and calls `toString` on the protobuf, which gives the whole TextFormat dump. That lands in `explain()`, in the SQL UI node description, and in `SparkListenerSQLExecutionStart` in the event log. With the merge you just added in `buildIcebergWriteProto`, the dump now includes `catalog_properties { key: "s3.secret-access-key" ... }` for an S3A table whose keys live in the Hadoop configuration.

Spark's own redaction will not catch this. `argString` redacts through `redactMapString` only for args that are Scala `Map`s, and a protobuf message falls past that to the plain `toString`. `CometIcebergNativeScanExec.stringArgs` sidesteps exactly this by emitting `output` plus a short descriptor instead of the proto, and it carries the same property bag. Something like `Iterator(output, s"$dataLocation, $writerMode")` here would keep the node readable and keep the secret out of the log.

Your new "write proto forwards fs.s3a.*" test looks like a good place to pin it, by asserting the exec's `simpleString` does not contain the secret value.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
```diff
@@ -263,6 +275,15 @@ object CometIcebergNativeWrite extends CometOperatorSerde[IcebergWriteExec] with
       .map(_ => s"$key=true (variant shredding changes the parquet schema)")
   }
 
+  // iceberg-java throws NumberFormatException at write time for a non-integer level, while the
+  // native translation would silently substitute the codec default. Fall back so the failure
+  // behaviour matches the stock path.
+  private val requireParseableCompressionLevel: TriggerRule = ctx =>
+    ctx.properties
+      .get(PropertyKeys.ParquetCompressionLevel)
+      .filter(v => scala.util.Try(java.lang.Integer.parseInt(v)).isFailure)
```
andygrove: This checks that the level parses as a Java int, but the native side is stricter than that. parquet-rs enforces zstd `1..=22`, gzip `0..=9` and brotli `0..=11`, and `build_writer_properties` returns an error outside those, which surfaces as a task failure rather than a fallback.

iceberg-java does not validate the level at all. I pulled `Parquet$WriteBuilder$Context` out of the 1.10.0 runtime jar to check, and unlike the four size properties, `compressionLevel` is kept as a raw `String` and dropped straight into `parquet.compression.codec.zstd.level` / `zlib.compress.level` / `compression.brotli.quality`. No `propertyAsInt`, no `checkArgument`. So `write.parquet.compression-level=0` with zstd, or a negative zstd fast level, writes fine on the stock path and dies mid-task on this one.

Could this rule resolve the codec (`resolveCompression` is right there in `IcebergWriteProtoTranslation`) and decline anything the corresponding parquet-rs level type would reject? That is the same treatment `requirePositiveIntParquetSizes` just gave the size properties. Boundary cases worth pinning in the detection suite next to the existing `fast` test: zstd `0`, zstd `-3`, zstd `23`, gzip `-1`, brotli `12`.

andygrove (commented):
Thanks for turning unikdahal's points around so quickly. Both of the last two are cleanly resolved, and I checked the pieces I could rather than taking them on trust, recorded here so nobody re-does the work:

- The write-side `table.io()` check really does exclude the `EncryptingFileIO` family. `isCompatibleFileIO` is `classNameInHierarchy(...) || startsWith(ENCRYPTING_FILE_IO_PREFIX)` and `requireRecognizedTableFileIO` uses only the first half, so the comment above it is accurate.
- `requirePositiveIntParquetSizes` matches iceberg-java exactly. I decompiled `Parquet$WriteBuilder$Context` out of the 1.10.0 runtime jar: row-group size, page size, page row limit and dict size are each read through `PropertyUtil.propertyAsInt` and `checkArgument`ed to be `> 0`. Same four keys, same rule.
- The gate map and the translation map really are the same map, so the new `parseJavaInt` throw is a tripwire and not a reachable path. `ctx.properties` is `tableProperties ++ writeProperties` and `effectiveProperties` is `properties ++ resolvedWriteProperties`, both built from the same two reflection calls. It also sits inside `convert`'s catch, so even the tripwire degrades to a fallback rather than a failure.
- The OSS restore is split correctly by `AccessMode`, and the docs table no longer lists `oss`.
- The new Hadoop-derived merge is a no-op for local `file://` tables, since `extractObjectStoreOptions` has no prefixes registered for the `file` scheme and `hadoopToIcebergS3Properties` drops whatever is left. So the common test path is unaffected by it.

## CI

A fresh run is in flight on `7443bcf8`: 14 green, 6 still pending, nothing red yet.

Two things left, both on the newly added code, and both are inline.

spark/src/main/scala/org/apache/comet/serde/operator/CometIcebergNativeWrite.scala
jordepic: Confirmed against the pinned parquet-rs 58 source: `CompressionLevel` bounds are exactly zstd `1..=22`, gzip `0..=9`, brotli `0..=11`, and `build_writer_properties` errors outside them — while, as you dug out, iceberg-java keeps the level as a raw string with no validation at all. The rule is now `requireNativeSupportedCompressionLevel`, delegating to a new `IcebergWriteProtoTranslation.compressionLevelRejection` that resolves the effective codec via `resolveCompression` and declines any level the corresponding parquet-rs type would reject (codecs without a native level concept — snappy/lz4/none — ignore the property on both sides, so any int passes). All five of your boundary cases are pinned in the detection suite next to the existing `fast` test, plus zstd `22` staying Compatible so the gate is not over-broad, and the range matrix (rejected and accepted boundaries per codec) is unit-tested in `IcebergWriteProtoTranslationSuite`.

spark/src/main/scala/org/apache/spark/sql/comet/CometIcebergWriteExec.scala
jordepic: Yikes — right, and this became load-bearing the moment the fs.s3a merge landed. Added the `stringArgs` override modeled on `CometIcebergNativeScanExec`: `Iterator(output, s"$dataLocation, $writerMode")`, so explain()/UI/event-log render a short descriptor instead of the protobuf TextFormat dump. Pinned exactly where you suggested: the fs.s3a forwarding test now also asserts `simpleString(Int.MaxValue)` does not contain the injected access key (and does contain the data location, so the node stays readable).

jordepic:
@andygrove two infra failures on the latest run, everything else green (74 passing) — could you re-run the failed jobs?

- `Spark 4.0, JDK 21 [exec]` died in toolchain setup: rustup could not download `channel-rust-stable.toml.sha256` (network flake, no project code ran).
- `macos-14 [scans]` is the known hotspot SIGTRAP flake on those runners — the JVM crash-reporter loop (`error occurred during error reporting … SIGTRAP`, then "Too many errors, abort") fired mid-`ParquetTimestampLtzAsNtzSuite`, a parquet read suite this PR doesn't touch.

andygrove (approved):
LGTM. Thanks @jordepic! Let's merge this and keep iterating.

cc @comphead @hsiang-c. Note that the feature is still disabled by default.

unikdahal:
Thanks again for all the work here and for addressing the earlier review feedback.

Since this has already merged, I moved the remaining findings into follow-up issues:

#5636  — update the pinned iceberg-rust revision to include the already-upstreamed partition-path escaping fix and add interoperability coverage.
#5637  — preserve or fail closed on Hadoop GCS configuration for `HadoopFileIO + gs://` native writes.

Leaving these here for visibility and future tracking.

Issue #4294, opened 2026-05-12
Allocate Comet's parquet reader buffers from ArrowUtils.rootAllocator to enable zero-copy PyArrow UDF runner


andygrove:
## What is the problem the feature request solves?

PR #4234 replaces the row-based input path in `CometMapInBatchExec` with a `CometArrowPythonRunner` that consumes columnar batches directly. The end-to-end speedup on Spark 4.0 is **1.92x narrow / 1.79x mixed / 1.42x wide** (vs the prior row-path numbers of 1.32x / 1.32x / 1.09x).

The PR is _bulk-copy_, not true zero-copy. For every batch, Comet's buffers are `Unsafe.copyMemory`'d into Spark's allocator before the Arrow IPC stream is unloaded. The reason is that Comet's Parquet readers each construct their own independent `RootAllocator`, separate from `ArrowUtils.rootAllocator`. Arrow Java rejects cross-root transfers (`A buffer can only be associated between two allocators that share the same root`), so `TransferPair.transfer()` cannot move buffers from Comet's allocator into the Python runner's `VectorSchemaRoot`.

If Comet's reader allocators shared a root with `ArrowUtils.rootAllocator`, the bulk-copy step in `CometColumnarPythonInput.writeNextBatchToArrowStream` would collapse to a pointer-swap, and the wide-row speedup should climb further (the wide-row case is where the copy cost is most concentrated; narrow rows are bottlenecked on Python fork/IPC).

Related: #957, #4234.

### Change 1: allocator parent

The five files under `common/src/main/java/org/apache/comet/parquet/` that today declare `new RootAllocator()` would instead allocate as children of `ArrowUtils.rootAllocator`:

- `ColumnReader.java:55`
- `NativeColumnReader.java:41`
- `NativeBatchReader.java:154`
- `ArrowConstantColumnReader.java:46`
- `ArrowRowIndexColumnReader.java:39`

Plus anywhere else \`new RootAllocator()\` appears in the read path.

Result: every Comet buffer is in `ArrowUtils.rootAllocator`'s family. `makeTransferPair(target).transfer()` succeeds.

### Change 2: trait simplification (downstream)

`CometColumnarPythonInput.copyVector` collapses from a tree walk + per-buffer memcpy to a single \`TransferPair.transfer()\` call per column. `CometVectorIpcCopier` in \`common\` is no longer needed and can be deleted. About 100 lines of code go away.

### Coupled question: shading

[...]

PR #5368, opened 2026-08-15
perf: serialize Python input directly from Comet Arrow vectors

### LINKED ISSUE #4383: Drop the per-batch Comet→Spark buffer copy in CometColumnarPythonInput

PR #4234 streams Comet's columnar batches to the Python worker via `CometColumnarPythonInput.writeNextBatchToArrowStream`. Per batch, the current path does two copies of every buffer:

1. **Comet's `FieldVector` → destination IPC root.** `copyVector` walks the source/destination trees in lockstep, allocates destination child vectors sized from the source, and `ArrowBuf.setBytes`-copies each buffer. The destination root is a child of `ArrowUtils.rootAllocator`, set up by Spark's `PythonArrowInput` trait.
2. **Destination IPC root → pipe.** `VectorUnloader.getRecordBatch` walks the root and produces an `ArrowRecordBatch`; `MessageSerializer.serialize` writes bytes to the `DataOutputStream` that feeds the Python worker's stdin.

Copy 2 is structural: Spark's transport to Python is fork + pipe + Arrow IPC, so the bytes must reach the pipe at least once. Copy 1 is droppable — we can build the IPC stream directly from Comet's vectors and skip the intermediate root entirely. After this change the path is at the single-copy floor for JVM-to-Python Arrow transport.

The cross-allocator constraint discussed in #4294 is independent of this: even after copy 1 is dropped, JVM-side zero-copy via `TransferPair` is blocked because Comet's source `FieldVector`s are imported from native via Arrow C Data Interface (their buffers route `release` through FFI), while Spark's destination IPC root is a child of `ArrowUtils.rootAllocator`. The two reference managers cannot share buffers. The path here doesn't try to: it bypasses the intermediate root entirely.

`spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala:217-241` (`serializeBatches`) already does the equivalent walk for the shuffle path: build a `VectorSchemaRoot` from Comet's `FieldVector`s, run `ArrowStreamWriter.writeBatch` against it, write the resulting bytes to a stream. The wire format is the same Arrow IPC the Python worker reads.

For `CometColumnarPythonInput`, the shape is slightly different because Spark's `PythonArrowInput` trait owns the `ArrowStreamWriter` and writes the schema header once via `writer.start()`. The per-batch contract is "append one `RecordBatch` to the already-open stream", not "write a full IPC stream". The implementation outline:

- The trait's destination `root` is still constructed (its schema is what `writer.start()` serialises into the stream header). It is never populated.
- Per batch, build an `ArrowRecordBatch` directly from Comet's source vectors:
  - Allocate one small validity buffer for the wrapping struct (all-`0xff`).
  - Collect `ArrowFieldNode`s and `ArrowBuf`s in the same depth-first order `VectorUnloader.appendNodes` uses, with Comet's `FieldVector` buffers in place of the destination root's.
  - Pass the assembled batch to `MessageSerializer.serialize`.

Net result: per-buffer `setBytes` and per-child `allocateNew` go away. The struct validity buffer (a few bytes per batch) is the only remaining JVM-side allocation on the input path.

## Additional context

- Reference comments live in `spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometColumnarPythonInput.scala` and `docs/source/user-guide/latest/pyarrow-udfs.md` (PR #4234), both already framed as "two copies, one droppable" so the doc updates will be local.
- Benchmark targets to beat: `spark/src/test/resources/pyspark/benchmark_pyarrow_udf.py`. Wide-row case (1.42x today) is where copy 1 dominates and should see the largest delta. Narrow / mixed (1.92x / 1.79x today) are bottlenecked on Python fork/IPC and should improve modestly.
- Related #4234 (introduces the operator), #4294 (proposes a more invasive allocator-tree change to enable `TransferPair`-based zero-copy; the path proposed here is orthogonal and lands the larger share of the win without that refactor).
- Originally surfaced in @mbutrovich's review on #4234 (item 3, "two copies → one copy").

sunchao:

Closes #4383.

Comet already produces Arrow-backed columnar batches during native execution. A Python `mapInArrow` or `mapInPandas` worker also consumes Arrow batches, so the data reaching the Python boundary is already in the right physical representation. However, the existing Spark 4.x Comet runner still makes a complete intermediate copy before sending each batch to the worker.

For example, even an identity UDF pays this cost:

```python
def identity(batches):
    yield from batches

result = events.mapInArrow(identity, events.schema)
```

For every input batch, the current runner allocates writer-owned destination vectors and copies each column's validity, offset, and value buffers into them. Nested arrays, structs, and maps repeat the same work recursively. The Arrow IPC writer then reads those copied buffers and writes their contents to the Python worker's pipe:

```text
Before

Comet's existing Arrow buffers
        |
        | allocate destination vectors and copy every column buffer
        v
writer-owned Arrow vector tree
        |
        | serialize Arrow IPC bytes
        v
Python worker pipe
```

That first copy does not transform the values, change the schema, or help the worker interpret the data. It only adds per-batch allocation, CPU work, and memory-bandwidth pressure. The cost becomes more visible with wide schemas, variable-width values, nested columns, or many batches per partition.

As an illustration, an 8,192-row batch containing roughly 16 MiB of Arrow column buffers currently requires roughly another 16 MiB of destination buffers and an extra 16 MiB memory copy before those same bytes are written to the pipe. After this change, the only new Arrow data buffer needed for that batch is the wrapping struct's validity bitmap: 8,192 bits, or 1 KiB. The 16 MiB example is illustrative; the exact savings depend on the batch layout.

The IPC write itself remains necessary because the Python worker is a separate process. This change removes the avoidable copy before that write.

The runner now treats Comet's existing Arrow vectors as the data source for the outgoing IPC record batch, instead of first materializing equivalent writer-owned vectors. The stream keeps its existing schema header and Python-worker contract, but the root used to advertise that schema becomes schema-only. When a batch arrives, the runner constructs the record-batch view directly over the original column buffers and hands it to the same Arrow stream writer.

```text
After

Comet's existing Arrow buffers
        |
        | describe those same buffers as an Arrow IPC record batch
        v
Arrow IPC writer
        |
        | serialize Arrow IPC bytes
        v
Python worker pipe
```

Python expects the input columns beneath one non-null struct, so the outgoing record batch adds the corresponding struct field node and a small all-valid bitmap before the existing column buffers. Arrow's normal vector-unloading logic preserves the depth-first layout of nested structs, lists, and maps, along with null counts and buffer metadata. The worker therefore receives the same logical schema, column values, and stream framing as before; the intermediate column-buffer copy simply disappears.

The native-backed vectors remain owned by their original Comet batch throughout this process. Their buffers are retained while the IPC message is written synchronously and released afterward, including when serialization fails. No buffers are transferred between Arrow allocators, no borrowed vectors are closed, and no shared-root allocator changes are required. This keeps the change separate from the allocator work discussed in #4294.

The runner also checks every input batch recursively against the advertised stream schema. Arrow types, nested layouts, dictionary encodings, and extension metadata must match; field names, nullability annotations, and ordinary metadata such as Parquet field IDs may differ. The stream header permits nullable values and nested fields beneath maps while preserving non-nullable map entries and keys. This accepts compatible native union branches without allowing changes to how the buffers are interpreted.

The optimization remains on the existing opt-in Spark 4.x execution path. Spark 3.5 continues to use its current fallback, and configurations requiring incompatible large-variable-width Arrow layouts still fall back to Spark. Regression coverage, CI suite registration, user documentation, and benchmark setup are updated to reflect the new data path.

## How was this PR tested?

The August 29 schema-compatibility follow-up was validated on Linux with JDK 17 and Spark 4.0.4 / Scala 2.13:

- **9 tests passed** in `CometArrowPythonRunnerSuite`, covering borrowed buffers, nested/null layouts, empty batches, zero-column batches, write-failure cleanup, and schema compatibility. The new safety tests reject changed widths, nesting, decimal scales, dictionaries, and extension metadata.
- **5 tests passed** in `CometMapInBatchSuite`, covering the JVM plan rewrite. The ineffective Scala "end-to-end" test was removed: it only collected Parquet and never invoked a Python UDF.
- **All 117 tests passed** in `test_pyarrow_udf.py` with real PySpark workers, Python 3.12, PyArrow 25.0.1, and pandas 3.0.5. These exercise both accelerated and fallback plans, including multiple source batches, nested values, chained UDFs, and the large-variable-type fallback. The union regressions cover 34 supported API/mode/shape cases with different nested nullability, nested names, and Parquet field IDs. They verify the complete output multiset and map nullability constraints. Struct map keys are covered through `mapInArrow`; Spark's own pandas conversion rejects that key type before the UDF runs.
- The Spark 4.0 Maven reactor packaged successfully. ScalaStyle and Spotless checks passed, as did all **178** Linux/macOS suite-registration checks from `python3 dev/ci/check-suites.py`.

The original review probes were also rerun against the rebuilt jar: all 20 accelerated/fallback schema cases passed, including the six map/name cases that failed on the previous PR head.

The benchmark configuration follow-up now uses `CometShuffleManager` with Comet shuffle left enabled at its default. A Spark 4.0.4 smoke run with 1,000 rows, no warmup, and one iteration per case completed all **12 combinations**: three workloads, both `mapInArrow` and `mapInPandas`, and vanilla/optimized execution. The existing plan assertions passed in each mode. This was a functional smoke check, not a new performance measurement.

The two new FFI tests export a 16 KiB string vector with `Data.exportVector`, import it into a second allocator, and close the original vector before serialization. They verify the shared data address, balanced reference counts and allocator usage, continued readability of the imported vector, and final release of all allocations. Both successful writes and injected write failures are covered, with the writer limited to 1 KiB so a payload copy cannot fit. The focused JVM suites also passed when rerun without a pre-existing Maven temporary directory.

The earlier Spark 4.1 local rerun was blocked during dependency resolution by timeouts from the configured Maven mirror; it did not reach the Spark-module compilation or focused tests. No repository configuration was changed. The existing CI matrix continues to cover Spark 4.0 and 4.1. Earlier validation of the original implementation included the five original Arrow tests on Spark 4.1 and compilation under Spark 3.5.

One Python regression sends 37 rows containing nested structs, arrays, maps, and nulls through a worker in six source batches:

```text
37 rows -> 7 + 7 + 7 + 7 + 7 + 2
```

It verifies both the returned values and the batch boundaries observed by Python, covering source-vector turnover.

The earlier local wide-schema end-to-end benchmark measured approximately **1.24x for `mapInArrow`** and **1.27x for `mapInPandas`**, comparing the optimized Comet path with vanilla Spark. Those workload-dependent numbers include Python and IPC overhead and do not isolate the incremental benefit of this PR.

Separately, [andygrove's review](https://github.com/apache/datafusion-comet/pull/5368#pullrequestreview-5053913289) measured the previous Comet `copyVector` + `writeBatch` path against `serializeBatch` on real Comet scan batches, writing to a discarding stream:

| Workload | Previous Comet path | Direct serialization | Reported speedup |
| --- | --- | --- | --- |
| 50 mixed columns, 500k rows | 42.9 ms | 10.8 ms | 3.95x |
| 2 bigint columns, 4M rows | 14.3 ms | 4.3 ms | 3.29x |
| Nested array/string and struct, 400k rows | 14.1 ms | 1.0 ms | 13.79x |

These isolated timings were reported by the reviewer before this follow-up and were not independently reproduced here. They use a different baseline from the end-to-end Comet-versus-Spark comparison above.

### DIFF
--- a/docs/source/user-guide/latest/pyarrow-udfs.md
+++ b/docs/source/user-guide/latest/pyarrow-udfs.md
@@ -203,9 +203,12 @@ on the unoptimized path.
   session time zone such a UDF can diverge from the unoptimized path. Set
   `spark.comet.exec.pyarrowUDF.enabled=false` for those UDFs.
 - `spark.sql.execution.arrow.useLargeVarTypes=true` is not supported. With this conf enabled,
-  Spark expects `StringType` and `BinaryType` to use Arrow's 8-byte-offset variants, while
-  Comet's source vectors always use 4-byte offsets. `EliminateRedundantTransitions` skips the
-  rewrite for this incompatible layout and vanilla Spark handles the operation.
+  Spark supplies `large_string` and `large_binary` input columns with 8-byte offsets. Native
+  Comet vectors use 4-byte offsets, and direct serialization advertises their matching `string`
+  and `binary` types. This produces a valid IPC stream, but does not preserve the input types
+  requested by the configuration. `EliminateRedundantTransitions` therefore skips the rewrite
+  and vanilla Spark handles the operation. Comet can read `large_string` and `large_binary`
+  columns returned by a Python worker; that output support does not widen the input vectors.
 - Comet writes input Arrow IPC record batches directly from its existing vector buffers. The
   only additional Arrow buffer is the validity bitmap for the non-null struct that wraps the
   input columns. Writing the IPC bytes to the Python worker's pipe still requires one copy;
--- a/docs/source/user-guide/latest/pyarrow-udfs.md
+++ b/docs/source/user-guide/latest/pyarrow-udfs.md
@@ -203,18 +203,11 @@ on the unoptimized path.
   session time zone such a UDF can diverge from the unoptimized path. Set
   `spark.comet.exec.pyarrowUDF.enabled=false` for those UDFs.
 - `spark.sql.execution.arrow.useLargeVarTypes=true` is not supported. With this conf enabled,
-  Spark widens `StringType` and `BinaryType` to Arrow's 8-byte-offset variants in the
-  destination IPC root, while Comet's source vectors always use 4-byte offsets. The buffer-copy
-  path cannot bridge that mismatch, so `EliminateRedundantTransitions` skips the rewrite and
-  vanilla Spark handles the operation.
-- Each batch is copied twice on the JVM side: once from Comet's vectors into Spark's
-  destination IPC root (per-buffer `setBytes`), and a second time inside the IPC writer when
-  `VectorUnloader` / `MessageSerializer.serialize` walks the root and writes bytes to the
-  pipe to the Python worker. The pipe write is structural (Spark's transport to Python is
-  fork + pipe + Arrow IPC, so the buffer bytes must reach the pipe at least once); dropping
-  the first copy by serialising directly from Comet's vectors is tracked in
-  [#4294](https://github.com/apache/datafusion-comet/issues/4294). Even after that,
-  true zero-copy at the JVM boundary is blocked because Comet's source `FieldVector`s are
-  imported from native via Arrow C Data Interface (their buffers route `release` through FFI),
-  while Spark's destination IPC root is a child of `ArrowUtils.rootAllocator`. The two
-  reference managers cannot share buffers via `TransferPair`.
+  Spark expects `StringType` and `BinaryType` to use Arrow's 8-byte-offset variants, while
+  Comet's source vectors always use 4-byte offsets. `EliminateRedundantTransitions` skips the
+  rewrite for this incompatible layout and vanilla Spark handles the operation.
+- Comet writes input Arrow IPC record batches directly from its existing vector buffers. The
+  only additional Arrow buffer is the validity bitmap for the non-null struct that wraps the
+  input columns. Writing the IPC bytes to the Python worker's pipe still requires one copy;
+  that copy is inherent to Spark's process-based Python transport. This path does not transfer
+  buffers between Arrow allocators or change their ownership.
--- a/spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
@@ -112,10 +112,9 @@ case class EliminateRedundantTransitions(session: SparkSession)
       // 4.1+ matches the renamed `MapInArrowExec`.
       //
       // Falls back to vanilla Spark when `spark.sql.execution.arrow.useLargeVarTypes` is enabled:
-      // CometArrowPythonRunnerBase.copyVector does raw `setBytes` on each Arrow buffer, but Comet's
-      // source string/binary vectors always use 4-byte offsets while the destination root is
-      // allocated with 8-byte offsets when this conf is on. The buffer counts match but the
-      // offset width does not, so a direct memcpy would corrupt the offsets.
+      // Comet's source string/binary vectors use 4-byte offsets, while Spark expects 8-byte
+      // offsets when this conf is on. Direct IPC serialization cannot change the source vectors'
+      // physical layout, so forwarding them under Spark's widened schema would corrupt the stream.
       //
       // `EligibleMapInBatch` matches whenever the operator would run natively if the feature were
       // enabled. When it is disabled (the default) we leave the vanilla Spark operator in place
--- a/spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
@@ -112,9 +112,10 @@ case class EliminateRedundantTransitions(session: SparkSession)
       // 4.1+ matches the renamed `MapInArrowExec`.
       //
       // Falls back to vanilla Spark when `spark.sql.execution.arrow.useLargeVarTypes` is enabled:
-      // Comet's source string/binary vectors use 4-byte offsets, while Spark expects 8-byte
-      // offsets when this conf is on. Direct IPC serialization cannot change the source vectors'
-      // physical layout, so forwarding them under Spark's widened schema would corrupt the stream.
+      // Native Comet string/binary vectors use 4-byte offsets. The IPC schema follows these
+      // vectors, so the stream is internally consistent, but the worker would receive string /
+      // binary instead of the large_string / large_binary input types requested by this conf.
+      // Keep the fallback to preserve Spark's input type contract.
       //
       // `EligibleMapInBatch` matches whenever the operator would run natively if the feature were
       // enabled. When it is disabled (the default) we leave the vanilla Spark operator in place
--- a/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
+++ b/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
@@ -176,10 +176,7 @@ private[python] trait CometArrowPythonRunnerBase
             .getValueVector
             .asInstanceOf[FieldVector]
         }
-        val childNames = inputStructType.fieldNames
-        val batchFields = sourceVectors.zipWithIndex.map { case (vector, i) =>
-          renamed(vector.getField, childNames(i), forceNullable = true)
-        }
+        val batchFields = sourceVectors.map(_.getField)
 
         if (arrowWriter == null) {
           // Build the schema-only struct root once from the first batch's child fields.
@@ -192,13 +189,17 @@ private[python] trait CometArrowPythonRunnerBase
           // rather than the session zone vanilla Spark would label it with; this is a documented
           // limitation (see pyarrow-udfs.md), not a value difference, since the stored instant is
           // identical.
-          streamFields = batchFields
+          val childNames = inputStructType.fieldNames
+          streamFields = batchFields.zipWithIndex.map { case (field, i) =>
+            renamed(field, childNames(i), forceNullable = true)
+          }
           startWriter(streamFields, dataOut)
         }
 
-        // Compare normalized fields: union branches may differ only in nested nullability.
+        // Union branches may differ in names, nullability, or descriptive metadata. Only
+        // differences that change how the advertised schema interprets the buffers are invalid.
         require(
-          batchFields == streamFields,
+          CometArrowPythonRunnerBase.hasCompatibleSchema(streamFields, batchFields),
           s"Arrow input schema changed between batches: expected $streamFields, got $batchFields")
 
         CometArrowPythonRunnerBase.serializeBatch(
@@ -290,11 +291,14 @@ private[python] trait CometArrowPythonRunnerBase
    * materialize the struct. Keeping the type and structure intact means the advertised schema
    * still mirrors the Comet source vectors serialized directly into each record batch.
    */
-  private def renamed(field: Field, name: String, forceNullable: Boolean): Field = {
-    // A Map's descendants must keep their original nullability: Arrow requires the entries struct
-    // (and its key) to be non-nullable, and `MapVector.createVector` rejects a nullable entries
-    // struct. Stop forcing nullable once we enter a Map subtree.
-    val childrenForceNullable = forceNullable && !field.getType.isInstanceOf[ArrowType.Map]
+  private def renamed(
+      field: Field,
+      name: String,
+      forceNullable: Boolean,
+      isMapEntry: Boolean = false): Field = {
+    // Map entries and keys must stay non-nullable, but values and fields nested inside a
+    // complex key may be nullable. Do not propagate the restriction through the whole subtree.
+    val isMap = field.getType.isInstanceOf[ArrowType.Map]
     val children = field.getChildren
     val newChildren =
       if (children.isEmpty) children
@@ -307,7 +311,8 @@ private[python] trait CometArrowPythonRunnerBase
           renamed(
             child,
             if (child.getName == null) s"_$idx" else child.getName,
-            childrenForceNullable)
+            forceNullable = !isMap && !(isMapEntry && idx == 0),
+            isMapEntry = isMap)
         }.asJava
     // Force the field nullable where allowed. Comet's FFI-imported vectors may carry a
     // non-nullable Arrow `Field` even for columns that contain nulls (Comet uses positional schema
@@ -323,6 +328,22 @@ private[python] trait CometArrowPythonRunnerBase
 }
 
 private[python] object CometArrowPythonRunnerBase {
+
+  // Extensions can change interpretation even when their underlying storage types match.
+  private val extensionMetadataKeys = Seq(
+    ArrowType.ExtensionType.EXTENSION_METADATA_KEY_NAME,
+    ArrowType.ExtensionType.EXTENSION_METADATA_KEY_METADATA)
+
+  /** Names, nullability and ordinary field metadata do not change the IPC buffer layout. */
+  private[python] def hasCompatibleSchema(expected: Seq[Field], actual: Seq[Field]): Boolean = {
+    expected.size == actual.size && expected.zip(actual).forall { case (left, right) =>
+      left.getType == right.getType &&
+      left.getDictionary == right.getDictionary &&
+      extensionMetadataKeys.forall(key =>
+        left.getMetadata.get(key) == right.getMetadata.get(key)) &&
+      hasCompatibleSchema(left.getChildren.asScala.toSeq, right.getChildren.asScala.toSeq)
+    }
+  }
 
   /**
    * Serialize source vectors directly beneath the non-null struct advertised in the IPC stream.
--- a/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
+++ b/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
@@ -20,7 +20,7 @@
 package org.apache.spark.sql.execution.python
 
 import java.io.{DataInputStream, DataOutputStream}
-import java.nio.channels.{Channels, WritableByteChannel}
+import java.nio.channels.Channels
 import java.util.ArrayList
 import java.util.concurrent.atomic.AtomicBoolean
 
@@ -29,8 +29,8 @@ import scala.jdk.CollectionConverters._
 import org.apache.arrow.memory.{ArrowBuf, BufferAllocator}
 import org.apache.arrow.vector.{FieldVector, VectorSchemaRoot, VectorUnloader}
 import org.apache.arrow.vector.complex.StructVector
-import org.apache.arrow.vector.ipc.{ArrowStreamReader, ArrowStreamWriter}
-import org.apache.arrow.vector.ipc.message.{ArrowFieldNode, ArrowRecordBatch}
+import org.apache.arrow.vector.ipc.{ArrowStreamReader, ArrowStreamWriter, WriteChannel}
+import org.apache.arrow.vector.ipc.message.{ArrowFieldNode, ArrowRecordBatch, MessageSerializer}
 import org.apache.arrow.vector.types.pojo.{ArrowType, Field, FieldType}
 import org.apache.spark.{SparkEnv, TaskContext}
 import org.apache.spark.api.python.{BasePythonRunner, PythonRDD, PythonWorker, SpecialLengths}
@@ -109,7 +109,7 @@ private[python] trait CometArrowPythonRunnerBase
       private val allocator =
         CometArrowAllocator.newChildAllocator(s"stdout writer for $pythonExec", 0, Long.MaxValue)
       private var currentGroup: Iterator[ColumnarBatch] = _
-      private var arrowWriter: CometArrowPythonRunnerBase.DirectArrowStreamWriter = _
+      private var arrowWriter: ArrowStreamWriter = _
       private var writeRoot: VectorSchemaRoot = _
 
       // The runner's input schema is a single struct column ("struct") whose children are the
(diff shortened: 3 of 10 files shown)


spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
```diff
@@ -107,9 +109,8 @@ private[python] trait CometArrowPythonRunnerBase
       private val allocator =
         CometArrowAllocator.newChildAllocator(s"stdout writer for $pythonExec", 0, Long.MaxValue)
       private var currentGroup: Iterator[ColumnarBatch] = _
-      private var arrowWriter: ArrowStreamWriter = _
+      private var arrowWriter: CometArrowPythonRunnerBase.DirectArrowStreamWriter = _
```
peterxcli:
```suggestion
      private var arrowWriter: ArrowStreamWriter = _
```

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
```diff
@@ -167,49 +170,37 @@ private[python] trait CometArrowPythonRunnerBase
...
-          structVec.getValidityBuffer.memoryAddress(),
-          0xff.toByte,
-          validityBytes)
-        writeRoot.setRowCount(numRows)
-        arrowWriter.writeBatch()
+        CometArrowPythonRunnerBase.writeDirectBatch(
+          arrowWriter,
+          sourceVectors,
+          cometBatch.numRows(),
+          allocator)
```
peterxcli: then use `MessageSerializer.serialize`
```suggestion
        // This root only borrows the Comet vectors; do not close it.
	  val sourceRoot =
	    new VectorSchemaRoot(sourceVectors.map(_.getField).asJava, sourceVectors.asJava, numRows)
	  val sourceBatch = new VectorUnloader(sourceRoot).getRecordBatch
	
	  try {
	    val validityBytes = (numRows.toLong + 7L) / 8L
	    val structValidity = allocator.buffer(validityBytes)
	
	    try {
	      if (validityBytes > 0) {
	        structValidity.setOne(0L, validityBytes)
	      }
	      structValidity.writerIndex(validityBytes)
	
	      val nodes = new ArrayList[ArrowFieldNode](sourceBatch.getNodes.size() + 1)
	      nodes.add(new ArrowFieldNode(numRows, 0))
	      nodes.addAll(sourceBatch.getNodes)
	
	      val buffers = new ArrayList[ArrowBuf](sourceBatch.getBuffers.size() + 1)
	      buffers.add(structValidity)
	      buffers.addAll(sourceBatch.getBuffers)
	
	      val wrappedBatch = new ArrowRecordBatch(
	        numRows,
	        nodes,
	        buffers,
	        sourceBatch.getBodyCompression,
	        sourceBatch.getVariadicBufferCounts,
	        true)
	
	      try {
	        MessageSerializer.serialize(
	          new WriteChannel(Channels.newChannel(dataOut)),
	          wrappedBatch)
	      } finally {
	        wrappedBatch.close()
	      }
	    } finally {
	      structValidity.close()
	    }
	  } finally {
	    sourceBatch.close()
	  }
	
	  pythonMetrics("pythonDataSent") += dataOut.size() - startData
	  true
}
```



ref: https://github.com/apache/spark/blob/v4.1.3/sql/core/src/main/scala/org/apache/spark/sql/execution/python/PythonArrowInput.scala#L134-L166

peterxcli (commented):
how about we remove `writeDirectBatch` and use `MessageSerializer.serialize` to replace the `writer.writeBatch`, which needs to `copyVector(cometBatch -> structVec)` previously,  then we could restore to use `ArrowStreamWriter`,  delete `DirectArrowStreamWriter` and .

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
sunchao: Updated in 64a0114a7. The runner now uses a standard `ArrowStreamWriter` for the stream schema and end marker, and `DirectArrowStreamWriter` has been removed.

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
sunchao: Updated in 64a0114a7. Each batch now gets a fresh `WriteChannel` over `dataOut` and is written with `MessageSerializer.serialize(writeChannel, wrappedBatch)`, matching the Spark 4.1 implementation. I kept the batch assembly and retain/release logic in a small `serializeBatch` helper so the allocator, nested/null, multi-batch, and failure-cleanup regression tests can exercise it directly. All five tests pass on both Spark 4.0 and Spark 4.1.

peterxcli (approved):
LGTM! thanks for the update.

andygrove (commented):
I checked this out locally and ran the JVM suites plus some extra probes on real Comet batches. The change looks correct to me. I could not run `test_pyarrow_udf.py` or the benchmark script here because I have no PySpark install available on this machine, so I leaned on CI for those and wrote JVM-level substitutes instead.

What I ran: `CometArrowPythonRunnerSuite` and `CometMapInBatchSuite` on both Spark 4.0 / Scala 2.13 and Spark 4.1, all green. On top of that I wrote a throwaway suite that takes real Comet scan batches, runs them through `serializeBatch`, decodes the IPC bytes back with an `ArrowStreamReader`, and compares every cell against the source vectors. That covered mixed primitives with nulls, decimals, timestamps, dates, an all-null column, nested list/struct/map/array-of-array at `spark.comet.batchSize=7` so batches turn over, low cardinality strings, an empty result set, and sliced vectors. All of it round trips exactly. I also snapshotted reference counts recursively across the whole FFI-imported vector tree before and after `serializeBatch` and they come back exactly balanced, and confirmed the source vectors are not mutated.

On performance, I timed the serialization step on its own against real Comet scan batches, writing into a discarding stream, so the numbers isolate the part this PR actually changes. Old `copyVector` plus `writeBatch` versus the new `serializeBatch`, order alternated per batch with a warmup iteration discarded:

| workload | old | new | speedup |
|---|---|---|---|
| 50 column mixed, 500k rows | 42.9 ms | 10.8 ms | 3.95x |
| 2 bigint columns, 4M rows | 14.3 ms | 4.3 ms | 3.29x |
| nested `array<string>` and `struct`, 400k rows | 14.1 ms | 1.0 ms | 13.79x |

Record batch payload sizes came out identical, the only byte difference being the one-time schema message. The 1.24x / 1.27x end-to-end numbers in the description include the Python interpreter and the pipe, so they understate the part this PR controls. Worth adding a JVM-isolated number like the above to the description.

One thing that is not from this PR but is load bearing for it. The `end-to-end: rewrite-on output matches rewrite-off output` test in `CometMapInBatchSuite` collects `spark.read.parquet(pathStr)` twice and never calls `mapInArrow`, so the `pyarrowUDF.enabled` flag it toggles cannot affect either plan. It passes unconditionally, and it is skipped entirely unless `PYSPARK_PYTHON` is set. Since this PR rewrites exactly what that test is meant to guard, would you either add the missing `.mapInArrow(...)` or drop it and rely on `test_pyarrow_udf.py`?

Three smaller comments inline.

spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
```diff
@@ -112,10 +112,9 @@ case class EliminateRedundantTransitions(session: SparkSession)
       // 4.1+ matches the renamed `MapInArrowExec`.
       //
       // Falls back to vanilla Spark when `spark.sql.execution.arrow.useLargeVarTypes` is enabled:
-      // CometArrowPythonRunnerBase.copyVector does raw `setBytes` on each Arrow buffer, but Comet's
-      // source string/binary vectors always use 4-byte offsets while the destination root is
-      // allocated with 8-byte offsets when this conf is on. The buffer counts match but the
-      // offset width does not, so a direct memcpy would corrupt the offsets.
+      // Comet's source string/binary vectors use 4-byte offsets, while Spark expects 8-byte
+      // offsets when this conf is on. Direct IPC serialization cannot change the source vectors'
+      // physical layout, so forwarding them under Spark's widened schema would corrupt the stream.
```
andygrove: The comment says forwarding Comet's 4-byte-offset vectors under Spark's widened schema "would corrupt the stream". That was true of `copyVector`, but after this change the advertised schema comes from `sourceVectors.map(_.getField)`, so the header and the buffers agree and the stream stays internally consistent.

I think the fallback should still stay, but for different reasons: the worker would see `string` where `useLargeVarTypes=true` promises `large_string`, and the worker's `large_*` output would then come back through `ArrowStreamReader` into `CometVector.getVector`. Could you update the comment here and the matching bullet in `pyarrow-udfs.md` to say that instead?

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
```diff
@@ -167,49 +168,37 @@ private[python] trait CometArrowPythonRunnerBase
...
           // identical.
           val childNames = inputStructType.fieldNames
-          val childFields = (0 until cometBatch.numCols()).map { i =>
-            val vecField =
-              cometBatch.column(i).asInstanceOf[CometDecodedVector].getValueVector.getField
-            renamed(vecField, childNames(i), forceNullable = true)
+          val childFields = sourceVectors.zipWithIndex.map { case (vector, i) =>
+            renamed(vector.getField, childNames(i), forceNullable = true)
           }
           startWriter(childFields, dataOut)
```
andygrove: `copyVector` used to fail loudly via its `require(srcBufs.size == dstBufs.size)` if a later batch's layout stopped matching the destination tree. Now that the schema header is written once from the first batch and each subsequent batch's buffers go straight to the wire, a layout change would instead produce a well-formed record batch that the worker silently misreads.

I could not find a path where Comet changes layout mid-partition, so this may well be unreachable, but would you consider capturing the child fields in `startWriter` and asserting `sourceVectors.map(_.getField)` still matches them per batch? It keeps the fail-loud behaviour for about three lines.

spark/src/test/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerSuite.scala
```diff
@@ -0,0 +1,329 @@
...
+    val reader = new ArrowStreamReader(new ByteArrayInputStream(bytes), allocator)
+    try {
+      f(reader)
+    } finally {
+      reader.close()
+      allocator.close()
+    }
+  }
+
+  test("direct batches retain borrowed buffers without copying them into the writer allocator") {
```
andygrove, line 127: All five tests here build their source vectors with a plain `RootAllocator`. In production the source is always FFI-imported, where the buffers are owned by `ReferenceCountedArrowArray` and the validity bitmap is synthesized by `BitVectorHelper.loadValidityBuffer` at import time. That is precisely the reference manager that the refcount and `getAllocatedMemory` assertions are trying to pin down.

Would you add one case that exports a vector with `Data.exportVector` and re-imports it into a second allocator before calling `serializeBatch`? I checked this locally against real Comet scan batches and the counts do balance, so this is a coverage question rather than a bug.

spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
sunchao: Updated in 1de4491c. The code comment, user guide, and fallback test docstring now explain that the IPC schema matches the source vectors, so the stream is valid, but it does not provide the large input types requested by the configuration. The fallback stays in place to preserve that input contract.

I also clarified that Comet already reads large_string / large_binary output through Utils and CometPlainVector; that output support does not widen the input vectors.

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
sunchao: Added in 1de4491c. The writer snapshots the first batch's raw Arrow fields and requires every subsequent batch to match before serializing it. I compare raw fields with raw fields because the advertised fields have their names and nullability normalized.

The full Spark 4.0 Python suite passed, including the multiple-batch, nested-source, and chained-UDF cases.

spark/src/test/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerSuite.scala
sunchao: Added two FFI cases in 1de4491c: successful serialization and an injected write failure. Each exports with Data.exportVector, imports into a second allocator, verifies the shared data address, and closes the original vector before serialization so the imported ownership keeps the buffers alive.

Both cases check balanced buffer reference counts, unchanged allocator usage, readable imported values, and zero outstanding allocations after the imported vector is closed. A 16 KiB payload is serialized with a 1 KiB writer allocator to rule out a payload copy. All seven Arrow tests pass on Spark 4.0.

sunchao:
Thanks for the detailed review and the additional probes. Addressed the follow-ups in 1de4491c.

I removed the ineffective Scala end-to-end test. CometMapInBatchSuite now keeps its five plan-rule checks, while actual worker execution is covered by test_pyarrow_udf.py, which exercises the accelerated and fallback plans and checks their results. All 83 Python cases passed locally on Spark 4.0.4, including source-batch turnover and chained UDFs.

The three inline requests are addressed as described in their threads. I also added your isolated JVM measurements to the PR description with attribution, explicitly separated from the earlier Comet-versus-Spark end-to-end numbers and marked as not independently reproduced in this update.

Validation: 12 focused JVM tests passed (7 Arrow serialization tests and 5 plan-rule tests), the Spark 4.0 Maven reactor packaged successfully, ScalaStyle and Spotless passed, and all 178 suite-registration checks passed. The FFI tests also passed with no pre-existing Maven temporary directory. The local Spark 4.1 rerun was blocked by dependency-mirror timeouts before the Spark-module compilation/tests; the existing CI matrix covers both Spark profiles.

spark/src/test/resources/pyspark/benchmark_pyarrow_udf.py
```diff
@@ -84,6 +84,8 @@ def _build_spark() -> SparkSession:
         .config("spark.plugins", "org.apache.spark.CometPlugin")
         .config("spark.comet.enabled", "true")
         .config("spark.comet.exec.enabled", "true")
+        # Keep Comet's scan and execution rules active with Spark's default shuffle manager.
+        .config("spark.comet.shuffle.enabled", "false")
```
comphead: why shuffle enabled false? Earlier we assessed that this mode should be only for testing mode

spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
```diff
@@ -112,10 +112,10 @@ case class EliminateRedundantTransitions(session: SparkSession)
       // 4.1+ matches the renamed `MapInArrowExec`.
       //
       // Falls back to vanilla Spark when `spark.sql.execution.arrow.useLargeVarTypes` is enabled:
-      // CometArrowPythonRunnerBase.copyVector does raw `setBytes` on each Arrow buffer, but Comet's
-      // source string/binary vectors always use 4-byte offsets while the destination root is
-      // allocated with 8-byte offsets when this conf is on. The buffer counts match but the
-      // offset width does not, so a direct memcpy would corrupt the offsets.
+      // Native Comet string/binary vectors use 4-byte offsets. The IPC schema follows these
```
comphead, line 115: not sure why this was changed, it doesn't relate to PR

docs/source/user-guide/latest/pyarrow-udfs.md
```diff
@@ -203,18 +203,14 @@ on the unoptimized path.
...
-  `VectorUnloader` / `MessageSerializer.serialize` walks the root and writes bytes to the
-  pipe to the Python worker. The pipe write is structural (Spark's transport to Python is
-  fork + pipe + Arrow IPC, so the buffer bytes must reach the pipe at least once); dropping
-  the first copy by serialising directly from Comet's vectors is tracked in
-  [#4294](https://github.com/apache/datafusion-comet/issues/4294). Even after that,
-  true zero-copy at the JVM boundary is blocked because Comet's source `FieldVector`s are
-  imported from native via Arrow C Data Interface (their buffers route `release` through FFI),
-  while Spark's destination IPC root is a child of `ArrowUtils.rootAllocator`. The two
-  reference managers cannot share buffers via `TransferPair`.
+  Spark supplies `large_string` and `large_binary` input columns with 8-byte offsets. Native
```
comphead, line 206: mmm, does that mean we need to map `large_string` to `LargeStringArray` in native?

comphead (approved):
Thanks @sunchao I dont see blockers for this PR, rather some questions

spark/src/test/resources/pyspark/benchmark_pyarrow_udf.py
sunchao: Updated in 5dd4782c. The benchmark now configures `CometShuffleManager` and leaves Comet shuffle enabled at its default, instead of disabling shuffle to keep the extension active with Spark's default manager.

Smoke-tested all three workloads with both `mapInArrow` and `mapInPandas`, in vanilla and optimized modes, on Spark 4.0.4. All 12 combinations completed and the existing plan assertions passed. This was a small functional run, not a new performance measurement.

spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala
sunchao: The old comment described raw buffer copying in `copyVector`, which this PR removes. I updated the explanation for direct serialization: the IPC schema matches the source buffers, so the stream is valid, but it would expose `string` / `binary` instead of the large input types requested by `useLargeVarTypes=true`.

Only the explanation changed here; the existing fallback behavior is unchanged.

docs/source/user-guide/latest/pyarrow-udfs.md
sunchao: `large_string` corresponds to Arrow's `LargeUtf8` / `LargeStringArray`, but no new native mapping is needed for this PR because the large-type configuration still falls back to Spark.

Supporting it on the accelerated path would require producing or converting inputs to actual 64-bit-offset arrays, not just changing the schema label. Comet already handles large string/binary output from Python on the JVM side; that is separate from widening the inputs sent to Python. I have left that additional support out of this PR and kept the fallback.

sunchao:
Merged to main, thanks!

PR #5369, opened 2026-08-15
fix: preserve Catalyst nullability and field IDs in native Parquet writes

### LINKED ISSUE #5305: Native Parquet writer derives schema nullability and field IDs from Arrow rather than Catalyst
### Describe the bug

`ParquetWriterExec` builds the output Parquet schema by renaming the fields of the incoming Arrow schema:

```rust
let fields: Vec<_> = input_schema.fields().iter().enumerate()
    .map(|(i, field)| Arc::new(field.as_ref().clone().with_name(&column_names[i])))
    .collect();
```

So two pieces of schema information come from whatever the Comet scan/serde produced, not from the Catalyst schema Spark intends to write:

1. **Nullability.** Spark writes each column `required` or `optional` according to the Catalyst `StructField.nullable` of the *target* schema. Comet inherits the Arrow field's nullability, which can differ — a column Spark would mark `required` may come out `optional`, or vice versa. That changes the file's schema as seen by every reader, and a spuriously `required` column is the dangerous direction.

2. **Field IDs.** `spark.sql.parquet.fieldId.write.enabled` makes Spark emit Parquet field IDs from the Catalyst metadata. Comet emits none, so `ParquetFieldIdIOSuite`-style round trips and any consumer keyed on field ID (Iceberg, Delta) lose them silently.

### Expected behavior

Take nullability and field IDs from the Catalyst schema. `WriteJobDescription.dataColumns` is already plumbed through to `CometWriteFilesExec` for column names (#5293), so the attribute list with its `nullable` flag and metadata is available at the same point — it needs to reach the native side as a serialized schema rather than just names.

### Additional context

Found while reviewing #5293. Same family as #3425 (INT96) and #3427 (footer metadata keys): the written file is readable but is not byte-comparable with what Spark writes, which matters most for files other engines read — see #4658 and the plan to write Iceberg data files natively.

sunchao:

Closes #5305.

A Parquet write has two different schemas: the schema of the Arrow batches arriving at the native writer, and the schema Spark intends to persist in the Parquet file. The first describes how data is transported through execution. The second is the durable contract every future reader sees in the file footer. Comet currently treats them as the same schema, even though they are not.

The native writer receives data through a placeholder scan that marks top-level fields nullable and does not retain their Catalyst field metadata. It then builds the Parquet schema by renaming those incoming Arrow fields. The values are written successfully, but Spark's intended nullability and Parquet field IDs are already gone.

For example, suppose the Catalyst write schema is:

```text
required_number: LONG NOT NULL   [field ID = 11]
optional_text:   STRING NULL     [field ID = 22]
```

Today, Comet writes a footer equivalent to:

```text
message schema {
  optional int64 required_number;
  optional binary optional_text (STRING);
}
```

Spark expects:

```text
message schema {
  required int64 required_number = 11;
  optional binary optional_text (STRING) = 22;
}
```

The file may still look correct when read back immediately by column name, which makes the problem easy to miss. However, `required_number` has silently become nullable, and both columns have lost the stable identities needed for schema evolution. For example, an evolved read schema should still identify the same columns after both a rename and a reorder:

```text
Written schema:   original_number [11], original_text [22]
Requested schema: renamed_text [22],    renamed_number [11]
```

Without field IDs in the footer, Spark cannot recover that mapping. Its field-ID-aware reader fails by default with:

```text
Spark read schema expects field Ids, but Parquet file schema doesn't contain any field Ids.
```

Setting `spark.sql.parquet.fieldId.read.ignoreMissing=true` avoids the exception but can return nulls instead, turning a visible compatibility failure into incorrect results.

Nested collections introduce another, less obvious version of the same problem. A schema using Delta-style column-mapping metadata can assign separate IDs to the synthetic fields inside lists and maps:

```text
tags [200]
  element [201]

attrs [300]
  key   [301]
  value [302]
```

Because `element`, `key`, and `value` are not Catalyst `StructField`s, their IDs are stored on the nearest parent field under `parquet.field.nested.ids`. Copying only ordinary struct-field IDs would still produce incomplete Parquet schemas for these nested cases.

This PR makes the Catalyst write schema, rather than the intermediate Arrow input schema, the source of truth for the Parquet file. At the Spark/native boundary, the existing writer operator now carries the complete target schema alongside the incoming data. The native side reconstructs that schema and uses it to initialize the Parquet writer, so the footer receives Spark's actual column names, required/optional annotations, and field IDs.

The conceptual change is:

```text
Before: Arrow input schema -> rename columns -> Parquet file schema

After:  Catalyst target schema -> native writer schema -> Parquet file schema
        Arrow input batches ---------------------------> row data
```

The schema travels recursively rather than stopping at top-level columns. Ordinary nested struct fields retain their own IDs and nullability, while list-element and map-key/value IDs are recovered from the nearest parent field's Delta-style metadata. Relative collection paths are preserved through nesting and restart at each nested struct, matching how those IDs are represented in Catalyst.

Incoming Arrow batches can still use their existing execution-oriented field names and metadata. They are aligned with the target write schema without weakening validation of physical data types or nullability. Spark's `spark.sql.parquet.fieldId.write.enabled` setting remains authoritative: its default-enabled behavior preserves IDs, and explicitly disabling it suppresses IDs consistently at every nesting level. Older serialized plans without a target write schema retain their existing behavior.

The change applies directly to Comet's existing Spark 3.5/4.0 native Parquet writer. It preserves compatibility with Delta-shaped and Iceberg-style field-ID schemas, but does not claim to replace or accelerate Delta's separate transactional writer. It is independent of the writer-seam refactor in #5293.

## How was this PR tested?

The new Scala regressions execute the actual `CometNativeWriteExec` path and inspect the footer of every generated Parquet part file. They verify required versus optional fields, top-level and nested IDs, list/map synthetic IDs, the default/explicitly enabled/disabled field-ID settings, and a real Spark read with both renamed and reordered columns. The read-back test disables Comet so the resulting file is interpreted by Spark's own field-ID reader.

The nullability regression was first reproduced against the unfixed writer, where it failed with `OPTIONAL did not equal REQUIRED`.

The complete Spark 3.5 native writer suite passes, including all three new regressions and existing coverage for complex types, compression, and save modes:

```bash
./mvnw -o -Pspark-3.5 test -Dtest=none \
  -Dsuites=org.apache.comet.parquet.CometParquetWriterSuite
```

Result: **33 tests passed**.

The focused native Parquet regressions also pass under Spark 4.0:

```bash
./mvnw -o -Pspark-4.0 test -Dtest=none \
  '-Dsuites=org.apache.comet.parquet.CometParquetWriterSuite native parquet'
```

Result: **3 tests passed**.

Rust coverage validates absent versus zero-valued field IDs, nested list/map metadata, preserved nullability, and a real written Parquet footer:

```bash
cargo test -p datafusion-comet execution::serde::tests --lib
cargo test -p datafusion-comet execution::operators::parquet_writer::tests --lib
cargo fmt --all -- --check
```

Results: **3 serde tests passed** and **2 writer tests passed**; four pre-existing HDFS integration tests remain ignored because they require a running HDFS cluster. Maven Spotless and Scalastyle checks also passed under both Spark profiles.

### DIFF
--- a/native/core/src/execution/operators/parquet_writer.rs
+++ b/native/core/src/execution/operators/parquet_writer.rs
@@ -543,8 +543,9 @@ impl ExecutionPlan for ParquetWriterExec {
                 // Rename columns in the batch to match output schema
                 let renamed_batch = if !column_names.is_empty() {
                     // Collection field IDs exist on the target schema, not on arrays produced by
-                    // the placeholder Scan. Ignore nested field names/metadata during validation
-                    // while still checking data types and nested/top-level nullability.
+                    // the placeholder Scan. Both schemas use the same Catalyst data types, and
+                    // disabling field-name matching still recursively validates nested nullability;
+                    // only nested field names and metadata are ignored.
                     RecordBatch::try_new_with_options(
                         Arc::clone(&schema_for_write),
                         batch.columns().to_vec(),
@@ -710,6 +711,68 @@ mod tests {
         let element = list.get_fields()[0].get_fields()[0].get_basic_info();
         assert_eq!(element.repetition(), Repetition::OPTIONAL);
         assert_eq!(element.id(), 23);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_parquet_writer_rejects_mismatched_nested_nullability() -> Result<()> {
+        let values = ListArray::from_iter_primitive::<Int32Type, _, _>([
+            Some(vec![Some(1)]),
+            Some(vec![Some(2)]),
+        ]);
+        let DataType::List(input_element) = values.data_type() else {
+            panic!("expected list input");
+        };
+        assert!(input_element.is_nullable());
+
+        let input_schema = Arc::new(Schema::new(vec![Field::new(
+            "col_0",
+            values.data_type().clone(),
+            true,
+        )]));
+        let batch = RecordBatch::try_new(Arc::clone(&input_schema), vec![Arc::new(values)])?;
+
+        let target_element = Field::new("element", DataType::Int32, false).with_metadata(
+            HashMap::from([(PARQUET_FIELD_ID_META_KEY.to_string(), "23".to_string())]),
+        );
+        let output_schema = Arc::new(Schema::new(vec![Field::new(
+            "values",
+            DataType::List(Arc::new(target_element)),
+            true,
+        )]));
+
+        let memory_source = MemorySourceConfig::try_new(&[vec![batch]], input_schema, None)?;
+        let input = Arc::new(DataSourceExec::new(Arc::new(memory_source)));
+        let temp_dir = tempfile::tempdir()?;
+        let work_dir = format!("file://{}", temp_dir.path().display());
+        let writer = ParquetWriterExec::try_new(
+            input,
+            work_dir.clone(),
+            work_dir,
+            None,
+            None,
+            ParquetCompression::None,
+            0,
+            vec!["values".to_string()],
+            Some(output_schema),
+            HashMap::new(),
+        )?;
+
+        let mut stream = writer.execute(0, SessionContext::new().task_ctx())?;
+        let error = stream
+            .try_next()
+            .await
+            .expect_err("mismatched nested nullability must be rejected");
+        let message = error.to_string();
+        assert!(
+            message.contains("Failed to rename batch columns"),
+            "unexpected error: {message}"
+        );
+        assert!(
+            message.contains("column types must match schema types"),
+            "unexpected error: {message}"
+        );
 
         Ok(())
     }
--- a/native/core/src/execution/operators/parquet_writer.rs
+++ b/native/core/src/execution/operators/parquet_writer.rs
@@ -34,7 +34,7 @@ use crate::parquet::parquet_support::is_hdfs_scheme;
 #[cfg(feature = "hdfs-opendal")]
 use crate::parquet::parquet_support::{create_hdfs_operator, prepare_object_store_with_configs};
 use arrow::datatypes::{Schema, SchemaRef};
-use arrow::record_batch::RecordBatch;
+use arrow::record_batch::{RecordBatch, RecordBatchOptions};
 use async_trait::async_trait;
 use datafusion::{
     error::{DataFusionError, Result},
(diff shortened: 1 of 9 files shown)


native/core/src/execution/operators/parquet_writer.rs
```diff
@@ -533,13 +542,17 @@ impl ExecutionPlan for ParquetWriterExec {
...
-                    RecordBatch::try_new(Arc::clone(&schema_for_write), batch.columns().to_vec())
-                        .map_err(|e| {
-                            DataFusionError::Execution(format!(
-                                "Failed to rename batch columns: {}",
-                                e
-                            ))
-                        })?
+                    // Collection field IDs exist on the target schema, not on arrays produced by
+                    // the placeholder Scan. Ignore nested field names/metadata during validation
+                    // while still checking data types and nested/top-level nullability.
```
parthchandra: Is the nested nullability guaranteed to match with the target schema here?

spark/src/test/scala/org/apache/comet/parquet/CometParquetWriterSuite.scala
```diff
@@ -143,6 +145,168 @@ class CometParquetWriterSuite extends CometTestBase {
...
+            }
+          }
+        }
+
+        checkAnswer(spark.read.parquet(outputPath), data)
+      }
+    }
+  }
+
+  test("Spark reads native parquet output by field ID after columns are renamed and reordered") {
```
parthchandra: Maybe also add coverage for nested types.  Maybe a test that renames a struct field and a list element or map key/value?

native/core/src/execution/operators/parquet_writer.rs
sunchao: Yes. The placeholder scan and target schema both use the same Catalyst data types, and the scan normalizes incoming arrays before they reach the writer. `with_match_field_names(false)` ignores nested field names and metadata, but Arrow still recursively validates nullability for list elements, struct fields, and map entries. I clarified the comment and added `test_parquet_writer_rejects_mismatched_nested_nullability`, which verifies the writer rejects a nullable list element against a non-nullable target even when the actual data contains no nulls.

spark/src/test/scala/org/apache/comet/parquet/CometParquetWriterSuite.scala
sunchao: Added a field-ID read/write round trip covering a renamed nested struct field, a renamed field inside an `array<struct>`, and a renamed map alongside reordered top-level columns. The array element and map key/value IDs are preserved as well. Spark treats `element`, `key`, and `value` as structural names, so those fields cannot be renamed independently; their IDs are verified directly in the existing Parquet-footer regression.

parthchandra (approved):
lgtm

Issue #5363, opened 2026-08-15
[EPIC] Reimplement Scala expression microbenchmarks so they measure expression cost, not scan and result transfer


andygrove:
## Background

We now have ~25 Scala end-to-end microbenchmark suites for expressions, almost all of which funnel through `runExpressionBenchmark` in `spark/src/test/scala/org/apache/spark/sql/benchmark/CometBenchmarkBase.scala`. I reviewed them to check whether they actually isolate expression evaluation cost. They largely do not — a large and non-uniform share of each timed region is Parquet scan plus columnar-to-row conversion of the result set, and there are several fairness bugs that make individual rows misleading.

These benchmarks are what we use to justify optimization work and to catch regressions, so it is worth reimplementing the harness rather than patching individual suites.

## Problem 1: scan and result transfer are inside the timed region

Every timed case is:

```scala
spark.sql("SELECT expr(c1) FROM parquetV1Table").noop()
```

Two costs that are not expression evaluation are inside the timer:

**Parquet scan.** Read on every iteration, and critically the two arms use *different* readers — Spark's vectorized reader vs. Comet's native scan. So the contamination does not cancel. The `relative` column is a ratio of sums, `(scan_spark + expr_spark + c2r_spark) / (scan_comet + expr_comet + c2r_comet)`, which for any expression cheaper than the scan converges to the *scan* ratio rather than the expression ratio.

**Result transfer + columnar-to-row.** The `noop` sink writer is a `DataWriter[InternalRow]`, so a `ColumnarToRowExec` sits on top of every plan. This is known and tolerated — `CometPlanChecker.findFirstNonCometOperator` explicitly whitelists `ColumnarToRowExec`, `CometColumnarToRowExec`, `CometNativeColumnarToRowExec` and `WholeStageCodegenExec`. Since `spark.comet.exec.columnarToRow.native.enabled` defaults to `false`, the Comet arm pays Arrow FFI export to JVM `CometVector` to `UnsafeRow`, while Spark emits rows directly from a single fused codegen stage.

The second cost tracks the *output* type of the expression, so contamination varies a lot within a single suite:

| Output type | Affected suites | c2r cost |
| --- | --- | --- |
| int / bool / long | comparison, hash, `length`/`ascii`/`instr`, `like`/`rlike`, `unix_timestamp`, cast-to-numeric | small |
[...]

Issue #1367, opened 2025-02-03
Remote HDFS tests with Minikube


comphead:

comphead:
Closed by #1515

PR #5371, opened 2026-08-15
test: make expression benchmark harness fair and reproducible

andygrove:

Part of #5363. This covers checklist items 3, 4, 8 and 9 of that epic; the epic stays open.

## Rationale for this change

`runExpressionBenchmark` in `CometBenchmarkBase` backs 27 benchmark suites. Four defects make its output untrustworthy, and they are worth fixing before the larger harness rework in #5363 so that later results can be compared against something sound.

**Constant folding was excluded for the Comet arm only.** `cometExecConfigs` set `spark.sql.optimizer.excludedRules` to `ConstantFolding`; the Spark case ran with folding enabled. For `select space(2) from parquetV1Table` in `CometStringExpressionBenchmark`, the Spark plan was:

```
*(1) Project [   AS space(2)#1366]
```

a folded literal doing no per-row work, while Comet evaluated `space(2)` for every row. That row reported a Comet regression that does not exist. The conf was also assigned rather than appended, clobbering any pre-existing exclusions.

**Nothing checked that the benchmarked expression survived optimization.** Rules that are not excluded, such as `SimplifyCasts` on a no-op cast, can remove the expression entirely and leave a case that measures only the scan.

**The not-fully-native warning went to `println`.** `Benchmark(output = output)` writes the results table to the `.txt`, but the warning only reached the console, so a reader of a results table cannot tell that a row labelled "Comet" actually ran on Spark.

**Table data came from an unseeded `scala.util.Random`,** so runs are not comparable. This is not only cosmetic: `CometStringExpressionBenchmark` derives `REPEAT(CAST(value AS STRING), 10)` from it, so `lpad(c1, 150, 'x')` is sometimes padding and sometimes truncating.

All changes are in `CometBenchmarkBase.scala`. No call sites change.

- `spark.sql.optimizer.excludedRules` is now built once and applied to both arms, appending to the existing value rather than overwriting it.
- A structural check warns when the Spark baseline plan does no work beyond scanning and projecting bare attributes or literals. Any other node counts as work, so the join and aggregate suites, which legitimately project bare attributes, are unaffected.
- Both that warning and the existing not-fully-native warning go through a helper that writes to `output` as well as the console, so they land in the results file directly above the affected table.
- The base table is generated from a SplitMix64 mix of the row id instead of `Random.nextLong()`. Seeding a driver-side `Random` would not have worked, since the closure runs per row on the executor. The helper lives in a companion object because referencing a trait method from a Dataset closure captures `this`, which is not serializable.

Two things reviewers may want to weigh in on:

- The checks warn rather than fail. Failing would block regenerating results for suites that currently fall back, which later items in #5363 address.
- The plan check runs the Spark arm once, untimed, before benchmarking. Previously only the Comet arm was pre-run. This costs an extra run per case but warms both arms symmetrically.

Results files are not regenerated here. The final item of #5363 regenerates them once, after the baseline-case and aggregate-sink work lands.

Benchmarks do not run in CI, so this was verified by running them locally. Compiles clean under the `spark-4.1`, `spark-3.5` and `spark-3.4` profiles.

- **Reproducibility.** `SELECT sum(hash(value)), count(*)` over the generated table returned `sum=-293787312920 count=65536` on three independent runs across JVM restarts.
- **The `space(2)` fix.** With folding excluded on both arms the Spark plan becomes `*(1) Project [space(2) AS space(2)#1367]`, evaluated per row, rather than the folded literal above.
- **Both warning paths.** Exercised with a temporary suite containing a bare-column projection and a case with `spark.comet.exec.project.enabled=false`. Both warnings appeared in the generated `.txt` above the correct table, with the offending plan. A control case using `abs(c1)` produced no warning. The temporary suite was removed.
- **No false positives.** `CometStringExpressionBenchmark` (31 expressions), `CometCastNumericToNumericBenchmark` and `CometPredicateExpressionBenchmark` were run in full. None produced a trivial-plan warning.

Running the suites with the warnings visible surfaced fallbacks that were previously console-only:

- `translate` in `CometStringExpressionBenchmark` falls back to a JVM `Project`.
- Eight `c_short` cases in `CometCastNumericToNumericBenchmark` do the same, for example `Project [cast(c_short#19 as int) AS c_short#528]`.

Those rows have been reporting Spark timings under a "Comet" label. I will file a separate issue rather than address them here.

Two corrections to the epic text that came out of this work, noted for whoever picks up the remaining items:

- #5363 refers throughout to "committed results files". `spark/benchmarks` is listed in `.gitignore`, so no Scala microbenchmark results are tracked in git; the only committed benchmark results are the TPC JSON files under `benchmarks/results/`. The `.txt` files are local artifacts pasted into PRs and issues by hand, which is where a console-only warning does its damage.
- The epic expects the plan assertion to catch the `In` predicate in `CometPredicateExpressionBenchmark`. It does not, and cannot: Spark retains `FilterExec` above the Parquet scan even when the filter is pushed down, so that plan contains real work. Confirmed by running the suite. Moving the `In` into the SELECT list remains a separate manual fix.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/benchmark/CometBenchmarkBase.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/benchmark/CometBenchmarkBase.scala
@@ -29,7 +29,7 @@ import org.apache.parquet.crypto.keytools.mocks.InMemoryKMS
 import org.apache.spark.SparkConf
 import org.apache.spark.benchmark.Benchmark
 import org.apache.spark.sql.{DataFrame, DataFrameWriter, Row, SparkSession}
-import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, Literal}
+import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, Expression, Literal}
 import org.apache.spark.sql.catalyst.optimizer.ConstantFolding
 import org.apache.spark.sql.comet.CometPlanChecker
 import org.apache.spark.sql.comet.util.Utils
@@ -157,6 +157,16 @@ trait CometBenchmarkBase
             |likely removed by the optimizer.
             |Spark plan:""".stripMargin + "\n" + plan.treeString)
       }
+      val rowInvariant = rowInvariantProjections(plan)
+      if (rowInvariant.nonEmpty) {
+        warn(
+          benchmark,
+          s"""WARNING: these projections reference no input column, so their value is the same
+             |for every row: ${rowInvariant.mkString(", ")}
+             |An engine may evaluate them once per batch and broadcast the result rather than
+             |per row, so the two arms are not necessarily doing comparable work. Benchmark the
+             |expression over a column instead.""".stripMargin)
+      }
     }
 
     // Check that the plan is fully Comet native before running the benchmark. Unlike the check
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/benchmark/CometBenchmarkBase.scala
```diff
@@ -125,50 +131,103 @@ trait CometBenchmarkBase
...
 
-    benchmark.addCase("Spark") { _ =>
-      withSQLConf(CometConf.COMET_ENABLED.key -> "false") {
-        spark.sql(query).noop()
-      }
-    }
+    // Constant folding is excluded so that expressions over literal arguments are still evaluated
+    // per row. It must be excluded for both arms: if only Comet excludes it, Spark folds the
+    // expression away and does no per-row work, and the comparison is meaningless.
+    val noConstantFolding =
```
sunchao, line 137: [P2] Could we make this benchmark depend on an input column rather than just disable `ConstantFolding`? The existing `SELECT space(2) FROM parquetV1Table` still compares different amounts of work: Spark's `StringSpace` calls `UTF8String.blankString(2)` once per row, while Comet's DataFusion `space` UDF receives `ColumnarValue::Scalar`, evaluates it once per batch, and broadcasts the result. The structural plan check also accepts the retained `ProjectExec`, so it does not flag the mismatch. Materializing an integer column and benchmarking `space(column)` would make both engines evaluate the expression across the rows.

sunchao (approved):
Thanks for improving the benchmark harness. The deterministic inputs and persisted fallback warnings are helpful. I left one [P2] comment about making the existing space(2) case measure comparable work.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometBenchmarkBase.scala
andygrove: Fixed. Thanks for suggesting that.

Issue #1636, opened 2025-04-10
Benchmark automation


andygrove:
### What is the problem the feature request solves?

I have been spending significant time manually running benchmarks, both during development and when preparing to release Comet. I have been using a mix of open-source scripts and custom scripts. I want other contributors to be able to run the same benchmarks, both locally and in the cloud.

I recently documented [how to run benchmarks in AWS](https://datafusion.apache.org/comet/contributor-guide/benchmarking_aws_ec2.html) towards this goal.

My goals for this issue are:

1. Ensure that all scripts, configs, and documentation are located in the datafusion-comet repo so that anyone can run the same benchmarks by just cloning the repo and running a script
2. Provide recommended configs for different environments and scale factors
3. Have some subset of benchmarks run nightly against the main branch, with the results posted somewhere for later analysis and reporting (this could be as simple as pushing to a GitHub repo)

PR #5374, opened 2026-08-15
feat: add micro benchmark runner and EC2 guide

andygrove:

Part of #1636.

## Rationale for this change

Comet has around 35 micro benchmark suites, but running them is a manual, one-suite-at-a-time
exercise (`SPARK_GENERATE_BENCHMARK_FILES=1 make benchmark-<class>`), and the results are written
to `spark/benchmarks`, which is git ignored. That makes it tedious to produce a full set of numbers
and impossible to publish them without force adding files.

Running the suites on a laptop also produces noisy numbers because of thermal throttling,
efficiency cores, and background load. A dedicated EC2 instance gives repeatable results, but the
existing EC2 guide only covers TPC-H and TPC-DS.

- `benchmarks/micro/run.py`, a dependency-free Python entry point with `setup`, `run`, `collect`,
  `publish` and `all` subcommands. It installs the prerequisites, clones or updates the repository
  at a chosen ref, builds in release mode, runs each suite in its own JVM with per-suite logs,
  timings and timeouts, continues past failures, and copies the results into
  `benchmarks/results/micro` together with a `RUN-INFO.md` recording the instance type, CPU,
  memory, JDK, Rust and Comet commit that produced them. The script can be downloaded on its own
  before the repository is cloned.
- The suites are discovered from the sources in
  `spark/src/test/scala/org/apache/spark/sql/benchmark`, so a newly added benchmark runs with no
  change to the script. The three TPC suites, `CometC2RIsolatedBench` and `CometReadHdfsBenchmark`
  are excluded from the default set, with the reason reported by `run --list`, leaving the 35
  suites that generate their own data.
- `print-benchmark-args` in the `Makefile`, which exposes the `MAVEN_OPTS` and `mvnw` arguments
  that `benchmark-%` uses so that `run.py` reads the invocation rather than restating it.
  `BENCH_HEAP` defaults to the previous `20g`, so that `benchmark-%` can also be run by hand on an
  instance with less memory.
- `benchmarks/results/micro` as the tracked location for published results. `spark/benchmarks`
  stays ignored and is used as scratch output.
- A new contributor guide, `benchmarking_micro_ec2.md`, covering instance creation on
  `m7i.xlarge`, manual installation of every prerequisite (git, make, gcc, cmake,
  protobuf-compiler, JDK 17, Rust, and a note that Maven is provided by `./mvnw`), running,
  collecting, publishing, guidance on getting comparable numbers, and troubleshooting.
- `dev/ci/check-benchmark-runner.py`, run from the `preflight` lint job, which imports the runner,
  checks its suite names against the benchmark sources, and exercises the pure helpers and the log
  parsing.
- A fix for the `jdk17` profile activation in `pom.xml`. `<jdk>17</jdk>` is a prefix match, not a
  floor, so on JDK 21 or later `java.version` stayed at its default of 11 and the Spark 4.x sources
  were compiled against the Java 11 API, failing with "Class java.lang.Record not found". The
  activation is now the range `[17,)`.
- A RAT exclusion for the result text files, and a prettier exemption for the generated
  `RUN-INFO.md`, whose tables do not match prettier's column alignment and would otherwise
  fail the markdown lint job on every results PR.

The runner was exercised locally against a release build: suite listing and `--only` / `--skip`
filtering, a dry run of the Maven invocation, a real run of `CometArithmeticBenchmark` that
completed in 4m35s and produced a results file which `collect` then picked up, and the publish step
including the generated `RUN-INFO.md` and the git and `gh` commands. Suite discovery was checked to
select exactly the 35 suites the previous hand-maintained list named.
`dev/ci/check-benchmark-runner.py` passes. The `jdk17` activation change was checked by evaluating
`maven.compiler.source` on JDK 11, 17, 21 and 22 with no profile, `-Pspark-3.5` and `-Pspark-4.1`:
11 and 17 are unchanged, 21 and 22 now resolve to 17 rather than 11. Documentation changes were
checked with prettier.


dwsmith1983:
@andygrove I see you have a bunch of benchmarks PRs. I ran Comet with Iceberg and benched marked at 100GB and 1TB. I have the details if you want them.
https://dustinsmith.info/blog/aws-parquet-column-index/

benchmarks/micro/run.py
```diff
@@ -0,0 +1,1066 @@
...
+        cwd=comet_home,
+        env=env,
+    )
+    if not value:
+        log("could not evaluate extraJavaTestArgs, continuing without it")
+        return ""
+    return " ".join(value.split())
+
+
+def suite_command(suite, args, jdk_major=None):
```
mbutrovich: `suite_command()`'s docstring says it mirrors the `benchmark-%` Makefile target (`Makefile:31-33`), duplicating its exec flags and `MAVEN_OPTS` composition, including a second `./mvnw help:evaluate -Dexpression=extraJavaTestArgs` call. Understandable, since that target's `release` dependency would rebuild before every suite, but it's now two sources of truth for the same invocation that can drift. Could the Makefile expose the argument list (e.g. `print-benchmark-args`) for `run.py` to read instead of retyping it?

pom.xml
mbutrovich, line 793: Root cause behind several of this PR's commits. `<jdk>17</jdk>` is an exact match, not a floor, so JDK 21 needs `-Pjdk17` passed explicitly. Maven's `<jdk>` activation supports ranges. Would `<jdk>[17,)</jdk>` fix this at the source for every Maven invocation, making `profile_args()`'s detection unnecessary? Possibly separate-PR scope, but worth a tracking issue if so.

benchmarks/micro/run.py
```diff
@@ -0,0 +1,1066 @@
...
+}
+
+# Suites deliberately left out of the default set, with the reason. They can
+# still be run by naming them in a --suites file.
+EXCLUDED_SUITES = {
+    "CometTPCHQueryBenchmark": "needs TPC-H data, pass --data-location",
+    "CometTPCDSQueryBenchmark": "needs TPC-DS data, pass --data-location",
+    "CometTPCDSMicroBenchmark": "needs TPC-DS data, pass --data-location",
+    "CometC2RIsolatedBench": "prints to stdout only, writes no results file",
+}
```
mbutrovich: Three unlinked hand-maintained string-keyed lists, nothing ties them to the actual classes in `spark/src/test/scala/org/apache/spark/sql/benchmark`. A suite could land in both `DEFAULT_SUITES` and `EXCLUDED_SUITES` with no error, and a new benchmark class silently never runs until someone remembers to add it. Worth discovering suites from source and diffing against these lists, even just as a `--list --check` warning?

benchmarks/micro/run.py
```diff
@@ -0,0 +1,1066 @@
...
+    if java_home:
+        env["JAVA_HOME"] = java_home
+        java_bin = Path(java_home) / "bin"
+        if str(java_bin) not in env.get("PATH", ""):
+            env["PATH"] = f"{java_bin}{os.pathsep}{env.get('PATH', '')}"
+    env["COMET_CONF_DIR"] = str(comet_home / "conf")
+    return env
+
+
+def profile_args(args, jdk_major=None):
```
mbutrovich: `profile_args`, `maven_extra_java_args`, `suite_command` take `jdk_major=None`, but every call site computes it via `require_supported_jdk()` first and always passes an int, so the `None` branch is dead. They also take the whole `args` namespace just to read `args.profile`. Make `jdk_major` required and pass `args.profile` directly instead of `args`?

benchmarks/micro/run.py
```diff
@@ -0,0 +1,1066 @@
...
+    except OSError:
+        pass
+    try:
+        for line in Path("/proc/meminfo").read_text().splitlines():
+            if line.startswith("MemTotal"):
+                kilobytes = int(line.split()[1])
+                info["memory"] = f"{kilobytes / 1024 / 1024:.1f} GiB"
+                break
+    except OSError:
+        pass
```
mbutrovich: The `/proc/cpuinfo` and `/proc/meminfo` reads are the same shape (find prefixed line, split, break) written twice. Worth a `read_proc_field(path, prefix)` helper called twice instead?

benchmarks/micro/run.py
```diff
@@ -0,0 +1,1066 @@
...
+# ---------------------------------------------------------------------------
+
+
+def cmd_all(args):
+    comet_home = cmd_setup(args)
+    args.comet_home = str(comet_home)
+    run_dir = cmd_run(args)
+    if run_dir is None:
+        return
+    args.run_dir = str(run_dir)
```
mbutrovich: Relays `comet_home`/`run_dir` to `cmd_run`/`cmd_collect` by writing them back onto the shared `args` namespace rather than passing them explicitly. Would passing them directly (or as optional params) be easier to follow than the implicit side channel?

benchmarks/micro/run.py
mbutrovich, line 1: No CI job touches this script, not even a syntax check (`PR Benchmark Check` only covers `native/benches` and the Scala benchmark sources, correctly skipping here). Worth adding `python3 -m py_compile benchmarks/micro/run.py`, or a small `pytest` over the pure helpers (`format_duration`, `protoc_asset_name`, `profile_args`, `load_suites` filtering), so a future edit that breaks `setup`/`collect`/log parsing gets caught before someone runs it on an EC2 box?

mbutrovich (commented):
First pass, mostly minor stuff. Thanks @andygrove!

benchmarks/micro/run.py
andygrove: Done. The Makefile now defines `BENCH_MAVEN_OPTS` and `BENCH_MVN_ARGS`, uses them in `benchmark-%`, and exposes them through a `print-benchmark-args` target. `run.py` shells out to `make print-benchmark-args BENCH_HEAP=<heap> PROFILES=<profile>` and parses the KEY=VALUE lines, so the exec flags, the heap and the `extraJavaTestArgs` evaluation all have one definition. The target itself still cannot be used, for the reason you spotted, but nothing is retyped now.

I also added `$(PROFILES)` to the `spark_jvm_17_extra_args` evaluation while I was there, since it was evaluating the property without the profile the rest of the invocation uses.

pom.xml
andygrove: Good catch, and yes. Fixed in this PR rather than deferred, since it removes the workaround rather than documenting it: the profile now activates on `<jdk>[17,)</jdk>`. I checked `maven.compiler.source` across JDK 11, 17, 21 and 22 with no profile, `-Pspark-3.5` and `-Pspark-4.1`; 11 and 17 are unchanged and 21 and 22 now resolve to 17 instead of 11.

`profile_args()` is gone, along with the `-Pjdk17` handling in the runner and the corresponding troubleshooting section in the guide. The minimum-JDK check stays, since a JDK older than 17 still fails.

benchmarks/micro/run.py
andygrove: Agreed, and I went further than the warning: `DEFAULT_SUITES` is gone. The suites are now discovered from `spark/src/test/scala/org/apache/spark/sql/benchmark` (a top-level `object` that mixes in a `*Benchmark*` base or declares its own `main`), and the default set is everything discovered minus `EXCLUDED_SUITES`. A new benchmark class runs with no change to `run.py`, and a suite cannot be in both lists.

Discovery immediately found one the hand-maintained list had missed: `CometReadHdfsBenchmark`. It is now excluded with a reason, since it starts a local HDFS mini cluster. The 35 suites the old list named are exactly what discovery selects.

`EXCLUDED_SUITES` and `SUITE_TIMEOUT_OVERRIDES` are still keyed by name, so a rename would silently un-exclude a suite or drop a timeout override. `stale_suite_entries()` diffs them against the sources and warns on every `run`, and the new CI check fails on it.

benchmarks/micro/run.py
andygrove: Both fixed. `profile_args()` and `maven_extra_java_args()` are gone entirely (the pom fix and the Makefile target replaced them), and `suite_command(suite, mvn_args, profile)` now takes what it uses. `require_supported_jdk()` no longer returns the version, since nothing needed it once the `-Pjdk17` handling went away.

benchmarks/micro/run.py
andygrove: Done, added `read_proc_field(path, prefix)` and called it for both.

benchmarks/micro/run.py
andygrove: Yes, clearer. `cmd_run`, `cmd_collect` and `cmd_publish` now take `comet_home` and `run_dir` as optional parameters that default to resolving from `args`, and `cmd_all` passes them explicitly. Nothing is written back onto the namespace.

benchmarks/micro/run.py
andygrove: Added `dev/ci/check-benchmark-runner.py`, run from the `preflight` lint job next to `check-suites.py`. It imports the runner (which is the syntax check), asserts that discovery finds a plausible set and that `EXCLUDED_SUITES` and `SUITE_TIMEOUT_OVERRIDES` still name real benchmarks, checks the `--only`/`--skip` filtering, and feeds a synthetic failing log through `failure_excerpt` to cover the exception-first and Maven-fallback paths. Plain asserts and stdlib only, matching `check-suites.py`, since the lint job has no pytest.

andygrove:
Thanks @mbutrovich, all seven addressed in the two commits above.

The two that changed more than the runner: the `jdk17` profile now activates on `<jdk>[17,)</jdk>` in `pom.xml`, which fixes the JDK 21 problem for every Maven invocation and let me delete the `-Pjdk17` workaround, and the `Makefile` now exposes the benchmark invocation through `print-benchmark-args` so `run.py` reads it instead of restating it. `DEFAULT_SUITES` is gone in favour of discovering the suites from the benchmark sources, which turned up `CometReadHdfsBenchmark` sitting in neither list. Details in the individual replies.

mbutrovich (approved):
Thanks for the quick revision, @andygrove!

Issue #5269, opened 2026-08-05
arrays_overlap panics on nested double arrays built from expressions


andygrove:
## Describe the bug

`arrays_overlap` over `array<array<double>>` panics in native code when both sides of the comparison are expression-constructed nested lists rather than plain columns.

```
org.apache.comet.CometNativeException: native panic: index out of bounds: the len is 2 but the index is 2
Comet native panic: panicked at arrow-array-58.4.0/src/array/list_array.rs:352:19
```

`list_array.rs:352` is `GenericListArray::value`, so something in `arrays_overlap_list_generic` (`native/spark-expr/src/array_funcs/arrays_overlap.rs`) indexes past the end of one of its two list arrays.

## Steps to reproduce

As a Comet SQL file test:

```sql
statement
CREATE TABLE t_panic(a array<array<double>>) USING parquet

statement
INSERT INTO t_panic VALUES
  (array(array(-0.0))),
  (array(array(0.0))),
  (array(array(1.0, -0.0))),
  (array(array(double('NaN')))),
  (array(array(1.0))),
  (array(array(-0.0), cast(NULL as array<double>))),
  (array(cast(NULL as array<double>)))

query
SELECT arrays_overlap(array(array(a[0][0])), array(array(cast(0.0 as double)))) FROM t_panic
```

Reproduced on `apache/main` at `f262b13d2`, Spark 4.1 profile, JDK 17, macOS.

## What I could and could not narrow down

The panic needs all seven rows. None of these reproduce it on their own:

- any single row from the table, including the two rows carrying a `NULL` inner list
- a two-row table combining a multi-element inner list with a `NULL` inner list
- the `int` equivalent of the same query
- `arrays_overlap(a, array(array(cast(0.0 as double))))`, a plain column against a constant
- `arrays_overlap(array(a[0]), array(array(0)))`, only the outer list constructed

The failing task index moves with the number of rows, so this looks like a batch or partition boundary interacting with the per-row `left.value(i)` / `right.value(i)` slicing in `arrays_overlap_list_generic`, rather than anything about a specific row's contents. Worth checking whether the two sides can end up with different `len()` when one of them arrives as a broadcast scalar and the other is a per-row constructed array.

## Additional context

Found while reviewing #5235, which changes the same function. The panic reproduces identically with and without that PR applied, so it is independent of it.

peterxcli:
take

PR #5376, opened 2026-08-15
fix: canonicalize NaN in flat arrays_overlap float keys

### LINKED ISSUE #5270: arrays_overlap on flat float arrays does not canonicalize NaN like Spark
## Describe the bug

`arrays_overlap` over a flat `array<double>` or `array<float>` column disagrees with Spark when either side holds a NaN whose bit pattern is not the canonical positive NaN.

Comet's flat path keys its hash set on the raw bits:

```rust
// native/spark-expr/src/array_funcs/arrays_overlap.rs
impl OverlapKey for f64 {
    type Key = u64;
    fn overlap_key(self) -> u64 {
        self.to_bits()
    }
}
```

Spark's `ArraysOverlap` uses `fastEval` for atomic element types, which is a `java.util.HashSet` over boxed values. `java.lang.Double.equals` compares `doubleToLongBits`, and that collapses every NaN payload and sign onto one pattern. So Spark treats `-NaN` and `NaN` as the same element and Comet does not.

## Steps to reproduce

The negation has to happen at query time. A `-double('NaN')` literal is canonicalized before it reaches Parquet, so putting one in the data does not reproduce it.

```sql
statement
CREATE TABLE t_flat(x double) USING parquet

statement
INSERT INTO t_flat VALUES (double('NaN')), (0.0), (1.0)

query
SELECT arrays_overlap(array(-x), array(double('NaN'))) FROM t_flat
```

```
!== Spark Answer - 3 ==     == Comet Answer - 3 ==
 [false]                    [false]
 [false]                    [false]
![true]                     [false]
```

Reproduced on `apache/main` at `f262b13d2`, Spark 4.1 profile, JDK 17, macOS.

## What the fix looks like

`overlap_key` needs to canonicalize NaN before taking the bits. It must **not** use `normalize_float` from `math_funcs/internal/normalize_nan.rs`, because that also folds `-0.0` into `0.0`, and the flat path is required to keep those distinct.

That last part is easy to get wrong, so spelling it out. Spark's flat and nested paths genuinely differ on signed zero:

| | signed zero | NaN sign / payload |
|---|---|---|
| flat (`fastEval`, `HashSet` of boxed `Double`) | `-0.0` and `0.0` are **distinct** | all NaN are **equal** |
| nested (`bruteForceEval`, `ordering.equiv` -> `SQLOrderingUtil.compareDoubles`) | `-0.0` and `0.0` are **equal** | all NaN are **equal** |

Verified directly against the JDK:

```
HashSet{0.0}.contains(-0.0) = false
HashSet{NaN}.contains(-NaN runtime) = true
compareDoubles(-0.0, 0.0) = 0
compareDoubles(-NaN, NaN) = 0
```

Comet's flat path already matches Spark on signed zero. Only the NaN column is wrong. The nested path is being fixed for both in #5235.

## Additional context

`array_position` does not have this problem. `position_float` uses `(search_is_nan && v.is_nan()) || v == search_val`, which matches `compareDoubles` on both signed zero and NaN.

Found while reviewing #5235.

sunchao:

`arrays_overlap` answers a simple question: do two arrays contain at least one common value? For floating-point arrays, Spark treats every representation of `NaN` as the same value, even when their underlying IEEE 754 sign bits or payloads differ. Comet's native implementation compared those raw bit patterns instead. As a result, enabling Comet could silently change the answer to an otherwise ordinary Spark SQL query.

Consider a Parquet table containing a `NaN`:

```sql
CREATE TABLE t_flat(x DOUBLE) USING parquet;
INSERT INTO t_flat VALUES (DOUBLE('NaN')), (0.0D), (1.0D);

SELECT x,
       arrays_overlap(array(x), array(-x)) AS overlap
FROM t_flat;
```

For the first row, `-x` is still a `NaN`, but negating it at execution time changes its sign bit. Spark correctly considers that value equal to the original `NaN`. On Spark 3.4–4.1, before this change, Comet incorrectly treated the two representations as different:

| Input `x` | Spark 3.4–4.1 | Comet before | Comet after |
| --- | --- | --- | --- |
| `NaN` | `true` | `false` | `true` |
| `+0.0` | `false` | `false` | `false` |
| `1.0` | `false` | `false` | `false` |

The `+0.0` row is also a compatibility control: `-x` becomes `-0.0`, and Spark 3.4–4.1 considers the two zeros different for flat arrays. Fixing `NaN` comparisons must not change that result.

Spark 4.2 and later intentionally behave differently. Under [SPARK-54918](https://github.com/apache/spark/blob/0e25e40348772702b621a8664482510c17e99812/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/NormalizeFloatingNumbers.scala#L102-L113), Spark's optimizer normalizes both floating-point `arrays_overlap` operands before evaluation. Consequently, `+0.0` and `-0.0` overlap, and both Spark and Comet return `true`. The native comparison must still preserve distinct zero keys for earlier Spark versions; on Spark 4.2+, the operands have already been normalized before those keys are constructed.

On Spark 3.4–4.1, the mismatch also affects SQL null semantics. If the first array additionally contains `NULL`, Spark returns `true` because the two `NaN` values overlap, while the previous Comet implementation returns `NULL` because it misses that definite match:

```sql
SELECT arrays_overlap(
  array(x, CAST(NULL AS DOUBLE)),
  array(-x)
)
FROM t_flat
WHERE isnan(x);

-- Spark: true
-- Comet before: NULL
-- Comet after: true
```

Runtime negation in these examples is intentional. Parquet canonicalizes floating-point `NaN` values when writing them, so storing a negative or custom-payload `NaN` in the input file is not enough to reproduce the bug. A distinct representation must be produced after the scan.

The signed-zero boundary is therefore version-specific:

```sql
SELECT arrays_overlap(array(0.0D), array(-0.0D));

-- Spark 3.4–4.1 and Comet: false
-- Spark 4.2+ and Comet: true (operands are normalized first)
```

A general-purpose floating-point normalizer that also merges signed zero inside Comet's comparison key would fix the NaN mismatch but introduce a Spark 3.4–4.1 compatibility regression.

Closes #5270.

The change gives Comet's existing flat-array comparison the same floating-point equality contract as Spark. When Comet derives a comparison key for a `FLOAT` or `DOUBLE`, every `NaN` representation is mapped to one canonical key. All other values retain their original bit patterns. This preserves distinct signed zeros when Spark 3.4–4.1 passes them through unchanged, while respecting Spark 4.2's earlier normalization of the operands.

Applying the correction at the shared comparison-key boundary fixes both ways the existing implementation checks for overlap: direct comparisons for small arrays and hash-based lookups for larger arrays. The execution plan, fast-path structure, null propagation, and behavior for non-floating-point values are otherwise unchanged.

This PR intentionally does not change nested-array comparisons. Spark uses a different equality contract for nested values, and the separate nested signed-zero issue is addressed in #5235.

Mixed scalar/array argument broadcasting is another pre-existing issue, tracked separately in #5269. This change does not alter that argument-dispatch behavior.

## How was this PR tested?

The native Rust regressions construct positive, negative, and signaling `NaN` bit patterns directly for both floating-point widths. They verify NaN equivalence, distinct signed-zero keys, SQL null handling, and both sides of the small-array/hash-lookup threshold. All 23 focused `arrays_overlap` tests pass.

Most integration coverage now lives in the existing `expressions/array/arrays_overlap.sql` fixture, as requested in review. Its Parquet-backed queries generate noncanonical NaNs after the scan and exercise `FLOAT` and `DOUBLE`, reversed operands, nullable arrays, signed-zero controls, and arrays large enough to use hash lookup. The SQL-file harness compares Comet with the running Spark version and checks native execution, so it also covers Spark 4.2's different signed-zero behavior without duplicating version-specific expected results.

A small Scala test remains to inspect the actual floating-point bits. It confirms that Parquet canonicalizes stored NaNs and that native runtime negation produces noncanonical NaNs after the scan—something SQL equality alone cannot establish.

Both focused integration tests pass locally on Spark 3.5, Spark 4.0, and Spark 4.2. Maven Spotless checks also pass for all three profiles.

Native formatting, compilation, and focused tests, from `native/`:

```bash
cargo fmt --all -- --check
cargo build
cargo test -p datafusion-comet-spark-expr --lib \
  array_funcs::arrays_overlap::tests -- --nocapture
```

The focused Spark commands were run from the repository root with JDK 17 configured. Clean reactor builds were used when switching profiles:

```bash
./mvnw -B -ntp test -Pspark-3.5 -Dtest=none \
  '-Dsuites=org.apache.comet.CometSqlFileTestSuite expressions/array/arrays_overlap.sql' \
  -Dscalastyle.skip=true
./mvnw -B -ntp test -Pspark-3.5 -Dtest=none \
  '-Dsuites=org.apache.comet.CometArrayExpressionSuite runtime NaN representations' \
  -Dscalastyle.skip=true

./mvnw -B -ntp clean test -Pspark-4.0 -Dtest=none \
  '-Dsuites=org.apache.comet.CometSqlFileTestSuite expressions/array/arrays_overlap.sql' \
  -Dscalastyle.skip=true
./mvnw -B -ntp test -Pspark-4.0 -Dtest=none \
  '-Dsuites=org.apache.comet.CometArrayExpressionSuite runtime NaN representations' \
  -Dscalastyle.skip=true

./mvnw -B -ntp clean test -Pspark-4.2 -Dtest=none \
  '-Dsuites=org.apache.comet.CometSqlFileTestSuite expressions/array/arrays_overlap.sql' \
  -Dscalastyle.skip=true
./mvnw -B -ntp test -Pspark-4.2 -Dtest=none \
  '-Dsuites=org.apache.comet.CometArrayExpressionSuite runtime NaN representations' \
  -Dscalastyle.skip=true
```

### DIFF
--- a/native/spark-expr/src/array_funcs/arrays_overlap.rs
+++ b/native/spark-expr/src/array_funcs/arrays_overlap.rs
@@ -272,10 +272,9 @@ fn range_has_null(nulls: Option<&NullBuffer>, range: Range<usize>) -> bool {
     nulls.is_some_and(|n| n.null_count() > 0 && n.slice(range.start, range.len()).null_count() > 0)
 }
 
-/// Projects a native value onto a hashable key whose equality matches the Arrow compare kernels:
-/// the value itself for integral types, the bit pattern for floats. Arrow orders floats by total
-/// order rather than IEEE semantics (NaN equals NaN, and 0.0 does not equal -0.0), which is
-/// exactly bit equality.
+/// Projects a native value onto a Spark-compatible hashable key. Floating-point keys canonicalize
+/// every NaN representation while preserving the distinct bit patterns of positive and negative
+/// zero.
 trait OverlapKey: Copy {
     type Key: Hash + Eq + Copy;
 
@@ -299,7 +298,11 @@ impl OverlapKey for f32 {
     type Key = u32;
 
     fn overlap_key(self) -> u32 {
-        self.to_bits()
+        if self.is_nan() {
+            f32::NAN.to_bits()
+        } else {
+            self.to_bits()
+        }
     }
 }
 
@@ -307,7 +310,11 @@ impl OverlapKey for f64 {
     type Key = u64;
 
     fn overlap_key(self) -> u64 {
-        self.to_bits()
+        if self.is_nan() {
+            f64::NAN.to_bits()
+        } else {
+            self.to_bits()
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


comphead (commented):
Thanks @sunchao looks solid to me, would be great to have tests mostly in `sql` file 
```
src/test/resources/sql-tests/expressions/array/arrays_overlap.sql
```

sunchao:
Thanks @comphead! Moved the behavioral cases into the existing `arrays_overlap.sql` in 247e0b173. The SQL suite now compares against the running Spark version and covers runtime-generated NaNs, signed zeros, nulls, reversed operands, and both the short-array and hash-lookup paths for `FLOAT` and `DOUBLE`.

I kept a small Scala raw-bit test because SQL equality cannot distinguish NaN representations. It verifies that runtime negation produces noncanonical NaNs after Parquet has canonicalized the stored values.

Both focused integration tests pass locally on Spark 3.5, 4.0, and 4.2, and all 23 targeted Rust tests pass.

comphead (approved):
Thanks @sunchao

Issue #4295, opened 2026-05-12
[FEATURE] Native scan support for VariantType columns (Iceberg + Spark 4.0)


Shekharrajak:
### What is the problem the feature request solves?

Today, Comet falls back to JVM Spark for any query touching a VariantType column. This eliminates Comet's acceleration for the entire query, including unrelated operators that could otherwise run natively. This issue tracks adding native execution for Variant column scans, both for plain Parquet and Iceberg tables, by consuming new Variant primitives from arrow-rs and iceberg-rust.

### Describe the potential solution

```
spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala (line 729):

case s: StructType if isVariantStruct(s) =>
  fallbackReasons +=
    s"Unsupported $name of type VariantType (shredded; not supported by $scanImpl scan)"
  false
```

### Additional context

Follow-ups

•  Comet support for parse_json / to_variant (write side)
•  Comet support for schema_of_variant (metadata operation)
•  Predicate pushdown on variant_get(...) = literal into Parquet's column index
•  Variant in shuffle / spill paths (separate issue)

Shekharrajak:
Ref https://github.com/apache/arrow-rs/pull/9963 
https://github.com/apache/arrow-rs/issues/8480
https://github.com/apache/iceberg-rust/pull/2423
https://github.com/apache/iceberg-rust/issues/2411

kazuyukitanimura:
Changed the priority to high because this is an important feature

peterxcli:
I would like to implement first step of this - basic native VARIANT projection for ordinary Parquet (protobuf/type identity, Arrow-to-Spark bridge, unshredding/reordering, and Spark VariantType vectors).

peterxcli:
take

PR #5377, opened 2026-08-15
feat: Support native scans with unprojected Spark 4 VARIANT columns

sunchao:

Spark 4 introduces `VARIANT` for semi-structured data, so ordinary analytical tables can now contain a mixture of familiar typed columns and a JSON-like `VARIANT` column. Adding that one column should not prevent Comet from accelerating queries that never read it.

For example, consider a table such as:

```sql
CREATE TABLE events (
  event_id BIGINT,
  payload VARIANT,
  event_type STRING
) USING parquet;

SELECT event_id, event_type
FROM events
WHERE event_type = 'purchase';
```

The query only needs `event_id` and `event_type`; it does not decode, filter, or otherwise inspect `payload`. Nevertheless, the existing scan paths make decisions using the complete table schema instead of the columns the native reader actually needs.

For an Iceberg table, that means the presence of `payload` causes the entire native scan to fall back, even though the query projects only supported scalar columns. For a regular Parquet table, Spark can correctly prune `payload` from the requested projection, but Comet still serializes the original full relation schema when building its native plan. Because `VARIANT` has no native Spark-type serialization, the supposedly supported scan can fail during plan construction.

The practical result is that introducing a `VARIANT` column can either disable native acceleration for unrelated queries or make those queries fail outright. This PR addresses that narrower, immediately useful part of [#4295](https://github.com/apache/datafusion-comet/issues/4295): queries that do not need `VARIANT` data should continue to run natively, while queries that actually need to read `VARIANT` must still fall back safely to Spark.

The change makes native-scan eligibility follow the actual data projection, while respecting the different ways Spark Parquet scans and Iceberg scans describe their input schemas.

For a regular Parquet scan, Spark has already determined which columns and nested fields are required. Comet now derives the schema sent to native execution from that information instead of blindly serializing every field in the relation schema. A completely unrequested `VARIANT` column is omitted. If a requested struct contains both a supported field and an unrequested `VARIANT` sibling, Comet uses Spark's already-pruned version of that struct, so a query such as `SELECT details.label FROM nested_events` can still execute natively when `details` has type `STRUCT<label: STRING, payload: VARIANT>`. Because removing a field changes subsequent positions, the native projection is rebuilt against the pruned schema so that later data columns, partition values, and file metadata continue to refer to the correct fields.

Iceberg requires a different boundary. Its complete table schema is still passed to `iceberg-rust`, which can represent `VARIANT` in schema metadata but cannot materialize a projected `VARIANT` field or a projected parent that contains one. The scan rule therefore distinguishes between roots that the reader will actually project and roots that remain entirely untouched. An omitted root may contain `VARIANT`; a projected root is still checked strictly, as are every other unsupported type and any scan whose projected field IDs cannot be determined safely. This also preserves fallback for empty and metadata-only projections, because the current Iceberg reader interprets an empty physical projection as a request for all columns.

That distinction must use Iceberg field IDs rather than column names. Column names can change while their IDs remain stable, which matters for time travel:

```sql
-- Snapshot 123 was created while the column was named "details".
ALTER TABLE iceberg_events RENAME COLUMN details TO renamed_details;

SELECT details.label
FROM iceberg_events VERSION AS OF 123;
```

The historical query exposes the old name, while the current table schema exposes the new one. Matching names would incorrectly treat the `VARIANT`-bearing root as unprojected and let the native reader fail later. Matching the stable field ID identifies the same logical root across both schemas and correctly keeps this unsupported nested projection on Spark.

This PR deliberately does not claim to decode `VARIANT` natively. Direct projections, `variant_get` predicates, shredded or otherwise unsupported nested forms, and delete paths that would require unsupported `VARIANT` materialization continue to fail closed. Full native `VARIANT` execution remains follow-up work once Comet's Spark-facing representation and upstream native-reader support can carry the type end to end.

## How was this PR tested?

The Spark 4.0 Parquet SQL regression checks both native-plan selection and Spark-answer parity for omitted top-level `VARIANT` columns, columns appearing after the omitted field, supported nested siblings, nullable parent structs, partition columns, and file metadata. It also verifies that direct `VARIANT` projections and predicates still fall back, that SQL `NULL` remains distinct from a `VARIANT` containing JSON `null`, and that an ordinary `STRUCT<value: BINARY, metadata: BINARY>` is not mistaken for `VARIANT`.

The Spark 4.0 Iceberg coverage exercises supported native scalar projections and filters, projected nested-root fallback, empty and metadata-only projections, and the historical rename shown above. The historical-rename regression was run against the previous implementation first and failed because a native scan was incorrectly selected; it passes with field-ID-based validation. Iceberg 1.10 has a separate Spark-reader bug when an annotated `VARIANT` column is physically present but omitted from its projection, so the reference read intentionally includes the `VARIANT` column and removes it from the collected expected rows. This preserves a real Spark-versus-Comet comparison over the same files without relying on the unrelated upstream bug.

A Spark 3.5 Iceberg `VERSION AS OF` regression also passes, and the shared tests compile under both Spark 3.5 and Spark 4.0 despite their different `withSQLConf` return signatures.

```bash
mvn -o -ntp -Pspark-4.0 -Dtest=none \
  '-Dsuites=org.apache.comet.CometIcebergNativeSuite variant' test

mvn -o -ntp -Pspark-4.0 -Dtest=none \
  '-Dsuites=org.apache.comet.CometSqlFileTestSuite variant' test

mvn -o -ntp -Pspark-3.5 -Dtest=none \
  '-Dsuites=org.apache.comet.CometIcebergNativeSuite schema evolution - read old snapshot after column drop' test
```

The focused runs passed three Iceberg `VARIANT` cases, one Parquet SQL file, and one Spark 3.5 historical-snapshot case. Spotless, Scalastyle, and `git diff --check` also passed.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
@@ -550,6 +550,9 @@ case class CometScanRule(session: SparkSession)
           try {
             val fullSchema = IcebergReflection.toSparkSchema(metadata.tableSchema)
             val projectedDataColumns = scanExec.output.filterNot(_.isMetadataCol)
+            // DataTypeSupport recursively dispatches back to this override for struct fields,
+            // array elements, and map entries, so Variant is allowed at any nesting depth only
+            // when its entire top-level Iceberg field is unprojected.
             val unprojectedTypeChecker = new CometScanTypeChecker() {
               override def isTypeSupported(
                   dt: DataType,
--- a/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
@@ -540,15 +540,30 @@ case class CometScanRule(session: SparkSession)
             }
           }
 
-        // Comet serializes the whole table/scan schema to native, not just projected columns, so a
-        // type the native reader does not support (e.g. variant) breaks the scan even when that
-        // column is not projected. The readSchema allow-list only covers projected columns, so run
-        // the same allow-list over the full schema Comet may serialize. Reflection failure also
-        // falls back.
+        // The whole Iceberg table schema is serialized to native, but iceberg-rust can represent
+        // Variant in that schema as long as no projected field contains one. Check projected
+        // roots strictly and allow Variant only under entirely unprojected roots. Other unsupported
+        // types still fail closed everywhere. An empty data projection is also strict because
+        // iceberg-rust currently interprets an empty field-id list as a request for every column.
         val schemaTypesSupported =
           try {
             val fullSchema = IcebergReflection.toSparkSchema(metadata.tableSchema)
-            typeChecker.isSchemaSupported(fullSchema, fallbackReasons)
+            val projectedDataColumns = scanExec.output.filterNot(_.isMetadataCol)
+            val unprojectedTypeChecker = new CometScanTypeChecker() {
+              override def isTypeSupported(
+                  dt: DataType,
+                  name: String,
+                  reasons: ListBuffer[String]): Boolean =
+                isVariantType(dt) || super.isTypeSupported(dt, name, reasons)
+            }
+            val resolver = session.sessionState.conf.resolver
+
+            fullSchema.fields.forall { field =>
+              val isProjected = projectedDataColumns.isEmpty ||
+                projectedDataColumns.exists(attr => resolver(attr.name, field.name))
+              val checker = if (isProjected) typeChecker else unprojectedTypeChecker
+              checker.isTypeSupported(field.dataType, field.name, fallbackReasons)
+            }
           } catch {
             case e: Exception =>
               fallbackReasons += "Iceberg reflection failure: could not verify column " +
@@ -715,7 +730,7 @@ case class CometScanRule(session: SparkSession)
             true
         }
 
-        // Check for unsupported struct types in delete files
+        // Check for unsupported struct and Variant types in delete files
         val deleteFileTypesSupported = {
           var hasUnsupportedDeletes = false
 
@@ -766,12 +781,13 @@ case class CometScanRule(session: SparkSession)
                     }
                     fieldInfo match {
                       case Some((fieldName, fieldType)) =>
-                        if (fieldType.contains("struct")) {
+                        if (fieldType.contains("struct") || fieldType.equalsIgnoreCase(
+                            "variant")) {
                           hasUnsupportedDeletes = true
                           fallbackReasons +=
                             s"Equality delete on unsupported column type '$fieldName' " +
                               s"($fieldType) is not yet supported by iceberg-rust. " +
-                              "Struct types in equality deletes " +
+                              "Struct and Variant types in equality deletes " +
                               "require datum conversion support that is not yet implemented."
                         }
                       case None =>
--- a/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
@@ -541,8 +541,9 @@ case class CometScanRule(session: SparkSession)
           }
 
         // The whole Iceberg table schema is serialized to native, but iceberg-rust can represent
-        // Variant in that schema as long as no projected field contains one. Check projected
-        // roots strictly and allow Variant only under entirely unprojected roots. Other unsupported
+        // Variant in that schema as long as no projected field contains one. Match projected
+        // roots by field ID so historical snapshots still identify renamed columns, check them
+        // strictly, and allow Variant only under entirely unprojected roots. Other unsupported
         // types still fail closed everywhere. An empty data projection is also strict because
         // iceberg-rust currently interprets an empty field-id list as a request for every column.
         val schemaTypesSupported =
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -540,15 +540,40 @@ case class CometScanRule(session: SparkSession)
...
+        // roots by field ID so historical snapshots still identify renamed columns, check them
+        // strictly, and allow Variant only under entirely unprojected roots. Other unsupported
+        // types still fail closed everywhere. An empty data projection is also strict because
+        // iceberg-rust currently interprets an empty field-id list as a request for every column.
         val schemaTypesSupported =
           try {
             val fullSchema = IcebergReflection.toSparkSchema(metadata.tableSchema)
-            typeChecker.isSchemaSupported(fullSchema, fallbackReasons)
+            val projectedDataColumns = scanExec.output.filterNot(_.isMetadataCol)
+            val unprojectedTypeChecker = new CometScanTypeChecker() {
```
parthchandra, line 556: This is being called for all the top level variant fields. What about a nested variant field?

spark/src/test/resources/sql-tests/expressions/misc/variant.sql
```diff
@@ -22,14 +22,26 @@
 -- MinSparkVersion: 4.0
 
 statement
-CREATE TABLE test_variant(id INT, v VARIANT) USING parquet
+CREATE TABLE test_variant(id INT, v VARIANT, tail STRING) USING parquet
```
parthchandra, line 25: Should we also add a test with `ARRAY<VARIANT>` and/or `MAP<STRING, VARIANT>`

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
sunchao: Nested Variant fields are handled recursively. `DataTypeSupport.isTypeSupported` dispatches back to this overridden checker for struct fields, array elements, and map keys/values, so an unprojected top-level Iceberg field may contain Variants at any depth. If that top-level field is projected, the strict checker still rejects it and the scan falls back to Spark. I clarified this in the comment and extended the Iceberg regression to cover nested structs, `ARRAY<VARIANT>`, and `MAP<STRING, VARIANT>`.

spark/src/test/resources/sql-tests/expressions/misc/variant.sql
sunchao: Added explicit `ARRAY<VARIANT>` and `MAP<STRING, VARIANT>` coverage for both Parquet and Iceberg. The tests verify that native scans remain enabled when the collection columns are unprojected and fall back to Spark when either collection is projected. They also cover SQL-null elements and map values, Variant JSON null, and null collection roots.

spark/src/test/scala/org/apache/comet/CometIcebergNativeSuite.scala
```diff
@@ -5160,6 +5160,165 @@ class CometIcebergNativeSuite
...
+               array(CAST(NULL AS VARIANT)), map('sql-null', CAST(NULL AS VARIANT))),
+              (3, NULL, NULL, NULL)
+          """)
+
+          var sparkScalarRows = Seq.empty[Row]
+          // Iceberg 1.10 cannot materialize nested Variant values in arrays or maps, but
+          // reading their sizes still projects every Variant-bearing root for Spark parity.
+          withSQLConf(CometConf.COMET_ENABLED.key -> "false") {
+            sparkScalarRows = spark
+              .sql(s"SELECT id, nested, size(variants), size(variants_by_key) " +
```
peterxcli: this would fix the ci lint failure
```suggestion
              .sql("SELECT id, nested, size(variants), size(variants_by_key) " +
```

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
```diff
@@ -766,12 +794,13 @@ case class CometScanRule(session: SparkSession)
                     }
                     fieldInfo match {
                       case Some((fieldName, fieldType)) =>
-                        if (fieldType.contains("struct")) {
+                        if (fieldType.contains("struct") || fieldType.equalsIgnoreCase(
+                            "variant")) {
```
peterxcli, line 798: Could we add a new test for this new gate to test spark fallback on this case?

peterxcli (commented):
Other LGTM

spark/src/test/scala/org/apache/comet/CometIcebergNativeSuite.scala
sunchao: Fixed by removing the unnecessary `s` prefix from the first SQL string. The second string still interpolates `$table`.

spark/src/main/scala/org/apache/comet/rules/CometScanRule.scala
sunchao: Added `variant equality deletes fall back to Spark`, which writes a real Iceberg equality-delete file keyed on a `VARIANT` column, verifies the same scalar projection is native before the delete and falls back afterward, and checks the exact fallback reason. The assertion is intentionally plan-only because Iceberg itself lacks a Variant equality comparator. Variant values are constructed reflectively so the Spark 3.4 / Iceberg 1.5 profile still compiles.

spark/src/test/scala/org/apache/comet/CometIcebergNativeSuite.scala
```diff
@@ -5160,6 +5160,224 @@ class CometIcebergNativeSuite
...
+          val value = variantsClass
+            .getMethod("of", java.lang.Integer.TYPE)
+            .invoke(null, Integer.valueOf(2))
+          val variant = Class
+            .forName("org.apache.iceberg.variants.Variant")
+            .getMethod(
+              "of",
+              Class.forName("org.apache.iceberg.variants.VariantMetadata"),
+              Class.forName("org.apache.iceberg.variants.VariantValue"))
+            .invoke(null, metadata, value)
```
peterxcli: asked ai how to clean this up. elsewhere.
```suggestion
          val records = IcebergGenerics
		    .read(Spark3Util.loadIcebergTable(spark, table))
		    .where(Expressions.equal("id", 2L))
		    .select("data")
		    .build()
		
		  val variant =
		    try {
		      val rows = records.iterator()
		      assert(rows.hasNext)
		      rows.next().getField("data")
		    } finally {
		      records.close()
		    }
```

peterxcli (approved):
just one nit, other lgtm, thanks for updating the pr

parthchandra (approved):
lgtm. pending ci

spark/src/test/scala/org/apache/comet/CometIcebergNativeSuite.scala
sunchao: Done, thanks!

Issue #2837, opened 2025-12-02
Add support for map inputs to `explode`


andygrove:
### What is the problem the feature request solves?

PR https://github.com/apache/datafusion-comet/pull/2836 adds initial support for `explode` for array inputs. We should also add support for map inputs.

rich7420:
take

PR #5381, opened 2026-08-16
test: add explode operator microbenchmark

andygrove:

N/A. This adds a benchmark for an existing operator, so there is no behavior change to track.

## Rationale for this change

`CometExplodeExec` has no benchmark. Generator cost is driven by dimensions that other operator
benchmarks do not exercise: how far each input row fans out, whether the generator emits a
position column or preserves rows with null and empty arrays, the element type being unnested,
how deeply the exploded array sits in its input row, and how many columns have to be replicated
alongside the generated one. Without numbers for those, there is no way to tell whether a change
to the operator helped, and no way to see which shapes Comet declines to convert.

A new `CometExplodeBenchmark` with 15 cases in five groups:

- **Fan-out**: `explode` over `array<bigint>` of length 2, 10, and 100.
- **Generator variants**: `explode`, `posexplode`, `explode_outer`, `posexplode_outer`.
- **Element type**: `array<bigint>`, `array<string>`, `array<struct<a bigint, b string>>`.
- **Nested input**: an `array<struct>` at depth 1, the same array eight struct accessors down,
  and a second generator chained onto the first. 100K rows rather than 256K, because each row
  carries 20 events.
- **Carried columns**: `explode` alone against `explode` plus three passthrough columns.

One in ten rows holds a null array and another one in ten holds an empty array. Without that,
`explode_outer` and `explode` would produce identical output and the variant group would be
measuring the same query twice. The empty array is built with `slice`, not `array()`, because
`array()` types as `array<null>`.

Each array column gets its own temp view rather than sharing one wide table, so a case is never
charged for scanning an array column it does not read. The nested group is the exception: its
depth-1 and depth-8 columns hold the same array at two depths, and keeping them in one file is
what makes them the same array.

### Isolating the operator

Six things had to be arranged before a case measures the dimension it is named after. Each is
documented at the point in the file that depends on it.

**Cases count the generated columns rather than writing them.** `.noop()` writes
`DataWriter[InternalRow]`, so the Comet arm would convert every generated row while the Spark arm,
whose `GenerateExec` already emits rows, does not. That is 419K conversions at fan-out 2 and 21M
at fan-out 100 — a cost scaling with the very dimension the group exists to measure, and not one
a scan baseline could subtract. Terminating in an aggregate puts the only row boundary above the
final exchange. `CometColumnarToRowBenchmark` measures that conversion on its own.

**The position column is summed, not counted.** `pos` is declared non-null, so `NullPropagation`
rewrites `count(pos)` to `count(1)` and no position value is ever read; `Count` does not read a
non-null argument's value in any case. That matters because the two engines produce positions
very differently: Spark hands over the loop index it already has, while Comet materializes a
parallel `List<Int32>` through `ListPositionsExpr` and unnests it alongside the values.
`sum(pos)` is `9437130` for `posexplode` and `posexplode_outer` alike — the outer variant's extra
52,430 rows carry a null position, which `sum` skips.

**Every sink is checked before it is timed.** Each case declares the aggregate row it must
produce, and both engines run it once, untimed, against that row. A sink that quietly stops
consuming what it names still reports a rate, just a better one, and the results table cannot
tell the difference. `count(pos)` was exactly that until this PR.

**`InferFiltersFromGenerate` is excluded, for both engines.** It infers `size(arr) > 0 AND arr IS
NOT NULL` below a generator but matches only on `outer = false`, so with it enabled `explode` and
`posexplode` received 209,714 rows and an extra filter while their outer variants received all
262,144 — and the rate was normalized on rows the non-outer arms never processed. Worth knowing
when reading these numbers: real queries do get that filter, so a plain `explode` in production
usually sees an array column with no nulls and no empty rows.

**The struct dataset's string field matches the string dataset's.** A field of `s1` through `s10`
holds 10 distinct values where the standalone string column holds 260,000-odd, which puts one
under the writer's 1 MiB dictionary page threshold and the other over it. The element-type group
would have been comparing Parquet encodings as much as element types.

**Both carried-column cases scan the same four columns.** `SELECT explode(arr)` alone would prune
`k`, `s` and `v` from the scan, so the difference between the two cases would include an entire
extra Parquet read of three columns, one of them 262K distinct strings. An always-true filter
referencing all three sits below the generator; column pruning then drops them from the
generator's input, so the "alone" case still does not replicate them. What this does not
equalize is the three extra counts the carried case runs over the generated rows, which are
cheaper than the three gathers they exist to measure but are not free.

Confirmed on the executed plans — both cases report
`ReadSchema: struct<arr:array<bigint>,k:bigint,s:string,v:bigint>`, and only the carried case
reaches `CometExplode` with the columns attached:

```
alone:    CometExplode explode(arr#36), [col#17532L], [col#17532L]
            +- CometProject [arr#36], [arr#36]
                 +- CometFilter [arr#36, k#37L, s#38, v#39L], ...

carried:  CometExplode explode(arr#36), [col#18682L], [k#37L, s#38, v#39L, col#18682L]
            +- CometFilter [arr#36, k#37L, s#38, v#39L], ...
```

Every column a case counts is nullable. `NullPropagation` rewrites `count(c)` to `count(1)` when
`c` is not, which would leave the carried columns unreferenced and let pruning drop them before
the generator, and would also split the variant group, since `outer` forces the generated column
nullable and the plain variants do not.

### The nested group

The shape is the one asked for in review: a customer profile whose event list sits at
`profile.account.settings.preferences.notifications.activity.sessions.events`, holding a second
array of four-field structs inside each element.

The requested outer container was a map keyed by platform, which is where a real schema would
put it. It is an `array<struct<platform, entries>>` here because `CometExplodeExec` refuses map
inputs (#2837), so with a map the first generator falls back to Spark's `GenerateExec` and the
Comet arm stops being a measurement of Comet at all. Worth adding once #2837 lands.

The whole event struct is counted rather than one of its fields, so nested schema pruning cannot
narrow the exploded element and leave the case measuring a two-column gather. It does prune the
siblings no case reads, confirmed on the executed plan: the depth-8 scan projects `profile` as
`struct<account:struct<...sessions:struct<events:array<...>>>>` with `billing`, `addresses` and
`device` gone, and the depth-1 scan does not project `profile` at all. The two depths therefore
read the same leaf payload and differ only in the struct wrapper.

All 15 cases complete, none trips the harness warning for a plan that is not fully Comet native,
and every case's aggregate matches the row it declares under both engines.

```sh
SPARK_GENERATE_BENCHMARK_FILES=1 make benchmark-org.apache.spark.sql.benchmark.CometExplodeBenchmark
```

Apple M3 Max, JDK 17, `local[1]`, Spark 4.1 / Scala 2.13, release native build (Best ms):

| Case | Spark | Comet | Relative |
| ---- | ----- | ----- | -------- |
| fan-out 2 / 10 / 100 | 34 / 41 / 137 | 23 / 35 / 188 | 1.5X / 1.2X / 0.7X |
| explode / posexplode / explode_outer / posexplode_outer | 41 / 40 / 42 / 47 | 36 / 43 / 45 / 53 | 1.1X / 0.9X / 0.9X / 0.9X |
| bigint / string / struct | 42 / 63 / 61 | 37 / 61 / 86 | 1.1X / 1.0X / 0.7X |
| depth 1 / depth 8 / chained | 69 / 84 / 96 | 156 / 167 / 181 | 0.4X / 0.5X / 0.5X |
| alone / plus 3 carried columns | 49 / 51 | 44 / 57 | 1.1X / 0.9X |

**These supersede the table I posted earlier, which overstated Comet's margin.** The earlier run
was taken on a contended machine, and at `local[1]` contention lands almost entirely on the Spark
arm. Across five runs the Comet arm reproduces to within a few ms on every case while the Spark
arm swings by up to 2x. To rule out the review fixes as the cause I also re-ran the **unchanged**
previous revision in the same environment and got today's numbers, not the old ones.

The benchmark has a built-in noise detector, which is worth using before trusting any run of it:
`explode array<bigint>[10]` is timed three times, once in each of the first three groups. In the
run above those three read 41 / 41 / 42 ms for Spark and 35 / 36 / 37 ms for Comet. When they
disagree, the machine was busy and the table should be thrown away — in one of my five runs the
same case came out at 40 ms and 71 ms for Comet in two different groups.

Reading the table: Comet wins at low fan-out over `array<bigint>` and loses everywhere the
element is wide or the fan-out is large. It is roughly 2x slower than Spark on the nested
`array<struct>` cases, ~1.4x slower on `array<struct>` at fan-out 10, and ~1.4x slower at fan-out
100. `posexplode` and the outer variants are a small loss. Those directions hold in every run and
look like the first things worth optimizing.

Correctness of the query forms themselves is already covered by `CometGenerateExecSuite`, which
tests explode and explode_outer over simple, empty, and null arrays.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
@@ -21,6 +21,9 @@ package org.apache.spark.sql.benchmark
 
 import java.io.File
 
+import org.apache.spark.sql.catalyst.optimizer.InferFiltersFromGenerate
+import org.apache.spark.sql.internal.SQLConf
+
 /**
  * Benchmark to measure performance of Comet's explode operator (`CometExplodeExec`) against
  * Spark's `GenerateExec`, across the dimensions that drive generator cost: fan-out, generator
@@ -30,11 +33,25 @@ import java.io.File
  *   SPARK_GENERATE_BENCHMARK_FILES=1 make benchmark-org.apache.spark.sql.benchmark.CometExplodeBenchmark
  * }}}
  *
- * `runExpressionBenchmark` reports whole-query totals, so the times below also include the
- * Parquet scan, the result transfer, and the per-iteration query planning. That fixed cost is a
- * large share of the total at fan-out 2, where it compresses the ratio between the two engines,
- * and a small one at fan-out 100. Issue #5363 tracks reporting operator cost against a scan
- * baseline instead; when that lands, this paragraph should go.
+ * Every case counts the generated columns rather than writing them, so the only row boundary is
+ * one row per partition. Writing them with `.noop()` would put a columnar-to-row conversion of
+ * every generated row inside the Comet arm and none inside the Spark arm, whose `GenerateExec`
+ * already emits rows: 419K conversions at fan-out 2 and 21M at fan-out 100, scaling with the very
+ * dimension the case is meant to isolate. `CometColumnarToRowBenchmark` measures that conversion
+ * on its own.
+ *
+ * Times are still whole-query totals and include the Parquet scan, the counting aggregate and its
+ * exchange, and per-iteration planning. The scan is a large share of the total at fan-out 2,
+ * where it compresses the ratio between the two engines, and a small one at fan-out 100. Issue
+ * #5363 tracks reporting operator cost against a scan baseline instead; when that lands, this
+ * paragraph should go.
+ *
+ * `InferFiltersFromGenerate` is excluded, for both engines. It infers `size(arr) > 0 AND arr IS
+ * NOT NULL` below a generator but matches only on `outer = false`, so leaving it enabled hands
+ * `explode` and `posexplode` 209,714 rows and an extra filter while their outer variants get all
+ * 262,144 — a comparison of two different plans, reported against a row count neither arm
+ * processes. Worth knowing when reading these numbers: real queries do get that filter, so a
+ * plain `explode` in production usually sees an array column with no nulls and no empty rows.
  */
 object CometExplodeBenchmark extends CometBenchmarkBase {
 
@@ -50,15 +67,40 @@ object CometExplodeBenchmark extends CometBenchmarkBase {
    *
    * The empty array is built with `slice`, not `array()`, because `array()` types as
    * `array<null>` and would give that row's column a different element type.
+   *
+   * Elements are wrapped in a never-taken null branch so the array types as `containsNull`; see
+   * [[nullableExpr]].
    */
   private def arrayColumn(elementExpr: String, len: Int): String = {
-    val full = s"transform(sequence(1, $len), x -> $elementExpr)"
+    val full = s"transform(sequence(1, $len), x -> ${nullableExpr(elementExpr, "x = 0")})"
     s"""CASE
        |  WHEN id % 10 = 0 THEN NULL
        |  WHEN id % 10 = 1 THEN slice($full, 1, 0)
        |  ELSE $full
        |END AS arr""".stripMargin
   }
+
+  /**
+   * Types `expr` as nullable without ever evaluating to null, `guard` being a predicate that is
+   * never true.
+   *
+   * Every column these queries count has to be nullable: `NullPropagation` rewrites `count(c)` to
+   * `count(1)` when `c` is not, and the counted column then has no reader at all. For the
+   * carried-column case that is fatal, because column pruning goes on to drop `k`, `s` and `v`
+   * from the generator's input, which is the entire dimension being measured. It would also split
+   * the variant group, since `outer` forces the generated column nullable and the plain variants
+   * do not. Parquet columns are usually optional in practice anyway.
+   */
+  private def nullableExpr(expr: String, guard: String): String = s"IF($guard, NULL, $expr)"
+
+  /**
+   * The string element, shared by the string and struct datasets so that the element-type cases
+   * differ in element type alone. A struct field of `s1` through `s10` would hold 10 distinct
+   * values against this column's 260,000-odd, which stays under the writer's 1 MiB dictionary
+   * page threshold where this one does not, so the comparison would also be measuring the
+   * difference between a dictionary-encoded column and a plain one.
+   */
+  private val stringElement = "concat('str_', CAST(id + x AS STRING))"
 
   /**
    * The temp views the benchmark reads, each with the expressions that build it.
@@ -70,14 +112,13 @@ object CometExplodeBenchmark extends CometBenchmarkBase {
     "arr_len2" -> Seq(arrayColumn("id + x", 2)),
     "arr_len10" -> Seq(arrayColumn("id + x", 10)),
     "arr_len100" -> Seq(arrayColumn("id + x", 100)),
-    "arr_str10" -> Seq(arrayColumn("concat('str_', CAST(id + x AS STRING))", 10)),
-    "arr_struct10" -> Seq(
-      arrayColumn("struct(id + x AS a, concat('s', CAST(x AS STRING)) AS b)", 10)),
+    "arr_str10" -> Seq(arrayColumn(stringElement, 10)),
+    "arr_struct10" -> Seq(arrayColumn(s"struct(id + x AS a, $stringElement AS b)", 10)),
     "arr_carry" -> Seq(
       arrayColumn("id + x", 10),
-      "id AS k",
-      "CAST(id AS STRING) AS s",
-      "id * 2 AS v"))
+      s"${nullableExpr("id", "id < 0")} AS k",
+      s"${nullableExpr("CAST(id AS STRING)", "id < 0")} AS s",
+      s"${nullableExpr("id * 2", "id < 0")} AS v"))
 
   /** Writes `selectExprs` over `numRows` rows to Parquet and registers it as a temp view. */
   private def createView(dir: File, name: String, selectExprs: Seq[String]): Unit = {
@@ -86,48 +127,91 @@ object CometExplodeBenchmark extends CometBenchmarkBase {
     spark.read.parquet(path).createOrReplaceTempView(name)
   }
 
+  /**
+   * A query that applies `generator` to `view`'s `arr` column, carries `carried` through the
+   * generator alongside it, and counts every column that comes out.
+   *
+   * The position column of the `posexplode` variants is counted too, because a generator whose
+   * second output nothing reads is not the generator being named.
+   */
+  private def countGenerated(
+      generator: String,
+      view: String,
+      carried: Seq[String] = Nil,
+      where: Option[String] = None): String = {
+    val generated = if (generator.startsWith("posexplode")) Seq("pos", "col") else Seq("col")
+    val alias =
+      if (generated.length == 1) s"AS ${generated.head}"
+      else generated.mkString("AS (", ", ", ")")
+    val projectList = (carried :+ s"$generator(arr) $alias").mkString(", ")
+    val filter = where.map(w => s" WHERE $w").getOrElse("")
+    val counts = (carried ++ generated).map(c => s"count($c)").mkString(", ")
+    s"SELECT $counts FROM (SELECT $projectList FROM $view$filter)"
+  }
+
   override def runCometBenchmark(mainArgs: Array[String]): Unit = {
     withTempPath { dir =>
       withTempTable(datasets.map(_._1): _*) {
         datasets.foreach { case (name, selectExprs) => createView(dir, name, selectExprs) }
 
-        // Cardinality is input rows for every case, so the numbers are per scanned row rather
-        // than per generated row. Fan-out is named in the case title: the 100-element case emits
-        // roughly 50 times as many rows as the 2-element case from the same 256K inputs.
-        runBenchmark("Explode - fan-out") {
-          Seq(2, 10, 100).foreach { len =>
-            runExpressionBenchmark(
-              s"explode array<bigint>[$len]",
-              numRows,
-              s"SELECT explode(arr) FROM arr_len$len")
-          }
-        }
+        // `runExpressionBenchmark` appends ConstantFolding to whatever the caller has already
+        // excluded, and applies the result to both arms, so setting this here excludes both.
+        withSQLConf(SQLConf.OPTIMIZER_EXCLUDED_RULES.key -> InferFiltersFromGenerate.ruleName) {
 
-        runBenchmark("Explode - generator variants") {
-          Seq("explode", "posexplode", "explode_outer", "posexplode_outer").foreach { generator =>
-            runExpressionBenchmark(
-              s"$generator array<bigint>[10]",
-              numRows,
-              s"SELECT $generator(arr) FROM arr_len10")
-          }
-        }
-
-        runBenchmark("Explode - element type") {
-          Seq("bigint" -> "arr_len10", "string" -> "arr_str10", "struct" -> "arr_struct10")
-            .foreach { case (elementType, view) =>
+          // Cardinality is input rows for every case, so the numbers are per scanned row rather
+          // than per generated row. Fan-out is named in the case title: the 100-element case
+          // emits roughly 50 times as many rows as the 2-element case from the same 256K inputs.
+          runBenchmark("Explode - fan-out") {
+            Seq(2, 10, 100).foreach { len =>
               runExpressionBenchmark(
-                s"explode array<$elementType>[10]",
+                s"explode array<bigint>[$len]",
                 numRows,
-                s"SELECT explode(arr) FROM $view")
+                countGenerated("explode", s"arr_len$len"))
             }
-        }
+          }
 
-        runBenchmark("Explode - carried columns") {
-          runExpressionBenchmark("explode alone", numRows, "SELECT explode(arr) FROM arr_carry")
-          runExpressionBenchmark(
-            "explode plus 3 carried columns",
-            numRows,
-            "SELECT k, s, v, explode(arr) FROM arr_carry")
+          runBenchmark("Explode - generator variants") {
+            Seq("explode", "posexplode", "explode_outer", "posexplode_outer").foreach {
+              generator =>
+                runExpressionBenchmark(
+                  s"$generator array<bigint>[10]",
+                  numRows,
+                  countGenerated(generator, "arr_len10"))
+            }
+          }
+
+          runBenchmark("Explode - element type") {
+            Seq("bigint" -> "arr_len10", "string" -> "arr_str10", "struct" -> "arr_struct10")
+              .foreach { case (elementType, view) =>
+                runExpressionBenchmark(
+                  s"explode array<$elementType>[10]",
+                  numRows,
+                  countGenerated("explode", view))
+              }
+          }
+
+          // Both cases read all four columns. The filter is always true and references k, s and
+          // v below the generator, where column pruning then drops them from the generator's
+          // input rather than replicating them; without it, `explode alone` would prune three
+          // columns from the Parquet scan and the difference between the two cases would be
+          // scan and string-decode work as much as replication. What the filter does not
+          // equalize is the three extra counts the carried case runs over the generated rows.
+          // Those are cheaper than the three gathers they exist to measure, but not free.
+          runBenchmark("Explode - carried columns") {
+            val readAllColumns = Some("k >= 0 AND v >= 0 AND length(s) > 0")
+            runExpressionBenchmark(
+              "explode alone",
+              numRows,
+              countGenerated("explode", "arr_carry", where = readAllColumns))
+            runExpressionBenchmark(
+              "explode plus 3 carried columns",
+              numRows,
+              countGenerated(
+                "explode",
+                "arr_carry",
+                carried = Seq("k", "s", "v"),
+                where = readAllColumns))
+          }
         }
       }
     }
--- a/spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
@@ -1,0 +1,169 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.spark.sql.benchmark
+
+import java.io.File
+
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.SparkSession
+
+import org.apache.comet.CometSparkSessionExtensions
+
+/**
+ * Benchmark to measure performance of Comet's explode operator (`CometExplodeExec`) against
+ * Spark's `GenerateExec`, across the dimensions that drive generator cost: fan-out, generator
+ * variant, element type, and the number of columns replicated alongside the generated one. To
+ * run:
+ * {{{
+ *   SPARK_GENERATE_BENCHMARK_FILES=1 make benchmark-org.apache.spark.sql.benchmark.CometExplodeBenchmark
+ * }}}
+ *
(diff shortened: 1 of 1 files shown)


sunchao (commented):
## Summary

This first review evaluates the new explode benchmark at commit `bf0324024d27985096bc590153110a3fe9b7c5df`. The coverage dimensions are useful, but four benchmark-design issues prevent several reported comparisons from isolating the native explode operator.

## Prior state and problem

`CometExplodeExec` previously lacked a dedicated Scala microbenchmark, making it difficult to characterize fan-out, generator variants, nested element types, and columns carried through generation. Existing expression benchmarks already include Parquet scan and result-transfer costs, which become especially significant for an operator that multiplies its input cardinality.

## Design approach

The change adds a standalone `CometExplodeBenchmark` that materializes deterministic 256K-row Parquet-backed temporary views and compares Spark with Comet through the shared `runExpressionBenchmark` harness. Its cases cover fan-outs of 2, 10, and 100; four generator variants; bigint, string, and struct elements; and zero versus three carried columns.

The generated data includes exactly 26,215 null arrays and 26,215 empty arrays, with 209,714 nonempty arrays, and the relevant native explode operator is enabled by default. However, Catalyst filters null and empty arrays before non-outer generators, the noop sink forces Comet's generated columnar output through row conversion, and mismatched source distributions and projected Parquet schemas introduce additional work that is unrelated to the intended operator comparisons.

## Key design decisions

The benchmark intentionally stores each array shape in a separate Parquet-backed view, reports rates per original input row, and delegates warmup, iteration control, and Spark-versus-Comet configuration to the common benchmark helper. Null and empty arrays are deliberately included to distinguish outer generators, while the carried-column case reuses a wider source view.

## Implementation sketch

`arrayColumn` builds typed arrays with deterministic null and empty branches; `createView` writes each generated dataset and registers a temporary view. Four `runBenchmark` groups then call `runExpressionBenchmark`, whose Spark and Comet cases execute the corresponding SQL query via Spark's row-oriented noop writer.

This is a benchmark-only change and does not modify production explode semantics. The benchmark-only CI route compiles and lints against Spark 4.0, while the broader Spark integration matrix is intentionally skipped; reported rates use 262,144 original input rows even though optimized non-outer generators receive only 209,714 rows.

## Suggested improvements

Keep exploded results columnar or use a matched conversion baseline; apply generator-filter optimization controls symmetrically so all variants receive the intended inputs; align string distributions and Parquet encoding across element-type datasets; and hold the projected scan schema constant, or subtract a matched scan baseline, when comparing carried columns. The inline P2 comments identify the exact affected cases and their concrete measurement consequences.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
```diff
@@ -0,0 +1,135 @@
...
+   * Each array column gets its own view rather than sharing one wide table, so that a case is
+   * never charged for scanning an array column it does not read.
+   */
+  private val datasets: Seq[(String, Seq[String])] = Seq(
+    "arr_len2" -> Seq(arrayColumn("id + x", 2)),
+    "arr_len10" -> Seq(arrayColumn("id + x", 10)),
+    "arr_len100" -> Seq(arrayColumn("id + x", 100)),
+    "arr_str10" -> Seq(arrayColumn("concat('str_', CAST(id + x AS STRING))", 10)),
+    "arr_struct10" -> Seq(
+      arrayColumn("struct(id + x AS a, concat('s', CAST(x AS STRING)) AS b)", 10)),
```
sunchao: [P2] Match string cardinality across the element-type datasets

`arr_str10` contains 262,151 distinct `str_<id+x>` values, while this struct's string field contains only the 10 values `s1` through `s10`, across the same 2,097,140 non-null elements. The default Parquet writer enables dictionaries with a 1 MiB dictionary-page threshold, so the struct field stays dictionary-encoded while the high-cardinality standalone string field exceeds that threshold. Consequently the advertised string-versus-struct comparison also measures substantially different Parquet encoding, decoding, compression, and scan work. Make `struct.b` row-varying like the standalone string case, or explicitly match the value distribution and encoding before comparing element types.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
```diff
@@ -0,0 +1,135 @@
...
+    withTempPath { dir =>
+      withTempTable(datasets.map(_._1): _*) {
+        datasets.foreach { case (name, selectExprs) => createView(dir, name, selectExprs) }
+
+        // Cardinality is input rows for every case, so the numbers are per scanned row rather
+        // than per generated row. Fan-out is named in the case title: the 100-element case emits
+        // roughly 50 times as many rows as the 2-element case from the same 256K inputs.
+        runBenchmark("Explode - fan-out") {
+          Seq(2, 10, 100).foreach { len =>
+            runExpressionBenchmark(
```
sunchao: [P2] Isolate explode from per-generated-row columnar-to-row conversion

`runExpressionBenchmark` executes `.noop()`, which writes to a `DataWriter[InternalRow]`. Therefore the Comet case converts every generated columnar output row to a Spark row, whereas Spark's `GenerateExec` already emits rows; the existing `CometColumnarToRowBenchmark` deliberately uses this same `.noop()` pattern to measure that conversion. Here fan-out 2 produces 419,428 rows, while fan-out 100 produces 20,971,400 rows, so conversion work scales 50x with the dimension attributed to explode and also varies with strings, structs, and carried columns. This is not fixed result-transfer overhead, and a scan-only baseline will not remove it. Consume generated batches without a row boundary, or include a matched columnar-to-row baseline so the benchmark isolates `CometExplodeExec`.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
```diff
@@ -0,0 +1,135 @@
...
+              s"explode array<bigint>[$len]",
+              numRows,
+              s"SELECT explode(arr) FROM arr_len$len")
+          }
+        }
+
+        runBenchmark("Explode - generator variants") {
+          Seq("explode", "posexplode", "explode_outer", "posexplode_outer").foreach { generator =>
```
sunchao: [P2] Prevent optimizer filtering from changing generator inputs

Catalyst's `InferFiltersFromGenerate` inserts `size(arr) > 0 AND arr IS NOT NULL` below `explode` and `posexplode`, but explicitly does not apply to their outer variants. With these generated datasets, the non-outer operators therefore receive only 209,714 rows, while `explode_outer` and `posexplode_outer` receive all 262,144; none of the 52,430 null or empty arrays reaches the non-outer operator. The purported generator-variant comparison consequently mixes an extra upstream filter and different operator input cardinalities, and its non-outer rate is normalized using rows the generator never processes. Exclude `InferFiltersFromGenerate` symmetrically for Spark and Comet, preserving the helper's existing `ConstantFolding` exclusion, or otherwise give both variants matched operator inputs.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
```diff
@@ -0,0 +1,135 @@
...
+                s"SELECT explode(arr) FROM $view")
+            }
+        }
+
+        runBenchmark("Explode - carried columns") {
+          runExpressionBenchmark("explode alone", numRows, "SELECT explode(arr) FROM arr_carry")
+          runExpressionBenchmark(
+            "explode plus 3 carried columns",
+            numRows,
+            "SELECT k, s, v, explode(arr) FROM arr_carry")
```
sunchao: [P2] Keep the Parquet scan schema constant when measuring carried columns

`SELECT explode(arr)` prunes `k`, `s`, and `v` from the Parquet scan, whereas this query must scan and decode all three additional columns before explode runs. Across 262,144 input rows, that adds 786,432 scalar values, including 262,144 distinct strings, although both cases produce the same 2,097,140 generated rows. The measured difference therefore combines extra Parquet I/O and string decoding with carried-column replication, so it cannot isolate the stated carried-column dimension. Keep the projected scan schema identical between the two cases, or subtract a schema-matched scan baseline.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
andygrove: Fixed. The struct's string field is now the same expression as the standalone string column, so both hold the same 260,000-odd distinct values and neither stays under the dictionary page threshold. The two datasets now differ only in that the struct wraps the string alongside a bigint, which is what I wanted the case to isolate.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
andygrove: You're right, and this was the one that mattered most. Every case now terminates in a count aggregate instead of `.noop()`, so the only row boundary is above the final exchange rather than one per generated row. A matched baseline wouldn't have worked here for the reason you give: the conversion scales 50x across the fan-out group, so there is no fixed quantity to subtract.

The old numbers sat flat at roughly 2x across every dimension, which in hindsight is exactly what a shared cost proportional to generated rows looks like. They now vary from 1.2x to 1.8x, and the variation is the useful part: Comet's margin narrows as fan-out grows, posexplode costs it 6ms over explode where Spark gets slightly cheaper, and posexplode_outer and array<struct> are its two weakest shapes.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
andygrove: Fixed by excluding the rule, symmetrically. `runExpressionBenchmark` appends ConstantFolding to whatever the caller already has in `spark.sql.optimizer.excludedRules` and applies the result to both arms, so setting it once around the four groups covers Spark and Comet and keeps the existing exclusion.

Worth recording somewhere, since it cuts the other way: the filter is not a benchmark artifact. It fires in any real query with a non-outer generator over an attribute, so a plain `explode` in production usually reaches the operator with an array column that has no nulls and no empty rows. I've noted that in the class comment. It also means the null-free path is worth optimizing for specifically.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
andygrove: Fixed by holding the scan schema constant. Both cases now sit under an always-true filter referencing k, s and v, so the scan reads and decodes all four columns either way. Column pruning still drops the three from the generator's input in the `alone` case, so it does not replicate them, which is the difference I wanted to keep.

Checked on the executed plans rather than assumed. Both report `ReadSchema: struct<arr:array<bigint>,k:bigint,s:string,v:bigint>`, and only the carried case reaches the operator with the columns attached:

```
alone:    CometExplode explode(arr#7), [col#12L], [col#12L]
            +- CometProject [arr#7], [arr#7]
                 +- CometFilter [arr#7, k#8L, s#9, v#10L], ...

carried:  CometExplode explode(arr#7), [col#21L], [k#8L, s#9, v#10L, col#21L]
            +- CometFilter [arr#7, k#8L, s#9, v#10L], ...
```

What this doesn't equalize is the three extra counts the carried case runs over the generated rows. They're cheaper than the three gathers they're there to measure, but not free, and I've said so in the comment rather than pretend otherwise.

One thing that fell out of the rewrite and isn't obvious: every counted column has to be nullable. `NullPropagation` turns `count(c)` into `count(1)` when `c` isn't, which would leave k, s and v unreferenced and let pruning drop them before the generator, quietly deleting this dimension entirely. The data generator now types them nullable without ever producing a null.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
```diff
@@ -0,0 +1,219 @@
...
+      view: String,
+      carried: Seq[String] = Nil,
+      where: Option[String] = None): String = {
+    val generated = if (generator.startsWith("posexplode")) Seq("pos", "col") else Seq("col")
+    val alias =
+      if (generated.length == 1) s"AS ${generated.head}"
+      else generated.mkString("AS (", ", ", ")")
+    val projectList = (carried :+ s"$generator(arr) $alias").mkString(", ")
+    val filter = where.map(w => s" WHERE $w").getOrElse("")
+    val counts = (carried ++ generated).map(c => s"count($c)").mkString(", ")
```
sunchao: [P2] Make the position aggregate consume position values

Could both position cases use a value-dependent aggregate such as `sum(pos)`, with an untimed result and plan check? The nullable array-element wrapper changes `col`, but `posexplode` still declares `pos` non-null. In the maintained Spark 3.5/4.0 sources, `NullPropagation` therefore rewrites `count(pos)` to `count(1)`. Even without that rewrite, `Count.updateExpressions` does not read non-null argument values. The outer variant checks only the position's null flag.

This leaves the position values unconsumed despite the helper's stated intent. Spark supplies its existing loop index as the position, while Comet's planner constructs and unnests a separate positions list. A value-dependent sink makes the comparison exercise the output it names. On this dataset, `sum(pos)` should be `9437130` for both variants. Please retain the whole-query/aggregate-cost caveat when rerunning the measurements.

sunchao (commented):
The four original comments are addressed in the revised source by matching the string inputs, using an aggregate sink, excluding the optimizer rule symmetrically, and matching the carried-column scan inputs. One new P2 remains in the position sink, detailed inline.

Current CI reports 7 successful checks and 11 skipped checks, with no failures. The benchmark compile/lint job passed for Spark 4.0 on the merge of this head into the reported base. The main Spark test matrices were skipped.

This follow-up is based on the exact source, maintained Spark 3.5/4.0 definitions, and an independently rerun input-enumeration check. I did not rerun Spark, JNI, or benchmark timings. The reported 1.2–1.8x ratios and executed-plan observations remain author-reported whole-query evidence.

comphead (commented):
Thanks @andygrove can we have a bench for wide and nested row.

Something like 

```
import org.apache.spark.sql.SparkSession

val spark = SparkSession.builder().getOrCreate()

val df = spark.sql("""
  SELECT
    1L AS customer_id,
    'US' AS region,

    named_struct(
      'account',
      named_struct(
        'id', 'account-1',
        'name', 'John',
        'settings',
        named_struct(
          'locale', 'en_US',
          'timezone', 'PST',
          'preferences',
          named_struct(
            'language', 'en',
            'theme', 'dark',
            'notifications',
            named_struct(
              'enabled', true,
              'frequency', 'daily',
              'channels', array('email', 'push'),
              'activity',
              named_struct(
                'last_seen', '2026-09-03',
                'sessions',
                named_struct(
                  'session_id', 'session-1',
                  'device',
                  named_struct(
                    'type', 'mobile',
                    'os', 'iOS',
                    'attributes',
                    map(
                      'version', '18',
                      'model', 'iPhone'
                    )
                  ),
                  'events',
                  map(
                    'web',
                    array(
                      named_struct(
                        'type', 'click',
                        'timestamp', 100L,
                        'payload', map('page', 'home', 'button', 'buy'),
                        'attributes', map('source', 'google')
                      ),
                      named_struct(
                        'type', 'view',
                        'timestamp', 200L,
                        'payload', map('page', 'product'),
                        'attributes', map('source', 'direct')
                      )
                    ),
                    'mobile',
                    array(
                      named_struct(
                        'type', 'open',
                        'timestamp', 300L,
                        'payload', map('screen', 'home'),
                        'attributes', map('source', 'push')
                      ),
                      named_struct(
                        'type', 'purchase',
                        'timestamp', 400L,
                        'payload', map('item', 'book'),
                        'attributes', map('source', 'mobile')
                      )
                    )
                  )
                )
              )
            )
          )
        )
      ),

      'billing',
      named_struct(
        'currency', 'USD',
        'country', 'US',
        'payment_methods', array('card', 'paypal')
      ),

      'addresses',
      array(
        named_struct('city', 'San Jose', 'country', 'US'),
        named_struct('city', 'New York', 'country', 'US')
      ),

      'metadata',
      map(
        'source', 'web',
        'campaign', 'summer'
      )
    ) AS profile
""")

// Map -> Array -> Struct, at ~10 levels of nesting.
val result = df
  .select(
    col("customer_id"),
    col("region"),
    explode(
      col(
        "profile.account.settings.preferences." +
        "notifications.activity.sessions.events"
      )
    ).alias("platform", "events")
  )
  .select(
    col("customer_id"),
    col("region"),
    col("platform"),
    explode(col("events")).alias("event")
  )
  .select(
    col("customer_id"),
    col("region"),
    col("platform"),
    col("event.type").alias("event_type"),
    col("event.timestamp"),
    col("event.payload"),
    col("event.attributes")
  )

result.show(false)
```

but on 100k rows adtaset

spark/src/test/scala/org/apache/spark/sql/benchmark/CometExplodeBenchmark.scala
andygrove: Fixed, and you were right that nothing was reading a position. Both cases now end in `sum(pos)`, which comes out at 9437130 for `posexplode` and `posexplode_outer` alike — the outer variant's extra 52,430 rows carry a null position and `sum` skips them, so the two stay comparable.

I took the "untimed result check" further than the position cases, because the failure mode you found is one the results table can never show: a sink that stops consuming what it names still reports a rate, just a better one. Every case now declares the aggregate row it must produce, and both engines run it once, untimed, against that row before the case is timed. The expected values are derived from the dataset parameters rather than pasted in, so they follow if the row count or fan-out changes.

On the plan check: the harness already refuses to run a case whose Comet plan is not fully native, and all 15 are. I confirmed on the executed plan that `sum(pos)` survives to the operator rather than being folded away — `CometHashAggregate [pos, col], [Partial], [partial_sum(pos), partial_count(col)]` sitting directly on `CometExplode posexplode(arr), [pos, col], [pos, col]`.

I kept the aggregate-cost caveat, and there is a bigger one to record. The numbers moved, and not because of this change. My earlier table was taken on a contended machine, and at `local[1]` contention lands almost entirely on the Spark arm. Re-running four times, the Comet arm reproduces to within a few ms everywhere while the Spark arm swings by up to 2x. To be sure the review fixes were not the cause I re-ran the *unchanged* previous revision in the same environment and got today's numbers, not the old ones. The corrected table is in the description; the short version is that Comet only wins at low fan-out over `array<bigint>`, and is roughly 2x slower on nested `array<struct>` and ~1.4x slower at fan-out 100. I would rather the benchmark say that than flatter us.

andygrove:
Thanks @comphead — added, as a new `Explode - nested input` group over 100K rows.

I had to change one thing from your example, and it turned out to be the interesting part. `events` in your query is a `map<string, array<struct>>`, and `CometExplodeExec` refuses map inputs (#2837), so the first `explode` falls back to Spark's `GenerateExec` and the Comet arm stops measuring Comet at all — it becomes Spark behind a columnar-to-row transition. So the outer container is an `array<struct<platform, entries>>` here instead, which keeps everything else: the event list still sits at `profile.account.settings.preferences.notifications.activity.sessions.events`, each element still holds a second array of four-field structs, and the query still chains two generators and carries `customer_id` and `region` through both. Worth adding the map case as soon as #2837 lands, and I'd expect it to be the worst of the lot.

Three cases: the array at depth 1, the same array eight struct accessors down, and the chained second explode. Depth 1 and depth 8 hold the same array written twice into one file, so the pair isolates what the struct chain costs. The whole event struct is counted rather than one of its fields, otherwise nested schema pruning narrows the exploded element and the case quietly stops measuring a wide row. Pruning does drop the siblings nothing reads — confirmed on the plan, the depth-8 scan projects `profile` as `struct<account:struct<...sessions:struct<events:array<...>>>>` with `billing`, `addresses` and `device` gone.

Results (Apple M3 Max, `local[1]`, Spark 4.1, best ms):

| Case | Spark | Comet | Relative |
| ---- | ----- | ----- | -------- |
| depth 1 | 69 | 156 | 0.4X |
| depth 8 | 84 | 167 | 0.5X |
| depth 8, then inner array | 96 | 181 | 0.5X |

So this shape is a clear loss for us — Comet is about 2x slower than Spark on it, and the reproducible part is the element being a wide struct rather than the nesting itself: depth 8 costs Comet only 11ms over depth 1, where it costs Spark 15ms. The plain `array<struct<bigint, string>>` case at fan-out 10 is 0.7X for the same reason. Good call asking for this; it is the largest gap the benchmark found and I think it is where the operator work should start.

sunchao (approved):
Re-reviewed `f79d0b8e`. The previous position-consumption issue is addressed by `sum(pos)`, and the untimed expected-result checks cover both engine settings. I also checked the new shallow/deep nested-array and chained-generator cases, the carried-column sinks, and the retained distinction between whole-query timings and operator-only cost. No new actionable P1/P2 finding.

Validation here was source review against the maintained Spark branches plus a data-shape check of the 15 expected result rows. I did not execute Spark or reproduce the benchmark timings locally. Current CI metadata shows six successful checks and 11 skipped checks. The author-reported benchmark results remain separate from that validation.

comphead (approved):
Thanks for having benchmark, a good follow up ticket would be perf improvement ticket referring to this bench result

andygrove:
> Thanks for having benchmark, a good follow up ticket would be perf improvement ticket referring to this bench result

Filed https://github.com/apache/datafusion-comet/issues/5685

PR #5392, opened 2026-08-18
fix: Native shuffle fails with a 2GB task serialization OOM on jobs with many partitions

### LINKED ISSUE #5391: Native shuffle fails with a 2GB task serialization OOM on jobs with very many partitions
### Describe the bug

A job with a huge number of scan partitions fails at stage submission:

  ```
  Task serialization failed: java.lang.OutOfMemoryError: Required array length 2147483639 + 794 is too large
      at java.io.ByteArrayOutputStream.ensureCapacity(...)
      at org.apache.spark.serializer.JavaSerializerInstance.serialize(...)
      at org.apache.spark.scheduler.DAGScheduler.submitMissingTasks(DAGScheduler.scala:1528)
  ```

  A single `byte[]` can't exceed ~2GB, and the driver is serializing one object bigger than that. The failing job had ~38.7M partitions (multi-petabyte dataset), native scan into a native shuffle. The same job runs fine on plain Spark, so it's something Comet does differently, not just scale.

  ### Analysis

  `DAGScheduler.submitMissingTasks` serializes the `(RDD, ShuffleDependency)` pair once into a single broadcast byte array, which must fit in ~2GB. On the native shuffle path, `CometShuffleDependency.nativeShuffleSpec` is a non-transient field holding a `NativeExecContext` whose `perPartitionByKey: Map[String, Array[Array[Byte]]]` carries one
  serialized scan plan-data blob (the partition's file list) per map partition. With ~38.7M partitions that's ~38.7M protobufs baked into the broadcast task binary, so it exceeds 2GB — even though each write task only ever reads its own partition's slice.

  Spark's `FileScanRDD` keeps per-partition file lists `@transient`; each partition's files ride in its own `Partition` object and are serialized separately per task, so the full list  is never included in the shared broadcast blob.

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

parthchandra:

  Closes #5391.

  ## Rationale for this change
  
  A native-shuffle job with a very large number of partitions fails at stage submission, before any task runs:
 
(From the issue:
  Task serialization failed: java.lang.OutOfMemoryError: Required array length 2147483639 + 794 is too large
      at java.io.ByteArrayOutputStream.ensureCapacity(...)
      at org.apache.spark.scheduler.DAGScheduler.submitMissingTasks(...)

  `DAGScheduler` serializes the `(RDD, ShuffleDependency)` pair into a single broadcast byte array that must fit in ~2GB. On the native-shuffle path, `CometShuffleDependency.nativeShuffleSpec` is a non-transient field holding a `perPartitionByKey` map with one scan-plan-data blob (the partition's file list) per map partition. The failing job had ~38.7M
  partitions, so ~38.7M protobufs got baked into that one blob and blew the limit — even though each task only reads its own slice. Plain Spark avoids this by keeping per-partition file lists `@transient` and shipping them per task; Comet's own `CometExecRDD` does the same. The native shuffle path was the exception.

  Mirror `CometExecRDD`: keep the map off the serialized dependency and give each task only its slice.

  - `CometNativeShuffleInputRDD` takes `perPartitionByKey` as a `@transient` arg and, in `getPartitions` (driver-side), slices out each partition's entry onto its `Partition` object. That slice flows to the writer, which injects it instead of indexing the full map.
  - `NativeExecContext.perPartitionByKey` is now `@transient` so no build path can serialize the full map, and we also empty it when building the dependency.
  - `commonByKey` is unchanged — it's sized by scan count, not partition count, and the writer still needs it.

  - A unit test serializes the input RDD at 10 vs 10,000 partitions and checks the size stays roughly constant, and that each partition carries only its own slice.
  - An end-to-end test runs a native shuffle over a multi-partition native scan and compares results against Spark with an all-native plan, so a wrong slice would fail it.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDDSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDDSuite.scala
@@ -19,48 +19,83 @@
 
 package org.apache.spark.sql.comet.execution.shuffle
 
+import org.apache.spark.HashPartitioner
 import org.apache.spark.serializer.JavaSerializer
 import org.apache.spark.sql.CometTestBase
+import org.apache.spark.sql.comet.{CometMetricNode, NativeExecContext}
+import org.apache.spark.sql.execution.metric.SQLMetrics
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+import org.apache.comet.serde.OperatorOuterClass.Operator
 
 /**
- * Ensure that serialized RDD does not overflow in size with a very large number of partitions
+ * Ensure the serialized shuffle-map-stage task binary does not grow with the number of
+ * partitions.
+ *
+ * `DAGScheduler.submitMissingTasks` broadcasts the serialized `(stage.rdd, stage.shuffleDep)`
+ * pair, which must fit in ~2GB. On the native-shuffle path the scan plan data is one serialized
+ * blob per map partition; the leak lived under
+ * `CometShuffleDependency.nativeShuffleSpec.execContext`, not on the thin RDD. So this builds
+ * both carriers and serializes the pair, which catches a regression in either the RDD or the
+ * dependency guards without needing a real 2GB allocation.
+ *
  * Lives in the `execution.shuffle` package so it can construct the `private[shuffle]`
- * [[CometNativeShuffleInputRDD]] directly.
+ * [[CometNativeShuffleInputRDD]] and the `private[comet]` [[NativeExecContext]] directly.
  */
 class CometNativeShuffleInputRDDSuite extends CometTestBase {
 
-  test("serialized RDD size is independent of partition count") {
+  test("serialized (rdd, dep) task binary size is independent of partition count") {
     val sc = spark.sparkContext
     val ser = new JavaSerializer(sc.getConf).newInstance()
 
-    // One scan key with a 1KB blob per map partition. Pre-fix the whole array serializes into the
-    // RDD, so 10000 partitions add ~10MB versus 10.
-    def buildRDD(numPartitions: Int): CometNativeShuffleInputRDD = {
+    // One scan key with a 1KB blob per map partition -- the per-partition plan data. Build both
+    // objects the DAGScheduler serializes for a shuffle-map stage: the thin RDD, and the
+    // CometShuffleDependency whose NativeShuffleSpec holds a NativeExecContext with this map.
+    def build(numPartitions: Int): (
+        CometNativeShuffleInputRDD,
+        CometShuffleDependency[Int, ColumnarBatch, ColumnarBatch]) = {
       val perPartitionByKey =
         Map("scan-0" -> Array.fill(numPartitions)(new Array[Byte](1024)))
-      new CometNativeShuffleInputRDD(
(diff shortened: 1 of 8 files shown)


spark/src/test/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDDSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+        sc,
+        inputRDDs = Seq.empty,
+        numPartitionsParam = numPartitions,
+        shuffleScanIndices = Set.empty,
+        perPartitionByKey = perPartitionByKey)
+    }
+
+    val smallSize = ser.serialize(buildRDD(10)).limit()
+    val large = buildRDD(10000)
+    val largeSize = ser.serialize(large).limit()
```
sunchao: [P2] Exercise the actual shuffle task binary

Could this also construct a native shuffle dependency and compare the serialized size of `(rdd, shuffleDependency)` for 10 and 10,000 partitions? That is what Spark broadcasts for a shuffle-map stage. Before this fix, the large map lived under `CometShuffleDependency.nativeShuffleSpec.execContext`, not on the thin RDD. In an isolated Spark serialization probe, restoring that dependency leak left the RDD at 880 bytes while the pair grew from 13,322 to 10,342,982 bytes. The current RDD-only assertion therefore stays green if those guards regress. Testing the actual pair would protect the reported 2GB failure without requiring a 2GB allocation.

sunchao (dismissed):
Reviewed `1b12802e123c020e4fd9bbfbce091bae818dee0b` in five independent scopes. The production change looks correct. Driver-side slicing preserves the old per-partition lookup, the shared scan data remains available to the writer, and the original context is still available for range sampling. I found no introduced execution defect.

An isolated probe using Spark 3.5.9's real Java serializer and RDD/dependency classes confirmed that the parent task-binary pair grows from 13,322 to 10,342,982 bytes when increasing from 10 to 10,000 partitions. The PR's pair stays at 2,922 bytes, and a separately serialized partition retains the correct scan-data slice. This probe stubs unrelated native/protobuf/metrics types and does not execute native SQL. Exact-head CI has 61 passing checks and 9 skipped. I also confirmed that both new tests ran in the Linux Spark 3.4 and macOS Spark 4.0 shuffle jobs.

Approving the implementation. I left one P2 comment asking for the size regression to cover the actual `(RDD, ShuffleDependency)` pair so it protects the original failure path.

comphead (approved):
Thanks @parthchandra lgtm, however I never hit that error when was testing on workloads with 20k+ partitions

spark/src/test/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDDSuite.scala
parthchandra: Good suggestion; added ! This will now fail if someone removes the `@transient` on `NativeExecContext.perPartitionByKey`.

parthchandra:
> Thanks @parthchandra lgtm, however I never hit that error when was testing on workloads with 20k+ partitions

@comphead this happens at petabyte scale. The failure occurred at 38M partitions.

parthchandra:
Merged. Thank you @sunchao, @comphead !

Issue #5207, opened 2026-08-01
Do not accelerate a query stage that contains a columnar-to-row fallback (enable whole-stage revert by default)


andygrove:
## What is the problem the feature request solves?

When someone tries Comet for the first time on a query where Comet accelerates the scan and then immediately falls back to Spark row-based execution, the result is often *slower* than plain Spark: the columnar-to-row (C2R) conversion has to be paid, and the surrounding Spark operators lose whole-stage codegen because a Comet operator sits underneath them. This is a bad first impression, and the user has no obvious signal that anything is wrong — the plan looks like Comet "worked".

Comet already has a whole-stage fallback for this, but it is disabled by default and, as currently written, cannot be enabled at the strictest setting.

### What exists today

`RevertNativeForTransitionHeavyStages` (`spark/src/main/scala/org/apache/comet/rules/RevertNativeForTransitionHeavyStages.scala`), registered in `postColumnarTransitions` ahead of `EliminateRedundantTransitions`, counts `ColumnarToRowTransition` nodes within a stage (stopping at `QueryStageExec` / shuffle / broadcast boundaries) and, above a threshold, rewrites every `CometExec` back to `cometExec.originalPlan`, strips the Comet transitions, re-inserts Spark's, and wraps in `RowToColumnarExec` when the parent exchange needs columnar input. AQE applies it per query stage; non-AQE applies it via `transformUp` over shuffles plus the result stage.

- `spark.comet.exec.transitionRevert.enabled` — default `false`
- `spark.comet.exec.transitionRevert.maxTransitions` — default `2`

Adjacent but much narrower: `spark.comet.shuffle.revertRedundantColumnar.enabled` (default `true`) only unwinds a Comet columnar shuffle sandwiched between two non-Comet hash aggregates.

### 1. Exempt output-boundary C2R transitions from the count

`countTransitions` currently counts the trailing C2R at the plan root, so a *fully* accelerated query such as `CometNativeScan -> CometProject -> ColumnarToRowExec(root)` counts as 1 transition. Setting `maxTransitions=0` today would therefore revert every query, including ones with no mid-stage fallback at all.

A C2R at the output boundary is not a fallback — it is the unavoidable handoff to the consumer. `countTransitions` should skip any C2R reachable from the stage root through only output-only nodes:

- the stage root itself
- `DataWritingCommandExec`
- `WriteFilesExec`
- `V2CommandExec` (covers append / overwrite / CTAS)
- `ExecutedCommandExec`

[...]

PR #5394, opened 2026-08-19
feat: add spark.comet.explain.planOnly.enabled

### LINKED ISSUE #5335: Add mode to run Comet planning but execute with Spark, so users can assess potential Comet coverage

andygrove:

Closes #5335. Alternate approach to #5345, based on review discussion there.

Heads up: I used an LLM to help draft this. The design is mine, but the code and prose have been shaped with LLM assistance, so review with that in mind.

## Rationale for this change

Users evaluating Comet on a workload need a way to estimate how much of it Comet would accelerate without actually changing execution. Turning Comet on and comparing runs carries real risk. This adds a mode that builds the Comet plan Comet would have executed, logs it, and lets Spark run the query unchanged.

- New config `spark.comet.explain.planOnly.enabled`, default off.
- When set, `CometScanRule` and `CometExecRule` short-circuit at the top of their `apply` and return the plan untouched.
- `CometExecRule` builds a preview by calling both rules' private `_apply` methods directly (with an explicit `forPreview = true` parameter that skips the short-circuit on the recursive call), logs the resulting Comet plan at WARN level, then returns the original Spark plan.
- The plan-only branch sits at the top of the exec-enabled block so `normalizePlan`/`RewriteJoin`/`tagUnsafePartialAggregates` are not run and thrown away.
- The WARN is deduped per SQL execution ID (bounded LRU) so under AQE a query gets one report, not one per stage.
- Doc section in `understanding-comet-plans.md`.

The estimate is Scala-side only. The native plan is never handed to DataFusion, so anything that would have failed inside DataFusion still counts as accelerated. That is called out in the config docstring and the user guide.

New tests in `CometExecRuleSuite` cover:

- V1 and V2 Parquet scans, each with AQE on and off.
- Scalar subquery, so the recursive preview builds through the subquery plan without leaving any Comet operators behind in the executed plan.
- Same query with the config off, asserting Comet operators do appear (sanity check that we haven't accidentally disabled Comet).

Each test asserts the executed plan has zero `CometPlan` operators.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
@@ -115,6 +115,47 @@ object CometExecRule {
    */
   val SKIP_COMET_BROADCAST_TAG: org.apache.spark.sql.catalyst.trees.TreeNodeTag[Unit] =
     org.apache.spark.sql.catalyst.trees.TreeNodeTag[Unit]("comet.skipCometBroadcast")
+
+  /**
+   * Bounded set of SQL execution IDs for which plan-only mode has already logged its report. Used
+   * to dedupe the WARN under AQE, where `CometExecRule` fires once against the full initial plan
+   * and again per stage.
+   */
+  private val PLAN_ONLY_REPORTED_LIMIT = 1024
+  private val planOnlyReportedIds: java.util.LinkedHashSet[String] =
+    new java.util.LinkedHashSet[String]()
+
+  /**
+   * Thread-local re-entry guard: set to true while `reportPlanOnlyCoverage` is running the rule
+   * recursively to build the preview plan. Prevents the outer plan-only branch in `_apply` from
+   * firing on the nested call, so the recursive invocation executes the normal transform.
+   */
+  private[rules] val planOnlyPreviewInProgress: ThreadLocal[Boolean] =
+    new ThreadLocal[Boolean] {
+      override def initialValue(): Boolean = false
+    }
+
+  private[comet] def markPlanOnlyReported(executionId: Option[String]): Boolean = {
+    executionId match {
+      case None => true
+      case Some(id) =>
+        planOnlyReportedIds.synchronized {
+          val added = planOnlyReportedIds.add(id)
+          if (planOnlyReportedIds.size > PLAN_ONLY_REPORTED_LIMIT) {
+            val it = planOnlyReportedIds.iterator()
+            it.next()
+            it.remove()
+          }
+          added
+        }
+    }
+  }
+
+  private[comet] def clearPlanOnlyReported(): Unit = {
+    planOnlyReportedIds.synchronized {
+      planOnlyReportedIds.clear()
+    }
+  }
 }
 
 /**
@@ -573,6 +614,26 @@ case class CometExecRule(session: SparkSession)
     newPlan
   }
 
+  /**
+   * Build the Comet plan we would have executed and log it. Called from `_apply` in plan-only
+   * mode; the built plan is discarded. `CometScanRule` and this rule both skip work when
+   * `spark.comet.explain.planOnly.enabled` is set, so we use their bypass flags to force normal
+   * behavior while we compute the preview.
+   */
+  private def reportPlanOnlyCoverage(plan: SparkPlan): Unit = {
+    val preview = CometScanRule.withForceApply {
+      CometExecRule.planOnlyPreviewInProgress.set(true)
+      try {
+        _apply(CometScanRule(session).apply(plan))
+      } finally {
+        CometExecRule.planOnlyPreviewInProgress.set(false)
+      }
+    }
+    logWarning(
+      s"[Comet plan-only]\n" +
+        new ExtendedExplainInfo().generateExtendedInfo(preview))
+  }
+
   private def _apply(plan: SparkPlan): SparkPlan = {
     // We shouldn't transform Spark query plan if Comet is not loaded.
     if (!isCometLoaded(conf)) return plan
@@ -604,6 +665,19 @@ case class CometExecRule(session: SparkSession)
       // formats are incompatible. This runs before transform() so the tags are checked
       // during the bottom-up conversion. Tags persist through AQE stage creation.
       tagUnsafePartialAggregates(planWithJoinRewritten)
+
+      // Plan-only mode: `CometScanRule` skipped work, so `plan` here is still pure Spark. Build
+      // a preview by rerunning both rules against a scan-wrapped copy and log the coverage
+      // report, then return `plan` unchanged so nothing is offloaded to native.
+      if (CometConf.COMET_EXPLAIN_PLAN_ONLY_ENABLED.get() &&
+        !CometExecRule.planOnlyPreviewInProgress.get()) {
+        val executionId = Option(
+          session.sparkContext.getLocalProperty(SQLExecution.EXECUTION_ID_KEY))
+        if (CometExecRule.markPlanOnlyReported(executionId)) {
+          reportPlanOnlyCoverage(plan)
+        }
+        return plan
+      }
 
       var newPlan = transform(planWithJoinRewritten)
 
--- a/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
@@ -131,21 +131,6 @@ object CometExecRule {
           size() > PLAN_ONLY_REPORTED_LIMIT
       })
 
-  /**
-   * Thread-local re-entry guard: set while `reportPlanOnlyCoverage` is building the preview so
-   * both `CometScanRule` and `CometExecRule` run their normal transforms on the nested pass
-   * instead of short-circuiting.
-   */
-  private[rules] val planOnlyPreviewInProgress: ThreadLocal[Boolean] =
-    ThreadLocal.withInitial(() => java.lang.Boolean.FALSE)
-
-  private[rules] def withPreview[T](f: => T): T = {
-    val prev = planOnlyPreviewInProgress.get()
-    planOnlyPreviewInProgress.set(true)
-    try f
-    finally planOnlyPreviewInProgress.set(prev)
-  }
-
   private[comet] def markPlanOnlyReported(executionId: Option[String]): Boolean = {
     executionId match {
       case None => true
@@ -600,7 +585,7 @@ case class CometExecRule(session: SparkSession)
   }
 
   override def apply(plan: SparkPlan): SparkPlan = {
-    val newPlan = _apply(plan)
+    val newPlan = _apply(plan, forPreview = false)
     if (showTransformations && !newPlan.fastEquals(plan)) {
       logInfo(s"""
            |=== Applying Rule $ruleName ===
@@ -612,17 +597,15 @@ case class CometExecRule(session: SparkSession)
 
   /**
    * Build the Comet plan we would have executed and log it. Called from `_apply` in plan-only
-   * mode; the built plan is discarded. Both rules short-circuit under plan-only mode, so we set
-   * `planOnlyPreviewInProgress` to force the nested calls to run their normal transforms.
+   * mode; the built plan is discarded. Passes `forPreview = true` through the nested calls so
+   * both rules run their normal transforms instead of short-circuiting.
    */
   private def reportPlanOnlyCoverage(plan: SparkPlan): Unit = {
-    val preview = CometExecRule.withPreview {
-      _apply(CometScanRule(session).apply(plan))
-    }
+    val preview = _apply(CometScanRule(session)._apply(plan), forPreview = true)
     logWarning(s"[Comet plan-only]\n${new ExtendedExplainInfo().generateExtendedInfo(preview)}")
   }
 
-  private def _apply(plan: SparkPlan): SparkPlan = {
+  private def _apply(plan: SparkPlan, forPreview: Boolean): SparkPlan = {
     // We shouldn't transform Spark query plan if Comet is not loaded.
     if (!isCometLoaded(conf)) return plan
 
@@ -643,8 +626,7 @@ case class CometExecRule(session: SparkSession)
       // `plan` is still pure Spark; `reportPlanOnlyCoverage` rebuilds a scan-wrapped copy for
       // the preview. Placed before `normalizePlan`/`RewriteJoin`/`tagUnsafePartialAggregates`
       // so their work is not wasted on the discarded outer pass.
-      if (CometConf.COMET_EXPLAIN_PLAN_ONLY_ENABLED.get() &&
-        !CometExecRule.planOnlyPreviewInProgress.get()) {
+      if (!forPreview && CometConf.COMET_EXPLAIN_PLAN_ONLY_ENABLED.get()) {
         val executionId = Option(
           session.sparkContext.getLocalProperty(SQLExecution.EXECUTION_ID_KEY))
         if (CometExecRule.markPlanOnlyReported(executionId)) {
--- a/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
@@ -32,7 +32,7 @@ import org.apache.spark.sql.comet._
 import org.apache.spark.sql.comet.execution.shuffle.{CometColumnarShuffle, CometNativeShuffle, CometShuffleExchangeExec}
 import org.apache.spark.sql.comet.util.Utils
 import org.apache.spark.sql.execution._
-import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanExec, AQEShuffleReadExec, BroadcastQueryStageExec, ShuffleQueryStageExec}
+import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanExec, AQEShuffleReadExec, BroadcastQueryStageExec, QueryStageExec, ShuffleQueryStageExec}
 import org.apache.spark.sql.execution.aggregate.{BaseAggregateExec, HashAggregateExec, ObjectHashAggregateExec}
 import org.apache.spark.sql.execution.command.{DataWritingCommandExec, ExecutedCommandExec}
 import org.apache.spark.sql.execution.datasources.WriteFilesExec
@@ -117,32 +117,89 @@ object CometExecRule {
     org.apache.spark.sql.catalyst.trees.TreeNodeTag[Unit]("comet.skipCometBroadcast")
 
   /**
-   * SQL execution IDs for which plan-only mode has already logged its report, as a bounded LRU.
-   * Dedupes the WARN under AQE, where `CometExecRule` fires once against the full initial plan
-   * and again per stage. Same synchronized-`LinkedHashMap` LRU pattern used by
-   * `IcebergPlanDataInjector.commonCache`.
+   * A bounded set of keys, used for plan-only reporting state. Evicts in LRU order once `limit`
+   * keys are held, so a long-lived driver retains a fixed amount of reporting state. Same
+   * synchronized-`LinkedHashMap` pattern used by `IcebergPlanDataInjector.commonCache`.
    */
-  private val PLAN_ONLY_REPORTED_LIMIT = 1024
-  private val planOnlyReportedIds: java.util.Map[String, java.lang.Boolean] =
-    java.util.Collections.synchronizedMap(
-      new java.util.LinkedHashMap[String, java.lang.Boolean](16, 0.75f, false) {
-        override def removeEldestEntry(
-            eldest: java.util.Map.Entry[String, java.lang.Boolean]): Boolean =
-          size() > PLAN_ONLY_REPORTED_LIMIT
-      })
+  private class BoundedKeySet(limit: Int) {
+    private val keys: java.util.Map[String, java.lang.Boolean] =
+      java.util.Collections.synchronizedMap(
+        new java.util.LinkedHashMap[String, java.lang.Boolean](16, 0.75f, true) {
+          override def removeEldestEntry(
+              eldest: java.util.Map.Entry[String, java.lang.Boolean]): Boolean = size() > limit
+        })
 
-  private[comet] def markPlanOnlyReported(executionId: Option[String]): Boolean = {
+    /** Adds `key`, returning true if it was not already present. */
+    def add(key: String): Boolean = keys.put(key, java.lang.Boolean.TRUE) == null
+
+    def contains(key: String): Boolean = keys.containsKey(key)
+  }
+
+  private val PLAN_ONLY_REPORTED_LIMIT = 1024
+
+  /** `executionId:planFingerprint` keys that plan-only mode has already reported. */
+  private val planOnlyReportedPlans = new BoundedKeySet(PLAN_ONLY_REPORTED_LIMIT)
+
+  /** Execution IDs whose plan-only report came from the query-stage-prep rule. */
+  private val planOnlyPrepReportedIds = new BoundedKeySet(PLAN_ONLY_REPORTED_LIMIT)
+
+  /**
+   * Whether plan-only mode should report `plan`, recording that it did so.
+   *
+   * Spark applies this rule many times during one SQL execution, and only some of those
+   * applications correspond to a plan the user is asking about:
+   *
+   *   - Each scalar subquery and DPP subquery is prepared as its own top-level plan, and that
+   *     happens *before* the outer plan reaches the conversion rules. Keying the report on the
+   *     execution ID alone therefore let a nested subquery consume the slot and suppressed the
+   *     outer plan, which is the plan being evaluated. Keying on the execution ID *and* the plan
+   *     gives the outer plan its own report and each separately prepared subquery theirs.
+   *   - Under AQE the rule also runs once per query stage (as a columnar rule) and again on every
+   *     re-optimization (as a query-stage-prep rule). Those are re-planning of a plan already
+   *     reported, so a `plan` containing query stages is skipped, and once the query-stage-prep
+   *     rule has reported an execution the columnar applications for it stay quiet.
+   *
+   * @param queryStagePrep
+   *   whether the calling rule instance is registered as a query-stage-prep rule.
+   */
+  private[comet] def shouldReportPlanOnly(
+      executionId: Option[String],
+      plan: SparkPlan,
+      queryStagePrep: Boolean): Boolean = {
     executionId match {
-      case None => true
-      case Some(id) => planOnlyReportedIds.put(id, java.lang.Boolean.TRUE) == null
+      case None =>
+        // No execution ID means the plan is being built outside an action (`df.explain`, or
+        // reading `queryExecution.executedPlan` directly). AQE creates no stages in that case, so
+        // there is nothing to dedupe against.
+        true
+      case Some(id) =>
+        if (plan.exists(_.isInstanceOf[QueryStageExec])) {
+          // An AQE stage plan or a re-optimized plan: a re-plan of what we already reported.
+          false
+        } else {
+          if (queryStagePrep) {
+            planOnlyPrepReportedIds.add(id)
+          } else if (planOnlyPrepReportedIds.contains(id)) {
+            return false
+          }
+          // The plan's structural hash identifies it: node tags are not part of it, so the same
+          // plan applied twice keys the same and is reported once.
+          planOnlyReportedPlans.add(s"$id:${plan.hashCode()}")
+        }
     }
   }
 }
 
 /**
  * Spark physical optimizer rule for replacing Spark operators with Comet operators.
+ *
+ * @param queryStagePrep
+ *   true for the instance registered with `injectQueryStagePrepRule`, which under AQE sees the
+ *   whole initial plan, and false for the one registered as a columnar rule, which under AQE sees
+ *   one query stage at a time. Only plan-only reporting reads this; see
+ *   [[CometExecRule.shouldReportPlanOnly]].
  */
-case class CometExecRule(session: SparkSession)
+case class CometExecRule(session: SparkSession, queryStagePrep: Boolean = false)
     extends Rule[SparkPlan]
     with ShimSubqueryBroadcast {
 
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -589,6 +621,20 @@ case class CometExecRule(session: SparkSession)
...
     } else {
+      // Plan-only mode: build the Comet plan Comet would have executed, log it, and return
+      // the original plan unchanged. `CometScanRule` also short-circuits in this mode, so
+      // `plan` is still pure Spark; `reportPlanOnlyCoverage` rebuilds a scan-wrapped copy for
+      // the preview. Placed before `normalizePlan`/`RewriteJoin`/`tagUnsafePartialAggregates`
+      // so their work is not wasted on the discarded outer pass.
+      if (!forPreview && CometConf.COMET_EXPLAIN_PLAN_ONLY_ENABLED.get()) {
+        val executionId = Option(
+          session.sparkContext.getLocalProperty(SQLExecution.EXECUTION_ID_KEY))
+        if (CometExecRule.markPlanOnlyReported(executionId)) {
```
sunchao: [P2] Keep the report slot for the outer query

Could we avoid consuming the execution-ID entry while Spark is preparing a nested subquery? For a fresh action on `SELECT id FROM range(10) WHERE id > (SELECT max(id) FROM range(3))`, Spark prepares the scalar subquery before the outer plan under the same execution ID. The subquery records the ID here, so the later outer-query call is suppressed. I reproduced this planning order with the same deduplication logic on Spark 3.5.2 with AQE both disabled and enabled. The only report describes `max(id)`, not the outer scan/filter or the workload being evaluated. Please distinguish root-query reporting from subquery/stage preparation and add an action-based test that captures the warning and checks that the outer plan appears.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -573,7 +595,17 @@ case class CometExecRule(session: SparkSession)
...
   }
 
-  private def _apply(plan: SparkPlan): SparkPlan = {
+  /**
+   * Build the Comet plan we would have executed and log it. Called from `_apply` in plan-only
+   * mode; the built plan is discarded. Passes `forPreview = true` through the nested calls so
+   * both rules run their normal transforms instead of short-circuiting.
+   */
+  private def reportPlanOnlyCoverage(plan: SparkPlan): Unit = {
+    val preview = _apply(CometScanRule(session)._apply(plan), forPreview = true)
```
sunchao: [P2] Include post-columnar reversion in the coverage preview

Could the preview run transition insertion and Comet's post-columnar rules before computing coverage? Normal planning subsequently runs `RevertNativeForTransitionHeavyStages` and `EliminateRedundantTransitions`, but this path stops before them. With `spark.comet.exec.transitionRevert.enabled=true`, `spark.comet.exec.transitionRevert.maxTransitions=0`, and Comet project execution disabled, the [existing regression case](https://github.com/apache/datafusion-comet/blob/eb514de1ff76785b81a4940d6a2adb3ea149ab07/spark/src/test/scala/org/apache/comet/rules/RevertNativeForTransitionHeavyStagesSuite.scala#L196-L220) shows that the real executed plan has zero `CometExec` nodes. This preview still counts the operators that the configured Scala-side rule removes, and its transition count is calculated before Spark inserts those transitions. That is separate from the documented uncertainty about DataFusion planning failures. Please compare the report with the real post-columnar plan for this configuration.

sunchao (commented):
## Summary

Reviewed `eb514de1ff76785b81a4940d6a2adb3ea149ab07` against `a28ac348f5a68500d96fe872b49e453d96e4bab6` across five independent scopes. Reusing the existing conversion rules is a sensible way to keep eligibility checks aligned with normal Comet planning. I found two P2 issues that affect the accuracy and completeness of the report. Both are described inline.

## Prior state and problem

The existing explain output describes a plan that has already been converted for Comet execution. Evaluating an unfamiliar workload therefore requires enabling Comet on the execution path. This change adds an opt-in way to inspect potential acceleration while retaining Spark execution.

## Design approach

The new configuration defaults to false. The public scan rule leaves the incoming plan alone, while the execution rule builds a disposable preview by invoking scan conversion and execution conversion directly. The explicit `forPreview` argument prevents that preview from taking the plan-only shortcut again.

I did not find a verified native-execution leak in the ordinary read path. The main concerns are reporting correctness: nested subquery planning can claim the execution-ID report slot before the outer query, and the preview stops before Scala-side transition insertion and transition-driven reversion.

A local Spark 3.5.2 planning-order probe using the same deduplication logic reproduced the first issue with AQE both off and on. In both cases the scalar aggregate was reported first and the outer filter was suppressed. The second issue is supported by the exact-head rule ordering and the existing `RevertNativeForTransitionHeavyStagesSuite` regression case. I did not run the full Comet JVM/native suite. `git diff --check` passed. At the final CI check, 12 checks had succeeded, 5 were running, 1 was queued, 7 were skipped, and none had failed.

## Key design decisions

Keeping the feature disabled by default limits ordinary behavior changes. Returning the original plan avoids reconstructing a Spark plan from converted operators. A bounded execution-ID cache also limits retained reporting state, but the report owner must distinguish the root query from recursively prepared subqueries.

## Implementation sketch

The implementation adds the configuration and user guide, exposes scan conversion within the rules package, adds the preview/report branch to `CometExecRule`, and tests V1/V2 scans with AQE on and off plus a scalar subquery and a config-off sanity check. Those tests check plan isolation, but they do not inspect the warning emitted by an actual query action.

Plan-only mode requires Comet execution to be enabled, emits a driver warning, and deliberately does not call DataFusion's native planner. That documented native-planning limitation is reasonable. It should remain distinct from omitting a known Scala-side reversion or reporting only an inner subquery.

## Suggested improvements

Please make the root query own the once-per-execution report, and calculate coverage after the applicable columnar-transition and Comet post-columnar rules. Add action-based log assertions for scalar subqueries under both AQE modes and compare the preview with normal execution when transition reversion is enabled.

andygrove:
Both fixed in 2f94707. On the first one — I couldn't find a way to identify the root plan at rule-application time. Subqueries are prepared before the outer plan with nothing to distinguish them, and matching the incoming plan against the root QueryExecution's output breaks down for commands and write plans, which would then get no report at all. So instead of one report owned by the root, each independently planned plan gets one: the outer query plus one per separately prepared subquery. Dedupe is now keyed on execution ID and plan, and AQE's per-stage and post-re-optimization applications are skipped as re-plans of something already reported, so the stage spam the execution-ID cache was there to prevent is still gone. Does that seem like a reasonable trade to you, or would you rather see the root identified even if some plan shapes drop out of reporting?

The second one was as you described. The preview now inserts transitions and runs both post-columnar rules. RevertNativeForTransitionHeavyStages needed a new entry point for this: its `apply` takes the AQE branch when AQE is on, which only judges the topmost stage, whereas the preview holds a plan that hasn't been split into stages, so it needs every shuffle boundary visited. The new test builds your regression configuration, reads the coverage out of the captured report, and compares it against `CometCoverageStats` for the plan Comet really executes — before the fix the report claimed 4/4 with no transitions against a real 3/4 with one.

What the preview still can't match is AQE re-planning: it describes the pre-adaptive plan and applies the post-columnar rules to it in one pass rather than per stage. That's called out in the user guide alongside the native-planning caveat.

spark/src/test/scala/org/apache/comet/rules/CometExecRuleSuite.scala
```diff
@@ -797,4 +799,164 @@ class CometExecRuleSuite extends CometTestBase {
...
+
+        withSQLConf(
+          CometConf.COMET_EXEC_TRANSITION_REVERT_ENABLED.key -> "true",
+          CometConf.COMET_EXEC_TRANSITION_REVERT_MAX_TRANSITIONS.key -> "0") {
+          // What Comet really executes with reversion enabled.
+          val df = sql(query)
+          df.collect()
+          val executed = CometCoverageStats.forPlan(df.queryExecution.executedPlan)
+
+          val reports = withSQLConf(CometConf.COMET_EXPLAIN_PLAN_ONLY_ENABLED.key -> "true") {
```
sunchao: [P1] Preserve the captured report across Spark 3.x withSQLConf

Could the `withSQLConf` block be moved inside `capturePlanOnlyReports`, or could these assertions run inside the configuration block? Spark 3.4 and 3.5 define `withSQLConf(...)(f: => Unit): Unit`, unlike the generic Spark 4.x helper, so `reports` is inferred as `Unit` here. The exact-head [Spark 3.4](https://github.com/apache/datafusion-comet/actions/runs/32280513769/job/96158604209) and [Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/32280513769/job/96158604194) checks both fail test compilation on the subsequent `size`, `mkString`, and `head` calls. This prevents the supported Spark 3.x test builds from compiling.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -115,12 +115,91 @@ object CometExecRule {
...
+   *
+   * @param queryStagePrep
+   *   whether the calling rule instance is registered as a query-stage-prep rule.
+   */
+  private[comet] def shouldReportPlanOnly(
+      executionId: Option[String],
+      plan: SparkPlan,
+      queryStagePrep: Boolean): Boolean = {
+    executionId match {
+      case None =>
```
sunchao: [P2] Deduplicate adaptive reports when the execution ID is absent

Could the no-ID path retain plan-scoped reporting state instead of returning `true` for every invocation? The public `df.rdd.count()` path can build and execute AQE stages without installing `spark.sql.execution.id`. A Spark 3.5.2 probe using this exact decision logic and both rule registrations produced five report decisions for `SELECT id % 2 AS k, count(*) AS n FROM range(20) GROUP BY id % 2`: initial preparation, the adaptive wrapper, the exchange stage, adaptive re-optimization, and the final stage. A fresh `collect()` produced one. Because this branch bypasses both the stage check and deduplication, plan-only mode rebuilds previews and emits overlapping coverage summaries for those ordinary RDD-backed workloads, contrary to the documented suppression of stage/re-optimization reports. Please cover `df.rdd.count()` and planning via `executedPlan` before an action in the reporting tests.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -573,7 +652,31 @@ case class CometExecRule(session: SparkSession)
...
+   * `CometSparkSessionExtensions.CometExecColumnar.postColumnarTransitions`), which can revert
+   * whole stages back to Spark and drop redundant transitions. Those steps run here too, so the
+   * report describes the plan that would really have executed and counts the transitions that
+   * would really have been there, rather than the pre-transition conversion result.
+   *
+   * `RevertNativeForTransitionHeavyStages` is applied with `applyToAllStages` because the preview
+   * holds the whole plan at once, whereas under AQE Spark hands that rule one stage at a time.
+   */
+  private def reportPlanOnlyCoverage(plan: SparkPlan): Unit = {
+    val converted = _apply(CometScanRule(session)._apply(plan), forPreview = true)
```
sunchao: [P2] Include converted scalar subqueries in the outer coverage report

Even with AQE disabled, an acceleratable scalar subquery is counted as Spark in the outer report. Its independently converted preview has already been discarded, and these two conversion passes walk ordinary plan children, leaving the original `ScalarSubquery.plan` in the outer preview. `ExtendedExplainInfo` then traverses those expression-owned plans and includes their operators in the percentage. For the new scalar-subquery test query, the separate subquery warning can therefore report acceleration that is missing from the outer query's coverage. A constructed-plan probe using the exact-head formatter/serializer and real Comet project nodes reports 1/5 with the untouched subquery versus 2/5 after replacing only its plan. Please carry the converted subquery previews into the outer preview, or exclude separately reported subqueries from that report's counts, and compare its coverage with normal Comet planning.

coderfender:
Seems like there are CI failures :)

coderfender (approved):
LGTM  pending CI

andygrove:
All three fixed in 1683a65.

The Spark 3.x compile break was exactly as you described — the assertions now run inside the `withSQLConf` block rather than binding its result. `-Pspark-3.4` and `-Pspark-3.5` both `test-compile` clean here, and `CometExecRuleSuite` passes on 3.5 as well as on the default 4.1.

On the no-execution-ID path: rather than scoping state to an ID that may not exist, the report is now marked on the plan itself. Catalyst copies tags onto replacement nodes, so the mark survives the rewrites Spark makes between one application of the rule and the next, which is what lets a later application recognise a stage or final-plan pass as a re-plan of something already reported. That turned out to matter beyond the no-ID case: I found a plan with no exchanges gets a final-plan pass under AQE whose hash no longer matches the prep-time plan, because the subquery was substituted in between, so the old dedupe let it through as a second report. I kept the execution-scoped hash alongside the mark for the one case a mark can't cover — a subquery referenced from two places is prepared twice as separate but identical plans. Both `df.rdd.count()` and `executedPlan` before an action are covered by tests now, under AQE on and off.

The subquery one was worse than the double-count you found, and getting the outer report to agree with normal planning took two more things. A subquery reaches the outer plan fully prepared, so its aggregates are already inside `WholeStageCodegenExec` with transitions inserted, and handing that to the conversion pass leaves them unconverted — the preview was reporting a subquery as falling back that Comet actually accelerates. Those artifacts are stripped before previewing. And `ReuseExchangeAndSubquery` is the last thing `QueryExecution.preparations` does, after the columnar rules, so the plan the preview sees still has one copy of the subquery per reference; replaying that rule at the top level collapses them the way the executed plan does. With AQE off the outer report is now exactly `CometCoverageStats` for the plan Comet really executes — 6/7 with 2 transitions on your scalar-subquery query, against 6/12 with 5 transitions before — and there's a test asserting that equality.

One thing I could not make match, and I'd like your read on it. Under AQE the outer report is produced from the prep-rule application, and at that point `PlanAdaptiveSubqueries` hasn't run, so the subquery is still a logical plan hanging off the expression. Extended explain walks it anyway and counts those logical nodes as un-accelerated Spark operators, so the same query reports 2/7 under AQE against 6/7 off. Reporting operators we never evaluated as fallbacks is the part that bothers me. The alternatives I can see are to exclude un-planned subqueries from the counts, which trades a pessimistic number for an optimistic one, or to report from the post-adaptive application instead of the prep one, which is only reachable for queries AQE never cuts into stages. For now it's documented, with a pointer to the per-subquery reports. Would you rather I pursue one of those, or is documenting it acceptable for a first cut?

The remaining red check is `spark-sql-sql_hive-1`, which failed downloading `apache-maven-3.9.6-bin.zip` from Maven Central — unrelated to this PR.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -573,7 +686,93 @@ case class CometExecRule(session: SparkSession)
...
+    plan.transformAllExpressions { case subquery: ExecSubqueryExpression =>
+      subquery.withNewPlan(previewSubquery(subquery.plan))
+    }
+  }
+
+  private def previewSubquery(subquery: BaseSubqueryExec): BaseSubqueryExec = subquery match {
+    // Reuse bookkeeping: the plan to preview is one level further down.
+    case reused: ReusedSubqueryExec => reused.copy(child = previewSubquery(reused.child))
+    case other =>
+      val preview = buildPreview(stripPreparation(other.child), topLevel = false)
```
sunchao: [P2] Apply stage reversion to the prepared DPP child

Could the DPP broadcast child be previewed through its own preparation boundary before restoring the exchange wrapper? With AQE disabled, Spark's `PlanDynamicPruningFilters` prepares the child before constructing `BroadcastExchangeExec`. This path instead passes that exchange into `buildPreview`, so `applyToAllStages` treats it as a boundary and leaves the reconverted child unreverted. I reproduced this on the PR head with a partitioned Parquet fact table joined to a filtered Parquet dimension, `spark.comet.exec.transitionRevert.enabled=true`, `spark.comet.exec.transitionRevert.maxTransitions=0`, and `spark.comet.exec.project.enabled=false`: the outer report says 4/12 accelerated (33%), while `CometCoverageStats` for the normally executed plan says 2/12 (16%). Both executions returned the same ten correct rows. The separate DPP report correctly shows 0/3, but its child is counted as accelerated again in the outer report. Please preserve the child's normal preparation scope and add a non-AQE DPP coverage comparison.

sunchao (commented):
Re-reviewed `1683a65` with five agents. The 41-test CometExecRuleSuite and 96 normal/preview plan comparisons passed on Spark 4.1.3. Targeted probes against locally compiled JVM and native code reproduced the two reporting issues below; the third comment identifies a test coverage gap.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -115,12 +116,124 @@ object CometExecRule {
...
+   *   still be reported.
+   */
+  private def isReapplication(
+      plan: SparkPlan,
+      queryStagePrep: Boolean,
+      aqeEnabled: Boolean): Boolean = {
+    plan.exists(p =>
+      p.isInstanceOf[QueryStageExec] || p.isInstanceOf[AdaptiveSparkPlanExec] ||
+        p.getTagValue(PLAN_ONLY_REPORTED).isDefined) ||
+    (aqeEnabled && !queryStagePrep && plan.isInstanceOf[Exchange])
```
sunchao: [P2] Keep report state when AQE replaces an empty plan

Could report ownership survive replacement of the entire physical tree? With plan-only mode and AQE enabled, `spark.sql("SELECT id % 2 AS k, count(*) AS n FROM range(20) WHERE id < 0 GROUP BY id % 2").collect()` emits an 83% report for the initial aggregate and then a second 0% report for `EmptyRelation`. After the empty shuffle materializes, AQE creates a fresh plan containing neither `QueryStageExec` nor the reporting tag. Its structural hash also differs, so both guards admit another report under the same execution ID. I reproduced the two actual Comet warnings on Spark 4.1.3; AQE disabled emits one. Please retain query/subquery reporting ownership across this rewrite and cover an empty adaptive query.

spark/src/test/scala/org/apache/comet/rules/CometExecRuleSuite.scala
```diff
@@ -797,4 +799,238 @@ class CometExecRuleSuite extends CometTestBase {
...
+    }
+  }
+
+  for {
+    useV1 <- Seq(true, false)
+    aqe <- Seq(true, false)
+  } {
+    val label = s"${if (useV1) "V1" else "V2"} scan, AQE=$aqe"
+    test(s"plan-only mode: $label") {
+      withParquetTable((0 until 100).map(i => (i, i % 5)), "tbl") {
```
sunchao: [P3] Select V2 before creating the Parquet fixture

Could `USE_V1_SOURCE_LIST` be configured outside `withParquetTable`? The fixture reads Parquet and registers `tbl` before `runPlanOnlyAndAssertReverted` changes that setting. Changing it afterward does not replace the existing logical relation, so both tests labeled "V2 scan" still exercise `FileSourceScanExec`. A Spark 4.1.3 probe confirmed this with AQE on and off; creating the fixture after the config change produces `BatchScanExec`. Actual V2 plan-only execution passed separate checks, but these committed tests do not cover that path.

andygrove:
All three fixed in c75a87c.

The DPP one was as you described. `PlanDynamicPruningFilters` prepares the build plan and wraps it in the exchange afterwards, so the preparation boundary — and the stage `RevertNativeForTransitionHeavyStages` judged — is the exchange's child. The preview now descends through the wrapper and puts it back, and on your configuration the outer report reads 2/12 with 3 transitions, matching `CometCoverageStats` for the executed plan exactly. There's a test that builds a partitioned fact table joined to a filtered dimension with reversion forced on and asserts that equality.

The empty adaptive plan I reproduced too, and it turned out the hash was not the only guard that missed: the mark could not reach the new tree either, since `AQEPropagateEmptyRelation` keeps the plan it replaced in `EmptyRelationExec`'s `innerChildren` rather than its children. Instead of chasing it through there I went with a third guard: once AQE has cut a query stage for an execution, everything that arrives afterwards is AQE re-planning something already reported, so the rest of the execution is suppressed. That is ordering-safe for subqueries because they are all compiled before the first stage is cut. The exception is a subquery AQE only plans once stages are under way — a DPP subquery, since `PlanAdaptiveDynamicPruningFilters` is a stage optimizer rule — which now gets no report of its own under AQE. That seemed a better trade than an extra bogus 0% report, but say the word if you'd rather keep it. Noted in the user guide, and tested under AQE on and off.

You were right about the V2 tests, and the cause is that `withParquetTable` resolves the relation through `spark.read` and registers the result as a temp view, so the source list is baked in at creation. The config now wraps the fixture, and each variant asserts it actually got a `FileSourceScanExec` or a `BatchScanExec` so it cannot silently stop covering the path again.

Separately, while building the DPP fixture I hit an unrelated pre-existing bug and filed #5486: with AQE on, DPP in play, and `spark.comet.exec.transitionRevert.enabled=true`, the query throws `SubqueryAdaptiveBroadcastExec does not support the execute() code path`. It reproduces on this PR's base commit with plan-only mode never enabled, and only when `transitionRevert` is on — so it's out of scope here, but it matters for #5207. It's why the new DPP test is non-AQE only, beyond the report being pre-adaptive anyway.

`CometExecRuleSuite` passes on Spark 3.4 (42, 2 pre-existing version-guard cancellations), 3.5 (43, 1) and 4.1 (44), and `RevertNativeForTransitionHeavyStagesSuite`, `CometScanRuleSuite`, `CometCoverageStatsSuite` and `CometDppFallbackRepro3949Suite` are green.

sunchao (commented):
Re-reviewed `c75a87c` with five agents. The 60 existing tests across five suites and 132 non-AQE coverage comparisons passed on Spark 4.1.3. Targeted probes reproduced the two reporting issues below.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -115,12 +116,143 @@ object CometExecRule {
...
+      plan: SparkPlan,
+      queryStagePrep: Boolean,
+      aqeEnabled: Boolean): Boolean = {
+    if (isQueryStage(plan, queryStagePrep, aqeEnabled)) {
+      // Record that AQE has started executing this query before dropping the stage itself.
+      executionId.foreach(planOnlyStagedExecutions.add)
+      false
+    } else if (isReapplication(plan)) {
+      false
+    } else if (executionId.exists(planOnlyStagedExecutions.contains)) {
```
sunchao: [P2] Suppress empty AQE reports without an execution ID

The new guard only retains state when an execution ID exists. With AQE and `spark.comet.explain.planOnly.enabled=true`, an empty grouped query through the JVM bridge used by [PySpark's `df.rdd`](https://github.com/apache/spark/blob/v4.1.3/python/pyspark/sql/classic/dataframe.py#L157-L161) still emits the initial 83% report followed by a misleading 0% `EmptyRelation` report:

```sql
SELECT id % 2 AS k, count(*) AS n
FROM range(20)
WHERE id < 0
GROUP BY id % 2
```

I reproduced this on the current head with Spark 4.1.3 by invoking `javaToPython()` and counting the resulting RDD; AQE disabled emits one report. The action has no SQL execution ID, and the replacement tree has no reporting tag. Could suppression survive this AQE replacement without requiring an execution ID, with a regression covering this RDD path?

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -115,12 +116,143 @@ object CometExecRule {
...
+      false
+    } else {
+      plan.setTagValue(PLAN_ONLY_REPORTED, ())
+      // Node tags are not part of a plan's structural hash, so the mark set above does not
+      // perturb the key. Without an execution ID there is nothing to scope the state to, and the
+      // checks above have already ruled out the repeat applications AQE makes, so report.
+      executionId.forall(id => planOnlyReportedPlans.add(s"$id:${plan.hashCode()}"))
```
sunchao: [P3] Canonicalize nested subqueries before deduplicating reports

Projecting the same nested scalar subquery twice and calling `.collect()` with AQE disabled produces four reports under one execution ID: the inner aggregate, the outer aggregate twice, and the main query. For a Parquet table `tbl` with integer columns `_1` and `_2`, the reproducer is:

```sql
SELECT _1,
  (SELECT max(_2) FROM tbl
   WHERE _1 > (SELECT min(_2) FROM tbl)) AS a,
  (SELECT max(_2) FROM tbl
   WHERE _1 > (SELECT min(_2) FROM tbl)) AS b
FROM tbl
```

On Spark 4.1.3 at the current head, the two outer-aggregate reports are byte-for-byte identical, and both normal and plan-only physical plans reuse that subquery through `ReusedSubqueryExec`. The raw plan hash misses this equivalence, causing redundant previews and warnings. Could the key use a canonical fingerprint, with a nested-subquery reuse regression? This is a reporting issue; results and coverage values still match.

sunchao (commented):
[P2] Re-reviewed `53db476c`. The original aggregate/no-ID case is addressed, but the [existing empty-AQE reporting issue](https://github.com/apache/datafusion-comet/pull/5394#discussion_r3867696864) still has a source-verified residual. When Spark removes an outer local sort before Comet's preparation rule, Comet marks the surviving logical Join, while AQE retains the original, unmarked logical Sort. Empty propagation replaces the Join and then the Sort; the replacement root inherits the Sort's unmarked tags and can pass the reporting guards again without a SQL execution ID.

The additional trigger is a two-partition Range merge join on `a.id % 7 = b.id % 7`, filtered by `a.id < 0`, with `SORT BY a.id % 7` and execution via `queryExecution.toRdd.count()`. The added aggregate/no-ID and nested-subquery regressions do not cover this removed-root shape. This path was traced through current source and Spark 4.1.3, not executed in a runtime reproduction.

andygrove:
I am moving this to draft. This seems overly complex. I will explore alternative approaches.

andygrove:
The reason that this is so complex is because we have two separate rules - CometScanRule and CometExecRule. If these get combined into a single rule then we can just create the comet plan and discard it and no longer need to roll things back. Working on it ....

andygrove:
Rebased onto #6082, which composes `CometScanRule` and `CometExecRule` into a single `CometRule`. That PR should go in first; until it does, the diff here includes it.

The reason for splitting it out is that most of the awkwardness in this PR came from the two rules being registered separately. With one rule there is a single place that holds the whole conversion, so plan-only mode gets a single short-circuit and a single call to build the plan it reports on:

- `CometScanRule` is back to its version on main. It no longer knows about plan-only mode at all, and its `_apply` is private again.
- `CometExecRule` loses the `forPreview` parameter, the plan-only branch in `_apply`, and the `queryStagePrep` constructor parameter. It is now purely operator conversion, and this PR touches 4 lines of it rather than 287.
- The reporting state, the reporting decision and the preview machinery live on `CometRule`. `buildPreview` calls `convert(...)` rather than reaching into both rules' private `_apply`s.

One behavioural note. The plan-only check used to sit inside `CometExecRule`'s exec-enabled branch, so it never engaged with `spark.comet.exec.enabled=false`. `CometRule.planOnlyApplies` keeps that scoping, along with the `isCometLoaded` and streaming guards, so a plan the conversion rules would have left alone is not diverted into a report and Comet's columnar shuffle still applies with exec disabled.

Also fixed here: `IcebergWriteStrategy` is registered with `injectPlannerStrategy`, so it runs ahead of both conversion rules and was not covered by the short-circuit — an Iceberg V2 write was still being offloaded under plan-only mode. Neither `IcebergCommitExec` nor `IcebergWriteExec` extends `CometPlan`, so the existing "no Comet operators" assertions could not see it. It now declines in plan-only mode, with a test in `CometIcebergWriteDetectionSuite` and a note in the user guide about the coverage that costs.

Local runs: `CometExecRuleSuite` 62, plus `CometIcebergWriteDetectionSuite`, `CometScanRuleSuite`, `RevertNativeForTransitionHeavyStagesSuite` and `CometScanSchemeFallbackSuite` at 81 between them. `test-compile` clean on the default profile and `-Pspark-3.4`.

andygrove:
rebased on main

andygrove:
@sunchao on the removed-root-sort residual from your last review: this is handled in 0896e36ed. Rather than chasing the mark through the rewrite, the query-stage-prep rule now recognizes a re-plan that AQE collapsed to nothing directly: its logical link has `maxRows == 0`. A genuinely empty query is still reported, because it is planned once and reaches the columnar rule instead. The empty-plan test matrix now includes your two-partition range merge join with `SORT BY a.id % 7`, under both `collect()` and `queryExecution.toRdd.count()`, with AQE on and off.

andygrove:
@sunchao @comphead could you help review?

sunchao:
Reviewed `a80ae3fce` with **five agents**. I found **three reproducible reporting issues** to address before merging.

1. **P2 — `explain.format=fallback` removes the coverage report.**  
   [CometRule.scala:154](https://github.com/apache/datafusion-comet/blob/a80ae3fce500cd180dd87eecb3edeff2b1dd7671/spark/src/main/scala/org/apache/comet/rules/CometRule.scala#L154) uses a formatter that honors the existing explain-format setting. A supported Parquet projection reports **2/2 operators, 100%** with `verbose`, but logs only `[Comet plan-only]` with `fallback`. Render the plan and coverage explicitly, independent of that setting.

2. **P2 — Initially empty AQE queries can receive no report.**  
   [CometRule.scala:90](https://github.com/apache/datafusion-comet/blob/a80ae3fce500cd180dd87eecb3edeff2b1dd7671/spark/src/main/scala/org/apache/comet/rules/CometRule.scala#L90) treats `maxRows == 0` as evidence of adaptive re-planning. However, `SELECT id FROM range(0) DISTRIBUTE BY id` satisfies this during initial planning. It emits **zero reports with AQE enabled**, versus one with AQE disabled. The rule also marks the plan as reported, preventing later recovery. Distinguish initial preparation from an actual re-plan.

3. **P3 — Repeated subqueries rebuild previews without an execution ID.**  
   [CometRule.scala:95](https://github.com/apache/datafusion-comet/blob/a80ae3fce500cd180dd87eecb3edeff2b1dd7671/spark/src/main/scala/org/apache/comet/rules/CometRule.scala#L95) bypasses canonical deduplication when the ID is absent. The PR’s repeated nested-subquery query emits **seven reports, only three distinct**, through `queryExecution.toRdd.count()` with AQE off; AQE on emits five. `collect()` emits three. This repeats preview construction and identical warnings. Deduplication needs query-scoped ownership that also works without an execution ID.

The overall design is sound: composing conversion in `CometRule` and sharing post-columnar rules keeps eligibility aligned with normal planning. The subquery handling reflects real Spark preparation boundaries. Reporting ownership remains the fragile part. I found no additional substantiated result-correctness, native-lifetime, or default-off performance defect.

**Validation:** Current CI has 23 successful checks and 14 skipped. Its execution group passed 999 tests, including all 62 `CometExecRuleSuite` tests. Spark SQL and Iceberg SQL matrices were skipped. [CI run](https://github.com/apache/datafusion-comet/actions/runs/35676248431)

Local probes used Spark 4.1.3/JDK 17 with freshly compiled current Scala changes over cached dependencies and native library. Disk constraints prevented a clean full build. These are targeted integration reproductions; no whole-query performance claim is made.

andygrove:
@sunchao all three addressed in 31230882a.

1. Plan-only reports now render through `ExtendedExplainInfo.generateVerboseInfo`, which always emits the annotated tree and coverage summary. The `verbose` format uses the same method, so its output is unchanged. There's a test with the format set to `fallback`.

2. The `maxRows == 0` guess is gone. A prep-rule pass counts as an AQE re-plan only when it runs inside `AdaptiveSparkPlanExec.reOptimize`, so initial preparation is never mistaken for one. `SELECT id FROM range(0) DISTRIBUTE BY id` now gets one report with AQE on. The old guess also missed a re-plan that isn't itself empty: `count(*)` over a join that becomes empty at runtime was reported twice. Both shapes are in the empty-plan matrix under AQE on/off and `collect`/`toRdd.count`, and both fail against the previous commit.

3. Subquery dedupe no longer uses the execution ID. Spark prepares a query's subqueries on the same thread just before the query itself, so the canonical hashes are kept in a thread-local record. It is cleared when the query's own plan arrives, and again once execution passes begin. A subquery preparation is recognised by `QueryExecution.prepareExecutedPlan` (non-AQE and DPP) or `InsertAdaptiveSparkPlan.compileSubquery` (AQE) on the call stack. Your nested-subquery query now gives 3 reports under AQE on/off with both `collect()` and `queryExecution.toRdd.count()`, and that matrix is now tested.

These checks read Spark method names off the stack. I checked that `reOptimize`, `prepareExecutedPlan` and `compileSubquery` exist unchanged from 3.4 through 4.2. If one were ever renamed, the result would be duplicate reports; execution isn't affected. The plan-only tests pass on Spark 3.4 and 4.1, and `CometExecRuleSuite`, `RevertNativeForTransitionHeavyStagesSuite`, `CometScanRuleSuite` and `CometIcebergWriteDetectionSuite` pass on 4.1.

andygrove:
@sunchao could you take another look? I'd like to get this one into the 1.1.0 release. Thanks!

PR #5395, opened 2026-08-19
perf: vectorize Map in spark_size via offset buffer

### LINKED ISSUE #5266: Use the offset buffer for Map in spark_size
Follow-up from #5233 
`spark_size` for `List` / `LargeList` / `FixedSizeList` now reuses Arrow's `length` kernel. 
The `Map` branch in `native/spark-expr/src/array_funcs/size.rs` is still a per-row loop that appends `MapArray::value_length(i)` into an `Int32Builder`.
`length` does not accept `MapArray`, so Map cannot reuse the kernel. It was kept out of #5233 on purpose: different technique, and its own correctness and benchmark surface.     

`MapArray::offsets()` is the entry-range offset buffer; for each row, `offsets[i+1] - offsets[i]` is the entry count (same value `value_length(i)` returns one row at a time). Vectorize that and apply Spark's null → `-1` rewrite with the same `set_indices` patch the list path uses                                                                                          
                                                                                                                             
### Additional context                                                                                                                   
Before merging:                                                                                                                          
- Add a Map shape (with and without nulls) to `benches/array_size.rs` — there is no Map coverage today. The win is expected to be smaller than for List, since `value_length` already reads the offsets; only per-row dispatch and builder overhead go away. If the benchmark comes back flat, closing this as not worth the churn is fine.                                                                            
- Add a sliced-`MapArray` test, mirroring the sliced-List test added in #5233. `MapArray::offsets()` returns the sliced offsets so `windows(2)` stays correct, but pin the invariant.

0lai0:

Closes #5266.

## Rationale for this change

Follow-up to #5233. Arrow's `length` kernel does not accept `MapArray`, so Map still went through a per-row `Int32Builder` loop over `MapArray::value_length(i)`. Each `value_length` is already O(1) from the offset buffer; the cost was the per-row `is_null` branch plus builder appends, which also blocked autovectorizing the offset diffs.

- `spark_size_map` differences `MapArray::offsets()` with `windows(2)` in one shot. `offsets()` returns the sliced buffer, so sliced inputs stay correct.
- Extract `rewrite_nulls_to_minus_one` so List and Map share the null → `-1` rewrite (`into_parts` + `set_indices`). The list-like `null_count() == 0` fast path is unchanged: return the `length` kernel output as-is and skip the downcast / `Int32Array::clone` that the helper would otherwise pay.
- Criterion shapes for Map with and without nulls (`benches/array_size.rs`).
- Unit tests for no-null Map and sliced Map (mirrors the sliced-List test from #5233).
- Docs table: Map array path ~6-13x; `length` does not support Map.

The with-nulls Map path still materializes `Vec<i32>` then clones again inside `rewrite_nulls_to_minus_one`. Writing `-1` during the first collect would drop that second allocation; left as a follow-up because the path is already ~6.5x faster than the loop in #5233.

- Unit tests in `native/spark-expr/src/array_funcs/size.rs`:
  `test_spark_size_map_no_nulls`, `test_spark_size_sliced_map_array` (slice(2, 3) → `[3, -1, 1]`).
- Existing SQL coverage in `sql-tests/expressions/array/size.sql` (column map, literal map fallback, `cast(NULL as map)`).
- Criterion (`benches/array_size.rs`), 8192 rows, M-series Mac, arrow 58.4.0. Same session, only `size.rs` swapped, bench file held fixed:

| shape                      | before  | after  | speedup |
| -------------------------- | ------- | ------ | ------- |
| spark_size: map, no nulls  | 5.85 µs | 452 ns | ~13x    |
| spark_size: map (10% null) | 8.22 µs | 1.27 µs | ~6.5x   |

List / LargeList moved by less than noise, as expected (only the Map path changed).


sunchao:
@0lai0 can you check and fix code style issue below?
```
Checking formatting...
[warn] docs/source/contributor-guide/optimizing_expressions.md
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```
Otherwise the PR LGTM

sunchao (approved):

sunchao:
Merged, thanks!

PR #5400, opened 2026-08-20
chore: fix clippy warnings for Rust 1.98

### LINKED ISSUE #5399: CI: Rust 1.98 Clippy lints break rust-test
## Description

The `PR Build (Linux) / ubuntu-latest/rust-test` job started failing after Rust stable moved from 1.97.1 to 1.98.0.

The workflow uses `RUST_VERSION: stable` and runs:

    cd native
    cargo clippy --color=never --all-targets --workspace -- -D warnings

Rust 1.98 introduces new Clippy warnings that are treated as errors because of `-D warnings`.

Example failing job:

https://github.com/apache/datafusion-comet/actions/runs/32406126292/job/96546147984

A recent `main` run using Rust 1.97.1 passed the same rust-test job:

https://github.com/apache/datafusion-comet/actions/runs/32392348862/job/96502048719

## Errors

    error: using `chunks_exact` with a constant chunk size
      --> spark-expr/src/bloom_filter/spark_bit_array.rs:80:64

    help: consider using `as_chunks::<8>()` instead

    error: manual implementation of `isolate_lowest_one`
      --> spark-expr/src/nondetermenistic_funcs/internal/mersenne.rs:152:12

    help: consider using `.isolate_lowest_one()`

## Notes

Comet declares `rust-version = "1.88"` in `native/Cargo.toml`.

The `as_chunks::<8>()` suggestion is compatible with Rust 1.88.

However, `isolate_lowest_one()` is only available since Rust 1.97, so applying that suggestion directly would raise the effective MSRV. A targeted Clippy allow/comment or MSRV-aware Clippy configuration may be preferable.

This appears to be a repo-wide CI issue rather than a regression from any specific PR.

_No response_

### Additional context

_No response_

ywskycn:

Closes #5399.

### DIFF
--- a/native/spark-expr/src/nondetermenistic_funcs/internal/mersenne.rs
+++ b/native/spark-expr/src/nondetermenistic_funcs/internal/mersenne.rs
@@ -148,6 +148,8 @@ impl SparkMersenneTwister {
 
     /// Port of `BitsStreamGenerator.nextInt(int n)`. The caller always passes a
     /// strictly positive `n`, matching Spark's `random.nextInt(i + 1)`.
+    // `isolate_lowest_one` requires Rust 1.97, newer than Comet's Rust 1.88 MSRV.
+    #[allow(clippy::manual_isolate_lowest_one)]
     pub(crate) fn next_int(&mut self, n: i32) -> i32 {
         if (n & n.wrapping_neg()) == n {
             // n is a power of two.
--- a/native/spark-expr/src/nondetermenistic_funcs/internal/mersenne.rs
+++ b/native/spark-expr/src/nondetermenistic_funcs/internal/mersenne.rs
@@ -149,7 +149,7 @@ impl SparkMersenneTwister {
     /// Port of `BitsStreamGenerator.nextInt(int n)`. The caller always passes a
     /// strictly positive `n`, matching Spark's `random.nextInt(i + 1)`.
     // `isolate_lowest_one` requires Rust 1.97, newer than Comet's Rust 1.88 MSRV.
-    #[allow(clippy::manual_isolate_lowest_one)]
+    #[allow(unknown_lints, clippy::manual_isolate_lowest_one)]
     pub(crate) fn next_int(&mut self, n: i32) -> i32 {
         if (n & n.wrapping_neg()) == n {
             // n is a power of two.
--- a/native/spark-expr/src/bloom_filter/spark_bit_array.rs
+++ b/native/spark-expr/src/bloom_filter/spark_bit_array.rs
@@ -77,8 +77,8 @@ impl SparkBitArray {
     pub fn merge_be_words(&mut self, incoming: &[u8]) {
         debug_assert_eq!(self.data.len() * 8, incoming.len());
         let mut bit_count: usize = 0;
-        for (word, chunk) in self.data.iter_mut().zip(incoming.chunks_exact(8)) {
-            *word |= u64::from_be_bytes(chunk.try_into().unwrap());
+        for (word, chunk) in self.data.iter_mut().zip(incoming.as_chunks::<8>().0) {
+            *word |= u64::from_be_bytes(*chunk);
             bit_count += word.count_ones() as usize;
         }
         self.bit_count = bit_count;


native/spark-expr/src/nondetermenistic_funcs/internal/mersenne.rs
```diff
@@ -148,6 +148,8 @@ impl SparkMersenneTwister {
 
     /// Port of `BitsStreamGenerator.nextInt(int n)`. The caller always passes a
     /// strictly positive `n`, matching Spark's `random.nextInt(i + 1)`.
+    // `isolate_lowest_one` requires Rust 1.97, newer than Comet's Rust 1.88 MSRV.
+    #[allow(clippy::manual_isolate_lowest_one)]
```
sunchao: [P2] Keep this allowance compatible with pre-1.98 Clippy

On Clippy 1.97.1 (and 1.91), this newly introduced lint name is itself unknown, so the documented `cargo clippy --all-targets --workspace -- -D warnings` workflow fails with `unknown lint: clippy::manual_isolate_lowest_one` (`-D unknown-lints` is implied by `-D warnings`). I reproduced this with the unmodified Mersenne module from both commits: the exact base passes and this head fails at this attribute. Ordinary Rust compilation still succeeds. Since these toolchains are newer than the declared Rust 1.88 minimum, please make the allowance backward-compatible, for example `#[allow(unknown_lints, clippy::manual_isolate_lowest_one)]`; that form passes both Clippy 1.97.1 and the installed 1.99 nightly.

sunchao (commented):
## Summary

Reviewed `e6572056cc0e05473dae51aa436550750a9be604` against `91f9fbed24ee28f99b23ed559e80e818c9696044`. All five independent reviewers inspected the full exact diff and their surrounding-code scopes. The two edits address the Rust 1.98 Clippy failures without changing the intended Bloom-filter or random-number behavior. One P2 compatibility issue remains: the newly named Clippy allowance itself fails under the documented warnings-as-errors command on pre-1.98 toolchains.

## Prior state and problem

The Linux Rust workflow follows `stable` and treats every Clippy warning as an error. Rust 1.98 introduces diagnostics for constant-size `chunks_exact` and the manually isolated low bit. Comet declares Rust 1.88 as its minimum, so adopting the newer integer method directly would not preserve that compatibility goal.

## Design approach

The Bloom merge now borrows fixed-size byte-array chunks and decodes each copied array as a big-endian word. The Mersenne Twister implementation keeps the existing Java-compatible calculation and suppresses only the new Clippy diagnostic on `next_int`.

Both chunk forms visit the same complete eight-byte groups, and the production Bloom caller validates the word count and passes an exactly sized slice. Word ordering, OR semantics, and cardinality updates are unchanged; the PRNG executable statements are also unchanged. Direct-source differential and golden-vector probes found no runtime regression.

The new slice API is available at Rust 1.88. The inline finding is narrower: older Clippy versions do not recognize the new lint name, and `-D warnings` promotes that unknown-lint diagnostic to an error. Unmodified base/head Mersenne-module probes pass at the base and fail at the head on Clippy 1.91 and 1.97.1. Ordinary Rust compilation is unaffected. The exact-head Rust 1.98 CI job passes Clippy and reports 895 tests passed, with four skipped; no full workspace suite was rerun locally.

## Key design decisions

Retaining the existing low-bit calculation preserves the older compiler API surface and avoids changing Spark-compatible random sequences. Limiting the suppression to one function also keeps unrelated Clippy checks active. The allowance needs to tolerate Clippy versions that predate this particular lint.

## Implementation sketch

In `spark_bit_array.rs`, `as_chunks::<8>().0` supplies `&[u8; 8]` items, so `from_be_bytes(*chunk)` replaces the fallible slice conversion without allocating. In `mersenne.rs`, only the explanatory comment and lint attribute are added; the bounded-draw and rejection-sampling paths remain intact.

Rust 1.98's strict lint job is repaired, and no query-result or serialized Bloom-filter change was found. However, contributors still using an older toolchain such as 1.97.1 now encounter an unknown-lint error in the same documented local Clippy workflow. This is a lint-workflow regression, not a normal `cargo build` failure.

## Suggested improvements

Make the item-level allowance backward-compatible, for example by adding `unknown_lints` to the same `allow` attribute. That narrow form was checked on Clippy 1.97.1 and the installed 1.99 nightly. Keep the existing runtime implementation and verify the strict lint command on one pre-1.98 toolchain as well as current stable.

native/spark-expr/src/nondetermenistic_funcs/internal/mersenne.rs
ywskycn: fixed

sunchao (approved):

Issue #4496, opened 2026-05-28
Do another audit sweep for string collation differences


andygrove:
### What is the problem the feature request solves?

The recent audit of string functions (https://github.com/apache/datafusion-comet/pull/4461) seemed to gloss over potential compatibility issues related to string collation. The audit skill may need enhancing.

PR #5403, opened 2026-08-21
fix: match Spark ordering in native array extrema

### LINKED ISSUE #5401: [Bug] array_min and array_max differ from Spark on signed-zero ties
## Describe the bug

`array_min` and `array_max` do not preserve Spark's choice of the first element when `+0.0` and `-0.0` compare equal. The corrected fixtures in #5393 expose this for `array_min`. The reverse zero order also exposes it for `array_max`.

At upstream commit `91f9fbed24ee28f99b23ed559e80e818c9696044`, both expressions are classified as compatible without checking `spark.comet.exec.strictFloatingPoint`. They therefore use the differing native implementation even when strict floating-point behavior is requested.

## Reproduction

Use a Parquet column so the extrema are evaluated at runtime:

```sql
SET spark.sql.optimizer.excludedRules=org.apache.spark.sql.catalyst.optimizer.ConstantFolding;
SET spark.comet.exec.strictFloatingPoint=true;

CREATE TABLE comet_extrema_zero_repro (id INT, arr ARRAY<DOUBLE>) USING parquet;
INSERT INTO comet_extrema_zero_repro VALUES
  (1, array(double('0.0'), double('-0.0'))),
  (2, array(double('-0.0'), double('0.0')));

SELECT id, array_min(arr), array_max(arr)
FROM comet_extrema_zero_repro ORDER BY id;
```

The same issue occurs for `ARRAY<FLOAT>`.

| Input | Spark min | Native min | Spark max | Native max |
| `[+0.0, -0.0]` | `+0.0` | `-0.0` | `+0.0` | `+0.0` |
| `[-0.0, +0.0]` | `-0.0` | `-0.0` | `-0.0` | `+0.0` |

Spark 3.5.8 and 4.0.1 preserve the first zero. A direct probe of the pinned DataFusion 54.1.0 / Arrow 58.4.0 UDFs confirms the native results for both floating-point widths and on both sides of the array-length-32 kernel boundary.

## Expected behavior

Strict floating-point mode should use Spark-compatible evaluation. The native implementation should eventually preserve Spark's first-equal-element behavior as well. A narrow first fix can route strict-mode extrema through the existing Spark-codegen fallback, while keeping this issue open for native parity.

Relevant code: `CometArrayMin` and `CometArrayMax` in `spark/src/main/scala/org/apache/comet/serde/arrays.scala`. This is a follow-up to #5271 and #5393, not a regression introduced by those fixture changes.

sunchao:

Closes #5401.

## Rationale for this change

Spark treats signed zeros as equal and keeps the first equal element in `array_min` and `array_max`. Native evaluation can instead select the opposite zero sign:

| Floating-point array column `a` | Expression     | Spark  | Before this change |
| ------------------------------- | -------------- | ------ | ------------------ |
| `[+0.0, -0.0]`                  | `array_min(a)` | `+0.0` | `-0.0`             |
| `[-0.0, +0.0]`                  | `array_max(a)` | `-0.0` | `+0.0`             |

For nested values, this can select a different whole element. Given structs `[(-0.0, 2), (+0.0, 1)]`, Spark's maximum compares the second field after the equal zero keys and returns `(-0.0, 2)`. Nested comparisons also require Spark's nulls-first ordering.

- Add native floating-point scans and recursive list/struct comparators that preserve the first winning value, including zero signs and NaN bits. All NaNs compare equal and greater than numbers. Other scalar element types retain their existing DataFusion implementation.
- Keep floating-point and binary-collated inputs native in both strict and non-strict modes. Non-default string collations, including nested strings, use Spark's JVM codegen dispatcher; disabling the dispatcher falls back to Spark unless incompatible execution is explicitly enabled. Native collation support remains tracked in #4496.
- Copy selected nested results without retaining buffers sized for losing candidates. Use Arrow's flat-list copy where appropriate and bounded initial child capacity for sparse/deep results. This bounds retained result buffers, not peak process memory.
- Limit custom layout handling to the `List`/`Struct` representations supplied by Spark. Consolidate min/max SQL fixtures, raw-bit cases, and benchmark setup; retain focused ordering, scalar/slice, delegation, collation, and memory regressions. Previously ignored signed-zero cases now require native execution.
- Update compatibility documentation and use Arrow's fallible copy APIs, replacing deprecated calls rejected by the previous revision's Rust CI.

Local checks for the simplified revision:

- Spark **4.1.3** production/test compilation and **12 `SqlFileTestParserSuite` tests** passed, along with Spotless and Scalastyle. Both combined SQL fixtures also passed SQL syntax and explicit native/dispatch assertion checks; this does not execute the queries.
- **Seven extracted-kernel tests** passed against locked Arrow **59.3.0**, covering raw-bit float ordering, sliced inputs, nested comparisons, and result-buffer bounds. This isolated harness excludes DataFusion UDF integration, scalar conversion, delegation, and the empty-batch wrapper.
- Rust formatting, changed Markdown formatting, and `git diff --check` passed.
- **Full native/JNI validation remains blocked locally:** the configured Cargo registry lacks locked DataFusion **55.1.0**, so the focused native command stopped before compilation. No dependency versions or registries were changed. The nine committed native tests, Spark integration fixtures, and full Comet suite have not been run on this revision.

Performance remains a tradeoff. Earlier measurements against DataFusion 54.1.0 found approximately **1.5–1.6× latency** for long, null-free FLOAT arrays; these measurements were not repeated against 55.1.0. Arrow's winner alone cannot identify when correction is needed: for `array_max([-NaN, 1.0])`, Arrow can choose `1.0` while Spark chooses NaN, so checking only zero/NaN winners is insufficient.

Isolated before/after kernel probes checked equal outputs and retained-buffer sizes. They supported retaining the compact list/struct copy helper. One Float64 maximum case was about **12% slower** with normal inlining; separately measured float kernels were within **1%**, indicating code-generation sensitivity. These probes do not establish full native or end-to-end Spark performance parity, and this PR makes no general speedup claim.


spark/src/test/resources/sql-tests/expressions/array/array_min.sql
```diff
@@ -49,16 +49,16 @@ INSERT INTO test_array_min_double VALUES
...
+-- Native parity is tracked by https://github.com/apache/datafusion-comet/issues/5401.
 statement
 CREATE TABLE test_array_min_dbl_negzero(arr array<double>) USING parquet
 
 statement
 INSERT INTO test_array_min_dbl_negzero VALUES
   (array(0.0, double('-0.0'), 1.0))
 
-query ignore(array_min signed-zero: Spark +0.0, Comet -0.0)
+query ignore(https://github.com/apache/datafusion-comet/issues/5401)
```
comphead: is it any chance to use `query expect_fallback` so once this fixed we would know the test should be addressed as well, instead of ignoring it

sunchao: For cases where fallback is expected, I agree that `query expect_fallback(...)` is a better regression check than `ignore`.

I've reworked `array_min` and `array_max` locally to follow Spark's ordering natively. In that revision, these signed-zero cases use plain `query` assertions, so they must execute natively and match Spark in both strict and non-strict modes. The separate non-default-collation tests use `query expect_fallback(...)`, since those cases still require Spark by default.

The rework is ready locally but hasn't been pushed yet, so the current PR diff still contains the old `ignore`. I'll update the branch once I can publish it.

comphead:
thanks @sunchao I'm spinning up the Claude to make a first round of review, considering the PR size

andygrove:

This is a good piece of work. Fixing native evaluation rather than routing through the codegen dispatcher is the right call, and the recursive `spark_comparator` with nulls-first ordering, dictionary decoding, and struct lexicographic comparison looks careful. I particularly appreciate that the description reports the slowdown honestly instead of burying it, and that the tests compare raw bits rather than numeric equality.

The float slowdown is the thing I want to talk about.

**1.5x to 1.6x on plain float arrays**

`float_extrema` gives up Arrow's vectorized min/max and does a scalar loop with a per-element null check and two branches. That is a real cost paid by every user on every `array_min` and `array_max` over a float column, in order to fix a case that only arises when the array contains both zero signs or multiple NaN encodings.

There is a fast path available here. Arrow's kernel and Spark disagree only when the winning value is a zero or a NaN. For any other winner, all tied candidates are bit-identical, so "keep the first" and "keep whichever Arrow found" produce the same bytes. So you could run Arrow's kernel first, and only when the result is `0.0`, `-0.0`, or NaN fall back to the scalar first-occurrence scan for that one list. Arrays with no zeros and no NaNs, which is essentially all real data, would then run at DataFusion's speed.

Would that work? If there is a reason it does not, it would be good to say so in the code, because a permanent 1.6x on the common path to fix an edge case is a hard trade to justify otherwise.

**`FixedSizeList` input reaches an `exec_err`**

`needs_spark_ordering` returns true for `FixedSizeList` as an element type, and `spark_comparator` handles it. But `invoke_with_args` matches only `DataType::List` and `DataType::LargeList` on the outer array and returns `exec_err!("{} does not support type {other}")` for anything else. So `array_max` over a `FixedSizeList` column would now fail at runtime rather than falling back.

I could not find a path where Comet produces a `FixedSizeList` here, so this may well be unreachable. If it is, could the match arm say so, along the lines of "Comet only produces List and LargeList; FixedSizeList is unreachable"? And if it is reachable, this needs either handling or a plan-time `Unsupported`.

**Two smaller things**

`spark_comparator`'s `Dictionary` arm calls `cast(array, value_type)`, which materializes the whole decoded values array on every invocation. For a heavily dictionary-encoded column that is a full decode per batch. Is that acceptable, or worth a note that dictionary elements are the slow path?

The `Cargo.toml` change adds a trailing newline to an unrelated line. Not a problem, just noting it so it does not look accidental in the diff.

sunchao:
Updated in [9cd863dfa](https://github.com/apache/datafusion-comet/commit/9cd863dfaeaf27fe79a8710ac1e05ac26867a517).

I checked the proposed Arrow-first fast path against the pinned Arrow/DataFusion implementation. It is not safe to trigger the corrective scan only when Arrow returns zero or NaN: for `array_max([-NaN, 1.0])`, Arrow returns `1.0`, while Spark returns NaN. The mismatch reproduces for FLOAT and DOUBLE with list lengths 2, 32, and 1,024. I added that counterexample beside `float_extrema` so a future optimization does not reintroduce it. A safe vectorized path would also have to establish that the input contains no exceptional values, and would need to be measured across null densities and lengths. I have not substituted an unmeasured two-pass algorithm here.

The reported 1.5–1.6× latency applies to the measured long, null-free FLOAT microbenchmark, not every floating array or an end-to-end Spark query. That cost remains explicit in the description; this clarification does not claim a performance improvement.

The `FixedSizeList` concern does not reach the fallback arm through the supported function signature: `invoke_with_args` first calls the delegated DataFusion `return_type`, which rejects an outer fixed-size list. Supporting fixed-size lists as nested elements is a different capability. The code now states that boundary rather than assuming that Comet can never construct such an Arrow array.

Dictionary values are decoded once when the comparator is constructed for a child array/batch, and the comparator retains that decoded array for its index comparisons. This is not a decode per element comparison; I added that scope to the comment. I have not claimed the decode is free or that deep dictionary nesting was benchmarked.

The earlier ignored-case requests are already implemented as native assertions, with separate collation-fallback checks. The final newline in `Cargo.toml` has no behavioral effect. This update only changes explanatory source comments: runtime code, tests, and the previously reported performance results are unchanged. `cargo fmt --all --check` and `git diff --check` pass; no new full Spark build is claimed.

sunchao:
Updated in [07c8de2f7](https://github.com/apache/datafusion-comet/commit/07c8de2f7e72b837e5cbc4824a48ffb79e9c187c).

Merged Apache main `525c05b52` in [07c8de2f7](https://github.com/apache/datafusion-comet/commit/07c8de2f7e72b837e5cbc4824a48ffb79e9c187c) to resolve the branch conflicts without rewriting its history. The resolution retains both the array-extrema and upstream signed-zero guide sections, plus all benchmark registrations from both sides. The extrema implementation, native/Spark regressions, and admission code remain unchanged.

The merged source passes **684 expression-library and 35 common-library unit tests**, including all **24 extrema tests**. Rust formatting, the guide's Prettier check, and the PR diff check against the merged main commit also pass. The description now distinguishes this fresh validation from the earlier optimized benchmarks, full-workspace checks, and Spark results, which were not rerun for this merge.

sunchao:
The [macOS Spark 4.0 scans job](https://github.com/apache/datafusion-comet/actions/runs/33099403856/job/98625302173) at `07c8de2f7` crashed during native thread cleanup, just after `native scan on fake fs` passed. This was a JVM `SIGBUS`, not a ScalaTest assertion failure in the extrema changes.

I downloaded the exact native artifact and verified its ZIP digest against the job log. Symbolication places the saved return address at `hdfsThreadDestructor + 0x50`, matching the prior crash in #5023 and the libhdfs fix in PR #5036. The HDFS/JNI/storage code, fake-filesystem suite and dependency lockfile are unchanged from merged main `525c05b52`. This identifies the known libhdfs teardown failure; it does not make the failed job a passing result. The parent workflow is still active, and I have not claimed a clean macOS rerun.

rich7420:
@sunchao thanks for the patch! please resolve the conflicts

andygrove (commented):
`CometArrayMax` and `CometArrayMin` in `spark/src/main/scala/org/apache/comet/serde/arrays.scala` now report `Incompatible` for non-UTF8_BINARY collations, but they do not mix in `CodegenDispatchFallback`, so a collated `array_max` falls the whole projection back to Spark. `CometArrayIntersect` and `CometArrayJoin` in the same file, and `CometReverse` in `collectionOperations.scala`, hit the identical collation problem and take the dispatcher route instead, which keeps the projection in the Comet pipeline while Spark's own `doGenCode` supplies the comparison. `ArrayMin.doGenCode` reaches `CodeGenerator.genComp`, which emits `semanticCompare` with the collation id for a collated `StringType`, so the dispatcher looks like it would be exactly right here. The collation fixtures assert `expect_fallback` with `spark.comet.exec.scalaUDF.codegen.enabled=true`, so this reads as deliberate. Was the mixin considered and ruled out? If it was, recording why next to `ArrayExtremaSupport` would help, because as it stands this serde and the three neighboring ones disagree for no visible reason.

I went back over my earlier suggestion of running Arrow's kernel first and rescanning only when the winner is a zero or a NaN, and your `array_max([-NaN, 1.0])` counterexample is right. Arrow orders floats with `total_cmp`, which sorts a negative NaN below `-Infinity`, so the disagreement is invisible from the winner alone. I checked `float_comparator` and `float_extrema` against `SQLOrderingUtil.compareDoubles`, and the nested nulls-first ordering against `PhysicalArrayType.interpretedOrdering` and `InterpretedOrdering`, and they line up. The empty-array and all-null cases return null the way `nullSafeEval` does.

One thing for when you resolve the conflicts. `main` has moved to DataFusion 55.0 and Arrow 59.2 while this branch is still on 54.1 and 58.4. I diffed `min_max.rs` in `datafusion-functions-nested` and `ord.rs` in `arrow-ord` across those versions and neither changed behavior, so the delegation and the `make_comparator` assumptions should still hold, but the green run on this head predates the bump.

andygrove (commented):
Still just the one question from last time, about the `CodegenDispatchFallback` mixin, so this is not blocked on anything else I have found.

While I was back in here I checked a few more of the ordering claims and they all hold. `hasNonDefaultStringCollation` recurses through arrays, maps and structs, so the collation guard covers a collated string nested inside a struct element and not only a top-level one. The replacement predicate in `float_extrema` matches `SQLOrderingUtil.compareDoubles` composed with the first-wins rule in `ArrayMax.nullSafeEval` on every combination I could construct, including both NaN signs against each other, both zero signs, and NaN against infinity. `compare_ranges` with `nulls_first` reproduces `PhysicalArrayType.interpretedOrdering`, nulls first then shorter array first, and the struct comparator matches `InterpretedOrdering` under `NullsFirst`.

The conflicts are all shallow. `Cargo.toml` is just the bench list, `comet_scalar_funcs.rs` is an import list, and the two doc files are ordinary drift. The DataFusion 55 and Arrow 59.2 bump from #5262 is in main now, so the green run on this head predates it.

sunchao:
Updated in [ee57cac60](https://github.com/apache/datafusion-comet/commit/ee57cac609d6e42fbbf136ca43f690e7443a00b5), with the main merge in [5522ab750](https://github.com/apache/datafusion-comet/commit/5522ab750a33f23e47811f1a9011effc0ac7c734).

@andygrove The `CodegenDispatchFallback` mixin is appropriate here. Both extrema serdes now use it for non-default collations, including nested arrays and structs. Spark's bound expression preserves the collation IDs through closure serialization, and its generated comparator supplies the collation-aware ordering. Ordinary binary-string and floating-point inputs continue to select the native implementation; `hasConditionalNativeDefault` keeps that distinction in the generated docs.

The collation SQL fixtures now require `expect_dispatch(...)`, with `expect_native(...)` controls for binary strings and floats. A separate regression checks fallback with the dispatcher disabled in both strict modes. The new JVM-only regression covers all eight min/max and nested-type combinations through serialization and code generation.

@rich7420 The conflicts are resolved against main `3b942e593`, preserving the newer benchmark registrations, function imports, and documentation. This brings the branch to DataFusion **55.1.0** / Arrow **59.2.0**. The unrelated Cargo.toml final-newline change is also removed.

Validation: **64 JVM-only tests passed on Spark 4.1.3**, including the new regression; production/test compilation, Spotless, Scalastyle, Rust and changed-Markdown formatting, and the PR diff check passed. Spark 4.1 documentation generation passed; the generated extrema sections correctly describe conditional codegen dispatch.

Native execution remains unvalidated on this head: the configured Cargo registry lacks DataFusion 55.1.0, so the focused Rust command stopped during dependency resolution and ran zero tests. That also blocks rebuilding JNI and running the Spark integration fixtures locally. The description now separates these current checks from the historical native tests and performance measurements.

spark/src/test/resources/sql-tests/expressions/array/array_min.sql
sunchao: The signed-zero cases are now published as ordinary native `query` assertions rather than ignored tests, so a fallback or result mismatch fails the test. The separate non-default-collation fixtures now use `expect_dispatch(...)` with codegen enabled, following the later review request; dispatcher-disabled collation fallback is covered separately in `CometArrayExpressionSuite` with a checked fallback reason.

andygrove (approved):
LGTM. Thanks @sunchao

sunchao:
Rebased onto current Apache main [`dd68a531c`](https://github.com/apache/datafusion-comet/commit/dd68a531c413c287c1f5fbe12ade4edb488900e5); the updated head is [`ae755c459`](https://github.com/apache/datafusion-comet/commit/ae755c459e0acd95f72a53f538d959e2adec572c). Conflicts are resolved, preserving upstream benchmark registrations, imports, and documentation.

Rechecked the earlier feedback: both extrema retain `CodegenDispatchFallback` for non-default collations, explicit native/dispatch assertions, and dispatcher-disabled fallback coverage. The simplified implementation and regression fixtures are unchanged.

Fresh validation passed all **9 native extrema tests** against DataFusion **55.1.0** / Arrow **59.3.0**, plus **16 JVM parser/documentation tests**, Spark 4.1 production/test compilation, formatting, and diff checks. The prior local dependency blocker is gone. Requested Spark 4.1 SQL CI with `run-spark-4.1-tests`; those integration results are still pending. The PR description now reflects this validation.

PR #5408, opened 2026-08-21
fix: expose native memory usage to Spark

ywskycn:

Resolving the sub issue 5 in #5212


comphead (commented):
Thanks @ywskycn this looks useful, however it would be nice to find a consumer on Scala side and also check if this value accounted properly

ywskycn:
Thx @comphead . Updated with a unit test. Anything else you think we should include?

sunchao (approved):

sunchao:
Merged, thanks @ywskycn ! @comphead let me know if you have more comments on this PR. We can add follow-ups.

PR #5409, opened 2026-08-21
test: expand replace compatibility regression coverage

### LINKED ISSUE #5354: Run replace natively by default when the search string is a non-empty literal
## Summary

`replace` has a native kernel already (DataFusion `replace`), gated off behind
`spark.comet.expression.StringReplace.allowIncompatible`. Its entire incompatibility is **one input
condition**: Spark returns the input unchanged when the search string is empty, DataFusion inserts the
replacement between every character.

```
SELECT replace('hello', '', 'x')   -- Spark: hello      DataFusion: xhxexlxlxox
```

That condition is a plain emptiness check on an argument, so when the search argument is a `Literal`
(the overwhelmingly common shape) it is decidable at plan time. This is the narrowest compatibility gap
of any candidate assessed so far, and closing it needs no native code.

Assessed with the `suggest-native-expression` skill: **compatibility confidence High**, **native upside
Medium to High, measured**. Verdict: Recommended.

Prior art: [#3344](https://github.com/apache/datafusion-comet/issues/3344) and
[#4497](https://github.com/apache/datafusion-comet/issues/4497) are both closed. They reported the
divergence as a bug and it was fixed by routing `replace` through the dispatcher, which is the current
state. Neither covers making the native path the default for the safe subset. (Note for the record:
#3344 states the divergence backwards, claiming Spark produces `xhxexlxlxox`. #4497 has it right.)

## How it runs today

`object CometStringReplace extends CometScalarFunction[StringReplace]("replace") with NativeOptInAvailable`
(`spark/src/main/scala/org/apache/comet/serde/strings.scala:179`).

`getSupportLevel` returns `Compatible(nativeOptIn = ...)` unconditionally when the user has not opted
in, so **every** `replace` runs on the dispatcher, including the vast majority whose search string is a
non-empty literal and for which the native kernel is already exactly correct.

`getIncompatibleReasons()` is "Produces different results from Spark when the search string is empty",
which already names the whole gap.

## Native upside: Medium to High (measured)

Two measurements, both on this branch.

**1. The emitted kernel.** Dumped via `CometBatchKernelCodegen.generateSource` for
`StringReplace(BoundReference(0, StringType), Literal("XYZ"), Literal("QQ"))`:

```java
value_0 = CollationSupport.StringReplace.execBinary(value_1, references[0], references[1]);
...
Object utfBase_0 = value_0.getBaseObject();
if (utfBase_0 instanceof byte[]) {
  output.setSafe(i, (byte[]) utfBase_0, ..., utfLen_0);
} else {
  byte[] utfArr_0 = value_0.getBytes();     // allocation + copy
  output.setSafe(i, utfArr_0, 0, utfArr_0.length);
}
```

The important subtlety is in that output branch. `UTF8String.replace` returns `this` when there is no
match, and in the dispatcher the input is off-heap backed (`UTF8String.fromAddress`, so `base` is
`null`, not a `byte[]`). Non-matching rows therefore take the **`getBytes()` branch and allocate a copy
anyway**. Reading `UTF8String.replace` in isolation suggests non-matching rows are free. They are not.

**2. Allocated bytes per row**, measured with
`com.sun.management.ThreadMXBean.getThreadAllocatedBytes` over 8192 off-heap-backed 64-byte values,
100 iterations after 200 warmup passes, including the kernel's output-write branch:

| match density | bytes/row | ns/row |
| --- | --- | --- |
| 0% | 80.0 | 171.0 |
| 10% | 85.6 | 195.8 |
| 50% | 108.0 | 170.5 |
| 100% | 136.0 | 182.4 |

For contrast, the same measurement with heap-backed inputs (which is *not* what the dispatcher
provides) reports 0.0 bytes/row at 0% density. That difference is exactly the output-side copy above,
and it is why the off-heap variant is the one to trust.

So the allocation saving is **unconditional**, 80 to 136 bytes per row, roughly 650 KB to 1.1 MB of JVM
heap churn per 8192-row batch, none of it visible to Comet's native memory pool. A native kernel writes
straight into the Arrow output buffer, and for the pass-through case can copy buffer to buffer.

**What is not established:** the wall-clock win. Time per row is flat at roughly 170 to 195 ns and
barely moves with density, so this path is **scan-bound, not allocation-bound**. Whether native is
faster depends on whether Rust's substring search beats `UTF8String.find`'s byte-at-a-time loop, which
is plausible (memchr-accelerated) but unmeasured here because it needs a release build of the native
library. Both paths already exist, so this is cheap for the implementer: run `CometStringExpressionBenchmark`
with and without `spark.comet.expression.StringReplace.allowIncompatible=true` at several match
densities. It is in the acceptance criteria below rather than asserted here.

### Spark versions

Identical across 3.4.3, 3.5.9, 4.0.4, 4.1.3, and 4.2.0 for the default collation. 3.4.3 and 3.5.9 are
byte-identical; 4.0.4 adds `NullIntolerant` becoming an override, a `collationId` field, and routes
through `CollationSupport.StringReplace.exec`, whose `isUtf8BinaryType` branch is
`execBinary(src, search, replace) = src.replace(search, replace)`, i.e. the same call as 3.x. 4.0.4,
4.1.3, and 4.2.0 are byte-identical to each other. No shims needed.

### Hazard checklist

| Hazard | Applies? |
| --- | --- |
| Empty search string | **Yes, and this is the whole gap.** Spark short-circuits on `search.numBytes == 0` (`UTF8String.java`) and returns the input. DataFusion inserts between every character. Decidable at plan time for a literal search argument. |
| Collation on Spark 4.0+ | **Yes.** Non-UTF8_BINARY collations route to `execLowercase` / `execICU`, which the native path does not implement ([#4496](https://github.com/apache/datafusion-comet/issues/4496)). Out of subset, and the collation is on the expression's type, so this is plan-time detectable. |
| Empty **source** string | No. Spark returns the input; DataFusion's replace over an empty haystack also produces empty. Same result. |
| JVM formatting / locale / ICU | No. Byte-level search and splice, no locale involvement. |
| Invalid UTF-8 | No more than any other string expression; the search and replace are both byte sequences and Comet's ingress policy already applies. |
| Regex features, decimal, timezone, ANSI errors, lambdas, nondeterminism, host state | No |

Unlike `upper`/`lower` ([#5353](https://github.com/apache/datafusion-comet/issues/5353)), the guard here
is an **expression** property rather than a **data** property, so it needs no per-batch fallback
mechanism and is expressible in today's architecture.

## Proposed approach

No native code. This is a `getSupportLevel` change plus tests.

1. In `CometStringReplace.getSupportLevel`, return `Compatible(None)` when the search argument is a
   non-empty `Literal` **and** the collation is UTF8_BINARY. That case takes the native path by default.
2. Everything else keeps today's behaviour exactly: an empty literal search, a non-literal search, or a
   non-default collation returns `Compatible(nativeOptIn = ...)` and rides the dispatcher.
3. Keep `getIncompatibleReasons()` as is, since it still describes the opt-in path for the cases that
   stay on the dispatcher, and add a compatible note documenting the new default.
4. Consider a follow-up for the non-literal search case: a native kernel could implement Spark's
   short-circuit directly (`if search.is_empty() { return src }`), which would make the whole expression
   compatible rather than just the literal subset. That is a small kernel change in
   `datafusion-comet-spark-expr` (or upstream), and it would be the cleaner end state. Filing this as
   step 4 rather than step 1 because the plan-time guard delivers most of the value with no native
   change at all.

### Non-goals

- Collation support. Non-default collations stay on the dispatcher.
- Changing the `allowIncompatible` config surface.

## Acceptance criteria

- `replace(col, 'nonempty', r)` runs natively by default, asserted via the absence of the expression
  from the `[COMET-INFO: JVM codegen dispatcher: ...]` EXPLAIN segment.
- `replace(col, '', r)`, a non-literal search, and non-default collations demonstrably still route to
  the dispatcher, each with a test.
- A Comet SQL Test covering empty search, empty source, no-match, single match, multiple matches,
  overlapping candidates, and multi-byte UTF-8 values.
- `CometStringExpressionBenchmark` numbers, dispatcher versus native, at 0%, 10%, and 100% match
  density, in the PR description. If native turns out **not** to be faster despite the allocation
  saving, say so in the PR and let reviewers decide whether the reduced heap churn alone justifies the
  change.
- The `replace` entry in `docs/source/contributor-guide/expression-audits/string_funcs.md` updated.

---

Assessment recorded in `docs/source/contributor-guide/expression-audits/string_funcs.md` under
`## replace`. Earlier runs of the same skill produced #5347, #5349, #5351, and #5353.

sam-1112:

Related to #5354. This PR does not close it: default-native `replace` is deferred until the import-boundary work in #5310 has landed and safe native eligibility can be reassessed.

## Rationale for this change

#5354 proposed making native DataFusion `replace` the default when `search` is a non-empty `UTF8_BINARY` literal. Review of the earlier native-default attempt found expression-boundary mismatches that are not resolved by checking only whether `search` is empty:

- malformed string literals can change during native serialization;
- Spark's NULL short-circuit can differ from eager native child
  evaluation; and
- broadcasting large scalars into Arrow `Utf8` arrays can overflow
  32-bit offsets.

The earlier `BatchSplitExec` / `UnnestExec` changes were split into #5559. That PR was later closed as superseded by the more complete fix merged in #5362. 

This PR retains the current routing: `replace` uses the JVM codegen dispatcher by default, while native DataFusion `replace` remains available through `spark.comet.expression.StringReplace.allowIncompatible=true`.

The remaining changes provide compatibility and regression coverage so that a future native-default implementation cannot omit these boundary cases.

Tests only; there are no production-code or routing changes.

SQL result-parity coverage in `string_replace.sql` includes:

- overlapping matches (`replace('aaaa', 'aa', 'x')` → `'xx'`);
- multibyte searches and replacements, including CJK and emoji; and
- a replacement containing the search string, verifying that replacement is not applied recursively.

`CometCodegenSuite` adds result-parity and dispatcher tripwires for:

- malformed source, search, and replacement literals;
- a malformed literal nested inside the source expression;
- Spark's ANSI NULL short-circuit behavior with a throwing replacement;
- a throwing expression nested inside the source; and
- oversized source and replacement expressions.

These cases assert `JVM codegen dispatcher: replace`. Because every `replace` currently uses the dispatcher by default, this is a guard against unsafe future eligibility changes rather than a current per-case routing distinction.

The focused SQL and codegen tests pass on Spark 3.5 and Spark 4.1:

```bash
./mvnw test -Dtest=none \
  -Dsuites="org.apache.comet.CometSqlFileTestSuite string_replace"

./mvnw test -Dtest=none \
  -Dsuites="org.apache.comet.CometCodegenSuite replace compatibility boundary cases stay on JVM codegen dispatcher"
```


spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -178,29 +179,56 @@ object CometStringReplace
...
+   * The native kernel is also byte-level `UTF8_BINARY` only, so non-default collations stay on
+   * the dispatcher. https://github.com/apache/datafusion-comet/issues/4496
+   */
+  private def nativeSafeSearchSubset(expr: StringReplace): Boolean = {
+    val children = expr.children
+    if (children.length != 3) {
+      return false
+    }
+    val searchIsNonEmptyLiteral = children(1) match {
+      case Literal(v: UTF8String, _) => v != null && v.numBytes() > 0
```
sunchao: [P2] Exclude malformed search literals from the native-safe subset

Could this guard also reject search literals whose bytes change during native serialization? With a valid Parquet source column containing U+FFFD (`EF BF BD`), `replace(s, CAST(X'FF' AS STRING), 'x')` returns the source unchanged in Spark and the base dispatcher because byte `FF` is absent. Catalyst folds the cast to a non-empty `UTF8String`, so this check accepts it, but `CometLiteral` serializes it through `UTF8String.toString`, changing the search to U+FFFD. The head's native path consequently returns `x`. I reproduced this with `allowIncompatible=false` and entirely valid scan data. Please retain dispatcher routing for malformed search literals unless their byte semantics can be preserved natively.

spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -178,29 +179,56 @@ object CometStringReplace
...
       inputs: Seq[Attribute],
       binding: Boolean): Option[Expr] = {
-    if (CometConf.isExprAllowIncompat(getExprConfigName(expr))) {
-      // The native DataFusion `replace` avoids the JVM allocations of the codegen
-      // dispatcher but is not Spark-compatible for an empty search string, so it is
-      // only used when incompatibility is explicitly allowed.
+    if (CometConf.isExprAllowIncompat(getExprConfigName(expr)) || nativeSafeSearchSubset(expr)) {
+      // Native DataFusion `replace` matches Spark when search is a non-empty UTF8_BINARY
+      // literal (the common case, selected by default) and when the user has opted in.
       super.convert(expr, inputs, binding)
```
sunchao, line 210: [P2] Preserve NULL short-circuiting for replacement expressions

With ANSI enabled and Parquet rows `(s=NULL, n=0)` and `(s='a', n=1)`, `SELECT replace(s, 'a', CAST(1 / n AS STRING)) FROM t` succeeds in Spark and the base dispatcher, returning NULL and `'1.0'`. This native conversion instead raises `DIVIDE_BY_ZERO` with `allowIncompatible=false`. Spark's ternary expression skips the replacement when the source is NULL, whereas the native scalar-function expression evaluates every child for the batch before `replace` receives the source null mask. Could the native eligibility check account for this conditional evaluation, or retain dispatcher routing when the replacement can throw? A nullable-source/erroring-replacement regression would protect this behavior.

spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -178,29 +179,56 @@ object CometStringReplace
...
 
   override def convert(
       expr: StringReplace,
       inputs: Seq[Attribute],
       binding: Boolean): Option[Expr] = {
-    if (CometConf.isExprAllowIncompat(getExprConfigName(expr))) {
-      // The native DataFusion `replace` avoids the JVM allocations of the codegen
-      // dispatcher but is not Spark-compatible for an empty search string, so it is
-      // only used when incompatibility is explicitly allowed.
+    if (CometConf.isExprAllowIncompat(getExprConfigName(expr)) || nativeSafeSearchSubset(expr)) {
```
sunchao: [P2] Avoid broadcasting large literals in the new default path

For an 8,192-row Parquet batch containing only short strings `'a'`/`'b'`, `SELECT replace(s, 'notfound', repeat('x', 262144)) FROM t` should return those inputs unchanged. The base dispatcher does, but the head's native Comet projection fails with `CometNativeException: native panic: offset overflow`, even though the output is tiny. Pinned DataFusion 54.1.0 broadcasts every `Utf8` scalar before matching, so the unused 256 KiB replacement becomes a 2 GiB array and exceeds Arrow's i32 offsets at the default Comet batch size. A large search literal triggers the same problem. Please use scalar-aware native execution or retain dispatcher routing for these cases instead of enabling them solely from a non-empty search.

sunchao (commented):
## Summary

Reviewed head `3b0d9d405b477218fa6c664919f020ec1ae2bc99` against base `f8f9b3356d29e776e02a9fca7a6b3c630790db2c` with five independent review scopes. The patch keeps the implementation small and reuses the existing native kernel, but three P2 regressions survive verification: malformed search literals can change results, NULL-source rows can raise previously suppressed replacement-expression errors, and large scalar literals can overflow during batch broadcasting. All three reproduce with `spark.comet.expression.StringReplace.allowIncompatible=false`.

## Prior state and problem

Previously, the default `StringReplace` path serialized the complete Spark expression into the JVM codegen dispatcher. Native DataFusion execution required explicit incompatibility opt-in because empty-search behavior differs from Spark. Issue #5354 proposes recognizing the non-empty-literal subset at planning time to avoid dispatcher-side JVM allocations.

That motivation is understandable, and ordinary valid-UTF-8 replacement semantics agree. The compatibility boundary also includes literal serialization, conditional evaluation of children, and the scalar-to-array adaptation performed before the native kernel runs. Those boundaries explain the three findings even though this PR does not change the kernel itself.

## Design approach

`nativeSafeSearchSubset` requires three children, a non-empty `UTF8String` search literal, and no non-default string collation on any child. Both `getSupportLevel` and `convert` use this predicate, so eligible expressions become native by default while empty, non-literal, and collated searches retain dispatcher routing.

The replacement remains unrestricted and may be a column or expression. Explicit `allowIncompatible=true` still selects the existing native route for the other cases. The collation check is conservative and uses the existing version shims rather than introducing Spark-version-specific logic here.

The pinned DataFusion 54.1.0 UDF matched 9,000 seeded valid-UTF-8 cases against each of Spark 3.4.3, 3.5.8, 4.0.4, and 4.1.3. The corpus included NULL source/replacement, empty source/replacement, overlaps, multibyte and combining characters, NUL bytes, and replacement columns. A focused Spark 4.0.4 probe also checked parsing, routing, and interpreted/generated agreement for UTF8_BINARY, UTF8_LCASE, and UTF8_BINARY_RTRIM.

For the regressions, I separately compiled the exact base and head versions of `strings.scala` and ran controlled A/B SQL probes on the same cached Spark 4.0.4/Comet support runtime. Both sides executed Comet projections, with the base retaining the dispatcher and the head taking the native route. The malformed-literal wrong result, ANSI division-by-zero failure, and unused-large-literal overflow all appeared only with the head serde. Actual pinned DataFusion/Arrow probes independently corroborated literal replacement and broadcasting behavior. This is focused serde/kernel validation, not a fresh full-head Maven/native build or a full test-suite run.

The exact-head [CI workflow](https://github.com/apache/datafusion-comet/actions/runs/32502902238), Delta build gate, CodeQL, and PR-title workflows currently report `action_required`. They provide no executed CI test evidence for this head. The test runs and performance numbers in the PR description remain author-reported.

## Key design decisions

Reusing the existing kernel avoids a second replacement implementation, and retaining dispatcher behavior outside the intended subset limits the change. Leaving the incompatibility opt-in intact also preserves the existing configuration surface. However, search shape alone does not establish that the complete expression is compatible with native execution.

The disclosed performance trade-off is explicit: native takes 2.5–2.9 times the dispatcher timing in the supplied benchmark, while eliminating dispatcher-side JVM string allocations. The allocation estimate was not remeasured in that benchmark. I have not treated the disclosed slowdown itself as an additional finding or inferred a production throughput or total-memory improvement from those numbers.

## Implementation sketch

The introduced PR diff changes four files: the serde predicate and routing, the replacement audit entry, SQL cases, and the codegen routing test. The existing scalar-function registration and three-string DataFusion signature already provide the required native plumbing, so no new registration is missing.

I also inspected all nine paths in the direct base-to-head tree diff. The other five paths belong to newer base-only commits for signed-zero tests and workflow dependencies, not reversions introduced by this PR. The added SQL tests run through a harness that excludes constant folding, while the Scala routing test uses a column source and dispatcher annotations to distinguish execution paths.

For the newly native subset, Spark no longer receives the whole expression tree in a serialized dispatcher closure. String literals pass through native literal serialization, replacement expressions are evaluated eagerly as native children, and scalar strings are expanded for the batch before matching. The inline reproducers show observable changes at each of these boundaries.

These are default-behavior regressions rather than changes confined to users who opted into incompatibility. Ordinary replacement-column, overlap, and valid-Unicode controls continued to match, so the findings do not argue against native execution for a more carefully bounded subset.

## Suggested improvements

Please preserve byte-level behavior for malformed literals and Spark's NULL short-circuiting for replacement expressions before selecting the native default. The native path also needs scalar-aware handling, or dispatcher retention for affected calls, so a no-match replacement cannot overflow merely from broadcasting an unused literal.

Add targeted regressions for a valid U+FFFD source with a malformed search literal, a NULL source with an ANSI-erroring replacement expression, and a full batch with a large unused search/replacement literal. These tests should assert both the result or expected error behavior and the intended execution path. The three inline comments contain concrete reproducers and the relevant boundaries to address.

spark/src/main/scala/org/apache/comet/serde/strings.scala
sam-1112: Agreed — kernel compatibility is not enough if `CometLiteral` changes the search bytes. `CAST(X'FF' AS STRING)` folds to a non-empty `UTF8String`, so the old “non-empty literal” check accepted it. Serialization then went through `UTF8String.toString` and the search became U+FFFD (`EF BF BD`), which incorrectly matched a well-formed U+FFFD in the source. The default native-safe subset now rejects a string literal unless its bytes survive that round-trip:

```scala
Arrays.equals(v.getBytes, v.toString.getBytes(StandardCharsets.UTF_8))

File: spark/src/main/scala/org/apache/comet/serde/strings.scala (hunk 2 above)
Comment (sam-1112) [reply spark/src/main/scala/org/apache/comet/serde/strings.scala:210]: Agreed. Spark's ternary `eval` / `doGenCode` skips the replacement when `src` is NULL, so this query returns `NULL` and `'1.0'` under ANSI. The native path evaluates every child for the batch first, so `1 / 0` still runs and raises `DIVIDE_BY_ZERO`. I am not trying to prove in general whether an arbitrary replacement can throw. The default native-safe subset now only accepts a replacement that is a short well-formed literal, a null literal, or a column (`Attribute` / `BoundReference`). `CAST(1 / n AS STRING)` is none of those, so it stays on the dispatcher. `CometCodegenSuite` covers the reproducer: Parquet rows `(NULL, 0)` and `('a', 1)`, ANSI on, `replace(s, 'a', CAST(1 / n AS STRING))`. The result matches Spark and EXPLAIN still shows `JVM codegen dispatcher: replace`.

File: spark/src/main/scala/org/apache/comet/serde/strings.scala (hunk 3 above)
Comment (sam-1112) [reply spark/src/main/scala/org/apache/comet/serde/strings.scala]: Makes sense — we shouldn't native this just because the search is a non-empty literal. I didn't change the kernel to be scalar-aware. Until we can apply a big Utf8 scalar without broadcasting it, these stay on the dispatcher. Literals are capped at `Int.MaxValue / spark.comet.batchSize` (262,143 bytes at the default 8,192). `repeat('x', 262144)` is over that if it folds, and if it doesn't fold it isn't a literal/column anyway, so either way we dispatch. Large search literals get the same cap.

Test: `replace(s, 'notfound', repeat('x', 262144))` returns the original strings and stays on the dispatcher.

File: spark/src/main/scala/org/apache/comet/serde/strings.scala [hunk 4]
```diff
@@ -178,29 +182,84 @@ object CometStringReplace
...
+    }
+    val replacementIsSafe = children(2) match {
+      case Literal(null, _) => true
+      case Literal(v: UTF8String, _) => isNativeSafeStringLiteral(v, allowEmpty = true)
+      case _: Attribute | _: BoundReference => true
+      case _ => false
+    }
+    val utf8BinaryCollation =
+      !children.exists(c => QueryPlanSerde.isStringCollationType(c.dataType))
+    utf8BinaryCollation && searchIsSafe && replacementIsSafe
```
sunchao: [P2] Check the source subtree before declaring the call native-safe

Could this eligibility check cover `children(0)` as well? Under ANSI, with Parquet rows `(s=NULL,n=0)` and `(s='a',n=1)`, `replace(substring(s, 1, CAST(1 / n AS INT)), 'a', 'x')` still passes this guard and throws `DIVIDE_BY_ZERO`. The base dispatcher returns NULL and `'x'`, because Spark's `Substring` skips the length expression on the NULL row. Native conversion recursively evaluates that unchecked source subtree instead.

Source literals bypass the new byte/size checks too. With a replacement column `r='x'`, `replace(CAST(X'FF' AS STRING), 'a', r)` changes raw `FF` into `EF BF BD`. A 256 KiB folded source with an empty replacement column still overflows when broadcast to 8,192 rows. These all reproduced in the base/head A/B with `allowIncompatible=false`. Checking only a top-level source literal would miss malformed literal descendants such as `concat(CAST(X'FF' AS STRING), r)`, so please retain dispatcher routing unless the source subtree is safe too.

sunchao (commented):
Follow-up review of `d223e8f1f90d15abb594eec941b917a1cd2fa780` against `367ca64be4e62dd10dc4c932de9ca738355a45fa`, after five fresh independent review scopes. The three original concrete reproducers now pass. The two focused routing tests and the 11-query SQL fixture also pass. One unique P2 is attached for the unchecked source subtree, grouping its byte-preservation, conditional-evaluation, and scalar-size symptoms.

The [existing overflow finding](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3832146723) is not fully addressed by the new batch-size cap. With an 8,192-row Parquet input of short strings, `SELECT replace(e, 'notfound', repeat('x', 131072)) FROM t LATERAL VIEW explode(array(s, s)) a AS e` passes the new literal limit, but native `explode` emits a 16,384-row batch. Broadcasting the unused replacement then still fails with `native panic: offset overflow`. The base dispatcher returns all 16,384 unchanged short strings. This is further evidence for the existing unresolved thread, not a second new inline finding. The cap needs a genuine runtime batch bound or scalar-aware handling.

Verification used controlled A/B runs with the exact base/head serde compiled onto the same cached Spark 4.0.4 support runtime. Fresh pinned DataFusion 54.1.0/Arrow 58.4.0 probes independently corroborated the overflow paths, and 30,000 fresh valid-UTF-8 cases matched across Spark 3.4/3.5/4.0/4.1. No fresh full-head build or full-suite run was performed.

spark/src/main/scala/org/apache/comet/serde/strings.scala
sam-1112: Thanks for catching this. I updated the eligibility check so `children(0)` uses the same conservative whitelist as the replacement argument: only a column reference, `NULL`, or a short byte-preserving UTF-8 literal is accepted.

Nested source expressions now remain on the dispatcher, so the `substring` short-circuit case and malformed literals hidden under `concat` are not converted recursively to native execution.

I also added regression coverage for:
- the ANSI `substring` / `DIVIDE_BY_ZERO` case,
- a malformed source literal,
- a malformed literal nested under `concat`, and
- an oversized folded source literal.

These cases now preserve Spark results and use the JVM codegen dispatcher with `allowIncompatible=false`.

spark/src/main/scala/org/apache/comet/serde/strings.scala
sunchao: [P2] Account for batches expanded by native `explode`

This still reproduces on `db8e31d2cb9d1d63798fb26cbf7318920aecd394`. The literal cap uses `spark.comet.batchSize`, but native `explode` can emit a larger batch without rebatching it.

With an 8,192-row Parquet input containing only short strings `'a'`/`'b'`, the default batch size of 8,192, and `spark.comet.expression.StringReplace.allowIncompatible=false`:

```sql
SELECT replace(e, 'notfound', repeat('x', 131072))
FROM t
LATERAL VIEW explode(array(s, s)) a AS e
```

The 131,072-byte replacement passes the 262,143-byte cap. However, `explode` produces one 16,384-row batch, so broadcasting the unused replacement would require 2,147,483,648 bytes and fails with `CometNativeException: native panic: offset overflow`. The exact-base dispatcher returns all 16,384 unchanged strings. Both A/B plans contain `CometProject` and `CometExplode`, with only the head routing `replace` natively.

Could this use scalar-aware handling or an enforced runtime batch bound before enabling these calls by default? Please add a regression after `explode`, since the one-row oversized-literal test checks routing but cannot establish this bound.

Verification used freshly compiled base (`367ca64be4e62dd10dc4c932de9ca738355a45fa`) and head `strings.scala` on the same cached Spark 4.0.4/Comet support runtime. This was focused A/B validation, not a fresh full-head Maven/native build.

sam-1112:
Thanks for the follow-up and the detailed reproducer. I’ve addressed both remaining boundaries in the latest commit.

For the source subtree, native eligibility now only accepts a column reference, `NULL`, or a short byte-preserving UTF-8 literal. Nested expressions such as `substring` and `concat` remain on the dispatcher. I added regressions for the ANSI short-circuit case, malformed source literals (including under `concat`), and oversized source literals. For the `explode` overflow, native `UnnestExec` is now wrapped in `BatchSplitExec`, which uses the runtime `TaskContext` batch size to split oversized output before downstream projections. The exact 8,192 → 16,384 row reproducer with the 131,072-byte replacement now passes. The regression test verifies all 16,384 results, confirms the plan uses `CometExplodeExec`, and confirms `replace` remains native. I also added a Rust unit test for runtime batch splitting. Both focused Rust and Spark 4.1 tests pass.

native/core/src/execution/planner.rs
```diff
@@ -2116,11 +2116,20 @@ impl PhysicalPlanner {
...
+                    Arc::new(BatchSplitExec::new(unnest_exec));
 
                 Ok((
                     scans,
                     shuffle_scans,
-                    Arc::new(SparkPlan::new(spark_plan.plan_id, unnest_exec, vec![child])),
+                    Arc::new(SparkPlan::new(
+                        spark_plan.plan_id,
+                        bounded_unnest,
+                        vec![child],
```
sunchao: [P2] Preserve the explode metrics through the new wrapper

Replacing `UnnestExec` as the `SparkPlan` root also drops its SQL metrics. `to_native_metric_node` reads this root's `metrics()` and then its `SparkPlan` children; it never visits the wrapped native child. `BatchSplitExec::metrics()` returns only `batches_split`, which `CometExplodeExec` does not expose, so its existing input/output row and batch counters and `elapsed_compute` now remain zero for every native explode, even when no split is needed. A focused Spark 4.0.4 probe over 8,192 short-string rows returned all 16,384 exploded rows, but all five metrics were zero with the matching PR native artifact; the control runtime reported `input_rows=8192` and `output_rows=16384`. The exact-source Rust probe independently confirms that the populated unnest metrics are hidden by the wrapper. Please preserve those metrics, including `output_rows` (the existing additional-plan aggregation deliberately skips that field), and add a metric regression check.

sam-1112: Thanks for catching this. I kept `BatchSplitExec` as the execution root, but made it transparent to metric collection: `BatchSplitExec::metrics()` now returns its own split metrics together with all metrics from the wrapped `UnnestExec`.

This preserves `input_rows`, `output_rows`, `input_batches`, `output_batches`, and `elapsed_compute`. It does not use `additional_native_plans`, so `output_rows` is not dropped by that path’s deliberate filtering. The wrapper’s `batches_split` metric is retained as well.

I added two regression checks:

- The Spark reproducer with 8,192 input rows expanding to 16,384 rows now asserts `input_rows == 8192`, `output_rows == 16384`, and non-zero input/output batch counters on `CometExplodeExec`.
- The Rust unit test wraps an input with an `output_rows` metric, executes it through `BatchSplitExec`, and verifies that `BatchSplitExec::metrics()` exposes `output_rows == 10`.

The focused Rust test and the Spark 4.1 reproducer both pass.

sunchao:
Follow-up on `a96957cce86116427e6a056e878807c9c0bf7f05` after the five-agent re-review: [P2] two existing issue families remain partially unresolved in residual cases—the [source-subtree safety thread](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3836632726) and the [large-literal/runtime batch-expansion thread](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3832146723). The [explode-metrics issue](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3839055178) is fixed, and no unique new P1/P2 was found. This comment is only a current-head status reminder; I am not reposting the summary or the existing inline findings.

andygrove:

The core idea, admit a conservative literal subset to the native kernel by default rather than gating the whole expression behind `allowIncompatible`, is a good one, and the reasoning in the `nativeSafeSubset` doc comment about why kernel compatibility alone is not sufficient (literal round-tripping, eager child evaluation, offset overflow) is unusually clear.

Four things.

**`BatchSplitExec` does not belong in this PR**

174 lines of new operator plus a planner change that wraps every `UnnestExec` in a `BatchSplitExec`. That is a runtime change affecting every `explode` query in Comet, and it has nothing to do with `replace` except that the offset-overflow reasoning happens to reference it.

It needs its own review: what does the extra stream wrapper cost on the explode path, does the `SplitMetrics` show up in the SQL UI, and is the DataFusion 54.1.0 diagnosis right? None of that will get proper attention buried in a string-expression PR. Could you split it out and make this one depend on it?

**Columns can also contain malformed UTF-8**

`isNativeSafeStringArg` accepts any `Attribute` or `BoundReference`, and the doc comment justifies excluding nested expressions on the grounds that "a throwing or malformed child cannot hide inside the source tree". But a plain column read from Parquet can absolutely contain bytes that are not valid UTF-8, which is the whole subject of #4764 and #5310.

What happens when the native `replace` kernel gets a column with invalid UTF-8? If the answer is "the same thing as every other native string kernel, and #5310 fixes it at the import boundary", then say so and the concern goes away. If the answer is that this expression is newly exposing that path by default where it previously went through the dispatcher, that is worth knowing before flipping the default.

**The size cap is a heuristic with a sharp edge**

```scala
val maxBytes = Int.MaxValue / math.max(CometConf.COMET_BATCH_SIZE.get(), 1)
```

This assumes the batch reaching the kernel is at most `spark.comet.batchSize` rows, which the comment acknowledges is only true because of `BatchSplitExec` on the explode path. Are there other native operators that can emit an oversized batch? A join, an unnest inside a different shape, or a scan with a large row group? If any can, the cap is unsound and the failure mode is an Arrow offset overflow at runtime rather than a fallback.

Also, `COMET_BATCH_SIZE` is read at plan time. If someone changes it between planning and execution, or if AQE re-plans, does the cap still hold? A comment on why plan-time is sufficient would help.

**Test coverage for the newly-default path**

`string_replace.sql` gains 15 lines. Since this flips the default for a whole class of queries, I would want to see the SQL fixture cover the cases where a naive `replace` diverges from Spark: overlapping matches (`replace('aaa', 'aa', 'b')`), a replacement containing the search string, multibyte search and replacement, a search longer than the source, and a null source with a non-null search. Are those covered somewhere else already? The 193 lines added to `CometCodegenSuite` look like they are about dispatch routing rather than result correctness.

sunchao:
@andygrove, a few clarifications against `a96957cc` for [these questions](https://github.com/apache/datafusion-comet/pull/5409#issuecomment-5441723664):

- `BatchSplitExec` addresses the [previously demonstrated post-explode replacement overflow](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3837642570), so there is a dependency even though its effect is broader than `replace`. The [current wrapper forwards Unnest metrics](https://github.com/apache/datafusion-comet/blob/a96957cce86116427e6a056e878807c9c0bf7f05/native/core/src/execution/operators/batch_split.rs#L102); its own `batches_split` counter is not declared in the SQL operator metrics. It bounds downstream batch sizes, not peak memory while Unnest constructs the expanded batch. Wrapper overhead remains unmeasured in this follow-up.
- The [codegen tests check Spark result parity](https://github.com/apache/datafusion-comet/blob/a96957cce86116427e6a056e878807c9c0bf7f05/spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala#L393), not only routing. The [SQL fixture](https://github.com/apache/datafusion-comet/blob/a96957cce86116427e6a056e878807c9c0bf7f05/spark/src/test/resources/sql-tests/expressions/string/string_replace.sql#L45) includes overlap, multibyte search, searches longer than source rows, and NULL-source cases. I did not find explicit multibyte-replacement or replacement-containing-search cases in the changed fixtures.
- The malformed-column and general batch-cap questions are not resolved merely by the literal guard. [#5310](https://github.com/apache/datafusion-comet/pull/5310) remains open and unmerged; its import protection cannot be assumed present. I also did not establish a global cap across every producer/configuration/AQE transition. Those concerns overlap the existing source-safety/overflow discussions, rather than establishing another verified finding here.

This was a source/discussion check; I did not rerun runtime tests or benchmarks.

sam-1112:
Thanks Andy and Chao for the review and clarifications.

I agree that this PR has grown beyond the original `replace` routing change. The malformed-column path and the global runtime batch bound are not established well enough to call the current default-native subset safe.

I’ll do the following next:

1. Split `BatchSplitExec`, the `UnnestExec` integration, metric forwarding and SQL UI behavior, and wrapper-overhead evaluation into a separate PR. That runtime change needs an independent review regardless of what remains in this PR.

2. I will not assume that #5310 is already protecting column inputs. Until an import-boundary fix is merged, column source and replacement arguments will remain on the dispatcher. Default-native `replace` for column inputs can be reconsidered in a separate PR after that protection has landed.

3. I will remove the `Int.MaxValue / spark.comet.batchSize` eligibility heuristic. Without an enforced runtime batch bound, broadcasting a literal into a `Utf8` array cannot be considered safe. Combined with keeping column arguments on the dispatcher, this means that I will not enable native `replace` by default in this PR. I am not planning to make the kernel scalar-aware here.

4. I’ll add explicit SQL result-parity cases for:
   - a multibyte replacement; and
   - a replacement containing the search string.

The earlier regressions involving malformed literals, NULL short-circuiting, nested source expressions, the post-`explode` reproducer, and forwarded `UnnestExec` metrics pass on the current head, and the current CI is green. I’ll leave the source-safety and runtime-batch threads open until `BatchSplitExec` has been split out and the default-native change has been removed from this PR.

After splitting out `BatchSplitExec`, I’ll keep this PR only for the `replace` test coverage and for dropping the unsafe default-native subset. If nothing remains after that cleanup, I’ll close it and revisit default-native `replace` after the import-boundary and runtime batch-bound work has landed.

sam-1112:
`BatchSplitExec` and the `UnnestExec` runtime-boundary changes have now been split into #5559 for independent review.

Next, I’ll update this PR to remove those runtime changes and drop the unsafe default-native `replace` subset. The remaining changes will focus on `replace` compatibility and regression coverage.

sam-1112:
The cleanup is ready for another look.

This PR is now tests-only. I removed the default `replace` routing changes and the `BatchSplitExec` / `UnnestExec` runtime changes. It now just adds the requested multibyte-replacement and replacement-containing-search cases, along with the expression-boundary regression tests.

sunchao (approved):
Re-reviewed `47714ee29d1901d59ae02b45a7c533b681e0701d` against `392da2ca7a99fd223644eff650406f61580eb46b`, following the review at `80c0737c` and the later discussion.

The [current diff](https://github.com/apache/datafusion-comet/pull/5409/files) is tests-only. Removing the default-native eligibility change and the BatchSplit/Unnest changes withdraws the code behind the earlier P2 findings, including the [source-safety](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3836632726), [batch-bound](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3837642570), and [metrics](https://github.com/apache/datafusion-comet/pull/5409#discussion_r3839055178) concerns. Those no longer block this contribution. This does not establish safety for a future default-native implementation.

The remaining tests cover multibyte replacement and nonrecursive replacement, plus Spark result parity and dispatcher selection at the expression boundaries discussed earlier. The one-row large-literal cases guard routing. They do not establish a runtime batch-size bound. I found no new or remaining P1/P2 issue in the current two-file diff.

Validation: source and test-harness review, comparison with maintained Spark 3.5/4.0 semantics, and a clean diff check. I did not run the Scala suites locally or independently verify the author’s reported runs. [Current CI](https://github.com/apache/datafusion-comet/actions/runs/34453572020) requires approval and has no executed jobs. The successful labeling check provides no test-execution evidence.

sam-1112:
Thanks for the re-review!

andygrove (approved):
Tests only, and they land on exactly the boundary cases that came up in the earlier review. The dispatcher tripwires are the useful part, since they will fail loudly if someone widens native eligibility later without revisiting malformed literals, NULL short-circuiting or the broadcast size. LGTM.

sam-1112:
The assertion fix is now pushed. The new CI workflows are waiting for maintainer approval. Could someone approve them when convenient? Thanks!!

sunchao (approved):
Re-reviewed `9dcdeba8` against base `3810936b`. The assertion fix now checks `getCodegenDispatchExpressions(plan).contains("replace")`, so it still detects dispatcher routing when other expression names precede `replace` in explain output. Spark result comparison and the Comet operator check remain intact.

The contribution remains tests-only. The earlier P2 findings remain withdrawn with the removed production changes. I found no new or remaining P1/P2 issue in the current diff.

The [Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/34706096488/job/103592149830) and [Spark 4.1](https://github.com/apache/datafusion-comet/actions/runs/34706096488/job/103592149826) jobs both passed `string_replace.sql` and the dispatcher-boundary test. Their checkout `332fb9e6` has parents exactly this base and head and a tree identical to the reviewed head. I did not run separate local Scala suites.

CI is still red at **2026-09-12 21:10:43 UTC**: the [Rust job](https://github.com/apache/datafusion-comet/actions/runs/34706096488/job/103588176248) failed `cancelling_abort_keeps_the_guard_armed` because its first poll completed instead of yielding. It stopped after 146 passed and one failed test, leaving 1,268 tests unrun. The native code and workflow trees are identical to the base, so this failure is outside the authored test changes. Required Checks remains failed. Maintained Spark 3.4/4.1 source gaps remain.

sunchao (commented):
CI follow-up on unchanged head `9dcdeba8` and base `3810936b`: the [Rust rerun](https://github.com/apache/datafusion-comet/actions/runs/34706096488/job/104410449716) passed all 1,415 selected tests, including `cancelling_abort_keeps_the_guard_armed`, with five tests skipped. [Required Checks](https://github.com/apache/datafusion-comet/actions/runs/34706096488/job/104418832159) now passes. The current check summary is 40 successful and 11 skipped.

The original failed run and this passing rerun both checked out `332fb9e6`, whose parents are exactly this base/head and whose tree equals the reviewed head. The cancellation assertion, native code, dependency lock and workflows are unchanged. This establishes a passing rerun, but does not identify or fix the cause of the earlier first-poll assertion failure. The newer merge preview `5574f00f` was not the tested checkout.

The two-file test contribution is unchanged, the earlier P2 findings remain withdrawn with the removed production changes, and I found no new or remaining P1/P2. My [existing approval](https://github.com/apache/datafusion-comet/pull/5409#pullrequestreview-5188141417) stands. No separate local product tests were run.

PR #5412, opened 2026-08-22
fix: distinguish reflection failure from absent accessor in IcebergReflection

### LINKED ISSUE #5258: IcebergReflection: Option helpers cannot distinguish an absent accessor from a failed reflective call

`IcebergReflection`'s `Option`-returning helpers collapse two very different outcomes into `None`:

- the Iceberg version on the classpath does not declare this accessor, which is a normal, expected version difference; and
- the accessor exists but the reflective call threw, which is a real failure.

`extractFileLocation` (`spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala:266-280`) is the clearest case. It probes for `location()`, falls back to `path()`, and wraps the whole thing in `catch { case _: Exception => None }`, so a `None` means either "this is Iceberg < 1.7" or "the invoke blew up". `getFileFormat` (`:298-309`) has the same shape, and the pattern repeats across the object.

That matters because the two paths that call these want opposite things. `CometScanRule` can fall back to Spark with a message, so a quiet `None` is fine there. Serde runs after the plan is committed to the native scan, so a `None` that was really a failure has to fail the query. Today the serde callers cannot tell, so they either turn every `None` into a throw (`CometIcebergNativeScan.scala:288`, `:1027-1031` — correct, but the error cannot say what actually went wrong) or they pick a default and carry on.

This was raised in review on #5222: as long as the reflection layer can report to callers *why* they might get `None` back, the loud-at-serde-time / quiet-at-planning-time split can be enforced properly.

Give the helpers a way to distinguish the two. Options worth weighing:

- Return `Either[ReflectionFailure, Option[T]]`, or a small ADT with `Found` / `NotDeclared` / `Failed` cases, so callers pattern-match on the reason.
- Keep `Option` for "not declared" and let genuine invocation failures propagate as exceptions, so the quiet planning-path callers catch them explicitly and serde callers get them for free. Narrower change, and it matches what `findMethod` already does after #5222: it returns `None` only for `NoSuchMethodException` and lets everything else through.

The second is the smaller diff and fits the existing grain, but it moves the burden onto every planning-path caller to add a catch. Worth deciding before the two companion issues are fixed, since both of them need this distinction to do the right thing.

## Additional context

Companion issues cover the delete-file fields that fall back to wrong defaults on reflection failure, and the schema/metadata helpers that return empty collections. This one is the enabler for both: neither can cleanly separate "genuinely absent" from "reflection failed" until the helpers report it.

Note that the method cache added in #5222 does not itself introduce this ambiguity. `findMethod` caches a miss only for `NoSuchMethodException`, which is the legitimate version-difference signal; the conflation is in the higher-level helpers and predates that change.

Related: #5256 (delete-file defaults), #5257 (schema/metadata helpers returning empty collections).

unikdahal:

Closes #5258.

## Rationale for this change

`extractFileLocation`, `getFileFormat`, `getEqualityFieldIds` collapsed "accessor not
declared" and "accessor threw" into the same `None`/empty. Live bug: `CometScanRule`
reads an empty `getEqualityFieldIds` as "position delete" and goes native, even when
the empty was actually a swallowed failure.

`Option`/empty now means "not declared" only; genuine invoke failures propagate.
Matches `findMethod`'s existing contract, per review comment on the issue. No caller
changes needed at `CometScanRule` -- its existing outer catch just starts firing
correctly.

New `IcebergReflectionSuite` cases proving failures propagate instead of collapsing.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
+++ b/spark/src/main/scala/org/apache/comet/iceberg/IcebergReflection.scala
@@ -329,31 +329,22 @@ object IcebergReflection extends Logging {
    * Different Iceberg versions expose file paths differently:
    *   - Newer versions: location() returns String
    *   - Older versions: path() returns CharSequence
+   *
+   * `None` means neither accessor is declared; a genuine invoke failure propagates instead.
    */
-  def extractFileLocation(contentFileClass: Class[_], file: Any): Option[String] = {
-    try {
-      findMethod(contentFileClass, "location") match {
-        case Some(locationMethod) => Some(locationMethod.invoke(file).asInstanceOf[String])
-        case None =>
-          findMethod(contentFileClass, "path")
-            .map(_.invoke(file).asInstanceOf[CharSequence].toString)
-      }
-    } catch {
-      case _: Exception => None
+  def extractFileLocation(contentFileClass: Class[_], file: Any): Option[String] =
+    findMethod(contentFileClass, "location") match {
+      case Some(locationMethod) => Some(locationMethod.invoke(file).asInstanceOf[String])
+      case None =>
+        findMethod(contentFileClass, "path")
+          .map(_.invoke(file).asInstanceOf[CharSequence].toString)
     }
-  }
 
   /**
    * Extracts file location from ContentFile instance using dynamic class lookup.
    */
-  def extractFileLocation(file: Any): Option[String] = {
-    try {
-      val contentFileClass = loadClass(ClassNames.CONTENT_FILE)
-      extractFileLocation(contentFileClass, file)
-    } catch {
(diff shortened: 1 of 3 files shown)


unikdahal:
@mbutrovich @andygrove this implements the option B you both landed on in #5258, PTAL

sunchao (approved):
Reviewed the distinction between absent accessors and reflection invocation failures, including the callers' fallback handling. I found no P1/P2 issues.

The exact-head ClassLoaders, IcebergReflection, and IcebergReflectionSuite sources compiled against the cached Spark 4.0.4 dependencies, and all 17 reflection tests passed locally. Exact-head CI also passed.

PR #5417, opened 2026-08-22
test: fail unexpected vacuous fallback-invariance checks

sunchao:

The suite added in #5329 checks outcomes only after proving that the default and forced-fallback legs exercised different execution paths. However, an unprovable comparison currently becomes `SKIPPED-VACUOUS`, and the final assertion ignores those entries. The suite can therefore pass with `pass=0 fail=0 vacuous=25`, or stay green after losing an individual witness such as `AddMonths`.

Unexpected coverage loss in this fixed regression corpus should fail the test instead of requiring someone to notice a changed counter in the logs. This is test-coverage enforcement, not a product behavior change; other suites also guard the historical `AddMonths` bug.

### Which issue does this PR close?

Follow-up to #5329 and its original issue, #5328. No new issue is closed.

- Classify unexpected opaque plans, non-native defaults, and ineffective fallback toggles as `FAIL-BIND`, which reaches the existing final assertion.
- Permit only `values / StringTranslate / not-native-by-default` as the known vacuous exception. If `StringTranslate` gains native support, it may pass normally; the test does not require a fixed skip count.
- Keep verdict counters exclusive and update the suite documentation and assertion message to cover binding failures.
- Add ten regression scenarios in the same suite, covering all binding-gate branches and the expression, section, and reason boundaries of the exception.

Only `CometFallbackInvarianceSuite.scala` changes. No product code or CI workflow changes.

## How was this PR tested?

Rebuilt the native library from this upstream checkout with `cargo build --locked --offline`, then ran the focused root Maven command on Spark 4.1.3 with JDK 17:

```sh
./mvnw -o test -Dtest=none -Dsuites=org.apache.comet.CometFallbackInvarianceSuite
```

All **6 ScalaTest tests passed**, including the ten-scenario guard test in each of the three shuffle/C2R variants. The new guard test also failed in all three variants when compiled against the old implementation, before the binding fix.

Negative controls used the same upstream runtime in isolated JVMs. These counters were observed in every variant:

| Configuration | Pass | Fail | Vacuous | Result |
| --- | ---: | ---: | ---: | --- |
| Healthy baseline | 24 | 0 | 1 | Passed |
| `spark.comet.exec.project.enabled=false` | 0 | 24 | 1 | Failed as expected |
| `spark.comet.expression.AddMonths.enabled=false` | 23 | 1 | 1 | Failed as expected |

Spotless, scalastyle, and `git diff --check` passed. Local runtime validation is Spark 4.1.3 only; the remaining Spark profiles are left to CI.

### DIFF
--- a/spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometFallbackInvarianceSuite.scala
@@ -41,9 +41,9 @@ import org.apache.spark.sql.internal.SQLConf
  *      exception on the other is a failure with a named witness, not an error in the harness --
  *      exception parity is where several historical divergences have lived.
  *
- * 2. Every comparison is gated on evidence that the config flip actually moved execution. A query
- * whose executed plan proves nothing is reported SKIPPED-VACUOUS and never counted as a pass.
- * Without that gate an invariance sweep silently overstates its own coverage.
+ * 2. Every comparison is gated on evidence that the config flip actually moved execution.
+ * Unexpected binding gaps fail the test instead of silently reducing coverage. Only the known
+ * non-native-by-default StringTranslate value case is SKIPPED-VACUOUS, never counted as a pass.
  *
  * Queries deliberately avoid `ORDER BY`: a shuffle roots the executed plan at
  * `AdaptiveSparkPlanExec`, whose `children` is empty, so plan inspection would read zero for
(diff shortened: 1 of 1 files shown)


comphead (approved):
Thanks @sunchao just wondering is it some of existing Spark tests leading to this check? I'm asking to figure out if existing Spark CI testing is enough

sunchao:
@comphead no, we found out this issue via Codex review when backporting https://github.com/apache/datafusion-comet/pull/5329

Issue #3079, opened 2026-01-13
Hash partitioning not falling back to Spark for high precision decimal partition keys


andygrove:
### Describe the bug

Comet's murmur3 hash implementation is not compatible with Spark for high precision types and we should fall back to Spark if the partitioning keys contain these types.

We just need to add a check like this:

```
 271 -       case dt: DecimalType =>                                                                                                                       
 272 -         // High precision decimals (> 18) fall back to Spark because Spark converts them                                                            
 273 -         // to Java BigDecimal before hashing, which has different behavior than native hashing                                                      
 274 -         dt.precision <= 18  
```

andygrove:
This is trivial to fix but requires updating some golden files.

andygrove:
Acutally, this is not a bug. We do not need to guarantee that we allocate the same partitions as Spark.

andygrove:
Reopening. I closed this saying we don't need to allocate the same partitions as Spark, and that was wrong.

We have two shuffle implementations and both can appear in the same plan. `applyCometShuffle` picks native or columnar per exchange with no harmonization across the plan, so the two sides of a join can end up with different ones. The columnar path uses Spark's `HashPartitioning.partitionIdExpression` and the native path uses the Rust murmur3. On a decimal key with precision above 18 those disagree. So the native path does have to match Spark, not for its own sake, but to stay consistent with its sibling.

On main at `9a4d5f283`, joining a Parquet table to a JSON view on a `DECIMAL(38,0)` key:

```
CometSortMergeJoin [k#37], [k#40], Inner
:- CometSort [k#37, a#38], [k#37 ASC NULLS FIRST]
:  +- CometExchange hashpartitioning(k#37, 7), ENSURE_REQUIREMENTS, CometNativeShuffle
:     +- CometNativeScan parquet spark_catalog.default.wd_parquet[k#37,a#38]
+- CometSort [k#40, b#41], [k#40 ASC NULLS FIRST]
   +- CometColumnarExchange hashpartitioning(k#40, 7), ENSURE_REQUIREMENTS, CometColumnarShuffle
      +- FileScan json [k#40,b#41]
```

Comet returns zero rows where Spark returns four. Default `spark.comet.shuffle.mode=auto`, nothing exotic set. It reproduces with AQE enabled too, as long as coalescing doesn't collapse the fixture into a single partition, which is what you'd get at any real data volume. That's most likely why the suites have never caught it.

[...]

PR #5420, opened 2026-08-22
fix: preserve Spark AVG buffer and overflow semantics

### LINKED ISSUE #5418: Incorrect AVG result from empty native partial buffers
### Describe the bug

An empty native scalar AVG partial currently exports `(sum = NULL, count = 0)`. Spark initializes AVG buffers to `(sum = 0, count = 0)` and adds partial sums without coalescing nulls. In the supported native-partial / Spark-final path for non-decimal AVG, an empty partial can therefore turn a nonempty query result into `NULL`.

The same scalar-state distinction matters for decimal AVG: a legitimate empty buffer must be distinguishable from an overflow buffer. A fix must keep the final result of an empty AVG as `NULL` and preserve decimal overflow behavior.

### Steps to reproduce

With Comet installed and enabled in a fresh local Spark session (for example, `local[4]`), create four separate Parquet scan partitions and filter out all rows in three of them:

```python
import tempfile

for key, value in {
    "spark.sql.adaptive.enabled": "false",
    "spark.sql.files.maxPartitionBytes": "1048576",
    "spark.sql.parquet.filterPushdown": "false",
    "spark.comet.scan.enabled": "false",
    "spark.comet.convert.parquet.enabled": "true",
    "spark.comet.shuffle.enabled": "false",
    "spark.comet.testing.aggregate.finalMode.enabled": "false",
}.items():
    spark.conf.set(key, value)

with tempfile.TemporaryDirectory() as tmp:
    path = f"{tmp}/data"
    (spark.range(8, numPartitions=4)
        .selectExpr("id", "cast(id + 1 as int) quantity")
        .write.parquet(path))
    spark.read.parquet(path).createOrReplaceTempView("avg_partial_probe")
    spark.sql("SELECT avg(quantity) FROM avg_partial_probe WHERE id = 1").show()
```

The final-mode testing switch isolates the native-partial / Spark-final boundary. The query has a native partial aggregate and a Spark final aggregate, but returns `NULL` instead of `2.0`.

### Expected behavior

The result is `2.0`, matching Spark. If the filter removes every row, the final result remains `NULL`.

### Additional context

Reproduced against Apache Comet `main` at `2699f59b71788e17a2714910e166a3f83deed937` using Spark 4.0.4. This is a partial-buffer representation problem, not a request to enable mixed execution for decimal AVG.

sunchao:

Closes #5418. Includes the reviewed prerequisite #5421 at `4b5f1704d`; that PR should land first. Wide-decimal shuffle routing and its benchmark are separate in #6005.

## Rationale for this change

An empty native AVG partial exports `(NULL, 0)`, while Spark expects `(0, 0)`. When Spark merges that buffer with a nonempty partition, the null sum can erase the result. Decimal AVG must also distinguish empty input from overflow and keep overflow sticky through merging and shuffle.

- Initialize AVG sums to zero and use the count for empty results. Scalar decimal AVG uses `Some(0)` for empty input and `None` for overflow, eliminating redundant state flags and duplicated update loops.
- Preserve grouped decimal overflow through state export and shuffle; report ANSI errors when evaluating the affected group.
- Keep global and expanding-window decimal AVG/TRY_AVG in Spark for input precision at least 28, where Spark can retain a wider temporary sum until division and return a valid average.
- Keep grouped ANSI decimal AVG natively enabled. Document that batch finalization can report overflow for a group Spark would not consume after `LIMIT`. The blanket fallback, operator opt-in matrix, policy benchmark, and six associated plan snapshots have been removed.
- Preserve #5421's reviewed directional buffer policies and diagnostics. Repaired nondecimal AVG gets an explicit producer opt-in; decimal AVG buffers remain excluded from mixed execution. #5509 tracks the separate ObjectHash AVG discrepancy, and #6002 tracks decimal SUM, including expanding windows.

Beyond the prerequisite on the same main baseline, the AVG patch is 11 files, +571/-216. The total diff is 15 files, +1268/-247; there are no benchmark or golden-file changes.

At `42a6bad39`, local Spark 4.1.3 / JDK 17 full-reactor test compilation, Spotless, Scalastyle, Rust formatting, and whitespace checks pass. All 37 planner tests and 129 TPC-DS plan checks pass with golden generation disabled. Those planning checks use a verified main CI native library only for initialization.

Focused Rust and Scala regressions cover empty buffers crossing engines in both directions, sticky scalar overflow, grouped state round trips and deferred errors, columnar shuffle null-count preservation, DISTINCT, and the global/window precision boundary. Redundant AQE/ANSI permutations and the grouped fallback tests were removed.

A fresh local native build is blocked by the configured registry's missing locked DataFusion 55.1.0. [Current-head CI](https://github.com/apache/datafusion-comet/actions/runs/35276146413), with `run-all-spark-profiles`, is pending and will validate native execution and the supported Spark profiles.

### DIFF
--- a/native/spark-expr/src/agg_funcs/avg.rs
+++ b/native/spark-expr/src/agg_funcs/avg.rs
@@ -146,8 +146,11 @@ pub struct AvgAccumulator {
 
 impl Accumulator for AvgAccumulator {
     fn state(&mut self) -> Result<Vec<ScalarValue>> {
+        // Spark's final AVG adds partial sums without coalescing nulls. Empty partials
+        // must therefore carry a zero sum, even when update_batch was never called.
+        let sum = if self.count == 0 { Some(0.0) } else { self.sum };
         Ok(vec![
-            ScalarValue::Float64(self.sum),
+            ScalarValue::Float64(sum),
             ScalarValue::from(self.count),
         ])
     }
@@ -343,5 +346,52 @@ where
     fn size(&self) -> usize {
         self.counts.capacity() * std::mem::size_of::<i64>()
             + self.sums.capacity() * std::mem::size_of::<T>()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use arrow::array::Float64Array;
+
+    #[test]
+    fn empty_partial_state_matches_spark() -> Result<()> {
+        for values in [None, Some(vec![]), Some(vec![None, None])] {
+            let mut acc = AvgAccumulator::default();
+            if let Some(values) = values {
+                acc.update_batch(&[Arc::new(Float64Array::from(values))])?;
+            }
+            assert_eq!(
+                acc.state()?,
+                vec![ScalarValue::Float64(Some(0.0)), ScalarValue::Int64(Some(0))]
+            );
+            assert_eq!(acc.evaluate()?, ScalarValue::Float64(None));
+        }
+        Ok(())
+    }
+
+    #[test]
+    fn merge_empty_partial_states() -> Result<()> {
+        for nonempty in [false, true] {
+            let mut final_acc = AvgAccumulator::default();
+            let values = nonempty.then(|| vec![Some(2.0), Some(4.0)]);
+            for values in [None, values, Some(vec![None])] {
+                let mut partial = AvgAccumulator::default();
+                if let Some(values) = values {
+                    partial.update_batch(&[Arc::new(Float64Array::from(values))])?;
+                }
+                let state = partial
+                    .state()?
+                    .into_iter()
+                    .map(|value| value.to_array_of_size(1))
+                    .collect::<Result<Vec<_>>>()?;
+                final_acc.merge_batch(&state)?;
+            }
+            assert_eq!(
+                final_acc.evaluate()?,
+                ScalarValue::Float64(nonempty.then_some(3.0))
+            );
+        }
+        Ok(())
     }
 }
--- a/native/spark-expr/src/agg_funcs/avg_decimal.rs
+++ b/native/spark-expr/src/agg_funcs/avg_decimal.rs
@@ -300,14 +300,22 @@ fn make_decimal128(value: Option<i128>, precision: u8, scale: i8) -> ScalarValue
 
 impl Accumulator for AvgDecimalAccumulator {
     fn state(&mut self) -> Result<Vec<ScalarValue>> {
+        // Spark distinguishes an empty partial (zero sum) from an overflow (null sum).
+        let sum = if !self.is_not_null {
+            None
+        } else if self.count == 0 {
+            Some(0)
+        } else {
+            self.sum
+        };
         Ok(vec![
-            ScalarValue::Decimal128(self.sum, self.sum_precision, self.sum_scale),
+            ScalarValue::Decimal128(sum, self.sum_precision, self.sum_scale),
             ScalarValue::from(self.count),
         ])
     }
 
     fn update_batch(&mut self, values: &[ArrayRef]) -> Result<()> {
-        if !self.is_empty && !self.is_not_null {
+        if !self.is_not_null {
             // This means there's a overflow in decimal, so we will just skip the rest
             // of the computation
             return Ok(());
@@ -342,6 +350,15 @@ impl Accumulator for AvgDecimalAccumulator {
         // counts are summed
         self.count += sum(partial_counts).unwrap_or_default();
 
+        // Empty partials have non-null zero buffers. A null sum is an overflow marker;
+        // grouped partials also null the count. Do not let sum() skip these markers or
+        // let a later empty partial revive an accumulator that has already overflowed.
+        if !self.is_not_null || partial_sums.null_count() > 0 || partial_counts.null_count() > 0 {
+            self.is_not_null = false;
+            self.sum = None;
+            return Ok(());
+        }
+
         // sums are summed
         if let Some(x) = sum(partial_sums) {
             let v = self.sum.get_or_insert(0);
@@ -359,24 +376,23 @@ impl Accumulator for AvgDecimalAccumulator {
     }
 
     fn evaluate(&mut self) -> Result<ScalarValue> {
-        // Check for overflow during sum accumulation in ANSI mode.
-        // This matches Spark's DecimalDivideWithOverflowCheck behavior.
-        // `count` guards against reporting an overflow when there was nothing to sum: an
-        // empty or all-null input also leaves `sum` as None, and `is_empty` cannot
-        // distinguish those cases because the counts merged in `merge_batch` are never null.
-        if self.sum.is_none()
-            && !self.is_empty
-            && self.count > 0
-            && self.eval_mode == EvalMode::Ansi
-        {
+        // A grouped overflow can have a null count, leaving the merged count at zero.
+        // Check the overflow marker before treating a zero count as empty input.
+        let has_overflow =
+            !self.is_not_null || (self.sum.is_none() && !self.is_empty && self.count > 0);
+        if has_overflow && self.eval_mode == EvalMode::Ansi {
             let error = decimal_sum_overflow_error("avg");
             return Err(self.wrap_error_with_context(error));
         }
 
-        // Also check if is_not_null is false (indicates overflow)
-        if !self.is_not_null && self.count > 0 && self.eval_mode == EvalMode::Ansi {
-            let error = decimal_sum_overflow_error("avg");
-            return Err(self.wrap_error_with_context(error));
+        // An all-empty native final now merges zero sums. It must still return NULL,
+        // rather than dividing that sum by zero, in every evaluation mode.
+        if self.count == 0 || has_overflow {
+            return Ok(make_decimal128(
+                None,
+                self.target_precision,
+                self.target_scale,
+            ));
         }
 
         let scaler = 10_i128.pow(self.target_scale.saturating_sub(self.sum_scale) as u32);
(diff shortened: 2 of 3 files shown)


andygrove:

The `AvgAccumulator::state` fix is clearly correct and clearly valuable. Spark's `Average.mergeExpressions` is a plain `sum.left + sum.right` with no coalesce, so a partial that reports a null sum for an empty partition poisons the whole final. Getting decimal to distinguish "empty partial" from "overflowed partial" through the state round trip is the fiddly part and the new accumulator tests read well.

I have concerns about how much else is in here.

**The decimal precision restriction on native shuffle**

Changing `CometShuffleExchangeExec` so that hash partitioning keys with `precision > 18` fall back is by far the widest-reaching change in this PR, and the justification given is that different partition assignments hide an overflow in the `AVG(DISTINCT ...)` regression. That is a very large hammer for making one test deterministic. Every query that hash-partitions on a `DECIMAL(38, s)` key now loses native shuffle, whether or not it involves `AVG`.

The removed TODO points at #3079 and says wider decimals need `BigDecimal` conversion before hashing. If native hashing genuinely disagrees with Spark for `precision > 18`, that is a correctness bug in its own right and it deserves its own PR, its own tests, and a note in the migration guide, rather than arriving as a side effect of an `AVG` fix. If it is not a correctness bug and this is purely about the test, could the test pin the partitioning some other way, for example by constructing the input so the overflow occurs regardless of assignment?

Either way, could you say what this costs? A before and after on a query that groups by a `DECIMAL(38, 2)` key would settle whether this is a minor or a major regression.

**This PR contains #5421 verbatim**

That is already being discussed on the other thread so I will not repeat it, except to note that the diff here is harder to review because of it. Once the ordering is settled, rebasing so this PR shows only its own changes would help a lot.

**`aggregateSupportLevel` only covers the ungrouped case**

The check is `op.groupingExpressions.isEmpty`, and the docs say grouped averages stay native. I believe the reasoning is that Spark's grouped path stores the buffer in an `UnsafeRow` at the declared precision, so it overflows at the same point Comet does, while the ungrouped codegen path keeps a wider local. Could you state that explicitly in the comment? And does the same hold for `SortAggregateExec` and for grouped `ObjectHashAggregateExec` after a spill, where the buffer round-trips through a different representation?

**The relaxed ANSI condition in `AvgDecimalGroupsAccumulator::evaluate`**

The guard changes from `!is_not_null && count > 0 && ansi` to `!is_not_null && ansi`. That is deliberate, and the comment explains why. But it means a group whose `is_not_null` is false and whose count is zero now throws in ANSI mode where it previously did not. I convinced myself that an all-null group keeps `is_not_null` true and so is unaffected, but I would rather see that pinned by a test: ANSI mode, grouped decimal `AVG`, with one group that is entirely null alongside a group that really does overflow. If that test already exists somewhere in the 381 new lines in `CometAggregateSuite`, pointing at it in the description would be enough.

**Non-local return in `CometWindowExec`**

The new block uses `return None` from inside three levels of nested `match` inside `getSupportLevel`. It matches the existing style just below it, so I am not asking you to change it. But four nested matches to reach one condition is hard to follow. Would a small predicate helper, along the lines of `isMaxPrecisionDecimalAverage(windowExpr)`, make both this block and the existing one below it easier to read?

**Splitting**

Counting them up, this PR contains: the float `AVG` empty-partial fix, the decimal `AVG` overflow-marker plumbing, an ungrouped decimal `AVG` operator fallback, a window `AVG` fallback, a native shuffle restriction on wide decimals, and all of #5421. The first two are tightly coupled and obviously belong together. The shuffle restriction in particular looks separable and would be much easier to evaluate on its own.

sunchao:
Updated in [6dca7f116](https://github.com/apache/datafusion-comet/commit/6dca7f11670ca7ed4f7a1ec69731245d77497bc2).

I worked through the review points and added the wide-decimal routing benchmark, carried the directional buffer policy from #5421, and corrected the grouped-AVG explanation. The updated description includes current validation and measured costs, with earlier validation clearly identified as historical.

1. **Scope beyond AVG.** Yes, the hash-key check affects every native shuffle whose decimal key has precision greater than 18. In `auto` mode, eligible queries use columnar shuffle and can retain native aggregates. In `native` mode, they use Spark shuffle, which can also require aggregate fallback. Wide decimal payloads and range keys are not excluded by this hash-key check. The description and existing tuning guidance now make that distinction explicit.

2. **Hash compatibility versus test determinism.** This is a real partition-assignment mismatch, also tracked in #3079. Spark hashes wide decimals through `BigInteger` bytes, while the current native path uses the fixed-width little-endian integer representation. Decimal partial-sum overflow depends on which values meet in a partition, so matching the final numeric type is insufficient. The guard is not just a way to stabilize one test, and I kept it.

3. **Splitting the changes.** The planner repair is already separated as #5421. I kept the AVG state repair, admission restrictions and wide-key partitioning change together because the `AVG(DISTINCT ...)` regression needs compatible partial buffers and partition assignments. Splitting those into independently mergeable fixes would leave that regression incomplete. The unrelated, pre-existing ObjectHash case now has a separate issue, #5509.

4. **Focused shuffle tests.** The existing `wide decimal hash keys use Spark-compatible shuffle partitioning` regression covers precision 18/19/38 across `native`, `auto` and `jvm`, together with payload/range/single-partition controls. It passed again in the current 47-test run. The new benchmark also checks actual input partition counts, results and routes rather than inferring execution from configuration.

5. **Migration guide.** The migration guide excludes correctness and native-admission changes from its scope. The tuning guide is the existing home for the shuffle-mode behavior, so I retained that explanation there and expanded the PR description rather than adding a migration flag or suggesting users restore incompatible hashing.

6. **Measured performance.** For 1,048,576 rows and 10,000 `DECIMAL(38,2)` groups, the four-process ABBA run measured 137→501.5 ms in `native` mode and 137→188.5 ms in `auto` mode, averaging the two run means for each variant. All eight variant/mode comparisons matched all Spark results and asserted their routes. This isolates the guard by removing only that check while keeping the same current native library. It is not the entire PR against its base. The description includes both raw run means and the resulting plans. The cost is material, especially when `native` mode restores Spark aggregation, but the faster incompatible route is not a safe alternative.

7. **Relationship to #5421.** #5421 remains the prerequisite and should land first. Its latest fix conservatively rejects native AVG partials before Spark Final because that branch still emits `(NULL, 0)` for never-updated partials. This PR repairs that state and explicitly enables non-decimal NativePartialToSparkFinal again. Decimal buffers remain excluded. The positive counterpart passed 12 mixed-route combinations plus six native controls with this branch's own rebuilt library. I have not treated #5421's temporary AVG restriction as appropriate for the repaired implementation.

8. **Why the high-precision guard selects ungrouped AVG.** The original wording was too broad. Ordinary Spark hash aggregation stores grouped decimal buffers in `UnsafeRow`, whose precision checks differ from generated global aggregation's local buffer. That explains the intended guard but does not prove all grouped implementations equivalent. The description now names the representation and qualifies the ObjectHash exception.

9. **SortAggregateExec.** Spark sort aggregation is already outside this native conversion path. The two ANSI controls in the earlier review probe remained in Spark and matched Spark results. No additional admission guard is needed for that path.

10. **ObjectHashAggregateExec and spill.** This concern is valid as a separate existing issue. `SpecificInternalRow` can preserve a wider temporary decimal sum, including after ObjectHash falls back to sort, while native grouped AVG can lose it earlier. I filed #5509 with the `0.6, 0.6, -0.4` counterexample, an all-null group, default/forced-spill checks and exact native base/head evidence showing the discrepancy predates this PR. The description no longer claims that grouping alone establishes compatibility.

11. **All-null and overflow groups.** The existing Rust `grouped_overflow_with_zeroed_null_count_is_not_empty` test and Scala `grouped decimal AVG preserves overflow across columnar shuffle` regression cover the distinction, including the all-null group alongside overflow. I retained them instead of adding duplicate tests. Their earlier validation is identified separately from this follow-up's focused run.

12. **Return in windowExprToProto.** The return is in a regular match within the method, not a closure, so it is not a Scala non-local return. I did not extract a helper solely for that concern.

The current optimized native build, full Spark 4.1.3 JVM reactor/style checks, and all 47 focused Scala tests pass. The full historical cross-version suite was not rerun for this follow-up. CI on the pushed revision is still separate from these local results.

sunchao:
Updated in [ae98cef70](https://github.com/apache/datafusion-comet/commit/ae98cef70f6f023acfef48248c857410e8be5c57).

I resolved the main-branch conflict with a history-preserving merge of `525c05b52`. The only manual resolution combines the decimal and upstream shuffle-test imports, retaining both sets of tests. The AVG implementation and policy, wide-decimal guard, and benchmark source are unchanged.

The merged branch has a fresh default-release native build, a full Spark 4.1.3 JVM reactor/style pass, and **55 passing focused Scala tests**, including the 18 AVG interoperability query combinations and upstream's new RSS/JNI/protocol coverage. An additional untimed wide-decimal run matches all 10,000 Spark results and verifies both expected shuffle/aggregate routes. Each run's loaded native library was checked against the merged build.

The description now distinguishes this validation from the earlier `6dca7f116` performance measurements. Those numbers remain pinned to the measured revision; I did not relabel them as merged-head timings or repeat the full cross-version suite.

sunchao:
Fixed the grouped decimal AVG evaluation-order issue in [728c55e8d](https://github.com/apache/datafusion-comet/pull/5420/commits/728c55e8d1fb92e3ebb59e21d9ed5c393cbc17f2). Grouped decimal `AVG` in ANSI mode now keeps every stage evaluating that AVG in Spark. This prevents native batch emission from throwing for a group that Spark's LIMIT does not consume, while preserving errors when the overflowing group is consumed. The native empty/count-state and overflow repairs are unchanged.

The guard applies to every decimal precision and to the whole affected aggregate operator, including colocated aggregates, even without LIMIT. This intentionally gives up native aggregation for those ANSI queries. Grouped `TRY_AVG`, legacy decimal AVG, non-decimal AVG, and unrelated native work retain their existing eligibility. The description and generated-documentation source now state that scope; existing benchmark timings use legacy mode and do not measure this new fallback.

The full Spark 4.1.3 JVM reactor passed **139 tests** (107 aggregate, 31 planner, and the AVG SQL fixture), with the loaded native library verified against the unchanged native tree. Both new LIMIT regressions fail against the unchanged planner and pass with the fix, with AQE off/on. Consumed-overflow and native-control checks also pass, as do Spotless, Scalastyle, and CI's syntactic Scalafix check. CI for the new commit is pending.

sunchao:
Fixed both re-review findings in [97f605ede](https://github.com/apache/datafusion-comet/commit/97f605ede8120fa354371baaddb98cdc3cfb3c10).

The six affected TPC-DS plan expectations were regenerated with the supported harness. They now preserve the intended grouped ANSI decimal AVG fallback on Spark 4.x; the shared and Spark 3.x expectations are unchanged. The operator guide also explains that the guard applies when AVG still has a DECIMAL input after Spark optimization, including other functions in the same aggregate operator and queries without LIMIT.

Validation passed all 129 plan checks on each of Spark 3.4.3, 3.5.9, 4.0.4, 4.1.3, and 4.2.0 with generation disabled (645 total). Current JVM classes and the actual loaded native library were verified. No implementation or native code changed; no new native build or performance result is claimed. The PR description has been updated, and the new hosted CI run is pending.

sunchao:
Corrected the generated AVG note in [754a1df66](https://github.com/apache/datafusion-comet/commit/754a1df6687b2441f13551d6b6ee5f0f3d5de690) to say that grouped ANSI fallback applies when the input remains `DECIMAL` after Spark optimization, matching the operator guide.

This changes only the documentation reason; aggregate admission and native code are unchanged. Full reactor compilation, Spotless, and fresh GenerateDocs output checks passed. The generated Average page now includes the qualification.

andygrove (commented):
Following up on my earlier question about `aggregateSupportLevel` only covering the ungrouped case. The answer in `operators.scala` went further than I expected. Grouped ANSI decimal `AVG` now falls back at every precision, and because ANSI is the default on Spark 4.x that removes the whole aggregate operator from Comet for a very common shape. The six regenerated Spark 4.x TPC-DS plans contain eighteen aggregate operators reverting to Spark, and q1, q18, q30, q65, q81 and q18a are all rooted at `TakeOrderedAndProject`, which consumes the entire aggregate output. The lazy-`LIMIT` argument does not apply to any of them, so the guard is paying a real cost in exactly the cases where Spark would raise the same error. The benchmark in the description is legacy mode, so the cost of this particular fallback is not measured anywhere, and `Unsupported` leaves users no way to opt back in.

Could the guard be narrowed? Two things I would like your read on. `AvgDecimalGroupsAccumulator::merge_batch` in `avg_decimal.rs` around line 592 raises the ANSI overflow error while it is still consuming input, before any group is emitted. If that deferred to `evaluate`, the remaining divergence from Spark would be batch granularity rather than whole-input granularity. Could the operator check also look at whether a limit can actually truncate this aggregate's output, instead of firing unconditionally? If neither is practical, could the fallback sit behind a config so a user who does not care about the `LIMIT` case can keep native aggregation?

Separately, in `AvgDecimalAccumulator` the `is_empty` field looks like it no longer does anything. Its only reader is the `has_overflow` expression at `avg_decimal.rs` line 382, and after this change I cannot construct a state where `sum` is `None` while `is_not_null` is true and `count` is greater than zero, because every path that clears `sum` also clears `is_not_null`. The flag has also stopped meaning what the comment above line 347 says, since merging a batch of empty partials with non-null zero counts now sets it to false. Is it still load bearing, or can it go?

The accumulator work itself reads well. The state round trip for decimal, the zeroed-null-count case and the precision boundary at 27 versus 28 are all pinned by tests in both ANSI modes, which was the part I was most worried about last time.

sunchao:
Rebased onto current main in `45800b356`.

Thanks, Andy. I addressed the accumulator and configuration points:

- Grouped merge overflow now records a sticky invalid state and raises during evaluation. Tests cover precision and `i128` overflow, prefix emission, state round trips, and later updates in ANSI, legacy, and TRY modes.
- Removed the redundant scalar `is_empty` field.
- Changed grouped ANSI decimal AVG from `Unsupported` to `Incompatible`. Users can opt in with `spark.comet.operator.HashAggregateExec.allowIncompatible=true` or `spark.comet.operator.ObjectHashAggregateExec.allowIncompatible=true`, according to the operator. Decimal buffer boundaries and global/window restrictions still apply.

The default remains conservative. Deferring errors helps prefix emission, but native finalization can still evaluate an unconsumed group within a batch. I agree that queries which consume every group pay for a fallback they do not need. A top-level `TakeOrderedAndProjectExec` alone does not establish that guarantee: Spark can use `take(limit)` when its child already provides the requested ordering, and AQE can plan aggregate stages separately from their consumer. I documented that limitation and the opt-in.

Added `--grouped-ansi-avg` to compare default fallback and native opt-in on the same ANSI workload, with result and plan checks outside timing. Measurements remain pending: the configured local registry cannot resolve the rebased lockfile's DataFusion 55.1.0, preventing a fresh native build and benchmark run.

The [previous hosted run](https://github.com/apache/datafusion-comet/actions/runs/34438085805) passed 1,273 Rust tests and 862 Spark 4.1 execution tests. Its six failing jobs hit the same redundant string-interpolation prefix in the benchmark; I removed it. Those results cover the previous synthetic merge. The new revision needs its own hosted CI results. Full-reactor Spark 4.0.4 test compilation, syntactic Scalafix 0.14.6, Rust formatting, Spotless, Scalastyle, and whitespace checks pass locally.

andygrove (commented):
Two things before I can read this as an AVG fix.

The copy of #5421 in here predates its review. `supportsNativePartialToSparkFinal` still defaults to `supportsSparkPartialToNativeFinal(fn)`, which is the inherited default that #5421 replaced with nine explicit per-function opt-ins in `6c14796f5` after @comphead raised it, and `revertUnsafePartialAggregates` is the version from before the unrepaired-producer diagnostic in `4b5f1704d`. Merging as it stands would put both decisions back. Could you rebase onto #5421's head? `operators.scala` conflicts with main now as well.

The shuffle routing change is a separate fix riding along. `CometShuffleExchangeExec.supportedShuffleType` goes from `case _: DecimalType => true` with a TODO pointing at #3079 to `d.precision <= 18`, which resolves a hashing-compatibility question that has nothing to do with AVG, and your own numbers put it at 137 ms to 501.5 ms in `native` mode. Could it go in its own PR? That would let the routing change be benchmarked, reviewed and reverted independently of the accumulator work.

On the grouped ANSI decimal AVG guard, moving to `Incompatible` helps, but I do not think the trade has changed. This is the same divergence #5533 is about, which is that Comet evaluates a batch eagerly where Spark evaluates lazily and stops at a `LIMIT`, and over there the open question is whether to document the class or special-case the planner. Here we are picking the expensive side of it for a shape that is the default on Spark 4.x. All six regenerated TPC-DS plans are rooted at `TakeOrderedAndProject`, which consumes every group, so none of them benefits from the fallback and all of them pay for it. Would documenting this under "Known result-value divergences" in `compatibility/index.md` and keeping `Compatible()` be the better call? If the fallback stays, the opt-in deserves a line saying it is operator-wide, since `spark.comet.operator.HashAggregateExec.allowIncompatible=true` also accepts anything else that operator marks incompatible later.

The new expanding-frame guard in `CometWindowExec` only inspects `Average`. Spark's `AggregateProcessor` writes the window buffer into a `SpecificInternalRow`, which does not check decimal overflow the way `UnsafeRowWriter` does, so an expanding-frame decimal `SUM` has the same sticky-overflow divergence you are guarding AVG against. Should that be in the same match, or tracked alongside the decimal SUM point I left on #5421?

Last, this regenerates golden plans for the `-spark4_0` and `-spark3_5` tiers and the base tier, but only Spark 4.1 ran. Could you add `run-all-spark-profiles`?

The accumulator work itself reads well now. `AvgAccumulator` emitting `(0.0, count)` matches `Literal.default(sumDataType)` exactly, the scalar `is_empty` field is gone, and deferring the grouped merge overflow to `evaluate` is the right shape.

andygrove:
On the window question above, I filed the decimal SUM divergence as #6002, which covers the ever-expanding frame alongside the ungrouped aggregate and `ObjectHashAggregate`. Happy for the `CometWindowExec` guard to stay AVG-only here and for SUM to be handled there, as long as this PR is not the last place it gets written down.

sunchao:
Thanks, Andy. Rebased onto main `3b942e593` and pushed the follow-up in [c59caf673](https://github.com/apache/datafusion-comet/commit/c59caf673f11a5587a92ab4282a8508344a889a3).

- **#5421:** Synchronized the prerequisite with its reviewed head `4b5f1704d`, including the independent default-false policies, nine explicit producer opt-ins, restoration explanations, and unrepaired-producer diagnostic. Repaired nondecimal AVG now has its own explicit producer opt-in. I also retained the newer prerequisite regressions and changed their deliberately unsafe AVG fixtures to decimal AVG, since ordinary AVG is now safe.
- **Shuffle scope:** Moved the wide-decimal restriction, shuffle-specific expectations, tuning guidance, and benchmark into #6005. Its standalone COUNT benchmark is independent of AVG, and the historical AVG-workload measurements are identified as historical there. That draft includes #5421 because a new shuffle fallback must also restore incompatible collection buffers; a focused `collect_list`/`collect_set` regression covers the boundary. The AVG tests here select JVM shuffle explicitly where Spark's partition assignments or columnar buffer conversion are part of the test.
- **Grouped ANSI policy:** I retained the configurable `Incompatible` fallback as the conservative option. It still costs native coverage for queries that consume every group. The known-divergence section now describes the behavior, and the operator guide explicitly says that each opt-in accepts every incompatibility for that operator, including future additions. The remaining divergence is eager finalization of a group that Spark never consumes; deferring merge errors does not remove it. Grouped ANSI timings remain unmeasured.
- **Window SUM:** Kept the guard AVG-only as agreed and documented #6002 alongside the known decimal SUM divergence.
- **Profiles:** Added `run-all-spark-profiles`. The 11 shuffle-related golden changes moved to #6005; the six grouped-ANSI plan expectations remain here. Both PR descriptions now reflect their final scope and validation.

Local Spark 4.1.3 validation passed full-reactor test compilation, Spotless, Scalastyle, Rust formatting, all 37 planner tests, and all 129 TPC-DS plan checks with generation disabled. The planner checks use a verified main CI library for initialization and do not execute the changed native AVG implementation. The configured local registry still lacks locked DataFusion 55.1.0, so fresh native AVG execution and cross-profile results are pending [default-profile CI](https://github.com/apache/datafusion-comet/actions/runs/35273797534) and [additional-profile CI](https://github.com/apache/datafusion-comet/actions/runs/35273813526). Separately, #6005 passed seven focused execution tests, nine affected plan checks, and benchmark result validation using a verified CI library whose native source tree matches that branch exactly.

sunchao:
Andy, I simplified the AVG patch further in [42a6bad39](https://github.com/apache/datafusion-comet/commit/42a6bad3938e611f4737485d7454f692f8053040). This supersedes the grouped ANSI policy in my previous reply: grouped AVG now remains `Compatible()`, with the eager batch/LIMIT overflow difference documented under known divergences, as you suggested. That removes the blanket fallback, operator opt-in tests, benchmark, and all six associated golden snapshots.

The scalar decimal accumulator also uses one update loop, and the test coverage is consolidated around distinct failures: empty partials crossing engines in both directions, sticky overflow, grouped state round trips with a zeroed null-count payload, and the 27/28 precision boundary. The grouped columnar-shuffle test includes overflowing, valid, and all-null groups and checks legacy, TRY, and ANSI behavior.

The reviewed #5421 prerequisite remains intact, wide-decimal shuffle remains separate in #6005, and the window guard stays AVG-only with decimal SUM documented under #6002. Beyond the prerequisite on the same main baseline, the patch is now **+571/-216**, down from **+1475/-224**. No benchmark or golden-file changes remain here.

Full-reactor Spark 4.1.3 test compilation, formatting/style checks, all 37 planner tests, and all 129 TPC-DS plan checks pass. The planning checks use a verified main native library only for initialization. [Current-head CI](https://github.com/apache/datafusion-comet/actions/runs/35276146413) has `run-all-spark-profiles`; native execution results are pending because the configured local registry cannot resolve locked DataFusion 55.1.0.

andygrove (commented):
The compatibility argument reads well now, and I checked it against `Average.scala` rather than taking it on trust. `UnsafeRow.setDecimal` nulling the field when `changePrecision` fails is what makes the grouped case genuinely equivalent to Comet, and `doProduceWithoutKeys` keeping the buffer in `Decimal` mutable state is what makes the ungrouped case genuinely different. Deferring the grouped merge error to `evaluate` and dropping `count > 0` from the ANSI check are both right. I had not appreciated that the second is a fix rather than a relaxation, since the old guard would have swallowed the error after a columnar shuffle zeroed the count payload.

Three things I would still like before this goes in.

The docs do not say the ungrouped fallback takes the whole operator with it. The new Aggregation section in `docs/source/user-guide/latest/compatibility/operators.md` reads as if only the average falls back, but your own change to `final decimal avg` shows `SELECT MIN(a), MAX(a), COUNT(a), SUM(a), AVG(a) FROM $table` dropping from 2 native aggregates to 0. A single wide-decimal `AVG` costs a user native execution for every aggregate sharing that operator. Could the section say so? That seems like the thing a reader of that page would most want to know.

The fixture that justifies the ungrouped fallback never asserts the value that justifies it. Every case in `high-precision global decimal AVG and TRY_AVG fall back to Spark` uses `checkSparkAnswerAndNumOfAggregates(..., 0)`, and once the aggregate has fallen back Comet is Spark, so the answer comparison cannot fail. That pins the fallback without pinning the reason for it. Would you add a `checkAnswer` on the `AVG(v38)` cases asserting `0.6` at the result scale? I worked through `DecimalDivideWithOverflowCheck` and I believe Spark does return `0.6` where native returns `NULL`, but that is the entire justification for the guard and nothing in the suite records it. The `repartition(2, col("ord"))` case is the better one to pin, since it does not depend on the exchange being elided by `coalesce(1)`.

The grouped overflow round trip only runs over columnar shuffle. `grouped decimal AVG preserves overflow across columnar shuffle` pins `CometColumnarShuffle`, which is the harder case and the right one to have. But `spark.comet.shuffle.mode` defaults to `auto` and prefers native, and the grouping key there is an `int`, so the default path for that shape is the one not covered. Could the test loop over `jvm` and `native`? Removing the `partial_counts.null_count()` handling leaves the sum's null bit as the only thing carrying overflow across the exchange, so it would be good to have both writers on record.

Ordering note rather than a finding. This still carries #5421 in full, now at `4b5f1704d`, so the merge order still matters and #5421 needs to land first.

sunchao:
Thanks, Andy. Addressed the three points in [7e9026df4](https://github.com/apache/datafusion-comet/commit/7e9026df4968111be023dbe72610bc4e3df9ba0e):

- The aggregation guide now explicitly says the wide-decimal AVG/TRY_AVG fallback moves the whole operator to Spark, including sibling MIN/MAX/COUNT/SUM expressions, in both partial and final stages.
- The global fixture now uses explicit expected results while retaining the native-aggregate count assertions. It pins `0.6` at scale 38 for AVG/TRY_AVG, including the `repartition(2, col("ord"))` case. The existing `collect_list` case explicitly expects NULL: Spark's object aggregate materializes an overflowing partial buffer before division.
- The grouped overflow regression now loops over `jvm` and `native`, preserving the legacy AVG, ANSI TRY_AVG, ANSI error, valid-group, and all-null-group checks. It asserts the shuffle writer directly above the partial aggregate, so the input repartition cannot satisfy the assertion accidentally.

#5421 has merged, and the rebase removed it from this PR's diff. This follow-up changes two files. Both affected tests pass on Spark 4.1.3 / JDK 17 with the current-source native library; full-reactor compilation, Spotless, Scalastyle, and whitespace checks pass. [Current-head CI](https://github.com/apache/datafusion-comet/actions/runs/35946785166) is running with all Spark profiles and Spark 4.1 SQL tests enabled.

PR #5421, opened 2026-08-22
fix: revert unsafe partial aggregates after final fallback

### LINKED ISSUE #5419: Unsafe native partial aggregates survive child-triggered final fallback
### Describe the bug

The unsafe-partial aggregate prepass checks aggregate expression support before child conversion is known. A final aggregate may pass those checks and then fall back because its shuffle child is not native. The corresponding native partial aggregate can survive even when the aggregate is explicitly ineligible for mixed Spark/Comet execution.

For decimal AVG, an empty native partial can then poison the Spark final aggregate's sum and produce an incorrect `NULL` result. The problem occurs with AQE both disabled and enabled.

### Steps to reproduce

With Comet installed and enabled in a fresh local Spark session (for example, `local[4]`):

```python
import tempfile

for key, value in {
    "spark.sql.adaptive.enabled": "false",  # Also reproduces with true.
    "spark.sql.files.maxPartitionBytes": "1048576",
    "spark.sql.parquet.filterPushdown": "false",
    "spark.comet.scan.enabled": "false",
    "spark.comet.convert.parquet.enabled": "true",
    "spark.comet.shuffle.enabled": "false",
}.items():
    spark.conf.set(key, value)

with tempfile.TemporaryDirectory() as tmp:
    path = f"{tmp}/data"
    (spark.range(8, numPartitions=4)
        .selectExpr("id", "cast(200 as decimal(20,2)) amount")
        .write.parquet(path))
    spark.read.parquet(path).createOrReplaceTempView("decimal_avg_probe")
    spark.sql("SELECT avg(amount) FROM decimal_avg_probe WHERE id = 1").show()
```

Only one of the four scan partitions contains a matching row. The final aggregate falls back across the Spark shuffle, but the partial aggregate remains native. The query returns `NULL` instead of `200.000000`.

### Expected behavior

When the final aggregate actually falls back and its buffers are not declared safe for mixed execution, the feeding partial aggregate must also use Spark. This needs to happen before AQE materializes the partial output, and it must survive stage-only rule reapplication.

The result should be `200.000000`. Supported mixed aggregates such as MIN/MAX and fully native aggregate chains should retain native execution; unrelated native filters and scans should not be reverted.

### Additional context

Reproduced against Apache Comet `main` at `2699f59b71788e17a2714910e166a3f83deed937` using Spark 4.0.4. The existing expression-support prepass is still useful, but does not cover a final fallback caused by actual child eligibility. This also requires handling the intermediate PartialMerge/grouping and exchange chain for distinct aggregates.

sunchao:

Closes #5419. This planner repair is a prerequisite for the native AVG state fixes in #5420.

## Rationale for this change

An enabled native shuffle can become ineligible after its child converts to Comet. This can leave a native partial aggregate feeding a Spark final aggregate that expects a different buffer format. For example, `collect_list` and `collect_set` emit native array buffers, while Spark expects serialized binary state. Grouping by an unsupported array hash key or disabling native hash partitioning can expose this mismatch.

AVG also needs protection until #5420: an untouched native partial emits `(null, 0)`, whereas Spark expects `(0.0, 0)`. A COUNT/AVG query with one surviving value across four input partitions can otherwise return NULL instead of the average.

- Repair incompatible native-partial/Spark-final boundaries after conversion, restoring the feeding aggregate and exchange chain while retaining native work below it. Tag restored partials so AQE stage replanning keeps them in Spark.
- Check buffer compatibility independently in each direction. Preserve safe native COUNT partials, and keep native AVG partials from feeding Spark finals until their state is fixed.
- Share partial restoration and fallback explanations with the existing Celeborn path. If repair stops at an existing stage but an incompatible native buffer producer remains on the known aggregate/exchange path, warn once and attach an EXPLAIN reason without rewriting that stage.
- Cover actual shuffle fallback, empty AVG partitions, safe mixed execution, DISTINCT merge chains, and AQE replanning with focused regressions.

The independent opt-in policy reduces native aggregation coverage when the Final stays in Spark, particularly with `spark.comet.exec.shuffle.enabled=false`. For example, decimal SUM loses its native Partial, changing one native aggregate stage to zero. Decimal SUM has a matching `(sum, isEmpty)` layout, but its native accumulator makes intermediate precision overflow sticky or throws immediately. Spark's generated scalar SUM can recover after cancellation: `DECIMAL(38,38)` inputs `0.6, 0.6, -0.6` return `0.6`. The exclusion protects that behavior. [#5975](https://github.com/apache/datafusion-comet/issues/5975) tracks auditing and recovering the remaining native-partial coverage; ordinary-input sweeps alone do not establish safe interoperability.

Rebased onto Apache main `1c25b492bfb5d4615c7838cf24ec4da867263034`. The test-import conflict was resolved by retaining both imports. Range-diff and independent source review confirmed that all 10 PR commits and current-main behavior were preserved.

- Spark 4.0.4 / JDK 17: all 220 tests passed across `CometExecRuleSuite`, `CometCelebornShufflePlanningSuite`, `RevertNativeForTransitionHeavyStagesSuite`, and `CometAggregateSuite`. Two Spark 4.1-only cases were canceled as expected; two existing tests remain ignored.
- Root-reactor compilation, Spotless, Scalastyle, CI configuration/suite-registration checks, and `git diff --check` passed.
- Local Spark 4.1 validation was blocked before test compilation by a connection timeout fetching `jackson-bom:2.21.2` from the configured Maven mirror. The mirror configuration was retained.

Runtime validation used the verified Linux native library from [CI run 35467971490](https://github.com/apache/datafusion-comet/actions/runs/35467971490/job/105963957957), artifact `10591229783`. Its native source tree matches this PR exactly (`73a4ae2bcf57eb601a40bd6bd736c04bc82a4d15`). The archive digest was checked against GitHub, and the packaged and extracted runtime libraries matched SHA-256 `51c2cb340a2b8021c426c1fbc3de5d8184add13cd48b64669270b63858cc03df`. A separate local native build was blocked by missing JNI headers in the default JDK; no native source changes are included in this PR.

The regressions cover independent compatibility defaults, the restoration explanation with a non-native child, a prior refusal of only the lower shuffle, a conditional diagnostic across the stage-placeholder boundary, and decimal SUM cancellation in Legacy and ANSI modes. They also verify restored Partial tags and retained native input.

Hosted validation is requested for the rebased head through `run-all-spark-profiles` and `run-spark-4.1-tests`, covering Comet suites across Spark profiles and Spark's own SQL suites. These hosted results are separate from the local results above.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
@@ -614,7 +614,7 @@ case class CometExecRule(session: SparkSession)
       // during the bottom-up conversion. Tags persist through AQE stage creation.
       tagUnsafePartialAggregates(planWithJoinRewritten)
 
-      var newPlan = transform(planWithJoinRewritten)
+      var newPlan = revertUnsafePartialAggregates(transform(planWithJoinRewritten))
 
       // if the plan cannot be run fully natively then explain why (when appropriate
       // config is enabled)
@@ -1015,6 +1015,63 @@ case class CometExecRule(session: SparkSession)
   }
 
   /**
+   * The early tagging pass cannot know whether a Final's child will become native. Check the
+   * actual conversion result as well, before native blocks are serialized or AQE launches stages.
+   * Restore only the feeding aggregate/exchange chain; keep native work below its Partial.
+   */
+  private def revertUnsafePartialAggregates(plan: SparkPlan): SparkPlan = {
+    def revertChain(node: SparkPlan): Option[SparkPlan] = node match {
+      case agg: CometHashAggregateExec if agg.modes == Seq(Partial) =>
+        val partial = agg.originalPlan.withNewChildren(Seq(agg.child))
+        partial.setTagValue(
+          CometExecRule.COMET_UNSAFE_PARTIAL,
+          "Partial aggregate disabled: corresponding final aggregate " +
+            "cannot be converted to Comet and intermediate buffer formats are incompatible")
+        Some(partial)
+
+      case agg: CometHashAggregateExec
+          if agg.modes.forall(m => m == Partial || m == PartialMerge) =>
+        revertChain(agg.child).map(child => agg.originalPlan.withNewChildren(Seq(child)))
+
+      case agg: BaseAggregateExec
+          if agg.aggregateExpressions.nonEmpty &&
+            agg.aggregateExpressions.forall(_.mode == Partial) =>
+        // This producer already emits Spark buffers. Do not reach through it to an unrelated
+        // aggregate below it.
+        None
+
+      case agg: BaseAggregateExec
+          if agg.aggregateExpressions.forall(e => e.mode == Partial || e.mode == PartialMerge) =>
+        revertChain(agg.child).map(child => agg.withNewChildren(Seq(child)))
+
+      case CometSinkPlaceHolder(_, _, shuffle: CometShuffleExchangeExec) =>
+        revertChain(shuffle)
+      case shuffle: CometShuffleExchangeExec =>
+        revertChain(shuffle.child).map(child => shuffle.originalPlan.withNewChildren(Seq(child)))
+      case shuffle: ShuffleExchangeExec =>
+        revertChain(shuffle.child).map(child => shuffle.withNewChildren(Seq(child)))
+
+      case _: ShuffleQueryStageExec | _: ReusedExchangeExec =>
+        // A stage owns (and may already have materialized) its buffers. Never rewrite it here.
+        // The whole-plan QueryStagePrep pass must tag the Partial before stages are created;
+        // that tag keeps it in Spark when the rule is reapplied to the exchange in isolation.
+        None
+      case _ => None
+    }
+
+    plan.transformUp {
+      case agg: BaseAggregateExec
+          if agg.aggregateExpressions.map(_.mode).distinct == Seq(Final) &&
+            !QueryPlanSerde.allAggsSupportMixedExecution(agg.aggregateExpressions) =>
+        revertChain(agg.child)
+          // Rebuild native consumers and shuffles from their original Spark operators. Merely
+          // replacing their children would leave a native protobuf reading the old buffers.
+          .map(child => transform(agg.withNewChildren(Seq(child))))
+          .getOrElse(agg)
+    }
+  }
+
+  /**
    * Look for the bottom Partial-mode aggregate that feeds into the given plan (the child of a
    * Final). Walks through exchanges and AQE stages, and continues down through intermediate
    * aggregate stages whose modes are all Partial / PartialMerge - these are the PartialMerge (and
@@ -1043,8 +1100,8 @@ case class CometExecRule(session: SparkSession)
   /**
    * Conservative check for whether an aggregate could be converted to Comet. Checks operator
    * enablement, grouping expressions, aggregate expressions, and result expressions.
-   * Intentionally skips the sparkFinalMode / child-native checks since those depend on
-   * transformation state.
+   * Intentionally skips the child-native checks since those depend on transformation state;
+   * [[revertUnsafePartialAggregates]] checks the actual conversion result before execution.
    *
    * WARNING: this intentionally mirrors the predicate checks in `CometBaseAggregate.doConvert`
    * (operators.scala). Any change to the convertibility rules there must be reflected here or
--- a/spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala
@@ -25,9 +25,11 @@ import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{CometTestBase, DataFrame, Row}
 import org.apache.spark.sql.catalyst.expressions.Cast
+import org.apache.spark.sql.catalyst.expressions.aggregate.Partial
 import org.apache.spark.sql.catalyst.optimizer.EliminateSorts
-import org.apache.spark.sql.comet.CometHashAggregateExec
-import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
+import org.apache.spark.sql.comet.{CometFilterExec, CometHashAggregateExec}
+import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanHelper, ShuffleQueryStageExec}
+import org.apache.spark.sql.execution.aggregate.BaseAggregateExec
 import org.apache.spark.sql.functions.{avg, col, count_distinct, sum}
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
@@ -35,6 +37,7 @@ import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
 import org.apache.comet.CometConf
 import org.apache.comet.CometConf.COMET_EXEC_STRICT_FLOATING_POINT
 import org.apache.comet.CometSparkSessionExtensions.isSpark41Plus
+import org.apache.comet.rules.CometExecRule
 import org.apache.comet.testing.{DataGenOptions, FuzzDataGenerator, ParquetGenerator, SchemaGenOptions}
 
 /**
@@ -185,6 +188,73 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
           CometConf.COMET_ENABLE_FINAL_HASH_AGGREGATE.key -> (!disableFinal).toString) {
           checkSparkAnswer(
             s"SELECT percentile_approx(_2, 0.5), count(DISTINCT _3) FROM tbl$groupBy")
+        }
+      }
+    }
+  }
+
+  for (adaptive <- Seq(false, true)) {
+    test(s"decimal AVG falls back across a Spark shuffle (AQE=$adaptive)") {
+      withTempDir { dir =>
+        val path = s"${dir.getAbsolutePath}/data"
+        withSQLConf(CometConf.COMET_ENABLED.key -> "false") {
+          spark
+            .range(0L, 8L, 1L, 4)
+            .selectExpr("id", "CAST(200 AS DECIMAL(20, 2)) AS v")
+            .write
+            .parquet(path)
+        }
+
+        withSQLConf(
+          SQLConf.ADAPTIVE_EXECUTION_ENABLED.key -> adaptive.toString,
+          SQLConf.FILES_MAX_PARTITION_BYTES.key -> "1048576",
+          SQLConf.PARQUET_FILTER_PUSHDOWN_ENABLED.key -> "false",
+          CometConf.COMET_NATIVE_SCAN_ENABLED.key -> "false",
+          CometConf.COMET_CONVERT_FROM_PARQUET_ENABLED.key -> "true",
+          CometConf.COMET_SHUFFLE_ENABLED.key -> "false") {
+          withParquetTable(path, "decimal_avg_fallback") {
+            // The filter leaves three input partitions empty. Decimal AVG is not safe to mix
+            // between engines: a native empty partial can poison the Spark final's sum buffer.
+            val df = sql("SELECT AVG(v) FROM decimal_avg_fallback WHERE id = 1")
+            val initialPlan = stripAQEPlan(df.queryExecution.executedPlan)
+            checkAnswer(df, Seq(Row(new java.math.BigDecimal("200.000000"))))
+            for (plan <- Seq(initialPlan, df.queryExecution.executedPlan)) {
+              assert(collect(plan) { case agg: CometHashAggregateExec => agg }.isEmpty)
+              val partials = collect(plan) {
+                case agg: BaseAggregateExec
+                    if agg.aggregateExpressions.forall(_.mode == Partial) =>
+                  agg
+              }
+              assert(partials.size == 1)
+              assert(partials.forall(_.getTagValue(CometExecRule.COMET_UNSAFE_PARTIAL).isDefined))
+              // Falling back the aggregate must not discard the native filter/scan conversion.
+              assert(collect(plan) { case filter: CometFilterExec => filter }.nonEmpty)
+            }
+            if (adaptive) {
+              val stages = collect(df.queryExecution.executedPlan) {
+                case stage: ShuffleQueryStageExec => stage
+              }
+              assert(stages.nonEmpty && stages.forall(_.isMaterialized))
+            }
+
+            // Compatible buffers may still use a native Partial and a Spark Final.
+            val safe = sql("SELECT MIN(v), MAX(v) FROM decimal_avg_fallback WHERE id = 1")
+            checkAnswer(
+              safe,
+              Seq(Row(new java.math.BigDecimal("200.00"), new java.math.BigDecimal("200.00"))))
+            assert(collect(safe.queryExecution.executedPlan) { case agg: CometHashAggregateExec =>
+              agg
+            }.size == 1)
+
+            // The same unsafe buffer is valid when both aggregate stages execute in Comet.
+            withSQLConf(CometConf.COMET_SHUFFLE_ENABLED.key -> "true") {
+              val native = sql("SELECT AVG(v) FROM decimal_avg_fallback WHERE id = 1")
+              checkAnswer(native, Seq(Row(new java.math.BigDecimal("200.000000"))))
+              assert(collect(native.queryExecution.executedPlan) {
+                case agg: CometHashAggregateExec => agg
+              }.size == 2)
+            }
+          }
         }
       }
     }
@@ -513,7 +583,10 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
             dictionaryEnabled) {
             val n = if (nativeShuffleEnabled) 2 else 1
             checkSparkAnswerAndNumOfAggregates("SELECT _2, SUM(_1) FROM tbl GROUP BY _2", n)
-            checkSparkAnswerAndNumOfAggregates("SELECT _2, COUNT(_1) FROM tbl GROUP BY _2", n)
+            // COUNT is not declared safe for mixed execution, unlike the other aggregates here.
+            checkSparkAnswerAndNumOfAggregates(
+              "SELECT _2, COUNT(_1) FROM tbl GROUP BY _2",
+              if (nativeShuffleEnabled) 2 else 0)
             checkSparkAnswerAndNumOfAggregates("SELECT _2, MIN(_1) FROM tbl GROUP BY _2", n)
             checkSparkAnswerAndNumOfAggregates("SELECT _2, MAX(_1) FROM tbl GROUP BY _2", n)
             checkSparkAnswerAndNumOfAggregates("SELECT _2, AVG(_1) FROM tbl GROUP BY _2", n)
@@ -721,26 +794,29 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
             val path = new Path(dir.toURI.toString, "test")
             makeParquetFile(path, 1000, 20, dictionaryEnabled)
             withParquetTable(path.toUri.toString, "tbl") {
+              // Spark rewrites _7's small decimal SUM to Long; _8 and _9 remain decimal and
+              // cannot use a native Partial when the Final runs in Spark.
               val expectedNumOfCometAggregates = if (nativeShuffleEnabled) 2 else 1
+              val expectedNumOfDecimalAggregates = if (nativeShuffleEnabled) 2 else 0
 
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT _g2, SUM(_7) FROM tbl GROUP BY _g2",
                 expectedNumOfCometAggregates)
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT _g3, SUM(_8) FROM tbl GROUP BY _g3",
-                expectedNumOfCometAggregates)
+                expectedNumOfDecimalAggregates)
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT _g4, SUM(_9) FROM tbl GROUP BY _g4",
-                expectedNumOfCometAggregates)
+                expectedNumOfDecimalAggregates)
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT SUM(_7) FROM tbl",
                 expectedNumOfCometAggregates)
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT SUM(_8) FROM tbl",
-                expectedNumOfCometAggregates)
+                expectedNumOfDecimalAggregates)
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT SUM(_9) FROM tbl",
-                expectedNumOfCometAggregates)
+                expectedNumOfDecimalAggregates)
             }
           }
         }
@@ -1325,7 +1401,9 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
             val path = new Path(dir.toURI.toString, "test")
             makeParquetFile(path, 1000, 20, dictionaryEnabled)
             withParquetTable(path.toUri.toString, "tbl") {
+              // Only _7 is rewritten to a mixed-safe Long AVG by Spark's decimal optimizer.
               val expectedNumOfCometAggregates = if (nativeShuffleEnabled) 2 else 1
+              val expectedNumOfDecimalAggregates = if (nativeShuffleEnabled) 2 else 0
 
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT _g2, AVG(_7) FROM tbl GROUP BY _g2",
@@ -1333,11 +1411,12 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
 
               checkSparkAnswerWithTolerance("SELECT _g3, AVG(_8) FROM tbl GROUP BY _g3")
               assert(getNumCometHashAggregate(
-                sql("SELECT _g3, AVG(_8) FROM tbl GROUP BY _g3")) == expectedNumOfCometAggregates)
+                sql(
+                  "SELECT _g3, AVG(_8) FROM tbl GROUP BY _g3")) == expectedNumOfDecimalAggregates)
 
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT _g4, AVG(_9) FROM tbl GROUP BY _g4",
-                expectedNumOfCometAggregates)
+                expectedNumOfDecimalAggregates)
 
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT AVG(_7) FROM tbl",
@@ -1345,11 +1424,11 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
 
               checkSparkAnswerWithTolerance("SELECT AVG(_8) FROM tbl")
               assert(getNumCometHashAggregate(
-                sql("SELECT AVG(_8) FROM tbl")) == expectedNumOfCometAggregates)
+                sql("SELECT AVG(_8) FROM tbl")) == expectedNumOfDecimalAggregates)
 
               checkSparkAnswerAndNumOfAggregates(
                 "SELECT AVG(_9) FROM tbl",
-                expectedNumOfCometAggregates)
+                expectedNumOfDecimalAggregates)
             }
           }
         }
--- a/spark/src/test/scala/org/apache/comet/rules/CometExecRuleSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/rules/CometExecRuleSuite.scala
@@ -35,6 +35,7 @@ import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
 
 import org.apache.comet.{CometConf, CometExplainInfo}
 import org.apache.comet.CometSparkSessionExtensions.{isSpark35Plus, isSpark40Plus, isSpark42Plus}
+import org.apache.comet.CometSparkSessionExtensions.withFallbackReason
 import org.apache.comet.testing.{DataGenOptions, FuzzDataGenerator}
 
 /**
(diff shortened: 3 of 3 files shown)


sunchao:
cc @andygrove @comphead discovered this correctness issue when running TPC-DS benchmark internally with `spark.comet.shuffle.enabled=false`

andygrove:
I used an LLM (Claude Code) to help with this review, including building both PRs locally and running the sweeps below. I have gone over the results myself and I agree with them, but flagging the tooling up front.

Thanks for digging into this. The gap you identified is real, and the reasoning about `canAggregateBeConverted` skipping the child-native check matches what the comment in `operators.scala` says the tagging pass is supposed to cover. Before getting into the mechanics I want to raise something about the framing.

I swept every aggregate function through the Comet-Partial plus Spark-Final configuration, using `spark.comet.exec.shuffle.enabled=false` so the Final stays in Spark, over 8 rows in 4 Parquet files with a filter that leaves three partials empty. On Spark 4.1 with JDK 17, on current main, only two of the 26 aggregates I tried are wrong across that boundary:

```
avg(bigint)          comet=[null]                      spark=[1.0]
avg(decimal(20,2))   comet=threw ARITHMETIC_OVERFLOW   spark=[200.000000]
```

With this PR applied the decimal case is fixed and `AVG(bigint)` is still wrong, exactly as before. `CometAverage.supportsMixedPartialFinal` returns true for non-decimal input, so the allowlist waves through what looks like the same defect with the same root cause. That seems like the thing to fix first, and it would be worth a test either way, since it is the same bug this PR is named after.

Everything else the PR newly blocks was already correct in my sweep. `COUNT`, decimal `SUM`, `FIRST`, `LAST`, and the whole stddev/variance/covariance/correlation family all matched Spark on main, and all of them lose their native Partial with this change. I tried to justify that cost by hunting for a case where the guard earns it: decimal `SUM` overflowing `DECIMAL(38,0)`, long `SUM` overflow, `TRY_SUM`, `TRY_AVG`, grouped and filtered variants, under ANSI on and off. All of those matched with the native Partial retained, including ANSI throwing and legacy returning null. The comment on `CometSum.supportsMixedPartialFinal` about overflow detection not surviving the split did not reproduce in this direction, which makes sense to me because that comment describes the Spark-partial to Comet-final direction rather than this one. The same goes for the `COUNT` exclusion, which the docstring attributes to `PropagateEmptyRelationAfterAQE` and Spark 4.0 count-bug decorrelation. Both of those only bite when the Final becomes a `CometHashAggregateExec`.

The other thing I noticed is that #5420 already contains `revertUnsafePartialAggregates` verbatim, so this PR is a subset of it. When I keep only #5420's native `avg.rs` and `avg_decimal.rs` changes and drop the planner guard entirely, both AVG cases are fixed and every aggregate keeps its native Partial.

So on the evidence I have, the native change in #5420 fixes strictly more than this guard does and costs no native aggregation, while this guard fixes less and gives up native partials for a fair number of aggregates. Could you say what this PR adds once #5420 lands? The description asserts the safeguard is still needed after that, but I was not able to construct a case, and none of the four new tests exercises one. If the value is defense in depth against buffer mismatches we have not found yet, rather than a bug that is broken today, that seems like a reasonable position to me, but I think it should be argued on that basis and weighed against the measured cost. It also affects the ordering of the two PRs.

Two smaller things while I was in there. `supportsMixedPartialFinal` is a single direction-agnostic flag, but most of the exclusions behind it are justified only in the Spark-partial to Comet-final direction. Would it be worth splitting it into two predicates so this pass consults only the one that applies? And where `revertChain` returns `None`, the unsafe boundary is left in the plan with no signal, which is the same failure mode the PR exists to prevent. A plain warning there would be too noisy, since I see the `None` path taken benignly on q10 and q35 where the Partial is already Spark, but gating it on `findCometPartialAgg` finding a Comet Partial that `revertChain` failed to reach would make the gap between those two traversals detectable.

One last note, unrelated to this PR. The sweep also turned up that `PERCENTILE`, `APPROX_PERCENTILE`, `COLLECT_LIST` and `COLLECT_SET` all trip the strict check with `spark.comet.exec.shuffle.enabled=false`, reporting "Comet did not convert ObjectHashAggregate but recorded no fallback reason". The shuffle-enabled guard on `ObjectHashAggregateExec` declines without calling `withFallbackReason`. I will file that separately.

sunchao:
Thanks @andygrove . You’re right that the AVG examples alone don’t demonstrate what #5421 adds after #5420’s native fixes.

I’ve now reproduced a separate failure. I kept #5420’s native library and removed only the call to `revertUnsafePartialAggregates`. On Spark 4.1.3 / JDK 17, `COLLECT_LIST` and `COLLECT_SET` crash in Spark’s `Collect.deserialize` when grouping by a stored `DECIMAL(19,0)` key, with Comet shuffle enabled, mode native, and native hash partitioning enabled. This happens with AQE both off and on. Restoring the guard makes all four cases match Spark.

The decimal key makes the native shuffle fall back, leaving a Comet partial feeding a Spark final. That partial produces an array buffer, while Spark expects serialized binary state. The shuffle-disabled sweep doesn’t exercise this path because it prevents the ObjectHashAggregate partial from converting in the first place. The existing AVG regression passes with or without the guard.

So there is an independent correctness case for the planner repair. That still doesn’t justify every exclusion in the current compatibility flag. I agree we should distinguish the two directions and preserve native partials where their buffers are compatible.

The duplication is because #5420 currently includes #5421 as a prerequisite. I’ll clarify that relationship and update #5421’s tests and description around this separate failure.

andygrove:

I read the existing thread and do not want to relitigate the "what does this add on top of #5420" question, which is already being worked out between you two. Two things I did not see raised there.

**`modes.distinct == Seq(Final)` misses the distinct-rewrite final stage**

Both `tagUnsafePartialAggregates` and the new `revertUnsafePartialAggregates` gate on `agg.aggregateExpressions.map(_.mode).distinct == Seq(Final)`. Spark's `AggUtils.planAggregateWithOneDistinct` produces a fourth stage whose modes are `Final` for the regular aggregates and `Complete` for the distinct one, so a query like

```sql
SELECT AVG(CAST(v AS DECIMAL(20,2))), COUNT(DISTINCT k) FROM t GROUP BY g
```

lands on `Seq(Final, Complete)` and matches neither guard. The `Complete` half reads raw input so it is harmless on its own, but the `Final` half still consumes the buffer produced by stage three, which is exactly the boundary this PR exists to protect. Is that shape actually safe for some reason I am missing, or should the predicate be "contains `Final`" rather than "is exactly `Final`"? The description mentions distinct aggregations with intermediate merge stages as an affected case, which is what made me look.

The new `CometExecRuleSuite` test does cover `distinct=true`, but it asserts zero `CometHashAggregateExec` in the fallback configurations, and I think that outcome comes from the pre-existing tagging plus the `missingCometProducer` cascade rather than from `revertChain` reaching the stage-four aggregate. It would be worth adding an assertion that distinguishes those two mechanisms, otherwise the test passes whether or not the new pass handles the shape.

**Re-running `transform` over a partially converted subtree**

The `transformUp` body calls `transform(agg.withNewChildren(Seq(child)))`. The `child` returned by `revertChain` is Spark operators down to the point where native work resumes, so the subtree handed back to `transform` still contains `CometFilterExec`, `CometScanExec`, and friends from the first pass. Is `transform` idempotent over already-converted operators? I would expect it to be, but the comment above the call only explains why rebuilding is necessary, not that re-entry is safe. A note there would help, and if there is any doubt, a test that puts a second unsafe `Final` above the first would exercise the nested case where `transformUp` triggers this twice on overlapping subtrees.

Related to that, is there a bound on how much work this can redo? With `transformUp`, N stacked unsafe finals means N re-conversions of progressively larger subtrees. Probably irrelevant for real plans, but worth knowing it is bounded.

sunchao:
Updated in [f1e5868d8](https://github.com/apache/datafusion-comet/commit/f1e5868d823e1d3e708f4697b53157ebc8a332e0).

I split the compatibility checks by direction and restored COUNT's safe native-partial-to-Spark-final path. The reverse direction stays disabled for the documented AQE/count-bug reasons. The new execution test covers grouped, all-null, and empty inputs with AQE on/off, and both early and post-conversion fallback. I have not enabled the remaining functions based only on the ordinary-input sweep; their buffer layout, overflow modes, and merge behavior still need targeted validation.

The PR now commits an independent `collect_list`/`collect_set` reproduction with Comet shuffle enabled but native hash partitioning disabled, using integer keys so it does not depend on #5420. Each AQE mode checks Spark-equivalent results, zero native aggregate stages, the unsafe-partial tag, and preservation of the native filter; re-enabling eligible native shuffle checks the two-native-stage control. These buffers are native arrays versus Spark's serialized binary state, so fixing AVG's empty sum/count in #5420 does not fix this boundary. The description now leads with that concrete case. #5421 remains the prerequisite, and I am synchronizing this update into #5420.

For the DISTINCT concern, Spark's one-DISTINCT rewrite has intermediate PartialMerge and mixed Partial/PartialMerge stages, but its final stage contains Final-mode expressions for both ordinary and distinct functions. I checked this across the supported Spark sources and in an actual four-stage plan; there was no Final+Complete final stage. The existing regression also exercised the new repair: the early pass added zero unsafe tags, conversion produced one native aggregate, and post-conversion repair removed it. Reapplying conversion preserved the same tree. I therefore kept the Final-only condition and did not broaden it to address an unobserved mode combination.

`revertChain(None)` intentionally stops at an already-Spark producer, materialized/reused stage, or unrelated boundary. An unconditional warning/assertion would flag those safe cases. Existing Comet nodes pass through conversion, and repair invokes conversion rather than recursively invoking the repair pass, so repeated-tree work is bounded by the finite plan. The current tests do not show an idempotence or planning-time defect. The separate pre-existing missing ObjectHashAggregate fallback explanation is tracked in https://github.com/apache/datafusion-comet/issues/5500; it is not introduced by this buffer-compatibility change.

Validation for this update: full Spark 4.1.3 JVM reactor, Spotless, and Scalastyle passed, along with 44 planner/execution tests. Local execution reused the existing OSS native library and is not a fresh native build. CI for the new commit remains to be confirmed.

spark/src/main/scala/org/apache/comet/serde/aggregates.scala
```diff
@@ -106,6 +106,10 @@ object CometMax extends CometAggregateExpressionSerde[Max] {
 }
 
 object CometCount extends CometAggregateExpressionSerde[Count] {
+  // Both buffers are a single non-null Long. The AQE/count-bug restrictions documented on the
+  // reverse direction concern a Comet Final; retaining Spark's Final preserves those rewrites.
+  override def supportsNativePartialToSparkFinal(fn: Count): Boolean = true
```
ziting-openai, line 117: [P1] Do not let the new COUNT exemption expose incorrect AVG buffers

This newly enables a mixed aggregate such as `SELECT COUNT(*), AVG(value) FROM parquet_input WHERE id = 1` to run with a Comet Partial and Spark Final: `COUNT` now returns true here and non-decimal `AVG` already claims forward compatibility. Previously `COUNT` forced the *entire* aggregate into Spark, so this query was protected. With four input partitions and only one surviving `BIGINT`, the existing native AVG defect independently demonstrated in https://github.com/apache/datafusion-comet/pull/5421#issuecomment-5429161669 makes the newly admitted plan return `AVG = NULL` instead of `1.0`. The dependent native AVG fix in #5420 is still unmerged, so landing this prerequisite first introduces silent wrong results. Please keep mixed COUNT+AVG plans in Spark until that fix lands (or mark native AVG-to-Spark-final unsafe) and add a COUNT+AVG regression with empty partial partitions.

sunchao:
Updated in [46ad7eca3](https://github.com/apache/datafusion-comet/commit/46ad7eca3c45be564d4edf065ae3821167acf954).

Fixed the new [COUNT+AVG empty-partial report](https://github.com/apache/datafusion-comet/pull/5421#discussion_r3874119048) by declining AVG NativePartialToSparkFinal until its emitted state is repaired. COUNT-only interoperability, AVG SparkPartialToNativeFinal, and fully native execution remain enabled.

Both AQE variants reproduced `(1, NULL)` instead of `(1, 1.0)` against this branch's own rebuilt native code before the guard. The final native/JVM builds, style checks, and **49 focused planner/execution tests pass**. The description now explicitly explains why the earlier run with #5420's already-repaired native library did not establish this standalone boundary. The review thread has the detailed fix and test scope; no full-suite or multi-version rerun is claimed.

spark/src/main/scala/org/apache/comet/serde/aggregates.scala
ziting-openai: Verified fixed at `46ad7eca3c45be564d4edf065ae3821167acf954`: native AVG explicitly rejects NativePartial-to-SparkFinal compatibility, the mixed aggregate requires every expression to be compatible, and four-partition COUNT+AVG regressions cover the former `(1, NULL)` result with AQE both enabled and disabled.

sunchao: Confirmed and fixed in [46ad7eca3](https://github.com/apache/datafusion-comet/commit/46ad7eca3c45be564d4edf065ae3821167acf954). With this branch's freshly rebuilt native library, both AQE variants reproduced `(COUNT=1, AVG=NULL)` instead of `(1, 1.0)` on the published policy, with a Comet Partial feeding Spark Final.

`CometAverage.supportsNativePartialToSparkFinal` now explicitly returns false until the emitted-state fix in #5420 is present. COUNT-only NativePartialToSparkFinal remains enabled. AVG's existing non-decimal SparkPartialToNativeFinal support is unchanged, as is fully native COUNT+AVG execution. This restriction belongs to the prerequisite with the old AVG implementation; it is not being copied blindly into #5420's repaired native path.

The new regressions pin four Parquet partitions and cover one surviving value, an all-null value, and no rows with AQE on/off and both early and post-conversion fallback. They check the fallback tag and retained native filter, plus a two-stage native control. I updated the existing AVG plan expectations rather than leaving tests that claimed this unsafe direction was supported.

The exact native build and full Spark 4.1.3 JVM reactor/style checks pass, as do all **49 focused tests** (31 planner, 18 execution). The prior local run reused #5420's already-repaired native library and did not cover this standalone COUNT+AVG boundary; the new before/after runs use this branch's native source with the loaded library hash verified. No full-suite or multi-version run is claimed.

ziting-openai (approved):
Reviewed commit 46ad7eca3c45be564d4edf065ae3821167acf954: no P1 or significant P2 findings. Source review covered the changed execution and compatibility boundaries and the focused regression cases; runtime tests were not run.

Posted by Codex on behalf of ziting-openai using spark-pr-review-memo.

ziting-openai (approved):
Re-reviewed commit `c078ef136add779662b0e74d6bb7035869799284`: no P1 or significant P2 findings. The production code is unchanged from the previously approved head. The six added AQE-off/on cases cover the reported array-key fallback for percentile and collect_list, retain a safe native SUM partial, and inspect both initial and materialized plans.

Source review only; runtime tests and cross-version builds were not rerun. Hosted CI was still in progress at review time.

Posted by Codex on behalf of ziting-openai using spark-pr-review-memo.

ziting-openai:
I'm sorry for the unsolicited AI reviews on this PR. The approvals I posted on August 29 ([first review](https://github.com/apache/datafusion-comet/pull/5421#pullrequestreview-5056269476), [second review](https://github.com/apache/datafusion-comet/pull/5421#pullrequestreview-5059046592)) were the result of a misconfigured AI agent that should have been limited to OpenAI-owned repositories. Please disregard those approvals. I've corrected the configuration to prevent further reviews on non-OpenAI repositories. Neither approval opened an inline review thread, and I don't have permission to dismiss the approvals myself. Sorry for the noise.

comphead:
Checking it

comphead:
> **Note on this review:** generated with LLM assistance (Claude Code) while I worked through the PR. I have read the findings and agree with them, but flagging the tooling. Push back on anything that is wrong or already handled.

Four things at the Major level. I checked the two load-bearing claims first and both hold up, so none of this is about the diagnosis:

- The AVG `(null, 0)` claim is correct. `native/spark-expr/src/agg_funcs/avg.rs:143-152` has `sum: Option<f64>` and `state()` returns `ScalarValue::Float64(self.sum)`, while Spark's `Average.mergeExpressions` is the null-intolerant `sum.left + sum.right` over an initial `0.0`.
- The COUNT direction argument is correct. Both rules the old docstring cites key on the **Final** being a `BaseAggregateExec`: `LogicalQueryStage.scala:58` (`computeStats` bumps a 0 row count to 1 for a global aggregate) and `AQEPropagateEmptyRelation.scala:58`. Keeping the Final in Spark preserves both, and the Partial's engine does not participate in either match.

### 1. The direction split defaults one direction to the other

`supportsNativePartialToSparkFinal(fn) = supportsSparkPartialToNativeFinal(fn)` in `CometAggregateExpressionSerde.scala`. The premise of the split is that the two directions are independent, and AVG is the proof. With this default, non-decimal SUM, MIN, MAX, the three bitwise aggregates, HLL++ and `bloom_filter_agg` all claim forward-direction safety on the strength of an analysis that was only ever done for the reverse direction.

The AVG defect is a per-function property, namely the state a never-updated accumulator emits, and nothing forces that question to be answered when a new aggregate opts in. I checked the least obvious one and it happens to be fine: `spark_bloom_filter.rs:282` makes `state_as_bytes()` an alias for `spark_serialization()`, and `bloom_filter_agg.rs:171` always emits `Binary(Some(..))`, so an empty partition does not produce the null that Spark's `BloomFilterAggregate.deserialize` would NPE on (it calls `BloomFilter.readFrom` with no null check). That reads as luck rather than design.

Suggestion: default `supportsNativePartialToSparkFinal` to `false` and add explicit `= true` overrides for the functions that were actually checked, recording the empty-partition state in each comment. That is a handful of one-line overrides, and it makes the next contributor answer the right question instead of inheriting an answer to a different one.

### 2. `revertChain` bails out silently, leaving the plan it exists to prevent

The trailing `case _ => None` in `revertChain` feeds `getOrElse(agg)`, so an unrecognised node between the Spark Final and the native Partial ships the unsafe boundary with no signal at all.

I do not think the earlier reply covers the proposal that was made. An unconditional warning would indeed fire on the benign q10/q35 shape where the Partial is already Spark, but gating the signal on `findCometPartialAgg(agg.child)` returning a Comet Partial that `revertChain` failed to reach is false on exactly those cases. The two traversals accept different node sets: `findCometPartialAgg` handles `AQEShuffleReadExec` and `ShuffleQueryStageExec`, `revertChain` handles `CometSinkPlaceHolder` and `CometShuffleExchangeExec`. Any future divergence between them is a live crash, and today it is undetectable.

### 3. Fourth overlapping mechanism for one invariant

The Comet-Partial-to-Spark-Final boundary is now guarded in four places: `tagUnsafePartialAggregates`, `preserveSparkAggregateBuffers`, the new `revertUnsafePartialAggregates`, and `RevertNativeForTransitionHeavyStages.hasUnsafeMixedAggregateAtStageBoundary`.

All of them are still needed as written, since the Celeborn one also covers a Comet Final over a fallen-back Comet exchange, which the new pass does not match (it only matches a Spark `BaseAggregateExec` Final). But the first three duplicate the chain walk and the reason string, and they descend through different node sets, which is the same drift risk as finding 2. At minimum `preserveSparkAggregateBuffers` should call the shared `revertChain` rather than keeping its own recursive `restore`.

### 4. The reverted Partial gets the tag but no fallback reason

`revertChain` calls `setTagValue(COMET_UNSAFE_PARTIAL, ...)` without the matching `withFallbackReason`, unlike the sibling at `CometExecRule.scala:228-229`. The reason is only recovered when the re-entrant `transform` reaches `doConvert`, which requires the Partial's children to all be `CometNativeExec`. That holds in every new test, since there is always a `CometFilterExec` below, so the gap is untested. Where the Partial's child is not native, `explainFallback` and `EXTENDED` output silently lose the explanation for the one node this PR changes. One line next to the `setTagValue`.

sunchao:
@comphead updated in [6c14796f5](https://github.com/apache/datafusion-comet/commit/6c14796f5146b5b9908bb8d9b7133eabe9765c96).

1. **Independent defaults:** `supportsNativePartialToSparkFinal` now defaults to `false`. The eight previously inherited opt-ins are explicit and document their empty/all-null state. Existing admission decisions are preserved, including SUM's decimal/TRY exclusions and COUNT/AVG's directional differences. A policy test verifies that opting into Spark-state consumption cannot implicitly enable native-state production.

2. **Unrepaired paths:** you are right that my earlier reply addressed unconditional warnings rather than the conditional diagnostic proposed. When repair cannot reach a remaining incompatible native producer, it now warns and records the reason on the Spark Final. The read-only check handles AQE/reused stages and attribute-preserving unary wrappers, while stopping at already-Spark producers and completed aggregates. It does not rewrite stage data or infer buffer forwarding through arbitrary branching/changed-output operators. Tests cover these boundaries and ensure the warning is not repeated when the rule runs again, with fallback logging both enabled and disabled.

3. **Shared restoration:** Celeborn and post-conversion repair now use the same helper to restore a Partial, retain its current children, tag it, and record its explanation. I kept their walks separate: Celeborn must restore native ancestors such as Sort and stop at exchanges, whereas `revertChain` crosses unmaterialized exchanges and follows aggregate chains. Directly substituting that traversal would break the existing Sort-ancestor regression. The early tagging and later transition-reversion checks also retain their distinct planning responsibilities.

4. **Immediate explanation:** the shared helper records `withFallbackReason` alongside `COMET_UNSAFE_PARTIAL`. A synthetic non-native-child regression verifies that the explanation does not depend on another Partial serde call.

Validation: 208 tests passed on Spark 3.4.3/JDK 17 and 123 on Spark 4.1.3/JDK 21; compilation, Spotless, Scalastyle, and diff checks passed. These runs reused a native library built from identical native sources, not a fresh native rebuild. CI for the new commit is pending.

CI-list follow-up: [8d610448c](https://github.com/apache/datafusion-comet/commit/8d610448c7b12e8c0b976fa9d872a37d80b5fd6a) places the new suite beside the aggregate suite to avoid an overlapping insertion from main. The suite entries and tested production/test sources are unchanged.

andygrove (commented):
The new post-conversion pass only fires on a Spark `BaseAggregateExec` whose modes are exactly `Seq(Final)` (`CometExecRule.scala` around line 1224), but `tagUnsafePartialAggregates` counts `modes.contains(PartialMerge)` as a buffer consumer too (around line 1094). In a one-distinct rewrite the stage that reads the bottom Partial's buffer is a `PartialMerge`, and `canAggregateBeConverted(stage, PartialMerge)` returns true for a plain `AVG` there, so the early pass leaves it untagged. If that `PartialMerge` then stays in Spark because its own child exchange fell back, `revertChain` never runs against it and `hasUnrepairedNativePartial` is not consulted either, since it is only reached from the `Final` branch at line 1231.

What lets the Final above it convert to Comet in that shape is that `missingCometProducer` in `CometBaseAggregate.doConvert` (`operators.scala` around line 1659) checks `aggsNotSupportingSparkPartialToNativeFinal`, which is the other direction. Non-decimal `AVG` passes that check while now being unsafe in the direction this PR added. The new `CometExecRuleSuite` distinct case uses `COUNT(DISTINCT name)`, and `COUNT` is excluded from `supportsSparkPartialToNativeFinal`, which is exactly what keeps the Final in Spark and lets the repair work. Would the same case with `SUM(DISTINCT ...)` or `MIN(DISTINCT ...)`, where both directions are allowed, still come out fully repaired?

I could not construct a query that actually lands a Spark `PartialMerge` above a Comet Partial after conversion, so there may be something upstream that makes it impossible. If so, a sentence on the `revertUnsafePartialAggregates` docstring explaining why a `Final`-only trigger is sufficient would save the next reader the same trace.

Separately, my earlier points all look settled to me. The direction split now defaults `supportsNativePartialToSparkFinal` to false with per-function opt-ins, `AVG` is excluded until #5420, and the unreached-producer case records a diagnostic instead of failing silently. I checked the `HyperLogLogPlusPlus` opt-in against `native/spark-expr/src/agg_funcs/hll_plus_plus.rs`, where the words vector is initialised to zeros and `state()` emits `ScalarValue::Int64(Some(w))`, so the empty-partition claim in that comment holds.

sunchao:
Thanks @andygrove. Updated in [cb22f68c4](https://github.com/apache/datafusion-comet/commit/cb22f68c4a13ae3df2d57243ec0d50b842afb627) with global and grouped `AVG(id), SUM(DISTINCT id)` planner tests covering 18 configurations.

With native-only shuffle, a refused lower exchange keeps the intermediate merges and Final in Spark, so repair applies without relying on COUNT’s reverse-direction restriction.

Your broader concern does identify a limit: a synthetically forced lower-only refusal in `auto`/`jvm` mode can leave a native Partial below Spark PartialMerge while the Final becomes native. The Final-only repair does not inspect that boundary. I documented this and added assertions for the plan and repeat-rule stability.

I still don’t have a naturally occurring SQL trigger or wrong-result reproduction for that shape. Its bottom AVG groups by the DISTINCT key, even for the global query, so it does not emit the scalar accumulator’s untouched `(null, 0)` state. The new cases are planner checks, and the production repair predicate remains unchanged.

Five selected suites passed on both versions: 215 tests on Spark 3.4 and 219 on Spark 4.1. Fresh hosted CI is running.

comphead:
Thanks @sunchao doing a second round

comphead:
> **Note on this review:** generated with LLM assistance (Claude Code) while I worked through the second round. I have read the findings and agree with them, but flagging the tooling. Push back on anything that is wrong or already handled.

Second pass, against `fe66f74643`. My four points from the last round are addressed in the code and the direction split reads well now. The per-function opt-in comments each state the empty-partition state, which is the right question to be making a contributor answer, and the two I spot-checked hold up. Sharing `restoreSparkPartial` with the Celeborn path while keeping the two walks separate is a good resolution to the duplication point, and having the new execution tests check both the initial and the materialized AQE plan is exactly where this class of bug hides.

Three things at the Major level. Nothing I found makes results wrong, so this is about scope, diagnostics, and coverage rather than the fix itself.

Two of the three concern `fe66f74643`. That commit has an empty body and no comment on the thread, and it removes several things that were added in `6c14796f5` and `cb22f68c4` in response to the last round. I am raising the underlying gaps below rather than the commit, but if the removals were deliberate I would like to understand the reasoning, and if the rebase ate them then two of these resolve themselves.

### 1. The fallback is wider than the evidence, and nothing tracks the remainder

Outside the nine opt-ins, every aggregate now loses its native Partial whenever the Final stays in Spark. The sharpest case is decimal `SUM`. `CometSum.supportsNativePartialToSparkFinal` (`aggregates.scala:290`) and `supportsSparkPartialToNativeFinal` (`aggregates.scala:294`) are now character-for-character the same expression with different comments, and the new one excludes decimal without saying why. The reverse-direction comment blames overflow detection not surviving the split, which @andygrove already noted describes a Comet Final rather than this direction.

Looking at the native side, `sum_decimal.rs:112-120` declares `(sum: nullable decimal, is_empty: Boolean non-null)` matching Spark's `(sum, isEmpty)`, and `state()` at line 294 always emits both, so an empty partition produces `(null, true)`. Spark's decimal merge is `coalesce(coalesce(sum.left, zero) + sum.right, sum.left)` with `isEmpty.left && isEmpty.right`, which absorbs that cleanly. The cost is visible in this PR's own diff: `SUM(_8)` and `SUM(_9)` in `CometAggregateSuite` move from `expectedNumOfCometAggregates` to `expectedNumOfDecimalAggregates`, so 1 native aggregate becomes 0 when `nativeShuffleEnabled` is false, and that test was already asserting results as well as plan shape. Is there a failure mode in this direction I am missing, or can decimal SUM opt in?

The same question applies to the tail. `FIRST`, `LAST`, and the whole stddev/variance/covariance/correlation family were measured as matching Spark on main in this configuration in @andygrove's earlier sweep, and `variance.rs:100-117` emits `(count, mean, m2)` as three `Float64` values against Spark's `(n, avg, m2)` three doubles. I am not claiming those are safe, only that nobody has checked and the conservative default now costs a native Partial for all of them, plus `MinBy`/`MaxBy`, `Mode`, and `Percentile`. #5420 restores AVG but does not touch any of these. Could you file an issue to audit the remaining functions' emitted state for this direction and link it here, so that the default is a staging point rather than the end state? It would also help for the description to say plainly that this is a measurable native-aggregation regression for `spark.comet.exec.shuffle.enabled=false`, since that is the configuration the original report came from.

### 2. `revertChain` is back to bailing out silently

`CometExecRule.scala:1176` is `case _ => None` and line 1187 is `.getOrElse(agg)`, so an unrecognised node between the Spark Final and the native Partial ships the unsafe boundary with no signal. The conditional diagnostic added in `6c14796f5` was removed in `fe66f74643`.

The traversal sets still diverge, and I can name a shape that reaches this. `CometExchangeSink.createExec` builds `CometSinkPlaceHolder(nativeOp, op, op)` where `op` is the `ShuffleQueryStageExec` (`CometSink.scala:164-165`), so

```
HashAggregateExec(Final)                        // Spark, function outside the allowlist
  CometHashAggregateExec(PartialMerge)          // native, matches revertChain's second arm
    CometSinkPlaceHolder(_, _, ShuffleQueryStageExec(...))
```

matches the `Partial`/`PartialMerge` arm, recurses into the placeholder, falls to `case _ => None`, and leaves the native `PartialMerge` feeding the Spark Final with nothing logged and nothing on `EXPLAIN`. Refusing to rewrite a stage there is right. I could not construct a query that reaches it and I do not think you could either, which reads to me as the argument for the diagnostic rather than against it. Any future divergence between `findPartialAggInPlan` and `revertChain` is a live crash and is currently undetectable.

### 3. Coverage the removed tests were carrying

Three gaps, all opened by `fe66f74643`:

- Nothing covers `restoreSparkPartial` recording its reason when the restored Partial's child is not native. Every remaining test has a `CometFilterExec` or `CometProjectExec` below, so the explanation is always recovered by the re-entrant `transform` reaching `doConvert`. The `withFallbackReason` line that makes it independent of that is the one thing the last round asked for here, and it is now untested.
- Nothing pins the direction-split invariant. Opting a function into `supportsSparkPartialToNativeFinal` grants nothing in the other direction today, but no test fails if the two are re-collapsed.
- `unsafe aggregate buffers fall back when native shuffle is ineligible` lost its `"prior shuffle fallback"` case, so it now varies only `COMET_SHUFFLE_NATIVE_HASH_PARTITIONING_ENABLED` and no longer exercises a sticky prior refusal.

Related, `revertUnsafePartialAggregates` went from `private[rules]` back to `private`, so the pass can no longer be driven directly from a test. Could that go back along with the cases above?

---

CI is fully green across all five Spark profiles, Iceberg, the Spark SQL suites, and TPC-DS/TPC-H verification, so none of this is about the build.

comphead (approved):
Thanks @sunchao overall its lgtm, left some nits

sunchao:
Thanks @comphead. Addressed these in [4b5f1704d](https://github.com/apache/datafusion-comet/commit/4b5f1704dab8790363bd3ef1a90a5ed57d51ef10).

The removals in `fe66f74643` were deliberate simplification, rather than a rebase loss. They removed useful diagnostics and coverage along with the larger test matrix. This update restores the focused cases you identified while keeping the simpler structure.

1. **Decimal SUM and the remaining audit:** decimal SUM cannot opt in yet despite its matching buffer layout. With one ordered partition of `DECIMAL(38,38)` values `0.6, 0.6, -0.6`, Spark's generated scalar SUM returns `0.6` in both Legacy and ANSI. A control using the matching native library and asserting native Partial/Final stages returns NULL in Legacy and throws SUM overflow in ANSI: the native update makes the intermediate precision overflow sticky or throws before cancellation. The new regression verifies that the mixed plan keeps the Partial in Spark, returns `0.6`, records the fallback tag, and retains native input work. One small source correction: the untouched native scalar state is `(0, true)`, not `(null, true)`; the exclusion is about update semantics. I filed [#5975](https://github.com/apache/datafusion-comet/issues/5975) for the remaining forward-compatibility audits and updated the description to state the loss of native aggregation coverage with shuffle disabled, including decimal SUM going from one native stage to zero.

2. **Conditional diagnostic:** restored a read-only check along the known aggregate/exchange/AQE path. If repair fails with an incompatible native Partial or PartialMerge remaining, it warns once and attaches an EXPLAIN reason to the Spark Final. It covers the stage-placeholder shape you described without rewriting existing stages. Tests cover repeat passes, both fallback-logging settings, and benign boundaries. This is synthetic planner coverage; I am not claiming a naturally occurring SQL reproduction for that shape.

3. **Focused coverage:** restored tests for the immediate reason with a non-native child, independent direction defaults, and a sticky prior refusal of the lower shuffle. `revertUnsafePartialAggregates` is again `private[rules]` so the diagnostic can be tested directly.

Validation: 107 tests passed on Spark 4.1.3 and 105 on Spark 3.4.3, with two expected version-specific cancellations on 3.4. Compilation and style checks passed on both. Runtime checks used the refreshed [Linux CI native artifact](https://github.com/apache/datafusion-comet/actions/runs/34914245755/attempts/2), whose native source tree matches this update; the separate decimal SUM control also verified the loaded library. [CI for the new commit](https://github.com/apache/datafusion-comet/actions/runs/35054516968) is running.

andygrove (commented):
Your answer on the `PartialMerge` boundary settles it for me. I checked `planAggregateWithOneDistinct` on 3.4.4, 3.5.9, 4.0.4 and master, and step 4 uses `Final` for both the distinct and the non-distinct functions on all of them, so the `modes.distinct == Seq(Final)` trigger does fire for distinct rewrites and the gap you documented really is the synthetic one. The per-function opt-in comments each state the empty-partition state, which is the right question to make a contributor answer, and the ones I spot-checked hold.

One thing came out of your decimal SUM control that I think is bigger than this PR. The exclusion comment reads as a buffer-compatibility restriction, but the behaviour it describes is a wrong answer with the aggregate running entirely in Comet. Spark's `Sum.add` uses `DecimalAddNoOverflowCheck` for decimals, and the comment on that expression says the overflow check is skipped because writing the buffer to an `UnsafeRow` will do it. An ungrouped whole-stage-codegen aggregate keeps the buffer in local variables and never writes an `UnsafeRow`, so Spark never checks and `Decimal.$plus` retains the wide intermediate. Meanwhile `SumDecimalAccumulator::update_single` in `sum_decimal.rs` validates the precision on every update and latches the failure. That is exactly your `DECIMAL(38,38)` `0.6, 0.6, -0.6` result, and it does not need a mixed plan to appear. The same reasoning applies to a decimal `SUM` over an ever-expanding window frame, where Spark's `AggregateProcessor` buffers into a `SpecificInternalRow` and also skips the check, and to `ObjectHashAggregate`, which buffers the same way.

Could that get its own correctness issue? #5975 is shaped as an enhancement about recovering native partial coverage, and the new `CometAggregateSuite` case deliberately disables the native path, so nothing currently tracks the wrong answer itself. It also reads oddly next to #5420, which fixes precisely this class for decimal AVG by falling back once the sum precision reaches 38, while decimal SUM keeps running natively and returning NULL.

Separately, could you add the `run-all-spark-profiles` label before this merges? This changes which aggregates convert and rewrites expectations across `CometAggregateSuite`, but the PR tier only runs Spark 4.1. You ran 3.4 and 4.1 locally, so it is mostly insurance on 3.5 and 4.0. I am happy to approve once that comes back green.

andygrove:
Filed the decimal SUM divergence as #6002, so nothing here is blocked on it. The exclusion in `CometSum.supportsNativePartialToSparkFinal` is the right call either way.

PR #5440, opened 2026-08-23
perf: use Arrow cast for decimal rescale check

### LINKED ISSUE #5094: decimal_rescale_check: replace fused rescale and precision check with arrow decimal cast

`native/spark-expr/src/math_funcs/internal/decimal_rescale_check.rs` hand-rolls a fused Decimal128 rescale plus precision check: checked multiply on scale-up, add-half-then-divide HALF_UP on scale-down, bounds check against `10^p - 1`, an `i128::MAX` sentinel plus `null_if_overflow_precision` two-pass dance in legacy mode, and an error in ANSI mode.

Arrow's decimal-to-decimal `cast_with_options` implements the same thing in one pass: it rescales, rounds half-away-from-zero on scale reduction (matching HALF_UP for both signs), validates target precision, and either nulls (`safe: true`) or errors (`safe: false`) on overflow, with no sentinel needed.

Replace `rescale_and_check` with `cast_with_options(&array, &Decimal128(p_out, s_out), &CastOptions { safe: !fail_on_error, .. })`, keeping a thin wrapper to map the ANSI error (the current implementation also raises a generic compute error there, so no fidelity is lost).

### Additional context

- The file exists as a deliberate fusion of `CheckOverflow(Cast(...))` into one pass, but arrow's cast kernel is itself a single pass doing both jobs; benchmark before switching (`spark-expr/benches/decimal_rescale.rs` exists).
- The `i128::MAX` sentinel aliasing is not user-visible today (a legitimate value of `i128::MAX` always exceeds precision 38 bounds and would be nulled anyway), but removing the sentinel simplifies reasoning.

peterxcli:

Closes #5094.

## Rationale for this change

`DecimalRescaleCheckOverflow` duplicates Arrow's Decimal128 cast kernel, including rescaling, HALF_UP rounding, precision validation, and legacy overflow nulling. Reusing Arrow's single-pass implementation removes the duplicate arithmetic and sentinel/null-masking path while preserving legacy and ANSI behavior.

- Delegate Decimal128 rescaling and precision checks to `ColumnarValue::cast_to` with Arrow `CastOptions`.
- Remove the hand-written rescale, precision-bound, overflow-sentinel, and null-masking logic.
- Add coverage for scale-down rounding that crosses the output precision boundary in legacy and ANSI modes.
- Update the expression audit documentation for the Arrow-backed implementation.

- `cargo test -p datafusion-comet-spark-expr decimal_rescale_check --lib` (15 passed)
- `git diff --check`
- `cargo bench -p datafusion-comet-spark-expr --bench decimal_rescale -- --quick`

The quick Criterion run was used as a regression screen. Point estimates improved in all measured shapes, although the quick-mode samples were not statistically significant:

| Shape | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| Scale up, no overflow | 241.92 us | 39.708 us | 6.09x |
| Scale up, no overflow, nulls | 480.80 us | 45.135 us | 10.65x |
| Scale down, no overflow | 999.03 us | 42.780 us | 23.35x |
| Sparse overflow | 374.10 us | 36.719 us | 10.19x |
| Dense overflow | 364.63 us | 38.451 us | 9.48x |
| ANSI scale up, no overflow | 155.48 us | 69.466 us | 2.24x |

### DIFF
--- a/docs/source/contributor-guide/expression-audits/math_funcs.md
+++ b/docs/source/contributor-guide/expression-audits/math_funcs.md
@@ -121,9 +121,7 @@ Internal decimal wrapper emitted around every decimal `+ - * /`, `sum`, and `avg
 
 ## DecimalRescaleCheckOverflow (internal)
 
-Internal fused expression that rescales a Decimal128 value (changing scale) and checks output precision in one pass, replacing the `CheckOverflow(Cast(expr, Decimal128(p, s)))` pattern used by decimal-to-decimal casts. Native impl: `math_funcs/internal/decimal_rescale_check.rs`.
-
-- Performance (tuned 2026-07-15, PR [#4938](https://github.com/apache/datafusion-comet/pull/4938)): the legacy path ran `null_if_overflow_precision` (a second full pass that allocates a new array) on every batch to turn overflow sentinels into nulls, even when nothing overflowed. Now that pass runs only when a sentinel is present (`contains(&i128::MAX)`, short-circuiting), so the common no-overflow case skips the allocation. 8 to 26% faster on no-overflow shapes; overflow and ANSI shapes unchanged. Benchmark: `benches/decimal_rescale.rs`.
+Internal fused expression that replaces the `CheckOverflow(Cast(expr, Decimal128(p, s)))` pattern used by decimal-to-decimal casts. It delegates rescaling, HALF_UP rounding, and output-precision validation to Arrow's single-pass decimal cast kernel. Arrow writes nulls directly in legacy mode and returns an error in ANSI mode. Native impl: `math_funcs/internal/decimal_rescale_check.rs`; benchmark: `benches/decimal_rescale.rs`.
 
 ## e
 
--- a/native/spark-expr/src/math_funcs/internal/decimal_rescale_check.rs
+++ b/native/spark-expr/src/math_funcs/internal/decimal_rescale_check.rs
@@ -20,11 +20,10 @@
 //! Replaces the pattern `CheckOverflow(Cast(expr, Decimal128(p2,s2)), Decimal128(p2,s2))`
 //! with a single expression that rescales and validates precision in one pass.
 
-use arrow::array::{as_primitive_array, Array, ArrayRef, Decimal128Array};
-use arrow::datatypes::{DataType, Decimal128Type, Schema};
-use arrow::error::ArrowError;
+use arrow::compute::CastOptions;
+use arrow::datatypes::{DataType, Schema};
 use arrow::record_batch::RecordBatch;
-use datafusion::common::{DataFusionError, ScalarValue};
+use datafusion::common::DataFusionError;
 use datafusion::logical_expr::ColumnarValue;
 use datafusion::physical_expr::PhysicalExpr;
 use std::hash::Hash;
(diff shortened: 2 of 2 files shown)


sunchao (approved):
## Summary

First review of `a2f10a54ca137504a5320d1bf05647c26fdee4ca` against `bba47448ecddca5d6bf62e2ca843b64703cf734d`. After five independent scopes covering semantics, compatibility/edge cases, routing/resources, tests/builds, and discussion deduplication, I found no verified, introduced [P1] or [P2] issue. Approving the implementation with the validation limitations below; the full JVM CI matrix is not green.

## Prior state and problem

`DecimalRescaleCheckOverflow` duplicated decimal scaling, HALF_UP rounding, precision bounds, and overflow handling already implemented by Arrow. Its legacy array path wrote an `i128::MAX` sentinel, scanned for that sentinel, and conditionally performed another null-masking pass. Maintaining the custom array/scalar arithmetic alongside Arrow added both code and work on this path.

## Design approach

The evaluator retains its Decimal128 type guard and delegates conversion to `ColumnarValue::cast_to` with the requested output precision/scale. `safe: !fail_on_error` maps legacy overflow to null and ANSI overflow to an error. This reuses the existing Arrow 58.4.0/DataFusion 54.1.0 dependencies without changing planner eligibility or adding dependency requirements.

For valid input coefficients in the previously supported scale-delta range, the checked scale-up and signed HALF_UP scale-down paths agree with the removed implementation. Arrow's strict downscale fast-path condition still validates cases where rounding gains a digit. Scalar/array parity, nulls, non-byte-aligned slices, target metadata, child errors, and shared-buffer lifetimes were also checked. Metadata-dependent fast paths assume coefficients satisfy their declared source precision; synthetic violations were reproduced but no supported producer into this fusion was established.

Validation evidence:

- A local harness importing the unchanged expression passed all 15 tests; all 146 resolved registry dependency versions/checksums match this PR's lockfile. Separate probes, rerun during reconciliation, covered 167,504 metadata shapes / 31,498,960 input rows, 11,532 scalar/array comparisons, sliced/null arrays, and ownership/error cleanup. These are focused expression probes, not end-to-end Spark execution.
- The [Rust CI job](https://github.com/apache/datafusion-comet/actions/runs/32641493857/job/97199409870) ran 900 tests successfully, with 4 skipped, including all 15 affected tests. Workspace Clippy and bench compilation passed; Linux/macOS native builds passed. The CI merge commit's tree exactly matches the reviewed head. Local formatting and whitespace checks passed.
- All 35 failing main-CI jobs stop at the same preexisting test-compilation error: `CometAggregateSuite.scala:318` references absent `CometConf.COMET_EXEC_SHUFFLE_ENABLED`. The relevant test/config/shim blobs are identical at base and head. These failures provide no JVM runtime parity evidence. The separate [Delta build gate](https://github.com/apache/datafusion-comet/actions/runs/32641493734/job/97199156591) also failed, but suppresses the underlying Maven diagnostic; its cause is unproven. Benchmark timing estimates were not independently rerun.

## Key design decisions

The fusion remains limited to `CheckOverflow(Cast(Decimal128))` where the cast target matches the check's target; ordinary arithmetic overflow wrappers remain in place. Evaluation now derives source precision/scale from the evaluated value, consistent with the child's type contract. Identity/widening conversions may share immutable Arrow buffers, while rescaling allocates output values; the ownership probes found no premature release or retained input reference on the tested error paths.

## Implementation sketch

The patch removes the custom precision-bound helper, rescale helper, separate array/scalar branches, and sentinel/null-masking machinery. The new regression exercises `99.4`, `99.5`, and `-99.5` converted to precision 2, scale 0: legacy returns `99`, null, null, while ANSI reports overflow. Existing tests remain registered and executed, and the audit documentation is updated to describe the Arrow-backed implementation rather than the removed masking optimization.

Legacy null-versus-ANSI-error behavior is preserved for the reviewed valid-input domain, but low-level Arrow error variants/messages can change. The old fused evaluator already used generic Arrow errors rather than ordinary `CheckOverflow`'s structured Spark error/query-context path; this patch does not add that missing context. Large valid downward scale changes can now round to zero instead of the old blanket scale-delta error. Extreme negative-scale release behavior was not exercised end-to-end, and the reported quick benchmark speedups are not independently certified here.

## Suggested improvements

No in-scope [P1] or [P2] code correction is requested. The remaining validation work is to repair the existing Scala test-compilation blocker separately, restore the JVM/Spark matrix, and obtain the Delta gate's underlying diagnostic before treating those checks as successful integration evidence. Any stronger performance claim should use repeated benchmark measurements; this approval does not certify the quick-mode point estimates.

sunchao:
Merged, thanks @peterxcli !

PR #5443, opened 2026-08-23
fix: support wide years in native make_date

### LINKED ISSUE #5208: Native make_date rejects Spark-valid years outside chrono range
## Describe the bug

Native `make_date` uses chrono `NaiveDate`, which supports years from `-262143` through `262142`. Spark accepts wider years when their epoch-day value fits `DateType`, so Comet returns `NULL` with ANSI mode disabled or throws with ANSI mode enabled for dates Spark accepts.

For invalid dates outside chrono's range, Comet can also produce a factually incorrect error message. For example, it reports that year 300000 is not a leap year even though it is divisible by 400.

## Steps to reproduce

Run:

```sql
SELECT make_date(300000, 6, 15); -- Comet rejects this Spark-valid date outside the chrono year range
SELECT make_date(300000, 2, 29); -- Comet also incorrectly reports that 300000 is not a leap year
```

Spark returns `+300000-06-15` and `+300000-02-29`. Native Comet returns `NULL` in non-ANSI mode and throws in ANSI mode. The ANSI error for the second query incorrectly says that 300000 is not a leap year.

## Expected behavior

Native Comet should match Spark for dates representable by Spark `DateType` and should not produce factually incorrect error messages.

## Additional context

Found while reviewing #5167. The chrono range limitation predates that pull request; #5167 only fixes ANSI exception fidelity.

peterxcli:

Closes #5208.

## Rationale for this change

Native `make_date` uses chrono date construction, whose supported year range is narrower than Spark `DateType`. This causes Comet to reject Spark-valid dates such as `300000-06-15` and `300000-02-29`.

- Reuse the existing proleptic-Gregorian `ymd_to_epoch_day` helper and narrow its result to Arrow `Date32`.
- Remove the resolved chrono-range compatibility note.
- Add Rust unit coverage and native SQL parity coverage with ANSI mode both disabled and enabled.

- `cargo test --manifest-path native/Cargo.toml -p datafusion-comet-spark-expr datetime_funcs::make_date::tests` (5 passed)
- `make core`
- `./mvnw test -Dtest=none -Dsuites='org.apache.comet.CometSqlFileTestSuite make_date' -Dscalastyle.skip=true` (2 passed)
- `git diff --check`


peterxcli:
cc @andygrove PTAL, thanks!

sunchao (approved):
## Summary

Reviewed head `6bd7db5e57cf66fe05568b0fde046cd0c874eac8` against observed base `bba47448ecddca5d6bf62e2ca843b64703cf734d`, with the PR diff taken from merge base `2699f59b71788e17a2714910e166a3f83deed937`. Five independent review scopes covered calendar correctness, Spark compatibility and adversarial boundaries, routing/resource safety, tests/CI, and live-discussion deduplication. No verified PR-introduced P1/P2 finding remained after reconciliation; approving this code change with the validation limitations below. The overall CI matrix is not green.

## Prior state and problem

Native `make_date` constructed a chrono `NaiveDate`, whose year range is narrower than the dates representable by Spark's signed epoch-day storage. Consequently, valid inputs such as year 300000 produced NULL or an ANSI error, including the incorrect non-leap-year diagnosis for `300000-02-29` described in #5208. The problem was the native constructor's range, not Spark's ability to represent these dates.

## Design approach

The patch reuses the existing proleptic-Gregorian `ymd_to_epoch_day` helper instead of maintaining another calendar implementation. It validates the calendar components using widened integers, then accepts the result only when checked conversion to `i32` succeeds. This extends successful construction to the Date32 boundaries without wrapping an out-of-range epoch into another date.

The reviewed arithmetic handles BCE/year zero, Gregorian century rules, wide positive and negative years, and both inclusive Date32 endpoints. Invalid month/day handling, null propagation, and the existing ANSI error path are preserved. Removing `getCompatibleNotes` changes generated documentation, not runtime admission or fallback: `CometMakeDate` already used the default compatible support level.

Validation included:

- An exact-source Rust/Java `LocalDate` comparison using JDK 26, covering 3,387,392 valid Date32 cases and 192,089 invalid-calendar/message cases, with no differences in that domain.
- 5,940 direct Catalyst evaluations on JDK 17 across Spark 3.4.3, 3.5.9, 4.0.4, 4.1.3, and 4.2.0, compared with the old and new native scalar functions. The bounded matrix identified 71 corrected input triples and no introduced mismatch. Calendar-valid dates beyond Date32 retain a pre-existing error-category mismatch; this patch does not establish parity for those inputs.
- The five unchanged/new scalar unit tests compiled from exact source, checked-arithmetic probes, changed-file Rust formatting, `git diff --check`, and suite-registration checks passed.
- The [Rust CI job](https://github.com/apache/datafusion-comet/actions/runs/32641784604/job/97200272246) reports 899 passed and 4 skipped, including the new wide-year test and existing UDF-registration smoke tests. CI tested merge `30293cfd0d1ef3a2a2ccf16d71d5198f850f1148`; all six changed-file blobs match the reviewed head.

No full local Cargo/Maven suite or end-to-end Comet SQL/JNI run was performed. The six expression-matrix jobs did not reach the SQL fixtures: five encountered the base branch's missing `COMET_EXEC_SHUFFLE_ENABLED` test symbol, and [Linux Spark 4.0](https://github.com/apache/datafusion-comet/actions/runs/32641784604/job/97200756460) failed a Maven download with HTTP 403. Other red jobs were not all diagnosed; the Delta build gate suppresses the underlying failing command's output. These limits prevent treating this approval as proof of a fully passing JVM matrix or of every downstream temporal operation's wide-date support.

## Key design decisions

The helper is exposed only within the crate, keeping the change internal and avoiding a new public API or dependency. Widening the existing `i32` inputs before arithmetic keeps this caller safely within `i64` intermediate bounds; checked narrowing preserves Date32's storage contract. The existing DateType/Date32 mapping, `fail_on_error` transport, scalar expansion, and array/error ownership remain unchanged.

## Implementation sketch

`conversion_funcs::string` makes the existing helper crate-visible, and `conversion_funcs` reexports it for the datetime implementation. `make_date` replaces chrono construction with the helper call and checked narrowing, while retaining the surrounding UDF control flow. The Scala serde removes the resolved chrono-range note, and the Rust and ANSI/non-ANSI SQL fixtures add wide-year coverage.

Spark-valid wide dates such as `300000-06-15` and `300000-02-29` now produce their correct epoch-day values rather than NULL or a misleading error. Year zero and BCE dates remain supported, and successful construction includes the exact Date32 minimum and maximum. The new SQL queries are meaningful native-coverage checks: their runner disables constant folding, applies the file-specific ANSI setting, and checks both the answer and native operators.

## Suggested improvements

No blocking changes are requested for this patch. The remaining integration-verification follow-up is a green JVM/native-SQL run after the base compilation issue and Maven download failure are resolved. Existing beyond-Date32 error parity and downstream chrono limitations should be evaluated separately rather than reported as new regression findings on this constructor fix.

sunchao:
Merged, thanks @peterxcli !

PR #5445, opened 2026-08-24
fix: report native child spill metrics in shuffle tasks

### LINKED ISSUE #5382: Report native child-operator spill metrics in Spark task metrics for unified shuffle plans
## What / Why

Native shuffle execution can inline a spill-capable operator, such as a native sort or sort-merge join, beneath `ShuffleWriterExec` in one native plan. The writer and its child operators have separate `CometMetricNode`s and separate SQL spill metrics, but the shuffle writer's task-metric bridge currently publishes only spill counters from the root writer node.

As a result, the Spark SQL operator view can show a child operator spilling while the Spark Stages/task view reports only the shuffle writer's own spill volume. This discrepancy already exists independently of #5370, which deliberately limits its fix to native shuffle writer metrics.

For example, suppose the shuffle writer spills **32 MiB** to disk and a native sort running beneath it spills another **128 MiB**. The Spark task should report **160 MiB** of disk spill, but currently reports only the writer's **32 MiB**. In-memory spill bytes must also be included when a child operator exposes a trustworthy in-memory counter; compressed disk bytes must not be reused as a substitute for memory spill bytes.

## Proposed scope

- Account for spill-capable descendant operators in the native metric tree when reporting task-level spill metrics for unified shuffle plans.
- Preserve separate on-disk and in-memory spill semantics. If a child lacks an in-memory spill metric, provide an accurate source rather than copying compressed disk bytes.
- Avoid counting the same writer, child metric, or shared/reused accumulator more than once.
- Preserve completion-listener ordering so final native metrics are available for successful, failed, and canceled task attempts.
- Add focused native sort or sort-merge-join coverage that compares Spark SQL operator spill metrics with Spark task metrics on Spark 3.5 and 4.0.

Related: #3996 and #5370.

sunchao:

Closes #5382.

When a Spark task spills data, the task-level spill metrics are the natural place to understand how much memory pressure that task experienced. They populate the Spark UI's stage and task views and are commonly used when investigating slow shuffles, tuning memory, or comparing execution engines. For a native Comet shuffle, however, those totals currently describe only part of the work that actually happened.

Consider a query that sorts rows within each input partition before repartitioning them:

```scala
spark.table("orders")
  .sortWithinPartitions($"amount".desc)
  .repartition(4, $"customer_id")
  .count()
```

Comet can execute the sort and shuffle writer together within the same native shuffle task:

```text
CometShuffleExchangeExec
  CometSortExec
    Scan
```

Both the shuffle writer and the sort can spill independently. Their individual spill values are already available on their corresponding Spark SQL operators, but the existing task-metric bridge reads only the shuffle writer's metrics. Consequently, the SQL operator view and the task/stage view can tell conflicting stories about the same execution.

Suppose the shuffle writer spills 32 MiB to disk and the child sort spills another 128 MiB. Suppose also that the writer reports 96 MiB of in-memory spill, while the child sort exposes no in-memory spill counter:

| Spill metric | Shuffle writer | Child sort | Task total before | Task total after |
| Disk bytes spilled | 32 MiB | 128 MiB | **32 MiB** | **160 MiB** |
| Memory bytes spilled | 96 MiB | Not reported | **96 MiB** | **96 MiB** |

Before this change, the task appears to have spilled only 32 MiB to disk even though its own operators report five times that amount. This is not a difference in how much data the query actually spilled: it is missing accounting at the boundary between Comet's operator metrics and Spark's task metrics.

The unchanged memory total in this example is equally important. Disk spill bytes can represent compressed data written to storage, while memory spill bytes represent the corresponding in-memory volume. Treating the child's 128 MiB of disk spill as though it were also 128 MiB of memory spill would merely replace one misleading metric with another. A child should contribute to task-level memory spill only when it actually exposes a trustworthy memory-spill measurement.

The change makes task-level spill reporting represent the complete native shuffle task rather than only its outer shuffle writer.

Comet already maintains an operator-shaped tree of SQL metrics for the native plan. Instead of introducing a second accounting system, this PR connects the shuffle writer's existing metrics to that existing child-operator tree and reports the combined tree when the task completes. Disk spill is aggregated from the disk counters that operators actually expose; memory spill is aggregated independently from genuine memory counters. The traversal also recognizes shared metric objects, so a reused accumulator contributes once even if it appears through multiple paths in the tree.

Returning to the example above, the completed task now reports `32 MiB + 128 MiB = 160 MiB` of disk spill. Its memory-spill total remains 96 MiB because the child sort has no memory counter to contribute. If a different child operator exposes, for example, 384 MiB of memory spill, that value is included naturally and the task reports `96 MiB + 384 MiB = 480 MiB` instead. No memory value is inferred from compressed disk bytes.

The other subtlety is timing. Native operators can publish their final metrics only while their iterators are being closed at task completion, and a shuffle task can initialize nested input producers before its writer is constructed. Registering the reporting callback too late can therefore observe incomplete child metrics, particularly when an attempt fails. This PR establishes one task-level reporting callback before those producers are initialized, so Spark's completion-listener ordering first finalizes the native operators and then reads the complete metric tree. That single callback replaces the previous writer-only callback rather than layering additional reporting on top of it.

The result is intentionally narrow: it changes how already-existing metrics are collected into Spark's task totals, without changing query execution, spill behavior, native memory management, the planner, or JNI.

## How was this PR tested?

A Spark integration regression runs a real native sort beneath a native shuffle, applies enough memory pressure to make the sort spill, and then compares the Spark stage metrics with the SQL metrics of the actual executed operators. It verifies that task-level disk spill equals the writer's disk spill plus the child sort's disk spill exactly once. It also verifies that the child sort has no memory-spill counter and that the task's memory-spill total therefore remains equal to the writer's real memory metric.

A separate deterministic task-completion test models a nested input producer and a shuffle writer publishing their metrics at completion time. It verifies that both disk and memory values are collected after those final updates, for both successful and failed task attempts. Additional coverage constructs a nested metric tree with shared accumulators to ensure reused metrics are not counted multiple times. Existing shuffle spill, failed-attempt, scan-input, output, and task-binary-size tests continue to run alongside the new regressions.

Both focused suites passed on **Spark 3.5, Spark 4.0, and Spark 4.1**, with **12 passing tests per version**.

<details>
<summary>Focused validation commands</summary>

```bash
SUITES=org.apache.spark.sql.comet.CometTaskMetricsSuite,org.apache.spark.sql.comet.execution.shuffle.CometNativeShuffleInputRDDSuite

./mvnw -Pspark-3.5 test -Dtest=none -DfailIfNoTests=false -Dsuites="$SUITES"
./mvnw -Pspark-4.0 test -Dtest=none -DfailIfNoTests=false -Dsuites="$SUITES"
./mvnw -Pspark-4.1 test -Dtest=none -DfailIfNoTests=false -Dsuites="$SUITES"
```

</details>

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDD.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDD.scala
@@ -21,7 +21,7 @@ package org.apache.spark.sql.comet.execution.shuffle
 
 import org.apache.spark._
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.comet.CometExecRDD
+import org.apache.spark.sql.comet.{CometExecRDD, CometMetricNode}
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
 import org.apache.comet.CometShuffleBlockIterator
@@ -44,6 +44,8 @@ private[shuffle] class CometNativeShuffleInputRDD(
       sc,
       inputRDDs.map(rdd => new OneToOneDependency(rdd))) {
 
+  private[shuffle] var spillMetricNode: Option[CometMetricNode] = None
+
   override protected def getPartitions: Array[Partition] =
     (0 until numPartitionsParam).map { i =>
       // Resolve leaf-RDD partitions on the driver here (where their @transient fields are still
@@ -63,6 +65,7 @@ private[shuffle] class CometNativeShuffleInputRDD(
   override def compute(
       split: Partition,
       context: TaskContext): Iterator[Product2[Int, ColumnarBatch]] = {
+    spillMetricNode.foreach(_.reportSpillMetrics(context))
     val partition = split.asInstanceOf[CometNativeShuffleInputPartition]
     val (inputObjects, shuffleBlockIters) =
       CometExecRDD.resolveInputObjects(
--- a/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDD.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDD.scala
@@ -39,12 +39,11 @@ private[shuffle] class CometNativeShuffleInputRDD(
     var inputRDDs: Seq[RDD[_]],
     numPartitionsParam: Int,
     shuffleScanIndices: Set[Int],
+    spillMetricNode: CometMetricNode,
     @transient perPartitionByKey: Map[String, Array[Array[Byte]]] = Map.empty)
     extends RDD[Product2[Int, ColumnarBatch]](
       sc,
       inputRDDs.map(rdd => new OneToOneDependency(rdd))) {
-
-  private[shuffle] var spillMetricNode: Option[CometMetricNode] = None
 
   override protected def getPartitions: Array[Partition] =
     (0 until numPartitionsParam).map { i =>
@@ -65,7 +64,7 @@ private[shuffle] class CometNativeShuffleInputRDD(
   override def compute(
       split: Partition,
       context: TaskContext): Iterator[Product2[Int, ColumnarBatch]] = {
-    spillMetricNode.foreach(_.reportSpillMetrics(context))
+    spillMetricNode.reportSpillMetrics(context)
     val partition = split.asInstanceOf[CometNativeShuffleInputPartition]
     val (inputObjects, shuffleBlockIters) =
       CometExecRDD.resolveInputObjects(
--- a/spark/src/test/scala/org/apache/spark/sql/comet/CometTaskMetricsSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/comet/CometTaskMetricsSuite.scala
@@ -231,8 +231,8 @@ class CometTaskMetricsSuite extends CometTestBase with AdaptiveSparkPlanHelper {
 
         val shuffleWriteStages = store
           .stageList(null)
-          .filterNot(stage => stagesBefore.contains(stage.stageId))
-          .filter(_.shuffleWriteRecords > 0L)
+          .filter(stage =>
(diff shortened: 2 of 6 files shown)


spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometNativeShuffleInputRDD.scala
```diff
@@ -44,6 +44,8 @@ private[shuffle] class CometNativeShuffleInputRDD(
       sc,
       inputRDDs.map(rdd => new OneToOneDependency(rdd))) {
 
+  private[shuffle] var spillMetricNode: Option[CometMetricNode] = None
```
peterxcli: put into constructor so we dont forgot to set it?

peterxcli (approved):
other lgtm

peterxcli:
@sunchao https://github.com/apache/datafusion-comet/issues/5447 I filed a related issue, please feel free to take it, but if you dont have bandwidth to do it then I'm happy to help :)

spark/src/test/scala/org/apache/spark/sql/comet/CometTaskMetricsSuite.scala
```diff
@@ -160,6 +190,59 @@ class CometTaskMetricsSuite extends CometTestBase with AdaptiveSparkPlanHelper {
...
+        val writerDiskSpilled = exchange.metrics("spilled_bytes").value
+        val writerMemorySpilled = exchange.metrics("memory_spilled_bytes").value
+        val childDiskSpilled = childSorts.map(_.metrics("spilled_bytes").value).sum
+        assert(childDiskSpilled > 0L, "Native child sort did not spill")
+        assert(childSorts.forall(!_.metrics.contains("memory_spilled_bytes")))
+
+        val shuffleWriteStages = store
+          .stageList(null)
+          .filterNot(stage => stagesBefore.contains(stage.stageId))
+          .filter(_.shuffleWriteRecords > 0L)
```
comphead: a nit: maybe we can combine it into a single filter

comphead (approved):
Thanks @sunchao

sunchao:
Merged, thanks @comphead and @peterxcli for the review!

PR #5451, opened 2026-08-24
feat: native `spark_unbase64` kernel

### LINKED ISSUE #5347: Implement unbase64 natively instead of JVM codegen dispatch

0lai0:

Closes #5347 

## Rationale for this change

`UnBase64` currently runs through `CometCodegenDispatch`, which pays JVM codegen + per-row dispatch on the hot path. Every row crosses the JNI boundary into Spark's generated decoder, which defeats the point of running the surrounding plan natively. 
A native kernel replaces both the crossing and the per-row overhead; the macro bench (below) shows the JVM path at
1.3–1.8 µs/row across payload shapes, all reducible by 2.4x–3.5x.

`unbase64` is also the last common base64 direction still on the JVM dispatcher, `spark_base64` (encode) went native in #5350, so this closes out the pair.

- `native/spark-expr/src/string_funcs/unbase64.rs` — kernel + 17 unit tests. Ports `java.util.Base64.Decoder.decode0` with `isMIME = true` byte-for-byte, including the four terminal-shape error messages Spark surfaces as `IllegalArgumentException`. 256-entry compile-time `BASE64_LUT`; non-alphabet bytes skipped so `unbase64(base64(x))` round-trips through Spark's default CRLF-wrapped output. JDK quirk pinned: after padding, JDK's trailing loop tests `base64[b] < 0`, so extra `=` bytes are silently consumed — `YQ===` decodes to `a`. An earlier draft rejected these; the regression test is `extra_trailing_padding_after_valid_pair`.
- `native/spark-expr/benches/unbase64.rs` — criterion bench (short, long single-line, long CRLF-wrapped, tiny, no-nulls / dense / all nulls, error-first, scalar).
- `spark/src/main/scala/org/apache/comet/serde/strings.scala` — `CometUnBase64` is now `CometExpressionSerde with CodegenDispatchFallback`. `failOnError = true` routes back to the JVM dispatcher via `Unsupported(...)`; native strict RFC 4648 validation is explicitly a Non-goal on #5347.
- `spark/src/test/resources/sql-tests/expressions/string/unbase64.sql` — SQL fixture, happy-path only. Error shapes stay in Rust unit tests because Spark surfaces them as raw `IllegalArgumentException` with no SQL error class.
- `spark/src/test/scala/org/apache/spark/sql/benchmark/CometUnBase64Benchmark.scala` — macro bench.
- `docs/source/contributor-guide/expression-audits/string_funcs.md` — audit entry pinned across 3.4.3 / 3.5.9 / 4.0.4 / 4.1.3 / 4.2.0.

- **Rust unit tests** (17, all in `unbase64.rs`): empty input, padded/unpadded, `xx=` / `xxx=` shapes, embedded space, embedded CRLF, `\r\n`-wrapped round-trip against `spark_base64`, each of the four error shapes, JDK trailing-`=` quirk, `check_binary_capacity` boundary, Utf8/LargeUtf8 array dispatch, null propagation, wrong arity, wrong scalar type.
- **SQL fixture** at `spark/src/test/resources/sql-tests/expressions/string/unbase64.sql`, loaded by `CometScalarFunctionSuite`. Covers padded, unpadded, empty, NULL, embedded space, embedded CRLF, plus a round-trip over lengths 3 / 5 / 0 / NULL / 57 / 58 / 120 (crossing the 57-byte chunk boundary).
- **Local status (darwin arm64):**
  - `cargo fmt --check`, `cargo clippy -p datafusion-comet-spark-expr --tests --benches -- -D warnings` — clean.
  - `cargo test -p datafusion-comet-spark-expr` — 17/17 in the new module.
  - `mvn -pl spark -am test -Dtest=CometScalarFunctionSuite` — 14/14 including new fixture.

## Benchmarks

Both on darwin arm64, Apple M5, release build.

**Criterion** (`native/spark-expr/benches/unbase64.rs`), 8192 rows/batch, one null every 10 rows:

| shape                              | time / batch  | throughput  |
| ---------------------------------- | ------------: | ----------: |
| short (16 raw B)                   |      167.9 µs |  ~48.8M/s   |
| long single-line (200 B)           |      1.282 ms |   ~6.4M/s   |
| long CRLF-wrapped (200 B)          |      1.433 ms |   ~5.7M/s   |
| tiny (1 B / row)                   |       51.2 µs |   ~160M/s   |
| scalar literal (200 B)             | 183 ns / call |          —  |

**Macro bench** (`CometUnBase64Benchmark`), Spark vs Comet native, 8192 rows/batch, OpenJDK 17.0.18. Full results: `spark/benchmarks/CometUnBase64Benchmark-results.txt`.

| shape             | Spark (ns/row) | Comet (ns/row) | Speedup |
| ----------------- | -------------: | -------------: | ------: |
| short             |         1782.8 |          662.3 |    2.7x |
| long single-line  |         1621.0 |          677.5 |    2.4x |
| long CRLF-wrapped |         1597.1 |          646.7 |    2.5x |
| tiny              |         1366.0 |          385.1 |    3.5x |

No regressions. Tiny gets the largest speedup because Spark's `doGenCode` pays per-row codegen/dispatch cost that the native kernel amortises.


sunchao (commented):
## Summary

Reviewed head `adc6bd97017cb4dd1ee539f3817fe7f1b43cc632` against base `e0ab0a6fe60c05bd679654f0201ebe88319cc3a7`, with five independent full-diff passes covering all ten changed files. One verified P2 remains: recursively lowering the child of `UnBase64` can expose eager native argument evaluation where the previous whole-tree dispatcher preserved Spark's generated null-short-circuiting.

## Prior state and problem

Previously, `CometUnBase64` always used Spark's generated JVM decoder inside the Comet pipeline. That retained JVM string conversion, allocation, and per-row decoding work even when the surrounding operators ran natively. A native implementation must preserve MIME decoding, including ignored characters and CRLF-wrapped output from `base64`, rather than substitute a stricter standard decoder.

## Design approach

The PR adds a lookup-table MIME decoder, registers it as the native `unbase64` scalar function, and gives ordinary `UnBase64(failOnError = false)` an explicit binary-returning serializer. Strict `failOnError = true` expressions remain on `CodegenDispatchFallback`, preserving Spark's separate validator instead of broadening this change into native strict validation.

The decoder itself held up under the focused review: padded and unpadded input, ignored bytes, trailing padding, nulls, slices, scalar/array representation, and binary-capacity arithmetic were checked. Two independent differential probes covered 192,202 and 108,406 inputs against JDK MIME decoding without decoded-byte or success-versus-error differences. Unicode trailing-error byte positions are not identical, so this is not a claim of byte-for-byte diagnostic parity.

The remaining failure is in expression-tree evaluation, not the decoder loop. On Spark 3.5.9/JDK 17.0.19, a Parquet row with `n = NULL` and `bad = 'A'` makes `unbase64(concat(n, cast(unbase64(bad) AS STRING)))` return NULL in Spark, but the pinned head fails inside `CometProject` with `Last unit does not have enough valid bits`. Compiling the exact base-commit serializer source into an isolated classpath control restores NULL with the same native library. This isolates the regression to replacing whole-tree JVM dispatch with recursive native serialization.

Local validation passed `make core`, 17 targeted Rust tests, a 24-test exact-source Arrow harness, and nine additional native shape/conditional checks. The `unbase64`, `to_binary`, and existing `try_to_binary` SQL fixtures passed on Spark 3.5.9 and 4.1.3; the unbase64 fixture also passed on Spark 4.1.3 with JVM codegen dispatch disabled. Formatting and diff checks passed. Broader CI is still running; benchmark compilation/lint and the completed Java/Scala lint checks passed.

## Key design decisions

Keeping strict validation on the existing dispatcher is an appropriate compatibility boundary, and an explicit `BinaryType` return avoids depending on registry inference. Reusing a per-batch scratch buffer and retaining i32 binary offsets also fit the surrounding execution interfaces. However, selecting the native path must account for the child tree's conditional evaluation, not only the outer expression's `failOnError` flag.

## Implementation sketch

Scala chooses the support level and serializes the child, while the native factory routes the function name to the new scalar/array decoder. The decoder handles MIME padding and skipped-byte rules, builds nullable binary output, and guards large-array capacity. SQL fixtures cover column/literal and encode/decode round-trip shapes; Rust and JVM benchmarks cover the main payload shapes without measuring a folded-away decoder.

Ordinary unbase64 now executes in Rust, including when JVM codegen dispatch is disabled; strict conversion still uses Spark's validator and dispatcher, or falls back to Spark if that dispatcher is disabled. The new recursive path also changes which child expressions are evaluated before a NULL result is determined. The inline finding demonstrates that this can turn a previously successful query into a task failure.

## Suggested improvements

Preserve Spark's short-circuiting for affected child trees, either by retaining whole-expression dispatch for those cases or by implementing equivalent native conditional evaluation. Add a regression over non-foldable Parquet columns with an earlier NULL concat argument and a later failing expression, so successful scalar-decoder tests cannot hide this integration failure. Native strict validation and further throughput tuning can remain separate follow-ups.

spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -713,7 +713,34 @@ object CometBase64 extends CometExpressionSerde[Base64] {
...
+      " implemented natively"
+
+  override def getUnsupportedReasons(): Seq[String] = Seq(failOnErrorReason)
+
+  override def getSupportLevel(expr: UnBase64): SupportLevel = {
+    if (expr.failOnError) Unsupported(Some(failOnErrorReason)) else Compatible()
+  }
+
+  override def convert(expr: UnBase64, inputs: Seq[Attribute], binding: Boolean): Option[Expr] = {
+    val childExpr = exprToProtoInternal(expr.child, inputs, binding)
```
sunchao, line 751: [P2] Preserve null-short-circuiting when lowering the child tree

Recursively serializing `expr.child` changes evaluation of trees that previously ran entirely through Spark's codegen dispatcher. With a Parquet table `t(n STRING, bad STRING)` containing `(NULL, 'A')`, `SELECT unbase64(concat(n, cast(unbase64(bad) AS STRING))) FROM t` returns NULL in Spark's default generated execution and with the pre-change dispatcher, but this head fails with `Last unit does not have enough valid bits` (reproduced on Spark 3.5.9/JDK 17 with `CometProject` + `CometNativeScan`). Spark's generated concat stops after its first NULL argument; native `ScalarFunctionExpr` evaluates the inner decoder before concat can apply its null mask. An isolated control using the exact base serializer restores NULL. Please retain whole-tree dispatch for affected children, or preserve their short-circuit semantics natively, before marking all `failOnError = false` trees compatible.

0lai0: Thanks @sunchao for review. 
I went with whole-tree dispatch for non-leaf children.
`getSupportLevel` now returns `Compatible()` only for `Attribute` / `Literal` children. Anything else is `Unsupported` and stays on the codegen dispatcher via `CodegenDispatchFallback`, so Spark's `concat` short-circuit is preserved. Native `unbase64` is only used for `unbase64(col)` and `unbase64('literal')`.

Regression coverage:
- `unbase64.sql`: `CASE WHEN` over Parquet columns (NULL branch first, `unbase64('A')` in the other), works across 3.4–4.1.
- `unbase64_concat_short_circuit_spark_3_5.sql`: your exact `concat` query, gated to Spark 3.5. On 4.x Spark itself raises on `(NULL, 'A')`, so that form cannot be checked against a 4.x reference.

Kernel is unchanged. Local `CometSqlFileTestSuite` passes on both profiles:
`unbase64` / `to_binary` / `base64` on `-Pspark-4.1`, and both `unbase64.sql` and the concat fixture on `-Pspark-3.5`. 
The 3.5-only fixture is `SKIPPED` on 4.1 as expected.

sunchao (approved):

sunchao:
Merged, thanks @0lai0 for your contribution!

PR #5452, opened 2026-08-24
feat: support Map for `CreateArray` literal

### LINKED ISSUE #5448: Support `MapType(IntegerType,ArrayType(IntegerType,false),true)` for `CreateArray`

Part of #1937

Fallback reasons - List(Unsupported data type MapType(IntegerType,ArrayType(IntegerType,false),true), unsupported arguments for CreateArray)

_No response_

### Additional context

_No response_

comphead:

Closes #5448 .

 Extends `CometLiteral` to serialize complex `Literal` values that Spark's `ConstantFolding` produces from `map(...)`, `named_struct(...)`, or `array(...)` sub-trees. The native `Literal` proto encodes scalars and nested `ListLiteral` only, so folded values containing `MapType` or `StructType` previously fell back to Spark.

  ## Changes

  - `CometLiteral.getSupportLevel` reports `Compatible` for folded complex literals gated by `canExpandComplexLiteral`.
  - `CometLiteral.convert` calls `expandComplexLiteral(value, dataType)` which rebuilds the value as a tree of `CreateArray` / `CreateMap` / `CreateNamedStruct` over primitive-typed `Literal`s, then recurses via `exprToProtoInternal`. Existing `CometCreateArray` / `CometCreateMap` / `CometCreateNamedStruct` serdes take over from there.
  - Non-null child literals are wrapped in `KnownNullable`. DataFusion `make_array` asserts strict Arrow-type equality across siblings and would panic on a nullability mismatch. `CometKnownNullable` is a no-op on the wire.
  - Empty top-level `ArrayType` literals stay on the pre-existing `makeListLiteral` path (element type cannot be recovered from a childless `Create*`).

### DIFF
--- a/native/core/src/execution/planner.rs
+++ b/native/core/src/execution/planner.rs
@@ -49,7 +49,9 @@ use crate::execution::{
 };
 use crate::jvm_bridge::{jni_call, JVMClasses, ShufflePartitionPusher};
 use arrow::compute::CastOptions;
-use arrow::datatypes::{DataType, Field, FieldRef, Schema, TimeUnit, DECIMAL128_MAX_PRECISION};
+use arrow::datatypes::{
+    DataType, Field, FieldRef, Fields, Schema, TimeUnit, DECIMAL128_MAX_PRECISION,
+};
 use arrow::ffi_stream::FFI_ArrowArrayStream;
 use datafusion::functions_aggregate::bit_and_or_xor::{bit_and_udaf, bit_or_udaf, bit_xor_udaf};
 use datafusion::functions_aggregate::count::count_udaf;
@@ -214,6 +216,39 @@ fn make_all_fields_nullable(data_type: &DataType) -> DataType {
                 _ => data_type.clone(),
             }
         }
+        other => other.clone(),
+    }
+}
+
+/// Return a copy of a `Map` type with only the outer entries `value` field marked nullable, keeping
+/// the key field non-nullable and every nested key/value type byte-for-byte unchanged. Any non-`Map`
+/// type is returned unchanged.
+///
+/// This is the shallow counterpart of `make_all_fields_nullable`, used for `map_entries`.
+/// `map_entries` reuses the input map's entries array as its output list values but declares that
+/// element's `value` field nullable (`Struct(key non-null, value nullable)`), deriving both nested
+/// types verbatim from the input. Arrow's `GenericListArray::try_new` compares the declared element
+/// field's type against the reused values array's type in full, so the ONLY field that can mismatch
+/// is the entries `value` field's own `nullable` flag; widening just that field is sufficient.
+/// Recursing into nested types (as `make_all_fields_nullable` does) would additionally flip, say, a
+/// nested `Map`'s `valueContainsNull`, which then diverges from the Spark-serialized `return_type`
+/// and makes a downstream `make_array` see unequal map types and panic.
+fn widen_map_entry_value_nullable(data_type: &DataType) -> DataType {
+    match data_type {
+        DataType::Map(entries, sorted) => match entries.data_type() {
+            DataType::Struct(kv) if kv.len() == 2 && !kv[1].is_nullable() => {
+                let new_value = Arc::new(kv[1].as_ref().clone().with_nullable(true));
+                let new_kv: Fields = vec![Arc::clone(&kv[0]), new_value].into();
+                let new_entries = Arc::new(
+                    entries
+                        .as_ref()
+                        .clone()
+                        .with_data_type(DataType::Struct(new_kv)),
+                );
+                DataType::Map(new_entries, *sorted)
+            }
+            _ => data_type.clone(),
+        },
         other => other.clone(),
     }
 }
@@ -3490,6 +3525,17 @@ impl PhysicalPlanner {
             .collect::<Result<Vec<_>, _>>()?;
 
         let fun_name = &expr.func;
+        // `map_entries` needs its argument's entry `value` field widened to nullable first (only
+        // that outer field). See `widen_map_entry_value_nullable`.
+        let args = if fun_name == "map_entries" {
+            args.into_iter()
+                .map(|arg| {
+                    Self::coerce_child_to(arg, &input_schema, widen_map_entry_value_nullable)
+                })
+                .collect::<Result<Vec<_>, ExecutionError>>()?
+        } else {
+            args
+        };
         let input_expr_types = args
             .iter()
             .map(|x| x.data_type(input_schema.as_ref()))
@@ -3651,6 +3697,26 @@ impl PhysicalPlanner {
         }
         let nullable_type = make_all_fields_nullable(&child_type);
         Ok(Arc::new(CastExpr::new(child, nullable_type, None)))
+    }
+
+    /// Casts `child` so its type matches `widen(child_type)`, wrapping it in a `CastExpr` only when
+    /// the widened type differs. Used for the `map_entries` argument widening (with
+    /// `widen_map_entry_value_nullable`). Unlike `coerce_collect_child_nullability`, which casts
+    /// unconditionally as a normalization barrier for aggregate accumulators, this casts only when
+    /// the type actually changes. That is sufficient for the scalar `map_entries`, whose reused
+    /// entries array only needs its declared element type to line up.
+    fn coerce_child_to(
+        child: Arc<dyn PhysicalExpr>,
+        schema: &SchemaRef,
+        widen: impl Fn(&DataType) -> DataType,
+    ) -> Result<Arc<dyn PhysicalExpr>, ExecutionError> {
+        let child_type = child.data_type(schema.as_ref())?;
+        let widened = widen(&child_type);
+        if child_type.equals_datatype(&widened) {
+            Ok(child)
+        } else {
+            Ok(Arc::new(CastExpr::new(child, widened, None)))
+        }
     }
 
     fn create_aggr_func_expr(
--- a/spark/src/main/scala/org/apache/comet/serde/arrays.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/arrays.scala
@@ -29,6 +29,7 @@ import org.apache.spark.sql.types._
 
 import org.apache.comet.CometConf
 import org.apache.comet.CometSparkSessionExtensions.withFallbackReason
+import org.apache.comet.DataTypeSupport.{deepNullable, isComplexType}
 import org.apache.comet.serde.QueryPlanSerde._
 import org.apache.comet.shims.{CometExprShim, CometTypeShim}
 
@@ -100,7 +101,28 @@ object CometArrayAppend extends CometExpressionSerde[ArrayAppend] {
   }
 }
 
-object CometArrayContains extends CometExpressionSerde[ArrayContains] {
+object CometArrayContains
+    extends CometExpressionSerde[ArrayContains]
+    with CodegenDispatchFallback {
+
+  private val floatingPointReason: String =
+    "Spark compares array elements with ordering.equiv, so -0.0 matches +0.0 and all NaNs match " +
+      "each other; Comet's native array_contains compares the raw Arrow values bitwise"
+
+  override def getIncompatibleReasons(): Seq[String] = Seq(floatingPointReason)
+
+  override def getSupportLevel(expr: ArrayContains): SupportLevel = expr.left.dataType match {
+    // Native array_contains compares floating-point elements bitwise, disagreeing with Spark for
+    // -0.0/+0.0 and NaN. Report Incompatible (not Unsupported) for float/double element types (at
+    // any nesting level) so the expression routes through the JVM codegen dispatcher (Spark's own
+    // doGenCode) and stays native + Spark-exact under the default config, while non-float arrays
+    // keep the fast native kernel. Under allowIncompatible=true the native kernel is used
+    // as before.
+    case ArrayType(elementType, _)
+        if SupportLevel.containsType(elementType, classOf[FloatType], classOf[DoubleType]) =>
+      Incompatible(Some(floatingPointReason))
+    case _ => Compatible()
+  }
 
   override def convert(
       expr: ArrayContains,
@@ -384,7 +406,7 @@ object CometArrayJoin
   }
 }
 
-object CometArrayInsert extends CometExpressionSerde[ArrayInsert] {
+object CometArrayInsert extends CometExpressionSerde[ArrayInsert] with ArraysBase {
 
   override def getSupportLevel(expr: ArrayInsert): SupportLevel = Compatible()
 
@@ -392,9 +414,32 @@ object CometArrayInsert extends CometExpressionSerde[ArrayInsert] {
       expr: ArrayInsert,
       inputs: Seq[Attribute],
       binding: Boolean): Option[ExprOuterClass.Expr] = {
-    val srcExprProto = exprToProtoInternal(expr.children.head, inputs, binding)
+    val srcArray = expr.children.head
+    val item = expr.children(2)
+    val srcElementType = srcArray.dataType.asInstanceOf[ArrayType].elementType
+
+    // Native ArrayInsert requires the item's Arrow type to equal the source array's element type
+    // exactly, including nested nullability. For complex element types the two sides can disagree:
+    // a `CreateArray` source is widened to a deeply-nullable element type (see CometCreateArray),
+    // while a standalone item (e.g. `map(2, coalesce(id, 0))`) keeps Spark's Catalyst nullability.
+    // Cast BOTH sides in lockstep to the same deeply-nullable element type so their Arrow types
+    // match; Spark's `ArrayInsert.dataType` is `first.dataType.asNullable`, so this also
+    // matches the declared output element type. Casting only widens metadata and never
+    // changes values. Primitive element types are byte-identical on both sides, so the gate
+    // leaves them untouched.
+    val (srcChild, itemChild) = if (isComplexType(srcElementType)) {
+      val elementType = deepNullable(srcElementType)
+      val arrayType = ArrayType(elementType, containsNull = true)
+      val widenedSrc = if (srcArray.dataType == arrayType) srcArray else Cast(srcArray, arrayType)
+      val widenedItem = if (item.dataType == elementType) item else Cast(item, elementType)
+      (widenedSrc, widenedItem)
+    } else {
+      (srcArray, item)
+    }
+
+    val srcExprProto = exprToProtoInternal(srcChild, inputs, binding)
     val posExprProto = exprToProtoInternal(expr.children(1), inputs, binding)
-    val itemExprProto = exprToProtoInternal(expr.children(2), inputs, binding)
+    val itemExprProto = exprToProtoInternal(itemChild, inputs, binding)
     val legacyNegativeIndex =
       SQLConf.get.getConfString("spark.sql.legacy.negativeIndexInArrayInsert").toBoolean
     if (srcExprProto.isDefined && posExprProto.isDefined && itemExprProto.isDefined) {
@@ -455,7 +500,7 @@ object CometArrayUnion extends CometExpressionSerde[ArrayUnion] {
   }
 }
 
-object CometCreateArray extends CometExpressionSerde[CreateArray] {
+object CometCreateArray extends CometExpressionSerde[CreateArray] with ArraysBase {
   override def convert(
       expr: CreateArray,
       inputs: Seq[Attribute],
@@ -472,26 +517,19 @@ object CometCreateArray extends CometExpressionSerde[CreateArray] {
 
     // DataFusion's `make_array` asserts strict element-type equality in
     // `MutableArrayData::with_capacities` and panics on a mismatch. Spark's CreateArray is more
-    // permissive: its type coercion compares element types with `sameType`, which ignores
-    // nullability, so children that share a surface type but differ only in nested field
-    // nullability get no unifying cast. DataFusion tolerates container nullability differences
-    // (an `ArrayType.containsNull` / `MapType.valueContainsNull` mismatch is coerced), but NOT a
-    // struct field's nullability -- `array(struct(a not null), struct(a nullable))` panics inside
-    // `make_array_inner`. Decline only those cases (i.e. children that still differ after
-    // normalizing container nullability) so Spark's evaluator handles them.
-    //
-    // TODO: remove this decline once apache/datafusion#22366 lands; the upstream fix widens the
-    // element type via nullability-OR-merge and casts each child before MutableArrayData.
-    val normalizedTypes = children.map(c => normalizeContainerNullability(c.dataType))
-    if (normalizedTypes.distinct.size > 1) {
-      withFallbackReason(
-        expr,
-        "CreateArray children have mismatched data types: " +
-          children.map(_.dataType).distinct.mkString(", "))
-      return None
+    // permissive: its coercion compares element types with `sameType` (nullability ignored), so
+    // children that share a surface type but differ in nullability reach here as distinct types.
+    // Comet's native runtime types are also frequently MORE nullable than Spark's Catalyst types
+    // (`map_entries` forces the entry `value` field nullable, list elements are nullable, ...), so
+    // casting to Spark's declared element type does not reliably unify them. Cast every child to a
+    // deeply-nullable element type instead (every array/map/struct field nullable at all nesting
+    // levels; the cast only widens metadata and never changes values), so `make_array` always sees
+    // identical Arrow types. A child whose cast is unsupported declines below.
+    val elementType = deepNullable(expr.dataType.asInstanceOf[ArrayType].elementType)
+    val childExprs = children.map { c =>
+      val unified = if (c.dataType == elementType) c else Cast(c, elementType)
+      exprToProtoInternal(unified, inputs, binding)
     }
-
-    val childExprs = children.map(exprToProtoInternal(_, inputs, binding))
 
     if (childExprs.forall(_.isDefined)) {
       scalarFunctionExprToProto("make_array", childExprs: _*)
@@ -499,26 +537,6 @@ object CometCreateArray extends CometExpressionSerde[CreateArray] {
       withFallbackReason(expr, "unsupported arguments for CreateArray")
       None
     }
-  }
-
-  /**
-   * Rewrites a type so that container nullability (`ArrayType.containsNull`,
-   * `MapType.valueContainsNull`) is forced to `true` everywhere, while struct field nullability
-   * is left intact. Two CreateArray children whose types differ ONLY in container nullability are
-   * tolerated by DataFusion's `make_array` (coerced), so they normalize equal here; a difference
-   * in a struct field's nullability survives normalization and triggers the decline above.
-   */
-  private def normalizeContainerNullability(dt: DataType): DataType = dt match {
-    case ArrayType(elementType, _) =>
-      ArrayType(normalizeContainerNullability(elementType), containsNull = true)
-    case MapType(keyType, valueType, _) =>
-      MapType(
-        normalizeContainerNullability(keyType),
-        normalizeContainerNullability(valueType),
-        valueContainsNull = true)
-    case StructType(fields) =>
-      StructType(fields.map(f => f.copy(dataType = normalizeContainerNullability(f.dataType))))
-    case other => other
   }
 }
 
@@ -593,7 +611,7 @@ object CometElementAt extends CometExpressionSerde[ElementAt] {
   override def getSupportLevel(expr: ElementAt): SupportLevel = {
     expr.left.dataType match {
       case _: ArrayType => Compatible()
-      case _: MapType => Compatible()
+      case MapType(keyType, _, _) => MapKeySupport.keySupport(keyType)
       case _ => Unsupported(Some("Input must be an array or map"))
     }
   }
@@ -605,10 +623,9 @@ object CometElementAt extends CometExpressionSerde[ElementAt] {
     val childExpr = exprToProtoInternal(expr.left, inputs, binding)
     val ordinalExpr = exprToProtoInternal(expr.right, inputs, binding)
 
-    expr.left.dataType match {
+    val baseExpr = expr.left.dataType match {
       case _: MapType =>
-        val mapExtractExpr = scalarFunctionExprToProto("map_extract", childExpr, ordinalExpr)
-        mapExtractExpr
+        scalarFunctionExprToProto("map_extract", childExpr, ordinalExpr)
       case _ =>
         val defaultExpr =
           expr.defaultValueOutOfBound.flatMap(exprToProtoInternal(_, inputs, binding))
@@ -633,6 +650,49 @@ object CometElementAt extends CometExpressionSerde[ElementAt] {
           withFallbackReason(expr, "unsupported arguments for ElementAt")
           None
         }
+    }
+
+    // Spark's ElementAt is a BinaryExpression: for a NULL map/array it returns NULL WITHOUT
+    // evaluating the key/index child. Native scalar functions evaluate the key eagerly over the
+    // whole batch, so under ANSI (failOnError) a throwing key (e.g. a divide-by-zero) fires even on
+    // rows whose map/array is NULL, where Spark short-circuits. When the left can actually be NULL,
+    // guard the lookup with CASE WHEN left IS NOT NULL THEN <lookup> ELSE null so the key is only
+    // evaluated on the selected rows (DataFusion's CaseExpr filters the batch before the THEN
+    // branch), reproducing the short-circuit. Mirrors the CASE-WHEN idiom in CometArrayAppend /
+    // CometSize; the ELSE null literal carries the result type, as in CometArraysZip.
+    if (expr.failOnError && expr.left.nullable) {
+      val isNotNullExpr = createUnaryExpr(
+        expr,
+        expr.left,
+        inputs,
+        binding,
+        (builder, unaryExpr) => builder.setIsNotNull(unaryExpr))
+      val nullLiteralProto = exprToProto(Literal(null, expr.dataType), Seq.empty)
+      for {
+        base <- baseExpr
+        notNull <- isNotNullExpr
+        nullLit <- nullLiteralProto
+      } yield {
+        // The generic serde path attaches this ElementAt's expr_id and QueryContext to the CASE we
+        // return here, but the lookup that actually throws under ANSI (ListExtract, for the array
+        // case) is nested inside the THEN branch. Attach them to that inner Expr too, so a native
+        // INVALID_ARRAY_INDEX_IN_ELEMENT_AT / INVALID_INDEX_OF_ZERO error still renders Spark's
+        // `== SQL ... ==` query context. (map_extract ignores expr_id, so the map case is
+        // unaffected and never throws an index error anyway.)
+        val guardedBase = QueryPlanSerde.attachExprIdAndContext(expr, base)
+        val caseWhenExpr = ExprOuterClass.CaseWhen
+          .newBuilder()
+          .addWhen(notNull)
+          .addThen(guardedBase)
+          .setElseExpr(nullLit)
+          .build()
+        ExprOuterClass.Expr
+          .newBuilder()
+          .setCaseWhen(caseWhenExpr)
+          .build()
+      }
+    } else {
+      baseExpr
     }
   }
 }
--- a/spark/src/main/scala/org/apache/comet/serde/literals.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/literals.scala
@@ -22,10 +22,9 @@ package org.apache.comet.serde.literals
 import java.lang
 
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Attribute, CreateArray, CreateMap, CreateNamedStruct, Expression, KnownNullable, Literal}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, CreateArray, CreateMap, Expression, KnownNullable, Literal}
 import org.apache.spark.sql.catalyst.util.{ArrayData, MapData}
-import org.apache.spark.sql.types.{ArrayType, BinaryType, BooleanType, ByteType, DataType, DateType, DayTimeIntervalType, Decimal, DecimalType, DoubleType, FloatType, IntegerType, LongType, MapType, NullType, ShortType, StringType, StructType, TimestampNTZType, TimestampType}
+import org.apache.spark.sql.types.{ArrayType, BinaryType, BooleanType, ByteType, DataType, DateType, DayTimeIntervalType, Decimal, DecimalType, DoubleType, FloatType, IntegerType, LongType, MapType, NullType, ShortType, StringType, TimestampNTZType, TimestampType}
 import org.apache.spark.unsafe.types.UTF8String
 
 import com.google.protobuf.ByteString
@@ -230,11 +229,21 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
    * True when a non-null Literal of this type is not encodable in the native `Literal` proto and
    * `convert` should try to rebuild it from primitive-typed Literals. The native proto today
    * carries scalars and nested `ListLiteral`s (arrays of arrays / arrays of scalars). It does not
-   * carry Map or Struct values, so any Literal whose type contains a `MapType` or a `StructType`
-   * needs to be expanded before serialization.
+   * carry Map values, so any Literal whose type contains a `MapType` needs to be expanded before
+   * serialization.
+   *
+   * `StructType` (at any nesting depth) is deliberately excluded. Native `CometCreateNamedStruct`
+   * (`spark/src/main/scala/org/apache/comet/serde/structs.scala`) uses `values_to_arrays` in
+   * `native/spark-expr/src/struct_funcs/create_named_struct.rs`, which returns a 1-row
+   * `StructArray` whenever all children are scalar values. That collides with the row count of
+   * the surrounding batch and fails `make_array`'s length check. Field nullability is likewise
+   * inferred from the concrete child expressions, and `CometKnownNullable`
+   * (`spark/src/main/scala/org/apache/comet/serde/contraintExpressions.scala:99`) drops the tag
+   * on the wire, so wrapping a non-null child in `KnownNullable` does not carry across. Fall back
+   * to Spark for those shapes.
    */
   private def needsExpansion(dataType: DataType): Boolean = dataType match {
-    case _: MapType | _: StructType => true
+    case _: MapType => true
     case ArrayType(et, _) => needsExpansion(et)
     case _ => false
   }
@@ -242,24 +251,48 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
   /**
    * True when the Literal is a non-null complex value that we can rebuild from primitive Literals
    * via [[expandComplexLiteral]]. Empty top-level containers are excluded because a synthesized
-   * `Create[Array|Map]` with no children cannot recover the original element type.
+   * `Create[Array|Map]` with no children cannot recover the original element type. A folded
+   * `MapData` with duplicate keys is also excluded: Spark's `CreateMap.eval`
+   * (`sql/catalyst/.../complexTypeCreator.scala:250`) feeds every entry through
+   * `ArrayBasedMapBuilder`, which throws under the default `MAP_KEY_DEDUP_POLICY=EXCEPTION`
+   * (`sql/catalyst/.../util/ArrayBasedMapBuilder.scala`). Rebuilding a folded literal that came
+   * from `from_json` or a similar source would then throw where the original literal had executed
+   * cleanly.
    */
   private def canExpandComplexLiteral(expr: Literal): Boolean = {
-    val value = expr.value
-    if (value == null || !needsExpansion(expr.dataType)) return false
+    if (expr.value == null) return false
     expr.dataType match {
-      case _: ArrayType => value.asInstanceOf[ArrayData].numElements() > 0
-      case _: MapType => value.asInstanceOf[MapData].numElements() > 0
-      case _: StructType => true
+      case at: ArrayType if needsExpansion(at) =>
+        expr.value.asInstanceOf[ArrayData].numElements() > 0
+      case MapType(kt, _, _) =>
+        val mapData = expr.value.asInstanceOf[MapData]
+        mapData.numElements() > 0 && !hasDuplicateMapKeys(mapData.keyArray(), kt)
       case _ => false
     }
   }
 
   /**
-   * Rebuild a folded complex Literal as an equivalent tree of `CreateArray` / `CreateMap` /
-   * `CreateNamedStruct` over primitive-typed Literals. Callers must gate on
-   * [[canExpandComplexLiteral]] so `dataType` is one of the three complex cases and top-level
-   * containers are non-empty.
+   * True when the folded map's key array contains any duplicate values. Uses Catalyst internal
+   * equality (`UTF8String`, `Decimal`, `ArrayData`, `InternalRow`, and boxed primitives all
+   * implement `equals` / `hashCode`), matching `ArrayBasedMapBuilder`'s own dedup semantics.
+   */
+  private def hasDuplicateMapKeys(keys: ArrayData, keyType: DataType): Boolean = {
+    val n = keys.numElements()
+    if (n < 2) return false
+    val seen = new java.util.HashSet[Any](n)
+    var i = 0
+    while (i < n) {
+      val k = if (keys.isNullAt(i)) null else keys.get(i, keyType)
+      if (!seen.add(k)) return true
+      i += 1
+    }
+    false
+  }
+
+  /**
+   * Rebuild a folded complex Literal as an equivalent tree of `CreateArray` / `CreateMap` over
+   * primitive-typed Literals. Callers must gate on [[canExpandComplexLiteral]] so `dataType` is
+   * one of the two complex cases and top-level containers are non-empty and have unique keys.
    */
   private def expandComplexLiteral(value: Any, dataType: DataType): Expression =
     dataType match {
@@ -282,15 +315,6 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
           Seq(k, v)
         }
         CreateMap(children, useStringTypeWhenEmpty = false)
-      case StructType(fields) =>
-        val row = value.asInstanceOf[InternalRow]
-        val children = fields.zipWithIndex.flatMap { case (f, i) =>
-          val v =
-            if (row.isNullAt(i)) Literal(null, f.dataType)
-            else asNullable(Literal(row.get(i, f.dataType), f.dataType))
-          Seq(Literal(f.name), v)
-        }
-        CreateNamedStruct(children.toSeq)
       case other =>
         throw new IllegalStateException(s"expandComplexLiteral called on $other")
     }
--- a/spark/src/main/scala/org/apache/comet/serde/literals.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/literals.scala
@@ -22,17 +22,17 @@ package org.apache.comet.serde.literals
 import java.lang
 
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.catalyst.expressions.{Attribute, Literal}
-import org.apache.spark.sql.catalyst.util.ArrayData
-import org.apache.spark.sql.types.{ArrayType, BinaryType, BooleanType, ByteType, DateType, DayTimeIntervalType, Decimal, DecimalType, DoubleType, FloatType, IntegerType, LongType, NullType, ShortType, StringType, TimestampNTZType, TimestampType}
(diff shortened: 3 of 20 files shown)


spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,82 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+        CreateMap(children, useStringTypeWhenEmpty = false)
+      case StructType(fields) =>
+        val row = value.asInstanceOf[InternalRow]
+        val children = fields.zipWithIndex.flatMap { case (f, i) =>
+          val v =
+            if (row.isNullAt(i)) Literal(null, f.dataType)
+            else asNullable(Literal(row.get(i, f.dataType), f.dataType))
+          Seq(Literal(f.name), v)
+        }
+        CreateNamedStruct(children.toSeq)
```
sunchao: [P2] Preserve scalar/batch cardinality for reconstructed structs

With normal constant folding, `SELECT id, named_struct('a', 1) FROM t` now reaches native `CreateNamedStruct` instead of retaining Spark projection fallback. For all-scalar children, its evaluator calls `ColumnarValue::values_to_arrays` and returns a one-row `StructArray`, regardless of the input batch size. With three rows in one Parquet batch, the query fails with `Array length 1 does not match expected length 3`; arrays containing such structs similarly fail the `make_array` row-count check. I reproduced this on Spark 3.5.9 and 4.1.3, while the exact base literal serializer returns all three rows correctly. Please preserve scalar semantics/broadcast to the input batch size, or retain fallback for these literals until the native constructor supports them.

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,82 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+            else asNullable(Literal(vals.get(i, vt), vt))
+          Seq(k, v)
+        }
+        CreateMap(children, useStringTypeWhenEmpty = false)
+      case StructType(fields) =>
+        val row = value.asInstanceOf[InternalRow]
+        val children = fields.zipWithIndex.flatMap { case (f, i) =>
+          val v =
+            if (row.isNullAt(i)) Literal(null, f.dataType)
+            else asNullable(Literal(row.get(i, f.dataType), f.dataType))
```
sunchao: [P2] Preserve struct-field nullability in the native representation

`KnownNullable` is discarded by `CometKnownNullable.convert`, and native `CreateNamedStruct::fields` infers each field's nullability from its serialized child. With normal folding, `array(named_struct('a', CAST(NULL AS INT)), named_struct('a', 1))` is a single literal with a unified nullable struct type, so the outer `CometCreateArray` type guard passes. Recursive expansion then produces native struct fields with `nullable=true` and `nullable=false`. DataFusion 54.1 preserves those per-input flags during struct coercion, and Arrow panics in `MutableArrayData` with `Arrays with inconsistent types`. This reproduces on Spark 3.5.9 and 4.1.3 and the base serializer safely falls back. Please carry/normalize the field nullability on the native wire, or decline expansion when it cannot be preserved; the Catalyst-only wrapper does not prevent this panic.

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,82 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+        val keys = mapData.keyArray()
+        val vals = mapData.valueArray()
+        val children = (0 until keys.numElements()).flatMap { i =>
+          val k = if (keys.isNullAt(i)) Literal(null, kt) else Literal(keys.get(i, kt), kt)
+          val v =
+            if (vals.isNullAt(i)) Literal(null, vt)
+            else asNullable(Literal(vals.get(i, vt), vt))
+          Seq(k, v)
+        }
+        CreateMap(children, useStringTypeWhenEmpty = false)
```
sunchao: [P2] Preserve existing map entries without reapplying constructor deduplication

A folded `MapData` need not have been produced by `CreateMap`. For example, Spark folds `from_json('{"a":1,"a":2}', 'MAP<STRING,INT>')` into a map literal containing keys `[a,a]` and values `[1,2]`, and Spark/the base serializer execute it successfully. Rebuilding that value as `CreateMap` sends it through `ArrayBasedMapBuilder` again, so projecting it alongside a column from a Parquet table now throws `[DUPLICATED_MAP_KEY]` under the default policy. I reproduced the regression on Spark 3.5.9 and 4.1.3 with normal folding and a native Comet projection. Please preserve the literal's existing entries directly, or retain fallback when reconstruction would change their semantics.

sunchao (commented):
## Summary

Reviewed `ce1cfe973ee916f6e8de6373b7b4a48f99bcd776` against `e0ab0a6fe60c05bd679654f0201ebe88319cc3a7`. Reconstructing folded complex literals addresses the fallback reported in #5448, and the ordinary array-of-maps-of-arrays case executes successfully in a Comet projection. Three reproducible P2 correctness regressions remain: struct batch cardinality, struct-field nullability, and duplicate map-key handling.

## Prior state and problem

Spark constant folding can replace an entire `array(map(...), ...)` expression with a complex `Literal`. The existing literal protocol encodes scalars and nested lists, but not populated maps or structs, so these folded values generally caused projection fallback even when the corresponding constructors were supported. The missing representation is therefore at the literal-serialization boundary, not simply in `CreateArray` argument dispatch.

## Design approach

The change recognizes non-null literal types containing maps or structs and reconstructs `CreateArray`, `CreateMap`, or `CreateNamedStruct` expressions around their stored values. It then reuses ordinary expression serialization recursively. This avoids extending the literal protobuf; maps use the existing JVM codegen dispatcher inside the Comet pipeline, while arrays and structs use their native expression paths.

The common map-of-array case matches Spark on both Spark 3.5.9 and 4.1.3. However, reconstruction must preserve the value's already-established semantics as well as its apparent Catalyst type. The inline findings show three places where that contract changes: scalar struct values become one-row arrays, field nullability is inferred again from individual values, and existing map entries are subjected to constructor duplicate-key validation.

Validation used `make core` before JVM tests, clean reactor builds when switching Spark profiles, and normal-folding probes over a three-row Parquet batch with native-scan/projection assertions. All seven probes pass with the exact base literal serializer substituted into the otherwise unchanged production build; at this head, five probes fail across the three reported causes and two controls pass, on both Spark versions. These were serializer differential controls, not separate full-base builds.

## Key design decisions

Excluding empty top-level arrays and maps avoids synthesizing an untyped zero-argument constructor. Keeping primitive leaves on their existing serialization paths also limits the protocol surface. Both are useful constraints, but the `KnownNullable` wrapper is only a Catalyst annotation: its serializer forwards the child without carrying that annotation to the native struct expression. Likewise, reusing `CreateMap` is not equivalent to transporting arbitrary existing `MapData`.

## Implementation sketch

`needsExpansion` identifies maps, structs, and arrays containing them; `canExpandComplexLiteral` gates admission; and `convert` delegates the reconstructed expression back to `exprToProtoInternal`. Array elements, map entries, and struct fields are read from Catalyst's internal containers. Nested complex children are expanded as they pass through the existing serdes, while unsupported children can still decline and retain Spark fallback.

This enables standalone folded maps and structs, not only maps used as `CreateArray` children, so their existing native consumers become part of the compatibility boundary. The changed SQL fixture exercises constructor paths with `ConstantFolding` disabled by `CometSqlFileTestSuite`; it therefore does not validate the new folded-literal path. Separately, the fixture currently fails both dictionary variants on Spark 3.5.9 and 4.1.3 at line 67, where the non-folded map children have different value nullability. That failure exposes an existing constructor-path issue rather than proving that literal expansion is covered.

## Suggested improvements

Preserve scalar/batch shape and struct-field nullability before admitting folded structs, and preserve existing map entries without reapplying constructor deduplication. Add normal-folding regression coverage over multirow inputs for the three inline cases, retaining assertions that the intended Comet path actually executes. Also resolve the newly failing fixture case and retain explicit empty/null-container fallback checks.

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +224,107 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+   * cleanly.
+   */
+  private def canExpandComplexLiteral(expr: Literal): Boolean = {
+    if (expr.value == null) return false
+    expr.dataType match {
+      case at: ArrayType if needsExpansion(at) =>
+        expr.value.asInstanceOf[ArrayData].numElements() > 0
+      case MapType(kt, _, _) =>
+        val mapData = expr.value.asInstanceOf[MapData]
+        mapData.numElements() > 0 && !hasDuplicateMapKeys(mapData.keyArray(), kt)
```
sunchao, line 302: [P2] Preserve Spark map-key equality for newly admitted literals

A folded map can now reach native `map_extract`, whose Arrow-key comparison does not match Spark's key equality. With normal folding and Parquet `id` values 1, 2, 3, for example:

```sql
SELECT id, element_at(
  map(CAST(0 AS DOUBLE), 7),
  CAST(concat('-', CAST(id - 1 AS STRING), '.0') AS DOUBLE))
FROM t
```

On Spark 3.5.9, Spark returns `7, NULL, NULL`, but this head returns `NULL, NULL, NULL`: the first lookup is `-0.0`, which must match the `+0.0` key. The same regression occurs for array-of-double keys. Collated string keys hit this boundary too: on Spark 4.0.4 and 4.1.3, a folded `MAP<STRING COLLATE UTF8_LCASE, INT>` containing `A1 -> 7` returns `NULL` for a dynamic `a1` lookup instead of `7`, because the wire type loses collation and native lookup is bytewise. These are single-key maps, not duplicate-key revalidation. All probes select `CometProject` at this head and return the correct results with the exact-base literal serializer retaining Spark projection. Please decline expansion for unsupported key semantics, or keep these lookups on a Spark-compatible path.

comphead: checking, if so, means the current Spark test CI is not enough.

comphead:
Thanks @sunchao addressed comments

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,104 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+      case MapType(kt, vt, _) =>
+        val mapData = expr.value.asInstanceOf[MapData]
+        val keys = mapData.keyArray()
+        if (mapData.numElements() == 0 || hasUnsafeMapKeyType(kt) ||
+          hasDuplicateMapKeys(keys, kt)) {
+          None
+        } else {
+          val values = mapData.valueArray()
+          val children = (0 until keys.numElements()).flatMap(i =>
+            Seq(literalAt(keys, i, kt), asNullable(literalAt(values, i, vt))))
```
sunchao: [P2] Preserve map-value nullability when expanding folded literals

Could reconstruction retain the literal's declared `valueContainsNull`? With normal constant folding and a Parquet table containing `id` values 1, 2, 3:

```sql
SELECT id, array(map(1, 2), map(2, coalesce(id, 0))) AS a
FROM t
```

Both original array children have `MapType(IntegerType, IntegerType, false)`, so the existing array type check passes. Wrapping the folded map's value in `KnownNullable` changes only that reconstructed map to `valueContainsNull=true`. The dynamic sibling stays `false`, and native `make_array` panics with `Arrays with inconsistent types passed to MutableArrayData`. Unlike the earlier non-folded constructor example, this rewrite introduces the mismatch after checking two identical input types.

I reproduced this at `31f28b2e` on Spark 3.5.9 and 4.1.3 with native `CometProject` assertions. Replacing only the literal serializer with the exact base version restores Spark projection and the correct three-row result. Please preserve the declared map-value nullability and add a normal-folding regression test for this mixed literal/dynamic case.

spark/src/main/scala/org/apache/comet/serde/literals.scala
sunchao: [P2] Nested map values still bypass the key-semantics guard

Retested `31f28b2e`: the direct floating/collated-key cases now fall back correctly, but maps nested inside map values still escape the guard. With normal folding, ANSI mode enabled, and Parquet `id` values 1, 2, 3:

```sql
SELECT id, element_at(
  element_at(map(1, map(CAST(0 AS DOUBLE), 7)), id),
  CAST(concat('-', CAST(id - 1 AS STRING), '.0') AS DOUBLE)) AS v
FROM t
```

Spark returns `7, NULL, NULL`, while the native Comet projection returns `NULL, NULL, NULL` on both Spark 3.5.9 and 4.1.3. The outer key type is `INT`, so it passes the guard. `CometCreateMap` then closure-serializes the whole expression through the JVM dispatcher. The inner map literal never revisits `CometLiteral`, leaving its double keys available to native `map_extract`. A positive-zero lookup control passes, and the native lookup-key expression was checked to produce negative zero for the first row.

Nested `UTF8_LCASE` map keys have the same bypass on Spark 4.1.3: an `A1 -> 7` entry fails a dynamic `a1` lookup, while the exact-case control passes. Replacing only the literal serializer with the exact base version restores the expected results.

Could the admission check inspect every contained map's key type, including maps inside values, or keep their lookups on a Spark-compatible path? Please cover a dynamic outer lookup so constant folding cannot extract the inner map into a separately guarded literal.

spark/src/main/scala/org/apache/comet/serde/literals.scala
sunchao: [P2] Duplicate binary keys still evade the duplicate-key check

Retested `31f28b2e`: the string-key example now falls back correctly, but casting those keys to binary still admits the literal:

```sql
SELECT id, CAST(
  from_json('{"a":1,"a":2}', 'MAP<STRING,INT>')
  AS MAP<BINARY,INT>) AS m
FROM t
```

Spark's map cast preserves both entries. The new `.distinct` check compares the extracted `Array[Byte]` keys by identity, whereas `ArrayBasedMapBuilder` compares binary keys by their contents using interpreted ordering. Reconstructing the admitted literal therefore still throws `DUPLICATED_MAP_KEY` under `EXCEPTION`. Under `LAST_WIN`, it silently drops the first entry instead of preserving the original map.

I reproduced both policies at this head on Spark 3.5.9 and 4.1.3 with normal folding, three-row Parquet input, and native `CometProject` assertions. Replacing only the literal serializer with the exact base version preserves both entries under either policy.

Could the duplicate check use Spark-compatible key equality and include this binary-key case under both deduplication policies?

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,104 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+
+  /**
+   * True when the map key type has Spark equality semantics that native map lookup (Arrow
+   * bytewise comparison) cannot honor, in which case declining expansion keeps the projection on
+   * Spark. `NormalizeFloatingNumbers` makes Spark treat `+0.0` and `-0.0` as the same key and
+   * canonicalises NaN, and a non-default collation compares under rules native applies as
+   * `UTF8_BINARY`. Both are checked at every nesting level of the key type.
+   */
+  private def hasUnsafeMapKeyType(kt: DataType): Boolean =
+    hasNonDefaultStringCollation(kt) ||
```
sunchao: [P2] Preserve nullable components of complex map lookup keys

This admits folded maps whose key type is `ArrayType(IntegerType, false)`, but Spark permits a dynamic lookup key containing a null element. With normal folding and Parquet `id` values `1, 2, 3`:

```sql
SELECT id, element_at(
  map(array(1), 7),
  array(IF(id = 2, CAST(NULL AS INT), id))) AS v
FROM t
```

Spark returns `7, NULL, NULL`. Native `map_extract` instead coerces the lookup to the literal key's exact Arrow type; the inserted cast aborts with `Non-nullable field of ListArray "item" cannot contain nulls`. The null is inside the lookup array, not a null map key. This needs a nullability-compatible lookup type or fallback for these literals.

Reproduced on Spark 3.5.9 and 4.1.3 with native `CometProject` assertions, in both ANSI modes. The non-null lookup control and dispatcher-off path pass; replacing only the literal serializer with the exact base version also restores the correct rows. No floating-point keys, collations, or duplicate keys are involved.

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,104 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+        }
+      case MapType(kt, vt, _) =>
+        val mapData = expr.value.asInstanceOf[MapData]
+        val keys = mapData.keyArray()
+        if (mapData.numElements() == 0 || hasUnsafeMapKeyType(kt) ||
+          hasDuplicateMapKeys(keys, kt)) {
+          None
+        } else {
+          val values = mapData.valueArray()
+          val children = (0 until keys.numElements()).flatMap(i =>
```
sunchao: [P2] Guard nested map literals before native map_entries

A map stored inside a map value is handed to the JVM dispatcher as a literal, so its original `valueContainsNull=false` survives. This enables another previously falling-back query with normal folding over Parquet `id` values `1, 2, 3`:

```sql
SELECT id, map_entries(
  element_at(map(1, map(1, 2)), id)) AS e
FROM t
```

The inner map reaches native `map_entries`, which declares a nullable `value` field but reuses entry arrays whose field is non-nullable. Arrow's `ListArray::new` then panics with a child-type mismatch instead of returning Spark's `[{1, 2}], NULL, NULL`. There are no mixed map siblings here: the inner type has not been changed; the incompatible schema is created by its consumer.

Reproduced on Spark 3.5.9 and 4.1.3 with native `CometProject` assertions, in both ANSI modes. Nullable-inner-map, `map_keys`, and `map_values` controls pass; the exact-base literal serializer and dispatcher-off path restore Spark's result. Please retain fallback for unsupported nested-map shapes until their native consumers preserve matching entry types.

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,129 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+  private def hasDuplicateMapKeys(keys: ArrayData, keyType: DataType): Boolean = {
+    val n = keys.numElements()
+    if (n < 2) {
+      false
+    } else if ((0 until n).exists(keys.isNullAt)) {
+      true
+    } else {
+      val ordering = TypeUtils.getInterpretedOrdering(keyType)
+      val sorted = (0 until n).map(i => keys.get(i, keyType)).sorted(ordering)
+      sorted.sliding(2).exists(pair => ordering.compare(pair.head, pair.last) == 0)
```
sunchao: [P2] Decline non-orderable map keys before requesting an ordering

`TypeUtils.getInterpretedOrdering` is not defined for every valid Spark map-key type. Spark supports `CalendarIntervalType` keys using hash equality, so normal folding produces a valid two-entry literal for:

```sql
SELECT id, map(make_interval(1), 1, make_interval(2), 2) AS m FROM t
```

With Parquet `id` values `1, 2, 3`, this new call throws `Type PhysicalCalendarIntervalType does not support ordered operations` from `CometLiteral.getSupportLevel`, before the dispatcher can decline the unsupported type. The keys are distinct and non-null. I reproduced the planning failure with the exact current serde on Spark 3.5.9 and 4.1.3; Spark and the exact-base serde return all three rows correctly. Please retain fallback for non-orderable key types, or use the matching Spark hash/equality path instead of unconditionally requesting an ordering.

spark/src/main/scala/org/apache/comet/serde/arrays.scala
```diff
@@ -502,20 +503,25 @@ object CometCreateArray extends CometExpressionSerde[CreateArray] {
...
   private def normalizeContainerNullability(dt: DataType): DataType = dt match {
     case ArrayType(elementType, _) =>
       ArrayType(normalizeContainerNullability(elementType), containsNull = true)
-    case MapType(keyType, valueType, _) =>
+    case MapType(keyType, valueType, valueContainsNull) =>
       MapType(
         normalizeContainerNullability(keyType),
         normalizeContainerNullability(valueType),
-        valueContainsNull = true)
+        valueContainsNull)
```
sunchao: [P2] Keep array nullability inside map types significant

The recursion still erases `ArrayType.containsNull` inside a `MapType`, even though DataFusion 54.1 cannot coerce two differing map types. With normal folding and a Parquet `id INT` column containing `1, 2, 3`:

```sql
SELECT id, array(map(1, array(1)), map(2, array(id))) AS a FROM t
```

The folded map has value type `ArrayType(IntegerType,false)` and the dynamic sibling has `ArrayType(IntegerType,true)`. Both maps retain `valueContainsNull=false`. Normalization admits them, but native `make_array` receives unequal Arrow map types and panics with `Arrays with inconsistent types passed to MutableArrayData`. This reproduces with the exact current serde on Spark 3.5.9 and 4.1.3 with `CometProject` asserted; replacing `id` with `coalesce(id,0)` passes, as does the exact-base serde retaining Spark projection. Unlike the earlier `valueContainsNull` finding, reconstruction preserves both original map flags here. Please preserve nested array nullability across the map boundary, or unify the complete map types before calling `make_array`.

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,129 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
+        val keys = mapData.keyArray()
+        if (mapData.numElements() == 0 || hasDuplicateMapKeys(keys, kt)) {
+          None
+        } else {
+          val values = mapData.valueArray()
+          val children = (0 until keys.numElements()).flatMap(i =>
+            Seq(
+              literalAt(keys, i, kt),
+              withNullability(literalAt(values, i, vt), valueContainsNull)))
+          Some(CreateMap(children, useStringTypeWhenEmpty = false))
```
sunchao: [P2] Preserve NULL short-circuiting for newly admitted map lookups

Admitting the nested map literal exposes eager evaluation of a lookup key that Spark skips after a NULL map. With ANSI enabled and Parquet `id` values `1, 2, 3`:

```sql
SELECT id, element_at(
  element_at(map(1, map(0, 7)), id),
  id % (id - 2)) AS v
FROM t
```

Spark returns `7, NULL, NULL`: for `id=2`, the inner lookup returns NULL, so the outer `ElementAt` never evaluates the remainder. The native `map_extract` scalar expression evaluates every argument over the batch first and raises a divide/remainder-by-zero error. All map keys are integers, so the new key-type guards do not prevent it. I reproduced this with the exact current serde on Spark 3.5.9 and 4.1.3 with `CometProject` asserted; the exact-base serde and a nonzero-divisor control pass. Please preserve per-row lazy key evaluation, or retain fallback for these newly admitted expressions until the native consumer has Spark's NULL behavior.

sunchao (commented):
Re-reviewed `97544ebe6e60ef558ebb5dc0f6b34aa7ec1ea242` against `e0ab0a6fe60c05bd679654f0201ebe88319cc3a7`, including the increment from `31f28b2e0ea5341716f2824be3b7a26bc8c82acb` and all prior discussions. Three additional P2 regressions are attached below.

The existing **[P2] key-equality family** ([original thread](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3846862434), [nested-map follow-up](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3848126069)) also remains reproducible through `map_contains_key`, although the reported `element_at`/subscript cases are now guarded. With Parquet `id` values `1, 2, 3` and normal folding:

```sql
SELECT id, map_contains_key(
  element_at(map(1, map(CAST(0 AS DOUBLE), 7)), id),
  CAST(concat('-', CAST(id - 1 AS STRING), '.0') AS DOUBLE)) AS present
FROM t
```

Spark returns `true, NULL, NULL`; Comet returns `false, NULL, NULL` on Spark 3.5.9 and 4.1.3. Spark lowers membership to `array_contains(map_keys(...), key)`, bypassing `MapKeySupport`. The analogous nested `UTF8_LCASE` map with stored `A1` and dynamic `a1` lookup also returns `false` on 4.1.3. Positive-zero and exact-case controls pass. I am keeping this with the existing family rather than adding another inline finding.

Validation used the exact head/base Scala serde sources compiled into isolated overlays over existing native builds, with normal folding, multirow Parquet input, and native scan/projection assertions. The relevant native paths were unchanged; this was **not** a full current-head native rebuild, and I did not locally rerun the changed `map_entries` planner. Base controls passed all 7 applicable probes on 3.5.9 and all 9 on 4.1.3; current-source probes reproduced the findings above and below. Current GitHub checks are green.

native/core/src/execution/planner.rs
```diff
@@ -3384,6 +3384,16 @@ impl PhysicalPlanner {
             .collect::<Result<Vec<_>, _>>()?;
 
         let fun_name = &expr.func;
+        // `map_entries` reuses the input map's entries array but declares the entry `value` field
+        // nullable, so widen the argument first or Arrow rejects the child type. See
+        // `coerce_child_fields_nullable`.
+        let args = if fun_name == "map_entries" {
+            args.into_iter()
+                .map(|arg| Self::coerce_child_fields_nullable(arg, &input_schema))
```
sunchao: [P2] Widen only the outer map_entries value field

This recursive cast also changes nullability inside the map's key/value types, although `map_entries` only needs the outer entry's `value` field to be nullable. With normal folding and Parquet `id INT` values `1, 2, 3`:

```sql
SELECT id, array(
  map_entries(map(1, IF(id = 1, map(1, 2), NULL)))[0].value,
  map(2, coalesce(id, 0))) AS a
FROM t
```

Both Catalyst array children are `MapType(INT, INT, false)`, so the array guard accepts them. The outer map already has `valueContainsNull=true`, and the pre-change native query returns all three rows correctly. This cast unnecessarily widens the *inner* map to `valueContainsNull=true`; `ListExtract` and `GetStructField` preserve that changed type while the sibling remains `false`. DataFusion 54.1 cannot coerce those unequal map types, and `make_array` panics with `Arrays with inconsistent types passed to MutableArrayData`.

Please preserve the nested key/value types and widen only the outer entry-value flag. This is distinct from the original `map_entries` child-type failure: that call succeeds here, and the new cast introduces a mismatch in its downstream consumer. Verified with a passing native SQL control and a differential kernel probe using the exact cast helper, DataFusion's cast/map_entries/make_array, and Comet's extraction kernels; this was not a full current-head native rebuild.

sunchao (commented):
Re-reviewed `fecf175ab8f7ad434fa53a255851d4df90349728` against `e0ab0a6fe60c05bd679654f0201ebe88319cc3a7`, including the increment from `97544ebe6e60ef558ebb5dc0f6b34aa7ec1ea242` and all prior discussions. One additional **[P2]** is attached: the recursive `map_entries` cast changes nested map types and can cause a downstream `make_array` panic. This is in the full PR diff, not introduced by the latest refactor.

The existing **[P2]** findings still reproduce: [interval-key ordering](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509162), [array nullability inside map values](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509174), [NULL-map eager key evaluation](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509181), and the [key-equality family](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3846862434) / [nested-map follow-up](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3848126069) through `map_contains_key`, as documented in [review 5023100588](https://github.com/apache/datafusion-comet/pull/5452#pullrequestreview-5023100588). No duplicate inline findings for these.

Fresh exact-serde probes on Spark 3.5.9 and 4.1.3 reproduced those existing failures; all base and positive controls passed. The new finding has a passing native SQL control with the pre-change planner plus a differential DataFusion 54.1 / Arrow 58.4 kernel reproduction using the exact current cast helper and actual Comet extraction kernels. These are focused controls over existing native builds, not a full current-head native rebuild.

comphead:
Thanks @sunchao for iterating, Ive added all the reported tests and now it passed with `query` accelerated fully natively or fallback to Spark, in some cases like collations or float keys

sunchao (commented):
Re-reviewed `ba011bbf1538f3cf1e6fc22d3ea189e81ae0a13f` against `e0ab0a6fe60c05bd679654f0201ebe88319cc3a7`, including the full PR and the increment from `fecf175ab8f7ad434fa53a255851d4df90349728`. One new **[P2]** is attached for large folded maps reaching an invalid dispatcher code path.

Two existing **[P2]** discussions remain:

- [NULL-map lookup-key evaluation](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509181): normal folding with ANSI enabled still raises `DIVIDE_BY_ZERO` instead of returning `7, NULL, NULL`.
- [map_entries type compatibility](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3849285194): the original construction panic is fixed, but the entry struct's value nullability still causes a downstream array panic. With Parquet `id INT` values `1, 2, 3`, this fails in both ANSI modes:

```sql
SELECT id, array(
  map_entries(element_at(map(1, map(1, true)), id))[0],
  named_struct('key', 1, 'value', id IS NOT NULL)) AS a
FROM t
```

I am keeping that variant in the existing discussion, not adding a duplicate inline finding. The other nine reported mechanisms are addressed at this head.

Validation on `chao-reviews-1`: exact-head native and JVM builds; Spark 4.0.4/JDK 17; 80 Array/Map tests and six SQL-fixture runs passed. Failure probes asserted native projection/scan; the exact-base literal-serde overlay and dispatcher-off controls restore correct results. The overlay uses the same head native library, not a separate full-base build. No full-project or cross-version suite claim.

spark/src/main/scala/org/apache/comet/serde/literals.scala
```diff
@@ -215,4 +225,168 @@ object CometLiteral extends CometExpressionSerde[Literal] with Logging {
...
```
sunchao: [P2] Retain a safe execution path for large folded map literals

Rebuilding every stored entry as constructor children newly sends large folded maps through the dispatcher's nested generated classes. With normal folding and a Parquet `id INT` column containing `1, 2, 3`:

```sql
SELECT id, element_at(
  map_from_arrays(sequence(1, 100000), sequence(1, 100000)),
  id) AS v
FROM t
```

Spark returns `(1,1), (2,2), (3,3)`, but this head selects `CometProject` and fails with `IllegalAccessError`: `SpecificCometBatchKernel$NestedClass_8.CreateMap_0$` accesses the protected `CometBatchKernel.references` field. The generated helper uses that field for its map builder. The schema-field-count admission check does not constrain map entry count, and there is no execution-time fallback.

This reproduced twice on Spark 4.0.4/JDK 17. The exact-base literal serializer and dispatcher-off controls both return the correct rows; a 10,000-entry native control passes. Please retain fallback for expansions the dispatcher cannot execute, preserve the compact literal representation, or fix generated-helper reference access before admitting them.

spark/src/main/scala/org/apache/comet/serde/arrays.scala
```diff
@@ -472,27 +472,20 @@ object CometCreateArray extends CometExpressionSerde[CreateArray] {
...
-      return None
+    // permissive: its coercion compares element types with `sameType` (nullability ignored), so
+    // children that share a surface type but differ in nullability reach here as distinct types.
+    // Comet's native runtime types are also frequently MORE nullable than Spark's Catalyst types
+    // (`map_entries` forces the entry `value` field nullable, list elements are nullable, ...), so
+    // casting to Spark's declared element type does not reliably unify them. Cast every child to a
+    // deeply-nullable element type instead (every array/map/struct field nullable at all nesting
+    // levels; the cast only widens metadata and never changes values), so `make_array` always sees
+    // identical Arrow types. A child whose cast is unsupported declines below.
+    val elementType = deepNullable(expr.dataType.asInstanceOf[ArrayType].elementType)
```
sunchao, line 528: **[P2] Keep widened array types consistent with downstream consumers**

`deepNullable` changes the Arrow type produced by this array without reconciling independently serialized arguments and declared result types of its consumers. For Parquet INT `id` values `1,2,3`:

```sql
SELECT array_insert(
  array(map(1, coalesce(id, 0))),
  2,
  map(2, coalesce(id, 0)))
FROM t
```

Both Catalyst maps have non-nullable values. Only the array element is widened here, so the inserted map remains non-nullable and native `ArrayInsert` rejects their unequal types (`Type mismatch in ArrayInsert`). This fails with normal folding on Spark 4.0.4/JDK 17; Spark, dispatcher-disabled execution, and exact prior/base array-serializer controls all return the correct rows (the prior/base controls also execute this projection natively). The same metadata drift causes `slice` declared-return-type errors and `IF` schema errors on all-false batches. Please keep consumer arguments/result types consistent with this widening, or retain fallback where that cannot be satisfied.

sunchao (commented):
Re-reviewed `b64837f9057813d0791ab53b896388f9f4b433d8`, including the full increment from `ba011bbf1538f3cf1e6fc22d3ea189e81ae0a13f`. The [100,000-entry dispatcher failure](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3865543786) and [downstream `map_entries` case](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3849285194) now pass with native execution. One new P2 is inline below.

Three existing P2 families remain (examples use a Parquet INT `id` column containing `1,2,3`):

- **[P2] [NULL-map key evaluation](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509181):** with ANSI enabled, `element_at(element_at(map(1,map(0,7)),id),id % (id-2))` still throws where Spark returns `7,NULL,NULL`. With ANSI disabled, using `element_at(array(0),id-2)` as the outer key also throws for index zero. Base-literal and dispatcher-disabled controls return the correct rows with normal folding. Folding-disabled failures on both implementations establish the older kernel origin, not safety for newly admitted folded maps.
- **[P2] [Equality of map contents](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3846862434):** the original key cases are guarded, but values still expose the same comparison issue. `array_contains(map_values(element_at(map(1,map(1,CAST(0 AS DOUBLE))),id)),CAST(concat('-',CAST(id-1 AS STRING),'.0') AS DOUBLE))` returns `false,NULL,NULL` instead of Spark's `true,NULL,NULL`. A `UTF8_LCASE` value/lookup of `A1`/`a1` also fails. Positive-zero, exact-case, base-literal and dispatcher-disabled controls pass.
- **[P2] [Scalar/batch cardinality](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3845993182):** the original folded-literal trigger now falls back. With `spark.sql.optimizer.excludedRules=org.apache.spark.sql.catalyst.optimizer.ConstantFolding` and all three rows in one Parquet batch, `array(named_struct('a',1),named_struct('a',id))` now fails with `Expected length: 1, found length: 3`. Two-dynamic-struct and normal-folding controls pass; prior/base array serializers correctly fall back.

Validation on `chao-reviews-1`: exact-head native build; Spark 4.0.4/JDK 17 with `-Pspark-4.0`; 79 focused array/map tests and six selected SQL-fixture runs passed. Separate boundary probes verified native projection/scan and the failures above. Prior/base comparisons are isolated serializer overlays on the current-head runtime, not complete prior/base builds. No full-project or cross-version runtime claim.

comphead:
| Query | Was | Fix |
  |---|---|---|
  | **4** `array(named_struct('a',1), named_struct('a',id))` | native crash "mixed length" | `CreateNamedStruct` returns a scalar struct so `make_array` broadcasts it |
  | **5** `array_insert(array(map(1,coalesce(id,0))), 2, map(2,coalesce(id,0)))` | native "Type mismatch" | serde casts source + item to a common deeply-nullable element type |
  | **3** `array_contains(map_values(element_at(map(1,map(1,DOUBLE 0)),id)), -0.0)` | wrong answer (bitwise `-0.0≠+0.0`) | float element types route through codegen dispatch (Spark-exact) |
  | **1** `element_at(element_at(map(1,map(0,7)),id), id%(id-2))` ANSI on | native throws `2%0`; Spark returns NULL | `CASE WHEN left IS NOT NULL` short-circuit under ANSI |


  ## Followup
  
  - **Query 2** `element_at(array(0), id-2)` (ANSI off) — the only one that already passed; test not yet added. Needs a home: safe choice is `CometArrayExpressionSuite` (folding on). A plain SQL-file `query` 
  might also work, but only after confirming the "literal `element_at` + constant-folding-off" Spark codegen bug does not fire with a per-row column index. One quick run decides it.
  - **Query 1** — fix committed but riskiest: it changes plan shape for *every* ANSI `element_at` over a nullable map/array, and the reviewer initially rejected it before the `left.nullable` gate. Validate
  with the ANSI suites + `CometPlanStabilitySuite`, and consider splitting into its own PR.

spark/src/main/scala/org/apache/comet/serde/arrays.scala
```diff
@@ -634,6 +651,42 @@ object CometElementAt extends CometExpressionSerde[ElementAt] {
...
+    // whole batch, so under ANSI (failOnError) a throwing key (e.g. a divide-by-zero) fires even on
+    // rows whose map/array is NULL, where Spark short-circuits. When the left can actually be NULL,
+    // guard the lookup with CASE WHEN left IS NOT NULL THEN <lookup> ELSE null so the key is only
+    // evaluated on the selected rows (DataFusion's CaseExpr filters the batch before the THEN
+    // branch), reproducing the short-circuit. Mirrors the CASE-WHEN idiom in CometArrayAppend /
+    // CometSize; the ELSE null literal carries the result type, as in CometArraysZip.
+    if (expr.failOnError && expr.left.nullable) {
+      val isNotNullExpr = createUnaryExpr(
+        expr,
+        expr.left,
```
sunchao, line 666: [P2] Evaluate a stateful lookup operand only once

`expr.left` was already serialized into `baseExpr`, so serializing it again here creates a second native instance of stateful descendants. CASE evaluates this condition over the full batch, but evaluates the lookup only over selected rows. With ANSI enabled and ordered INT `id = [1, 2]` in the first batch of partition 0:

```sql
SELECT element_at(
  element_at(
    array(CAST(NULL AS ARRAY<BIGINT>),
          array(monotonically_increasing_id(), CAST(NULL AS BIGINT))),
    id),
  id - 1)
FROM t
```

The source-derived results are `[NULL, 1]` for Spark and the previous serializer, but `[NULL, 0]` here: only `id = 2` reaches the independent THEN-side counter, which starts at zero. Spark's Project optimizer does not hoist this nested nondeterminism, and native planning/projection does not memoize the two serialized instances. The NULL first row also short-circuits before index-zero validation in the previous native path, so that control does not throw.

This is source-verified, not a fresh runtime reproduction. Please materialize the left operand once, or retain a safe evaluation path when this guard would duplicate stateful descendants.

sunchao (commented):
Re-reviewed `a9b24c35` with five fresh independent reviewers. The scalar-struct cardinality, direct `array_insert`, floating-point membership, and reported deterministic ANSI lookup cases are addressed in source. The existing [NULL-key evaluation discussion](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509181) is only partially addressed; [collated map-value comparisons](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3846862434) and [nested-type consistency for `slice`/conditionals](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3867906876) also remain under their existing discussions.

The inline below is a distinct P2 introduced by the ANSI guard. These dispositions are source-verified; no fresh runtime reproduction was performed.

sunchao (commented):
Re-reviewed `1f63556a`, including the complete changes since `a9b24c35`. [P2] The existing [stateful-operand issue](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3874895480), [NULL-short-circuit issue](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509181) in `GetMapValue`/`TryElementAt`, [collated-value comparison issue](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3846862434), and [downstream array-type mismatch](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3867906876) in `Slice` remain. The fixed floating-point-key and direct `ArrayInsert` cases are not being raised again.

This update changes only SQL fixtures; the stateful-operand regression still runs with ANSI disabled, so it does not exercise the new ANSI guard. Exact-dependency native component controls corroborate the duplicated-counter and Slice-type failures. This was source/component validation, not a fresh full Spark SQL/JNI reproduction. No new distinct P1/P2 inline finding.

comphead:
> The existing https://github.com/apache/datafusion-comet/pull/5452#discussion_r3874895480, https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509181 in GetMapValue/TryElementAt, https://github.com/apache/datafusion-comet/pull/5452#discussion_r3846862434, and https://github.com/apache/datafusion-comet/pull/5452#discussion_r3867906876 in Slice remain.

I'm not sure why the agent complains, all that queries it referred to, have been added to tests and passing ok

sunchao (commented):
Re-reviewed `67fbae59`, including the update from `1f63556a`. Four source-supported P2 residuals remain in the existing discussions:

- [P2] [NULL-left lookup](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3856509181): `try_element_at` and bracket lookup can still evaluate a throwing key when the map is NULL; index-zero keys also expose the non-ANSI route.
- [P2] [Stateful ANSI operand](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3874895480): the guard constructs separate left operands over different row sets. The added stateful fixture runs with ANSI disabled.
- [P2] [Collated map values](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3846862434): folded integer-key maps containing `'A1' COLLATE UTF8_LCASE` still reach byte-based `array_contains` comparison with `'a1' COLLATE UTF8_LCASE`.
- [P2] [Slice result typing](https://github.com/apache/datafusion-comet/pull/5452#discussion_r3867906876): `slice(array(map(1,coalesce(id,0))),1,1)` retains stricter declared nested nullability than the widened array produces. The direct ArrayInsert case is fixed.

CI is now successful; the inspected expression jobs ran on recorded merge `590deafe`. These residuals are source-derived, not new runtime reproductions.

comphead:
Thanks @sunchao for iterating, last review statement is slightly misleading, is it feasible to provide missing tests or gaps? 

> Four source-supported P2 residuals remain in the existing discussions

It sounds for me the agent looks to the comments/reviews rather than code itself?

sunchao:
Hmm let me check

sunchao (commented):
@comphead — that wording was unclear. I linked existing discussions to avoid duplicate inline comments; the correctness evidence was the code at `67fbae59`, compared with the base and Spark/native implementations. These are **proposed regression cases with source-derived expectations, not reproductions I have executed**.

Use scanned Parquet `t(id INT)`, Comet enabled and `spark.comet.exec.scalaUDF.codegen.enabled=true`, with default compatibility settings. Cases 1 and 3 need constant folding enabled. The [SQL-file harness](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/test/scala/org/apache/comet/CometSqlFileTestSuite.scala#L88-L132) appends its folding exclusion after file/matrix configs; the existing [Scala Parquet test pattern](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/test/scala/org/apache/comet/CometMapExpressionSuite.scala#L285-L298), with an explicit `withSQLConf("spark.sql.optimizer.excludedRules" -> "")`, is one concrete placement.

1. **[P2] NULL-left lookup:** ANSI enabled; IDs `1,2,3`.

   ```sql
   SELECT try_element_at(element_at(map(1, map(0, 7)), id),
                         id % (id - 2)) FROM t;
   ```

   Spark should return `7, NULL, NULL`, skipping the key when the inner map is NULL. The current native path evaluates the key and reaches division by zero at `id=2`: the [new guard](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L655-L682) requires `failOnError`, which TRY clears. Also cover `element_at(map(1, map(0, 7)), id)[id % (id - 2)]`: [bracket lookup](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/maps.scala#L117-L133) has no such guard. The [strict-map regression](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/test/resources/sql-tests/expressions/map/element_at_map_ansi.sql#L30-L36) does not exercise those routes. For ANSI-off coverage, replace the key with `element_at(array(0), id - 2)`; `% 0` does not throw there.

2. **[P2] Stateful ANSI operand:** ANSI enabled; a controlled first batch containing ordered IDs `1,2` in partition 0.

   ```sql
   SELECT element_at(
     element_at(array(CAST(NULL AS ARRAY<BIGINT>),
                      array(monotonically_increasing_id(), CAST(NULL AS BIGINT))), id),
     id - 1) FROM t;
   ```

   Spark should return `NULL, 1`; the current path predicts `NULL, 0`. The [rewrite](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L623-L669) serializes the operand twice, creating separate [stateful counters](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/native/spark-expr/src/nondetermenistic_funcs/monotonically_increasing_id.rs#L35-L74); the THEN copy sees only selected rows. The [existing stateful query](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/test/resources/sql-tests/expressions/array/element_at.sql#L18-L37) runs with the harness's ANSI-off default. Simply enabling ANSI on that entire fixture would introduce unrelated out-of-bounds errors; use the two-row case.

3. **[P2] Collated map values:** Spark 4.0; IDs `1,2,3`, folding enabled.

   ```sql
   SELECT array_contains(
     map_values(element_at(map(1, map(1, 'A1' COLLATE UTF8_LCASE)), id)),
     'a1' COLLATE UTF8_LCASE) FROM t;
   ```

   Spark should return `true, NULL, NULL`; the current native path predicts `false, NULL, NULL`. [Literal admission](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/literals.scala#L295-L334) protects map keys, while [ArrayContains](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L104-L134) admits this value comparison without preserving collation semantics. The [collated-key fixture](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/test/resources/sql-tests/expressions/map/element_at_map_collation.sql#L18-L35) tests a different path. This example uses explicit `COLLATE`; a CAST-only example is not the basis for the expectation.

4. **[P2] Slice nested result type:** nullable scanned IDs `1,NULL`; either ANSI setting.

   ```sql
   SELECT slice(array(map(1, coalesce(id, 0))), 1, 1) FROM t;
   ```

   Spark should return `[{1:1}]` and `[{1:0}]`. [CreateArray now widens the map's value nullability](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L503-L541), but [Slice declares the original nested type](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L465-L487), while [native slicing retains the input field](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/native/spark-expr/src/array_funcs/array_slice.rs#L175-L180). The predicted failure is a column/schema type mismatch. The [scalar-element regression](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/test/scala/org/apache/comet/CometArrayExpressionSuite.scala#L1307-L1312) and [nested-array fixture](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/test/resources/sql-tests/expressions/array/slice.sql#L237-L261) do not cover this constructed-map case.

The Spark contract check covers pinned 3.5/4.0 sources for cases 1, 2 and 4, and 4.0 only for case 3. Each regression should compare answers and require the intended Comet project/expression route; an operator check alone does not prove every expression ran natively, since map construction intentionally uses the JVM dispatcher.

comphead:
Thanks @sunchao I'm proposing to have this PR as is, as it actually refers to support map inside array literal. 

Assessing the ticket scope and proposed fixes:

- no array literal, its most likely existing issue, we need to address.

```
SELECT try_element_at(element_at(map(1, map(0, 7)), id),
                      id % (id - 2)) FROM t;
```

- its not array of map

```
SELECT element_at(
  element_at(array(CAST(NULL AS ARRAY<BIGINT>),
                   array(monotonically_increasing_id(), CAST(NULL AS BIGINT))), id),
  id - 1) FROM t;
```

- this is good, yes, collation support is weak in Comet

```
SELECT array_contains(
  map_values(element_at(map(1, map(1, 'A1' COLLATE UTF8_LCASE)), id)),
  'a1' COLLATE UTF8_LCASE) FROM t;
```


- here reported problem for `slice` rather that array of map

```
SELECT slice(array(map(1, coalesce(id, 0))), 1, 1) FROM t;

```



I'm totally down we need to fix all of issues above, they just beyond the scope of the original issue. what about creating smaller follow up PRs to address issues above?

sunchao (commented):
@comphead, smaller follow-ups are reasonable, provided this PR preserves the previous safe behavior for the affected cases. The remaining **[P2]** concerns come from the actual BASE-to-HEAD changes:

- The TRY/bracket and collation examples contain folded map literals. [BASE declines those non-null complex literals](https://github.com/apache/datafusion-comet/blob/e0ab0a6fe60c05bd679654f0201ebe88319cc3a7/spark/src/main/scala/org/apache/comet/serde/literals.scala#L43-L64); [HEAD newly admits and rebuilds them](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/literals.scala#L43-L86), exposing the consumer limitations. That admission is not restricted to an outer array-of-map literal.
- The stateful example exercises the [new ANSI guard](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L623-L696), which serializes the left operand twice. [BASE has no second serialization](https://github.com/apache/datafusion-comet/blob/e0ab0a6fe60c05bd679654f0201ebe88319cc3a7/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L601-L639). This change affects nullable array lookups beyond literal expansion.
- The Slice example does construct an array of maps. The [new `deepNullable` casts](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L520-L534) change its nested map type, while [Slice still declares the original element type](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L465-L484). BASE passes that single map child through without the new cast.

So a narrow solution here is fine: restrict the newly enabled paths or retain safe fallback until the corresponding consumer fix lands; broader collation/collection improvements can remain separate PRs. The scope distinction above is from pinned source comparison, not an additional executed reproduction.

sunchao:
@comphead yes, since this PR has grown to somewhat large, please feel free to fix the remaining issues in the follow-ups.

comphead:
> @comphead yes, since this PR has grown to somewhat large, please feel free to fix the remaining issues in the follow-ups.

Filed https://github.com/apache/datafusion-comet/issues/5544, address them there

comphead:
Resolved conflicts, appreciate if we can get this approved and move fwd with followups

sunchao (approved):
Re-reviewed ec9832366. LGTM with the remaining issues tracked in #5544 as agreed. Please wait for the remaining CI checks before merging.

sunchao:
The final pass of `ec9832366` identified two **P1 issues**, now recorded with reproductions and validation limits in [#5544](https://github.com/apache/datafusion-comet/issues/5544#issuecomment-5464398708):

- Newly admitted malformed UTF-8 map literals can reach native string operations and abort the JVM with **SIGABRT**. This was reproduced with the debug native build; release-build failure behavior was not tested.
- The ANSI `element_at` guard reevaluates nullable stateful operands, silently changing values and NULL placement in already-supported primitive-array queries. A simple counter example returns only four of eight expected non-NULL values. This raises the priority of the previously reported stateful-operand P2.

Both were independently reproduced on Spark 4.1.3 / JDK 17, and the exact-base serializer controls pass on the same native/JVM runtime. Those controls are not a full base build.

These results change my earlier LGTM recommendation: I recommend restoring safe fallback for these cases before merging. The broader fixes can remain follow-ups in #5544.

comphead:
Thanks @sunchao taking a look into followup

PR #5458, opened 2026-08-24
docs: document C2R cost for wide/nested schemas in tuning guide

### LINKED ISSUE #5455: Document C2R cost for wide/nested schemas

Recently working with wide/nested schemas and found c2r is very expensive. 

Image

It would be reasonable to document that entire fallback would be cheaper if schema is wide/complex. For example setting 

```
comet.exec.transitionRevert.enabled=true
comet.exec.transitionRevert.maxTransitions=1
```

_No response_

### Additional context

_No response_

DebadityaHait:

Closes #5455.

## Rationale for this change

The tuning guide documents the transitionRevert mechanism but never mentions schema shape as the cost driver. As reported in #5455, columnar-to-row conversion becomes very expensive for wide or deeply nested schemas, where reverting the entire stage can be cheaper than paying repeated transitions.

Extends the "Reducing Row/Columnar Conversion Overhead" section of the tuning guide with one short paragraph covering schema-shape-driven C2R cost and earlier-revert guidance using full config keys.

Docs-only change; validated with a clean local Sphinx build of the site (no warnings for the changed file). Config keys and defaults match their definitions in `CometConf.scala`.

### DIFF
--- a/docs/source/user-guide/latest/tuning.md
+++ b/docs/source/user-guide/latest/tuning.md
@@ -273,6 +273,14 @@ when the number of columnar-to-row transitions exceeds
 `spark.comet.exec.transitionRevert.maxTransitions` (default `2`). This trades native execution of a small
 subset of operators for eliminating conversion overhead across the stage.
 
+The cost of each conversion also grows sharply with schema shape: for wide or deeply nested schemas,
+columnar-to-row conversion is especially expensive because the conversion work scales with the number of
+columns and nested fields. If profiling shows these conversions dominating a query stage over such a schema,
+set `spark.comet.exec.transitionRevert.enabled=true` and lower
+`spark.comet.exec.transitionRevert.maxTransitions` to `1` so that the stage reverts entirely to Spark
+row-based execution once it needs more than one conversion, which can be cheaper than paying the conversion
+repeatedly across many operators.
+
 ## Metrics Overhead
 
 Comet exposes rich native operator metrics for observability (see [Metrics](metrics.md)), but they are
--- a/docs/source/user-guide/latest/tuning.md
+++ b/docs/source/user-guide/latest/tuning.md
@@ -273,13 +273,16 @@ when the number of columnar-to-row transitions exceeds
 `spark.comet.exec.transitionRevert.maxTransitions` (default `2`). This trades native execution of a small
 subset of operators for eliminating conversion overhead across the stage.
 
+### Entire-Plan Fallback for Wide or Deeply Nested Schemas
+
 The cost of each conversion also grows sharply with schema shape: for wide or deeply nested schemas,
 columnar-to-row conversion is especially expensive because the conversion work scales with the number of
-columns and nested fields. If profiling shows these conversions dominating a query stage over such a schema,
-set `spark.comet.exec.transitionRevert.enabled=true` and lower
-`spark.comet.exec.transitionRevert.maxTransitions` to `1` so that the stage reverts entirely to Spark
-row-based execution once it needs more than one conversion, which can be cheaper than paying the conversion
-repeatedly across many operators.
+columns and nested fields. If profiling shows these conversions dominating a query over such a schema, set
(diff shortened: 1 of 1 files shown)


comphead (commented):
Thanks @DebadityaHait lets have this note in separate paragraph under h3 heading, please make sure to emphasize, its not a stage fallback, rather the entire plan fallback, because c2r for complex deep schemas might be very expensive

sunchao (approved):
## Summary

This documentation-only change connects schema-dependent columnar-to-row conversion cost to Comet's existing transition-revert controls. I found no P1/P2 issue in the complete diff from `ec0f7975dcef8e658a39b273cd491352147ccd66` to `3849ef6d4c9741da99fbd985640f47d237c9cb1e`; it does not change execution code or configuration defaults.

## Prior state and problem

The tuning guide already explains why repeated row/columnar transitions can outweigh native execution benefits, but it does not identify wide and deeply nested schemas as a source of additional conversion work. Readers investigating those workloads could therefore miss a useful reason to profile transitions and consider an earlier fallback.

## Design approach

The new paragraph stays alongside the existing transition-revert explanation and provides both fully qualified configuration keys. Its recommendation is conditional on profiling showing conversion-dominated execution, and “can be cheaper” correctly leaves the performance decision workload-dependent rather than promising a universal improvement.

The keys and defaults match [the exact-head configuration definitions](https://github.com/apache/datafusion-comet/blob/3849ef6d4c9741da99fbd985640f47d237c9cb1e/spark/src/main/scala/org/apache/comet/CometConf.scala#L451-L475): reversion is disabled by default and the maximum is `2`. The rule uses a strict `transitionCount > maxTransitions` comparison, so setting the maximum to `1` permits zero or one C2R and reverts a stage with at least two.

The current implementation is stage-scoped in both AQE and non-AQE execution: transition counting and rewriting stop at query-stage, shuffle, and broadcast boundaries. Thus the paragraph's “stage” wording agrees with [the pinned implementation](https://github.com/apache/datafusion-comet/blob/3849ef6d4c9741da99fbd985640f47d237c9cb1e/spark/src/main/scala/org/apache/comet/rules/RevertNativeForTransitionHeavyStages.scala#L45-L145); these settings do not unconditionally revert every stage in the whole query.

## Key design decisions

Including the enable switch is important because changing the threshold alone has no effect while the rule is disabled. Lowering the threshold is an explicit tuning choice, while retaining the profiling prerequisite accounts for native-execution benefits, fixed conversion costs, and differences between JVM/native conversion paths.

## Implementation sketch

The patch adds eight lines to `docs/source/user-guide/latest/tuning.md` and changes no other file. A Markdown rendering check confirms one separate paragraph with all three inline-code spans intact and surrounding headings unchanged.

Validation included the full pinned diff, configuration/rule/conversion-path source inspection, and that rendering check. CI has five successful checks and twelve skipped checks, including runtime suites and site deployment; I did not independently run a full Sphinx build or a performance benchmark.

There is no runtime behavior change until a reader explicitly applies the documented settings. Opting in can give up native execution in qualifying stages, and required format bridges at stage boundaries may remain, so the advice should not be read as eliminating every conversion in the query.

## Suggested improvements

The existing suggestion to give this schema-specific advice its own h3 is a reasonable optional organizational improvement, not a correctness blocker. Any wording revision should retain the distinction between a per-stage C2R threshold and unconditional whole-query fallback.

sunchao:
@DebadityaHait could you address the comment from @comphead in the above? after that I'll merge this. Thanks

DebadityaHait:
Thanks @comphead @sunchao — addressed in `d13c7135`: the guidance now sits under its own h3 heading ("Entire-Plan Fallback for Wide or Deeply Nested Schemas") and states explicitly that the fallback applies to the entire plan, with Comet removing its native operators rather than leaving a mix of native and fallback operators joined by repeated conversions.

sunchao:
Thanks, merged!

PR #5459, opened 2026-08-24
ci: shard Iceberg Spark tests across four runners

sunchao:

The main Iceberg Spark test task is one of Comet CI's longest paths. In a [recent successful job](https://github.com/apache/datafusion-comet/actions/runs/32751494106/job/97518377579), test execution consumed about 67 minutes, including 18.7 minutes in `TestStructuredStreamingRead3` and 10.7 minutes in `TestRewriteDataFilesAction`. The job finished after 78 minutes including setup and compilation.

**Which issue does this PR close?** None. This is a CI scheduling improvement that retains the existing supported versions and test coverage.

Split the main Iceberg Spark test task across four isolated workers. Keep the long structured-streaming family together and assign all remaining classes automatically by a stable hash. New classes therefore receive a worker without maintaining a test allowlist. Nested classes stay with their enclosing class, and all classes remain available on the test classpath for inherited fixtures and framework discovery.

A workflow-side Gradle init script adds an exclusion predicate to the existing candidate selection. It does not replace Gradle's existing includes/excludes or JUnit's tag and parameterized-test handling. Each worker uploads both its unsharded and selected candidate inventories alongside the JUnit reports, with distinct names for reruns. A dependent job checks the complete index set, identical unsharded baselines, and selected sets whose disjoint union equals the baseline. Matrix indices and the Gradle partition count now come from one definition, independent of `strategy.job-total`.

Run a lightweight coverage-equivalence fixture once per Iceberg version in the existing short runtime job. Native compilation, extensions, runtime-test commands, and existing Comet/Gradle build flags are otherwise unchanged. This PR intentionally leaves repeated JVM compilation for separate work.

## How was this PR tested?

- The original fixture passed with Gradle 8.14.4/JDK 17 and Gradle 8.12.1/JDK 11. For this review follow-up, reran the updated fixture with Gradle 8.14.4/JDK 17: the four workers match the unsharded set of 10 candidate classes and 12 executed cases exactly once, and each captured unsharded inventory matches the independently executed baseline.
- Added 15 fast regression tests for real-inventory validation, including missing/extra/duplicate classes, missing/extra indices, mismatched counts/tasks/baselines, malformed metadata, and failed-job reruns. All pass and run in preflight.
- Covered nested, parameterized, inherited, dynamic, and newly named tests; existing include/exclude patterns, exclusion predicates, and JUnit tags; single-worker equivalence; and isolation from another Test task.
- Verified that a failing test still fails Gradle, and invalid worker indices or task names fail explicitly.
- Checked the streaming family in all four configured Iceberg releases, the change filters, and the unchanged native/extension jobs and existing test flags. Actionlint, Python syntax, Apache RAT, and `git diff --check` pass on the review follow-up. Documentation formatting was checked on the original revision but was not rerun locally for this follow-up.

A complete successful four-shard Iceberg run and its total elapsed time remain unmeasured. The previous unsharded job spent about 67 minutes testing and 11 minutes on setup/compilation. Assuming the same test work and 11 minutes of repeated setup per worker, four shards would use about **111 runner-minutes per version** (`67 + 4 × 11`), compared with 78 previously: **33 extra minutes, or roughly 42%**. This is an estimate, not a measured after result, and excludes the small inventory-check job. The existing successful shards 2/3/4 took 26m54s, 28m08s, and 25m05s, but shard 1 did not complete, so these cannot establish a full-run total or whole-CI critical-path improvement. Compilation sharing or fewer shards can be evaluated against a complete run; neither benefit is assumed here. The main test workers have new shard-suffixed check names, so any external requirements that refer to the old individual check should be checked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,9 @@ jobs:
       - name: Check pull request type labeling
         run: node --test dev/ci/pr-type-label.test.mjs
 
+      - name: Check Iceberg shard inventory validation
+        run: python3 dev/ci/test-iceberg-shards.py
+
       - name: Install actionlint
         run: |
           curl -sSfL https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash | bash
--- a/.github/workflows/iceberg_spark_test_reusable.yml
+++ b/.github/workflows/iceberg_spark_test_reusable.yml
@@ -111,7 +111,11 @@ jobs:
 
   iceberg-spark:
     needs: build-native
-    name: iceberg-spark/iceberg-${{ inputs.iceberg-full }}/spark-${{ inputs.spark-full }}/scala-${{ inputs.scala }}/java-${{ inputs.java }}
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4]
+    name: iceberg-spark/iceberg-${{ inputs.iceberg-full }}/spark-${{ inputs.spark-full }}/scala-${{ inputs.scala }}/java-${{ inputs.java }}/shard-${{ matrix.shard }}
     runs-on: ubuntu-24.04
     container:
       image: amd64/rust
@@ -142,7 +146,19 @@ jobs:
           rm -rf /root/.m2/repository/org/apache/parquet # somehow parquet cache requires cleanups
           ENABLE_COMET=true ENABLE_COMET_ONHEAP=true ./gradlew -DsparkVersions=${{ inputs.spark-short }} -DscalaVersion=${{ inputs.scala }} -DflinkVersions= -DkafkaVersions= \
             :iceberg-spark:iceberg-spark-${{ inputs.spark-short }}_${{ inputs.scala }}:test \
+            --init-script ../dev/ci/iceberg-test-shards.gradle \
+            -PcometShardTask=:iceberg-spark:iceberg-spark-${{ inputs.spark-short }}_${{ inputs.scala }}:test \
+            -PcometShardIndex=${{ matrix.shard }} -PcometShardCount=${{ strategy.job-total }} \
             -Pquick=true -x javadoc
+      - name: Upload Iceberg shard inventory and test reports
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: iceberg-spark-${{ inputs.iceberg-full }}-spark-${{ inputs.spark-full }}-scala-${{ inputs.scala }}-jdk${{ inputs.java }}-shard-${{ matrix.shard }}-attempt-${{ github.run_attempt }}
+          path: |
+            apache-iceberg/**/build/comet-shards/*.json
+            apache-iceberg/**/build/test-results/test/*.xml
+          retention-days: 7
 
   iceberg-spark-extensions:
     needs: build-native
@@ -206,6 +222,8 @@ jobs:
         uses: ./.github/actions/setup-iceberg-builder
         with:
           iceberg-version: ${{ inputs.iceberg-full }}
+      - name: Verify test sharding preserves discovery
+        run: python3 dev/ci/check-iceberg-shards.py --gradle "$PWD/apache-iceberg/gradlew"
       - name: Run Iceberg Spark runtime tests
         run: |
           cd apache-iceberg
--- a/.github/workflows/iceberg_spark_test_reusable.yml
+++ b/.github/workflows/iceberg_spark_test_reusable.yml
@@ -63,6 +63,9 @@ jobs:
   build-native:
     name: Build Native Library
     runs-on: ubuntu-24.04
+    outputs:
+      shard-matrix: ${{ steps.shards.outputs.matrix }}
+      shard-count: ${{ steps.shards.outputs.count }}
     container:
       image: amd64/rust
     steps:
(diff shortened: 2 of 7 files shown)


andygrove:

Cutting the Iceberg path from 78 minutes is worth doing, and I like that the split is a Gradle exclusion predicate layered on top of the existing candidate set rather than an allowlist someone has to maintain. Keeping `@Nested` classes with their enclosing class and leaving everything on the classpath are the right details to get right.

Three things I would want addressed.

**`strategy.job-total` will silently drop tests if the matrix ever gains a dimension**

The workflow passes `-PcometShardCount=${{ strategy.job-total }}`, which today equals 4 because `shard` is the only matrix key. If anyone later adds a second dimension, say `jdk: [11, 17]`, `job-total` becomes 8 while `matrix.shard` still ranges 1 through 4. Classes hashing to shards 5 through 8 would then be excluded from every worker and simply never run, with a green build. Could this pass `4` literally, or derive the count from the length of the shard list? A silent coverage hole in the Iceberg suite is exactly the kind of thing nobody notices for months.

**Nothing verifies that the four real shards cover the real test set**

`check-iceberg-shards.py` is a nice piece of work, but it validates the init script against a synthetic 10-class fixture in a different job. The actual Iceberg run uploads four `comet-shards/*.json` manifests as artifacts and then nothing reads them. The property we care about is "the union of the four real shards equals what the unsharded run would have selected, with no class appearing twice," and that is currently unchecked.

Would you add a small follow-up job that depends on the four shards, downloads the manifests, and asserts the union has no duplicates and matches the expected class count? That is cheap and it turns the artifact upload from something you look at during a postmortem into something that fails the build. Without it, the failure mode of a bad hash change is silently running fewer tests.

**The runner-minute cost is not stated**

The description quotes 67 minutes of test execution inside a 78-minute job, so roughly 11 minutes of setup, container pull, and compilation. That 11 minutes is now paid four times per Iceberg version, and the description explicitly defers the repeated-compilation work. Wall clock clearly improves. Total consumption on shared ASF runners goes up, and by how much is the number that matters for whether this is a net win. Could you put the before and after total runner minutes in the description? If the increase is significant it might be worth landing the compilation sharing first, or starting with two shards instead of four.

**One smaller thing**

The comment in `CometIcebergTestShard.owner` records concrete minute figures from a 1.11 run. Those will go stale and there is nothing that notices when the buckets drift apart. Since each shard's duration is visible in the Actions UI anyway, would it be better to say "sized against a 1.11 run in <month>, recheck if a shard becomes the critical path" rather than baking in numbers that will silently become wrong?

sunchao:
Updated in [913354ca0](https://github.com/apache/datafusion-comet/commit/913354ca0475770421f961f29aa78d6f1e94a2c0).

Addressed the count and real-inventory validation gaps in this update.

The workflow matrix and Gradle partition count now come from the same `SHARD_COUNT` definition. Neither uses `strategy.job-total`, so adding another matrix dimension cannot silently change ownership to nonexistent shard indices.

Each real Iceberg worker now captures the unsharded Gradle candidate set with only our predicate temporarily disabled, restores the predicate, and records the selected set before running tests. A dependent job downloads the matching configuration's artifacts, requires every shard index, checks that the independently captured baselines agree, and verifies that the selected sets are disjoint and cover the baseline exactly once. It selects each shard's latest available attempt so failed-job reruns can retain earlier successful workers. There is no fixed release-specific class count. JUnit execution is still checked separately by the synthetic fixture, since candidate classes are not executed test cases.

Validation: all 15 inventory-validator regressions pass; the updated Gradle 8.14.4/JDK 17 fixture verifies 10 candidate classes and 12 executed cases exactly once, including nested/parameterized/inherited/dynamic cases, upstream exclusions, another task's isolation, and failure propagation. Actionlint 1.7.12, Python syntax, Apache RAT, and `git diff --check` pass. The real Iceberg artifact check is wired into CI but awaits a complete hosted run; I have not represented it as a locally completed full Iceberg suite.

I also made the cost estimate explicit in the description: using the previous 67 test minutes plus 11 setup/compile minutes, four workers are approximately **111 runner-minutes versus 78** per version, or **+33 minutes / +42%**, excluding the small inventory-check job. That assumes unchanged total test work and repeated setup, and is not a measured after result. Shards 2/3/4 completed in 26m54s, 28m08s, and 25m05s, but shard 1 did not finish, so those partial results cannot establish the full total or whole-CI critical-path improvement. Compilation sharing or fewer shards should be judged against a complete run.

The balancing comment remains explicitly historical: it describes the Iceberg 1.11 run used to size the shards, not a current timing guarantee.

andygrove (approved):
`check-iceberg-shards.py` now defines `SHARD_COUNT` once and feeds it to both the workflow matrix and `-PcometShardCount`, so `strategy.job-total` is out of the picture. That was the silent-drop risk I flagged if the matrix ever grew another dimension, and it's gone now.

`iceberg-spark-shard-coverage` covers the other gap I raised. It downloads each shard's manifest, keeps the latest attempt per index, and fails if any index is missing or the selected candidates don't union to the unsharded baseline exactly once. That turns the coverage property into something the real Iceberg run checks on every PR, not just the synthetic Gradle fixture.

The runner-minute cost is in the description now, roughly 111 minutes versus 78 per version, clearly marked as an estimate rather than a measured result. That's what I wanted stated plainly before this merges.

On the `CometIcebergTestShard.owner` comment, I'll take the point that it documents how the buckets were sized rather than promising a current runtime, so I won't press on that further.

andygrove:
@sunchao this is conflicting with main now. Could you rebase? Everything was green beforehand.

sunchao:
Thanks @andygrove . Just rebased

andygrove:
Rebase checked. I diffed the net change against `913354ca0`, the head I approved, and it is byte-identical, so this is a pure rebase and my approval stands. Good to merge once CI comes back.

sunchao:
Merged, thanks!

PR #5460, opened 2026-08-24
chore: add md formatting to `make format`

comphead:

Closes #.

## Rationale for this change

Adding md files formatting to `make format` to have all formatting under the same umbrella

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,11 @@ clean:
 bench:
 	cd native && RUSTFLAGS="-Ctarget-cpu=native" cargo bench $(FEATURES_ARG) $(filter-out $@,$(MAKECMDGOALS))
 format:
+	@if command -v prettier >/dev/null 2>&1; then \
+		prettier -w "**/*.md"; \
+	else \
+		echo "prettier not found, skipping markdown formatting (npm i -g prettier)"; \
+	fi
 	cd native && cargo fmt
 	./mvnw compile test-compile scalafix:scalafix -Psemanticdb $(PROFILES)
 	./mvnw spotless:apply $(PROFILES)
--- a/docs/source/contributor-guide/native_shuffle.md
+++ b/docs/source/contributor-guide/native_shuffle.md
@@ -49,7 +49,6 @@ Native shuffle (`CometExchange`) is selected when all of the following condition
    columnar output. Row-based Spark operators require JVM shuffle.
 
 3. **Supported partitioning type**: Native shuffle supports:
-
    - `HashPartitioning`
    - `RangePartitioning`
    - `SinglePartition`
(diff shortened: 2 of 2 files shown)


peterxcli (approved):
make sense. thanks!

sunchao (approved):
## Summary

Reviewed `ec0f7975dcef` → `e3f26c88e926`: no introduced, actionable P1/P2 findings. This is a focused contributor-workflow change that brings Markdown formatting into `make format`; the two-file diff leaves Comet's native/JVM implementation and runtime configuration unchanged.

## Prior state and problem

`make format` previously ran Rust formatting, Scalafix, and Spotless, while Markdown required a separate Prettier invocation. CI already checked Markdown with the same recursive file pattern, so the usual local formatting entry point did not cover an existing CI requirement.

## Design approach

The new recipe checks for an installed `prettier` executable and runs it in write mode over the quoted `**/*.md` pattern. If it is unavailable, the recipe prints an installation hint and continues with the existing formatters; it does not install dependencies automatically.

The quoted pattern reaches Prettier intact, and current Prettier 3.x observes the repository's ignore rules, consistent with its [documented ignore handling](https://prettier.io/docs/ignore). Isolated probes using exact base/head Makefiles verify missing-tool behavior, argument passing, failure propagation, serial recipe ordering under `-j8`, dry-run behavior, and invocation through `-C`; an existing file named `format` suppresses the target on both base and head and is not a new regression.

The exact before/after Markdown snapshots have identical normalized ASTs and rendered HTML under markdown-it-py (CommonMark with table support), and Prettier 3.9.6 checks pass on both. CI now reports 43 successful checks and 10 skipped, with the [macOS Spark 4.0 scans retry](https://github.com/apache/datafusion-comet/actions/runs/32787119106/job/97682544169) passing at the unchanged head. The [original failed attempt](https://github.com/apache/datafusion-comet/actions/runs/32787119106/job/97628182198) has a crash artifact showing a native-thread SIGSEGV during `_pthread_tsd_cleanup`; the workflow invokes Cargo/Maven directly and never runs the changed `format` target. This was an undiagnosed runtime crash in an unchanged execution path, not evidence of a formatting regression. Local validation used isolated formatter/stub probes and Markdown comparisons, not a full Sphinx or Rust/JVM build.

## Key design decisions

Reusing an installed formatter keeps Node/Prettier optional for contributors who only need Rust/JVM formatting. Markdown enforcement remains in CI, so the local skip path does not weaken the repository's existing formatting gate.

## Implementation sketch

Five lines add a POSIX-shell conditional at the beginning of the existing `format` recipe, followed by the same Cargo and two Maven commands. The documentation change removes six optional blank lines before nested lists without changing the described shuffle implementation.

With Prettier installed, `make format` now rewrites matching Markdown files before running the other formatters. An actual Prettier failure stops the target, whereas an absent executable emits a warning and preserves the previous formatting workflow.

## Suggested improvements

No implementation change is requested for this diff. The original native-thread teardown crash is a separate follow-up if it recurs; the successful retry does not establish its underlying cause.

PR #5461, opened 2026-08-25
fix: release native shuffle reservation after spill failure

peterxcli:

Part of #5212 (finding 13).

## Rationale for this change

`MultiPartitionShuffleRepartitioner::spill` takes ownership of its buffered batches and partition indices before calling `PartitionWriter::write`. If a write failed, the early return skipped reservation release and pinned/repeated-spill cleanup even though those batches had already been dropped.

The deterministic reproducer left 96 bytes reserved after the injected failure. It also retained one pinned buffer and 32 repeated-spill bytes while the buffered batches and indices were already empty.

- Preserve the first partition-writer error while running the existing spill cleanup unconditionally.
- Drop the taken partitioned batches before releasing their reservation.
- Add a fake failing `PartitionWriter` test covering exact reservation, pinned/repeated state, and spill-metric deltas on both success and failure paths.

- `cargo test -p datafusion-comet-shuffle --lib` (35 tests)
- `cargo clippy -p datafusion-comet-shuffle --lib --tests -- -D warnings`

### DIFF
--- a/native/shuffle/src/partitioners/multi_partition.rs
+++ b/native/shuffle/src/partitioners/multi_partition.rs
@@ -566,15 +566,17 @@ impl<T: PartitionWriter> MultiPartitionShuffleRepartitioner<T> {
 
         with_trace("shuffle_spill", self.tracing_enabled, || {
             let num_output_partitions = self.partition_indices.len();
-            let mut partitioned_batches = self.partitioned_batches();
-
-            for partition_id in 0..num_output_partitions {
-                self.partition_writer.write(
-                    partition_id,
-                    &mut partitioned_batches.produce(partition_id, &self.metrics.interleave_time),
-                    &self.metrics,
-                )?;
-            }
+            let write_result = {
+                let mut partitioned_batches = self.partitioned_batches();
+                (0..num_output_partitions).try_for_each(|partition_id| {
+                    self.partition_writer.write(
+                        partition_id,
+                        &mut partitioned_batches
+                            .produce(partition_id, &self.metrics.interleave_time),
+                        &self.metrics,
+                    )
+                })
+            };
 
             let memory_spilled_bytes = self
                 .reservation
@@ -585,7 +587,7 @@ impl<T: PartitionWriter> MultiPartitionShuffleRepartitioner<T> {
             self.pinned_buffers.clear();
             self.repeated_spill_buffer_bytes = 0;
             self.metrics.spill_count.add(1);
-            Ok(())
+            write_result
         })
     }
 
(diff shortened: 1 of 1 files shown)


sunchao:
cc @ywskycn are you also working on the same thing?

peterxcli:
@ywskycn sorry I dont know that. if you have prior work, please feel free to close this, happy to review then :)

sunchao (approved):
## Summary

Reviewed the complete one-file diff at head `b44eafe4797d90ea11e31fc0aa7e7556b8dc3578` against base `bd7dc601aa9a99289530598d4778c40e6ded7772`. Five independent nathree/Ultra review scopes covered correctness and edge cases, compatibility and semantics, tests and CI, performance/concurrency/resource safety, and adversarial cross-checking and discussion deduplication. No outstanding verified P1/P2 findings remain; approving this focused spill-error cleanup.

## Prior state and problem

`spill()` transfers buffered batches and partition indices into `PartitionedBatchesProducer` before writing partitions. Previously, `write(...)?` returned immediately on a writer error: the producer was dropped, but `reservation.free()`, pinned-buffer cleanup, repeated-spill accounting reset, and spill-metric updates were skipped. The repartitioner could therefore retain reservation/accounting state for batches it no longer owned until its eventual destruction.

## Design approach

The partition loop now retains its `Result` using `try_for_each`, preserving ascending partition order and stopping at the first error. An explicit inner scope ensures the producer and its remaining batches and indices are dropped before reservation release. The existing cleanup then executes for either result, after which the original result is returned.

The error remains an error at the `spill()` boundary; subsequent partitions are not written after a failure. The concrete multi-partition spill writer keeps its batch writer/coalescer local to the write call, so those temporary batch references also drop before reservation release. The caller propagates an insertion failure without proceeding to final shuffle output, so this change does not turn a partial write into successful output.

`MemoryReservation::free()` clears the reservation, preventing a later destructor from releasing the same bytes again. `unreserved_bytes` remains an input to spill metrics, not an amount returned to the pool. Successful partitioning, serialization, writer contracts, and the single-partition and empty-schema paths are unchanged. The synchronous, exclusively borrowed spill loop adds no suspension point, lock, batch copy, or new allocation.

Validation: the [Rust CI job](https://github.com/apache/datafusion-comet/actions/runs/32855355627/job/97826718236) ran `RUST_BACKTRACE=1 cargo nextest run` and reported **901 passed, 4 skipped**, including **35 passing shuffle tests** and `spill_write_error_releases_buffered_memory`. Its checkout, `dd5ec42bdfa2b68b8abe653e51cf4f367a0bc34c`, merges the exact head and base above and has the same source tree as the reviewed head. The PR check snapshot has **63 successful and 9 skipped checks**, with no failures or pending checks. These are independently inspected CI results; no local build or test run was performed for this review.

## Key design decisions

- Keep error selection separate from cleanup: retaining the first `Result` avoids continuing writes or replacing the original failure.
- Drop the owned producer before releasing its memory charge, making physical ownership and reservation lifetime agree on both result paths.
- Clear `pinned_buffers` and `repeated_spill_buffer_bytes` when the buffer drains, while preserving the different lifetime of `spill_accounted_input_buffers` until the outer input finishes. The existing saturating accounting formula is retained.

## Implementation sketch

The production change is confined to the write-result scope and the final return value in `MultiPartitionShuffleRepartitioner::spill`. The regression first completes a successful spill, then buffers another slice sharing the same backing allocation, verifies a nonzero reservation and repeated-buffer state, and injects an error before the first partition iterator is consumed.

The assertions cover the unchanged error, write short-circuiting, zero reservation and pool usage, drained batches/indices, cleared pinned/repeated state, retained outer-input deduplication state, and exact metric deltas. Source tracing shows why the original early return would fail the post-error reservation assertion; that base-version failure was not executed locally. Partial-consumption and later-partition failures were checked through ownership/control-flow analysis rather than separate injected tests.

A failed spill now promptly releases its drained input reservation and clears the associated buffer tracking before returning the failure. Spill count and memory-spill metrics also include this failed attempt's cleanup, with already-accounted shared input buffers excluded as before. Physical spill-byte accounting remains the writer's responsibility; the change does not claim that discarded input was successfully persisted or add retry or panic-recovery semantics.

## Suggested improvements

No required P1/P2 changes remain after the five scopes and the parent review. Existing REST and GraphQL discussions contain contributor coordination only, with no technical review finding to duplicate.

sunchao:
Merged, thanks @peterxcli !

Issue #5462, opened 2026-08-25
Track exact SpillWriter page bytes for O(1) memory usage sampling


peterxcli:
## What is the problem the feature request solves?

`SpillWriter.getMemoryUsage()` walks every entry in `allocatedPages` to sum the current page bytes. This is normally cheap because shuffle pages are large, but it is called from JVM shuffle spill accounting and from the bypass-shuffle "largest writer first" selection path. A writer with many small pages can therefore pay repeated O(page count) sampling costs.

This is follow-up performance work for finding #12 in #5212. The correctness fix should remain narrowly scoped; this issue tracks whether exact O(1) page-byte accounting is worthwhile.

Using `allocatedPages.size() * configuredPageSize` would not be exact:

- `CometBoundedShuffleMemoryAllocator` allocates `max(pageSize, required)`, so an oversized record can create a larger page.
- `CometUnifiedShuffleMemoryAllocator` can receive a smaller partial grant that still satisfies `required`.
- Allocator totals cannot be used generically because allocator scope differs and the bounded allocator is shared by multiple writers.

## Describe the potential solution

Maintain exact writer-local allocated page bytes in `SpillWriter`:

- Add each successfully allocated block's actual `MemoryBlock.size()` at both allocation sites.
- Subtract the actual byte count returned by `allocator.free(block)` for each freed block.
- Return the tracked value from `getMemoryUsage()` in O(1).
- Continue adding the pointer-array usage separately in `SpillSorter.getMemoryUsage()`.

The value needs cross-thread visibility because `CometDiskBlockWriter.ArrowIPCWriter` memory can be read while selecting another writer to spill. Avoid making the base allocation/free methods uniformly synchronized without first checking the existing `currentWriters` and per-writer lock ordering.

## Additional context

Before implementation, establish that this improves end-to-end behavior rather than only a scan microbenchmark:

- Cover normal, oversized, and partial page allocations with exact byte assertions.
- Cover free/reset, repeated or idempotent frees, and partial-failure accounting.
- Exercise cross-thread `ArrowIPCWriter` reads.
- Run warmed repeated A/B measurements for 1, 16, 256, and 1,024 pages, plus tiny and representative end-to-end shuffle spills.
- Keep the change only if the end-to-end gain is measurable and spill bytes, output, and correctness remain unchanged.

peterxcli:
take

peterxcli:
seems like there is no sigificant improvement by doing this.
```java
[...]

Issue #12, opened 2024-02-12
Join support


viirya:
This is an umbrella ticket for adding Join support to Comet. In Spark, there are basically three types of Join operators: BroadcastJoin, HashJoin, SortMergeJoin. In DataFusion, two Join operators are supported: HashJoin, SortMergeJoin (experimental).

### SortMergeJoin

- [x] Relax join keys constraint from Column to any physical expression for physical join operators https://github.com/apache/arrow-datafusion/pull/8991
- [x] Support join filter for SortMergeJoin https://github.com/apache/arrow-datafusion/pull/9080
- [ ] Support SortMergeJoin operator - will be open sourced later
...

### BroadcastJoin

- [ ] Add BosonBroadcastExchangeExec support broadcasting the result of Comet native operator - will be open sourced later

SortMergeJoin and HashJoin are blocked by shuffle support and some patches. We've implemented shuffle internally which will be open sourced later.

mwiewior:
Hey @viirya - it's really awesome ! Do you happen to know when  approx these BHJ and shuffle components will be open sourced?

viirya:
Hi @mwiewior

I will open source shuffle components soon. For join support, it is in development now. We already have working code but get some test failure on Spark tests. Once these test failures are resolved, we will open source it.

viirya:
For shuffle, we just remove it from the initial open source code. So the code is ready, we only need to add it back to Comet.

mwiewior:
Great news ! Thx !

coderfender:
https://github.com/apache/datafusion-comet/issues/198 https://github.com/apache/datafusion-comet/issues/199

coderfender:
I will be working on https://github.com/apache/datafusion-comet/issues/199 next

coderfender:
Working on cartesian join. Fixing errors at the moment

PR #5463, opened 2026-08-25
fix: track peak memory before JVM shuffle spill

peterxcli:

Part of #5212 (finding #12). This does not close the parent audit epic.

Follow-up performance work for O(1) page-byte accounting is tracked in #5462.

## Rationale for this change

`CometShuffleExternalSorter.spill()` frees the active sorter's data pages without first updating `peakMemoryUsedBytes`. If an earlier spill has a larger working set than later spills, the final getter only observes the reset pointer array and under-reports the actual peak.

On unmodified upstream main, a deterministic two-spill reproducer reached a known first-spill high-water mark of 1,792 bytes but reported 1,024 bytes.

- Update peak memory immediately before an active spill sorter's pages are freed.
- Add a focused two-spill regression whose first working set is larger and which does not poll the peak between spills.
- Assert exact allocator/spill relationships, two spill files, record count, disk-byte accounting, partition lengths, and cleanup.

- `make core`
- `./mvnw -q -Prelease test -Dtest=none -Dsuites="org.apache.spark.sql.comet.CometTaskMetricsSuite JVM shuffle peak memory includes a larger earlier spill" -Djacoco.skip=true`
  - Exactly 1 test expected and run; 1 succeeded.
- `./mvnw -q spotless:check -DskipTests -Dscalastyle.skip=true`
- `git diff --check`

The regression failed against unmodified upstream main with `1024 did not equal 1792` after its preceding spill-count, record-count, and exact byte-accounting assertions passed. It passes with this change.

A warmed release-mode A-B-B-A check used 5 warmups and 15 recorded samples per block for tiny and representative two-spill workloads. Combined medians were 1.705 ms before versus 1.698 ms after for the tiny workload, and 52.65 ms before versus 53.90 ms after for the representative workload; both differences were within measured variability. Spill bytes, record counts, partition lengths, and spill-file SHA-256 digests were unchanged across all recorded samples.

### DIFF
--- a/spark/src/main/java/org/apache/spark/shuffle/sort/CometShuffleExternalSorter.java
+++ b/spark/src/main/java/org/apache/spark/shuffle/sort/CometShuffleExternalSorter.java
@@ -196,6 +196,7 @@ public final class CometShuffleExternalSorter implements CometShuffleChecksumSup
     activeSpillSorter.setSpillInfo(spillInfo);
 
     activeSpillSorter.writeSortedFileNative(false, tracingEnabled);
+    updatePeakMemoryUsed();
     final long spillSize = activeSpillSorter.freeMemory();
     activeSpillSorter.reset();
 
--- a/spark/src/test/scala/org/apache/spark/sql/comet/CometTaskMetricsSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/comet/CometTaskMetricsSuite.scala
@@ -23,13 +23,17 @@ import java.io.File
 
 import scala.collection.mutable
 
-import org.apache.spark.{SparkConf, SparkContext, Success}
+import org.apache.spark.{SparkConf, SparkContext, SparkEnv, Success, TaskContext}
 import org.apache.spark.executor.ShuffleReadMetrics
 import org.apache.spark.executor.ShuffleWriteMetrics
+import org.apache.spark.memory.{TaskMemoryManager, TestMemoryManager}
 import org.apache.spark.scheduler.SparkListener
 import org.apache.spark.scheduler.SparkListenerJobStart
 import org.apache.spark.scheduler.SparkListenerTaskEnd
+import org.apache.spark.shuffle.comet.CometShuffleMemoryAllocator
+import org.apache.spark.shuffle.sort.CometShuffleExternalSorter
 import org.apache.spark.sql.CometTestBase
+import org.apache.spark.sql.catalyst.expressions.UnsafeRow
 import org.apache.spark.sql.comet.execution.shuffle.CometNativeShuffle
 import org.apache.spark.sql.comet.execution.shuffle.CometShuffleExchangeExec
 import org.apache.spark.sql.execution.SparkPlan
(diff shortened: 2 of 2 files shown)


sunchao (approved):
## Summary

Reviewed the complete two-file diff at head `901455dab14f16ff62146e25a56d5e6f17b1abcf` against base `bd7dc601aa9a99289530598d4778c40e6ded7772`. Five independent read-only scopes covered correctness and edge cases, compatibility and semantics, tests and CI, performance/concurrency/resource safety, and adversarial cross-checking with discussion deduplication. No outstanding, factually verified P1/P2 findings remain; approving this focused accounting fix.

## Prior state and problem

The regular spill path freed the active sorter's data pages and reset its pointer array without first preserving that working set in `peakMemoryUsedBytes`. A later getter or final cleanup could therefore observe only a smaller subsequent working set or the reset array, losing the larger earlier peak. This addresses finding 12 of #5212; it does not claim to resolve the other memory-accounting issues in that epic.

## Design approach

Reuse the existing `updatePeakMemoryUsed()` helper once each successful nonempty spill has finished writing, immediately before `freeMemory()` and `reset()`. The helper preserves the maximum of the existing peak and current writer-local page-plus-pointer-array usage, so earlier spills remain represented after their buffers are released.

`SpillSorter.writeSortedFileNative()` retains the JVM data pages and pointer array through its return, making the new sampling point early enough to capture both. Explicit spills, record-threshold spills, and allocation-pressure spills reach this same path. The existing close/cleanup sampling and `CometUnsafeShuffleWriter` propagation into task peak-execution-memory reporting remain intact.

The added call uses the existing synchronized sorter getter without changing managed-buffer allocation, serialization, exception handling, or lock ordering. The test's off-heap configuration creates a fresh unified allocator, so its independent `getUsed()` readings cannot be contaminated by the shared bounded allocator or accidentally update the sorter's peak.

Direct [CI logs for run 32865394451](https://github.com/apache/datafusion-comet/actions/runs/32865394451) show the new regression passing on Linux with Spark 3.4, 3.5, 4.0, 4.1, and 4.2, and on macOS with Spark 4.0. All six relevant jobs completed successfully. Their merge commit `da1ac50ea43eaf8cc2a1580981032b78f3bba25a` has the exact pinned head/base as parents and the same tree as the reviewed head. The complete check set is 63 successful and 9 skipped checks. No local JVM/native test or benchmark was rerun for this review; the author's reported pre-fix failure and A/B timings were not independently reproduced.

## Key design decisions

- Keep accounting writer-local, using actual page sizes plus pointer-array capacity. This avoids relying on allocator-wide totals whose scope and behavior differ between allocator implementations.
- Use a deliberately larger first spill and avoid calling the peak getter until both spills and close have completed. The getter itself updates the peak, so polling it earlier would mask the original defect.
- Hold the initial pointer array at 128 longs and use 256-byte pages with 21 then 5 records. Neither batch grows the pointer array, and both supported record-header widths produce three pages for the first spill and one for the second.

## Implementation sketch

The changed sequence is `writeSortedFileNative(false, ...) -> updatePeakMemoryUsed() -> freeMemory() -> reset()`, followed by the existing spill-byte metric update. The test derives each working set independently from allocator usage and checks its relationship to bytes freed during that spill. It then verifies the retained peak after close, two spill entries, 26 written records, matching file/partition/disk-byte totals, and no remaining task allocations after cleanup.

Reported JVM shuffle peak memory now retains a larger earlier spill instead of being reduced to what remains visible after that spill. In this fixture, the first and second working sets are 1,792 and 1,280 bytes respectively, while the reset pointer array alone is 1,024 bytes. The production change only records the existing working set; it leaves row output, spill thresholds, spill-file layout, spill-byte accounting, and buffer-release order unchanged.

## Suggested improvements

No P1/P2 change is needed for this PR. The additional sample traverses the current page list once per nonempty spill; exact O(1) page-byte accounting is already scoped separately in #5462. Keeping that optimization separate preserves the narrow correctness fix and lets the follow-up validate oversized/partial page sizes, lifecycle accounting, cross-thread visibility, and measured performance on its own.

sunchao:
Merged, thanks @peterxcli !

Issue #20, opened 2024-02-14
How to use/test Comet?


jinwenjie123:
Hi Team,

I am trying to evaluate the performance of using the comet plugin.
But I did not find any documentations about how to use the comet plugin after compiling it.
Specifically, how to use it in the cluster mode.

Thanks

viirya:
I think we will provide more documents on this.

Currently, https://github.com/apache/arrow-datafusion-comet/blob/main/bin/comet-spark-shell contains the simply usage of how to come out necessary configuration in Spark side to use Comet.

You just need to build Comet, distribute the jar with Spark and set up necessary configs to trigger it.

Btw, we haven't open source some features which are performance related. So the performance number you run might be not really accurate.

jinwenjie123:
Hi Team,

I also noticed that when comet detects unsupported features and fallback to Spark engine. Does it come with the cost of extra time to convert columnar data to row based data format ?
(Since I found some other Open Source projects which use native library, like Gluten, need to spend extra time to do this when the fallback happens.)

Thank you for your time.

viirya:
Yes, that's correct. At the boundary between Comet operator and Spark operator, we need ColumnarToRow operator to convert from ColumnarBatch to Spark InternalRow which takes some time cost. Overall we expect that the gain obtained from native operators can be bigger than these cost. And with more native operator support, we can reduce such fallback and the cost of columnar to row conversion.

jinwenjie123:
Hi Team,

I am wondering will we have a documentation about what kinda of data types are supported by Comet/Datafusion (Like, Decimal(16, 6), interger .... ) and will not fallback to vanilla spark and cause the regression.

Or where I can lookup to find related information. Since this is very important while we are trying to evaluate whether to use Comet or not.

Thank you so much !

viirya:
Hello, I answered in #64.

comphead:
Can it be closed via https://github.com/apache/arrow-datafusion-comet/pull/125? @viirya

viirya:
Okay

viirya:
@jinwenjie123 Feel free to open new issues on Comet usage if you still have other questions. Thanks.

Issue #19, opened 2024-02-14
Planning to publish Roadmap?


okue:
Hi, I would like to inquire about a roadmap for Comet.

I am currently exploring the option of replacing Spark workloads with native engines, and I'm keen to understand the future plans and scope of this plugin. Any information on the roadmap would be greatly appreciated.

Thank you!

sunchao:
Thanks @okue . Yes, will add a roadmap into doc soon. We used to have it internally.

alamb:
Maybe anyone else who is interested in using / contributing to comet could use this ticket to explain their use case / any features they are interested in helping with

milenkovicm:
I'd like to put a suggestion, based on my experience a lot of production spark workloads have some kind of UDF, good chunk of those UDFs are very simple and can be expressed as SQL expressions. 

I assume that in case UDF is used, comet will fall back to classic spark execution, which might not be optimal (I might be wrong, apologise if I am, I did not do my homework to check comet code in depth). My suggestion is to consider adding functionality like https://nvidia.github.io/spark-rapids/docs/additional-functionality/udf-to-catalyst-expressions.html which can speed up UDF in comet case as well.

I believe there is nothing GPU specific in that code, and it can be reused, just not sure what would be the best approach. 

Maybe @andygrove would be able to help

dbtsai:
Five years ago, we were delving into bytecode analysis for similar purposes at Apple, as detailed in this video: https://www.youtube.com/watch?v=FWg8iF34RJw. Our implementation differed from Nvidia's, and we encountered many edge cases that didn't work. Consequently, we halted further investment in this area. Nonetheless, we remain optimistic that we can refine the process to reliably handle simpler use-cases, prompting us to consider revisiting this approach.

milenkovicm:
> Five years ago, we were delving into bytecode analysis for similar purposes at Apple, as detailed in this video: https://www.youtube.com/watch?v=FWg8iF34RJw. Our implementation differed from Nvidia's, and we encountered many edge cases that didn't work. Consequently, we halted further investment in this area. Nonetheless, we remain optimistic that we can refine the process to reliably handle simpler use-cases, prompting us to consider revisiting this approach.

Thanks for sharing @dbtsai!

From my experience some projects mostly have "simple" udfs others may have mostly  complex functions. It'll be hard to get it right for all of them. 
Simple functions usually deal with data validation and they may be called all over the place, so speeding them up may make sense

[...]

PR #5464, opened 2026-08-25
fix: RAII for memory pool registration 

peterxcli:

Part of #5212 (finding #20 and the correctness portion of finding #19).

## Rationale for this change

Tracing registers each execution context's memory pool in `THREAD_MEMORY_POOLS`, but registration currently happens without a lifetime guard. A failure during later plan creation or during the metrics update in `releasePlan` skips unregistration and retains the pool indefinitely.

Debug tracing also registers each plan's distinct `LoggingMemoryPool` wrapper. Two contexts sharing one 4,096-byte task pool therefore report 8,192 bytes instead of the 4,096-byte ground truth.

On unmodified upstream main, a deterministic failure reproducer left 100 registry entries after plan-creation failures and 200 after an additional 100 metrics failures. Those entries reported 819,200 bytes and retained 200 extra references to the shared base pool.

- Register the task-shared base pool before constructing a per-plan logging wrapper.
- Store registration in a drop guard so every post-registration creation error removes its entry.
- Move the guard out before the fallible release metrics update, preserving the raw execution context for the existing JVM retry while cleaning the tracing registry on error.
- Add a focused regression covering two debug wrappers over one base pool, exact reserved-byte deduplication, repeated creation/metrics failures, and retained-reference cleanup.

The existing registry locking and scan algorithm are unchanged.

- `cargo test -p datafusion-comet --no-default-features execution::jni_api::tests::thread_memory_pool_registration_is_scoped_and_deduplicates_base_pool --profile ci -- --exact --nocapture --test-threads=1`
  - 1 passed.
- `cargo test -p datafusion-comet --no-default-features --lib --profile ci -- --test-threads=1 --skip parquet::objectstore::s3::tests::`
  - 126 passed.
- `cargo clippy -p datafusion-comet --no-default-features --lib --tests --profile ci -- -D warnings`
- `cargo fmt --check --all`
- `git diff --check`

An unfiltered library run passed 155 of 164 tests. Eight S3 credential tests failed because this host has no usable native root certificates; a representative failure reproduced with the unmodified-main binary. The remaining order-sensitive cache-refresh failure passed independently on both versions.

A warmed optimized A-B-B-A comparison used 31 interleaved samples per arm:

| Workload | Upstream main | This PR | Change |
| --- | ---: | ---: | ---: |
| Tiny, one context/pool | 95.4 ns/op | 98.9 ns/op | +3.6% |
| Two contexts sharing one pool | 110.5 ns/op | 109.0 ns/op | -1.3% |
| Registration lifecycle | 193.4 ns/op | 187.9 ns/op | -2.9% |

Per-arm IQRs overlapped for all three workloads.

### DIFF
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -148,6 +148,36 @@ fn register_memory_pool(thread_id: u64, context_id: i64, pool: Arc<dyn MemoryPoo
         .insert(context_id, pool);
 }
 
+struct ThreadMemoryPoolRegistration {
+    thread_id: u64,
+    context_id: i64,
+    registered: bool,
+}
+
+impl ThreadMemoryPoolRegistration {
+    fn new(thread_id: u64, context_id: i64, pool: Arc<dyn MemoryPool>) -> Self {
+        register_memory_pool(thread_id, context_id, pool);
+        Self {
+            thread_id,
+            context_id,
+            registered: true,
+        }
+    }
+
+    fn unregister_and_total(mut self) -> usize {
+        self.registered = false;
+        unregister_and_total(self.thread_id, self.context_id)
+    }
+}
+
+impl Drop for ThreadMemoryPoolRegistration {
+    fn drop(&mut self) {
+        if self.registered {
+            unregister_and_total(self.thread_id, self.context_id);
+        }
+    }
+}
+
 /// Unregister a context's pool and return the remaining total reserved for the thread.
 fn unregister_and_total(thread_id: u64, context_id: i64) -> usize {
     let mut map = get_thread_memory_pools().lock();
@@ -362,6 +392,8 @@ struct ExecutionContext {
     /// it has to travel with the plan. `None` when no driving Spark task is present (unit tests,
     /// direct native driver runs). Lifetime is as for `task_context` above.
     pub class_loader: Option<Arc<Global<JObject<'static>>>>,
+    /// Removes this context's tracing memory-pool entry on every exit path.
+    memory_pool_registration: Option<ThreadMemoryPoolRegistration>,
 }
 
 /// Accept serialized query plan and return the address of the native query plan.
@@ -445,6 +477,13 @@ pub unsafe extern "system" fn Java_org_apache_comet_Native_createPlan(
             let memory_pool =
                 create_memory_pool(&memory_pool_config, task_memory_manager, task_attempt_id);
 
+            // Register the shared base pool before wrapping it for per-plan debug logging. The
+            // guard removes the entry if any later plan setup step fails.
+            let rust_thread_id = get_thread_id();
+            let memory_pool_registration = tracing_enabled.then(|| {
+                ThreadMemoryPoolRegistration::new(rust_thread_id, id, Arc::clone(&memory_pool))
+            });
+
             let memory_pool = if logging_memory_pool {
                 Arc::new(LoggingMemoryPool::new(task_attempt_id as u64, memory_pool))
             } else {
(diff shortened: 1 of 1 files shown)


sunchao (approved):
## Summary

Reviewed the complete one-file diff at head `a34fc27b3d0a9ba622a07faa621325718acc9496` against base `bd7dc601aa9a99289530598d4778c40e6ded7772`. The change scopes tracing-registry ownership to a native execution context and makes shared-pool accounting independent of per-plan debug wrappers.

Five independent passes covered correctness and edge cases; compatibility and semantics; tests and CI; performance, concurrency, and resource safety; and adversarial cross-checking and discussion deduplication. After reconciliation and source validation, I found no outstanding P1/P2 issue in this scoped change and am approving it.

## Prior state and problem

`THREAD_MEMORY_POOLS` owns an `Arc` for each registered context, but previously registration and unregistration were separate calls without a lifetime guard. A failure after registration during context construction, or an error from the metrics update at the start of `releasePlan`, could bypass tracing-registry cleanup and retain that reference.

The registry already deduplicates pools by their allocation address. When memory debugging is enabled, however, registering each context's distinct `LoggingMemoryPool` wrapper gives the same task-shared base pool multiple identities. Its reservation can consequently be counted once per wrapper instead of once per base allocation.

## Design approach

`ThreadMemoryPoolRegistration` records the registering thread and context IDs and unregisters on drop. `createPlan` constructs this optional guard immediately after obtaining the base memory pool, before creating a logging wrapper, and transfers it into `ExecutionContext` only when construction succeeds.

`releasePlan` takes the guard out before the fallible metrics update. An error then drops the local guard, while the raw context remains alive for the JVM's existing opportunity to retry close. On a normal release, the consuming unregister method returns the remaining total for the existing tracing emission and disables a second unregister from `Drop`.

The runtime still receives the logging wrapper when configured, so consumer registration, allocation limits, growth, shrinking, and debug logging retain their existing behavior. Only the identity stored in the tracing registry changes. Task-shared and global pool factories return cloned base `Arc`s, while per-plan pools remain distinct, matching the registry's pointer-based deduplication.

The context and DataFusion reservations retain their own pool references, so removing a tracing reference does not free a pool still used by execution. The guard preserves the original thread ID for cleanup, and taking its `Option` prevents a later release retry from unregistering the same registration again. Tracing-disabled execution does not create a registration.

This change is scoped to the tracing registry. It does not repair the separate pre-existing task-shared `num_plans` rollback problem or guarantee disposal of a raw context after a persistent metrics failure; those should not be confused with this guard's cleanup guarantee.

## Key design decisions

- Register the shared base allocation, while retaining per-plan wrappers for actual execution and logging.
- Let Rust ownership cover fallible construction paths instead of adding cleanup beside each `?` return.
- Move registration ownership before metrics publication without prematurely freeing the raw context used by the existing JVM close path.
- Preserve the existing registry mutex, per-thread buckets, and deduplicating scan; this is not a redesign of the registry's locking or complexity.

## Implementation sketch

The patch adds the guard and one optional field to `ExecutionContext`, moves registration before wrapper creation, and replaces the manual release-time unregister with the guard's consuming method. The guard's ordinary destructor handles early returns, while the successful release path explicitly obtains the remaining reservation total.

The new regression test constructs two wrappers over one reserved base pool, checks registry entry counts and deduplicated bytes, repeats error-scope cleanup, and verifies that registry references do not keep the pool alive. These are direct guard tests rather than JNI fault-injection tests.

With tracing and memory-debug logging enabled, contexts sharing one base pool now contribute that pool's reservation once per thread. Failed post-registration construction and failed release metrics updates stop retaining a tracing-registry entry. Normal successful release continues to emit the remaining registered total.

I inspected the [native CI job](https://github.com/apache/datafusion-comet/actions/runs/32867771917/job/97869882161): 901 tests passed and 4 were skipped, including an explicit pass for the new regression. Its tested merge commit has the exact reviewed base and head as parents. Five additional local source-extracted lifecycle checks passed, covering explicit unregister, shared/distinct pools, error and unwind cleanup, release retry ownership, and cross-thread destruction; those checks use pool/mutex stand-ins and are not native/JNI integration runs.

CI is not entirely green: the [Spark 4.1 SQL shard](https://github.com/apache/datafusion-comet/actions/runs/32867771917/job/97883458742) failed before SQL tests started because downloading Maven 3.9.6 returned HTTP 429. This is an infrastructure failure, not a test failure attributable to this diff. No workflow was rerun as part of this review.

## Suggested improvements

No P1/P2 code changes are requested for this patch. The remaining CI action is to rerun the Spark 4.1 shard that failed downloading Maven; until that succeeds, the current checks should not be described as entirely green.

sunchao:
Merged, thanks @peterxcli !

Issue #5465, opened 2026-08-25
Make fair_unified account for spillable consumers


peterxcli:
### What is the problem the feature request solves?

`fair_unified` currently gives every registered memory consumer equal weight when it calculates a fair share. As a result, unspillable consumers reduce the share available to operators that can actually respond to memory pressure by spilling.

DataFusion's `FairSpillPool` handles the two kinds of consumers separately: it subtracts memory held by unspillable consumers, then divides the remainder among spillable consumers.

This difference appears in normal native sorts. `ExternalSorter` is spillable, while `ExternalSorterMerge` is not, but both currently increase Comet's divisor.

The focused fix for #5212 (finding #1) deliberately leaves this policy unchanged. It only fixes the reservation used in the current fair-share check. Spillability-aware accounting should be evaluated separately so that a policy change is not mixed into that root-cause fix.

### Describe the potential solution

Match DataFusion's accounting model:

- Track spillable and unspillable consumers and their reserved bytes separately.
- For a spillable request, divide the memory left after unspillable reservations among the spillable consumers.
- Admit unspillable requests only while aggregate reservations remain within the pool limit.
- Preserve the existing JNI acquisition, release, and partial-grant rollback behavior.
- Keep `reserved()` and `memory_limit()` truthful.

This needs the full mixed-consumer accounting, not just a different divisor. For example, with a 32 MiB pool and a 10 MiB merge reservation, giving the sorter a full 32 MiB share would admit 42 MiB.

A focused test should cover mixed spillable and unspillable reservations, and a native sort test should show that merge memory reduces the sorter's available share.

### Additional context

The per-reservation fix confirms that the low-concurrency path is reachable:

- 200k rows: 5 spills on main, 2 after the fix.
- 2m rows: 43 spills on main, 16 after the fix.
- 2m spilled bytes: 232,467,592 → 130,125,664.

The timing ranges overlapped, so this is spill and capacity evidence, not a speed claim. We have not run a spillable-only A/B yet, and the remaining spills do not prove that this policy change will help.

Fair-pool policy is performance-sensitive: #1369 reported TPC-H q3 slowing from about 16 seconds to 3.4 minutes. Before implementing this, we should run warmed repeated A/B on the same tiny and representative low-concurrency workloads and compare spill count, spilled bytes, capacity, and task tail. If that does not show a material improvement in a real workload, we should leave the current policy alone.

PR #4582, opened 2026-06-03
feat: prototype allocator-level OOM protection (OomGuard breaker + cooperative real-usage accounting)

andygrove:
## Which issue does this PR close?

Relates to #4576. This is an exploratory prototype covering both halves of that issue: the RSS circuit breaker ("OomGuard") and cooperative online accounting that feeds the real allocator balance into DataFusion's `MemoryPool`. It is not a complete, production-default implementation, so it does not close the issue.

## Rationale for this change

Comet's memory accounting relies on voluntary `MemoryPool` reservations, which miss allocations made by Arrow buffers, join scratch space, and expression kernels. Two problems follow:

1. When real native memory exceeds the container limit, the OS/YARN/Kubernetes kills the entire executor JVM, losing every task and all cached data on it.
2. Because the pool undercounts real usage, users compensate by manually lowering `spark.comet.exec.memoryPool.fraction`, a crude, workload-specific knob.

This prototype attacks both by tracking the real bytes the global allocator hands out and using that signal in two layers:

- A cooperative gate that rejects pool growth (triggering a DataFusion spill and retry) once real usage plus the request would exceed the off-heap budget. This lets the pool act on real usage rather than tracked reservations alone, so `fraction` no longer needs tuning.
- A last-resort, executor-global circuit breaker that fails a single task with a retriable `ResourcesExhausted` error instead of letting the executor get OOM-killed.

The byte-tracking allocator adapts the `AccountingAllocator` from apache/datafusion#22626 rather than depending on it, since that code lives in DataFusion's test-only `sqllogictest` crate.

## What changes are included in this PR?

Gated behind a new `oom-guard` cargo feature; the default build is unchanged with zero added per-allocation overhead.

Allocator accounting and circuit breaker:
- `native/core/src/execution/memory_pools/oom_guard.rs` (new): `AccountingAllocator<A>` wrapping the inner global allocator; a single process-wide balance with per-thread drift settled at a 64 KiB threshold; `arm`/`disarm`/`stamp_current_thread`/`current_balance`; a typed `OomGuardPanic`, raised via `panic_any` on an armed, stamped thread that crosses the limit, with reentrancy protection so the panic's own boxing allocation does not recurse.
- `native/core/src/lib.rs`: under the `oom-guard` feature, installs the wrapper as `#[global_allocator]` over jemalloc / mimalloc / system; mutually exclusive cfgs leave the default build untouched.

Cooperative real-usage gate:
[...]

PR #5466, opened 2026-08-25
fix: enforce fair memory share per reservation

peterxcli:

Addresses finding 1 and the fair-pool portion of finding 16 in #5212.

Follow-up #5465 tracks spillable versus unspillable consumer accounting separately.

## Rationale for this change

`CometFairMemoryPool` calculates a per-consumer fair share but compares it with pool-wide usage. With a 32 MiB pool and two consumers, one reservation holding 10 MiB and the requester holding 6 MiB make total usage equal the 16 MiB fair share. Main then rejects any positive growth from the requester, even though that requester has 10 MiB of fair-share headroom.

The fair pool also has a fixed configured capacity, so reporting an unknown memory limit is inaccurate.

- Compare the fair share with the requesting reservation's size plus the requested growth.
- Report `MemoryLimit::Finite(pool_size)` for `CometFairMemoryPool`.
- Add a focused regression test for the 10 MiB + 6 MiB reservation case and finite-limit reporting.
- Preserve the existing policy of counting all registered consumers. Restricting the count to spillable consumers is the separate policy and annotation change tracked by #5465.

Focused native test:

```shell
cargo test --release -p datafusion-comet fair_share_uses_requesting_reservation_and_reports_pool_limit --lib
```

Formatting and diff checks:

```shell
cargo fmt --all -- --check
git diff --check
```

A warmed local A/B used `local[1]`, a 32 MiB fair pool, one sort partition, three warmups, and seven measured runs:

| Workload | Spills | Native spilled bytes | Wall time, median (range) | Max task, median (range) |
| 200K rows | 5 → 2 | 13,014,520 → 13,013,008 | 92 (78–121) → 99 (87–115) ms | 51 (44–65) → 54 (51–76) ms |
| 2M rows | 43 → 16 | 232,467,592 → 130,125,664 | 406 (372–507) → 369 (326–632) ms | 371 (338–481) → 341 (300–603) ms |

Spill counts and bytes were identical across the seven runs for each build. Runtime ranges overlap, so this PR makes no speed claim; the demonstrated benefit is restoring the requesting consumer's fair-share capacity and avoiding premature spills.

### DIFF
--- a/native/core/src/execution/memory_pools/fair_pool.rs
+++ b/native/core/src/execution/memory_pools/fair_pool.rs
@@ -24,10 +24,9 @@ use jni::objects::{Global, JObject};
 
 use crate::{errors::CometResult, jvm_bridge::JVMClasses};
 use datafusion::common::resources_err;
-use datafusion::execution::memory_pool::MemoryConsumer;
 use datafusion::{
     common::DataFusionError,
-    execution::memory_pool::{MemoryPool, MemoryReservation},
+    execution::memory_pool::{MemoryConsumer, MemoryLimit, MemoryPool, MemoryReservation},
 };
 use parking_lot::Mutex;
 
@@ -42,6 +41,17 @@ pub struct CometFairMemoryPool {
 struct CometFairPoolState {
     used: usize,
     num: usize,
+}
+
+fn fair_limit_exceeded(
+    pool_size: usize,
+    num: usize,
+    reservation: &MemoryReservation,
+    additional: usize,
+) -> Option<(usize, usize)> {
+    let used = reservation.size();
+    let limit = pool_size.checked_div(num).expect("overflow in checked_div");
+    (limit < used.saturating_add(additional)).then_some((used, limit))
 }
 
 impl Debug for CometFairMemoryPool {
@@ -142,21 +152,15 @@ impl MemoryPool for CometFairMemoryPool {
 
     fn try_grow(
         &self,
-        _reservation: &MemoryReservation,
+        reservation: &MemoryReservation,
         additional: usize,
     ) -> Result<(), DataFusionError> {
         if additional > 0 {
             let mut state = self.state.lock();
             let num = state.num;
-            let limit = self
-                .pool_size
-                .checked_div(num)
-                .expect("overflow in checked_div");
-            // We use state.used instead of reservation.size() because DataFusion 53+
-            // calls pool.try_grow() before incrementing the reservation's atomic size,
-            // so reservation.size() would not include prior grows.
-            let used = state.used;
-            if limit < used + additional {
+            if let Some((used, limit)) =
+                fair_limit_exceeded(self.pool_size, num, reservation, additional)
+            {
                 return resources_err!(
                     "Failed to acquire {additional} bytes where {used} bytes already reserved and the fair limit is {limit} bytes, {num} registered"
                 );
@@ -186,5 +190,30 @@ impl MemoryPool for CometFairMemoryPool {
 
     fn reserved(&self) -> usize {
         self.state.lock().used
+    }
+
+    fn memory_limit(&self) -> MemoryLimit {
+        MemoryLimit::Finite(self.pool_size)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::execution::memory_pool::UnboundedMemoryPool;
+
+    #[test]
+    fn fair_share_uses_requesting_reservation_and_reports_pool_limit() {
+        let backing: Arc<dyn MemoryPool> = Arc::new(UnboundedMemoryPool::default());
+        let other = MemoryConsumer::new("other").register(&backing);
+        let requesting = MemoryConsumer::new("requesting").register(&backing);
+        other.grow(10);
+        requesting.grow(6);
+
+        assert_eq!(fair_limit_exceeded(32, 2, &requesting, 10), None);
+        assert_eq!(fair_limit_exceeded(32, 2, &requesting, 11), Some((6, 16)));
+
+        let pool = CometFairMemoryPool::new(Arc::new(Global::null()), 32);
+        assert!(matches!(pool.memory_limit(), MemoryLimit::Finite(32)));
     }
 }


native/core/src/execution/memory_pools/fair_pool.rs
```diff
@@ -142,21 +152,15 @@ impl MemoryPool for CometFairMemoryPool {
...
-                .pool_size
-                .checked_div(num)
-                .expect("overflow in checked_div");
-            // We use state.used instead of reservation.size() because DataFusion 53+
-            // calls pool.try_grow() before incrementing the reservation's atomic size,
-            // so reservation.size() would not include prior grows.
-            let used = state.used;
-            if limit < used + additional {
+            if let Some((used, limit)) =
+                fair_limit_exceeded(self.pool_size, num, reservation, additional)
```
sunchao, line 162: **[P2] Preserve the aggregate configured pool limit**

This replaces the only aggregate admission bound with a per-reservation check. With `spark.memory.offHeap.size=64m` and `spark.comet.exec.memoryPool.fraction=0.5`, `pool_size` is 32 MiB. A can register and reserve 24 MiB; after B registers, B's request for 16 MiB now passes (`0 + 16 <= 32 / 2`), leaving **40 MiB reserved against the configured 32 MiB budget**. The base rejects B's request. Similarly, `new_empty()` shares its consumer registration while starting a separate size counter, so sibling reservations can collectively exceed the budget; DataFusion's native sort merge uses this pattern.

Spark's JVM memory manager only applies its larger off-heap budget, not Comet's fraction, so it can grant these requests. This consumes the headroom that the [tuning guide](https://github.com/apache/datafusion-comet/blob/a658c1960ea2ae27cfa4584e77800163fab16645/docs/source/user-guide/latest/tuning.md#L48-L51) explicitly reserves to avoid OOM from untracked allocations. Please retain the corrected requester check and add a separate aggregate `state.used` versus `pool_size` bound; this does not require the spillability-policy change in #5465.

Both excess-allocation sequences were reproduced against unchanged head/base pool code in a harness with mocked JNI grants; Spark's larger backing limit was verified separately in source.

sunchao (commented):
## Summary

Reviewed the full single-file diff at head `a658c1960ea2ae27cfa4584e77800163fab16645` against base `bd7dc601aa9a99289530598d4778c40e6ded7772`. The reservation-based fair-share calculation fixes the reported premature rejection, but one **P2** remains: the change also removes enforcement of Comet's aggregate configured memory budget. That finding is reported inline.

## Prior state and problem

The previous implementation divided `pool_size` by the registered-consumer count, then compared that per-consumer allowance with `state.used`, which is the total across the pool. With a 32 MiB pool, two consumers holding 10 MiB and 6 MiB exhaust the old 16 MiB comparison even though the requester still has 10 MiB of its own allowance available. DataFusion updates a reservation's size after each successful grow, so subsequent sequential calls do include previous allocations; the removed comment incorrectly implied otherwise. The pool also inherited `MemoryLimit::Unknown` despite having a configured size.

## Design approach

The new helper compares the requesting reservation's current size plus the additional request against `pool_size / num`. Saturating addition avoids wraparound in that comparison, and `memory_limit()` now exposes the configured size through `MemoryLimit::Finite`. The existing policy of counting every registered consumer is retained, leaving the spillability policy in #5465 separate from this correction.

The sequential reservation calculation agrees with DataFusion 54.1's API ordering, and the finite-limit API is supported by the pinned dependency. Registration, aggregate bookkeeping, JNI acquisition/release, and partial-grant rollback are otherwise unchanged. However, individual reservation bounds do not imply an aggregate bound: registrations can arrive after earlier allocations, and `new_empty()`/`split()` create separate reservation sizes while sharing one consumer registration. Spark's backing memory manager does not enforce Comet's smaller `spark.comet.exec.memoryPool.fraction` budget, so the inline P2 needs an independent aggregate check.

Validation separates actual CI from the focused harness. The [Rust CI job](https://github.com/apache/datafusion-comet/actions/runs/32871988838/job/97882030555) tested the merge of these pins and passed 901 tests with 4 skipped, including the added regression; the current PR checks are 64 successful and 9 skipped. A separate Linux harness used the unchanged head/base pool implementations and checksum-verified DataFusion reservation source: all six checks passed in both debug and optimized modes, including assertions that the head accepts the excess reservations while the base rejects them. That harness mocked JNI grants/releases and supporting mutex/error adapters; it was not a native Spark integration test.

## Key design decisions

Keeping the fair-share calculation tied to the requester is the right correction for the original 10+6 example. Keeping `state.used` as the total remains necessary for bookkeeping and for enforcing the configured aggregate budget independently of fairness. The unchanged spillability policy and unchanged JNI rollback behavior keep this PR's intended scope narrow; neither needs redesign to address the P2.

## Implementation sketch

`fair_limit_exceeded` returns the requesting reservation's usage and its calculated allowance when growth would exceed that allowance. `try_grow` calls this helper while holding the existing state lock, then follows the existing JNI acquisition and accounting path. The new test exercises the helper with real DataFusion reservations backed by an unbounded pool, plus the finite-limit accessor on a Comet pool with a null JVM handle. It does not call the Comet pool's allocating `try_grow` path, so it does not cover aggregate admission or JNI behavior.

The requesting consumer can now use its remaining allowance even when another consumer already holds memory, which addresses the original premature-spill condition. Callers can also observe the configured finite limit instead of an unknown limit. Without the additional aggregate guard, some valid reservation sequences can consume the headroom deliberately reserved by a fraction below 1.0; the inline example reaches 40 MiB against a configured 32 MiB pool. The reported workload timings were not independently rerun, so this review makes no performance claim from them.

## Suggested improvements

Address the **P2** by retaining the new per-reservation comparison and separately rejecting growth that would push aggregate `state.used` above `pool_size`. Add a regression that exercises actual pool admission when a consumer registers after another has allocated, or when sibling reservations share a registration, while preserving the 10+6+10 success case. This can remain a narrow capacity fix without taking on the spillability-policy work in #5465.

comphead:
I'm currently experimenting with https://github.com/apache/datafusion-comet/pull/4582 
DF has proven mem accountability issues which tracked  https://github.com/apache/datafusion/issues/23393

native/core/src/execution/memory_pools/fair_pool.rs
```diff
@@ -142,21 +152,15 @@ impl MemoryPool for CometFairMemoryPool {
...
+            {
                 return resources_err!(
                     "Failed to acquire {additional} bytes where {used} bytes already reserved and the fair limit is {limit} bytes, {num} registered"
                 );
```
andygrove, line 166: I want to back up @sunchao's point about losing the aggregate bound, with two specifics that I think make it more concrete.

The first is that there is a documentation claim this breaks. `docs/source/user-guide/latest/tuning.md` lines 66 to 67 say the shared pool "ensures that the combined memory usage stays within the per-task limit". After this change nothing in the fair pool enforces that. `state.used` becomes pure bookkeeping and the only real ceiling left is Spark's `TaskMemoryManager`, which applies the full off-heap size and knows nothing about `spark.comet.exec.memoryPool.fraction`.

The second is that the sibling-reservation case is not hypothetical for the operators we actually run. `new_empty()` and `split()` show up in `sorts/sort.rs:644`, `sorts/streaming_merge.rs:248`, `sorts/stream.rs:187` and `:253`, and `aggregates/row_hash.rs:1368`. Each sibling shares one consumer registration but carries its own size counter, so each one independently gets a full `pool_size / num` allowance. A single sort can hold several multiples of its intended share.

In fairness, upstream `FairSpillPool` has the same late-registration overshoot property, so this is not a hazard we are inventing. But our pool takes a `pool_size` and after this change uses it only as a divisor. Could you keep the corrected per-reservation check and add `state.used + additional > pool_size` next to it? If you would rather not, then I think `tuning.md` lines 66 to 70 need updating in this PR, since the `pool_size / num_reservations` wording on line 70 also gets less accurate once the denominator is registrations and the numerator is per-reservation.

native/core/src/execution/memory_pools/fair_pool.rs
```diff
@@ -44,6 +43,17 @@ struct CometFairPoolState {
...
 }
 
+fn fair_limit_exceeded(
+    pool_size: usize,
+    num: usize,
+    reservation: &MemoryReservation,
+    additional: usize,
+) -> Option<(usize, usize)> {
+    let used = reservation.size();
+    let limit = pool_size.checked_div(num).expect("overflow in checked_div");
```
andygrove, line 53: Since this is moving into a helper anyway, could it use `.unwrap_or(pool_size)` the way `FairSpillPool` does with `.checked_div(state.num_spill).unwrap_or(spill_available)`? The current message describes the wrong failure, because `checked_div` returns `None` on division by zero rather than overflow, and a panic in here aborts the process instead of surfacing as a Spark error. The helper is now a free function that is directly callable with `num = 0`, so it is a bit easier to hit than it was before.

native/core/src/execution/memory_pools/fair_pool.rs
```diff
@@ -187,4 +191,29 @@ impl MemoryPool for CometFairMemoryPool {
...
+        let backing: Arc<dyn MemoryPool> = Arc::new(UnboundedMemoryPool::default());
+        let other = MemoryConsumer::new("other").register(&backing);
+        let requesting = MemoryConsumer::new("requesting").register(&backing);
+        other.grow(10);
+        requesting.grow(6);
+
+        assert_eq!(fair_limit_exceeded(32, 2, &requesting, 10), None);
+        assert_eq!(fair_limit_exceeded(32, 2, &requesting, 11), Some((6, 16)));
+
+        let pool = CometFairMemoryPool::new(Arc::new(Global::null()), 32);
```
andygrove, line 216: This test only reaches the free function with a hardcoded `num` of 2, so nothing exercises `try_grow` itself. If someone reverted `try_grow` back to passing `state.used`, this would still pass.

The `other` reservation also has no bearing on either assertion, since `fair_limit_exceeded` never looks at other consumers. The test would behave identically without it, so it does not really reproduce the 10 plus 6 case in its name.

Would it be possible to register both consumers on the `CometFairMemoryPool` so that `num` comes from the pool's own bookkeeping rather than a literal? I realize `try_grow` is awkward to reach because `acquire` needs a live JVM. If putting the JNI calls behind a small trait is more than you want to take on here, then a comment explaining why the test stops at the helper would at least make the gap visible to the next person reading it.

On the `memory_limit()` half, I checked and `Global::null()` is safe to drop without a JVM in jni 0.22.4 because `Drop` short-circuits on a null reference, so no concern there.

andygrove (commented):
Thanks for digging into this. I went and checked the removed comment against datafusion-execution 54.1.0 and you are right that it was wrong. `MemoryReservation::try_grow` calls `pool.try_grow(self, additional)` and then does `self.size.fetch_add(capacity)`, so `reservation.size()` is missing only the current call, not the prior grows. DataFusion's own `FairSpillPool::try_grow` compares `reservation.size() + additional > available`, which is exactly what you have here, so this moves us onto the upstream semantics.

I think the description undersells how bad the old behavior was. Comparing `state.used + additional` against `pool_size / num` capped the pool's total usage at `pool_size / num`, not just each consumer's share, so with 8 registered consumers a 32 MiB pool was limited to 4 MiB overall. Since `fair_unified` is the default off-heap pool, that seems worth spelling out, and it suggests the spill-count improvements in your A/B numbers are the floor rather than the ceiling of what this fixes.

One thing to be aware of on the testing side. No CI job here exercises off-heap mode. The Iceberg suites are the only ones that set `spark.memory.offHeap.enabled=true` and they are skipped on this PR, so the green checks are not telling us anything about `CometFairMemoryPool`. That puts a fair amount of weight on the unit test, which is part of why I left a comment on it.

I have also left a comment supporting @sunchao's point about the aggregate bound, with a couple of specifics that I think make the case stronger.

dwsmith1983:
@peterxcli Andy asked on #5613 that we agree an order between #5466 and #5613, since both touch `try_grow` in `fair_pool.rs`.

#5613 keeps the fair-limit check and the reservation as one step under the lock, then makes the JVM call with no lock held. #5466 changes what that check is measured against. The two compose: the per-reservation basis applies to the reserve-under-lock step rather than replacing it, and the two comments explaining the `state.used` basis get rewritten for the new basis rather than merged.

Either order works. If #5466 lands first I rebase #5613 onto it and carry the per-reservation basis into the locked step. If #5613 goes first, your change lands on a `try_grow` whose JVM call is already outside the lock, so the rebase is the admission check and its comments.

andygrove:
Thanks @peterxcli for finding this and working out how `MemoryReservation::try_grow` orders its size update. That analysis is what showed the removed comment was wrong. #5847 builds on your change. It keeps the per-consumer share, adds the pool-total cap from the review here, charges sibling reservations against their consumer's share, and reports the finite limit. So I'm closing this as superseded. The underlying bug is tracked in #5961.

PR #5469, opened 2026-08-26
fix: normalize scalar float sort and window rank keys

### LINKED ISSUE #5468: Scalar NaN and signed-zero sort keys split WindowGroupLimit rank peers
### Describe the bug

Native `WindowGroupLimitExec` can drop rows that Spark retains when a scalar `FLOAT` or `DOUBLE` ordering key contains different NaN representations or both zero signs. This is a follow-up to #4870.

Spark considers all NaNs equal and greater than non-NaN values, and treats `-0.0` and `+0.0` as equal. Arrow row encoding compares their raw total-order representations instead. Consequently, native `RANK` / `DENSE_RANK` cutoffs can split one Spark peer group. Native sorting also needs to agree: with a secondary ordering key, sorting by raw NaN payload or zero sign can make equal peers noncontiguous.

### Steps to reproduce

On Spark 3.5 or later, enable native execution and `spark.comet.exec.localTableScan.enabled=true`, keeping `spark.comet.exec.strictFloatingPoint=false`.

Build a local table with runtime NaNs, for example `java.lang.Double.longBitsToDouble(0x7ff8000000000001L)` and `java.lang.Double.longBitsToDouble(0xfff8000000000002L)`, plus finite values and a partition containing both zero signs. Avoid a Parquet round trip that can canonicalize NaN payloads. Filter a query of this shape to its top rank:

```sql
SELECT * FROM (
  SELECT *, RANK() OVER (PARTITION BY p ORDER BY v DESC NULLS LAST) AS rnk
  FROM floating_window_peers
) WHERE rnk <= 1
```

A regression fixture retaining eight rows in Spark returns only four in unpatched Comet. This reproduces for both `FLOAT` and `DOUBLE` with native Sort, Window, and WindowGroupLimit operators. Adding a secondary ordering key also exposes the sorting inconsistency.

### Expected behavior

Scalar floating-point sort keys, window peer comparisons, and native range-partition keys/boundaries should use consistent Spark-compatible normalization. Only comparison keys should change; returned values must preserve their original NaN payloads and zero signs.

### Additional context

The existing strict-floating-point fallback remains an opt-out. Floating-point values nested in arrays or structs are a separate compatibility limitation.

sunchao:

A filter such as `RANK() <= 1` must keep every row tied for first place. Native execution can currently drop some of those rows when the ordering column contains signed zeros or different NaN representations. This PR fixes that result mismatch for scalar `FLOAT` and `DOUBLE` keys, following the native `WindowGroupLimitExec` support added in #4870.

### A rank cutoff can discard a tied row

Consider a table `measurements` whose `DOUBLE` column `v` contains `-0.0`, `+0.0`, and `1.0`:

```sql
SELECT v, rnk
FROM (
  SELECT v, RANK() OVER (ORDER BY v ASC) AS rnk
  FROM measurements
) ranked
WHERE rnk <= 1;
```

Spark considers the two zeros equal, so both belong at rank 1. The native comparison instead distinguishes their IEEE-754 representations:

| Input value | Rank in Spark | Rank with the previous native comparison |
| --- | --- | --- |
| `-0.0` | 1 | 1 |
| `+0.0` | 1 | 2 |
| `1.0` | 3 | 3 |

The query therefore keeps **both zero rows in Spark, but only the negative-zero row before this fix**. `DENSE_RANK` has the same problem at the cutoff. Once `WindowGroupLimitExec` has discarded a qualifying row, the window calculation above it cannot recover that row.

NaNs expose the same mismatch. Spark treats every NaN representation as equal and greater than every non-NaN value. For input containing two differently encoded NaNs and `1.0`, `ORDER BY v DESC` followed by `rnk <= 1` should keep both NaNs. Arrow's raw total ordering can separate them; a NaN with its sign bit set can even sort below finite values. The regressions in this PR construct distinct NaN encodings directly so the test actually exercises this distinction.

The fix gives the native sorting and window operators a common representation for comparing scalar floating values. It reuses Comet's existing `NormalizeNaNAndZero` expression to map every NaN to one canonical NaN and both zero signs to positive zero **in the comparison keys**. Arrow can then compare those keys using its existing machinery while agreeing with Spark about which values are tied.

This normalization must happen before sorting as well as when assigning ranks. For example, with `ORDER BY v, secondary`, the rows `(-0.0, 1)` and `(+0.0, 1)` are peers and must precede the rows whose secondary key is `2`. A sort that distinguishes the zero signs can instead produce:

```text
(-0.0, 1), (-0.0, 2), (+0.0, 1), (+0.0, 2)
```

Even a corrected peer comparison would see the first peer group split apart. A streaming rank limit could reach the cutoff at `(-0.0, 2)` and stop before seeing the other qualifying row. The shared comparison-key construction therefore supplies normalized keys to Sort, Window, and WindowGroupLimit, keeping their ordering and peer decisions consistent.

That consistency includes `PARTITION BY`, not just `ORDER BY`. An ordinary query such as `RANK() OVER (PARTITION BY p ORDER BY id)` must still work when `p` is a `DOUBLE` column containing only `1.0` and `2.0`. Spark may already have normalized `p`; wrapping it again in Sort while leaving Window unchanged produces different expressions for the same key. DataFusion uses those expressions to recognize partition ordering, so the mismatch causes execution to fail even though the values compare equally. The planner therefore reuses already-normalized expressions and constructs matching keys for Sort, Window, and WindowGroupLimit. This preserves ordinary windows as well as the rank-limit path.

Native range partitioning uses the same rule for its sampled boundaries. An incoming positive zero and a boundary containing negative zero must compare equal, just as they do during sorting. Normalizing both sides keeps the shuffle's comparisons consistent with the operators that consume its output.

The original input values are preserved: selecting `v` still returns its original zero sign or NaN payload. Only the temporary keys used for comparison are normalized. The scope remains scalar `FLOAT` and `DOUBLE`; the documented limitations for floats nested in arrays or structs and the existing `spark.comet.exec.strictFloatingPoint` fallback policy are unchanged.

## How was this PR tested?

The correctness results in the following paragraphs and table were recorded during the earlier implementation and rebase. They are historical local validation, not a fresh full-suite rerun for this review follow-up. The newly run sort benchmark and checks are described after those results.

The regressions verify both the returned rows and the execution path. The Spark cutoff tests require native Sort and WindowGroupLimit operators, compare results with Spark, and check that the input and output retain their original floating bits. Native tests cover compound keys, peer groups spanning batches, and agreement between range boundaries and incoming keys. The previously ignored signed-zero SQL regression is also enabled.

Additional regressions protect ordinary windows partitioned by floating columns: `RANK`, `PERCENT_RANK`, and `NTILE` must match Spark for single and compound partitions. A filtered `RANK` query also checks the Sort → WindowGroupLimit → Window path. A native planner test verifies that both bare and already-normalized partition keys retain recognizable ordering metadata.

With the comparison-key fix removed, the native peer/range regressions and both Spark FLOAT/DOUBLE cutoff regressions fail. In the Spark fixture, unpatched Comet returns **4 rows where Spark returns 8**. The fixed runs pass.

On Spark 4.1, the partition-ordering cases were also checked against the library before the key-consistency fix: all six ordinary-window cases failed, while the two WindowGroupLimit controls passed. All eight pass with the fix.

The final review reran all three native planner regressions and the eight Spark 4.1 partition-key regressions; all passed. It also checked the three query shapes behind the Spark SQL CI failures on the earlier revision: a count window partitioned by floating columns, a rank window with computed aggregate/partition keys, and a correlated `EXISTS ... LIMIT 1` query that Spark rewrites into a window. Each query matched Spark with native Sort and Window execution asserted. The same probes passed with the pinned base library. These were focused query probes, not reruns of the full CI shards.

| Local validation | Result |
| --- | --- |
| Native workspace unit tests | 909 passed; 4 existing tests ignored |
| Spark 3.4: floating partition-key regressions | 6 passed; 2 WindowGroupLimit cases skipped because Spark 3.4 has no such operator |
| Spark 3.5: floating partition-key regressions | 8 passed |
| Spark 4.0: floating partition-key regressions | 8 passed |
| Spark 4.1: full window suite, all WindowGroupLimit SQL cases, sort fallback tests, and native range shuffle | 87 passed |

`make core`, `cargo fmt --all -- --check`, and `cargo clippy --all-targets --workspace -- -D warnings` passed. JVM formatting checks and Spark 4.1 Scala style checks also passed.

The review follow-up adds a reproducible [scalar floating-key sort benchmark](https://github.com/apache/datafusion-comet/blob/38825e26a1324297a46821a06fe5ce33e8b47779/native/spark-expr/benches/sort_float_keys.rs). It executes DataFusion 54.1.0 `SortExec` with a bare `Column` key and with Comet's `NormalizeNaNAndZero` key in the same optimized binary, representing the before/after key-construction change for previously bare keys. Already-normalized keys are reused by the planner. This measures execution and output draining for an in-memory, single-partition sort; it does not measure Spark planning, JNI, I/O, or a distributed range shuffle. Input generation and physical-plan construction are outside the timed region.

Both variants sort 262,144 rows in 32 batches of 8,192, with one current-thread Tokio runtime, ascending/nulls-last ordering, and no fetch limit. The finite shape has no nulls; the mixed shape has approximately 20% nulls, 10% payload/sign NaNs, 10% signed zeros, and 60% finite values. The release profile uses thin LTO and one codegen unit. Measurements used an AMD EPYC-Milan CPU pinned to one core, Rust 1.98.0, 30 Criterion samples per case, a one-second warmup, and a three-second measurement target. Other workspace builds and tests were paused. A second pass reversed the order of the two variants.

| Shape | Bare then normalized: median ms (change) | Normalized then bare: median ms (change) |
| --- | --- | --- |
| FLOAT finite | 23.617 → 23.711 (+0.4%) | 23.314 → 23.897 (+2.5%) |
| FLOAT mixed | 18.873 → 19.763 (+4.7%) | 18.811 → 18.487 (−1.7%) |
| DOUBLE finite | 27.064 → 30.155 (+11.4%) | 27.058 → 28.843 (+6.6%) |
| DOUBLE mixed | 21.977 → 24.723 (+12.5%) | 22.282 → 23.440 (+5.2%) |

Each cell reports bare → normalized even when execution order was reversed. FLOAT differences were small and the mixed-shape direction changed between passes. DOUBLE normalization had a measurable cost in both passes: +6.6–11.4% for finite values and +5.2–12.5% for mixed values. These are observed repeat ranges, not confidence intervals or end-to-end Spark performance claims. Normalization deliberately changes the comparison semantics for the mixed input. Folding it into sorting or encoding could avoid temporary arrays, but that optimization is not implemented by this correctness fix.

Untimed validation checks every row identity, original floating bits and nulls, ordering, and zero spills. Peak DataFusion pool reservations were identical for bare and normalized keys: 16 MiB / 16.0625 MiB for FLOAT finite/mixed and 18 MiB / 18.0625 MiB for DOUBLE finite/mixed. These are reserved-pool peaks, not process RSS or a complete allocation high-water mark: they do not include the temporary normalized values buffers. Each normalization evaluation allocates 4 bytes per row for FLOAT or 8 for DOUBLE (32/64 KiB for an 8,192-row array), sharing the validity bitmap. That is a per-evaluation size, not total allocation traffic; sorting and merging may evaluate the key more than once.

To reproduce from `native/`:

```sh
cargo bench --locked -p datafusion-comet-spark-expr --bench sort_float_keys -- --noplot
COMET_SORT_BENCH_REVERSE_VARIANTS=1 cargo bench --locked -p datafusion-comet-spark-expr --bench sort_float_keys -- --noplot
```

For this follow-up, the release benchmark build, all eight benchmark smoke cases, release Clippy for this benchmark with `-D warnings`, `cargo fmt --all -- --check`, and `git diff --check` passed. The broader correctness suites above were not rerun for this benchmark/documentation-only commit.

The guide now links [#5506](https://github.com/apache/datafusion-comet/issues/5506) for narrowing strict-mode admission of corrected scalar sort keys and [#5507](https://github.com/apache/datafusion-comet/issues/5507) for nested ORDER BY/rank compatibility. Spark 4.1.3 diagnostic queries reproduced nested array/struct ORDER BY and array RANK mismatches for both FLOAT and DOUBLE; struct RANK already fell back and matched Spark. All eight strict-mode controls matched Spark, and four scalar planning controls confirmed the conservative strict-mode fallback. Those tracker probes reused a prior native build with the unchanged nested path and matching JVM admission sources; they are not a full rebuild or correctness-suite run of this PR head.

Closes #5468.

### DIFF
--- a/docs/source/user-guide/latest/compatibility/floating-point.md
+++ b/docs/source/user-guide/latest/compatibility/floating-point.md
@@ -28,15 +28,18 @@ to Spark in some cases, especially when the data contains both positive and nega
 case that is not of concern for many users. If it is a concern, setting `spark.comet.exec.strictFloatingPoint=true`
 will make relevant operations fall back to Spark.
 
-## Ordering: signed zero (`-0.0` vs `+0.0`)
+## Ordering: NaN and signed zero (`-0.0` vs `+0.0`)
 
 Spark's `ORDER BY`, `RANK`, `DENSE_RANK`, and window frame comparisons route through
-`SQLOrderingUtil.compareDoubles` / `compareFloats`, which explicitly define `-0.0 == 0.0`. Comet's
-native sort and `WindowGroupLimitExec` use the `arrow-row` row-format encoder for `ORDER BY` keys,
-which applies Rust's total-ordering transform to the raw IEEE-754 bits. Under that encoding `-0.0`
-sorts strictly less than `+0.0`, so a partition that mixes the two zeros can produce a rank
-distribution that differs from Spark. For example, `RANK() OVER (ORDER BY v ASC)` over
-`[-0.0, 0.0, 1.0]` filtered to `rk <= 1` returns two rows in Spark (both zeros tied at rank 1) but
-one row in Comet (`-0.0` at rank 1, `+0.0` at rank 2). If your workload materially mixes `-0.0`
-and `+0.0` in a ranked column, prefer Spark for that stage or normalize the column to `+0.0`
-upstream.
+`SQLOrderingUtil.compareDoubles` / `compareFloats`, which equate all NaN representations and
+define `-0.0 == 0.0`. NaN sorts above every non-NaN value.
+
+For scalar `FLOAT` and `DOUBLE` keys, Comet normalizes NaNs and signed zeros before native
+sorting, window peer comparisons, and `WindowGroupLimitExec` rank comparisons. Native range
+partitioning normalizes its keys and sampled boundaries in the same way. Only comparison keys
+are normalized; returned values retain their original NaN representations and zero signs.
+
+Floating-point values nested in arrays or structs still use Arrow's raw total ordering and can
+produce different ordering or rank results from Spark. The existing
+`spark.comet.exec.strictFloatingPoint=true` fallback policy is unchanged, including its
+conservative fallback for scalar floating-point sort keys.
--- a/docs/source/user-guide/latest/compatibility/floating-point.md
+++ b/docs/source/user-guide/latest/compatibility/floating-point.md
@@ -39,7 +39,10 @@ sorting, window peer comparisons, and `WindowGroupLimitExec` rank comparisons. N
 partitioning normalizes its keys and sampled boundaries in the same way. Only comparison keys
 are normalized; returned values retain their original NaN representations and zero signs.
 
-Floating-point values nested in arrays or structs still use Arrow's raw total ordering and can
-produce different ordering or rank results from Spark. The existing
-`spark.comet.exec.strictFloatingPoint=true` fallback policy is unchanged, including its
-conservative fallback for scalar floating-point sort keys.
+Native sorting of floating-point values nested in arrays or structs still uses Arrow's raw total
+ordering. Nested keys can therefore produce different ordering or rank results from Spark; see
+[#5507](https://github.com/apache/datafusion-comet/issues/5507).
+
+The existing `spark.comet.exec.strictFloatingPoint=true` fallback policy is unchanged, including
+its conservative fallback for scalar floating-point sort keys. Narrowing that scalar-sort
+admission policy is tracked in [#5506](https://github.com/apache/datafusion-comet/issues/5506).
--- a/native/core/src/execution/planner.rs
+++ b/native/core/src/execution/planner.rs
@@ -949,7 +949,17 @@ impl PhysicalPlanner {
     ) -> Result<PhysicalSortExpr, ExecutionError> {
         match spark_expr.expr_struct.as_ref().unwrap() {
             ExprStruct::SortOrder(expr) => {
-                let child = self.create_expr(expr.child.as_ref().unwrap(), input_schema)?;
+                let child =
+                    self.create_expr(expr.child.as_ref().unwrap(), Arc::clone(&input_schema))?;
+                let data_type = child.data_type(input_schema.as_ref())?;
+                // Spark treats every NaN as a peer and equates signed zeros. Normalize only
+                // comparison keys, preserving the original values in the output. Sort, Window,
+                // and WindowGroupLimit must agree so peers remain contiguous for compound keys.
+                let child = if matches!(data_type, DataType::Float32 | DataType::Float64) {
+                    Arc::new(NormalizeNaNAndZero::new(data_type, child)) as Arc<dyn PhysicalExpr>
+                } else {
+                    child
+                };
                 let descending = expr.direction == 1;
                 let nulls_first = expr.null_ordering == 0;
 
@@ -3441,10 +3451,14 @@ impl PhysicalPlanner {
                     }
                 }
 
-                // Convert the collection of ScalarValues to collection of Arrow Arrays
+                // Normalize boundary arrays just like the incoming sort keys, so equal NaNs
+                // and signed zeros are assigned to the same range partition.
                 let arrays: Vec<ArrayRef> = scalar_values
                     .iter()
-                    .map(|scalar_vec| ScalarValue::iter_to_array(scalar_vec.iter().cloned()))
+                    .map(|scalar_vec| {
+                        ScalarValue::iter_to_array(scalar_vec.iter().cloned())
+                            .map(|array| NormalizeNaNAndZero::normalize_array(&array))
+                    })
                     .collect::<Result<Vec<_>, _>>()?;
 
                 // Create a RowConverter and use to create OwnedRows from the Arrays
@@ -4815,13 +4829,15 @@ mod tests {
     };
 
     use arrow::array::{
-        Array, ArrayRef, DictionaryArray, Int32Array, Int8Array, ListArray, RecordBatch,
-        StringArray,
+        Array, ArrayRef, DictionaryArray, Float32Array, Float64Array, Int32Array, Int8Array,
+        ListArray, RecordBatch, StringArray,
     };
     use arrow::datatypes::{DataType, Field, FieldRef, Fields, Schema, SchemaRef};
     use datafusion::catalog::memory::DataSourceExec;
+    use datafusion::common::ScalarValue;
     use datafusion::config::TableParquetOptions;
     use datafusion::datasource::listing::PartitionedFile;
+    use datafusion::datasource::memory::MemorySourceConfig;
     use datafusion::datasource::object_store::ObjectStoreUrl;
     use datafusion::datasource::physical_plan::{
         FileGroup, FileScanConfigBuilder, FileSource, ParquetSource,
@@ -4830,8 +4846,10 @@ mod tests {
     use datafusion::logical_expr::AggregateUDF;
     use datafusion::logical_expr::ScalarUDF;
     use datafusion::physical_expr::expressions::Column;
-    use datafusion::physical_expr::PhysicalExpr;
+    use datafusion::physical_expr::{LexOrdering, PhysicalExpr};
+    use datafusion::physical_plan::sorts::sort::SortExec;
     use datafusion::physical_plan::ExecutionPlan;
+    use datafusion::prelude::SessionConfig;
     use datafusion::{assert_batches_eq, physical_plan::common::collect, prelude::SessionContext};
     use datafusion_physical_expr_adapter::PhysicalExprAdapterFactory;
     use datafusion_spark::function::aggregate::collect::{SparkCollectList, SparkCollectSet};
@@ -4843,9 +4861,10 @@ mod tests {
     };
     use crate::jvm_bridge::{JavaShufflePartitionPusher, ShufflePartitionPusher};
 
-    use crate::execution::operators::ExecutionError;
+    use crate::execution::operators::{ExecutionError, PartitionedRankLimitExec, WindowFnKind};
     use crate::execution::planner::literal_to_array_ref;
     use crate::execution::planner::parse_file_scan_tasks_from_common;
+    use crate::execution::shuffle::CometPartitioning;
     use crate::parquet::parquet_support::SparkParquetOptions;
     use crate::parquet::schema_adapter::SparkPhysicalExprAdapterFactory;
     use datafusion_comet_proto::spark_expression::expr::ExprStruct;
@@ -4856,6 +4875,9 @@ mod tests {
         spark_expression::{self, literal},
         spark_operator,
         spark_operator::{operator::OpStruct, Operator},
+        spark_partitioning::{
+            partitioning::PartitioningStruct, BoundaryRow, Partitioning, RangePartition,
+        },
     };
     use datafusion_comet_spark_expr::EvalMode;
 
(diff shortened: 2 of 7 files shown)


andygrove:

This is a thorough piece of work. Reusing `NormalizeNaNAndZero` rather than inventing a new comparator is the right call, and I like that `create_normalized_key_expr` is shared by sort, window partition keys, window group limit, and range partitioning so the four cannot drift apart. The `floating_window_partition_keys_preserve_ordering` test that checks `get_ordered_partition_by_indices` still returns the prefix is a good catch, since silently losing the sorted window mode would have been an easy regression to miss. Flipping the `ignore(...)` in `window_group_limit_rank.sql` back to a real assertion is nice to see.

A few things I would like resolved before merge.

**No performance data for the sort path**

`create_sort_expr` now wraps every scalar `FLOAT` and `DOUBLE` sort key, which means `native/core/src/execution/planner.rs:1498` adds a full column copy per batch to every native sort on a floating key, plus the same for window ordering and range partitioning. `unary` allocates a fresh values buffer each time. That is probably fine relative to the row encoding and the sort itself, but "probably fine" is not a number. Could you run a sort benchmark with a float key before and after and put the result in the description? If it turns out to be measurable, it would be worth discussing whether the normalization can be folded into the row encoder instead of materializing an intermediate array.

**`strictFloatingPoint` now falls back for cases this PR fixes**

The docs change says the `spark.comet.exec.strictFloatingPoint=true` fallback policy is unchanged, "including its conservative fallback for scalar floating-point sort keys." After this PR those scalar keys agree with Spark, so users who turn strict mode on are paying a fallback for something that is now correct. Is narrowing that fallback to the nested array and struct cases in scope here, or should it be a follow-up? If it is a follow-up I would like an issue filed and linked from this PR, because a stale conservative fallback is the kind of thing that quietly stays for a year.

**Tracking issue for the nested case**

Same question for the nested array and struct divergence the docs now call out. There is no issue link in the new `floating-point.md` text. A user who reads that paragraph and is affected has nowhere to go. Could you file one and link it from the doc?

**Sort-merge join builds an expression it throws away**

At `planner.rs:2145` the `create_sort_expr` result is used only for `sort_expr.options`, and the expression itself is discarded. That call now constructs a `NormalizeNaNAndZero` node for nothing on float keys. It is not a correctness problem, but it does suggest that site wants a smaller helper that only decodes direction and null ordering. Worth splitting out?

**Silent passthrough in `normalize_array`**

`NormalizeNaNAndZero::normalize_array` returns the input untouched for any type that is not `Float32` or `Float64`, including `Dictionary(_, Float64)`, `List<Float64>`, and structs with float fields. In the range-partitioning call site that is consistent with `create_normalized_key_expr`, which also only handles the two scalar types, so I do not think there is a bug here today. But the function is `pub` and the passthrough is silent. Could the doc comment say explicitly that nested and dictionary-encoded floats are deliberately not normalized, and that callers must match `create_normalized_key_expr`'s coverage? Otherwise the next caller gets a silently wrong answer with no compile error.

**One test question**

In `CometWindowExecSuite`, the new tests set `COMET_EXEC_STRICT_FLOATING_POINT` to `false` explicitly, which is already the default. Is that there to document intent, or to guard against the default flipping? If the former, a short comment saying so would help, since otherwise it reads like the tests only pass in non-strict mode.

comphead (commented):
Thanks @sunchao this PR looks good to me from manual review, let me do 1 pass of LLM review as well

sunchao:
Updated in [38825e26a](https://github.com/apache/datafusion-comet/commit/38825e26a1324297a46821a06fe5ce33e8b47779).

The follow-up adds the sort benchmark and links the two remaining compatibility issues from the guide. I checked the six points separately:

1. **Sort cost:** [the new benchmark](https://github.com/apache/datafusion-comet/blob/38825e26a1324297a46821a06fe5ce33e8b47779/native/spark-expr/benches/sort_float_keys.rs) runs DataFusion 54.1.0 `SortExec` with bare versus normalized keys in the same release binary, using 262,144 rows, 32 batches, one partition/thread, and both finite/non-null and null/NaN/signed-zero inputs. Two passes reverse the variant order. FLOAT median changes were small (−1.7% to +4.7%, with mixed-input direction varying); DOUBLE cost was measurable in both passes: +6.6–11.4% finite and +5.2–12.5% mixed. The description contains both passes and reproduction commands. This is a SortExec benchmark, not an end-to-end Spark result. Peak pool reservations were unchanged, but they exclude temporary normalized arrays: each evaluation adds a 4/8-byte-per-row values buffer and shares validity. Already-normalized keys are reused, so this measures previously bare keys rather than another copy at every call. Folding normalization into sorting/encoding is a worthwhile optimization to investigate separately; this change does not claim the cost is free.

2. **Strict-mode admission:** tracked in [#5506](https://github.com/apache/datafusion-comet/issues/5506), now linked from the guide. The follow-up is scoped to scalar Sort admission with strict-mode execution tests; this PR retains the existing guards for other consumers and nested keys.

3. **Nested ordering/rank:** tracked in [#5507](https://github.com/apache/datafusion-comet/issues/5507), also linked from the guide. Diagnostic SQL reproduced array/struct ORDER BY and array RANK differences for FLOAT and DOUBLE with native operators asserted. Struct RANK already falls back in those probes; it is not a reproduced native wrong result. Strict-mode controls matched Spark. The issue records the reused-native-build limitation, so these are evidence for the remaining limitation, not a fresh full-head validation claim.

4. **Sort-merge-join options:** the observation is correct, but that site constructs and discards a planning-time expression. It does not evaluate the normalizer or copy a column there; it already built an expression just to obtain options before this PR. I left the optional helper cleanup out of this correctness fix.

5. **`normalize_array` coverage:** the existing doc comment specifies scalar floating comparison keys and says other types are unchanged. Its Float32/Float64 match and `create_normalized_key_expr` use the same gate: dictionary, list, and struct arrays deliberately pass through. There is no recursive or dictionary-normalization promise in that API; the nested limitation now has the explicit tracker above.

6. **Explicit `strictFloatingPoint=false` in tests:** it pins the native execution path under test, independently of externally supplied configuration or a future default change. The tests assert native Sort/WindowGroupLimit operators; `true` deliberately still invokes the conservative fallback tracked by #5506. I kept the setting for that reason.

The new release benchmark built, all eight smoke cases passed, and benchmark Clippy (`-D warnings`), formatting, and diff checks passed. I also corrected the stale validation paragraph: the earlier full-suite results are historical, not rerun for this benchmark/documentation-only follow-up.

comphead:
One thing to consider is https://issues.apache.org/jira/browse/SPARK-54918 which fixes +0, -0 in Spark 4.2.0

sunchao:
@comphead Thanks for the pointer. I checked [the SPARK-54918 patch](https://github.com/apache/spark/commit/a4aa4adf4fd0f790d7b71b322292050f6cf8db38). It extends normalization to array set operations such as `array_distinct` and `array_union`.

For the scalar sorting and ranking cases here, Spark already treats `-0.0` and `+0.0` as equal before 4.2. The problem is that Comet’s native comparison can distinguish them, splitting rows that should share a rank and potentially dropping qualifying rows at a rank cutoff. 

This PR normalizes the comparison keys consistently across sorting and ranking, so the fix is still needed with Spark 4.2.

comphead (approved):
Thanks @sunchao LGTM

PR #5470, opened 2026-08-26
fix: preserve aggregate result identity during exchange reuse

sunchao:

No issue is automatically closed. This is a standalone correctness fix for aggregate exchange reuse.

## Rationale for this change

Spark can reuse a shuffle when another branch of the query has already computed the same data. That decision depends on the identity of the plan below the exchange. Comet currently compares an aggregate's grouping keys and aggregate functions without comparing its final result expressions. Two aggregates can therefore look equivalent even when they return different values.

For example, suppose a Parquet table `t(k, v)` contains `(0, 2)` and `(0, 3)`. This excerpt shows the query shape covered by the regression, including a shuffle above each final aggregate:

```scala
import org.apache.spark.sql.functions.col

def branch(result: String) =
  spark.sql(s"SELECT $result AS c, k FROM t GROUP BY k")
    .repartition(2, col("c"), col("k"))

val plus = branch("COUNT(*) + 1")
val minus = branch("COUNT(*) - 1")
plus.unionAll(minus).collect()
```

The expected `(c, k)` rows are `(3, 0)` and `(1, 0)`. Both branches count the same rows, but their final arithmetic differs. If Spark incorrectly reuses the first shuffle for the second, the query can finish successfully with `(3, 0)` twice.

The fix must also preserve legitimate reuse. Repeating `COUNT(*) + 1` under a different column alias still computes the same values, even though Spark assigns fresh expression IDs. Treating those plans as different would avoid incorrect sharing at the cost of unnecessary computation and shuffling.

Aggregate identity now includes the final result expressions, so equality and hashing distinguish computations such as `COUNT(*) + 1` and `COUNT(*) - 1` before Spark decides whether to share an exchange. The native aggregate already evaluates these projections; this change makes the plan's identity describe the result it actually produces.

To retain valid reuse, Comet also preserves Spark's original aggregate result attributes and uses them when normalizing plan identity. This is especially important for distinct aggregates, whose rewritten expressions do not always retain the original result IDs. Equivalent computations can still compare equal across fresh IDs and aliases, while different projections remain separate. This applies to both hash and object hash aggregate conversion. The scope is the aggregate implementation and its existing test suite; native execution kernels and dependencies are unchanged.

Five regressions in `CometAggregateSuite` cover count, distinct count, distinct count combined with sum, and collect-set size. COUNT runs with AQE both disabled and enabled; the other three cases run with AQE disabled. Each case checks that different final projections return the correct results and that equivalent projections with different aliases still reuse an exchange above a native final aggregate. The AQE case also requires an adaptive plan and traverses query-stage wrappers when checking reuse. All cases use Parquet input, native Comet shuffle, and exchange reuse enabled.

At commit `49e821a24`, the five regressions passed again on Spark 3.4.3 / Scala 2.12.17 / JDK 17, together with the full Scalafix CHECK using the `semanticdb` profile, packaging, Scalastyle, and Spotless checks. This commit only removes a redundant string-interpolation prefix from the test name. Spark 4.1 was not rerun for this lint-only cleanup.

Previously, at commit `2ea7ed82c`, all five regressions passed locally on each of Spark 3.4.3 / Scala 2.12.17 / JDK 17 and Spark 4.1.3 / Scala 2.13.17 / JDK 21: **10 tests passed, zero failures**. Reactor compilation, Scalastyle, and Spotless checks passed on both profiles. Native sources are unchanged; these runs used a verified Apache CI native library built from the exact upstream base `fefee03d`, rather than a fresh local native build.

For historical context, the [earlier macOS Spark 4.0 execution job](https://github.com/apache/datafusion-comet/actions/runs/32933515258/job/98075229919) validated commit `b448894b` before the rebase. That result is not validation of the current head. Hosted CI for `49e821a24` is pending after this update; the local results above do not imply full CI has passed.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/comet/operators.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/operators.scala
@@ -30,7 +30,7 @@ import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.internal.Logging
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, AttributeSet, Expression, ExpressionSet, Generator, NamedExpression, SortOrder}
+import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, AttributeSeq, AttributeSet, Expression, ExpressionSet, Generator, NamedExpression, SortOrder}
 import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, AggregateMode, CollectList, CollectSet, Final, Partial, PartialMerge, Percentile}
 import org.apache.spark.sql.catalyst.optimizer.{BuildLeft, BuildRight, BuildSide}
 import org.apache.spark.sql.catalyst.plans._
@@ -1974,6 +1974,7 @@ object CometHashAggregateExec
       op.output,
       op.groupingExpressions,
       op.aggregateExpressions,
+      op.aggregateAttributes,
       op.resultExpressions,
       op.child.output,
       op.child,
@@ -2024,6 +2025,7 @@ object CometObjectHashAggregateExec
       adjustOutputForNativeState(op),
       op.groupingExpressions,
       op.aggregateExpressions,
+      op.aggregateAttributes,
       op.resultExpressions,
       op.child.output,
       op.child,
@@ -2037,6 +2039,7 @@ case class CometHashAggregateExec(
     override val output: Seq[Attribute],
     groupingExpressions: Seq[NamedExpression],
     aggregateExpressions: Seq[AggregateExpression],
+    aggregateAttributes: Seq[Attribute],
     resultExpressions: Seq[NamedExpression],
     input: Seq[Attribute],
     child: SparkPlan,
@@ -2049,7 +2052,15 @@ case class CometHashAggregateExec(
   // modes is empty too.
   val modes: Seq[AggregateMode] = aggregateExpressions.map(_.mode).distinct
 
-  override def producedAttributes: AttributeSet = outputSet ++ AttributeSet(resultExpressions)
+  // Match Spark's aggregate canonicalization, including the original result attributes that
+  // rewritten DISTINCT aggregate expressions do not necessarily retain in their resultIds.
+  override lazy val allAttributes: AttributeSeq =
+    child.output ++ aggregateExpressions.flatMap(_.aggregateFunction.aggBufferAttributes) ++
+      aggregateAttributes ++
+      aggregateExpressions.flatMap(_.aggregateFunction.inputAggBufferAttributes)
+
+  override def producedAttributes: AttributeSet =
+    outputSet ++ AttributeSet(resultExpressions) ++ AttributeSet(aggregateAttributes)
 
   override protected def withNewChildInternal(newChild: SparkPlan): SparkPlan =
     this.copy(child = newChild)
@@ -2072,6 +2083,8 @@ case class CometHashAggregateExec(
         this.output == other.output &&
         this.groupingExpressions == other.groupingExpressions &&
         this.aggregateExpressions == other.aggregateExpressions &&
+        this.aggregateAttributes == other.aggregateAttributes &&
+        this.resultExpressions == other.resultExpressions &&
         this.input == other.input &&
         this.modes == other.modes &&
         this.child == other.child &&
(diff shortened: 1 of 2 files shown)


sunchao:
cc @andygrove @viirya @comphead

viirya (commented):
The fix makes sense to me. Since the native aggregate already evaluates `resultExpressions`, those expressions need to participate in plan identity; otherwise, different final projections can incorrectly share an exchange. Preserving the original `aggregateAttributes` also addresses normalization across fresh expression IDs, including DISTINCT rewrites. I like that the regressions check both correctness and preservation of legitimate reuse.

A few non-blocking suggestions:

- Could we use `override def allAttributes` instead of `lazy val`? This matches newer Spark versions and avoids retaining the attribute lookup structures for the lifetime of the plan.
- Could we add an AQE-enabled regression, at least for COUNT, to cover adaptive query-stage reuse? The current tests explicitly disable AQE.
- Please update the validation section: the cited CI run covers `b448894b`, while the current head is `73effbeac`. The patch content is unchanged, but the wording should distinguish historical validation from CI on the current head.

I did not find a blocking correctness issue in the changes.

sunchao:
Thanks @viirya. Addressed the test and validation feedback in `2ea7ed82c`:

- Added one AQE-enabled COUNT case to the existing regressions. It checks different final projections against Spark and explicit expected rows, requires an adaptive plan, and checks that equivalent aliases reuse an exchange above the native Final aggregate. The plan checks now traverse adaptive/query-stage wrappers.
- Refreshed the validation section to distinguish the current commit's local results from historical CI on `b448894b`. All five committed regressions pass on Spark 3.4.3 / JDK 17 and Spark 4.1.3 / JDK 21, for 10 passing tests. Hosted CI for the new commit is pending.

I kept `override lazy val allAttributes` in the shared implementation for Spark 3 compatibility. Both [Spark 3.4.3](https://github.com/apache/spark/blob/v3.4.3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/QueryPlan.scala#L614) and [Spark 3.5.9](https://github.com/apache/spark/blob/v3.5.9/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/QueryPlan.scala#L579) declare the inherited member as a `lazy val`; Scala rejects overriding it with a `def` (confirmed with Scala 2.12 and 2.13 compiler probes). I agree with the memory-retention rationale in newer Spark. Adopting `def` only on Spark 4 would require a version-specific implementation, so I left that outside this narrow correctness fix.

viirya (approved):
Thanks for addressing the feedback. The AQE test now checks reuse above the native Final aggregate and traverses adaptive/query-stage wrappers, covering the gap from the previous revision.

You’re also right about `allAttributes`. I verified that Spark 3.4/3.5 declares it as a `lazy val`, and confirmed with Scala 2.12 and 2.13 that overriding it with `def` fails. I missed that compatibility constraint in my earlier suggestion; keeping `lazy val` is appropriate here.

The updated validation section clearly distinguishes local results from historical CI. No further concerns from me; LGTM pending CI.

sunchao:
Merged, thanks @viirya !

andygrove:
Apologies for not reviewing before this PR was merged, but it was only marked as ready for review yesterday.

I went through it after the fact and the fix itself looks right to me. I checked out the merge commit and reverted `operators.scala` back to its parent while keeping the new tests, and exactly the five new tests fail, all of them on wrong answers rather than on plan assertions, so the regressions really are targeting the bug. I also tried removing just the `allAttributes` override while keeping the equals change. The same five fail, but this time on `finalAggregate(plus).sameResult(finalAggregate(same))` rather than on results, which confirms both halves are pulling their weight. The equals change buys correctness and the `allAttributes` override buys back the legitimate reuse. That override is also a character for character match of `HashAggregateExec.allAttributes` in Spark 4.1, so no concerns there either.

What did come out of the review is that the same defect is still live in four other operators, and I have filed #5824 for it. `CometHashJoinExec`, `CometBroadcastHashJoinExec` and `CometSortMergeJoinExec` all carry `joinType` as a constructor field and all leave it out of `equals` and `hashCode`. `CometBroadcastNestedLoopJoinExec` does include it, which makes me think the other three were simply missed. `LeftSemi` and `LeftAnti` have identical output, keys, condition and build side, so without `joinType` they canonicalize to the same plan. Usually the semi join's inferred `isnotnull(key)` makes the left subtrees differ and hides it, but writing the null check explicitly in both branches is enough to remove that protection. At that point an `EXISTS` / `NOT EXISTS` union returns the `EXISTS` rows twice under Comet and the correct rows under Spark, and it reproduces with all three join strategies.

`CometExplodeExec` has the same shape of problem. It writes `op.outer` into the protobuf but never stores it on the case class, so `equals` cannot see it, and `output` does not disambiguate because plain `explode` over an array with `containsNull = true` is already nullable. `InferFiltersFromGenerate` hides this one, but only when the generator input is a bare attribute, so `explode(s.arr)` against `explode_outer(s.arr)` collides on stock configuration.

Since you already have the context loaded, would you be up for picking up #5824 as a follow-up? The fixes look like about a line each. The other thing I would like to get out of it is a guard so we stop rediscovering this. A test that reflects over every `CometNativeExec` subclass and asserts that each constructor parameter is either referenced in `equals` or named on an explicit exclusion list would have caught all three of these, including this one, before they shipped. Does that seem worth doing to you?

PR #5471, opened 2026-08-26
fix: ignore structural tags when lifting expression coverage

sunchao:

No issue is automatically closed. This is a follow-up to the structural-expression tag filtering associated with #5229, which is already closed.

## Rationale for this change

Comet's extended explain reports how much of a query runs natively or through the JVM codegen dispatcher. Its expression-coverage counts come from metadata recording the expressions used by the plan. Stale names in that metadata can inflate coverage counts or add misleading dispatch labels, making it harder to understand what Comet actually accelerated.

Spark's rewrites can copy a tagged expression's metadata onto the process-wide `Literal.TrueLiteral` singleton. Later queries in that JVM share the same literal. Existing filtering ignores tags found directly on literals, but decimal promotion leaves another path for them to escape.

For example, suppose an earlier query left `greaterthanorequal` coverage on the shared literal. Consider this illustrative projection, where `amount` is `DECIMAL(10, 2)`:

```sql
SELECT named_struct('flag', true, 'sum', amount + amount) AS value
FROM payments;
```

There is no comparison in this projection. However, decimal promotion rebuilds its expression tree to add overflow checking. The current coverage lift collects tags from that rebuilt tree, including the stale literal tag, and copies them onto the projection's original output alias. An alias can legitimately hold coverage collected from its child expressions, so subsequent filtering accepts the copied name. The query's native coverage now includes `greaterthanorequal` even though the query does not contain it. Codegen-dispatch coverage can be contaminated in the same way.

This makes diagnostics depend on earlier queries in the JVM and can produce misleading coverage counts or unstable explain output. It does not corrupt query results.

The coverage lift now applies the existing structural-node filtering before copying native and codegen-dispatch names back to the original expression. Stale tags on literals and other nodes that cannot legitimately own coverage are discarded before they can be attached to an alias.

Genuine coverage still needs to survive the rewrite. In particular, decimal promotion introduces a `CheckOverflow` expression that is absent from the original tree; its coverage must still reach the original owner. The change preserves that behavior and keeps aliases as valid recipients. It changes expression metadata and the existing regression test, without changing query results, execution routing, or fallback decisions.

The expanded `CometCodegenSuite` regression seeds both coverage-tag categories on the shared literal, checks that decimal promotion excludes them while retaining real `checkoverflow` coverage, and checks an unrelated plan. The seeded tags deliberately stand in for earlier-query contamination; this is not a claim that the test reproduces the entire preceding query history.

At head `37004235`, the [Spark 4.0 / JDK 21 expression job](https://github.com/apache/datafusion-comet/actions/runs/32933543592/job/98075971394) passed this regression and the existing expression-coverage tests. Local JVM validation had been blocked by dependency resolution; the passing runtime evidence comes from CI.

CI on `37004235` is now green, with 63 passing checks and 9 skips. The [Spark 3.5 shuffle rerun](https://github.com/apache/datafusion-comet/actions/runs/32933543592/job/98261490884) and the Iceberg 1.11 [Spark](https://github.com/apache/datafusion-comet/actions/runs/32933543592/job/98261485434), [extensions](https://github.com/apache/datafusion-comet/actions/runs/32933543592/job/98261487058), and [runtime](https://github.com/apache/datafusion-comet/actions/runs/32933543592/job/98261486905) jobs passed. The review follow-up also compiled the full Spark 4.1.3 JVM reactor locally and passed five coverage/singleton tests; disabling the coverage lift made the new positive dispatcher regression fail. These local tests reused a previously built OSS native library and did not rebuild native code. CI for the review follow-up commit is not yet confirmed.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
@@ -1032,6 +1032,11 @@ object QueryPlanSerde extends Logging with CometExprShim with CometTypeShim {
    * Nodes that carry no computation of their own. They are excluded from the expression coverage
    * stats in extended explain because they appear in nearly every expression tree and would swamp
    * the names a user actually cares about.
+   *
+   * `CometExplainInfo.isNeverTagged` must be this set minus `Alias`: the read-side filter retains
+   * aliases because [[liftCoverageTags]] uses them to hold names from rewritten children. Keep
+   * both sets in sync. This coverage invariant does not exclude structural nodes from carrying
+   * `FALLBACK_REASONS`.
    */
   private def isStructuralExpr(expr: Expression): Boolean = expr match {
     case _: Attribute | _: BoundReference | _: Literal | _: Alias => true
--- a/spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
@@ -879,14 +879,10 @@ object QueryPlanSerde extends Logging with CometExprShim with CometTypeShim {
   }
 
   private def liftCoverageTags(from: Expression, to: Expression): Unit = {
-    val native = mutable.Set.empty[String]
-    val dispatched = mutable.Set.empty[String]
-    from.foreach { e =>
-      e.getTagValue(CometExplainInfo.NATIVE_EXPRS).foreach(native ++= _)
-      e.getTagValue(CometExplainInfo.CODEGEN_DISPATCH_EXPRS).foreach(dispatched ++= _)
+    val exprs = from.collect { case e: Expression => e }
+    Seq(CometExplainInfo.NATIVE_EXPRS, CometExplainInfo.CODEGEN_DISPATCH_EXPRS).foreach { tag =>
+      appendTagValues(to, tag, CometExplainInfo.collectExprTagValues(exprs, tag))
     }
-    appendTagValues(to, CometExplainInfo.NATIVE_EXPRS, native.toSet)
-    appendTagValues(to, CometExplainInfo.CODEGEN_DISPATCH_EXPRS, dispatched.toSet)
   }
 
   /**
--- a/spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
@@ -25,7 +25,7 @@ import org.apache.arrow.vector._
 import org.apache.spark.{SparkConf, SparkEnv, TaskContext}
 import org.apache.spark.sql.CometTestBase
 import org.apache.spark.sql.api.java.UDF1
-import org.apache.spark.sql.catalyst.expressions.{AttributeReference, BoundReference, CreateArray, CreateMap, CreateNamedStruct, Expression, Literal, MapConcat}
+import org.apache.spark.sql.catalyst.expressions.{Add, Alias, AttributeReference, BoundReference, CreateArray, CreateMap, CreateNamedStruct, Expression, Literal, MapConcat}
 import org.apache.spark.sql.catalyst.expressions.objects.Invoke
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.internal.SQLConf
@@ -358,7 +358,20 @@ class CometCodegenSuite
     val planted = Literal.TrueLiteral
     planted.setTagValue(CometExplainInfo.EXTENSION_INFO, Set("PLANTED_INFO"))
     planted.setTagValue(CometExplainInfo.NATIVE_EXPRS, Set("plantedexpr"))
+    planted.setTagValue(CometExplainInfo.CODEGEN_DISPATCH_EXPRS, Set("planteddispatch"))
     try {
+      // Decimal promotion rebuilds this projection. Its coverage lift must not copy the
+      // singleton's stale tags onto the Alias, which is a legitimate coverage owner.
+      val decimal = AttributeReference("amount", DecimalType(10, 2), nullable = false)()
+      val projection = Alias(
+        CreateNamedStruct(Seq(Literal("flag"), planted, Literal("sum"), Add(decimal, decimal))),
+        "value")()
+      assert(QueryPlanSerde.exprToProto(projection, Seq(decimal)).isDefined)
+      val native = projection.getTagValue(CometExplainInfo.NATIVE_EXPRS).getOrElse(Set.empty)
+      assert(native.contains("checkoverflow"), s"expected lifted decimal coverage, got: $native")
+      assert(!native.contains("plantedexpr"))
+      assert(projection.getTagValue(CometExplainInfo.CODEGEN_DISPATCH_EXPRS).isEmpty)
+
       withSQLConf(
         CometConf.COMET_EXTENDED_EXPLAIN_FORMAT.key ->
           CometConf.COMET_EXTENDED_EXPLAIN_FORMAT_VERBOSE,
@@ -381,6 +394,7 @@ class CometCodegenSuite
 
           val info = new ExtendedExplainInfo()
           assert(!info.getNativeExpressions(plan).contains("plantedexpr"))
+          assert(!info.getCodegenDispatchExpressions(plan).contains("planteddispatch"))
           val explain = info.generateExtendedInfo(plan)
           assert(!explain.contains("PLANTED_INFO"), s"tag leaked into:\n$explain")
         }
@@ -388,6 +402,7 @@ class CometCodegenSuite
     } finally {
       planted.unsetTagValue(CometExplainInfo.EXTENSION_INFO)
       planted.unsetTagValue(CometExplainInfo.NATIVE_EXPRS)
+      planted.unsetTagValue(CometExplainInfo.CODEGEN_DISPATCH_EXPRS)
     }
   }
 
--- a/spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
@@ -25,7 +25,7 @@ import org.apache.arrow.vector._
 import org.apache.spark.{SparkConf, SparkEnv, TaskContext}
 import org.apache.spark.sql.CometTestBase
 import org.apache.spark.sql.api.java.UDF1
-import org.apache.spark.sql.catalyst.expressions.{Add, Alias, AttributeReference, BoundReference, CreateArray, CreateMap, CreateNamedStruct, Expression, Literal, MapConcat}
+import org.apache.spark.sql.catalyst.expressions.{Add, Alias, AttributeReference, BoundReference, Cast, CreateArray, CreateMap, CreateNamedStruct, Expression, Hypot, Literal, MapConcat}
 import org.apache.spark.sql.catalyst.expressions.objects.Invoke
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.internal.SQLConf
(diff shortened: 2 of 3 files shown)


spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
```diff
@@ -855,14 +855,10 @@ object QueryPlanSerde extends Logging with CometExprShim with CometTypeShim {
   }
 
   private def liftCoverageTags(from: Expression, to: Expression): Unit = {
```
andygrove, line 881: What makes this filter safe is that `isStructuralExpr` at line 1002 never lets the serde tag an `Attribute`, `BoundReference`, or `Literal` in the first place, so `CometExplainInfo.isNeverTagged` can only ever discard copied tags. That invariant is now load-bearing for this fix, but the two lists live in different files under different names and neither comment mentions the other. Could `isNeverTagged` be derived from `isStructuralExpr` minus `Alias`, or failing that, could each comment name the other and state the subset relationship? Otherwise someone adding a node type to just one of them either reopens this path or quietly deletes real coverage.

spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
```diff
@@ -278,7 +279,20 @@ class CometCodegenSuite
     val planted = Literal.TrueLiteral
     planted.setTagValue(CometExplainInfo.EXTENSION_INFO, Set("PLANTED_INFO"))
     planted.setTagValue(CometExplainInfo.NATIVE_EXPRS, Set("plantedexpr"))
+    planted.setTagValue(CometExplainInfo.CODEGEN_DISPATCH_EXPRS, Set("planteddispatch"))
     try {
+      // Decimal promotion rebuilds this projection. Its coverage lift must not copy the
```
andygrove, line 379: Would it be better to split this out into its own test? It asserts something quite different from the rest of the test it is in, which is about the dynamic-pruning plan not reporting the planted name, and the test name and comment only describe that second half. A separate test named for the lift, something like "the coverage lift ignores stale tags on the shared TrueLiteral", would point straight at the mechanism when it fails, and it would sit next to "expression coverage stats survive the decimal promotion rewrite", which is the positive case for the same code path.

Related question. Is there a way to drive this through a real plan the way the neighbouring tests do, or does planting on the singleton force the direct `exprToProto` call? I ask because this is the only place in these suites that builds a Catalyst tree by hand, so if a plan-level version is workable it would be more consistent and would cover the roll-up and the explain rendering too.

andygrove (commented):
Thanks for chasing this down. Routing the lift through `collectExprTagValues` looks like the right place to me, because once a stale name has landed on an `Alias` the roll-up has no way to tell it apart from a real one.

I checked the write side and I agree the filter is lossless. `isStructuralExpr` at `QueryPlanSerde.scala:1002` already stops the serde from tagging an `Attribute`, `BoundReference`, or `Literal`, so anything this filter discards can only have arrived by `copyTagsFrom`. `CheckOverflow` is not structural, so the coverage the lift exists for still survives.

One question outside the diff. Does `liftFallbackReasons` just below have the same exposure? It walks every node in the rebuilt tree the way `liftCoverageTags` used to, and `CometExecRule.rollUpFallbackReasons` does not filter either, so a stale reason branded onto `Literal.TrueLiteral` would surface on an unrelated operator. The branding path looks like the one the test comment already describes, where `PlanDynamicPruningFilters` swaps a `DynamicPruningSubquery` for the singleton, and a subquery expression is very likely to be carrying a Comet fallback reason at that point. I do not think it belongs in this PR, partly because the fix cannot be symmetric there. Line 993 does legitimately tag literals with fallback reasons, so `isNeverTagged` would be the wrong filter. It also seems worse than a coverage count, since a spurious reason on the operator is enough to satisfy `reportUnexplainedFallback` and would hide a genuinely missing reason under `COMET_STRICT_FALLBACK_REASONS`. Could you file an issue and link it here so it does not get lost?

CI is green now, including the Spark 3.5 shuffle job and both Iceberg 1.11 jobs, so the last paragraph of the description is stale and worth refreshing.

spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
```diff
@@ -278,7 +279,20 @@ class CometCodegenSuite
...
+      // singleton's stale tags onto the Alias, which is a legitimate coverage owner.
+      val decimal = AttributeReference("amount", DecimalType(10, 2), nullable = false)()
+      val projection = Alias(
+        CreateNamedStruct(Seq(Literal("flag"), planted, Literal("sum"), Add(decimal, decimal))),
+        "value")()
+      assert(QueryPlanSerde.exprToProto(projection, Seq(decimal)).isDefined)
+      val native = projection.getTagValue(CometExplainInfo.NATIVE_EXPRS).getOrElse(Set.empty)
+      assert(native.contains("checkoverflow"), s"expected lifted decimal coverage, got: $native")
+      assert(!native.contains("plantedexpr"))
+      assert(projection.getTagValue(CometExplainInfo.CODEGEN_DISPATCH_EXPRS).isEmpty)
```
andygrove, line 389: This change routes `CODEGEN_DISPATCH_EXPRS` through the structural filter for the first time, and both of the new assertions for it are negative. Would it be worth adding a positive case, where an expression inside a promoted decimal tree really is routed through the JVM codegen dispatcher, checking that its name still reaches the original owner? Without one, a future change to `isNeverTagged` could drop genuine dispatch coverage across a decimal rewrite and these tests would still pass.

sunchao:
Updated in [95c744c4c](https://github.com/apache/datafusion-comet/commit/95c744c4c63af1d8a778245bd34b080d12c0904b).

Addressed the coverage follow-up in 95c744c4c63a: the two structural predicates now document their relationship, and a positive dispatcher regression proves a name created only on the promoted copy reaches the original alias. The full Spark 4.1.3 JVM reactor compiled; all five selected coverage/singleton tests passed. Removing only the lift made the new regression fail. Local execution reused a previously built OSS native library; no native changes are in this follow-up.

The fallback-reason provenance follow-up is tracked in https://github.com/apache/datafusion-comet/issues/5499. It covers stale-reason propagation and possible masking of the strict check while preserving legitimate literal fallback reasons. I kept the proposed DPP contamination origin explicitly unproven: Spark wraps `TrueLiteral`, and the rule ordering alone does not reproduce singleton branding.

I also refreshed the stale final CI paragraph. The earlier head's reruns are green; the new commit's CI remains to be confirmed.

spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala
sunchao: Updated in 95c744c4c63af1d8a778245bd34b080d12c0904b. Added reciprocal comments on `isStructuralExpr` and `isNeverTagged`: the read filter is the write-side structural set minus `Alias`, because rewritten-child coverage is lifted onto the original alias. I also narrowed the wording to coverage/info tags and explicitly excluded `FALLBACK_REASONS` from that invariant; literals can legitimately carry those reasons.

spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
sunchao: Updated in 95c744c4c63af1d8a778245bd34b080d12c0904b. I kept the seeded-singleton probe with its plan-level check so the two phases share one setup and guaranteed cleanup. The direct call pins the exact `Literal.TrueLiteral` identity and the original `Alias`; normal SQL optimization can change that shape. The neighboring decimal test already covers plan-level coverage, and the second half of this test covers DPP roll-up and extended explain. There are also existing direct Catalyst probes in this suite for `BoundReference` and map/array expressions.

spark/src/test/scala/org/apache/comet/CometCodegenSuite.scala
sunchao: Updated in 95c744c4c63af1d8a778245bd34b080d12c0904b. Added `codegen dispatch coverage survives the decimal promotion rewrite`. It places `Hypot` above a cast of decimal addition, so promotion rebuilds both `Hypot` and its alias. The test checks the emitted `JvmScalarUdf` dispatcher class, verifies that the original `Hypot` has neither the dispatch marker nor its coverage tag, and requires the original alias to receive `hypot`. This passed locally on Spark 4.1.3. As a mutation check, removing only `liftCoverageTags(newExpr, expr)` made this new assertion fail with `None did not contain Set("hypot")`.

sunchao:
**Update, August 27 at 19:20 UTC:** The retry completed successfully on the unchanged PR head `95c744c4c`. [Workflow attempt 2](https://github.com/apache/datafusion-comet/actions/runs/33095231283/attempts/2) passed.

The original [Spark 4.2 / JDK 17 shuffle job](https://github.com/apache/datafusion-comet/actions/runs/33095231283/job/98607452205) failed before the Java modules compiled or tests ran: Maven Central returned HTTP 429 while resolving `org.junit:junit-bom:5.10.1` through the RAT plugin dependencies. It was a dependency setup failure, not a shuffle-test failure.

GitHub initially rejected the retry while the parent workflow was running. Once that workflow completed, the rerun was accepted at 19:05 UTC and subsequently passed. No code change was needed for this failure.

andygrove (approved):
The reciprocal comments on `isStructuralExpr` and `CometExplainInfo.isNeverTagged` are what I was asking for, the new `codegen dispatch coverage survives the decimal promotion rewrite` test covers the positive dispatcher case that was missing, and thanks for filing #5499 for the fallback-reason side.

One correction to my earlier comment. I said `isStructuralExpr` is what stops the serde from tagging a structural node, but that guard only covers the `withNativeExpr` call in `exprToProtoInternal`. `CometScalaUDF.emitJvmCodegenDispatch` calls `withCodegenDispatchExpr` at `spark/src/main/scala/org/apache/comet/serde/CometScalaUDF.scala:174` with no such check. I read through the registrations and it is safe today, because `CometLiteral`, `CometAttributeReference` and `CometAlias` are plain `CometExpressionSerde` and neither `CometCodegenDispatch` nor `CodegenDispatchFallback`, so no structural node can reach that line. That is a different fact from the one the new comment states though, and this PR is what makes `CODEGEN_DISPATCH_EXPRS` depend on it. Would it be worth naming that second condition alongside the existing one, so that someone registering a dispatch serde knows it is load-bearing?

The rest reads correct to me. The lift now uses the same flatten-then-filter shape as `CometExecRule.rollUpInfoMessages`, `FALLBACK_REASONS` is left alone and called out as such in both comments, and none of the TPC-DS `extended.txt` coverage counts moved, which is a reasonable signal that nothing genuine is being discarded. CI is fully green.

andygrove:
@sunchao this is conflicting with main now. Could you rebase? CI was green before the conflict, so it should be ready to go after that.

andygrove:
Rebase checked. Against `95c744c4c`, the head I approved, the only differences are two import lines that moved with `main`, so the change itself is unaltered and my approval stands.

On the question I left open last time, about `CometScalaUDF.emitJvmCodegenDispatch` calling `withCodegenDispatchExpr` without the `!isStructuralExpr` guard that `exprToProtoInternal` applies to `withNativeExpr`: rereading it with the new comments, that condition is not load-bearing any more, and this PR is what makes it not load-bearing. `collectExprTagValues` filters on the *node* through `isNeverTagged`, so a coverage tag that reached a `Literal`, `Attribute` or `BoundReference` by any route, including a direct write at the dispatch site, is dropped on read rather than depending on nobody writing it. That is a better answer than the comment I asked for, and the two comments you added say the right thing about which set has to track which. Worth noting explicitly that `FALLBACK_REASONS` is excluded from the invariant in both places.

One coordination note, since it is my own PR and therefore my problem rather than yours. #5610 also touches expression coverage: it records the names of a dispatched subtree's descendants on the dispatched root's `CODEGEN_DISPATCH_EXPRS` tag. I checked the interaction and it is clean in both merge orders. Your filter operates on nodes and mine writes names onto a non-structural root, and mine already excludes attribute references and literals from the names it records, so neither one can undo the other. I will rebase #5610 behind this rather than the other way round.

sunchao:
Thanks for rechecking the rebase and the interaction with #5610. Agreed: the read-side filter addresses the earlier concern, while `FALLBACK_REASONS` remains separate. Landing #5471 first and rebasing #5610 afterward sounds good.

sunchao:
Merged, thanks @andygrove for the detailed review!

Issue #5352, opened 2026-08-13
Support Apache Celeborn remote shuffle service


pingzh:
### What is the problem the feature request solves?

Remote shuffle services such as Apache Celeborn are widely used in cloud-native environments. Comet does not currently support Celeborn as a remote shuffle service.

I plan to work on this.

### Describe the potential solution

The proposed implementation consists of approximately eight small PRs:

1. **Add an RSS partition writer and JNI callback.** Introduce a backend-neutral native partition writer and task-owned JVM callback for sending encoded Arrow IPC shuffle frames.

2. **Add partition-writer destinations to shuffle plans.** Extend the native shuffle plan to represent local versus remote partition writers while maintaining compatibility with existing shuffle plans.

3. **Add destination-aware native shuffle execution.** Allow native shuffle execution to write either to existing local shuffle files or to a remote partition writer without changing existing local behavior.

4. **Bind task-owned JVM pushers to native shuffle planning.** Connect each Spark map task’s JVM-side pusher to native shuffle execution through a task-scoped JNI handle.

5. **Add a Celeborn shuffle manager and partition pusher.** Introduce the Celeborn-specific adapter and a composite shuffle manager that delegates ordinary Spark shuffles to the existing Celeborn manager.

6. **Complete the Celeborn map-side push lifecycle.** Add bounded frame handling, backpressure, asynchronous push completion, cancellation, retries, speculative execution, commit coordination, and support for nested Arrow batches.

7. **Add a native-aware Celeborn shuffle reader.** Read remote Celeborn shuffle partitions through either Spark-compatible columnar reads or direct Comet native shuffle scans, with appropriate metrics and failure handling.

8. **Enable native-only Celeborn shuffle planning.** Enable the remote path only for explicitly configured, supported native shuffle plans; preserve fallback for unsupported plans, JVM shuffle, and Spark I/O encryption; add user-facing configuration and documentation.

```mermaid
flowchart TD
    Manager["Comet + Celeborn shuffle manager"]
    Fallback["Existing Spark / Celeborn shuffle"]

    Manager -->|"Supported native exchange"| Native
    Manager -->|"Ordinary or unsupported exchange"| Fallback

    subgraph Map["Map-side executor"]
        direction LR

        Native["Comet native operators"]
        Encode["Native partitioning + Arrow IPC"]
        Destination{"Partition writer"}
        Local["Existing local shuffle writer"]
        RSS["Backend-neutral RSS writer"]
[...]

PR #5473, opened 2026-08-26
feat: add RSS partition writer and task-owned JNI callback (1/n)

pingzh:

Part of #5352. This is the first foundational PR and does not close the issue.

## Rationale for this change

Comet’s native shuffle currently writes partition data to local files. Supporting remote shuffle services requires a transport-independent mechanism for delivering complete encoded shuffle blocks to task-owned JVM callbacks.

This PR introduces that foundation without changing shuffle planning, adding Celeborn integration, or modifying existing shuffle behavior.

- Add a Java `ShufflePartitionPusher` callback interface for task-owned shuffle output.
- Add a Rust callback abstraction and JNI adapter that:
  - Safely invokes callbacks from native execution threads.
  - Preserves Java exceptions across the JNI boundary.
  - Validates partition identifiers and payload lengths.
- Add `RssPartitionWriter`, which:
  - Encodes complete, length-prefixed Arrow IPC shuffle blocks.
  - Supports existing shuffle compression codecs and dictionary-encoded batches.
  - Enforces configurable frame limits and partition lifecycle ordering.
  - Preserves shuffle metrics and propagates callback failures.
- Expose the writer for integration in subsequent PRs.

- 15 new Rust tests cover frame boundaries, partition routing, all compression codecs, dictionary batches, error propagation, frame limits, and lifecycle validation.
- Three new JNI tests cover partition and payload validation.
- All 50 native shuffle unit tests pass.
- All 28 existing JVM native-shuffle regression tests pass.
- Full-workspace Rust Clippy, Rust formatting, ScalaStyle, Spotless, CI preflight, and the complete four-module Maven build all pass.

### DIFF
--- a/native/jni-bridge/src/shuffle_partition_pusher.rs
+++ b/native/jni-bridge/src/shuffle_partition_pusher.rs
@@ -1,0 +1,147 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::{check_exception, errors::CometError, JVMClasses};
+use datafusion::common::{DataFusionError, Result};
+use jni::objects::{Global, JMethodID, JObject, JValue};
+use jni::signature::{Primitive, ReturnType};
+use jni::Env;
+
+/// Receives a complete encoded shuffle block for one output partition.
+///
+/// Implementations must remain safe when invoked from native execution
+/// threads that do not inherit Spark's task-local JVM state.
+pub trait ShufflePartitionPusher: Send + Sync {
+    /// Sends one complete, length-prefixed Arrow IPC shuffle block.
+    fn push_partition_data(&self, partition_id: i32, data: &[u8]) -> Result<()>;
+}
+
+/// Invokes a task-owned JVM shuffle callback from any native execution thread.
+///
+/// The global callback reference keeps both the Java object and its class alive
+/// for the lifetime of the cached method ID. No thread-local JNI environment
+/// or Spark task context is retained between invocations.
+pub struct JavaShufflePartitionPusher {
+    callback: Global<JObject<'static>>,
+    push_method: JMethodID,
+}
+
+impl JavaShufflePartitionPusher {
+    /// Captures the callback while running on an attached JVM thread.
+    pub fn try_new(env: &mut Env<'_>, callback: &JObject<'_>) -> Result<Self> {
+        if callback.is_null() {
+            return Err(DataFusionError::Execution(
+                "Remote shuffle callback must not be null".to_string(),
+            ));
+        }
+
+        let callback_class = env.get_object_class(callback).map_err(CometError::from)?;
+        let push_method = env
+            .get_method_id(
+                &callback_class,
+                jni::jni_str!("pushPartitionData"),
+                jni::jni_sig!("(I[BI)V"),
+            )
+            .map_err(CometError::from)?;
+        let callback = env.new_global_ref(callback).map_err(CometError::from)?;
+
+        Ok(Self {
+            callback,
+            push_method,
+        })
+    }
+
+    fn checked_payload_length(partition_id: i32, payload_length: usize) -> Result<i32> {
+        if partition_id < 0 {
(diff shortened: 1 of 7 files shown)


native/shuffle/src/writers/rss/rss_partition_writer.rs
```diff
@@ -0,0 +1,214 @@
...
+                &mut Cursor::new(&mut self.frame),
+                &mut self.compression_context,
+                &metrics.encode_time,
+            )?;
+
+            if encoded_size == 0 {
+                continue;
+            }
+
+            if encoded_size > self.max_frame_size {
```
ziting-openai, line 143: **Withdrawn by Codex:** This feedback was posted by a misconfigured AI agent and should be disregarded. Sorry for the unsolicited review.

<details>

[P2] Split oversized batches before encoding instead of failing valid RSS writes. max_frame_size is checked only after ShuffleBlockWriter::write_batch has already encoded the entire Arrow batch into an unbounded Vec. Consequently a normal multi-row batch whose IPC output exceeds the transport frame limit fails the whole shuffle task even when every row would fit in its own complete frame, and the configured limit does not bound the allocation that can OOM an executor. The current oversized-frame test codifies this failure. Estimate/bound allocation before encoding and split batches at row boundaries into independently decodable complete frames; reject only a genuinely unsplittable row. AI-assisted review (Codex).

</details>

_Posted by Codex on behalf of ziting-openai to withdraw an earlier automated review._

sunchao (approved):
LGTM for this foundational scope.

Verified 50 native shuffle tests, 3 JNI validation tests, 9 additional adversarial cases, and JNI runtime probes for native-thread callbacks and exception preservation. The full Spark/Maven regression suite was not rerun.

The frame-size concern is a follow-up before RSS integration: the current contract caps callback payloads and intentionally rejects oversized frames; it does not promise automatic splitting or bounded serialization memory.

AI-assisted review (Codex).

native/jni-bridge/src/shuffle_partition_pusher.rs
```diff
@@ -0,0 +1,147 @@
...
+            return Err(DataFusionError::Execution(format!(
+                "Remote shuffle partition must be nonnegative, got {partition_id}"
+            )));
+        }
+
+        i32::try_from(payload_length).map_err(|_| {
+            DataFusionError::Execution(format!(
+                "Remote shuffle payload size {payload_length} exceeds the JVM array limit"
+            ))
+        })
```
xhumanoid, line 79: in specification it Integer.MAX_VALUE, but HotSpot implementation details Integer.MAX_VALUE-8
so maybe make sense to test it as well

https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java#L854-L866

```
    /**
     * A soft maximum array length imposed by array growth computations.
     * Some JVMs (such as HotSpot) have an implementation limit that will cause
     *
     *     OutOfMemoryError("Requested array size exceeds VM limit")
     *
     * to be thrown if a request is made to allocate an array of some length near
     * Integer.MAX_VALUE, even if there is sufficient heap available. The actual
     * limit might depend on some JVM implementation-specific characteristics such
     * as the object header size. The soft maximum value is chosen conservatively so
     * as to be smaller than any implementation limit that is likely to be encountered.
     */
    public static final int SOFT_MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
```

native/shuffle/src/writers/rss/rss_partition_writer.rs
xhumanoid: > Split oversized batches before encoding instead of failing valid RSS writes.

it important and need to introduce splitting before encoding, 
but not exclude requirement to check size after compression
because depends from data and codec, final size can be bigger than initial data

andygrove:
The `pub use RssPartitionWriter` doesn't actually expose anything usable yet. `PartitionWriter` is `pub(crate)`, so all three of `write`, `finish_partition`, and `finish_all` live on a trait that downstream crates can't import — a crate outside `datafusion-comet-shuffle` can call `try_new` and then do nothing with the result. I confirmed this with a throwaway integration test in `native/shuffle/tests/`: `error[E0599]: no method named 'write' found for mutable reference `&mut RssPartitionWriter``.

Since the stated goal here is to expose the writer for integration in subsequent PRs, could we either make `PartitionWriter` `pub` in this PR, or drop the re-export from `lib.rs` and add it in the PR that actually wires it up? If the integration is going to live inside the shuffle crate anyway (e.g. `ShuffleWriterExec` gaining an RSS mode), the second option seems cleaner — the writer stays `pub(crate)` alongside `LocalPartitionWriter` and the API surface doesn't grow.

andygrove (approved):
LGTM with one nit - can be handled in a future PR though. Thanks @pingzh!

sunchao:
Merged, thanks @pingzh !! Please address the follow-ups in a separate PR

pingzh:
@sunchao @andygrove , here is the follow up pr https://github.com/apache/datafusion-comet/pull/5475 , 🙏

ziting-openai:
Sorry for the noise. My automated review on this PR was posted by a misconfigured AI agent that should have been restricted to OpenAI-owned repositories. Please disregard that automated review. GitHub does not give my account permission to resolve [my inline thread](https://github.com/apache/datafusion-comet/pull/5473#discussion_r3860020195); could a maintainer please mark it resolved? This agent will not review non-OpenAI repositories again.

Issue #323, opened 2024-04-25
CometShuffleExchangeExec logical link is different to Spark ShuffleExchangeExec


viirya:
### Describe the bug

Enabling columnar shuffle by default PR #250 causes a lot of test failures in Spark `org.apache.spark.sql.execution.adaptive.AdaptiveQueryExecSuite`.

This is a tricky bug.

I found that it is because Comet sets incorrect logical link to `CometShuffleExchangeExec`. Because logical link is used in Spark AQE framework to do query planning, incorrect logical links cause test failures there.

PR #5483, opened 2026-08-26
fix: preserve current AQE logical-stage links on Comet operators

### LINKED ISSUE #5482: CometExecRule overwrites direct AQE LogicalQueryStage links during replanning
## Describe the bug

`CometExecRule` can overwrite the current AQE logical-stage link on a reused native operator with an older link from `originalPlan`.

Spark can represent a native final aggregate above a shuffle stage as a `LogicalQueryStage`. During replanning, Spark's planner reuses that physical aggregate and gives it a direct `SparkPlan.LOGICAL_PLAN_TAG` pointing to the current logical-stage object. Comet's subsequent logical-link repair unconditionally restores `originalPlan.logicalLink`, or clears the tags when the original link is absent.

The original logical aggregate is now stored inside a logical-stage leaf, rather than appearing as a node in the active logical tree. Restoring that older link breaks the correspondence between the current physical root and the current logical stage. A new exchange above the aggregate can inherit the stale link, but Spark's identity-based logical-stage replacement cannot find that old logical node in the active tree.

The unconditional restoration is present on OSS `main` at [`a2c6bd4b930174cf81e0ad2857d2feb422e081c5`](https://github.com/apache/datafusion-comet/blob/a2c6bd4b930174cf81e0ad2857d2feb422e081c5/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala#L647-L678).

## Steps to reproduce

The following planner-level regression can be added inside `CometExecRuleSuite`, using its existing `createSparkPlan` and `applyCometExecRule` helpers. It creates a native final aggregate above a shuffle query stage, then passes a `LogicalQueryStage` through Spark's actual planner twice. AQE execution is disabled only to construct the initial plan; the test explicitly exercises its planner reuse sequence.

```scala
import org.apache.spark.sql.execution.adaptive.{LogicalQueryStage, ShuffleQueryStageExec}
import org.apache.spark.sql.internal.SQLConf

test("CometExecRule preserves the current direct AQE logical link") {
  withSQLConf(
    SQLConf.ADAPTIVE_EXECUTION_ENABLED.key -> "false",
    SQLConf.WHOLESTAGE_CODEGEN_ENABLED.key -> "false",
    CometConf.COMET_SHUFFLE_MODE.key -> "native",
    CometConf.COMET_SPARK_TO_ARROW_SUPPORTED_OPERATOR_LIST.key -> "Range") {
    val plan = createSparkPlan(
      spark,
      "SELECT id % 3 AS k, SUM(id) AS total FROM range(0, 100, 1, 2) GROUP BY id % 3")
    val aggregate = applyCometExecRule(plan).asInstanceOf[CometHashAggregateExec]
    val shuffle = aggregate.child.asInstanceOf[CometShuffleExchangeExec]
    val originalLogicalPlan = aggregate.originalPlan.logicalLink.get
    var current: SparkPlan = aggregate.withNewChildren(
      Seq(ShuffleQueryStageExec(0, shuffle, shuffle.canonicalized)))

    (1 to 2).foreach { _ =>
      val logicalStage = LogicalQueryStage(originalLogicalPlan, current)
      val replanned = spark.sessionState.planner.plan(logicalStage).next()
      assert(replanned eq current)
      assert(replanned.getTagValue(SparkPlan.LOGICAL_PLAN_TAG).exists(_ eq logicalStage))

      current = applyCometExecRule(replanned)
      assert(current.getTagValue(SparkPlan.LOGICAL_PLAN_TAG).exists(_ eq logicalStage))
    }
  }
}
```

The final assertion specifies the required invariant: applying Comet's rule must not replace the direct link that Spark just assigned. The current repair branch instead restores `originalLogicalPlan`. The same preservation requirement applies when `originalPlan` has an inherited link or no logical link at all.

## Expected behavior

Preserve an existing direct `LogicalQueryStage` link on a `CometExec` when the rule revisits it. The tag must continue to reference the exact current logical-stage object after repeated replanning.

Keep the existing repair and clearing behavior for ordinary direct links and inherited links. In particular, merely inheriting a `LogicalQueryStage` link from an ancestor must not trigger preservation. Leave shuffle and broadcast exchange link handling unchanged, including the empty-link invariant from #323.

## Additional context

This follows public Spark planner behavior: [`LogicalQueryStageStrategy`](https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/LogicalQueryStageStrategy.scala#L64-L65) returns the existing physical plan, and [`SparkStrategies.plan`](https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkStrategies.scala#L78-L87) sets its direct logical link. [`AdaptiveSparkPlanExec.replaceWithQueryStagesInLogicalPlan`](https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L759-L782) replaces logical nodes by object identity.

A narrow fix is to skip logical-link repair only when `getTagValue(SparkPlan.LOGICAL_PLAN_TAG)` is a `LogicalQueryStage`. No Spark changes, configuration changes, or changes to native aggregate execution are needed.

Reproduced locally on OSS Spark 4.1.3 / Scala 2.13.17 with JDK 17 and a native library built from this OSS checkout. With the production rule unchanged, the direct-link regression fails at the assertion that Comet preserves the current logical-stage object. A generated-data SQL regression also returns the correct result but finishes with zero native broadcast hash joins instead of the expected two. With the narrow preservation guard, both regressions pass, along with the rest of `CometExecRuleSuite` and the existing shuffle logical-link test (33 tests total).

This establishes an OSS planner and native broadcast adaptation regression. It does not claim a particular cancellation pattern or benchmark result.

sunchao:

Closes #5482.

## Rationale for this change

Spark's Adaptive Query Execution (AQE) can use runtime statistics to replace a join with a broadcast hash join when an input is small enough. Comet can disrupt this adaptation for native aggregate inputs, leaving queries with correct results but without the expected native broadcast joins.

For example, the added regression joins several aggregates built from generated data. One of its join inputs is:

```sql
SELECT id % 64 AS k, SUM(id + 1) AS v
FROM range(0, 3072, 1, 4)
GROUP BY id % 64
```

This reduces 3,072 input rows to just 64 groups. The full regression uses two 64-row grouped inputs on the right sides of successive left joins. It disables automatic broadcast hash join selection during initial planning but allows AQE to select those joins later, with a 10 MiB adaptive broadcast threshold. The native broadcast hash joins must therefore come from adaptive replanning.

The local comparison shows why both the result and the final plan matter:

| Full regression query | Without the fix | With the fix |
| --- | ---: | ---: |
| Native broadcast hash joins in the final plan | 0 | 2 |
| Query result | `48738816` | `48738816` |

AQE relies on links between physical operators and logical plan nodes to update a query's plan as stages finish. When Spark represents a native final aggregate above a shuffle stage as a `LogicalQueryStage`, it reuses the physical aggregate and links it directly to that current logical-stage object. Comet's repair step then overwrites this fresh link with the older `originalPlan.logicalLink`, or clears it if the original has no link.

The older logical aggregate is now hidden inside the logical-stage leaf, no longer a node in the active logical tree. If a later exchange inherits that stale link, Spark cannot insert its query stage into the active tree: replacement looks for the exact referenced object, which is no longer there. AQE can then lose the new stage during subsequent replanning.

This PR lets Spark's current AQE stage assignment take precedence over Comet's saved original link. When `CometExecRule` revisits a native operator that already has a direct `LogicalQueryStage` link, it leaves that link intact. The reused operator can therefore stay associated with the stage Spark is currently planning, even across repeated replanning and even when its original plan has an older or missing link.

The exception is limited to direct stage links. An inherited link can come from an ancestor and does not establish that the operator represents that stage, so ordinary and inherited links still follow the existing repair and clearing rules. Shuffle and broadcast exchange handling also stays unchanged, including the empty-link behavior required by #323. The fix changes planning metadata; it requires no changes to Spark, native aggregate execution, or configuration. The broader audit of logical-link repair is tracked separately in #6034.

The planner regressions exercise Spark's actual reuse path through two successive replanning passes. They check that the exact current stage link survives whether the original plan has a direct, inherited, or absent link, and that ordinary and inherited links still receive the existing repair or clearing behavior.

A DPP execution regression observes Spark's preparation rules before and after Comet repairs a reused broadcast root. It verifies that the actual root reaches the direct-stage-link case, the physical plan and standard cost remain unchanged, and the temporary logical identity survives Comet's repair. An isolated copy with conflicting temporary and direct links also exercises Spark's stage-link assignment to verify that the temporary link takes precedence. The SQL result and native broadcast/DPP operators are checked, and the test skips Spark 3.4, where native AQE DPP is unsupported.

The execution regression runs the generated-data query with native shuffle and checks its result and adaptive broadcast joins. It also verifies that both grouped aggregates feeding the broadcasts retain direct stage links, each produce 64 rows, and record positive native compute time.

Local validation of the current change used Linux x86_64, JDK 17, Spark 4.1.3, and Scala 2.13.17, with the native library built from the rebased source using `make core` in debug mode with debug symbols disabled. Maven ran from the reactor root: all 35 `CometExecRuleSuite` tests and three targeted execution regressions passed (38 tests total). These cover native aggregate broadcast adaptation, shuffle logical links, and adaptive shuffled-join conversion. Scalastyle, Spotless, and `git diff --check` passed.

The original controlled before/after comparison on Spark 4.1.3 made both the direct-link planner regression and the SQL broadcast assertion fail when only the production guard was removed. CI is configured to run Spark SQL tests and Comet's supported Spark profiles for additional version coverage.


andygrove:

The write-up on this one is really good. The explanation of why a stale link makes AQE lose the stage during replanning is the clearest description of that failure mode I have seen, and the unit test in `CometExecRuleSuite` that walks two replanning rounds across all three original-tag shapes is exactly the right way to pin the behavior down.

Some things I would like to see addressed.

**Only one of the three link-repair blocks is guarded**

`CometExecRule.scala` has three near-identical repair blocks in that same `transform`, one for `CometExec`, one for `CometShuffleExchangeExec`, and one for `CometBroadcastExchangeExec`. The new guard is only on the first one. The rationale in the description talks about a later exchange inheriting the stale link, so it is not obvious to me why the exchange cases are safe to leave alone. Can an exchange ever carry a direct `LogicalQueryStage` tag that we would then clobber? If the answer is no, a sentence in the comment saying so would save the next reader the same question.

Related to that, those three blocks are copies of each other and now they have diverged. Would you be willing to pull the repair into a single private helper that all three cases call, with the guard living inside it? Right now a future change to the repair logic has to be made in three places and it is easy to miss one, which is roughly what happened here.

**Metric assertions in the `CometExecSuite` test**

The new test asserts `aggregate.longMetric("elapsed_compute").value > 0`. That does not tell us anything about the logical-link fix and it is the kind of timing assertion that eventually flakes on a loaded CI runner. The `output_rows == 64` check is useful because it confirms we found the right aggregate. Could the `elapsed_compute` one just be dropped?

**Query shape and runtime**

The regression builds four `range` scans totaling roughly 10k rows and joins them. That is fine for correctness, but `CometExecSuite` is already one of the slower suites. Did you check what this adds to the suite's wall clock? If it is meaningful, it might be worth shrinking the ranges, since the point is the number of groups on the build side and not the input row count.

**Tracking the general problem**

Is #5482 the right home for the broader question of whether Comet's link repair is still needed at all in its current form, or is that a separate issue? The repair exists because `originalPlan` is the source of truth for the link, and this PR establishes that it is not always the source of truth. It would be good to have that written down somewhere other than a code comment.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -646,6 +646,15 @@ case class CometExecRule(session: SparkSession)
...
       // Set up logical links
       newPlan = newPlan.transform {
+        case op: CometExec
+            if op
+              .getTagValue(SparkPlan.LOGICAL_PLAN_TAG)
+              .exists(_.isInstanceOf[LogicalQueryStage]) =>
+          // AQE replanning reuses this physical root and links it to the current logical stage.
+          // originalPlan can still point to a subtree hidden inside that logical leaf, which
+          // AQE cannot replace in the current logical plan. Only preserve a direct stage link,
+          // not a link inherited from an ancestor.
```
peterxcli, line 701: is it possible to link some spark code snippet here with version tag to show why this case is added? so readers can understand these comment with more context

sunchao:
Updated in [92b0baac9](https://github.com/apache/datafusion-comet/commit/92b0baac92bafdb866547fda3161c4ce3eab981a).

I checked the exchange question against the actual AQE/DPP lifecycle before widening the guard. On the ordinary exchange path, Spark [wraps exchanges in query-stage leaves](https://github.com/apache/spark/blob/v4.1.3/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L586-L637), so this transform does not revisit the exchange inside them.

There is an exception worth distinguishing: the manually inserted DPP broadcast can carry a direct `LogicalQueryStage` link as the whole adaptive-plan root. I observed this in real SQL with a grouped fact input joining a grouped, filtered dimension, and in a grouped self-join. In both observed replans, repair restored the original aggregate link, but the prepared physical tree was equal to the reused tree, both costs were `SimpleCost(0)`, and `TEMP_LOGICAL_PLAN_TAG` retained the original aggregate identity. With Spark's standard cost evaluator, [AQE rejects that unchanged replan](https://github.com/apache/spark/blob/v4.1.3/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L343-L370), and [new stage creation prioritizes the retained temporary link](https://github.com/apache/spark/blob/v4.1.3/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L715-L727). All three probe queries matched Spark and retained Comet DPP subqueries. This does not support an additional exchange guard for the lifecycle tested here, so I kept the existing exchange restore/clear behavior and the #323 empty-link contract. I also left helper refactoring out of this focused fix.

I retained `elapsed_compute > 0`: it checks a recorded native measurement, not a latency target. Comet exposes nanoseconds, and the locked DataFusion timer [adds at least one nanosecond per recorded duration](https://github.com/apache/datafusion/blob/54.1.0/datafusion/physical-expr-common/src/metrics/value.rs#L191-L204). A loaded CI runner does not create the proposed zero-duration failure.

For runtime, the existing [Spark 4.1 CI job](https://github.com/apache/datafusion-comet/actions/runs/33020356343/job/98354848063) recorded 471 ms for this regression in an approximately 128-second suite, so I kept the input sizes. #5482 records the concrete stale-stage-link defect; removing all original-plan repair would be separate work and still needs to preserve #323.

The follow-up adds the requested version-pinned Spark references beside the guard. Local validation rebuilt the full Spark 4.1.3 JVM reactor and passed all 31 planner tests plus five execution/AQE/DPP/shuffle-link regressions (36 total). The separate DPP probe also asserts physical-plan equality, equal cost, and preservation of the temporary logical identity. These local runs reused a previously built OSS native library; no native code changed.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
sunchao: Updated in 92b0baac92bafdb866547fda3161c4ce3eab981a. Added version-pinned Spark 4.1.3 links beside the guard: `LogicalQueryStageStrategy` returns the existing physical root, then `SparkStrategies.plan` assigns its direct logical link. The comment also distinguishes the ordinary exchange path, where the exchange is behind a query-stage leaf, and keeps the direct/inherited distinction explicit.

andygrove (commented):
The follow-up commit and your replies cover the points I raised earlier. The guard question, the `elapsed_compute` assertion, and the runtime numbers all check out on a closer look. I checked the `LogicalQueryStageStrategy` and `SparkStrategies.plan` sequence you cited against Spark 3.4.3, 3.5.8, and 4.0.1, not only 4.1.0. The reuse-by-identity return and the direct `setLogicalLink` call are unchanged across all of them, so the guard in `CometExecRule.scala` should generalize the way you describe. I also traced the other places that copy a `logicalLink` onto a Comet node, in `EliminateRedundantTransitions.scala` and `CometPlanAdaptiveDynamicPruningFilters.scala`. Both build a brand new destination node in the same statement, so neither can hit the clobber this PR fixes. On the metric assertion, DataFusion's `Time::add_duration` floors every recorded duration at 1 nanosecond, so `elapsed_compute > 0` can't read zero on a loaded runner either.

Your reply about the exchange cases mentions a separate local probe for the DPP broadcast scenario, where a direct `LogicalQueryStage` link can land on a `CometBroadcastExchangeExec` at the root of a DPP subquery's own adaptive plan. Leaving that case unguarded relies on Spark's cost evaluator rejecting a no-op replan, and on `TEMP_LOGICAL_PLAN_TAG` taking priority inside `setLogicalLinkForNewQueryStage`. That's real Spark behavior you found by testing, not something Comet's own code guarantees. The probe itself isn't part of this diff. Would it be worth committing it as a regression test next to the two already in `CometExecRuleSuite`? If either of those Spark internals shifts in a later release, this path could regress the same way the original bug did, with correct results and no native broadcast join.

Is #5482 still the right place to track whether the general link-repair mechanism should exist in its current form, or would a new issue be better? I don't see that broader question filed anywhere yet, and I'd rather it not end up living only in this comment thread.

sunchao:
Addressed the two remaining follow-ups from [the September 9 review](https://github.com/apache/datafusion-comet/pull/5483#pullrequestreview-5154584449).

The new `CometExecRuleSuite` regression executes a grouped fact/dimension join and requires both native broadcast joins and Comet DPP, as well as the expected rows. Observers around Comet's AQE preparation rules require the direct `LogicalQueryStage` broadcast-root case to occur, then check that the physical tree and standard cost are unchanged and the original temporary logical identity survives Comet's repair. A captured copy with conflicting direct and temporary links calls Spark's actual stage-link assignment helper and verifies that the temporary link wins. The observer state is scoped to the test, and Spark 3.4 skips this native AQE DPP case.

The broader question of which `originalPlan` repairs remain necessary is now tracked in #6034. It covers link ownership, repeated AQE replanning, DPP broadcast roots, and the #323 clearing contract; #5482 stays focused on the reproduced stale-stage-link defect.

Committed in [69b2ca86a](https://github.com/apache/datafusion-comet/commit/69b2ca86ae51601e468d23c56ce71f8df36bbafc). Local validation on Spark 4.1.3 / JDK 17 passed all 35 planner tests and three targeted execution regressions (38 total), using the native library built from the rebased source. Spotless, Scalastyle, and `git diff --check` passed. The Spark SQL and all-Spark-profile CI labels are enabled; those checks are pending.

andygrove (approved):
LGTM. Thanks @sunchao

Issue #5396, opened 2026-08-19
[EPIC] Criterion bench coverage for all native expressions


andygrove:
## Background

Follow-up to #5363. The Scala end-to-end microbenchmarks conflate expression cost with scan and columnar-to-row transfer, so a Rust-side Criterion bench is the right tool for actually attributing per-row time to an expression. We already have a bench harness under `native/spark-expr/benches/`, but coverage is ad hoc.

This issue is the catalog: what native expressions exist, which have Criterion benches, which don't, and a proposal for filling the gap deliberately (not one bench per one expression — there are hundreds of thin wrappers around identical shapes).

## Scope: what counts as a "native expression"

Two sources reach native code from Comet:

1. **Comet-owned kernels** in `native/spark-expr/src/`. These have their own Rust implementations that Comet is responsible for.
2. **DataFusion / datafusion-spark passthroughs** dispatched by function name via `CometScalarFunction(...)` in `spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala`. Comet's responsibility here is limited to argument massaging and correctness gating — the kernel itself is upstream.

For (2), a bench inside Comet's tree is redundant with upstream benches for the kernel itself, but *is* useful when Comet wraps the passthrough with pre/post processing (input coercion, timezone handling, NULL semantics) or when the compat gate (`allow_incompat`) selects a different path than the plain DF call. The catalog below treats those distinctly.

## Existing Criterion benches

40 bench files under `native/spark-expr/benches/`:

`aggregate`, `array_size`, `arrays_overlap`, `base64`, `bloom_filter_agg`, `cast_binary_to_string`, `cast_decimal_to_string`, `cast_float_to_decimal`, `cast_float_to_string`, `cast_from_boolean`, `cast_from_string`, `cast_int_to_decimal`, `cast_int_to_timestamp`, `cast_non_int_numeric_timestamp`, `cast_numeric`, `cast_string_to_date`, `cast_string_to_timestamp`, `ceil`, `check_overflow`, `checked_arithmetic`, `conditional`, `date_trunc`, `date_trunc_array_fmt`, `decimal_div`, `decimal_rescale`, `floor`, `get_json_object`, `make_decimal`, `map_sort`, `normalize_nan`, `padding`, `parse_url`, `regexp_extract`, `round`, `to_csv`, `to_json`, `to_time`, `unhex`, `unscaled_value`, `wide_decimal`

[...]

PR #5492, opened 2026-08-27
chore: Bench string functions

coderfender:

Part of #5396

## Rationale for this change

Stacked on the math-benches PR (`bench_math_scalar_expr`), which introduces the shared common/mod.rs harness this builds on. Please merge that first, or review this against it as the base.

`benches/levenshtein.rs, split.rs` — `spark_levenshtein` over two Utf8 columns

### DIFF
--- a/native/spark-expr/Cargo.toml
+++ b/native/spark-expr/Cargo.toml
@@ -222,3 +222,15 @@ harness = false
 [[bench]]
 name = "cast_int_to_decimal"
 harness = false
+
+[[bench]]
+name = "log"
+harness = false
+
+[[bench]]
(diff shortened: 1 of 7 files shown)


sunchao (approved):
Reviewed `17e55ec1` against `73ff2f3`; no actionable P1/P2 findings. The math prerequisite is already in base. CI type-checked the four new targets, and an isolated component using unchanged HEAD sources passed all 45 smoke cases plus separate output checks for all 45 fixtures. This is not performance or end-to-end Spark validation.

sunchao:
Merged, thanks @coderfender !

PR #5493, opened 2026-08-27
fix: make CometDiskBlockWriter spill registry per-task instead of executor-global

peterxcli:

Part of #5212 (positions 6, 7, 9, and 10). It does not close the epic.

## Rationale for this change

`CometDiskBlockWriter.currentWriters` is declared `static final`, so despite its comment ("of same shuffle task") it is an executor-global registry shared by every shuffle task in the JVM. This causes several problems, all reproduced with a deterministic two-task reproducer against current `main`:

- **Cross-task spilling with wrong accounting (position 6):** when task A cannot allocate a page, `spill()` sorts the global list and force-spills the largest writers regardless of owner. In the reproducer, task A force-flushed task B's 3,145,728 buffered bytes (2,751 rows written to B's file from A's thread, disk-spill metrics charged to B), counted all of it toward its own `totalFreed`, and spilled none of its own data.
- **ABBA deadlock (new finding):** `insertRow` holds the writer monitor and then takes the registry lock inside `spill()`, while a concurrently spilling task holds the registry lock and requests the victim writer's monitor. A benchmark with two concurrent tasks under a shared memory cap deadlocked in 3 of 3 rounds (confirmed via `ThreadMXBean.findDeadlockedThreads()`).
- **`ConcurrentModificationException` (position 7):** the spill comparator reads other tasks' `allocatedPages` lists unsynchronized while their owners mutate them; observed live in the same concurrent benchmark.
- **Unsynchronized cross-thread mutation (position 9):** `spilling` and `totalWritten` are mutated from other tasks' threads without proper synchronization.
- **Failed-task writer leak:** `stop(false)` frees memory but never removes writers from the static list, so they are retained for the executor lifetime (12/12 writers still strongly reachable after GC across three simulated failed tasks). The retained writers also keep row addresses that point into freed pages, so a later task spilling them would hand freed addresses to native code.
- **Dead code (position 10):** `spillingWriters` is never populated, so its loop in `freeMemory()` is unreachable.

Separately, the old spill loop has an **ordinary single-task assertion failure**: it could stop after spilling a large sibling while leaving the requesting writer's current page allocated. `SpillWriter.initialCurrentPage()` requires that page to be null. A single task with eight writers against an eight-page pool reaches this assertion within 20,000 mixed-size rows on `main`; no second task or concurrency is required.

**1. Task-owned spill registry.** Replace the static registry with a task-owned `LinkedList` created in `CometBypassMergeSortShuffleWriter.write()` and passed to each partition's `CometDiskBlockWriter`. Spilling now only ever touches the requesting task's writers, matching Spark's own shuffle semantics, and every writer is confined to its task's thread (which also removes the deadlock, the `ConcurrentModificationException`, and the cross-thread mutation of positions 7/9 structurally). Writers of a failed task become unreachable together with the task's shuffle writer, removing the leak. The dead `spillingWriters` field and its loop in `freeMemory()` are deleted.

**2. Caller-first spilling.** A writer that cannot allocate now flushes itself before considering siblings, which preserves the `initialCurrentPage()` contract and fixes the ordinary single-task assertion above. Siblings are sorted and spilled only if the caller did not release enough memory; the loop continues until the request is covered.

**3. Blocking allocation for the shared on-heap pool (`CometBoundedShuffleMemoryAllocator.allocateBlocking`).** The registry change removes a task's only recourse when the shared pool is full: previously it force-spilled other tasks' buffers. In the opt-in on-heap compatibility mode all tasks share one bounded allocator, so without a replacement a task whose own writers hold nothing spillable would fail with `SparkOutOfMemoryError` whenever other tasks held the pool (reproduced in review with a real Spark 3.5.9 shuffle). After callers spill the buffered data they can cheaply release, the replacement waits for other tasks to free pool memory instead of failing, and `free()` notifies waiters. Per-thread accounting includes memory the requesting task still retains, such as the unsafe sorter's pointer array or sibling partition pages.

The waiting logic keeps per-thread accounting (which thread allocated each page, how much each thread retains, and which threads are blocked waiting with what request size) in order to fail fast, before or during a wait, exactly when the wait provably cannot succeed:

- the request does not fit next to the memory the requester itself still retains; or
- every allocated byte is retained by threads that are themselves blocked waiting and none of their requests fits in the free pool, i.e. no thread is left that could ever free memory.

The wait is additionally bounded by `spark.comet.shuffle.jvm.memoryWaitTimeout` (internal, default 5 minutes): the holders a wait depends on can themselves be blocked on resources outside this pool — Spark execution memory, locks, I/O — possibly in a cycle that only a task waiting here could break, and review demonstrated that no detector local to this allocator classifies such holders reliably in either direction. On timeout the waiter fails with the managed `SparkOutOfMemoryError`, as the pre-PR code did, so its task releases memory, the other resource can progress, and Spark's task retry can recover.

To keep such cycles from forming in the first place, the unsafe writer also allocates its sorter lazily: `open()` moved from the constructor into `write()`, after the first `records.hasNext()`, so no Comet pool memory is retained while Spark evaluates or lazily materializes the shuffle input (which can block on Spark execution memory). Writer-held allocations that fatal errors would otherwise orphan (Spark only calls `stop(false)` for exceptions) are reclaimed on every path: the bypass writer frees its partition writers' pages and deletes their temporary files on any `Throwable` out of `write()`, `SpillSorter` frees everything it allocated when its constructor fails part-way, and a task-completion listener backstops the unsafe writer.

While blocked, the allocator checks cooperative `TaskContext` cancellation at most once per second and also handles Java interruption. Flag-only cancellation throws Spark's `TaskKilledException`; Java interruption remains a non-fatal `RuntimeException` with the `InterruptedException` cause. A wait logs when it starts and every 30 seconds while it lasts. The off-heap (`CometUnifiedShuffleMemoryAllocator`) path is unchanged because Spark's memory manager already arbitrates between tasks.

The caller-first policy has no measured regression. With 200,000 rows, 200 partitions, and a 16-page pool, median wall time was 27.3 s on this branch versus 28.1 s on base, with identical spilled and output byte counts. With four partitions and occasional 120 KiB rows, this branch spilled 546,665 bytes versus 693,031 on base.

`CometDiskBlockWriterSuite` covers both the registry and allocator behavior:

- Three writers owned by task A share one task registry while task B buffers pages in a separate registry. Memory pressure first verifies that the requester flushes itself even when a larger sibling alone would satisfy the request, then a 900 KiB row forces the spill loop through two siblings. Task B's rows, metrics, and file remain untouched.
- On-heap shared pool: a task with nothing to spill waits for another task to finish instead of failing; unsatisfiable requests fail fast; Java interruption remains non-fatal; flag-only `TaskContext` cancellation promptly throws `TaskKilledException`; a notified waiter is not declared deadlocked while another waiter's smaller request can proceed; cross-pool cycles unwind at the configured timeout; and fatal writer/sorter failures reclaim their memory and temporary files.


spark/src/main/java/org/apache/spark/sql/comet/execution/shuffle/CometBypassMergeSortShuffleWriter.java
```diff
@@ -176,6 +177,10 @@ public void write(Iterator<Product2<K, V>> records) throws IOException {
                   CometShuffleExternalSorter.MAXIMUM_PAGE_SIZE_BYTES,
                   memoryManager.pageSizeBytes()));
 
+      // This task's disk writers. Under memory pressure a writer spills its sibling writers in
+      // this list, never those of other tasks.
+      final LinkedList<CometDiskBlockWriter> taskWriters = new LinkedList<>();
```
sunchao, line 182: [P2] Preserve progress for the shared on-heap shuffle pool

In the explicitly enabled on-heap compatibility mode, `CometShuffleMemoryAllocator.getInstance()` still returns one `CometBoundedShuffleMemoryAllocator` shared by the executor. If task B holds that pool before A allocates its first page, A's new list contains nothing spillable, so `initialCurrentPage()` now throws `SparkOutOfMemoryError`. I reproduced this through a real Comet JVM shuffle on Spark 3.5.9 with 256-KiB pages and a 1-MiB shuffle budget: B idles with four pages (751 binary rows), then A fails on its first row. The same job with the two exact base Java classes returns all 752 rows; this head also succeeds with a 2-MiB budget. Please preserve safe allocation progress for this shared pool when isolating the spill registries; the maintained on-heap Spark SQL compatibility path still uses it.

peterxcli: Good catch, thanks — fixed in 3cee056cd. The spill registries stay per-task, and allocation progress on the shared on-heap pool is preserved by making the post-spill retry (`SpillWriter.initialCurrentPage`) call a new `allocateBlocking` on `CometBoundedShuffleMemoryAllocator`, which waits (up to 60s) for other tasks to free pool memory; `free()` notifies waiters. That retry point is only reached after the requesting task has already spilled everything it owns, so a waiting task holds no pool memory itself and the tasks still holding it can always progress and free it — no cross-task waiting cycle is possible. The off-heap allocator keeps the non-waiting default since Spark's unified memory manager already arbitrates between tasks there. Added an on-heap regression test mirroring your reproducer (1 MiB pool, 256 KiB pages, task B idle with 4 pages, task A's first row): it fails with the same `UNABLE_TO_ACQUIRE_MEMORY` error without the change and passes with it.

sunchao (commented):
[P2] Rechecked `3cee056cd1ded0240892ef53ebfef515fd6c1e4e`: the [existing shared-pool finding](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3869474061) is partially fixed. An independently released 1.5-second holder now succeeds, but a real Spark 3.5.9 Comet JVM shuffle with a 1-MiB pool and 256-KiB pages still raises `SparkOutOfMemoryError` at the 60-second allocation deadline when another task independently pauses for 65 seconds while holding four pages. The exact base Java-class overlay and current-head 2-MiB control each return all 752 correct rows. This remains limited to opt-in on-heap compatibility; the probe allows one task attempt, so normal Spark retries may recover.

spark/src/main/java/org/apache/spark/sql/comet/execution/shuffle/CometBypassMergeSortShuffleWriter.java
peterxcli: Thanks for re-testing — addressed in 4c12797e0. You're right that any finite deadline fails some workload, so I removed it: `allocateBlocking` now waits until memory is freed, mirroring how Spark's unified memory manager blocks a task until memory becomes available. Liveness still holds for the same reason as before (a waiter has already spilled everything it owns, so pool memory is only held by tasks that can progress and free it), interruption (task kill) still aborts the wait, and the allocator logs a warning when a task starts waiting so long waits are diagnosable. I validated your exact scenario locally: 1-MiB pool, 256-KiB pages, task B holding four pages (751 rows) pausing 65 s — task A's first insert waited ~65.1 s, then both tasks completed with all 752 rows.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
```diff
@@ -112,6 +118,36 @@ public synchronized MemoryBlock allocate(long required) {
...
+      try {
+        return allocateMemoryBlock(size);
+      } catch (SparkOutOfMemoryError e) {
+        if (!logged) {
+          logger.warn(
+              "Waiting for other tasks to free up {} bytes of Comet shuffle pool memory", size);
+          logged = true;
+        }
+        try {
+          wait();
```
sunchao: [P2] Check that allocation can make progress before waiting

In opt-in on-heap JVM shuffle, this retry also runs from the unsafe sorter, whose empty `spill()` retains its initial pointer array. I reproduced two real Spark 3.5.9 tasks with a 1 MiB pool, 256 KiB pages, `spark.shuffle.sort.bypassMergeThreshold=0`, pointer capacity 4096, and a first binary row of 976 KiB each: both request 999448 bytes, but their arrays retain 65536 bytes, leaving only 983040 available. Both executor tasks remain here after the input gate opens, and neither can reach cleanup to free its array. Each task alone and both with a 2 MiB pool complete correctly; the exact base Java overlay raises an error promptly. A sole oversized row also waits with an empty pool. Please handle retained allocations and unsatisfiable requests before entering an unbounded wait.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
```diff
@@ -112,6 +118,36 @@ public synchronized MemoryBlock allocate(long required) {
...
+        } catch (InterruptedException ie) {
+          Thread.currentThread().interrupt();
+          throw e;
```
sunchao: [P2] Preserve task-kill semantics when the wait is interrupted

Throwing the preceding `SparkOutOfMemoryError` converts an intentional task kill into a counted failure: it extends `OutOfMemoryError`, so Spark's killed-task handler for `InterruptedException | NonFatal(_)` does not match it and reports `ExceptionFailure`. In a real Spark 3.5.9 Comet JVM shuffle, `killTaskAttempt(id, true)` on a small-row allocation waiter produced `ExceptionFailure` and aborted the job with `local[2,1]`. The same-head control killed during an input wait reported `TaskKilled`; its retry returned all 752 correct rows. Both used a healthy holder with an independent finite release. Please preserve interruption/task-kill classification here so intentional kills do not consume the retry budget or abort the job when that budget is exhausted.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Addressed in b8fadf587. The allocator now tracks the pool memory retained per thread (owner recorded per page) and the set of threads blocked in `allocateBlocking`, and gives up before waiting when the wait can never succeed: (1) when the request does not fit next to the requester's own retained memory — this covers your sorter scenario (each task requesting 999448 bytes while retaining its 65536-byte pointer array fails immediately, since 999448 > 1048576 − 65536 even if everything else were freed) and the sole oversized row on an empty pool; and (2) when every allocated byte is retained by threads that are themselves blocked waiting, so nothing can be freed anymore (the mutual-block case where each request would individually fit). A new waiter wakes existing waiters so they re-evaluate (2) against the enlarged waiting set. Both cases have regression tests that fail fast (<25 ms) where they previously hung; the healthy-holder wait scenario is unaffected since a non-waiting holder keeps the wait alive.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Addressed in b8fadf587. The interrupt handler no longer rethrows the pending `SparkOutOfMemoryError`; it restores the interrupt flag and throws a `RuntimeException` wrapping the `InterruptedException`, which is NonFatal and therefore matched by Spark's killed-task handler — an intentional `killTaskAttempt` on a waiter is now classified as `TaskKilled` instead of `ExceptionFailure`, so it does not consume the retry budget. Added a regression test that interrupts a blocked waiter and asserts the surfaced error is not an `OutOfMemoryError` and carries the `InterruptedException` cause.

sunchao (commented):
[P2] Follow-up to the [allocation-progress thread](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3870526306): the new check at `CometBoundedShuffleMemoryAllocator.java:163` in `b8fadf58` can reject a satisfiable schedule. Notified threads remain in `waitingThreads` until allocation returns. With a 1-MiB pool and 256-KiB pages, after a healthy holder finishes, two unsafe-shuffle tasks each retain 32,768 pointer bytes and request 999,448 / 262,144 bytes. If the large waiter resumes first, it throws with 983,040 bytes free, although the smaller waiter can finish and release its array for it. In fresh Spark 3.5.9 opt-in on-heap tests, small-first arrival failed 10/10; exact `4c12797` Java overlays and current reversed arrival each returned three correct rows 10/10, and the matching 2-MiB control passed 3/3. Please let a satisfiable notified request proceed before declaring deadlock. This is an incremental regression against `4c12797`, reproduced with `local[3,1]`.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Follow-up addressed in de5d00a9b. You're right — the deadlock check treated notified-but-not-yet-resumed waiters as permanently blocked. `waitingThreads` now maps each waiter to its pending request size, and deadlock is only declared when, in addition to all allocated memory being retained by blocked waiters, none of their requests fits in the free pool. In your schedule, the large waiter that resumes first sees the small waiter's 262144-byte request fits the 983040 free bytes, so it keeps waiting; the small waiter proceeds, finishes, and releases its array, after which the large request fits (999448 ≤ 1048576 − 32768). The genuine mutual-block case is still detected, since there no waiter's request fits the free pool. Added a regression test with exactly your sizes (900-KiB holder, two waiters retaining 32768-byte arrays, 999448/262144-byte requests) asserting both waiters complete after the holder frees, regardless of wake order.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
```diff
@@ -112,6 +128,81 @@ public synchronized MemoryBlock allocate(long required) {
...
+            Thread.currentThread().interrupt();
+            // Not an allocation failure: stay non-fatal so that an intentional task kill is
+            // classified as TaskKilled rather than ExceptionFailure.
+            throw new RuntimeException(
+                "Interrupted while waiting for Comet shuffle pool memory", ie);
+          }
+        }
+      }
+    } finally {
+      waitingThreads.remove(self);
```
sunchao: [P2] Wake remaining waiters when the only fitting request is cancelled

`anyWaiterCanProceed()` can keep other threads waiting on an entry that this `finally` removes without notification. In a 1-MiB pool, let two large waiters each retain 32,768 bytes and request 999,448 bytes, while a zero-retained waiter requests 262,144 bytes. If the small waiter is interrupted, a healthy holder frees its last page, and the large waiters reacquire first, both wait again because the still-registered small request fits the 983,040 free bytes. The interrupted waiter then removes itself here. Neither remaining request fits, and there is no notification to re-evaluate the deadlock check, so they can remain blocked absent another pool event.

With the freshly compiled current allocator, a controlled legal monitor schedule reproduced this state in 2/20 trials (both survivors still waiting after 500 ms); notification alone restored termination. The freshly compiled `b8fadf58` control terminated 20/20. Source tracing maps the zero-retained request to a bypass writer waiting for its first page: its empty failure cleanup calls no allocator `free()`. Cancelling only that writer's independent job with `interruptOnCancel=true` need not retry it or cancel the unsafe waiters retaining pointer arrays; Spark task/memory-manager cleanup notifies different monitors. This mixed-job impact is source-inferred—the dedicated Spark cancellation control could not run within the disk reserve. Please notify/recheck surviving waiters when removing this entry.

andygrove:

The core of this is clearly right. A static registry that lets task A spill task B's buffers, then reads B's `allocatedPages` unsynchronized while B mutates it, is a genuine correctness and safety bug, and scoping the list to the task's `CometBypassMergeSortShuffleWriter` is the obvious fix. Deleting the dead `spillingWriters` field along the way is a bonus. The reproducer that shows the cross-task spill plus the `AssertionError` is convincing.

My concerns are all about the second half of the PR.

**The description does not mention `allocateBlocking`**

The "What changes are included in this PR?" section only describes the registry change. But this PR also adds about 80 lines of new blocking-allocation logic to `CometBoundedShuffleMemoryAllocator`, including a hand-rolled deadlock-avoidance check, `wait`/`notifyAll` coordination, and two new pieces of per-thread accounting. That is the riskiest code in the diff and a reader going by the description would not know it was here. Could you write up what it does and why it is needed?

I do understand the connection: once a task can no longer take memory from other tasks, it needs somewhere to go when the pool is full. But that is an argument that deserves to be made explicitly, and it may well be an argument for a second PR rather than a second half.

**`wait()` has no timeout**

`allocateBlocking` calls bare `wait()` and only ever wakes on `free()` or on another waiter joining. The two fail-fast checks cover the cases where the waiting set can prove progress is impossible. They do not cover a thread that holds pool memory, is not in `waitingThreads`, and is not going to free anything soon, for example one blocked in a long shuffle read or stuck on network I/O. In that situation `allocatedMemory > retainedByWaitingThreads()` holds, the check passes, and the task waits forever after a single `logger.warn`.

From an operator's point of view that is a hung task with one line of explanation. Would you add a bounded wait, either `wait(timeoutMillis)` with periodic re-logging so the stall is visible, or a configurable maximum after which it throws the original `SparkOutOfMemoryError`? An OOM that fails the task is much easier to diagnose than a job that never finishes.

**`retainedMemory` holds strong references to `Thread` objects**

Entries are removed from `retainedMemory` only when the retained total drops to zero through `free`. Any page that is never freed, which is exactly what happens on the task-failure paths this PR is partly about, leaves a permanent `Thread` key in an executor-lifetime map. `retainedByWaitingThreads` also walks the whole map under the allocator lock on every failed allocation. Would a `WeakHashMap`, or clearing the entry from a `TaskCompletionListener`, be safer here?

**Interrupt is rethrown as `RuntimeException`**

The comment says throwing a plain `RuntimeException` keeps an intentional task kill classified as `TaskKilled` rather than `ExceptionFailure`. I do not think the exception type is what drives that classification. Spark decides based on `TaskContext.isInterrupted` and the kill reason in `Executor.run`, and a `RuntimeException` surfacing out of the writer is not obviously different from any other. Could you point at where that behavior comes from, or verify it with a test that kills a task mid-wait and checks the reported failure reason?

**Test flakiness**

`CometDiskBlockWriterSuite` is 439 lines of two-thread coordination with `CountDownLatch` and `TimeLimits`, and it is now wired into both the Linux and macOS PR builds. That is the right place for it, but concurrency tests of this shape have a habit of becoming the flaky suite everyone learns to rerun. How long does it take, and are the timeouts generous enough for a loaded CI runner? If any assertion depends on one thread reaching a point before another without a latch enforcing it, that is the one to harden now rather than after the first flake.

peterxcli:
Thanks for the review @andygrove. Point-by-point:

**PR description** — agreed, that was a real gap. The description now has a dedicated section explaining `allocateBlocking`: why removing cross-task force-spill needs a replacement on the shared on-heap pool, how the per-thread accounting drives the two fail-fast checks, and why the off-heap path is untouched. On splitting into a second PR: the registry change alone is a regression for opt-in on-heap mode (@sunchao reproduced a real Spark 3.5.9 shuffle failing on the intermediate commit), so the two halves need to land together to keep every commit releasable.

**`wait()` has no timeout** — took your first option in cab20e38f: the wait is now `wait(30_000)` chunks that re-log every 30 s with elapsed time, requested bytes, free bytes, and waiter count, so a stall caused by a holder stuck outside the allocator is visible in executor logs. I kept it unbounded rather than adding a deadline because an earlier revision had exactly that (60 s cap) and @sunchao demonstrated it fails a legitimate workload whose healthy holder simply outlives the cap — any finite default has that problem, and it is also how Spark's own `UnifiedMemoryManager` behaves (its `lock.wait()` loop has no deadline either).

**`retainedMemory` Thread references** — I looked at this and I believe it is already safe, so no change: entries are removed eagerly the moment a thread's retained total reaches zero (`free` decrements per page owner), so the map cannot accumulate entries across normally-completing tasks. A lingering entry requires a page that is never freed, and such a page pins its owner thread through `pageOwners` regardless, so a `WeakHashMap` would not reclaim anything — and if the entry did vanish while the leaked page stayed allocated, the deadlock check would lose sight of memory that is genuinely unfreeable and waiters would hang instead of failing. A `TaskCompletionListener` cannot free those pages either (the leak means cleanup did not run). The map walk happens only on the failed-allocation path and its size is bounded by the number of concurrent shuffle tasks.

**Interrupt classification** — the exception type is what drives it. Spark's killed-task clause in `Executor.TaskRunner.run` (Spark 3.5.9, `core/src/main/scala/org/apache/spark/executor/Executor.scala`) is `case _: InterruptedException | NonFatal(_) if task != null && task.reasonIfKilled.isDefined => ... TaskKilled`. `reasonIfKilled` is necessary but not sufficient — the guard only matches `InterruptedException` or NonFatal throwables, and `SparkOutOfMemoryError` extends `OutOfMemoryError`, which `NonFatal` rejects, so it falls through to the generic `case t: Throwable` and becomes `ExceptionFailure` even for a killed task. @sunchao verified this end-to-end in [this thread](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3870526317): `killTaskAttempt(id, true)` on a waiter produced `ExceptionFailure` with the old rethrow and `TaskKilled` behaves correctly with the NonFatal wrapper. There is also a unit test asserting the surfaced error is not an `OutOfMemoryError` and carries the `InterruptedException` cause.

**Test flakiness** — the suite runs in ~8 s locally with 60 s `failAfter` budgets. Every cross-thread ordering the assertions depend on is enforced by a latch or a state poll (threads observed in `WAITING`/`TIMED_WAITING` before the event that releases them); the one bare `Thread.sleep(500)` is one-sided — if the other thread is late, the scenario degrades to a trivially-passing variant, never a failure. Your instinct was right in one concrete way: switching to a timed wait moved blocked threads from `WAITING` to `TIMED_WAITING` and the state polls caught it immediately; they now accept both states (also in cab20e38f). 10/10 consecutive local runs pass.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
```diff
@@ -112,6 +131,96 @@ public synchronized MemoryBlock allocate(long required) {
...
+          // This thread cannot free what it retains while it waits, so a request that does not
+          // fit next to its own retained memory can never be satisfied.
+          if (size > totalMemory - retainedMemory.getOrDefault(self, 0L)) {
+            throw e;
+          }
+          // The allocation just failed, so the request does not fit in the unallocated pool.
+          // Waiting can only succeed while some thread can still free memory: either a thread
+          // outside the waiting set retains pool memory, or another waiter's request fits in the
+          // free pool, in which case that waiter can proceed and eventually free what it retains.
+          if (allocatedMemory <= retainedByWaitingThreads() && !anyWaiterCanProceed()) {
```
sunchao, line 194: [P2] Avoid unbounded waits on orphaned constructor allocations

An owner outside `waitingThreads` is not necessarily able to free its allocation. `SpillSorter` first constructs `ShuffleInMemorySorter(allocator, 1, true)` (8 bytes), then allocates its default 4,096-long array (32 KiB). If the second allocation fails, expansion never frees the first array. This happens inside the unsafe-writer constructor, before Spark receives a writer to clean up; Spark's task-memory cleanup does not own Comet's private pages, and `SparkOutOfMemoryError` does not terminate the executor.

A source-derived on-heap JVM-shuffle case uses a 1-MiB pool, 256-KiB pages and three task slots: a healthy bypass task retains 1,032,192 bytes; an unsafe task then allocates 8 bytes and fails its 32-KiB allocation. After that independent job fails and the healthy holder finishes, 8 orphaned bytes remain. An already-running bypass task on a different thread requesting its first 1,048,576-byte page has zero retained bytes, but only 1,048,568 bytes are free. Every timeout sees `allocatedMemory=8 > retainedByWaitingThreads()=0`, so this branch waits again without a remaining release path. The payload sizes are reachable with binary rows; mixed bypass/unsafe routing can use 2 versus 3 partitions and bypass threshold 2.

The tiny constructor leak predates this PR; changing the subsequent prompt allocation failure into an unbounded wait is the regression. Please reclaim failed-constructor allocations or otherwise account for departed owners before relying on this progress condition. This is static source/cleanup-path verification, not an executed reproduction.

sunchao (commented):
Re-reviewed `cab20e38` with five independent reviewers. Timed reevaluation addresses the [previous missed-notification stall](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3873174944) in source. The distinct P2 below is source-verified; no current local runtime reproduction was performed.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Addressed in a21f3f19e by reclaiming the failed-constructor allocations — your first option, since it restores the invariant the progress condition relies on (every allocated byte has an owner with a cleanup path) instead of trying to enumerate departed owners, which is unreliable anyway with pooled executor threads that outlive their tasks. `SpillSorter`'s constructor now wraps the pointer-array allocation and, on any failure, frees both the array (if allocated but not yet adopted) and the sorter's initial one-entry array before rethrowing. I audited the other consumers of the bounded allocator for the same shape: `CometDiskBlockWriter` allocates only lazily after construction, `CometShuffleExternalSorter` allocates nothing before creating its `SpillSorter`, and the mid-task paths (`reset`, `growPointerArray`) belong to writers Spark already cleans up via `stop(false)` — so this constructor was the only orphan window. Added a regression test with your sizes: a holder retaining 1,032,192 bytes of the 1-MiB pool, a `SpillSorter` constructor that fails on its 32-KiB array after the 8-byte one succeeds, and after the holder frees, a full-pool 1,048,576-byte allocation must succeed — it fails forever if the 8 bytes leak.

sunchao (commented):
[P2] The constructor leak in the [existing orphan-allocation thread](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3873969645) is fixed at `a21f3f19`, but bypass failures can still orphan pages. With opt-in on-heap JVM shuffle, a 1-MiB pool, 256-KiB pages and bypass threshold 3, a three-partition task can spill its largest writer and retain 524,288 sibling bytes. Before its 512,024-byte retry, an independent four-partition unsafe task can retain a 32,768-byte pointer array and wait for 999,448 bytes. Only 491,520 bytes remain free, so the all-waiter check throws `SparkOutOfMemoryError` in the bypass task.

Both bypass `write()` and Spark's `ShuffleWriteProcessor` catch only `Exception`; `stop(false)` is skipped, and task-memory cleanup does not own these private pages. The unrelated unsafe job can then wait indefinitely, including through every 30-second recheck. Please reclaim the bypass pages on this error path. The cleanup gap is pre-existing; the introduced regression is unbounded waiting. This is verified against current source and Spark 3.5.9 contracts, not an executed reproduction.

peterxcli:
Addressed the [bypass orphan-pages finding](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5044733895) in a1a4a84ef. You're right that `stop(false)` is skipped for fatal errors: both `write()` and Spark's `ShuffleWriteProcessor` catch only `Exception`, and `SparkOutOfMemoryError` is an `Error`. `CometBypassMergeSortShuffleWriter.write()` now catches `Throwable` and frees all partition writers' buffered pages before rethrowing, so a fatal failure (including the all-waiter fail-fast in the blocking allocator) can no longer orphan pool pages that an unrelated waiter would then wait on indefinitely. `freeMemory()` is idempotent, so the `stop(false)` that still runs on Exception paths is unaffected, and `CometUnsafeShuffleWriter` already handles generic throwables via its `success`/`finally` pattern (`sorter.cleanupResources()` runs in `finally`), so the bypass writer was the remaining gap. Added a regression test that drives the real `write()` (real `BlockManager` from a local `SparkContext`, stub `ShuffleExecutorComponents`): it buffers three pages of the 1-MiB pool, fails mid-write with a `SparkOutOfMemoryError`, and then asserts a full-pool 1,048,576-byte allocation succeeds — impossible if any buffered page leaked.

sunchao (commented):
Re-reviewed `a1a4a84`. The earlier bypass-write failure and failed-constructor cleanup are addressed. [P2] One case remains under the [existing orphan-allocation discussion](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3873969645), in Spark 3.5.9's pre-write path.

The unsafe writer allocates its pointer array before [Spark evaluates `rdd.iterator`](https://github.com/apache/spark/blob/7c14a3c28b141cc97a330c4d0f5d2a6da7267f85/core/src/main/scala/org/apache/spark/shuffle/ShuffleWriteProcessor.scala#L53-L99). Comet's eager round-robin input sort can then throw a managed `SparkOutOfMemoryError`. The writer's `write()` cleanup has not started, the processor's `Exception` handler skips `stop(false)`, and Spark task-memory cleanup does not own the 32-KiB Comet allocation.

With on-heap JVM shuffle and a 1-MiB Comet pool/page size, a bypass task on another thread can then wait indefinitely on that abandoned allocation. Controlled runs with the actual 3.5.9 processor, writers and shared Spark memory manager reproduced head waiting versus base failing promptly; explicit orphan cleanup resumed native output. These used constrained managed memory, manual task contexts and a fixed CI native binary, not a full scheduler run.

Please give allocations made before `write()` task-completion cleanup, or defer them until the writer's cleanup scope owns them.

peterxcli:
Addressed the [pre-write orphan finding](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5047933186) in 844b30867, taking the "give allocations task-completion cleanup" option: `CometUnsafeShuffleWriter`'s constructor now registers a `TaskCompletionListener` that runs `sorter.cleanupResources()` if the sorter is still alive. Completion listeners run in `Task.run`'s `finally` regardless of whether the failure was an Exception or an Error, so this covers the window you found — a fatal error thrown while Spark evaluates the input iterator, after the sorter allocated its pointer array but before `write()`'s own cleanup scope begins — as well as any other path that skips `stop(false)`. `cleanupResources()` is idempotent (the pointer array free is guarded, page lists are cleared, spill-file deletion checks existence), so it is a no-op on every normal path where `write()` or `stop()` already cleaned up. I did not defer the allocation instead, since the sorter's insert paths assume the array exists and lazy allocation would move the failure into a hotter path. Added a regression test that constructs the writer exactly as Spark does pre-iterator, asserts the pointer array actually occupies the 1-MiB pool (a full-pool allocation fails), then calls `markTaskCompleted` without `write()` or `stop()` ever running and asserts the full-pool allocation now succeeds — which fails without the listener.

sunchao (commented):
[P2] [The orphan-allocation issue](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3873969645) remains partially addressed at `844b308`. `SpillSorter` still calls `serializeSchema` outside the cleanup block after adopting its pointer array. A nested integer field with `parquet.field.id=2147483648L` passes the ordinary Spark 3.5 schema/encoder path, then throws in the Parquet ID conversion. The outer `sorter` is never assigned, so the new completion listener cannot release the 32 KiB array; in the opt-in on-heap JVM path that orphan can keep another job's first-page allocation waiting indefinitely. Please cover later constructor failures as well. This is verified from source; the passing pre-write test does not exercise it, and no current runtime reproduction is claimed.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Follow-up addressed in 4b592b9c2. The constructor cleanup now covers the whole tail of `SpillSorter`'s constructor — `serializeSchema` and the `Native` library load included — not just the pointer-array allocation. An `adopted` flag keeps ownership single: before `expandPointerArray` adopts the array it is freed directly (the sorter still owns only its initial one-entry array), and after adoption `inMemSorter.free()` releases it — avoiding a double free, which would trip `TaskMemoryManager`'s already-freed assertion in off-heap mode. Extended the constructor-leak regression test with your exact trigger: a nested field with `parquet.field.id = 2147483648` makes `ParquetUtils.getFieldId` throw `IllegalArgumentException` after the array is adopted, and the test then asserts a full-pool 1,048,576-byte allocation succeeds — which fails while the adopted 32-KiB array leaks. Also hardened the bypass writer's failure cleanup to skip null slots in case a mid-loop writer construction failure leaves the tail of `partitionWriters` unassigned.

sunchao (commented):
[P2] Account for live owners blocked in Spark's execution-memory pool

At `4b592b9`, the [reported orphan allocations](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3873969645) are reclaimed, but the [allocation-progress condition](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3870526306) still permits a live cross-pool cycle.

In opt-in on-heap JVM shuffle on Spark 3.5.9, with round-robin pre-sorting, bypass disabled and 4096-entry pointer buffers, each unsafe writer holds 32 KiB before its input iterator is built. B can finish sorting while it alone owns more than three quarters of the fully growable Spark execution pool; A then blocks below its minimum share in Spark while retaining its Comet array. For B's first nonfinal 976-KiB binary payload, its approximately 999448-byte Comet request cannot fit in a 1-MiB pool with both arrays (983040 bytes free), yet fits beside its own array (1015808 bytes). A is outside `waitingThreads`, so B waits too. Each task holds memory the other needs; neither can reach cleanup, and timed retries see unchanged state.

BASE fails and unwinds at this retry instead. Please avoid retaining pre-write Comet allocations across blocking input construction, or otherwise break this cross-pool cycle before waiting indefinitely. This is verified from current and matching public 3.5.9 source, not an executed reproduction.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Addressed the [cross-pool cycle](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5050181096) in 3ca39cf33, taking your first option: the unsafe writer no longer retains pre-write Comet allocations across blocking input construction. `open()` moved from the constructor into `write()`, so the sorter's pointer array is only allocated after Spark has finished evaluating the input iterator — a task blocked in Spark's execution-memory pool during an eager input sort now holds zero Comet pool bytes, and in your scenario B's 999448-byte request fits beside its own array (1015808 free) and proceeds, breaking the cycle. Everything that touches the sorter already runs inside `write()`, and `stop()`, peak-memory accounting, and the task-completion listener all null-guard it; the listener stays as a backstop for fatal errors during `write()` itself. I preferred this over cross-pool cycle detection since the Comet allocator cannot observe waiters in Spark's pool. Reworked the regression test accordingly: it now asserts a full-pool allocation succeeds right after writer construction (nothing retained pre-write; this fails on the previous head), then drives `write()` into a fatal mid-iteration error and asserts the pool is fully reclaimed again afterwards.

sunchao (commented):
[P2] Following up on the [cross-pool finding](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5050181096): moving `open()` fixes eager input construction, but it still reserves the Comet pointer array before `records.hasNext()`. On Spark 3.5 with on-heap JVM shuffle, bypass disabled, constrained pools and large rows, ordinary lazy Sort or partial HashAggregate input can then block in Spark execution memory. A second task can retain its nonfinal Spark input pages while waiting for a Comet page that fits beside its own pointer array but not both arrays. The first task is absent from `waitingThreads`, so the checks keep allowing a stable two-pool wait and neither task reaches cleanup. BASE instead raised a managed allocation error and unwound. Please also prevent retention across this lazy input boundary, or make the retry unwind when that dependency cannot progress. This is a source-derived reachable schedule, not an executed reproduction.

peterxcli:
Addressed the [lazy-input cross-pool follow-up](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5051189659) in ab5c033ad with both halves of your suggestion, since neither alone closes it: deferral cannot help once rows are buffered (mid-write `hasNext()` on streaming input can block on Spark memory just as the first call can), and unwinding alone would leave the common window wider than necessary.

1. **No retention across the lazy input boundary**: the sorter is now allocated only after `records.hasNext()` has produced the first record, so a task blocked in Spark execution memory while its input materializes holds zero Comet pool bytes.
2. **The retry unwinds when the dependency cannot progress**: after at least one full wait interval, a Comet waiter throws the managed `SparkOutOfMemoryError` (as base did) when the memory it waits for is retained by threads parked in an *untimed* wait whose stack is inside `ExecutionMemoryPool`/`UnifiedMemoryManager` — i.e. blocked acquiring Spark execution memory, which may in turn only be freed by tasks blocked here. The detection is deliberately narrow so the previously-reviewed healthy-holder scenarios keep waiting: a sleeping holder is `TIMED_WAITING`, a latch-parked holder with an independent release has no Spark-memory frames, and a computing or I/O-bound holder is `RUNNABLE` — none match.

Added a regression test that builds the real two-pool state: one task owns the whole Spark execution pool (a real `UnifiedMemoryManager`), a holder thread retains 500 KiB of the 1-MiB Comet pool and then genuinely parks inside `ExecutionMemoryPool` acquiring Spark memory (verified from its stack), and a Comet waiter requesting 800 KiB unwinds with `SparkOutOfMemoryError` instead of hanging; once Spark memory frees, the holder resumes and the pool drains fully. Also reworked the unsafe-writer test to assert nothing is allocated before the first record is consumed.

sunchao (commented):
[P2] Re-reviewed `ab5c033ad98d`: the [allocator progress issue](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5051189659) remains in on-heap JVM shuffle. A coalesced cached-sort consumer can retain Comet pages while waiting on a cache lock whose producer waits for Spark memory held by a Comet waiter. The task waiting on that cache lock has no recognized Spark-memory frame, so the guard stays false across every retry; BASE's managed allocation failure releases the Spark holder. The admitted plan needs a used computed Project between coalesce and keyed repartition.

The [healthy-holder case](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3869474061) also remains: a waiting bypass allocation can be rejected while its Comet holder is parked behind an independently progressing Spark task, whereas BASE can spill that holder and satisfy the request. A newly registered waiter can trigger this before a full wait interval. These are source-derived witnesses on the pinned Spark 3.5 sources, not executed reproductions.

peterxcli:
Addressed the [two follow-up witnesses](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5052924155) in d9bf677f0 — and I think together they settle the design question, so I changed approach rather than patching the detector again. Your cache-lock chain shows any frame- or state-based check can be evaded transitively (the blocked holder's stack shows the lock, not the Spark-memory dependency behind it), and your second witness shows the same check can misfire on a holder that an independently progressing task would have released. Holders' dependencies live outside this allocator's view, so no local detector classifies them reliably in either direction.

The frame detector is removed. In its place the wait now has an overall bound, `spark.comet.shuffle.jvm.memoryWaitTimeout` (internal, default 5 minutes): on timeout the waiter throws the managed `SparkOutOfMemoryError`, exactly as BASE did, so its task releases memory, the resource on the other side of any cycle can progress, and Spark's task retry can recover. The two precise fail-fast checks (self-retained, all-waiters) are unchanged, and the interval re-logging remains. The new-waiter `notifyAll` concern is moot with the detector gone — a waiter now only gives up at its own deadline, never because another waiter's arrival changed an instantaneous classification.

On base-relative behavior, stated plainly: with cross-task force-spill removed (the core of this PR), some schedules BASE completed by force-spilling another task's buffers can only ever complete here if the holder frees within the bound; the generous default covers every healthy-holder timing raised in this review (1.5 s, 65 s, independent finite releases), undetectable cycles now unwind within the bound instead of hanging the executor, and operators can tune the tradeoff. The cross-pool regression test now exercises the timeout (holder genuinely parked in `ExecutionMemoryPool`, waiter unwinds at a 1 s configured bound, pool drains fully once Spark memory frees).

sunchao (approved):
[P2 follow-up: addressed] Re-reviewed `d9bf677f05d05eba08ee6a9931ec85af1b0dc1db` against [the prior review](https://github.com/apache/datafusion-comet/pull/5493#pullrequestreview-5052924155). The Spark-stack false-rejection path is removed, and the configured finite timeout provides a managed escape from the former cache-cycle wait. No outstanding verified introduced P1/P2 remains. Slow holders can still time out; the passing component CI evidence is from the separate merge, not standalone HEAD.

andygrove (commented):
Thanks for the thorough iteration on this. The registry change is clearly correct and well motivated, and the thread with @sunchao has already worked the blocking allocator hard, so I have nothing to add on the deadlock-detection design. Below is only what I did not see raised already.

### `this.allocator` can be null in `write()`'s `finally` now that `open()` moved

This looks like a regression from moving `open()` out of the constructor. In `CometUnsafeShuffleWriter.java:276` the `finally` block does `nativeLib.logMemoryUsage(offheapMemKey, this.allocator.getUsed())` with no null guard. `allocator` is assigned inside `open()`, and `open()` is now only reached from inside the `try`. So if `records.hasNext()` throws before the first record arrives, which is exactly the window this PR is hardening, and `spark.comet.tracing.enabled` is on, the `finally` throws an NPE that masks the original failure. Line 253 is fine because it sits after `open()`. Could you guard line 276 with `this.allocator != null`?

### Does `spill()` guarantee the calling writer was flushed?

`ArrowIPCWriter.spill(int required)` sorts `currentWriters` by memory usage and breaks as soon as `totalFreed >= required`. `required` is a single serialized row plus `uaoSize`, so flushing the single largest sibling satisfies it right away. When that sibling is not `this`, control returns to `insertRow`, `acquireNewPageIfNecessary` has already returned false, and `initialCurrentPage` runs with `currentPage` still non-null, tripping its `assert (currentPage == null)`. `pom.xml:117` puts `-ea` on the test JVM, so that is an `AssertionError` rather than a silently wasted page tail.

That is the same symptom the description attributes to the cross-task registry, and scoping the list to the task does not remove it. It needs two or more partition writers in the task with the caller not being the largest, which is the normal bypass-merge shape.

The same gap also breaks the precondition the new blocking path documents. `SpillWriter.initialCurrentPage` now says "Only called after `spill()` has flushed this writer's data", and `allocateBlocking` says a waiting task "holds no memory that other tasks could be waiting for". When `spill()` breaks early the caller enters the wait still holding its own full page, which is precisely the state the fail-fast reasoning assumes away. Would freeing at least `Math.max(pageSize, required)`, or always spilling the calling writer, close both?

### Every new test uses exactly one writer per task list

Each `CometDiskBlockWriter` in `CometDiskBlockWriterSuite` is constructed with its own fresh `new JLinkedList[CometDiskBlockWriter]()`, and the first test gives `writersA` and `writersB` one writer each. So the sibling-spill loop, which is the whole reason the list exists, never runs with more than one entry, and that is also why the case above does not show up. It might be worth extending the first test so task A owns two or three writers sharing `writersA`. That would cover the intended sibling spilling and would also answer the `initialCurrentPage` question.

### The fatal-error path frees pages but leaves the temp files

The new `catch (Throwable)` in `CometBypassMergeSortShuffleWriter.write()` walks `partitionWriters` and calls `freeMemory()`. `stop(false)` does the same walk but also deletes each writer's temp file, and the premise of the new catch block is that `stop(false)` never runs for an `Error`. So on that path the memory is reclaimed and the temp shuffle files stay on disk. Since the loop is already there, could the `file.delete()` from `stop(false)` be added to it? Setting `partitionWriters = null` afterwards would also keep the `Exception` path from trying to delete them a second time and logging a spurious "Error while deleting file".

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
```diff
@@ -75,6 +98,8 @@ public final class CometBoundedShuffleMemoryAllocator extends CometShuffleMemory
     this.pageSize = pageSize;
     this.totalMemory =
         CometSparkSessionExtensions$.MODULE$.getCometShuffleMemorySize(conf, SQLConf.get());
+    this.memoryWaitTimeoutMs =
+        (long) CometConf$.MODULE$.COMET_SHUFFLE_JVM_MEMORY_WAIT_TIMEOUT().get();
```
sunchao: [P2] Read the wait timeout from the task that is waiting

The per-task spill registry is a useful improvement. Could you resolve this timeout once per `allocateBlocking` call, or explicitly enforce executor-startup scope? `getInstance()` keeps this allocator in a static `INSTANCE`, but this getter reads the initializing task's `SQLConf`. Later tasks can carry a new `spark.comet.shuffle.jvm.memoryWaitTimeout` through session settings and task properties, yet their deadline still uses the first value.

For example, on a reused executor with the JVM/on-heap bypass path enabled (`spark.comet.exec.onHeap.enabled=true` for plugin setup), initializing the pool under `1s` and then changing the session setting to `5m` leaves a later first-page wait using the one-second deadline. If another nonwaiting task holds the needed memory past that deadline, a request that fits the pool can raise OOM despite that task's longer configured wait. Reversing the values also ignores a requested short cap. A regression test reusing the same allocator across two task settings would cover this.

andygrove (commented):
> **Note on this review:** as with my last one, this was produced with an LLM (Claude Code) at my request. Unlike last time it also checked out both this branch and the merge base, built them, ran the suites, wrote extra tests, and benchmarked, so the claims below are backed by things that actually ran rather than by reading the diff. I have still not personally re-derived each one, so push back on anything that is wrong or already handled.

All four points from my last review look properly addressed, and the two I care most about are backed by evidence now rather than by argument. I reverted just the `ArrowIPCWriter.spill()` hunk on your head commit and reran `CometDiskBlockWriterSuite`, and 2 of the 9 tests fail with the same `AssertionError` the fix targets, so that regression test is real rather than vacuous. The first test owning three writers is exactly what was missing.

I also ran an A/B benchmark of the bypass writer against base `2949fd0d2`, driving 200k rows through a `TestMemoryManager` capped at a few 256 KiB pages. There is no regression anywhere I could measure. At 200 partitions against a 16 page pool the median wall time is 27.3 s on this branch against 28.1 s on base, and the spilled and output byte counts are identical to the byte. On a 4 partition shape with occasional 120 KiB rows this branch spills 546,665 bytes against base's 693,031, so the new "flush myself, then siblings only if I am still short" policy never spills more than the old one did. Worth putting some of that in the description, since a shuffle change without numbers invites the question.

### The `spill()` fix is doing more than the description credits it with

The description attributes the `initialCurrentPage` assertion to the cross-task registry, and that is only half the story. I built base at `2949fd0d2` and ran a single task with 8 partition writers against an 8 page pool with mixed row sizes, and it throws `java.lang.AssertionError` at `SpillWriter.initialCurrentPage:170` inside the first 20k rows. No second task, no concurrency, nothing exotic. The old loop broke as soon as `totalFreed >= required`, and `required` is one serialized row, so the largest sibling always satisfied it on its own and the caller kept its page. Your head commit runs the identical workload clean with `-ea` on. I had to pass `-da` to get base through it at all.

Could you call the `spill()` change out as its own fix? As written the description reads as though scoping the list was sufficient, and a future reader will believe that. It also argues for this PR more strongly than the description currently does, because it means you are fixing a crash an ordinary bypass-merge shuffle can reach.

### `stop(false)`'s cleanup loop is now unreachable

`CometBypassMergeSortShuffleWriter.java:405`. The new `catch (Throwable)` sets `partitionWriters = null`, and `writePartitionedData` already nulls it on the success path, so `write()` cannot leave the field non-null on any exit. That makes the `if (partitionWriters != null)` block in `stop(false)` dead, and it is a near duplicate of the new loop apart from the null-slot skip and the `file.exists()` check that only the new one has.

Would you pull it out into a `private void cleanupPartitionWriters(Throwable failure)` and call it from both places? That stops the two copies drifting, and it leaves `stop(false)` correct if some later change reintroduces a path that reaches it.

### The comments still describe cross-thread spilling

The new class comment says writers belong to one task, which is the whole point of the change, but five comments inside `CometDiskBlockWriter` still say the opposite. Line 184 has "could be triggered by other thread", line 217 has "other threads may be spilling this writer at the same time", line 306 has "could be spilled by other threads" and sits above code with no `synchronized` at all, line 341 has "All threads are writing to the same file" where each writer owns its own file, and line 357 has "Other threads may be updating the metrics at the same time".

Could you update or drop them here? The new invariant is the interesting thing about this PR, and right now the file argues against it in five places. If the `synchronized (currentWriters)` blocks are staying as belt and braces, one line saying so would be worth more than what is there today.

### The `allocateBlocking` contract does not match the check below it

`CometShuffleMemoryAllocatorTrait.java:36` and `CometBoundedShuffleMemoryAllocator.java:148` both say callers must only use this after spilling their own data "so that a waiting task holds no memory that other tasks could be waiting for". Neither caller satisfies that. `CometShuffleExternalSorter.spill()` calls `SpillSorter.reset()`, which re-allocates the pointer array, and `ArrowIPCWriter.spill()` stops as soon as `totalFreed >= required` so sibling writers in the same task keep their pages. The self-retained check at line 178 exists precisely because of that.

Could the doc say what is actually true, something closer to "callers must have spilled what they can cheaply release, and whatever they still retain is accounted for by the self-retained check"? As written it states a precondition the next few lines assume away, and someone could reasonably delete that check on the strength of it.

### The sort in `spill()` is usually redundant now

`CometDiskBlockWriter.java:386`. Now that the caller flushes itself first, `totalFreed >= required` is already true by the time the sort runs in the common case, since the caller's page is at least `pageSize` and `required` is a single row. The loop then breaks on its first iteration and the sort was pure overhead. At the default `bypassMergeThreshold` of 200 that is roughly 1500 comparisons per spill, each walking two `allocatedPages` lists. Wrapping the sort in `if (totalFreed < required)` would skip it. I benchmarked 200 partitions against a 16 page pool and could not measure a difference, so treat this as tidiness rather than something to fix.

### One more case for the first test

`CometDiskBlockWriterSuite.scala:137`. Three writers is the right shape now, but every row it inserts is 1 KiB, so `required` is always covered either by the caller's own flush or by the first sibling and the loop never runs past one iteration. That iteration count is exactly what `acb71a1bd` restructured when it moved the `break` to the top of the body. One row larger than any single writer's buffer would cover it. I ran three writers against a 6 page pool with every third row at 700 KiB against a 256 KiB page and it passes, so this is about locking the behaviour in rather than a bug I suspect.

### One loose end

@sunchao's [last comment](https://github.com/apache/datafusion-comet/pull/5493#discussion_r3900888272) about resolving the wait timeout per `allocateBlocking` call looks addressed in `acb71a1bd`, and the shared-pool wait test discriminates properly, since the allocator is initialised under a task carrying `100ms` and the wait then happens under a task carrying `5s` with the holder releasing at 500 ms. The thread has no reply though, so it is easy to miss.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
```diff
@@ -123,6 +143,114 @@ public synchronized MemoryBlock allocate(long required) {
...
+                    + "{} bytes free, {} thread(s) waiting",
+                now - waitStart,
+                size,
+                totalMemory - allocatedMemory,
+                waitingThreads.size());
+          }
+          try {
+            wait(
+                Math.max(
+                    1L, Math.min(WAIT_LOG_INTERVAL_MS, memoryWaitTimeoutMs - (now - waitStart))));
```
sunchao: [P2] Could the retry loop also observe cooperative task cancellation? In on-heap mode, a task requesting its first page can retain no pool memory while another job's uncancelled task holds the pool without waiting, so both liveness guards permit this wait. If the waiter is cancelled with `interruptOnCancel=false`, Spark sets its `TaskContext` flag without interrupting the Java thread. This new retry can keep the cancelled task occupying executor capacity and delay task-exit cleanup until memory is freed or the allocation deadline expires (five minutes from wait entry by default), unless a later Java interruption or external termination intervenes. Could we check the task flag here and cover this flag-only case with a focused regression test?

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Addressed in acb71a1bd: allocateBlocking now resolves spark.comet.shuffle.jvm.memoryWaitTimeout for each call. The shared-pool regression initializes the singleton under a 100 ms task setting, then waits under a 5 s task setting while the holder releases at 500 ms, so it discriminates against retaining the first task setting.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
peterxcli: Addressed in 59b7cf2b2: allocateBlocking now checks Spark TaskContext cancellation on each retry, polls at most once per second, and notifies remaining waiters when it exits. The regression marks TaskContext interrupted without interrupting the Java thread and requires prompt TaskKilledException; CometDiskBlockWriterSuite passes 9/9.

peterxcli:
Addressed the follow-up review in 59b7cf2b2: extracted the shared partition-writer cleanup; corrected the task-confined comments and allocator contract; skipped the sibling sort after a sufficient self-spill; and expanded the three-writer regression with a 900 KiB row that requires two sibling spills. I also updated the PR description with the independent single-task crash and benchmark figures. Local validation: make format, make core, and CometDiskBlockWriterSuite 9/9.

sunchao (approved):
Re-reviewed `59b7cf2b27765a7803e0db6526699189cff522c9`. The flag-only task-cancellation concern is addressed: allocation retries check `TaskContext`, and the `finally` path removes the waiter and wakes peers. No new P1/P2 findings in the updated allocator, spill and cleanup paths.

This was a source review; I did not rerun local tests or benchmarks. CI is still running.

spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java
```diff
@@ -123,6 +147,122 @@ public synchronized MemoryBlock allocate(long required) {
...
+   * and none of their requests fits in the free pool. Because the holders it depends on may in turn
+   * be blocked on resources outside this pool that only a task waiting here can release, the wait
+   * is also bounded by `spark.comet.shuffle.jvm.memoryWaitTimeout`, after which the managed
+   * allocation error is thrown and Spark's task retry can recover. Task cancellation or Java
+   * interruption aborts the wait.
+   */
+  @Override
+  public synchronized MemoryBlock allocateBlocking(long required) {
+    long memoryWaitTimeoutMs =
+        (long) CometConf$.MODULE$.COMET_SHUFFLE_JVM_MEMORY_WAIT_TIMEOUT().get();
```
sunchao: Could this lookup move into the first actual wait? `allocateBlocking` can succeed immediately after the caller spills, so that path does not need to resolve the timeout.

Across two JVM forks, the current successful allocation/free path measured 329–336 ns and 432 B of Java heap per pair. The prototype with a capacity check and deferred lookup measured 110–126 ns and 112 B. The capacity-check-only control retained roughly the current cost, which points to the eager configuration lookup as the avoidable work on this path. These are component measurements and the absolute cost is small per page.

Read the setting once per waiting request, from that task's configuration, so this preserves the per-task timeout fix already made in this PR.

sunchao (commented):
The follow-up text is posted as a regular PR comment [here](https://github.com/apache/datafusion-comet/pull/5493#issuecomment-5529155657).

sunchao:
**The overall design is sound, and the abstraction boundaries are appropriate. I would keep the design and make two focused improvements to the compatibility allocator.** I checked the unchanged head, `59b7cf2b`, with five independent review scopes and local measurements.

1. **Avoid constructing an exception on every unsuccessful allocation retry.**

   In [`allocateBlocking()`](https://github.com/apache/datafusion-comet/blob/59b7cf2b27765a7803e0db6526699189cff522c9/spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java#L177), each wakeup attempts allocation and constructs a new `SparkOutOfMemoryError` if capacity is still insufficient. Exception construction, formatting, and stack unwinding happen while holding the shared allocator lock.

   This is measurable. With approximately 1,000 small frees per second, blocked requests that still cannot fit produced:

   | Blocked tasks | Current code: waiter heap allocation | Capacity-check prototype |
   |---|---:|---:|
   | 8 | 29.46 MiB | 0.22 MiB |
   | 32 | 54.24 MiB | 0.49 MiB |

   Each measurement covers **1.2 seconds**. For eight waiters, combined waiter CPU also fell from approximately **546 ms to 367 ms**.

   I would introduce one private `tryAllocateMemoryBlock()` helper that returns a shortage indication. The immediate allocation API converts shortage into an exception; the blocking API runs its existing waiting checks. Construct the exception only when the operation must fail. This preserves cancellation, timeout, notifications, and deadlock checks while removing substantial temporary allocation.

2. **Read the timeout setting only when the call actually needs to wait.**

   The [current lookup](https://github.com/apache/datafusion-comet/blob/59b7cf2b27765a7803e0db6526699189cff522c9/spark/src/main/java/org/apache/spark/shuffle/comet/CometBoundedShuffleMemoryAllocator.java#L165) happens even when spilling has already freed enough memory.

   For a successful `allocateBlocking()` followed by `free()`, two independent JVM runs measured approximately:

   | Implementation | Time per pair | Java heap allocated |
   |---|---:|---:|
   | Current head | 329–336 ns | 432 bytes |
   | Capacity check with deferred timeout lookup | 110–126 ns | 112 bytes |

   This is a small absolute cost per page, but the change is straightforward. Keep the setting specific to the requesting call; moving it back into the singleton constructor would reintroduce the earlier configuration bug.

The other performance choices look reasonable:

- **Task-owned writer lists improve contention and reduce the spill search space.**
- **Caller-first spilling usually avoids sorting siblings entirely.** The remaining full sort repeatedly scans writers’ page lists, but that cost predates this PR and now occurs less often. A linear victim search could help particular workloads; the evidence does not justify a more elaborate priority queue.
- **Ownership accounting adds modest normal-path overhead:** roughly 45–65 ns and 72 bytes of temporary Java heap per allocation/free pair in my component test. It occurs per page, and only in the compatibility allocator. The usual off-heap allocator avoids this machinery.
- The row, file, and metrics synchronization is largely inherited. Its presence should not be counted as newly introduced overhead.

**The added state is mostly necessary complexity.** Page ownership, retained bytes, and waiting request sizes answer different questions. The separate immediate and blocking allocation APIs also matter: callers must first get a chance to spill their buffers. Constructor cleanup, write-failure cleanup, and task-completion cleanup cover different lifetime boundaries.

The alternatives have clear tradeoffs:

| Alternative | Assessment |
|---|---|
| Current design with a nonthrowing allocation helper | Best fit here: simpler retry logic without losing safeguards. |
| Timeout-only waiting | Less code, but impossible requests can occupy task slots for five minutes. |
| Fixed per-task memory budgets | Simpler accounting, but unused capacity becomes stranded and uneven tasks spill more. |
| Fair allocation queue | Could improve fairness under contention, but adds scheduling and mixed-request-size handling. It still needs retained-memory safeguards. |
| Use Spark’s memory manager for every mode | Worth considering as broader architecture work; it changes how the compatibility mode’s separate off-heap budget is accounted for. |

The local prototype passed **14/14 focused allocator and writer regression tests**. These measurements establish allocator-level overhead on Spark 4.1.3/JDK 17; they do not establish an end-to-end query speedup. The PR’s reported spill benchmarks cover a different part of the change.

sunchao:
Thanks @peterxcli for the contribution! The above are some minor performance related follow-ups that Codex suggested. Please take a look and see if they make sense.

PR #5494, opened 2026-08-27
fix: make task-shared memory pool as ref-counted RAII guard

peterxcli:

Part of #5212 (positions 2, 3 and 8). Supersedes the closed draft #5217, replacing its explicit refcount guard with plain `Arc` lifetime.

## Rationale for this change

Entries in `TASK_SHARED_MEMORY_POOLS` leak whenever a plan's lifecycle does not run to completion, and a stranded entry is never reclaimed: keys are unique task attempt ids and nothing prunes the map. For the unified pool types each entry transitively holds a JNI global ref to the task's `CometTaskMemoryManager`, which pins its `TaskMemoryManager` and `TaskContext` for the lifetime of the executor.

There were two leak paths, plus an unsafe close path:

1. **`createPlan` failing after the pool was registered.** The pool is registered early in `createPlan`, but later steps can still fail — building the DataFusion session context (e.g. an invalid `spark.comet.datafusion.*` config), the key-unwrapper global ref, and the `TaskContext`/`ClassLoader` global refs. On the JVM side `plan` is a field initializer evaluated *before* the task-completion listener is registered, so a failed `createPlan` means `releasePlan` is never called.

2. **`releasePlan` failing before the release.** It ran `update_metrics(...)?` *before* the pool release and the `Box::from_raw`, so a metrics failure stranded both the registry entry and the native `ExecutionContext`.

3. **`CometExecIterator.close()` was not idempotent and could skip or repeat `releasePlan`.** `closed = true` was set last, so a throw from `currentBatch.close()`, `nativeUtil.close()` or a shuffle block iterator skipped `releasePlan` entirely (stranding the context), and the task-completion listener's retry re-entered the whole teardown.

I verified all three paths empirically against unmodified `main` with an instrumented build and deterministic failure injections (a bogus `spark.comet.datafusion.*` key to fail `createPlan` after registration, a metrics node that throws to fail `releasePlan`, and a shuffle block iterator whose `close()` throws):

| Injected failure (on unmodified main) | Observed |
|---|---|
| 10 × `createPlan` failure | registry grew 1→10, zero releases; 10/10 `CometTaskMemoryManager`s still pinned after 30 s of GC |
| teardown throw in `close()` | `releasePlan` never called; native context and pool entry stranded |
| metrics failure in `releasePlan` | `releasePlan` called **twice** (listener retry), context freed **zero** times |

With this change the same three injections leave the registry empty, free every native context exactly once, and the pinned managers become collectible.

**Tie the registry entry to the pool's own lifetime.** `TASK_SHARED_MEMORY_POOLS` now maps task attempt id → `Weak<TaskSharedMemoryPool>`, where `TaskSharedMemoryPool` is a transparent `MemoryPool` wrapper whose `Drop` removes its own entry. The `Arc` returned by `create_memory_pool` is the RAII handle: the pool stays registered exactly as long as the plan's session context (or any reservation) holds it, and both failure paths above are covered by construction — `createPlan` unwinding drops the `Arc`, and `releasePlan` dropping the `ExecutionContext` drops it too. A `ptr_eq` check in `Drop` prevents an old pool's drop from removing a racing replacement's entry. Using `Weak` in the map means the map itself never keeps a pool alive.

**`releasePlan` reclaims the `Box` up front and flushes metrics last**, so the context (and with it the pool reference and every JNI global ref) is freed even when the metrics update fails; the metrics error is still thrown.

**`CometExecIterator.close()` sets `closed` first and always calls `releasePlan`.** Teardown runs in a `try`/`catch`; the release runs unconditionally afterwards, with the teardown exception propagated and any release failure attached as a suppressed exception. This has to land together with the `releasePlan` change: now that the `Box` is always freed, a second `releasePlan` on the same pointer would be a use-after-free rather than a leak.

**Deletions that fall out of the above:** `MemoryPoolType::is_task_shared`, `handle_task_shared_pool_release`, the per-plan refcount (`PerTaskMemoryPool.num_plans`), and the `ExecutionContext::task_attempt_id` / `memory_pool_config` fields, all of which existed only to feed the old explicit release call. The duplicated task-shared arms of `create_memory_pool` collapse behind an `acquire_task_shared_pool` helper. The registry moves to `parking_lot::Mutex`, matching `fair_pool.rs` in the same module.

**New `CometExecIteratorLifecycleSuite`** with one deterministic reproducer per failure path, exercising the real JNI lifecycle:

- `createPlan` failure: 10 plans created with an unknown `spark.comet.datafusion.*` key (fails inside `createPlan` after pool registration) against `fair_unified` pools; a `WeakReference` per `CometTaskMemoryManager` plus a GC loop asserts none stay pinned. The test asserts the injected error message so a missing native library cannot make it pass vacuously.
- teardown throw: a `CometShuffleBlockIterator` stub whose `close()` throws; asserts the exception propagates and a second `close()` is a no-op.
- metrics failure: a `CometMetricNode` subclass that throws in `set_all_from_bytes`, armed after a real plan executed; asserts the failure propagates and a second `close()` is a no-op.

On unmodified `main` all three fail deterministically (first with "10 of 10 CometTaskMemoryManagers stayed reachable", the other two with a second `close()` re-throwing); with this change all three pass.

**Unit tests** in `task_shared.rs` cover the registry: plans in one task share a pool, different tasks do not, the entry survives until the last reference drops, dropping the `Arc` alone releases it (the `createPlan`-unwind case), and an old pool's drop does not remove a racing replacement's entry.

**Regression suites:** `CometAggregateSuite`, `CometNativeShuffleSuite`, `CometNativeSuite`, `CometTaskMemoryManagerSuite` and the new suite — 125 tests, all passing. Native: full `datafusion-comet` crate suite (167 passed), `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, `spotless:check` and `scalastyle:check` clean.

No API or configuration changes. Long-running executors no longer accumulate memory pool registry entries and pinned JVM task objects when `createPlan`, the final metrics update, or iterator teardown fails, and a teardown failure no longer risks a double release of the native plan.

One behavioral nuance: the registry entry now lives until the last `Arc` to the pool drops rather than until the last plan's `releasePlan` call, so a reservation that outlives its plan keeps the pool registered instead of the entry being removed while the pool is still in use.

### DIFF
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -986,6 +986,14 @@ pub extern "system" fn Java_org_apache_comet_Native_releasePlan(
     exec_context: jlong,
 ) {
     try_unwrap_or_throw(&e, |env| unsafe {
+        // A null pointer from the JVM would be undefined behaviour in `Box::from_raw`; panic
+        // instead, which `try_unwrap_or_throw` converts into a Java exception (this is the check
+        // `get_execution_context` performs for the other JNI entry points).
+        assert_ne!(
+            exec_context, 0,
+            "Comet execution context shouldn't be null!"
+        );
+
         // Reclaim ownership of the context up front so that it is always freed, even if updating
         // metrics below fails. Dropping it releases the memory pool and every JNI global ref the
         // context holds.
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -96,9 +96,7 @@ use std::{sync::Arc, task::Poll};
 use tokio::runtime::{Handle, Runtime};
 use tokio::sync::mpsc;
 
-use crate::execution::memory_pools::{
-    create_memory_pool, handle_task_shared_pool_release, parse_memory_pool_config, MemoryPoolConfig,
-};
+use crate::execution::memory_pools::{create_memory_pool, parse_memory_pool_config};
 use crate::execution::operators::{ScanExec, ShuffleScanExec};
 use crate::execution::shuffle::{read_ipc_compressed, CompressionCodec};
 use crate::execution::spark_plan::SparkPlan;
@@ -306,8 +304,6 @@ fn collect_op_names<'a>(op: &'a Operator, names: &mut std::collections::BTreeSet
 struct ExecutionContext {
     /// The id of the execution context.
     pub id: i64,
-    /// Task attempt id
-    pub task_attempt_id: i64,
     /// The deserialized Spark plan
     pub spark_plan: Operator,
     /// The number of partitions
@@ -340,8 +336,6 @@ struct ExecutionContext {
     pub debug_native: bool,
     /// Whether to write native plans with metrics to stdout
     pub explain_native: bool,
-    /// Memory pool config
-    pub memory_pool_config: MemoryPoolConfig,
     /// Whether to log memory usage on each call to execute_plan
     pub tracing_enabled: bool,
     /// Rust thread ID, used for aggregating tracing metrics per thread
@@ -528,7 +522,6 @@ pub unsafe extern "system" fn Java_org_apache_comet_Native_createPlan(
 
             let exec_context = Box::new(ExecutionContext {
                 id,
-                task_attempt_id,
                 spark_plan,
                 partition_count: partition_count as usize,
                 root_op: None,
@@ -545,7 +538,6 @@ pub unsafe extern "system" fn Java_org_apache_comet_Native_createPlan(
                 session_ctx: session,
                 debug_native,
                 explain_native,
-                memory_pool_config,
                 tracing_enabled,
                 rust_thread_id,
                 tracing_memory_metric_name: format!(
@@ -964,15 +956,11 @@ pub extern "system" fn Java_org_apache_comet_Native_releasePlan(
     exec_context: jlong,
 ) {
     try_unwrap_or_throw(&e, |env| unsafe {
-        let execution_context = get_execution_context(exec_context);
-
-        // Update metrics
-        update_metrics(env, execution_context)?;
-
-        handle_task_shared_pool_release(
-            execution_context.memory_pool_config.pool_type,
-            execution_context.task_attempt_id,
-        );
+        // Reclaim ownership of the context up front so that it is always freed, even if updating
+        // metrics below fails. Dropping it releases the memory pool and every JNI global ref the
+        // context holds.
+        let mut execution_context: Box<ExecutionContext> =
+            Box::from_raw(exec_context as *mut ExecutionContext);
 
         // Unregister this context's pool and emit the remaining total for the thread
         if execution_context.tracing_enabled {
@@ -984,8 +972,8 @@ pub extern "system" fn Java_org_apache_comet_Native_releasePlan(
             );
         }
 
-        let _: Box<ExecutionContext> = Box::from_raw(execution_context);
-        Ok(())
+        // Flush metrics last, as it is the only fallible step here.
+        update_metrics(env, &mut execution_context)
     })
 }
 
--- a/native/core/src/execution/memory_pools/mod.rs
+++ b/native/core/src/execution/memory_pools/mod.rs
@@ -52,34 +52,27 @@ pub(crate) fn create_memory_pool(
         ))
     }
 
-    fn task_shared(
-        task_attempt_id: i64,
-        create: impl FnOnce() -> Arc<dyn MemoryPool>,
-    ) -> Arc<dyn MemoryPool> {
-        acquire_task_shared_pool(task_attempt_id, create)
-    }
-
     let pool_type = memory_pool_config.pool_type;
     let pool_size = memory_pool_config.pool_size;
 
     match pool_type {
-        MemoryPoolType::GreedyUnified => task_shared(task_attempt_id, || {
+        MemoryPoolType::GreedyUnified => acquire_task_shared_pool(task_attempt_id, || {
             tracked(CometUnifiedMemoryPool::new(
                 comet_task_memory_manager,
                 task_attempt_id,
             ))
         }),
-        MemoryPoolType::FairUnified => task_shared(task_attempt_id, || {
+        MemoryPoolType::FairUnified => acquire_task_shared_pool(task_attempt_id, || {
             tracked(CometFairMemoryPool::new(
                 comet_task_memory_manager,
                 pool_size,
             ))
         }),
-        MemoryPoolType::GreedyTaskShared => task_shared(task_attempt_id, || {
+        MemoryPoolType::GreedyTaskShared => acquire_task_shared_pool(task_attempt_id, || {
             tracked(GreedyMemoryPool::new(pool_size))
         }),
         MemoryPoolType::FairSpillTaskShared => {
-            task_shared(task_attempt_id, || tracked(FairSpillPool::new(pool_size)))
+            acquire_task_shared_pool(task_attempt_id, || tracked(FairSpillPool::new(pool_size)))
         }
         MemoryPoolType::Greedy => tracked(GreedyMemoryPool::new(pool_size)),
         MemoryPoolType::FairSpill => tracked(FairSpillPool::new(pool_size)),
--- a/native/core/src/execution/memory_pools/mod.rs
+++ b/native/core/src/execution/memory_pools/mod.rs
@@ -34,93 +34,66 @@ use unified_pool::CometUnifiedMemoryPool;
 pub(crate) use config::*;
 pub(crate) use task_shared::*;
 
+/// Creates the memory pool for a native plan.
+///
+/// Task-shared pools use their returned `Arc` as the RAII handle, so they remain registered for as
+/// long as the plan or any of its reservations retain the pool.
 pub(crate) fn create_memory_pool(
     memory_pool_config: &MemoryPoolConfig,
     comet_task_memory_manager: Arc<Global<JObject<'static>>>,
     task_attempt_id: i64,
 ) -> Arc<dyn MemoryPool> {
     const NUM_TRACKED_CONSUMERS: usize = 10;
-    match memory_pool_config.pool_type {
-        MemoryPoolType::GreedyUnified => {
-            let mut memory_pool_map = TASK_SHARED_MEMORY_POOLS.lock().unwrap();
-            let per_task_memory_pool =
-                memory_pool_map.entry(task_attempt_id).or_insert_with(|| {
-                    let pool: Arc<dyn MemoryPool> = Arc::new(TrackConsumersPool::new(
-                        CometUnifiedMemoryPool::new(
-                            Arc::clone(&comet_task_memory_manager),
-                            task_attempt_id,
-                        ),
-                        NonZeroUsize::new(NUM_TRACKED_CONSUMERS).unwrap(),
-                    ));
-                    PerTaskMemoryPool::new(pool)
-                });
-            per_task_memory_pool.num_plans += 1;
-            Arc::clone(&per_task_memory_pool.memory_pool)
-        }
-        MemoryPoolType::FairUnified => {
-            let mut memory_pool_map = TASK_SHARED_MEMORY_POOLS.lock().unwrap();
-            let per_task_memory_pool =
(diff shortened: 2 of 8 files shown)


sunchao (commented):
Reviewed this head with five independent read-only reviewers. One verified P2 remains in exceptional JVM shuffle-input cleanup, reported inline.

Validation on chao-reviews-1 used isolated non-worktree source/build directories, a separate Cargo target and Maven repository, and JDK 17: `make core`; 5 task-shared pool tests; 130 selected Spark 4.0.4 JVM tests; 4 additional native ownership/concurrency probes; and 14 JNI lifecycle probes with `-Xcheck:jni`. The JNI probes checked creation failures, shared-pool retention, an actually executed native result row, exact exception/suppression behavior and collectability after release.

The inline issue has a real-file-descriptor control: head fails the cleanup assertion, while replacing only the iterator class with the exact base version passes; normal cleanup passes both. This is a focused lifecycle control, not a full base build or an end-to-end shuffle-fetch failure test. No full-suite claim. The inspected execution CI jobs pass; the current Spark 4.1 SQL-job failure is an HTTP 429 downloading Maven during setup, before tests.

spark/src/main/scala/org/apache/comet/CometExecIterator.scala
```diff
@@ -227,13 +227,33 @@ class CometExecIterator(
...
+      closed = true
+
+      var failure: Throwable = null
+      try {
+        if (currentBatch != null) {
+          currentBatch.close()
+          currentBatch = null
+        }
+        nativeUtil.close()
+        shuffleBlockIterators.values.foreach(_.close())
```
sunchao: [P2] Attempt cleanup of every shuffle input after a close failure

If an earlier shuffle input's `close()` throws, this `foreach` skips all later inputs, and the new `closed = true` makes the task-completion retry a no-op. This matters when an iterator closes before its raw shuffle inputs are drained: `SequenceInputStream.close()` can fail while fetching/closing remaining blocks, while another input already owns an open file stream. Spark's completion cleanup only calls `FileSegmentManagedBuffer.release()` (a no-op for its `FileInputStream`), and dropping the native plan does not invoke Java stream close, so the descriptor stays open until GC.

I reproduced this on Spark 4.0.4/JDK 17 with two real `CometShuffleBlockIterator`s, `SequenceInputStream`s and a file-backed managed buffer: after the first close throws and task completion runs, the later descriptor remains open at this head. With only the iterator class replaced by the exact base version, task completion closes it; normal cleanup passes both controls. Please attempt each resource's cleanup independently, accumulating exceptions while still releasing the native plan exactly once.

peterxcli: Good catch, thanks — fixed in a29ddff20. `close()` now attempts each resource's cleanup independently (`currentBatch`, `nativeUtil`, every shuffle input, then `releasePlan`), recording the first failure and attaching later ones as suppressed exceptions, so an early throw no longer skips the remaining inputs and the native plan is still released exactly once. Extended the teardown regression test with a second shuffle input that asserts it is closed after the first input's `close()` throws.

sunchao (approved):
Re-reviewed a29ddff20, including the full diff and complete increment from 2ba46fba, with five fresh independent reviewers. The [previous P2](https://github.com/apache/datafusion-comet/pull/5494#discussion_r3869989678) is fixed: fresh real-descriptor head/prior/base iterator controls confirm the later input closes on the first attempt at this head.

Fresh Spark 4.0.4/JDK 17 validation passed 38 selected repository tests, 14 JNI probes and 35 cleanup-dispatch cases, including ordered suppression of distinct exceptions and one native release. Five task-shared tests and four additional native probes also passed. The historical controls replace only the iterator class; no full-suite or end-to-end shuffle-fetch-failure claim. CI is still running.

No outstanding verified P1/P2 at this head.

native/core/src/execution/jni_api.rs
```diff
@@ -994,31 +986,23 @@ pub extern "system" fn Java_org_apache_comet_Native_releasePlan(
...
-
-        handle_task_shared_pool_release(
-            execution_context.memory_pool_config.pool_type,
-            execution_context.task_attempt_id,
-        );
+        // Reclaim ownership of the context up front so that it is always freed, even if updating
+        // metrics below fails. Dropping it releases the memory pool and every JNI global ref the
+        // context holds.
+        let mut execution_context: Box<ExecutionContext> =
+            Box::from_raw(exec_context as *mut ExecutionContext);
```
andygrove, line 1001: Reclaiming the `Box` up front is the right call, but it drops the null check that `get_execution_context` was doing with its `.expect("Comet execution context shouldn't be null!")`. A null or stale `jlong` coming from the JVM is now undefined behaviour rather than a panic that `try_unwrap_or_throw` converts into a Java exception.

I do not think it is reachable today, since `plan` is a `val` that only a successful `createPlan` can set. But this is a JNI entry point taking a raw pointer from Java, and the PR itself points out that a second `releasePlan` on the same pointer is now a use-after-free rather than a leak. Could you keep an assertion on the pointer before reclaiming it?

spark/src/main/scala/org/apache/comet/CometExecIterator.scala
```diff
@@ -227,13 +227,39 @@ class CometExecIterator(
...
+      }
+      attempt(nativeUtil.close())
+      shuffleBlockIterators.values.foreach(it => attempt(it.close()))
+
+      // Released last and exactly once, even if the teardown above failed: dropping the native
+      // execution context frees this plan's task-shared memory pool reference and several JNI
+      // global refs.
+      attempt(nativeLib.releasePlan(plan))
+
+      if (failure != null) {
```
andygrove: Throwing at this point means a teardown failure also loses the `memInUse != 0` warning further down, and that is exactly the diagnostic that would tell you whether the failed teardown stranded native memory.

Could the tracing call and the `memInUse` check go inside `attempt` blocks as well, with the throw moved to the very end of the method? Position 8 stays fixed either way, because `closed` is already set by then, so a throw out of `traceMemoryUsage()` can no longer trigger a listener retry.

spark/src/test/scala/org/apache/spark/CometExecIteratorLifecycleSuite.scala
```diff
@@ -0,0 +1,213 @@
...
+        new ThrowingMetricNode,
+        1,
+        0,
+        None,
+        Seq.empty)
+
+      failMetrics.set(true)
+      // Exhausting the iterator closes it, and the close propagates the metrics failure thrown
+      // by the native releasePlan call.
+      intercept[Throwable](iter.hasNext)
```
andygrove: The `createPlan` test is careful to assert the injected message so an unrelated failure cannot make it pass vacuously, but this one accepts any `Throwable`. That matters more here than usual, because `executePlan` also calls `update_metrics` on the periodic interval, so the throw could come from mid-execution rather than from `releasePlan` and the test would still pass without covering the path it is named after.

Could you assert the message contains `injected metrics update failure`, and ideally arm `failMetrics` only once the plan has produced its final batch, so the failure is pinned to the release path?

native/core/src/execution/memory_pools/mod.rs
```diff
@@ -34,94 +34,67 @@ use unified_pool::CometUnifiedMemoryPool;
...
+
+    fn tracked(pool: impl MemoryPool + 'static) -> Arc<dyn MemoryPool> {
+        Arc::new(TrackConsumersPool::new(
+            pool,
             NonZeroUsize::new(NUM_TRACKED_CONSUMERS).unwrap(),
-        )),
+        ))
+    }
+
+    fn task_shared(
```
andygrove: This forwards its arguments unchanged, so it is only a second name for `acquire_task_shared_pool`. Calling that directly in the four arms below would make it obvious at a glance which arms touch the registry. `tracked` is pulling real weight, so I would keep that one.

native/core/src/execution/memory_pools/task_shared.rs
```diff
@@ -15,46 +15,174 @@
...
+            let _pool = acquire(-1005);
+            assert!(is_registered(-1005));
         }
+        assert!(!is_registered(-1005));
+    }
+
+    #[test]
+    fn an_old_pool_does_not_remove_its_replacement() {
+        let old_pool = acquire(-1006);
+        TASK_SHARED_MEMORY_POOLS.lock().remove(&-1006);
```
andygrove, line 179: Removing the entry by hand does exercise the `ptr_eq` guard, but it does not show that the real code can reach that state. If the guard ever regresses the failure is silent and unpleasant: the old pool's `Drop` evicts a live replacement, so later plans in the same task each build their own pool and the per-task limit ends up enforced more than once.

Could you add a short concurrent churn test alongside this one? Something like a handful of threads each acquiring and immediately dropping the same task attempt id in a loop, then asserting the map is empty once they quiesce and that a fresh acquire comes back registered.

andygrove (commented):
Thanks for this, and for the failure-injection table in the description. That table is the part that makes the change easy to trust.

Tying the registry entry to the pool's own `Arc` is the right shape. I went through the failure paths independently and every resource `createPlan` acquires before the failure point is already RAII, so this genuinely closes the leak rather than relocating it, and position 3's `usize` underflow disappears along with the field it lived on. The `Drop`/acquire race checks out too. `Arc::drop` zeroes the strong count before running `Drop::drop`, so `Weak::upgrade` can never hand out a pool that is mid-drop, and the allocation cannot be reused while `Drop::drop` is still on the stack, so `ptr_eq` cannot hit an ABA. I also checked that nothing in Comet or in DataFusion 54.1.0 downcasts `dyn MemoryPool` outside DataFusion's own tests, so the extra wrapper layer is safe, and it follows the pattern `LoggingMemoryPool` already established.

Five things I would like addressed, left inline. The one I care most about is the null assertion that `releasePlan` lost.

native/core/src/execution/jni_api.rs
peterxcli: Restored in 9ea52ffa6 as an `assert_ne!(exec_context, 0, "Comet execution context shouldn't be null!")` before the `Box::from_raw`, with the same message `get_execution_context` uses — `try_unwrap_or_throw` converts the panic into a Java exception as before.

spark/src/main/scala/org/apache/comet/CometExecIterator.scala
peterxcli: Done in 9ea52ffa6 — the tracing call and the `memInUse` check now run inside `attempt` blocks and the throw moved to the very end of `close()`, so a teardown failure still emits the non-zero-memory warning (and any diagnostic failure is attached as suppressed).

spark/src/test/scala/org/apache/spark/CometExecIteratorLifecycleSuite.scala
peterxcli: Done in 9ea52ffa6 — the test now runs under `spark.comet.metrics.updateInterval=0`, which disables the periodic updates inside `executePlan`, so the releasePlan update is the only place the stub can fire; and it asserts the thrown message contains `injected metrics update failure` so an unrelated throw can't make it pass vacuously.

native/core/src/execution/memory_pools/mod.rs
peterxcli: Done in 9ea52ffa6 — dropped the pass-through and the four arms call `acquire_task_shared_pool` directly; `tracked` stays.

native/core/src/execution/memory_pools/task_shared.rs
peterxcli: Added in 9ea52ffa6: `concurrent_acquire_and_drop_leaves_a_consistent_registry` — 8 threads each acquire-and-drop the same task attempt id 1000 times, then the test asserts the registry entry is gone once they quiesce and that a fresh acquire comes back registered. The hand-removal test stays as the deterministic unit-level check of the `ptr_eq` guard.

sunchao (approved):
Re-reviewed `9ea52ffa68992bc5fe799423b07928ab18356df8` with five independent reviewers; no outstanding verified P1/P2 findings remain. Current CI passes the registry and lifecycle regressions, including the prior cleanup case. The reviewed regions were verified identical in CI's merge build; this is not a standalone exact-head build or full-suite claim. No fresh local tests were run.

sunchao:
Merged, thanks @peterxcli !

PR #5497, opened 2026-08-27
fix: report native operator spill metrics in Spark task metrics for non-shuffle stages

### LINKED ISSUE #5447: Report native operator spill metrics in Spark task metrics for non-shuffle stages
## What / Why

#5370 and #5445 bridge native spill metrics into Spark task metrics (`diskBytesSpilled` / `memoryBytesSpilled`), but only for native shuffle write tasks: `CometMetricNode.reportSpillMetrics` is registered solely on the `CometNativeShuffleInputRDD` path.

A spill-capable native operator running in a non-shuffle stage (e.g. a sort in a result stage, executed via `CometExecRDD` in `operators.scala`) still reports `spilled_bytes` only as a SQL operator metric. The Spark Stages/task view shows zero spill for the task, the same SQL-view vs task-view discrepancy that #5382 fixed for unified shuffle plans.

## Proposed scope

- Register a task-level spill reporting callback for native stages executed via `CometExecRDD`, aggregating over the stage's `CometMetricNode` tree as `sumMetricValues` does for shuffle tasks (shared accumulators counted once, disk and memory kept separate, no memory value inferred from disk bytes).
- Register it before the native iterator so final metrics are published first, including for failed attempts, matching the ordering established in #5445.

peterxcli:

Closes #5447.

## Rationale for this change

#5370 and #5445 bridge native spill metrics into Spark task metrics (`diskBytesSpilled` / `memoryBytesSpilled`), but only for native shuffle write tasks. A spill-capable native operator running in a non-shuffle stage (e.g. a sort in a result stage executed via `CometExecRDD`) still reports `spilled_bytes` only as a SQL operator metric, so the Spark Stages/task view shows zero spill for the task — the same SQL-view vs task-view discrepancy that #5382 fixed for unified shuffle plans.

- `CometExecRDD.compute` now registers `CometMetricNode.reportSpillMetrics` for the stage's metric tree, before input producers are resolved and before the native iterator is created, so reverse-ordered task completion listeners publish final native metrics first, including for failed attempts (matching the ordering established in #5445). Disk and memory totals stay separate and no memory value is inferred from disk bytes.
- Several native executions can run inside one Spark task with overlapping metric trees: `CometMetricNode.fromCometPlan` descends through union/coalesce input boundaries into nested native blocks, and a coalesced partition computes the same tree once per parent partition. All `reportSpillMetrics` registrations in a task (including the existing native shuffle writer path) now claim accumulators from a shared per-task identity seen-set keyed by task attempt id (one set per metric name, cleaned up on task completion), so each spill accumulator is counted exactly once per task while disjoint trees still all report.

New tests in `CometTaskMetricsSuite`:

- A unit test registering overlapping, repeated, and disjoint metric trees on one `TaskContext` (for both successful and failed attempts), with final metric values published by a completion listener to verify listener ordering, and the per-task seen-set discarded after completion.
- An end-to-end test that a native sort spilling in a shuffle-free result stage reports its `spilled_bytes` as the stage's `diskBytesSpilled`, with `memoryBytesSpilled` staying zero.
- An end-to-end test that a failed result-stage attempt (late-row ANSI division failure above a spilling sort) still preserves the disk spill metrics.

Existing spill-metric tests in `CometTaskMetricsSuite` and `CometNativeShuffleInputRDDSuite` continue to pass.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/comet/CometExecRDD.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/CometExecRDD.scala
@@ -98,6 +98,11 @@ private[spark] class CometExecRDD(
   }
 
   override def compute(split: Partition, context: TaskContext): Iterator[ColumnarBatch] = {
+    // Must precede resolveInputObjects and the CometExecIterator: completion listeners run in
+    // reverse registration order, so registering first means this listener runs last, after
+    // nested native blocks and the iterator have published their final metric values.
+    Option(context).foreach(nativeMetrics.reportSpillMetrics)
+
     val partition = split.asInstanceOf[CometExecPartition]
 
     val (inputObjects, shuffleBlockIters) =
--- a/spark/src/main/scala/org/apache/spark/sql/comet/CometMetricNode.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/CometMetricNode.scala
@@ -20,6 +20,7 @@
 package org.apache.spark.sql.comet
 
 import java.util.IdentityHashMap
+import java.util.concurrent.ConcurrentHashMap
 
 import scala.jdk.CollectionConverters._
 
@@ -63,9 +64,17 @@ case class CometMetricNode(metrics: Map[String, SQLMetric], children: Seq[CometM
     else children.flatMap(_.leafNodes)
   }
 
-  private[comet] def sumMetricValues(metricName: String): Long = {
-    val seenMetrics = new IdentityHashMap[SQLMetric, java.lang.Boolean]()
+  /**
+   * Sums `metricName` across this metric tree. A shared accumulator reachable through multiple
+   * paths in the tree contributes only once, and unset size metrics (initialized to -1) count as
+   * zero.
+   */
+  private[comet] def sumMetricValues(metricName: String): Long =
+    sumMetricValues(metricName, new IdentityHashMap[SQLMetric, java.lang.Boolean]())
 
+  private def sumMetricValues(
+      metricName: String,
+      seenMetrics: IdentityHashMap[SQLMetric, java.lang.Boolean]): Long = {
     def sumFromNode(metricNode: CometMetricNode): Long = {
       val nodeValue = metricNode.metrics.get(metricName).fold(0L) { metric =>
         if (seenMetrics.put(metric, java.lang.Boolean.TRUE) == null) {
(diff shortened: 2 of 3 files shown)


sunchao (approved):
## Summary

This review covers head `6d02c87a5b5deae93b2c0099c787d0f1f351f213` against base `525c05b52bb73bb873f05c32e30987132676bf5e`. The three-file change extends native spill reporting to non-shuffle tasks and coordinates reporters sharing SQL accumulators. All five independent reviewers completed the full diff and their specialized checks. No verified P1/P2 findings remain; the recommendation is approval at these pins.

## Prior state and problem

Previously, the task-metric bridge was registered only on the native shuffle input path. A spilling native sort in a result stage could therefore show spill in SQL operator metrics while Spark’s corresponding task/stage metrics remained zero. Adding reporters requires deduplication because input boundaries and coalesced partitions can make several metric trees reference the same SQL accumulator.

## Design approach

[`CometExecRDD.compute`](https://github.com/apache/datafusion-comet/blob/6d02c87a5b5deae93b2c0099c787d0f1f351f213/spark/src/main/scala/org/apache/spark/sql/comet/CometExecRDD.scala#L100-L104) registers its reporter before resolving input producers and creating the native iterator. Spark’s reverse completion-listener order lets iterators publish final native values before the bridge reads them. Shared identity sets, scoped by task attempt, then count each disk or memory accumulator once while preserving contributions from disjoint trees.

[CI run 33089261189](https://github.com/apache/datafusion-comet/actions/runs/33089261189) checked out `657a7959505a82a3e0622f5adb9117b4ecf06f6d`, whose tree `f21a07f83f4ffc899208d5d617c98abb4334d77a` is identical to the reviewed head. Native artifact `9654075656` was produced and downloaded by JVM consumers with matching upload/download digests.

All three new tests explicitly passed on Spark 3.4, 3.5, 4.0, 4.1 and 4.2. Corresponding execution suites reported 655/664/695/696/670 passes, 7/5/4/1/8 cancellations and three ignored tests each, with no failures. Existing native-shuffle metric checks also passed; Rust reported 976 passes and four skips. Canceled, ignored and skipped tests are not passing coverage. Broader CI remained in progress at the review handoff. No local builds or runtime probes were performed.

## Key design decisions

Deduplication uses `SQLMetric` object identity, rather than metric-tree identity, names or equal numeric values. Separate seen sets preserve the distinction between disk bytes written and memory bytes released; neither quantity is inferred from the other. Values are read at completion, unset negative values contribute zero, and task-scoped registry entries are removed after reporting.

## Implementation sketch

The implementation is confined to registration in `CometExecRDD`, aggregation and registry handling in `CometMetricNode`, and additions to `CometTaskMetricsSuite`. Native metric producers are unchanged. The [new tests](https://github.com/apache/datafusion-comet/blob/6d02c87a5b5deae93b2c0099c787d0f1f351f213/spark/src/test/scala/org/apache/spark/sql/comet/CometTaskMetricsSuite.scala#L88-L284) combine synthetic overlapping/repeated/disjoint trees with successful and failed native result-stage execution, covering both deduplication and completion ordering.

Non-shuffle native execution now contributes its published spill counters to task metrics. The successful sort test asserts native execution without a shuffle exchange and checks stage/SQL disk-byte equality. The failure test asserts native sort/project execution and retained positive spill bytes after an ANSI error.

This does **not** fix cumulative physical spill accounting across repeated coalesced executions. Existing absolute SQL accumulator updates can retain only the last execution’s value; this bridge transfers that final SQL value once, as specified.

## Suggested improvements

As optional follow-up work, real spilling union/coalesce cases would strengthen coverage beyond synthetic overlapping trees. Any effort to accumulate physical spill across repeated execution instances should address the underlying SQL metric publication semantics explicitly. These are coverage and accounting improvements, not verified P1/P2 findings or additional approval conditions for this change.

sunchao:
Merged, thanks @peterxcli !!

PR #5491, opened 2026-08-27
feat: bind task-owned RSS callbacks to native shuffle plans (4/n)

pingzh:
## Which issue does this PR close?

Part of #5352. This is the fourth foundational PR and does not close the issue.

Previous PRs:

- https://github.com/apache/datafusion-comet/pull/5473
- https://github.com/apache/datafusion-comet/pull/5476
- https://github.com/apache/datafusion-comet/pull/5481

## Rationale for this change

Previous PRs introduced RSS partition writers, destination-aware shuffle plans, and native RSS execution. However, serialized RSS plans cannot execute without connecting their task-owned JVM callback to the native planner.

This PR completes that connection while preserving existing local shuffle behavior and JNI compatibility.

## What changes are included in this PR?

- Add task-scoped RSS callback registration through a separate JNI method, preserving the existing `createPlan` signature.
- Pass optional shuffle callbacks through `CometExecIterator` into native execution contexts and the physical planner.
- Bind RSS shuffle plans to their task-owned callback and preserve callback isolation between tasks.
- Reject RSS plans without callbacks, RSS plans containing local output paths, and callbacks supplied to local shuffle plans.
- Apply the JVM’s conservative maximum byte-array size to RSS frame limits.
- Preserve original callback exceptions across JNI boundaries.
- Clean up native resources and all shuffle inputs when callback registration fails.

## How are these changes tested?

- Added planner tests covering RSS binding, task isolation, destination validation, and end-to-end native execution.
- Added Spark/JNI integration tests covering complete RSS frames, original callback exception identity, existing JNI signature compatibility, and registration-failure cleanup.
- Passed 105 Rust tests across planner, shuffle, JNI bridge, and protobuf suites.
- Passed all 35 tests in `CometNativeShuffleSuite`.
- Passed workspace-wide Clippy, formatting checks, and the full Maven reactor build.

PR #5476, opened 2026-08-26
feat: add partition writer destinations to native shuffle plans (2/n)

pingzh:
## Which issue does this PR close?

Part of #5352. This is the second foundational PR and does not close the issue.

Previous PR: 
- https://github.com/apache/datafusion-comet/pull/5473 

## Rationale for this change

Comet’s native shuffle plans currently assume all partition data is written to local files. Supporting remote shuffle services requires plans to explicitly describe their output destination while preserving compatibility with existing local shuffle behavior.

## What changes are included in this PR?

- Add a protobuf partition-writer descriptor supporting local and RSS destinations.
- Serialize local shuffle plans with both the new descriptor and existing output-file fields for backward compatibility.
- Update native planning to:
  - Support legacy plans without a partition-writer descriptor.
  - Resolve local output paths from the new descriptor.
  - Reject missing destinations, invalid paths, and conflicting legacy fields.
  - Reject RSS destinations until remote shuffle execution is introduced in a subsequent PR.
- Preserve existing shuffle execution behavior.

## How are these changes tested?

- Four protobuf tests verify local/RSS serialization and compatibility between legacy and updated plans.
- Nine native planner tests cover valid destinations, legacy fallback, malformed plans, conflicting paths, and unsupported RSS destinations.
- Three new JVM tests verify local, RSS, and legacy plan serialization.
- All 31 native-shuffle JVM tests, 50 native shuffle tests, and 25 JNI tests pass.
- Full-workspace Clippy, Rust formatting, ScalaStyle, Spotless, and the complete Maven reactor all pass.

PR #5501, opened 2026-08-27
feat: add Celeborn shuffle manager and partition pusher (5/n)

pingzh:

Part of #5352. This is the fifth foundational PR and does not close the issue.

Previous PRs:

## Rationale for this change

Previous PRs introduced backend-neutral RSS partition writers, destination-aware shuffle plans, native shuffle execution, and task-owned JVM callbacks.

Integrating Apache Celeborn additionally requires a Celeborn-specific partition pusher and a shuffle manager that preserves existing Celeborn behavior for ordinary Spark shuffles.

- Add `CelebornShufflePartitionPusher`, which:
  - Sends complete native shuffle frames through Celeborn’s existing client.
  - Captures shuffle, mapper, attempt, and partition metadata.
  - Validates frame boundaries, partition identifiers, and transport byte counts.
  - Preserves original client exceptions across reflective calls.
  - Avoids introducing a compile-time Celeborn dependency.

- Add `CometCelebornShuffleManager`, which:
  - Loads the existing Celeborn shuffle manager reflectively.
  - Delegates ordinary Spark shuffle registration, writing, reading, cleanup, and shutdown.
  - Rejects unsupported Comet shuffle paths until subsequent integration PRs.
  - Fails clearly when the optional Celeborn client is unavailable.

- Register the new test suites in Linux and macOS CI workflows.

Map-side lifecycle management, Celeborn shuffle reading, and native-only planning will be addressed in subsequent PRs.

- Added 11 partition-pusher tests covering frame validation, task metadata, transport accounting, client failures, and worker-thread execution.
- Added 12 shuffle-manager tests covering delegation, reader ranges, lifecycle handling, unsupported shuffle paths, and missing Celeborn dependencies.
- Passed all 35 existing native shuffle tests.
- Passed all 58 tests across the three relevant suites.
- Passed ScalaStyle, Spotless, Apache license-header verification, and CI suite-registration checks.

### DIFF
--- a/spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
+++ b/spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
@@ -34,6 +34,7 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
 
   private final Object shuffleClient;
   private final Method pushOrMergeData;
+  private final Method computeBatchCRC;
   private final int shuffleId;
   private final int mapId;
   private final int encodedAttemptId;
@@ -110,8 +111,48 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
           "Celeborn raw-push API must be an instance method returning an int");
     }
 
+    Method integrityMethod = null;
+    try {
+      integrityMethod =
+          shuffleClient
+              .getClass()
+              .getMethod(
+                  "computeBatchCRC",
+                  int.class,
+                  int.class,
+                  int.class,
+                  int.class,
+                  byte[].class,
+                  int.class,
+                  int.class);
+    } catch (NoSuchMethodException missing) {
+      // Older Celeborn clients account for integrity inside pushOrMergeData instead.
+      try {
+        for (Method method : shuffleClient.getClass().getMethods()) {
+          if (method.getName().equals("computeBatchCRC")) {
+            throw new IllegalArgumentException(
+                "Celeborn integrity-accounting API has an incompatible signature", missing);
+          }
+        }
+      } catch (SecurityException failure) {
+        throw new IllegalArgumentException(
+            "Cannot inspect the optional Celeborn integrity-accounting API", failure);
+      }
+    } catch (SecurityException failure) {
+      throw new IllegalArgumentException(
+          "Cannot resolve the optional Celeborn integrity-accounting API", failure);
+    }
+
+    if (integrityMethod != null
+        && (integrityMethod.getReturnType() != void.class
+            || Modifier.isStatic(integrityMethod.getModifiers()))) {
+      throw new IllegalArgumentException(
+          "Celeborn integrity-accounting API must be an instance method returning void");
+    }
+
     this.shuffleClient = shuffleClient;
     this.pushOrMergeData = pushMethod;
+    this.computeBatchCRC = integrityMethod;
     this.shuffleId = shuffleId;
     this.mapId = mapId;
     this.encodedAttemptId = encodedAttemptId;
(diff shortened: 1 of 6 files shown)


spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
```diff
@@ -0,0 +1,190 @@
...
+              + declaredBodyLength
+              + " body bytes, but contains "
+              + (length - Long.BYTES));
+    }
+
+    final int accepted;
+    try {
+      accepted =
+          (int)
+              pushOrMergeData.invoke(
```
ziting-openai, line 197: [P2] Record Celeborn's required integrity accounting before raw pushes

With `spark.celeborn.client.shuffle.integrityCheck.enabled=true`, invoking `pushOrMergeData` directly skips the only CRC/byte accumulation path. Celeborn v0.7 explicitly documents that external writers must use `pushDataWithCRC`; that method first calls `computeBatchCRC`, and mapper completion later publishes the accumulated per-partition CRCs and byte counts. This adapter submits the frames without updating either value, so reducer-side integrity validation receives missing/incorrect commit metadata and fails. Please invoke Celeborn's CRC accounting exactly once before each raw push (while preserving the required `skipCompress=true` behavior) and add a regression with integrity checks enabled.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
```diff
@@ -0,0 +1,190 @@
...
+      if (cause instanceof RuntimeException) {
+        throw (RuntimeException) cause;
+      }
+      if (cause instanceof Error) {
+        throw (Error) cause;
+      }
+      throw new IOException("Celeborn raw shuffle push failed", cause);
+    }
+
+    int expected = length + CELEBORN_BATCH_HEADER_BYTES;
```
ziting-openai: [P2] Do not reject successful encrypted Celeborn pushes

With supported Spark I/O encryption enabled (`spark.io.encryption.enabled=true`), Celeborn applies its `CryptoHandler` inside `pushOrMergeData` even when `skipCompress=true`. The encrypted payload is longer than the original Comet frame, and the real client returns `encrypted.length + 16`, not `length + 16`. This exact comparison consequently throws after every encrypted frame has already been submitted, failing/retrying the task and potentially duplicating remote data. Either account for the actual encrypted transport length or reject incompatible encryption before pushing, and cover a length-expanding crypto handler in the test client.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
ziting-openai: Verified fixed in 6a6f3d27: plaintext CRC/byte accounting now runs exactly once before each raw push, with coverage for multiple partitions, failures, encryption, and legacy clients. This thread can be resolved.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
ziting-openai: Verified fixed in 6a6f3d27: encrypted transport expansion is accepted while undersized responses remain rejected, and the new regression covers combined encryption and integrity accounting. This thread can be resolved.

sunchao (approved):
Reviewed head `6a6f3d27ca5d929200e668c5389f33627a2edc6d` against base `525c05b52bb73bb873f05c32e30987132676bf5e`. No remaining P1/P2 findings.

Independently verified the CRC accounting and encrypted-length fixes using real Celeborn 0.7 client methods and Spark crypto with an in-memory transport. Native core built successfully, followed by 28 focused pusher/manager tests on each of Spark 3.4, 3.5, 4.0, 4.1, and 4.2: 140 passing test executions.

Full Celeborn cluster integration was not run; broader CI remains in progress.

sunchao:
Merged, thanks @pingzh !

ziting-openai:
Sorry for the noise. My automated review on this PR was posted by a misconfigured AI agent that should have been restricted to OpenAI-owned repositories. Please disregard that automated review. My inline review threads are already resolved. This agent will not review non-OpenAI repositories again.

PR #5505, opened 2026-08-27
fix: cancel background batch producers before final metrics

### LINKED ISSUE #5504: Cancel background batch producers before collecting final plan metrics
### Describe the bug

The async I/O execution path discards its producer `JoinHandle`. When Spark closes a plan early, `releasePlan` snapshots metrics before the receiver and context are dropped. The producer can still own the stream, so stream-drop metric guards can update counters only after the final snapshot. Closing the consumer also does not explicitly request producer cancellation.

### Steps to reproduce

Start an async batch producer whose stream owns a drop-time metric guard and remains pending, then close its consumer before it finishes. A final snapshot taken before the producer releases the stream misses the guard's update. A producer inside synchronous decoding is a distinct case: it cannot observe cancellation until it yields.

### Expected behavior

Retain producer ownership, close the receiver, request cancellation, and wait only within a bounded budget before the final metric snapshot. Cooperative completion should expose drop-time updates; timeout must explicitly remain a potentially incomplete snapshot. There must be no claim that a timeout makes external JVM task state safe to tear down.

### Additional context

Extracted from #5453 to review cancellation behavior and close latency independently of new I/O counters. Existing producer ownership is `async move` over the stream and sender, not a captured raw `ExecutionContext` pointer. Cancellation remains cooperative; this issue does not promise preemption of synchronous decoding or complete accounting after a timeout.

sunchao:

Closes #5504.

## Rationale for this change

Early plan close can collect final metrics while the async batch producer still owns its stream. Updates performed when stream guards are dropped then occur after the snapshot. The producer handle is currently discarded, so closing the plan cannot explicitly cancel that work.

This extracts the producer-lifecycle change from #5453. It adds no scan counters, does not change object-store registration, and does not depend on the metrics PR.

Keep the producer `JoinHandle` in the execution context. On the Spark executor thread, drop the batch receiver, abort the producer, and wait for cooperative completion before collecting final metrics. The helper returns an explicit finished/timeout result and logs when the final snapshot may be incomplete. Waiting uses a 100 ms budget; a missing or already finished producer does not pay that delay, and a cooperative producer can stop earlier.

This is best-effort final accounting. The producer owns its stream and sender rather than a raw execution-context pointer, but synchronous decoding cannot be preempted by `abort`, and a timed-out producer may still consult external task state or update counters later. The waiting budget is not a memory-safety guarantee or a hard wall-clock limit under arbitrary scheduler delays.

Dedicated one-worker Tokio runtimes keep blocking stand-ins out of the global runtime. Four tests verify cooperative drop-time metric visibility, cancellation while a batch is queued, an explicitly incomplete timeout snapshot followed by a later update, and an already finished producer under exhausted Tokio cooperative budget. The blocked test releases its worker after shutdown returns, so it does not depend on a tight upper timing assertion.

A manual ignored test measures the helper with 100 already finished producers, 100 pending cooperative producers, and five blocked synchronous stand-ins. In the full debug-build test binary, observed median / p95 durations were **2.124 / 2.714 microseconds** for finished producers, **1.123 / 1.142 milliseconds** for cooperative pending producers, and **100.129 / 100.989 milliseconds** for the five blocked stand-ins. Cooperative cancellation exposed the drop-time counter before the snapshot; every blocked sample returned with that counter still incomplete. These are local helper/scheduler measurements, not end-to-end Spark close timings or a latency guarantee.

After merging Apache main `525c05b52` in [3f0b62a5e](https://github.com/apache/datafusion-comet/commit/3f0b62a5e021864bad8a2285f7eb0d1cc4b6a5ff), the exact merged native core library test target rebuilt successfully. All **five selected tests passed**: the four producer lifecycle regressions plus main's memory-pool registration regression. The manual timing test was not rerun. The merge preserves cancellation before the metrics snapshot and the upstream registration guard before the fallible update; both parents' tests remain unchanged. Formatting and the PR diff check against the merged main commit pass. No Spark/JNI release retry test was run.

Before that merge, the full native crate compiled with the two disjoint extracted changes combined at their shared base. All **four lifecycle regressions passed**, and the ignored manual measurement passed when invoked explicitly. All **90 Parquet tests** also passed with a confirmed zero process exit status. `cargo fmt --all --check` and `git diff --check` pass. Default native features and JDK 21 were used. The measurement isolates the shutdown helper and does not measure a full Spark task or production decoding workload.


sunchao:
Merged current Apache main in [3f0b62a5e](https://github.com/apache/datafusion-comet/commit/3f0b62a5e021864bad8a2285f7eb0d1cc4b6a5ff) to resolve the conflict without rewriting history. The resolution preserves both producer cancellation before the final snapshot and the upstream memory-pool registration cleanup before a fallible metrics update. Both sets of tests are retained.

The exact merged native core library test target compiled and **all five selected lifecycle/registration tests passed**. Rust formatting and the PR diff check against main also passed. The shutdown helper is unchanged; its earlier helper timings remain labeled as historical measurements, not a fresh Spark/JNI teardown benchmark.

andygrove (commented):
This shows as conflicting with main right now, and I think I know why. `native/core/src/execution/jni_api.rs`'s `releasePlan` was restructured by #5494 (the task-shared memory pool RAII change), which landed two days after this PR's last push. `task_attempt_id` and `memory_pool_config` came off `ExecutionContext`, `handle_task_shared_pool_release` was deleted, and the function now reclaims the `Box<ExecutionContext>` up front rather than leaving the raw pointer alive for a JVM-side retry. This PR's diff still calls `handle_task_shared_pool_release(execution_context.memory_pool_config.pool_type, execution_context.task_attempt_id)` around line 1070, so it won't compile against current main. The green checks above all ran against the head commit from the 27th, before that landed, so they don't tell us anything about the rebased state.

Once this is rebased, the cancellation block will need a new home. The order you have now, drop the receiver, abort the producer and wait, then take the metrics snapshot, still looks right to me. It should probably sit right after the box is reclaimed and before the rest of the teardown that #5494 added, so a cooperative producer's drop-time metric guards still run before the snapshot rather than after.

The cancellation logic itself reads correctly on its own terms. Dropping `batch_receiver` before calling `abort` is the right order. A producer parked in `tx.send(...).await` gets woken with a channel-closed error as soon as the receiver drops, so `stop_batch_producer` doesn't need to rely on abort for that case at all. The wait loop runs on the calling thread with `thread::sleep` rather than inside the tokio runtime, so a saturated runtime can't turn the 100ms budget into a wait on itself. The producer's async block only captures `stream` and `tx`, not a pointer back into `ExecutionContext`, so a producer still running in the background after `releasePlan` returns isn't reaching into memory that's already been freed. The four new tests drive `stop_batch_producer` directly against a dedicated single-worker runtime and cover the cooperative, saturated-channel, blocked, and already-finished cases, which matches what this helper needs.

One separate question. The timeout log around line 1054 uses `log::debug!`, while the rest of this file logs comparable one-off events with `info!`. This is the one signal that the final metrics snapshot may be missing data, so would `warn!` make it more likely an operator actually notices it?

PR #5513, opened 2026-08-27
feat: complete Celeborn map-side shuffle push lifecycle (6/n)

pingzh:

Part of #5352. This is the sixth foundational PR and does not close the issue.

Previous PRs:

## Rationale for this change

The existing Celeborn integration provides the shuffle manager and partition-pusher foundation but does not yet complete the native map-side shuffle lifecycle.

Production use requires bounded frames, executor-wide backpressure, asynchronous completion tracking, cancellation handling, speculative-attempt coordination, retry-safe commits, and support for nested Arrow data.

- Wire native Comet shuffle writers into Celeborn without creating local shuffle files.
- Add executor-wide byte admission covering native encoding, JNI copies, and Celeborn transport.
- Track asynchronous push completion and preserve admission until requests actually complete.
- Handle task cancellation, mapper completion, cleanup, and asynchronous transport failures.
- Coordinate shuffle generations, speculative attempts, retries, and Spark output commits.
- Handle stale commit ownership on Spark 3.4.
- Support Celeborn 0.6 and 0.7 mapper-completion APIs.
- Preserve existing CRC accounting, encrypted payload handling, and generic JNI callback compatibility.
- Split oversized batches into complete row-aligned frames and compact nested Arrow dictionaries, lists, maps, and structs.
- Add configurable limits:
  - `spark.comet.shuffle.rss.maxFrameBytes`
  - `spark.comet.shuffle.rss.maxInFlightBytes`
- Register the new native shuffle writer suite in Linux and macOS CI.

- 101 JVM tests across:
  - `CelebornShufflePartitionPusherSuite`
  - `CometCelebornShuffleManagerSuite`
  - `CometCelebornNativeShuffleWriterSuite`
  - `CometConfSuite`
  - `CometNativeShuffleSuite`
- 67 Rust shuffle tests, including bounded frames, nested Arrow compaction, backpressure, failed writers, and legacy callbacks.
- 17 native planner tests covering local and remote shuffle destinations.
- 3 JNI bridge tests.
- Scala formatting, Rust formatting, Scala style checks, and CI suite-registration checks all pass.

### DIFF
--- a/native/shuffle/src/writers/rss/rss_partition_writer.rs
+++ b/native/shuffle/src/writers/rss/rss_partition_writer.rs
@@ -18,11 +18,19 @@
 use crate::metrics::ShufflePartitionerMetrics;
 use crate::writers::partition_writer::PartitionWriter;
 use crate::ShuffleBlockWriter;
-use arrow::array::RecordBatch;
+use arrow::array::cast::AsArray;
+use arrow::array::{
+    Array, ArrayRef, BinaryViewArray, FixedSizeListArray, LargeListArray, LargeListViewArray,
+    ListArray, ListViewArray, MapArray, RecordBatch, RunArray, StringViewArray, StructArray,
+    UnionArray,
+};
+use arrow::buffer::OffsetBuffer;
+use arrow::datatypes::{DataType, Field, Int16Type, Int32Type, Int64Type};
 use arrow::ipc::writer::CompressionContext;
+use arrow_select::dictionary::garbage_collect_any_dictionary;
 use datafusion::common::{DataFusionError, Result};
 use datafusion_comet_jni_bridge::ShufflePartitionPusher;
-use std::io::Cursor;
+use std::io::{self, Cursor, Seek, SeekFrom, Write};
 use std::sync::Arc;
 
 /// Sends complete Comet shuffle blocks to a task-owned remote shuffle pusher.
@@ -42,9 +50,9 @@ pub(crate) struct RssPartitionWriter {
     num_partitions: usize,
     max_frame_size: usize,
     compression_context: CompressionContext,
-    frame: Vec<u8>,
     next_partition_to_finish: usize,
     finished: bool,
+    failed: bool,
 }
 
 impl RssPartitionWriter {
@@ -83,14 +91,14 @@ impl RssPartitionWriter {
             num_partitions,
             max_frame_size,
             compression_context: CompressionContext::default(),
-            frame: Vec::new(),
             next_partition_to_finish: 0,
             finished: false,
+            failed: false,
         })
     }
 
     fn validate_writable_partition(&self, partition_id: usize) -> Result<i32> {
-        if self.finished {
+        if self.finished || self.failed {
             return Err(DataFusionError::Execution(
                 "Remote shuffle writer has already finished".to_string(),
             ));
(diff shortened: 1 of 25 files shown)


ziting-openai (commented):
Native-first Celeborn initialization bypasses configured Spark shuffle encryption, and fixed maximum-frame admission serializes encoding under the default executor limits.

spark/src/main/scala/org/apache/comet/shuffle/CelebornShufflePusherFactory.scala
```diff
@@ -0,0 +1,337 @@
...
+      val celebornConfClass = ClassLoaders.loadClass(CELEBORN_CONF)
+      val userIdentifierClass = ClassLoaders.loadClass(CELEBORN_USER_IDENTIFIER)
+      val celebornConf =
+        sparkUtilsClass.getMethod("fromSparkConf", classOf[SparkConf]).invoke(null, conf)
+
+      def handleValue(name: String): AnyRef = handleClass.getMethod(name).invoke(handle)
+
+      val client = acquireClient(
+        shuffleClientClass
+          .getMethod(
```
ziting-openai: [P1] Preserve Spark shuffle encryption when creating the Celeborn client

This native path invokes the six-argument `ShuffleClient.get`, which in Celeborn 0.7 delegates to the seven-argument overload with `Optional.empty()` for its crypto handler. The normal Spark shuffle manager instead supplies `SparkCommonUtils.getCryptoHandler(conf)`. If a native Comet shuffle initializes an executor's shared Celeborn client first while `spark.io.encryption.enabled=true`, shuffle data is transmitted and stored as plaintext despite the application explicitly requiring encryption. Please prefer the crypto-aware overload with Spark's configured handler, retain a compatible fallback only for older clients, and add a native-first encrypted-shuffle regression.

native/shuffle/src/writers/rss/rss_partition_writer.rs
```diff
@@ -125,36 +132,616 @@ impl RssPartitionWriter {
...
+        if minimum_size > self.max_frame_size && batch.num_rows() > 1 {
+            return self.push_split_batch(partition_id, batch, metrics);
+        }
 
-            if encoded_size == 0 {
-                continue;
+        // Native IPC, its JNI byte array, and Celeborn's copied transport request overlap.
+        // Acquire all three copies before compaction/encoding to avoid allocation-before-admission
+        // and deadlocks caused by growing another task's reservation after partial acquisition.
+        let overlapping_copies = self.max_frame_size.checked_mul(3).ok_or_else(|| {
```
ziting-openai: [P2] Size encoding admission to the actual frame instead of its configured maximum

Every nonempty batch reserves `3 * self.max_frame_size` before compaction and encoding, regardless of its actual size. With the default 64 MiB frame maximum and 256 MiB executor-wide in-flight limit, each tiny frame therefore acquires 192 MiB plus its header, so two map tasks cannot encode concurrently on the same executor. The reservation is reduced only after native IPC encoding/compression finishes and Java claims the actual frame. This serializes CPU-intensive shuffle encoding across all executor cores under default settings. Please reserve a conservative per-batch upper bound while preserving atomic three-copy accounting, and add a concurrent small-frame regression.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
```diff
@@ -207,30 +479,503 @@ public void pushPartitionData(int partitionId, byte[] data, int length) throws I
...
+          }
+          if (asynchronousFailure == null) {
+            asynchronousFailure = observedFailure;
+            detectedFailure = asynchronousFailure;
+          }
+        }
+        long retainedRequests =
+            Math.max(
+                0L, observed.inFlightRequests.sum() - (observed.terminalFailureCredited ? 1L : 0L));
+        long completions = Math.max(0L, observed.submittedPushes - retainedRequests);
```
sunchao, line 792: [P2] Track failed transport completion after task cleanup

On `bd439e12137b12b81276b40907f7130371ef4c14`, cancelling a map with an outstanding push can permanently consume this executor-wide admission budget. Stock Celeborn 0.6.3 and 0.7.0 `cleanup()` sets the captured `PushState.exception` to `"Cleaned Up"`; if that request subsequently fails (for example, its connection closes or it times out), the stock raw-push callback returns immediately because the exception is already set, without calling `removeBatch()`. Thus the counter used here remains positive even after the transport has completed. The reconciler never releases the reservation, and later native maps sharing the client can wait indefinitely in `acquire()`.

I reproduced this with the unmodified Comet pusher and real Celeborn `ShuffleClientImpl`, `PushState`, cleanup, and raw callbacks, replacing only network delivery. A real native `ShuffleBlockWriter` frame containing 32 Int32 rows is 604 bytes. With valid `maxFrameBytes=604` and `maxInFlightBytes=2416`, reserve 1812 bytes, push the frame, abort the map, and deliver the held stock callback's `onFailure`. All original transport callbacks have completed, but the stock counter stays at 1 and available admission stays at 1796; another map's required 1828-byte admission blocks until interrupted. The head's successful-callback control releases all 2416 bytes, and the same failed-callback sequence on base `eabb5d4773091b983d8fce713f0e34b1cf93f877` does not block another map.

Please account for actual failed transport completion after cleanup rather than relying on this counter to eventually fall. The existing cancellation test only completes the cancelled request successfully, so it misses this failure path.

spark/src/main/scala/org/apache/comet/shuffle/CelebornShufflePusherFactory.scala
ziting-openai: Verified fixed at 26efda6a: native client initialization now selects the crypto-aware Celeborn 0.7 overload with the Spark handler, retains the 0.6 fallback, and covers native-first encrypted plus fail-closed behavior.

native/shuffle/src/writers/rss/rss_partition_writer.rs
ziting-openai: Verified fixed at 26efda6a: admission now scales to a conservative per-batch bound, retries release their full reservation, and the new two-thread regression confirms concurrent small-frame encoding under production limits.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
ziting-openai: [P2] This still reproduces at 16d71677 despite the new transport snapshot. Celeborn 0.6/0.7 removes a failed request from outstandingPushes before invoking its callback. If the transport thread pauses between removal and callback, abort() snapshots no request, cleanup() sets Cleaned Up, and the resumed callback returns without removeBatch(); cancelledTransportBaseline is never initialized, so executor-wide admission remains permanently reserved. The new regression only fails the transport after abort has already captured the request. Please cover the removal-before-snapshot/callback-after-cleanup interleaving and track completion independently of transport-map membership.

spark/src/test/scala/org/apache/comet/shuffle/CelebornShufflePartitionPusherSuite.scala
```diff
@@ -405,6 +1144,374 @@ class RecordingCelebornPushClient {
...
+  val pushDataRetryPool: ExecutorService = retryExecutor
+
+  def getDataClientFactory: RecordingCelebornTransportClientFactory = dataClientFactory
+
+  var openConnectionBeforePush: Boolean = false
+  var beforePushBegins: () => Unit = () => ()
+  var beforePushReturns: RecordingCelebornPushState => Unit = _ => ()
+  var retriesBeforeFailure: Int = 0
+  var retryCallback: RecordingCelebornTransportCallbackApi => Unit =
+    _.onFailure(new IOException("revive failed"))
```
sunchao: [P2] Keep the callback initializers valid under Scala 2.12

At exact head `520ca35f5dbaff1a9a47ad02bb7af027d93910d1`, these two new mutable callback initializers prevent the default Spark 3.4 and 3.5 builds from compiling their test sources. Scala 2.12 parses the bare `_` after a `var` declaration as its default initializer, then rejects the remaining `=>` at line 1284 and `.` at line 1287. Root Maven `test` with `-Pspark-3.5 -Pjdk17` and `clean test` with `-Pspark-3.4 -Pjdk17` both stop in `scala:testCompile` before any selected suite can execute. The current [Spark 3.5 CI job](https://github.com/apache/datafusion-comet/actions/runs/33125216230/job/98702084424) and [Spark 3.4 CI job](https://github.com/apache/datafusion-comet/actions/runs/33125216230/job/98702084393) report the same two errors.

For the causal control, this suite's exact base `eabb5d4773091b983d8fce713f0e34b1cf93f877` and previous `16d71677d5975bb2aaa0d835a77f781d487c7262` sources both pass the parser with Scala 2.12.17 and 2.12.18; the current source fails. Parenthesizing only these initializer expressions in a scratch copy restores parsing. Spark 4.0/Scala 2.13 passes all 63 focused tests, so that profile does not cover this compatibility regression. Please express both lambdas in a form Scala 2.12 accepts so normal Spark 3.4/3.5 test and package builds can complete.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
ziting-openai: Rechecked at be5f4af7d7876e862af9be172b95e2e704acf049 with three independent reviews. The callback tracker now retains ownership before request publication and releases it only after the original callback returns, including retry handoffs. This addresses the cleanup/failure leak and the removal-before-callback interleaving described above; map membership is no longer the completion signal. I traced the implementation and inspected the added regressions; I did not rerun the Celeborn integration tests.

_Posted by Codex on behalf of ziting-openai._

native/shuffle/src/writers/rss/rss_partition_writer.rs
```diff
@@ -125,35 +132,633 @@ impl RssPartitionWriter {
...
+                )
+            })?;
+            // Legacy Java callbacks retain the full JVM-soft frame limit but inherit a no-op
+            // reservation method whose argument is still a signed Java int. Concrete bounded
+            // pushers advertise a smaller frame limit, so their actual three-copy reservation
+            // fits; capping here preserves legacy callbacks and their failure propagation.
+            let reservation = overlapping_copies
+                .max(original_size.saturating_add(compaction_scratch))
+                .min(self.pusher.max_reservation_size());
+            self.pusher.reserve_partition_data(reservation)?;
```
sunchao: [P2] Charge codec workspace and allocated capacity before encoding

At exact head `be5f4af7d7876e862af9be172b95e2e704acf049`, the new `spark.comet.shuffle.rss.maxInFlightBytes` contract includes native encoding scratch, but this reservation covers estimated IPC/frame lengths and compaction only. The actual `ShuffleBlockWriter` also creates codec working buffers, and `BoundedBuffer` bounds its Vec's length rather than its allocated capacity. Consequently, a valid budget can be exceeded even by one admitted encoder, before transport completion or the separate callback-lifetime issue matters.

I independently compiled the exact current/base RSS writer sources with the unchanged block writer and measured successful live Rust heap allocation requests only during encoding, excluding input/schema/metric construction. With a 16,384-byte budget, the real Java pusher accepts a one-Int32-row LZ4 reservation of 12,687 bytes (12,703 including the Celeborn header; effective frame cap 5,456), but encoding peaks at 154,497 bytes and produces a 238-byte frame. The uncompressed control peaks at 1,160 bytes. This is not only a tiny-budget codec case: with 16,384 Int32 rows and a 262,144-byte budget, the accepted reservation is 215,424 + 16, while LZ4 peaks at 285,321 bytes. Without compression, that batch's 67,996-byte frame has 135,976 bytes of Vec capacity; adding its JNI and Celeborn copies requires at least 271,984 bytes, also above the budget. The Java reservation checks were exercised with stock Celeborn 0.6.3 and 0.7.0.

Base `eabb5d4773091b983d8fce713f0e34b1cf93f877` has the same codec allocation peaks but no encoding-admission protocol or byte-budget setting. The introduced defect is the new limit admitting work without accounting for memory it explicitly promises to cover; this is not a claim that the codec allocations themselves were introduced here. Please include codec workspace and actual buffer capacities in the bound, or reject budgets that cannot accommodate them, before allowing encoding to begin.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
```diff
@@ -175,62 +570,449 @@ public void pushPartitionData(int partitionId, byte[] data, int length) throws I
...
+  private void markSubmitted(
+      PushReservation reservation, ObservedPushState pushState, int acceptedBytes) {
+    int released;
+    synchronized (lifecycleLock) {
+      reservation.pushState = pushState;
+      reservation.submitted = true;
+      pushState.submittedPushes++;
+      int retained = Math.min(reservation.bytes, acceptedBytes);
+      released = reservation.bytes - retained;
+      reservation.bytes = retained;
```
sunchao: [P2] Retain native and JNI frame charges until their buffers are released

At exact head `be5f4af7d7876e862af9be172b95e2e704acf049`, `markSubmitted` shrinks admission to the Celeborn transport byte count while the caller still owns both the native output Vec and the JNI byte array. A fast transport completion can then release the remainder. The JNI local frame is popped only after `JavaShufflePartitionPusher::push_partition_data` returns from its callback, and the native RSS writer drops its output afterward. Another map can therefore acquire the released permits while those two frame copies remain live.

I reproduced this using the actual Rust RSS writer, real JavaShufflePartitionPusher JNI bridge, unmodified Java pusher, and stock Celeborn 0.6.3/0.7.0 request registration/removal/success callbacks. Only network delivery and scheduling before JNI return were controlled. Two maps each encode 2,048 Int32 rows without compression, producing 8,860-byte frames, with `maxFrameBytes=8860` and the valid shared budget `maxInFlightBytes=26596`. After the first production Java push returns but before JNI releases its array/native frame, all 26,596 permits are available. The second encoder enters; the two paused calls then own at least 35,440 bytes of native/JNI frame data alone, excluding capacity and all transport/allocator overhead. A control deferring semaphore release until the first native writer returns leaves zero permits, blocks the second encoder until those buffers are gone, and then lets both complete.

The exact base `eabb5d4773091b983d8fce713f0e34b1cf93f877` has these same frame lifetimes but no shared admission protocol; the faulty release transition is new in this PR. This is separate from the fixed cleanup/transport-callback leak and from codec workspace accounting. Please keep the native/JNI portion charged until those owners actually finish rather than releasing it when the raw Celeborn call returns or its transport callback completes.

sunchao (approved):
Follow-up review at `93b8336ed2f61dfc82d177b6e1bb994eb7e341aa` against base `eabb5d4773091b983d8fce713f0e34b1cf93f877`: verified the fixes for both prior admission P2s and rechecked the earlier fixes. No remaining P1/P2 found.

Validation: native core; 102 Rust tests; 128 JVM tests each on Spark 3.4, 3.5, and 4.0, and 76 each on 4.1 and 4.2; 96 allocation, 74 actual-JNI, and 48 stock Celeborn controls with old/base causal comparisons.

Tests used local/controlled transport, not a deployed Celeborn cluster. Encrypted native RSS is explicitly rejected; ordinary Spark crypto handling is retained. CI is still running.

sunchao:
Merged, thanks for the work @pingzh !!! cc @wForget @andygrove too

andygrove:
Thanks for this @pingzh, and sorry to arrive after the merge. I did a local pass over the merged commit and wanted to record what I found rather than let it sit in my notes.

The good news first, because it is the part I most wanted to check. The data path holds up. I wrote round-trip probes for the cases the suite does not cover, nullable sliced `List`, `LargeList`, `Map`, `FixedSizeList` and `Struct` with dictionaries nested inside them, plus `NaN`, `-0.0`, positive and negative infinity, subnormals, empty strings, composed versus decomposed UTF-8, astral plane characters and embedded NULs. Every one round-trips exactly, so the dictionary garbage collection, the offset rebasing and the row-aligned splitting all preserve values and row order. I also verified every reflected Celeborn member against the real v0.6.0 and v0.7.0 sources and they all resolve, including the four-argument versus five-argument `mapperEnd` split. Locally I get 77 Rust tests and 128 JVM tests green on Spark 3.5 with Scala 2.12, and no encode regression at Spark's default `spark.shuffle.compress=true`.

I did find four things worth following up on, so I filed them separately rather than burying them here:

- #5527, the two new configs cannot both hold at their shipped defaults. `maxFrameBytes` defaults to 64 MiB but the effective per-row cap at the default 256 MiB `maxInFlightBytes` is about 36 MiB, and a row above that fails the query outright after four stage reruns.
- #5528, the tracker assigns four fields that are `private final` in real Celeborn, and the visibility gap fails toward releasing admission while a payload is still in flight.
- #5529, the bootstrap hook is fatal to `createClient` and is never uninstalled, so a Comet-specific incompatibility can stop all Celeborn shuffle on the executor.
- #5530, Celeborn is not a build dependency in any scope, so the reflection is only ever exercised against hand-written stand-ins.

#5527 is the one I would prioritise, since it has a short reproducer and affects anyone running the defaults. There are also some smaller perf and polish items I noticed that I did not think justified their own issues, and I am happy to add them to #5352 as a checklist if that is useful.

pingzh:
hi @andygrove thanks for checking. I will also work on those follow up items. 🙏

ziting-openai:
Sorry for the noise. My automated review on this PR was posted by a misconfigured AI agent that should have been restricted to OpenAI-owned repositories. Please disregard that automated review. My inline review threads are already resolved. This agent will not review non-OpenAI repositories again.

PR #5516, opened 2026-08-28
fix: report bounded shuffle allocator memory usage

ywskycn:

Solving sub-issue 11 in #5212.

## Rationale for this change

`CometBoundedShuffleMemoryAllocator` tracks allocations through its own `allocatedMemory` counter because its allocations bypass Spark's memory manager. However, it inherited `MemoryConsumer.getUsed()`, which reports Spark's unused internal counter and always returned `0`.

This caused Comet shuffle memory usage logging to report zero even when the bounded allocator held pages or pointer arrays.

- Override `getUsed()` in `CometBoundedShuffleMemoryAllocator` to return the synchronized `allocatedMemory` counter.
- Document that the value represents allocator-wide usage shared across tasks.
- Add regression tests for page allocations, pointer arrays, repeated frees, failed allocations, exhausted budgets, and page table exhaustion.
- Register the new test suite in Linux and macOS CI.

- Added `CometBoundedShuffleMemoryAllocatorSuite`.
- Verified all five new tests failed before the fix because `getUsed()` returned `0`.
- Verified the allocator suite and existing `SpillSorterSuite` pass after the fix:

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -333,6 +333,7 @@ jobs:
               org.apache.comet.exec.CometAsyncShuffleSuite
               org.apache.comet.exec.DisableAQECometShuffleSuite
               org.apache.comet.exec.DisableAQECometAsyncShuffleSuite
+              org.apache.spark.shuffle.comet.CometBoundedShuffleMemoryAllocatorSuite
               org.apache.spark.shuffle.sort.SpillSorterSuite
           - name: "exec"
             value: |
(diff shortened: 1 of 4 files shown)


sunchao (approved):
Reviewed head `f4b20c1e30ff790b14305c83c0227df772a1c33e` with base `eabb5d4773091b983d8fce713f0e34b1cf93f877`. No P1/P2 findings.

Native core built before the focused JVM suites: 84 test executions passed across Spark 3.4, 3.5 (both Scala variants), 4.0, 4.1, and 4.2. An isolated exact-base allocator comparison reproduced all five regression failures; all five pass at head. Shared-task accounting checks also passed.

Broader CI is still running; full shuffle integration was not exercised.

sunchao:
Merged, thanks @ywskycn !

PR #5522, opened 2026-08-28
doc: fix benchmark examples for MacOS

xhumanoid:
regression in example documentation after #2418

### DIFF
--- a/docs/source/contributor-guide/benchmarking_macos.md
+++ b/docs/source/contributor-guide/benchmarking_macos.md
@@ -36,7 +36,7 @@ cargo install tpchgen-cli
 mkdir benchmark_data
 cd benchmark_data
 tpchgen-cli -s 100 --format=parquet
-export $BENCH_DATA=`pwd`
+export BENCH_DATA=`pwd`
 ```
 
 Create a temp folder for spark events emitted during benchmarking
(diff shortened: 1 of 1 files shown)


sunchao (approved):
Reviewed the three-line correction at 6f5dfa37d5b4ca7d29cef328dd8b3efa2ba36b6c. No P1/P2 findings in this change. The export now assigns `BENCH_DATA`, and both benchmark commands point at the directory containing the generated `<table>.parquet` files.

Checked the path contract against [tpchgen-cli 3.0.0](https://github.com/datafusion-contrib/tpcgen-rs/blob/4f6bf4c5ab40511c8fdef5888fc8d022e5e546d7/tpchgen-cli/src/output_plan.rs#L237-L258) and [the benchmark reader](https://github.com/apache/datafusion-benchmarks/blob/cb12c981e6608e0f2dcf919956ada8f1f1622d72/runners/datafusion-comet/tpcbench.py#L32-L48), alongside bounded Bash/dash/zsh controls on Linux and static checks. No data generation, Spark/Comet benchmark, or macOS execution was performed. The CI workflows remain `action_required`; this review does not claim a CI pass.

andygrove (approved):
Thanks @xhumanoid

PR #5526, opened 2026-08-28
feat: support NullType output types in codegen dispatch

### LINKED ISSUE #5525: Support NullType output types in codegen dispatch

Any expression whose output type contains `NullType` is rejected by the codegen dispatch gate and the whole operator falls back to Spark, even though a `NullType` column can only ever hold nulls.

Spark's untyped constructors leave `NullType` children behind, so this hits common literals:

- `map()` → `MapType(NullType, NullType)`
- `map('a', NULL)` → `MapType(StringType, NullType)`
- `array()` → `ArrayType(NullType)`
- `map_from_arrays(array(), array())`, `struct(map())`, and anything nested around them

`CreateMap` and friends are routed through codegen dispatch (`spark/src/main/scala/org/apache/comet/serde/maps.scala:214`), so a projection containing one of these literals is currently kicked out of the native plan with `codegen dispatch: unsupported output type ...`.

Why the gate rejects it today:

- `isSupportedDataType` (`spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala:85`) has no `NullType` case, so it falls through to `false`, recursively for `ArrayType` / `StructType` / `MapType` children.
- `canHandle` applies that same predicate to both the output type (`CometBatchKernelCodegen.scala:120`) and every `BoundReference` input (`CometBatchKernelCodegen.scala:172`).

The rejection is only needed on the input side: `CometScalaUDFCodegen.specFor` has no way to build an `ArrowColumnSpec` for a `NullVector`, so a `NullType` *input* must keep falling back. On the output side the kernel only needs to emit an all-null Arrow `NullVector`, which the rest of the pipeline already understands (`serializeDataType` maps `NullType` to its own type id and `Utils.toArrowField` maps it to `ArrowType.Null`).

Make the type gate asymmetric:

- Accept `NullType` (top-level or nested inside array / struct / map) for the **output** type in `canHandle`, and keep rejecting it for `BoundReference` **inputs**.
- Teach the output emitter (`CometBatchKernelCodegenOutput`) to map `NullType` to `NullVector` and to write it with `setNull` only, without reading a source value.
- Keep the output emitter's type surface in sync with the gate, as the existing doc comments require, so plan-time acceptance never turns into an execute-time exception.
- Update the Scala/Java UDF user guide: `NullType` arguments remain unsupported, `NullType` return types become supported.

Done when a `CometSqlFileTestSuite` fixture such as `SELECT map()`, `SELECT map('a', NULL)`, `SELECT array(map())`, and a `map()` column carried through `ORDER BY` runs without a Spark fallback, and unit tests in `CometCodegenSourceSuite` lock in the output/input asymmetry and the gate/emitter agreement.

### Additional context

No native changes are needed: the `NullType` support in serde and Arrow field conversion already exists; the missing piece is entirely in the JVM codegen gate and output emitter.

I have a patch for this ready and will open a PR referencing this issue.

Assisted-by: Claude Code (claude-fable-5)

grorge123:

Closes #5525.

## Rationale for this change

Untyped constructors such as `map()`, `map(k, NULL)` and `array()` leave `NullType` children in their output type. `CometBatchKernelCodegen.canHandle` rejected any output type containing `NullType`, so such expressions fell back to Spark and took their whole operator with them. Only the input side needs that restriction (`CometScalaUDFCodegen` cannot build a spec for a `NullVector`); on the output side the kernel just has to emit an all-null `NullVector`.

Admitting those outputs then exposed a set of downstream assumptions that `NullType` never arrives, on the JVM IPC paths and in native kernels. This PR closes them and adds a differential sweep so the next one is found here rather than in review.

**Codegen gate.** `canHandle` accepts `NullType` in output types (nested included) while still rejecting it in inputs, and rejects duplicate struct field names, which Arrow's `StructVector` collapses. A `NullType` child is always declared nullable on both sides of the FFI boundary (`Utils.declaredChildNullability`), since native kernels that rebuild a list around the input's child fail on the nullability mismatch.

**JVM IPC paths.** Arrow's `MinorType.NULL` factory rebuilds a `NullType` map key as nullable, which the IPC reader rejects; `Utils.newArrowStreamWriter` (now the only way to build an `ArrowStreamWriter`, scalastyle-enforced) repairs the key on every writer. `VectorSchemaRootAppender` loops forever on a `NullVector` directly under a struct; a broadcast build side with that shape stays on Spark's broadcast (`CometBroadcastExchangeExec` reports it `Unsupported`, naming the columns), since shipping it uncoalesced costs every consuming task one IPC stream per buffer, and `coalesceBroadcastBatches` keeps a defensive bypass on the same predicate. `array<null>` coalesces and stays native.

**Serde gates.** Refused over `NullType`-bearing inputs: `make_array` (single row), `array_union` (drops entries), `array_intersect` (returns the other side's entries; reported `Unsupported` so the codegen dispatcher runs it at every setting, since the `Incompatible` branch has no dispatcher fallback under `allowIncompatible`, and `array_except`'s unsupported element types likewise), `array_repeat` over an input whose native list item is non-nullable (`map_entries`; DataFusion's repeat rebuilds the item as nullable and fails on the mismatch, while a `CreateArray` child is exempt since native `make_array` emits a nullable item whatever Spark's `containsNull` says), `collect_list`/`collect_set` (nested nullability mismatch), `hash`/`xxhash64` (no Null arm), and `map_from_arrays` with a literal array beside a per-row one (the native map kernel reads a scalar list through its first row; pre-existing and independent of `NullType`, found while probing the `NullType` flavour). The null guards in `CometElementAt` (ANSI), `CometArrayAppend`, `CometMapFromArrays`, `CometArraysZip`, `CometCoalesce` and `CometSize` serialize their children into a predicate copy and a THEN branch, and a stateful child diverges from Spark's single evaluation either way it runs: native CASE evaluates the THEN copy on the rows the predicate selected, and a codegen-dispatched child (any lambda) is one cached kernel whose state both copies share. Any non-deterministic child inside a guard is refused, whichever argument is the nullable one (`NullGuard`); `CometSize` builds no guard for a non-nullable child or in legacy mode, where native already answers -1, so those stay native. `CometIf`, `CometCaseWhen` and `CometCoalesce` refuse a `NullType` result, since native CASE merges its branches' rows through Arrow's `merge_n`, which cannot build a `NullArray` with a validity bitmap; native `GetStructField` returns a scalar for a scalar struct input instead of a one-row array that a CASE result builder slices past, and rebuilds a `NullType` field as a fresh `NullArray`, since `element_at` on an out-of-range index hands over a Null child carrying a validity bitmap. `array_union`/`array_intersect`/`array_except` cast both sides to a deeply-nullable element type, since the native set-op kernel asserts identical nested nullability and a lambda variable arrives nullable where a literal field does not. The `coalesce`, set-op and non-`NullType` guard cases are reachable on main; they are included because the sweep found them and the fixes are shared.

Every gate above whose shape Spark's own `doGenCode` can evaluate (`array_repeat`, `array_union`, `make_array`, `IF`/`CASE`/`coalesce`, alongside the already enrolled `array_intersect`/`array_except`) mixes in `CodegenDispatchFallback`, so `Unsupported` routes the expression through the JVM codegen dispatcher and the projection stays in the Comet pipeline rather than falling back to Spark; the dispatcher takes `NullType` as an output but not as an input column, so a gated shape whose `NullType` side arrives as an upstream column still falls back. Enrolling `coalesce` exposed a dispatcher bug: a top-level `struct<.., x: null>` output failed to allocate, since `StructVector`'s constructor builds a `NullableStructWriter` over the field's children and that writer has no arm for a Null child; the output vector now adds its children after construction. Every serde whose support level moved publishes the reason through `getUnsupportedReasons`, so the generated compatibility pages carry the restriction, and the runtime note starts with the published text.

**Native `to_csv`.** Yields NULL for a row that renders to an empty string and reports itself nullable, as Spark does: Spark hands the row to univocity's `writeRowToString` with `skipEmptyLines` on, so a struct with a lone null field (under the default empty `nullValue`) is NULL, not `""`. Pre-existing and independent of `NullType`; the `allowIncompatible` sweep profile found it on Spark 3.4, whose interpreted `StructsToCsv` shows the NULL, while Spark 3.5+ crashes in its own generated code on that NULL (`nullSafeCodeGen` never marks the result null), so the sweep counts the case as invalid in Spark there.

**Native shuffle.** The row shuffle writer's field-major paths gain the `Null` struct field case their row-major path already had, so a struct holding a `NullType` field shuffles through the JVM columnar shuffle instead of panicking, and the writer recreates every `NullType`-bearing builder after each batch, since `NullBuilder::finish` keeps its length and a second writer batch used to panic or miscount; the sweep no longer tolerates any failure.

**Tests.** `CometNullTypeCompositionSuite` sweeps every `NullType` producer under consumers, operators and nesting containers, across ANSI, nullable, non-deterministic and cross-input (a stateful argument beside a nullable one) settings, with the optimizer's null and comparison simplifications excluded so no consumer folds to a literal, and under physical profiles that vary the batching and the exchange path (native batches of two rows, the JVM shuffle's bypass and sort-based writers with and without forced spills, native shuffle, AQE, native columnar-to-row) plus one that opts every registered serde into its native kernel through `allowIncompatible`; each sweep runs only under the profiles that can change its outcome (28 tests, about 8 minutes on an i7-12700K), with operators that partition, join and run a scalar subquery on the value itself, comparing Comet with Spark with no tolerated failures and enforcing floors on the compared and natively executed counts. A registry check fails the suite if a registered array, map, struct or any-type aggregate serde is reached by no template, so a serde added later cannot stay outside the sweep, and a gate check asserts, per gate, that the serde is enrolled in the dispatcher, that its reason is published, and that the query keeps its `CometProjectExec`. `CometInMemoryCacheSuite` round-trips `NullType` columns and children through the Arrow cache serializer. Unit tests in `CometCodegenSourceSuite`, `UtilsSuite` (including an exhaustive oracle for the coalesce bypass rule), `CometJoinSuite`, `CometColumnarShuffleSuite`, `test_pyarrow_udf.py`, and SQL-file witnesses that assert the native plan for every dispatched shape and pin the reason for every remaining fallback.

**Benchmark.** `CometNullTypeColumnsBenchmark` (separate commit) times the consumed `map(id, NULL)` projection and broadcast joins with a `NullType` build side against Spark and the prior fallback path, with an `array<null>` control, at default and small batch sizes and with 2 and 64 broadcast buffers; it checks that every arm returns the same rows and profiles plan, allocation, memory and IPC stream counts per arm. Every case is on par with or faster than the prior path: the gated build sides match it, and the `array<null>` join runs 2.0x Spark at 64 buffers against 1.1x when it fell back.

Spark 4.1 / Scala 2.13, on the head rebased onto `3810936b4`: `CometNullTypeCompositionSuite`, `CometSqlFileTestSuite`, `CometCodegenSourceSuite`, `CometCodegenSuite`, `UtilsSuite`, `CometJoinSuite`, `CometArrayExpressionSuite` and `CometExpressionSuite` pass, plus the `struct_nulls` and `get_struct_field` native unit tests. Earlier revisions: `make test-jvm` green; on the benchmark commit `CometNullTypeCompositionSuite`, `CometSqlFileTestSuite`, `CometCodegenSourceSuite`, `UtilsSuite`, `CometArrayExpressionSuite`, `CometMapExpressionSuite`, `CometJoinSuite`, `CometAggregateSuite` and the hash tests all pass, plus `pytest test_pyarrow_udf.py`. Spark 3.5 / Scala 2.12: `CometNullTypeCompositionSuite`, `CometSqlFileTestSuite`, `UtilsSuite`, `CometCodegenSourceSuite` pass. Spark 3.4 / Scala 2.12: `CometNullTypeCompositionSuite` and `CometCsvExpressionSuite` pass. Disabling any one gate turns the sweep red on the compositions it covers.

Assisted-by: Claude Code (claude-fable-5)

### DIFF
--- a/native/shuffle/src/spark_unsafe/row.rs
+++ b/native/shuffle/src/spark_unsafe/row.rs
@@ -572,6 +572,11 @@ fn append_nested_struct_fields_field_major(
                     }
                 }
             }
+            // A Null field carries no data: every row is null, whether or not the struct is.
+            DataType::Null => {
+                let field_builder = get_field_builder!(struct_builder, NullBuilder, field_idx);
+                field_builder.append_nulls(num_rows);
+            }
             _ => {
                 unreachable!(
                     "Unsupported data type of struct field: {:?}",
@@ -1020,6 +1025,11 @@ fn append_struct_fields_field_major(
                     }
                 }
             }
+            // A Null field carries no data: every row is null, whether or not the struct is.
+            DataType::Null => {
+                let field_builder = get_field_builder!(struct_builder, NullBuilder, field_idx);
+                field_builder.append_nulls(row_end - row_start);
+            }
             _ => {
                 unreachable!(
                     "Unsupported data type of struct field: {:?}",
@@ -1179,9 +1189,7 @@ fn append_columns(
         }
         DataType::Null => {
             let null_builder = downcast_builder_ref!(NullBuilder, builder);
-            for _ in row_start..row_end {
-                null_builder.append_null();
-            }
+            null_builder.append_nulls(row_end - row_start);
         }
         DataType::Timestamp(TimeUnit::Microsecond, _) => {
             append_column_to_builder!(
@@ -1418,6 +1426,14 @@ pub fn process_sorted_row_partition(
             .collect();
         let batch = make_batch(array_refs?, n)?;
 
+        // A finished `NullBuilder` keeps its length; see `recreate_null_type_builders`.
+        recreate_null_type_builders(
+            &mut data_builders,
+            schema,
+            batch_size,
+            prefer_dictionary_ratio,
+        )?;
+
         frozen.clear();
         let mut cursor = Cursor::new(&mut frozen);
 
@@ -1438,6 +1454,37 @@ pub fn process_sorted_row_partition(
         current_checksum.map(|c| c.finalize()),
         ipc_time.value() as i64,
     ))
+}
+
+/// Whether `dt` is `Null` or nests a `Null` anywhere below a list, struct or map.
+fn contains_null_type(dt: &DataType) -> bool {
+    match dt {
+        DataType::Null => true,
+        DataType::List(field) | DataType::LargeList(field) | DataType::Map(field, _) => {
+            contains_null_type(field.data_type())
+        }
+        DataType::Struct(fields) => fields.iter().any(|f| contains_null_type(f.data_type())),
+        _ => false,
+    }
+}
+
+/// Replaces every builder whose type holds a `Null` somewhere, after its batch has been
+/// finished. `NullBuilder::finish` keeps its length (a `NullArray` owns no buffers to hand
+/// over), so such a builder would carry this batch's rows into the next: a top-level Null column
+/// comes out longer than the batch, and a Null struct field longer than its parent panics in
+/// `StructBuilder::finish`. Every other builder resets on finish and is kept.
+fn recreate_null_type_builders(
+    builders: &mut [Box<dyn ArrayBuilder>],
+    schema: &[DataType],
+    batch_size: usize,
+    prefer_dictionary_ratio: f64,
+) -> Result<(), CometError> {
+    for (builder, datatype) in builders.iter_mut().zip(schema.iter()) {
+        if contains_null_type(datatype) {
+            *builder = make_builders(datatype, batch_size, prefer_dictionary_ratio)?;
+        }
+    }
+    Ok(())
 }
 
 fn builder_to_array(
@@ -1503,9 +1550,97 @@ fn make_batch(arrays: Vec<ArrayRef>, row_count: usize) -> Result<RecordBatch, Ar
 
 #[cfg(test)]
 mod test {
+    use arrow::array::StructArray;
     use arrow::datatypes::Fields;
 
     use super::*;
+
+    // The batch loop in `process_sorted_row_partition` reuses its builders. A `NullBuilder`
+    // keeps its length across `finish`, so without `recreate_null_type_builders` the second
+    // batch's Null struct field is twice as long as its parent (a panic in `StructBuilder::
+    // finish`) and a top-level Null column is twice as long as the batch.
+    #[test]
+    fn null_type_builders_start_every_batch_empty() {
+        let struct_type = DataType::Struct(Fields::from(vec![
+            Field::new("v", DataType::Int64, true),
+            Field::new("n", DataType::Null, true),
+        ]));
+        let nested_type = DataType::Struct(Fields::from(vec![Field::new(
+            "s",
+            struct_type.clone(),
+            true,
+        )]));
+        let schema = vec![struct_type.clone(), nested_type.clone(), DataType::Null];
+        assert!(schema.iter().all(contains_null_type));
+        assert!(!contains_null_type(&DataType::List(Arc::new(Field::new(
+            "item",
+            DataType::Int64,
+            true
+        )))));
+
+        let batch_size = 2;
+        let mut builders: Vec<Box<dyn ArrayBuilder>> = schema
+            .iter()
+            .map(|dt| make_builders(dt, batch_size, 1.0).unwrap())
+            .collect();
+
+        for batch in 0..2 {
+            let struct_builder = builders[0]
+                .as_any_mut()
+                .downcast_mut::<StructBuilder>()
+                .unwrap();
+            for row in 0..batch_size {
+                struct_builder
+                    .field_builder::<Int64Builder>(0)
+                    .unwrap()
+                    .append_value(row as i64);
+                struct_builder
+                    .field_builder::<NullBuilder>(1)
+                    .unwrap()
+                    .append_null();
+                struct_builder.append(true);
+            }
+            let nested_builder = builders[1]
+                .as_any_mut()
+                .downcast_mut::<StructBuilder>()
+                .unwrap();
+            for _ in 0..batch_size {
+                let inner = nested_builder.field_builder::<StructBuilder>(0).unwrap();
+                inner
+                    .field_builder::<Int64Builder>(0)
+                    .unwrap()
+                    .append_null();
+                inner.field_builder::<NullBuilder>(1).unwrap().append_null();
+                inner.append(true);
+                nested_builder.append(true);
+            }
+            builders[2]
+                .as_any_mut()
+                .downcast_mut::<NullBuilder>()
+                .unwrap()
+                .append_nulls(batch_size);
+
+            let arrays: Vec<ArrayRef> = builders
+                .iter_mut()
+                .zip(schema.iter())
+                .map(|(builder, dt)| builder_to_array(builder, dt, 1.0).unwrap())
+                .collect();
+            for (array, dt) in arrays.iter().zip(schema.iter()) {
+                assert_eq!(array.len(), batch_size, "batch {batch} of {dt}");
+            }
+            let outer = arrays[0].as_any().downcast_ref::<StructArray>().unwrap();
+            assert_eq!(
+                outer.column(1).len(),
+                batch_size,
+                "batch {batch} Null field"
+            );
+
+            recreate_null_type_builders(&mut builders, &schema, batch_size, 1.0).unwrap();
+            for (builder, dt) in builders.iter().zip(schema.iter()) {
+                assert_eq!(builder.len(), 0, "builder for {dt} after batch {batch}");
+            }
+        }
+    }
 
     #[test]
     fn test_append_null_row_to_struct_builder() {
--- a/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
+++ b/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
@@ -116,6 +116,26 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
   }
 
   /**
+   * Names that repeat within one struct, searched recursively through `dataType`. Spark keeps
+   * duplicate struct field names as distinct positional fields, but Arrow's `StructVector` keys
+   * its children by name (`ConflictPolicy.CONFLICT_REPLACE` by default), so
+   * `initializeChildrenFromFields` collapses the duplicates and the generated ordinal-based child
+   * casts hit a missing or differently typed vector. `CometCreateNamedStruct` refuses the same
+   * shape at the serde level, but whole-expression dispatch never consults that rule for a
+   * `named_struct` nested inside e.g. a `transform` lambda, so [[canHandle]] re-checks here.
+   */
+  private def duplicateStructFieldNames(dataType: DataType): Seq[String] = dataType match {
+    case st: StructType =>
+      val names = st.fieldNames.toSeq
+      val dups = names.diff(names.distinct).distinct
+      if (dups.nonEmpty) dups else st.fields.flatMap(f => duplicateStructFieldNames(f.dataType))
+    case ArrayType(inner, _) => duplicateStructFieldNames(inner)
+    case MapType(keyType, valueType, _) =>
+      duplicateStructFieldNames(keyType) ++ duplicateStructFieldNames(valueType)
+    case _ => Nil
+  }
+
+  /**
    * Mirrors `WholeStageCodegenExec.numOfNestedFields` so [[canHandle]] can reuse
    * `spark.sql.codegen.maxFields`.
    */
@@ -132,12 +152,18 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    * back cleanly rather than crashing the Janino compile at execute time.
    *
    * Checks every `BoundReference`'s data type and the root `expr.dataType` against
-   * [[isSupportedDataType]], rejects aggregates / generators / `Unevaluable`, and gates total
-   * nested-field count on `spark.sql.codegen.maxFields`.
+   * [[isSupportedDataType]] and [[duplicateStructFieldNames]], rejects aggregates / generators /
+   * `Unevaluable`, and gates total nested-field count on `spark.sql.codegen.maxFields`.
    */
   def canHandle(boundExpr: Expression): Option[String] = {
     if (!isSupportedDataType(boundExpr.dataType, allowNullType = true)) {
       return Some(s"codegen dispatch: unsupported output type ${boundExpr.dataType}")
+    }
+    val outputDups = duplicateStructFieldNames(boundExpr.dataType)
+    if (outputDups.nonEmpty) {
+      return Some(
+        s"codegen dispatch: duplicate struct field name ${outputDups.mkString(", ")} " +
+          s"in output type ${boundExpr.dataType}")
     }
     // Mirror WSCG's `spark.sql.codegen.maxFields` gate. Wide schemas blow the generated class's
     // typed input field count, the typed-getter switch, and the constant pool. Refuse here so the
@@ -187,12 +213,23 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
             "(aggregate, generator, or unevaluable)")
       case None =>
     }
-    val badRef = boundExpr.collectFirst {
-      case b: BoundReference if !isSupportedDataType(b.dataType) =>
-        b
-    }
-    badRef.map(b =>
-      s"codegen dispatch: unsupported input type ${b.dataType} at ordinal ${b.ordinal}")
+    boundExpr.collectFirst(Function.unlift(inputRejection))
+  }
+
+  /** Why `expr` cannot be read as a codegen input, if it cannot. */
+  private def inputRejection(expr: Expression): Option[String] = expr match {
+    case b: BoundReference if !isSupportedDataType(b.dataType) =>
+      Some(s"codegen dispatch: unsupported input type ${b.dataType} at ordinal ${b.ordinal}")
+    case b: BoundReference =>
+      val dups = duplicateStructFieldNames(b.dataType)
+      if (dups.isEmpty) {
+        None
+      } else {
+        Some(
+          s"codegen dispatch: duplicate struct field name ${dups.mkString(", ")} " +
+            s"in input type ${b.dataType} at ordinal ${b.ordinal}")
+      }
+    case _ => None
   }
 
   /**
--- a/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
+++ b/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
@@ -87,8 +87,15 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    * single child and the generated writer NPEs on the missing ordinal-1 vector.
    * `CometCreateNamedStruct` declines them on the native path for the same reason, but a struct
    * nested inside a dispatcher-built value (a `CreateMap` value) never reaches that check.
+   *
+   * `NullType` is output-only: [[CometBatchKernelCodegenOutput]] can write an all-null Arrow
+   * `NullVector`, but `CometScalaUDFCodegen.specFor` cannot build an [[ArrowColumnSpec]] for one,
+   * so a `NullType` input (nested or not) has to keep falling back to Spark.
    */
-  def isSupportedDataType(dt: DataType): Boolean = dt match {
+  def isSupportedDataType(dt: DataType): Boolean = isSupportedDataType(dt, allowNullType = false)
+
+  private def isSupportedDataType(dt: DataType, allowNullType: Boolean): Boolean = dt match {
+    case NullType => allowNullType
     case BooleanType | ByteType | ShortType | IntegerType | LongType => true
     case FloatType | DoubleType => true
     case _: DecimalType => true
@@ -96,13 +103,15 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
     case DateType | TimestampType | TimestampNTZType => true
     case dt if isTimeType(dt) => true
     case _: YearMonthIntervalType | _: DayTimeIntervalType | CalendarIntervalType => true
-    case ArrayType(inner, _) => isSupportedDataType(inner)
+    case ArrayType(inner, _) => isSupportedDataType(inner, allowNullType)
     case st: StructType =>
       // `fieldNames` rebuilds an array on each call, so read it once.
       val names = st.fieldNames
       names.distinct.length == names.length &&
-      st.fields.forall(f => isSupportedDataType(f.dataType))
-    case mt: MapType => isSupportedDataType(mt.keyType) && isSupportedDataType(mt.valueType)
+      st.fields.forall(f => isSupportedDataType(f.dataType, allowNullType))
+    case mt: MapType =>
+      isSupportedDataType(mt.keyType, allowNullType) &&
+      isSupportedDataType(mt.valueType, allowNullType)
     case _ => false
   }
 
@@ -127,7 +136,7 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    * nested-field count on `spark.sql.codegen.maxFields`.
    */
   def canHandle(boundExpr: Expression): Option[String] = {
-    if (!isSupportedDataType(boundExpr.dataType)) {
+    if (!isSupportedDataType(boundExpr.dataType, allowNullType = true)) {
       return Some(s"codegen dispatch: unsupported output type ${boundExpr.dataType}")
     }
     // Mirror WSCG's `spark.sql.codegen.maxFields` gate. Wide schemas blow the generated class's
--- a/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
+++ b/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
@@ -82,11 +82,8 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    * Type surface the kernel covers on both input and output sides. Recursive: complex types are
    * supported when their children are.
    *
-   * Duplicate struct field names are excluded, an output-side rule: Arrow addresses a
-   * `StructVector`'s children by name, so `named_struct('x', 10, 'x', 20)` collapses into a
-   * single child and the generated writer NPEs on the missing ordinal-1 vector.
-   * `CometCreateNamedStruct` declines them on the native path for the same reason, but a struct
-   * nested inside a dispatcher-built value (a `CreateMap` value) never reaches that check.
+   * Duplicate struct field names are a separate rule, [[hasDuplicateStructFieldNames]], so that
+   * [[canHandle]] can name them in its reason.
    *
    * `NullType` is output-only: [[CometBatchKernelCodegenOutput]] can write an all-null Arrow
    * `NullVector`, but `CometScalaUDFCodegen.specFor` cannot build an [[ArrowColumnSpec]] for one,
@@ -104,11 +101,7 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
     case dt if isTimeType(dt) => true
     case _: YearMonthIntervalType | _: DayTimeIntervalType | CalendarIntervalType => true
     case ArrayType(inner, _) => isSupportedDataType(inner, allowNullType)
-    case st: StructType =>
-      // `fieldNames` rebuilds an array on each call, so read it once.
-      val names = st.fieldNames
-      names.distinct.length == names.length &&
-      st.fields.forall(f => isSupportedDataType(f.dataType, allowNullType))
+    case st: StructType => st.fields.forall(f => isSupportedDataType(f.dataType, allowNullType))
     case mt: MapType =>
       isSupportedDataType(mt.keyType, allowNullType) &&
       isSupportedDataType(mt.valueType, allowNullType)
@@ -116,24 +109,33 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
   }
 
   /**
-   * Names that repeat within one struct, searched recursively through `dataType`. Spark keeps
-   * duplicate struct field names as distinct positional fields, but Arrow's `StructVector` keys
-   * its children by name (`ConflictPolicy.CONFLICT_REPLACE` by default), so
-   * `initializeChildrenFromFields` collapses the duplicates and the generated ordinal-based child
-   * casts hit a missing or differently typed vector. `CometCreateNamedStruct` refuses the same
-   * shape at the serde level, but whole-expression dispatch never consults that rule for a
-   * `named_struct` nested inside e.g. a `transform` lambda, so [[canHandle]] re-checks here.
+   * Spark keeps duplicate struct field names as distinct positional fields, but Arrow's
+   * `StructVector` keys its children by name (`ConflictPolicy.CONFLICT_REPLACE`), so
+   * `named_struct('x', 10, 'x', 20)` collapses into a single child and the generated
+   * ordinal-based child casts hit a missing or differently typed vector. `CometCreateNamedStruct`
+   * refuses the same shape on the native path, but whole-expression dispatch never consults that
+   * rule for a `named_struct` nested inside e.g. a `transform` lambda or a `CreateMap` value.
    */
-  private def duplicateStructFieldNames(dataType: DataType): Seq[String] = dataType match {
+  private def hasDuplicateStructFieldNames(dt: DataType): Boolean = dt match {
     case st: StructType =>
-      val names = st.fieldNames.toSeq
-      val dups = names.diff(names.distinct).distinct
-      if (dups.nonEmpty) dups else st.fields.flatMap(f => duplicateStructFieldNames(f.dataType))
-    case ArrayType(inner, _) => duplicateStructFieldNames(inner)
-    case MapType(keyType, valueType, _) =>
-      duplicateStructFieldNames(keyType) ++ duplicateStructFieldNames(valueType)
-    case _ => Nil
+      // `fieldNames` rebuilds an array on each call, so read it once.
+      val names = st.fieldNames
+      names.distinct.length != names.length ||
+      st.fields.exists(f => hasDuplicateStructFieldNames(f.dataType))
+    case ArrayType(inner, _) => hasDuplicateStructFieldNames(inner)
+    case MapType(k, v, _) => hasDuplicateStructFieldNames(k) || hasDuplicateStructFieldNames(v)
+    case _ => false
   }
+
+  /** Why `dt` cannot cross the kernel boundary as `side` ("output" or "input"), if it cannot. */
+  private def typeRejection(dt: DataType, allowNullType: Boolean, side: String): Option[String] =
+    if (!isSupportedDataType(dt, allowNullType)) {
+      Some(s"codegen dispatch: unsupported $side type $dt")
+    } else if (hasDuplicateStructFieldNames(dt)) {
+      Some(s"codegen dispatch: unsupported $side type $dt (duplicate struct field name)")
+    } else {
+      None
+    }
 
   /**
    * Mirrors `WholeStageCodegenExec.numOfNestedFields` so [[canHandle]] can reuse
@@ -152,18 +154,13 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    * back cleanly rather than crashing the Janino compile at execute time.
    *
    * Checks every `BoundReference`'s data type and the root `expr.dataType` against
-   * [[isSupportedDataType]] and [[duplicateStructFieldNames]], rejects aggregates / generators /
-   * `Unevaluable`, and gates total nested-field count on `spark.sql.codegen.maxFields`.
+   * [[isSupportedDataType]] and [[hasDuplicateStructFieldNames]], rejects aggregates, generators
+   * and `Unevaluable`, and gates total nested-field count on `spark.sql.codegen.maxFields`.
    */
   def canHandle(boundExpr: Expression): Option[String] = {
-    if (!isSupportedDataType(boundExpr.dataType, allowNullType = true)) {
-      return Some(s"codegen dispatch: unsupported output type ${boundExpr.dataType}")
-    }
-    val outputDups = duplicateStructFieldNames(boundExpr.dataType)
-    if (outputDups.nonEmpty) {
-      return Some(
-        s"codegen dispatch: duplicate struct field name ${outputDups.mkString(", ")} " +
-          s"in output type ${boundExpr.dataType}")
+    typeRejection(boundExpr.dataType, allowNullType = true, "output") match {
+      case Some(reason) => return Some(reason)
+      case None =>
     }
     // Mirror WSCG's `spark.sql.codegen.maxFields` gate. Wide schemas blow the generated class's
     // typed input field count, the typed-getter switch, and the constant pool. Refuse here so the
@@ -218,17 +215,9 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
 
   /** Why `expr` cannot be read as a codegen input, if it cannot. */
   private def inputRejection(expr: Expression): Option[String] = expr match {
-    case b: BoundReference if !isSupportedDataType(b.dataType) =>
-      Some(s"codegen dispatch: unsupported input type ${b.dataType} at ordinal ${b.ordinal}")
     case b: BoundReference =>
-      val dups = duplicateStructFieldNames(b.dataType)
-      if (dups.isEmpty) {
-        None
-      } else {
-        Some(
-          s"codegen dispatch: duplicate struct field name ${dups.mkString(", ")} " +
-            s"in input type ${b.dataType} at ordinal ${b.ordinal}")
-      }
+      typeRejection(b.dataType, allowNullType = false, "input")
+        .map(reason => s"$reason at ordinal ${b.ordinal}")
     case _ => None
   }
 
--- a/spark/src/test/scala/org/apache/comet/CometNullTypeCompositionSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometNullTypeCompositionSuite.scala
@@ -1,0 +1,728 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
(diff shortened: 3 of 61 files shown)


sunchao (commented):
Source review of this pinned change found the two newly admitted broadcast failure paths below. Neither example was executed locally. The supplied snapshot has four action-required workflows and no passing checks; the author-reported local test results were not independently verified.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
```diff
@@ -117,7 +126,7 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    * nested-field count on `spark.sql.codegen.maxFields`.
    */
   def canHandle(boundExpr: Expression): Option[String] = {
-    if (!isSupportedDataType(boundExpr.dataType)) {
+    if (!isSupportedDataType(boundExpr.dataType, allowNullType = true)) {
```
sunchao: [P2] Handle NullType map values before enabling native broadcast

A required build-side projection such as `map(id, NULL)` over primitive Parquet input now passes this gate and can enter a normal hinted Comet broadcast hash join with AQE disabled. `CometBroadcastExchangeExec` unconditionally calls `Utils.coalesceBroadcastBatches`, which creates an empty target and appends the first batch. In the pinned Arrow Java 18.3.0, the map-entries struct appender grows capacity before updating child counts, but its `NullVector` value child reports capacity equal to its still-zero count and `reAlloc()` does nothing. The struct therefore never reaches the positive entry count while its key and validity buffers keep growing until allocation or size failure. Finalizing the source batch does not fix this separate target. BASE rejects this output and leaves the intermediate project/broadcast in Spark. Please guard or fix coalescing before admitting this shape. This is source-derived, not an executed reproduction.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
```diff
@@ -81,18 +81,27 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
...
     case DateType | TimestampType | TimestampNTZType => true
     case dt if isTimeType(dt) => true
     case _: YearMonthIntervalType | _: DayTimeIntervalType | CalendarIntervalType => true
-    case ArrayType(inner, _) => isSupportedDataType(inner)
-    case st: StructType => st.fields.forall(f => isSupportedDataType(f.dataType))
-    case mt: MapType => isSupportedDataType(mt.keyType) && isSupportedDataType(mt.valueType)
+    case ArrayType(inner, _) => isSupportedDataType(inner, allowNullType)
+    case st: StructType => st.fields.forall(f => isSupportedDataType(f.dataType, allowNullType))
+    case mt: MapType =>
+      isSupportedDataType(mt.keyType, allowNullType) &&
```
sunchao, line 103: [P2] Preserve non-nullable NullType keys through broadcast IPC

Allowing NullType map keys also admits the valid empty result `transform_values(map(), (k, v) -> id)` over a primitive column when `spark.sql.legacy.createEmptyCollectionUsingStringType=false`. Retaining that result on a normal hinted native broadcast build side with AQE disabled exposes a separate failure. The first generated export preserves the non-nullable key field, but ordinary native-to-Java reimport uses Arrow 18.3.0's `Types.NULL` factory, which constructs `NullVector(name)` and changes its field to nullable. Runtime `getField()` reconstruction and IPC preserve that flag; the broadcast reader then throws `Map data key type should be a non-nullable` during schema initialization, before reading any entries. Zero entries do not avoid this check. The enclosing TransformValues remains nonfoldable on the inspected maintained Spark 3.5/4.0 paths; BASE rejects its output and retains Spark execution. Please preserve the key field across reimport/IPC or retain fallback for this shape. Source-derived, not an executed reproduction.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
grorge123: Fixed: `coalesceBroadcastBatches` now uses `hasNestedNullType` to check the schema, and ships the batches uncoalesced when a `NullType` sits inside a struct or map.

Tests: `UtilsSuite` ("coalesceBroadcastBatches ships maps with nested NullType uncoalesced") checks that such batches come back uncoalesced, and `CometJoinSuite` ("Broadcast HashJoin with NullType map columns on the build side") runs a hinted broadcast hash join with `map(_1, NULL)` and `transform_values(map(), (k, v) -> _1)` on the build side, with AQE on and off.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
grorge123: Since Arrow 18.3.0's `MinorType.NULL.getNewVector` is `new NullVector(field.getName())`, every `NullType` map key comes back nullable after any vector is created for it. We added `Utils.withNonNullableMapKeys`, which repairs the key flag, route all IPC writers through `Utils.newArrowStreamWriter` so the repair is applied in one place, and added a scalastyle rule as a lint check for `new ArrowStreamWriter`. The same repair is applied in `CometArrowStream.actualFieldOf` (the stream schema handed to native) and in `CometArrowPythonRunnerBase` before it builds its destination struct.

Tests: `UtilsSuite` pins the Arrow behaviour (asserts the key comes back nullable, so the workaround can be dropped once Arrow fixes it), checks `withNonNullableMapKeys` restores the flag, round-trips a `NullType` map key through `serializeBatches`, and checks `newArrowStreamWriter` keeps/returns the root a later row count must be set on. `CometJoinSuite` covers the broadcast path, `CometColumnarShuffleSuite` covers `Map[NullType, _]` and `Map[_, NullType]` through the JVM columnar shuffle, and `test_pyarrow_udf.py` covers `mapInArrow` with `Map[NullType, _]` input on Spark 4.1.3 in both accelerated and fallback modes. Also verified under the `spark-3.5` profile.

sunchao (commented):
Re-reviewed `db1768fa2aea540821391c7aef4cff2a4b8c43fe`. The two earlier witnesses are fixed in source. Two source-derived P2s remain; neither was executed here.

**[P2] Reject duplicate names in newly admitted struct outputs.** The [output gate](https://github.com/apache/datafusion-comet/blob/db1768fa2aea540821391c7aef4cff2a4b8c43fe/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala#L129) now admits `transform(array(id), x -> named_struct('a', x, 'a', NULL))` over an `INT` input. Whole-expression ArrayTransform dispatch skips the inner CreateNamedStruct duplicate-name fallback. Spark retains both fields, but Arrow 18.3.0's default struct policy replaces the first `a` with the later NullVector. Generated output setup still casts child ordinal 0 to IntVector, so the query fails before rows are written. BASE rejected this Null-bearing output. Please retain fallback for duplicate names recursively, or allocate children without losing their ordinals.

**[P2] Keep coalescing safe plain null lists.** The [new broadcast bypass](https://github.com/apache/datafusion-comet/blob/db1768fa2aea540821391c7aef4cff2a4b8c43fe/spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala#L500) also rejects plain `array<null>`, although its ListVector appender grows offsets/validity independently of the Null child and finalizes the child count. An already supported build payload such as `IF(rand(17L) < 0.5, array(NULL), array(NULL,NULL))`, selected after a join over a repartitioned broadcast build, therefore loses coalescing. Returning B original buffers makes each of P consuming tasks open B compression/IPC streams and schema roots instead of one: B×P setups rather than P. This is an operation-count consequence, not a measured speed ratio. Keep the struct/map protection, but allow the safe plain-list case to coalesce.

grorge123:
We re-applied the representation-level serde restriction inside canHandle.
Besides that, we found another problem: filter(array(), ...) and map_filter(map(), ...) leave containsNull = false on a NullType child, and downstream native kernels that rebuild a list around that input (map_entries, array_repeat, slice) panic on the nullability mismatch. We fixed it by always declaring a NullType child nullable on both sides of the FFI boundary.

spark/src/test/scala/org/apache/spark/sql/comet/util/UtilsSuite.scala
```diff
@@ -197,4 +212,270 @@ class UtilsSuite extends CometTestBase {
...
+        CometArrowConverters
+          .rowToArrowBatchIter(
+            Iterator.fill(numRows)(InternalRow(value)),
+            schema,
+            numRows,
+            "UTC",
+            CometArrowAllocator)
+          .next()
+      }
+      val bufs = Utils.serializeBatches(batches.iterator).map(_._2).toSeq
```
sunchao: [P2] Materialize serialization before closing the fixture batches

On Scala 2.12, `Iterator.toSeq` produces a lazy `Stream`: the first batch is serialized here, but the tail is not. The next line closes all three input batches. Closing their ListVectors clears the later value counts, so the remaining serializations construct zero-row roots and this fixture supplies 4 + 0 + 0 rows instead of 12. This matches the `array<void>` assertion in the [Spark 3.4 / Scala 2.12 CI job](https://github.com/apache/datafusion-comet/actions/runs/33394656333/job/99563322366). Please consume the iterator eagerly, for example with `toVector`, before closing the inputs. This is a fixture lifetime defect. That CI failure does not establish production coalescer row loss. I did not rerun the test.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
```diff
@@ -113,13 +142,19 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
...
    *
    * Checks every `BoundReference`'s data type and the root `expr.dataType` against
-   * [[isSupportedDataType]], rejects aggregates / generators / `Unevaluable`, and gates total
-   * nested-field count on `spark.sql.codegen.maxFields`.
+   * [[isSupportedDataType]] and [[duplicateStructFieldNames]], rejects aggregates / generators /
+   * `Unevaluable`, and gates total nested-field count on `spark.sql.codegen.maxFields`.
    */
   def canHandle(boundExpr: Expression): Option[String] = {
-    if (!isSupportedDataType(boundExpr.dataType)) {
+    if (!isSupportedDataType(boundExpr.dataType, allowNullType = true)) {
```
sunchao: [P2] Preserve single evaluation of nullable stateful outputs

With codegen dispatch and ANSI enabled, consider `element_at(transform(IF(monotonically_increasing_id() % 2 = 0, array(id), CAST(NULL AS array<bigint>)), x -> named_struct('id', x, 'n', NULL)), 1)` over a native LONG batch containing `id=0,1,2,3`. Spark evaluates the transform once per row, retaining the struct for `id=2`. This gate now admits that Null-bearing output, but the existing ElementAt ANSI conversion serializes its left subtree into both the CASE predicate and the lookup. Native CASE tests all four rows, then reevaluates the transform on the two selected rows. The second selected row receives an odd counter and becomes NULL. Please materialize the left value once or retain Spark fallback for this composition. Current BASE rejects this output. The previous reviewed head admitted it but lacked the duplicated-left CASE, so the newly failing combination is in this increment. Source-derived witness, not executed.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
```diff
@@ -81,21 +81,50 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
...
+   *
+   * `NullType` is output-only: [[CometBatchKernelCodegenOutput]] can write an all-null Arrow
+   * `NullVector`, but `CometScalaUDFCodegen.specFor` cannot build an [[ArrowColumnSpec]] for one,
+   * so a `NullType` input (nested or not) has to keep falling back to Spark.
    */
-  def isSupportedDataType(dt: DataType): Boolean = dt match {
+  def isSupportedDataType(dt: DataType): Boolean = isSupportedDataType(dt, allowNullType = false)
+
+  private def isSupportedDataType(dt: DataType, allowNullType: Boolean): Boolean = dt match {
+    case NullType => allowNullType
```
sunchao, line 92: [P2] Preserve row count for newly admitted all-null array producers

With codegen dispatch enabled, `array(aggregate(array(id), NULL, (acc, x) -> NULL))` over a live LONG input now sends a NullArray of the batch length to native `make_array`. The locked implementation's all-Null branch builds one list containing those nulls, rather than one list per row. With N > 1 input rows, the physical scalar-function row-count guard therefore raises an error for length 1 versus N. The ordinary nonfoldable ArrayAggregate survives the inspected Spark 3.5/4.0 optimizer rules. A folded `array(NULL)` or a one-row batch does not exercise this case. Please preserve batch cardinality in the consumer or retain fallback for this composition. BASE rejects this producer and stays in Spark. This is a previously unreported PR admission bug already present at the prior reviewed head, not a new dependency change. Source-derived, not executed.

sunchao (commented):
Re-reviewed `1d0a6f493bbd60453234748d548c85d77348a98c`. The previous duplicate-field, map-key schema, and broadcast-bypass issues are addressed in the inspected source. The five additional P2 cases below remain.

Could you add a small microbenchmark for consumed `map(id, NULL)` projections and the remaining struct/map broadcast bypass, with a plain `array<null>` control? Compare this head with Spark and safe prior paths using matched versions, settings, and warmup. Please confirm result equality and the executed native/fallback plans, and report throughput, allocations, peak/retained memory, and IPC setup counts for small/default batches and few/many build buffers.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
sunchao: [P2] Preserve the item field when repeating map entries

The [NullType-leaf normalization](https://github.com/apache/datafusion-comet/pull/5526#issuecomment-5478130993) still leaves `array_repeat(map_entries(map(id, NULL)), 2)` failing over a non-null native LONG input with codegen dispatch enabled. `map_entries` produces a list whose struct item is non-nullable. The registered SparkArrayRepeat wrapper delegates to native repeat, which rebuilds the inner list with a nullable item while its outer field still expects the original non-nullable item type. Arrow's ListArray constructor compares those nested fields, including nullability, and rejects the mismatch. The item here is a struct, so normalizing direct NullType leaves does not repair it. Please preserve that item field or keep this newly admitted composition in Spark. BASE rejects the Null-bearing map. The same admission already existed at the previous reviewed head. Source-derived, not executed.

sunchao: [P2] Retain real NULL entries in newly admitted array unions

With codegen dispatch enabled and `spark.sql.legacy.createEmptyCollectionUsingStringType=false`, `array_union(transform(array(id), x -> NULL), array())` over a live LONG input should return `[NULL]` for each row. The new gate admits the transform's List<Null> output, but native union treats `left.value_type().is_null()` as a reason to return `distinct(right)`, discarding the left list's actual entries. The empty right list is accepted by the locked Arrow row converter, so the result is `[]`, not `[NULL]`. The inspected Spark 3.5/4.0 type guards explicitly allow NullType here and their union implementation retains one NULL. Please handle Null elements using the list offsets or retain fallback for this composition. BASE refuses the producer. This previously unreported admission bug was already possible at the prior reviewed head. Source-derived, not executed.

spark/src/test/scala/org/apache/spark/sql/comet/util/UtilsSuite.scala
grorge123: Fixed: the serialization is forced with `toVector` before the inputs are closed, so the Scala 2.12 Stream laziness no longer drops the tail batches. Verified with UtilsSuite under `-Pspark-3.5` (Scala 2.12), which reproduces the CI environment.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
grorge123: Fixed: `CometElementAt` keeps a non-deterministic nullable collection in Spark under ANSI, since the null guard's two copies each hold their own kernel state and the THEN copy only sees the CASE-selected rows. Reproduced the witness first (Comet returned `[null]` where Spark keeps `[[2,null]]`), and it now matches. The same guard is built by `CometSize` (non-legacy mode), `CometArrayAppend`, `CometMapFromArrays` and `CometCoalesce`, so they share the gate (`NullGuard`); `CometSize` drops the guard in legacy mode, where native already answers -1. The coalesce case is reachable on main without NullType (`coalesce(IF(monotonically_increasing_id() % 2 = 0, array(id), NULL), array(id))` NPEs in columnar-to-row because the result is declared non-nullable); it is included since the sweep found it and the fix is one line on the shared guard.

Tests: expect_fallback queries in element_at_ansi.sql, map_from_arrays.sql and coalesce.sql, plus the non-deterministic dimension of CometNullTypeCompositionSuite.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
grorge123: Fixed: `CometCreateArray` keeps non-foldable NullType arguments in Spark - `make_array`'s all-Null branch (`SingleRowListArrayBuilder`) collapses the batch to one row, reproduced as "Expected: 4, Got: 1". All-literal NULLs arrive as scalars and broadcast correctly, so `array(NULL)` literals stay native.

Tests: expect_fallback query in create_array.sql.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
grorge123: Fixed: a dedicated `CometArrayRepeat` keeps `containsNull = false` non-NullType items in Spark, since DataFusion's repeat rebuilds the item field as nullable (reproduced the ListArray nested-field rejection). A NullType item is declared nullable on the FFI boundary and matches the rebuild, so plain `array<null>` stays native.

Tests: expect_fallback query in array_repeat.sql.

grorge123: Fixed: `CometArrayUnion` keeps NullType-element unions in Spark - the set-op kernel's `value_type().is_null()` branch returns `distinct(other side)`, reproduced as `[]` where Spark keeps `[NULL]`. The existing `array_union(array(), array())` / `array_union(array(), array(NULL))` queries move to expect_fallback with it (their operands are Null-typed; the empty cases are benign natively but the gate cannot see emptiness statically).

Tests: expect_fallback queries in array_union.sql, including the non-empty witness.

grorge123:
Beyond the five cases above, this revision adds `CometNullTypeCompositionSuite`, a differential sweep of the admitted shape space: every non-foldable NullType producer (`transform(array(id), x -> NULL)`, `map(id, NULL)`, `transform_values(map(), ...)`, `map_entries(map(id, NULL))`, `named_struct(..., NULL)`, an all-NULL `aggregate`) under every consumer, operator and nesting container Spark accepts, across ANSI on/off, a nullable non-deterministic wrapper, and native columnar-to-row on/off, comparing Comet with Spark.

The non-NullType flavour of the same item-field problem (`slice(map_entries(map(k, v)), ...)`, `slice(reverse(arrays_zip(...)))`, the #4789 nested-nullability contract) reproduces on main and is outside this PR's admission. The one native gap the sweep tolerates is also pre-existing: Comet's row shuffle writer has no case for a `Null` struct field (`Unsupported data type of struct field: Null`), matched by signature so every other failure stays red.

Assisted-by: Claude Code (claude-fable-5)

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
sunchao: [P2] Residual stateful guard case across different arguments

At `79fc84008a7dca16a4c24f28036649c0af2603b0`, the original nullable `element_at` witness is fixed, but the shared guard still misses a cross-input case. This is source-derived, not an executed reproduction. With a native LONG input `t` containing ordered ids 0, 1, 2, 3 in one partition/batch and codegen dispatch enabled, consider `SELECT id, arrays_zip(transform(array(id), x -> named_struct('i', monotonically_increasing_id(), 'n', NULL)), IF(id % 2 = 0, array(id), CAST(NULL AS ARRAY<BIGINT>))) AS z FROM t`. The first array is non-nullable and nondeterministic, while the second is nullable and deterministic, so neither matches this predicate. Spark evaluates the first array even on odd rows where the zip is NULL. `CometArraysZip.convert` instead puts the value-producing kernel inside a CASE whose THEN batch contains only even rows, changing the stateful child's evaluation sequence. The base rejected this NullType-bearing producer. This guard therefore misses a composition newly admitted by the PR. Could the guard preserve one evaluation before filtering or retain fallback when a nullable sibling filters a stateful child, with coverage for nullability and nondeterminism occurring in different arguments?

spark/src/test/scala/org/apache/comet/CometNullTypeCompositionSuite.scala
```diff
@@ -0,0 +1,543 @@
...
+        var knownGap = 0
+
+        for ((label, query) <- queries) {
+          Try(rowsOf(query, cometEnabled = false, ansi, nativeColumnarToRow)).toOption match {
+            case None =>
+              skipped += 1
+            case Some((sparkRows, _)) =>
+              compared += 1
+              Try(rowsOf(query, cometEnabled = true, ansi, nativeColumnarToRow)) match {
+                case Failure(e) if tolerated.exists(causeText(e).contains) =>
```
sunchao: [P2] Do not suppress shuffle regressions newly admitted by this PR

Source-derived, not executed. Over nonempty primitive Parquet `t`, use `SELECT /*+ REPARTITION(3, c) */ c FROM (SELECT transform(array(id), x -> named_struct('v', x, 'n', NULL)) AS c FROM t)` with native scan/codegen dispatch, CometShuffleManager, Comet shuffle enabled, shuffle mode `auto`, `spark.comet.shuffle.convertFromSparkPlan.enabled=false`, AQE off and JVM columnar-to-row. At the base the producer falls back, and the disabled conversion leaves its exchange in Spark. This PR admits a CometProject, so the array partition key rejects native shuffle but falls through to JVM columnar shuffle. The unsafe-row list writer then sends its struct's Null field to `append_field`, which reaches exactly the panic waived here. The writer defect is inherited, but its reachability under these settings is new. Please retain fallback for that schema or support Null struct fields, and do not waive failures that passed at the base under the same settings. With conversion enabled the base can also fail, so that is not a valid regression control.

sunchao (commented):
The five earlier concrete witnesses are addressed in the inspected source at `79fc84008a7dca16a4c24f28036649c0af2603b0`. A residual cross-input stateful-guard case remains in the [existing discussion](https://github.com/apache/datafusion-comet/pull/5526#discussion_r3921234624), and one newly reachable shuffle failure is detailed inline.

This pass used source analysis only. I did not run a Spark/Comet query, test or runtime reproduction. The current-head snapshot has no check runs or workflow runs, so CI validation is unavailable.

spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
grorge123: Fixed: reproduced the witness first (Comet returned `[[4,null],0]` and `[[5,null],2]` where Spark keeps `[[0,null],0]` and `[[2,null],2]`), and the non-NullType flavour diverges the same way, so this is the guard's rule being too narrow rather than a NullType path. The witness's stateful child is a lambda, which runs through the JVM codegen dispatcher; its kernel cache is keyed by the serialized expression, so the guard's predicate copy and THEN branch run one kernel instance and share its counter: the predicate consumes it for the whole batch and the THEN branch continues from there. A natively evaluated stateful child gets its own instance per copy, but native CASE evaluates the THEN branch on the rows the predicate selected, so it diverges as soon as the guard filters. `NullGuard` now refuses any non-deterministic child inside a guard, whichever argument is the nullable one, and `CometArraysZip`, `CometElementAt`, `CometArrayAppend`, `CometMapFromArrays`, `CometCoalesce` and `CometSize` share it; `CometSize` builds no guard at all for a non-nullable child (or in legacy mode), so `size(filter(arr, x -> x < monotonically_increasing_id()))` is evaluated once and stays native.

Tests: expect_fallback witness in arrays_zip.sql; `CometNullTypeCompositionSuite` gains a cross-input sweep that puts a non-nullable stateful producer (including one whose length records the counter) beside a nullable deterministic sibling under every consumer, and a nullable deterministic sweep so the guards' ELSE branches run natively; the `size` witnesses are in `CometArrayExpressionSuite`.

spark/src/test/scala/org/apache/comet/CometNullTypeCompositionSuite.scala
grorge123: Fixed: the exact query above does not fail at this head (reproduced under the listed settings: `CometProject` into `CometColumnarShuffle`, results match), because a list of structs goes through the writer's row-major `append_field`, which has a `Null` arm. The shape that did panic is a top-level struct column with a `Null` field, e.g. `REPARTITION(3) named_struct('v', id, 'n', NULL)`, which takes the field-major paths that lacked the arm; that was the failure the sweep tolerated. Both field-major paths now handle `Null` struct fields (every row null, as the row-major path does), and the sweep tolerates nothing: every producer under `repartition`, shuffle join, group-by and sort compares against Spark.

Tests: `CometNullTypeCompositionSuite` operator and nesting sweeps with the waiver removed.

sunchao: You're right. Correction to my [P2] example: at `79fc8400`, the array-of-struct path reaches [`append_field`](https://github.com/apache/datafusion-comet/blob/79fc84008a7dca16a4c24f28036649c0af2603b0/native/shuffle/src/spark_unsafe/row.rs#L272-L275), which already has a `DataType::Null` arm. My assertion that this path reaches the unsupported-struct-field panic was incorrect, and I withdraw that specific witness. Thanks for checking it.

I am reviewing the two field-major fixes and removal of the sweep waiver separately at `496c3f7f`. This correction is based on the pinned source, not an executed SQL reproduction.

native/shuffle/src/spark_unsafe/row.rs
```diff
@@ -1020,6 +1027,13 @@ fn append_struct_fields_field_major(
                     }
                 }
             }
+            // A Null field carries no data: every row is null, whether or not the struct is.
+            DataType::Null => {
+                let field_builder = get_field_builder!(struct_builder, NullBuilder, field_idx);
+                for _ in row_start..row_end {
```
sunchao: [P2] Reset NullType builders before reusing them for another batch

These new arms handle the first batch, but `process_sorted_row_partition` creates its builders outside the batch loop and `builder_to_array` calls `finish()` without replacing them. In the pinned Arrow 58.4.0, `NullBuilder::finish()` leaves its length unchanged. With the exact current conversion functions, two consecutive two-row batches of `struct<v:bigint,n:void>` panic on the second finish because the struct has length 2 while its Null child has length 4. The single-batch and typed-null two-batch controls pass. A nested struct fails the same way.

`SpillSorter` sends an entire destination partition to this native call, so one call can exceed `spark.comet.shuffle.jvm.batchSize`. The PR now admits producers such as `element_at(transform(array(id), x -> named_struct('v', x, 'n', NULL)), 1)` over native primitive input. With CometShuffleManager, JVM shuffle mode, `spark.comet.shuffle.convertFromSparkPlan.enabled=false`, and AQE off, source tracing shows their exchange can enter this writer where the base retained Spark fallback.

Could we reset or recreate Null-containing builders between batches, or retain fallback, and add coverage spanning more than one writer batch? I reproduced the builder failure in an isolated harness using production conversion source and pinned dependencies. The SQL/planner route was source-traced, not executed end to end.

sunchao (commented):
Re-reviewed `496c3f7f` against `ef62b463`. The broader nondeterminism guard addresses the cross-input case, and the new field-major NullType arms pass the single-batch controls. I found one additional P2 in builder reuse across batches, detailed inline.

I independently reran the exact-source native component harness: 8 tests passed and 3 failed on the same NullBuilder reuse defect. This is not a full Spark, JNI, or shuffle execution. The planner route is source-traced. Four workflows remain `action_required`, with no head or merge check results in the current snapshot. No fresh benchmark was run.

native/shuffle/src/spark_unsafe/row.rs
grorge123: Fixed: reproduced first with the sort-based writer (`spark.shuffle.sort.bypassMergeThreshold=0`, `spark.comet.shuffle.jvm.batchSize=2`, jvm shuffle mode, `convertFromSparkPlan.enabled=false`, AQE off, 16 rows into 2 partitions): `named_struct('v', id, 'n', NULL)` panics in `StructBuilder::finish` with `(2 != 4)` on the second batch, as you describe, and so do a nested struct and `element_at(transform(array(id), x -> named_struct('v', x, 'n', NULL)), 1)`. The same reuse defect hits every other Null-bearing shape: a top-level `NULL` column fails the batch's row-count check, `map(id, NULL)` fails "keys and values have unequal length", and `array<null>` comes back with the wrong row count. Base control at `ef62b463` under the same settings: the top-level `NULL` column and `array(named_struct('v', id, 'n', NULL))` fail there too (row-count check and the same `(2 != 4)` panic through the row-major arm), the struct shapes hit the field-major `unreachable` instead, and `map(id, NULL)`, `array<null>` and the `element_at` shape stayed in Spark on the base, so those three are admitted by this PR. `NullBuilder::finish` keeps its length in arrow 58.4.0 (a `NullArray` owns no buffers to hand over), so `process_sorted_row_partition` now recreates every builder whose type holds a `Null` anywhere (`contains_null_type`) after each batch; the other builders reset on finish and are kept. The bypass writer sends at most one batch per native call, which is why the single-batch controls and the sweep's `REPARTITION(3)` never saw it.

Tests: `null_type_builders_start_every_batch_empty` in row.rs drives two batches through the production `make_builders` / `builder_to_array` / recreate path for a Null field, a nested one and a top-level Null column; `CometColumnarShuffleSuite` "columnar shuffle spanning several native writer batches with NullType columns" runs the seven shapes above through `REPARTITION(300)` with `jvm.batchSize=2` and the spill threshold lifted, under both AQE settings, comparing with Spark.

grorge123:
Rebased onto main and pushed two commits on top of the fix.

**The struct/map broadcast bypass is gone.** Building the benchmark showed it was slower than the path it replaced: with the bypass in, a `map<bigint, null>` build side split into 64 uncoalesced buffers over 8 consumer tasks opened 512 IPC streams (one per buffer per task, `CometBatchRDD.compute`) and ran 0.9x Spark where the fallback path ran 1.2x, a cost that grows with buffers x tasks. The first commit replaces it with a planner gate: a build side with a `NullType` directly under a struct or map entry stays on Spark's broadcast (`CometBroadcastExchangeExec.getSupportLevel` reports it `Unsupported`, naming the columns), which is what happened before the dispatcher admitted `NullType` outputs. `array<null>` is insulated by the list and still coalesces natively. The gate and the coalescer's now defensive bypass share one predicate, `Utils.hasNullTypeUnderStruct`, so both lift together once Arrow's appender can grow a `NullVector` under a struct; `UtilsSuite` runs the real appender over every shape and pins the predicate to exactly the hanging ones, and `CometJoinSuite` asserts the fallback reason and that the plan carries Spark's `BroadcastExchangeExec`.

**The benchmark** is the second commit: `CometNullTypeColumnsBenchmark` (`SPARK_GENERATE_BENCHMARK_FILES=1 make benchmark-org.apache.spark.sql.benchmark.CometNullTypeColumnsBenchmark`). It covers the cases, arms, checks and dimensions you listed and nothing else, on Spark's default settings. Numbers below are from one run on this head: Spark 4.1 / Scala 2.13, JDK 17, `local[1]`, i7-12700K; Spark's `Benchmark` harness with its 2s warmup and 2+ timed iterations per arm, all arms in one session.

*Cases.* Consumed `map(c1, NULL)` projection and a `transform` to `array<null>` control, each over 10M rows; a hinted broadcast hash join (1M-row probe) whose build side projects an `array<null>` control (coalesces), a `map<bigint, null>` and a struct with a NULL field (the two gated shapes), with the build sized for 2 and for 64 broadcast buffers (16K and 512K rows: the native scan emits 8192-row batches and the broadcast collects one buffer per batch).

*Arms.* Spark; Comet (this head); Comet with the codegen dispatcher off, i.e. the prior path where a NullType projection falls back and takes its operator with it; and the two Comet arms again with `spark.comet.batchSize=1024`.

*Result equality and plans.* The benchmark computes an order-independent row digest of every arm's result and fails on any difference; all 8 cases report identical results across the 5 arms. It records, per arm, the first non-Comet operator of the final (post-AQE) executed plan: the Comet arm is fully native in every case except the two gated shapes, where it falls back at `HashAggregate` over Spark's broadcast exactly like the prior-path arm; the prior-path arm falls back at `Project` for the dispatched projections and at `HashAggregate` for the `array<null>` join.

*Throughput* (best time, ms):

| case | Spark | Comet | prior path |
|---|---|---|---|
| `map(c1, NULL)` consumed | 508 | 514 (1.0x) | 497 |
| `array<null>` control | 658 | 537 (1.2x) | 642 |
| join, 2 buffers: `array<null>` (control, coalesced) | 74 | 51 (1.4x) | 48 |
| join, 2 buffers: `map<bigint, null>` (Spark broadcast) | 63 | 45 (1.4x) | 44 |
| join, 2 buffers: struct (Spark broadcast) | 58 | 41 (1.4x) | 40 |
| join, 64 buffers: `array<null>` (control, coalesced) | 195 | 99 (2.0x) | 175 |
| join, 64 buffers: `map<bigint, null>` (Spark broadcast) | 272 | 241 (1.1x) | 247 |
| join, 64 buffers: struct (Spark broadcast) | 212 | 190 (1.1x) | 190 |

The gated shapes match the prior path at both sizes; the dispatched `map(c1, NULL)` projection is within noise of it (stdev 15 ms); the `array<null>` cases are where staying native pays.

*IPC setup counts.* Every consuming task decodes every broadcast buffer, so streams = buffers x consumer tasks, both read back from the executed plan and the consuming stage's task count. The `array<null>` build side coalesces 2 and 64 batches into 1 buffer, so its consumer opens 1 stream at either size; the gated shapes open none (Spark broadcast).

*Allocations and memory* (one profiled execution per arm after warmup, JVM-wide). JVM allocation for the joins is 42-556 MB on the Comet arms against 62-563 MB on Spark; the two projections allocate 2.2-2.5 GB on every arm (the row conversion of 10M rows). Heap peak is lower on the Comet arms in every case. Retained heap after GC and retained Arrow memory are within GC jitter (Arrow retained is 0 everywhere). Native join build memory, from the join's `build_mem_used` metric, is 0.3 MB at 2 buffers and 8 MB at 64, identical between default and small batches; Spark's `peakExecutionMemory` is 1-68 MB.

*Small batches.* `spark.comet.batchSize=1024` costs the row-conversion-heavy projections 0.8x of the default on both Comet arms and changes nothing for the joins. Two observations: the native Parquet scan emits 8192-row batches whatever this setting is, so the broadcast buffer count is set through the build size; and the small-batch projection arms peak at 1.6 GB of heap against 0.64 GB at the default, since the row conversion allocates per batch.

Assisted-by: Claude Code (claude-fable-5)

andygrove (commented):
Several of the new `Unsupported` branches sit on serdes that do not mix in `CodegenDispatchFallback`, so unlike the `array_intersect` and `array_except` cases you reason about in the comments, they never reach the JVM dispatcher and instead drop the whole projection back to Spark. `CometArrayRepeat` is the one I would look at first, because its gate fires for any element that is itself an array Spark typed with `containsNull = false`, so something like `array_repeat(array(c), 2)` over a non-nullable column now leaves the Comet pipeline entirely rather than staying native. `CometArrayUnion`, `CometCreateArray`, `CometSlice`, and the `NullType` result gate in `CometIf` / `CometCaseWhen` / `CometCoalesce` are in the same position. Since `CometBatchKernelCodegen.canHandle` accepts all of those output types after this change, would mixing in `CodegenDispatchFallback` keep them in the pipeline instead?

The new restrictions do not reach the generated compatibility pages. `CometArrayUnion`, `CometArrayIntersect`, `CometArrayExcept`, `CometSlice`, `CometArrayRepeat` and `CometCreateArray` each gained an `Unsupported` branch but still return an empty `getUnsupportedReasons()`, and `CometArraysZip` has one that does not mention the new null-guard restriction. `hash.scala` does this the right way by adding `unsupportedNullTypeReason` to `HashUtils.unsupportedReasons`, so it would be good to follow that wherever a support level moved in this PR. `CometArrayExcept` is the notable one, since its element-type restriction used to at least carry the text inline in the `Incompatible` note.

`CometNullTypeCompositionSuite` is now registered in both `pr_build_linux.yml` and `pr_build_macos.yml`, and it is large. By my count it expands to 37 tests, and the operator sweep alone runs 8 producers across 32 operator templates against both engines in each of six physical profiles, three of which set `spark.sql.shuffle.partitions` to 300 over 900 rows. The nesting sweep asserts a floor of 200 compared cases per profile on top of that. How long does the suite take end to end on one runner? If it is long, pruning the profile matrix on the consumer sweeps would still leave the shuffle and batching coverage intact.

grorge123:
Thanks, all three addressed in one commit on top, rebased onto main (`3810936b4`).

**Dispatcher enrollment.** `CometArrayRepeat`, `CometArrayUnion`, `CometCreateArray`, `CometIf`, `CometCaseWhen` and `CometCoalesce` now mix in `CodegenDispatchFallback`, so their `Unsupported` cases run through the JVM codegen dispatcher and the plan stays fully native. `CometSlice` needs no gate after the rebase: #5766's `convert` no longer serializes a return type, so the mismatch it guarded against cannot happen and `slice(map_entries(map(k, NULL)), 1, 1)` runs natively. On `array_repeat`, your example made me look closer: `array_repeat(array(c), 2)` runs natively on main because native `make_array` emits a nullable item whatever Spark's `containsNull` says; what fails is an input whose native item is really non-nullable (`map_entries`, "ListArray expected data type List(non-null Struct(..)) got List(Struct(..))"). The gate now exempts a `CreateArray` child, so `array(c)` stays on the kernel and `map_entries(..)` goes through the dispatcher. Enrolling `coalesce` exposed a dispatcher bug: a top-level `struct<.., x: null>` output failed to allocate (`StructVector`'s constructor builds a `NullableStructWriter` over the field's children, which has no arm for a Null child); the output vector now adds its children after construction, with a unit test. One boundary: the dispatcher accepts NullType as an output but not as an input column, so a gated shape whose NullType side arrives as an upstream column still falls back. The SQL file tests that asserted a fallback now assert the native plan, and `CometNullTypeCompositionSuite` checks per gate that the serde is enrolled, the reason is published, and the query keeps its `CometProjectExec`.

**Compatibility pages.** Every serde whose support level moved publishes the reason through `getUnsupportedReasons`, following `HashUtils`; the runtime note starts with the published text and the suite asserts that, so the two cannot drift apart.

**Suite size.** It was 37 tests in 10 min 49 s on an i7-12700K, so likely two to three times that on a runner, and the consumer sweeps were the bulk. Each sweep now runs only under the profiles that can change its outcome (a null guard is per row, so batching cannot affect the nullable sweep; a stateful producer only diverges across a batch boundary, so the kernel choice cannot affect that one; a non-deterministic child leaves every kernel, so neither reaches the last), the columnar-to-row-only physical profile is folded into the sort-writer one, and Spark's reference rows are cached across profiles with the same row count. That is 28 tests in about 8 min on the same machine, with the operator and nesting sweeps still under every shuffle path and both batch sizes; nothing removed had ever caught a failure the kept combinations did not.

Rebase notes: `CometElementAt` takes main's #5544 fix, the `NullArray` special case in `GetStructField` moved into #5769's shared `child_with_parent_nulls`, and the two duplicate-field-name checks are folded into one with main's reason wording.

Assisted-by: Claude Code (claude-fable-5)

PR #5531, opened 2026-08-28
feat: add raw Celeborn native shuffle reader (7/n)

pingzh:

Part of #5352. This is the seventh foundational PR and does not close the issue.

Previous PRs:

- https://github.com/apache/datafusion-comet/pull/5513

## Rationale for this change

The preceding PRs added native Celeborn map-side shuffle writes. Comet also needs a reduce-side reader that consumes native frames without Celeborn's row decompression, preserves AQE read ranges, and participates in Spark's fetch-failure recovery.

- Add a shared local/remote reader interface and route native Celeborn handles through the application-owned client.
- Read reducers lazily using one file-group snapshot, preserving shuffle generations, mapper ranges, attempt identities, and shuffle metrics. Support the stock Celeborn 0.6/0.7 raw-reader API and the newer coalesced-metadata overload without adding a Celeborn dependency.
- Validate frame boundaries before concatenating reducers and preserve `FetchFailedException` across native codec/IPC decoding. Keep metadata timeouts, cancellation, and cleanup failures distinct from corrupt shuffle data.
- Validate Arrow payloads on the remote path while retaining the existing local decoder entry point and fast path.
- Release prefetched batches and unconsumed Arrow structures on early termination, EOF, and failure, preserving suppressed cleanup errors.

Planner enablement remains for 8/n; this PR does not enable end-to-end Celeborn query planning.

- 177 JVM tests passed across the Celeborn reader/pusher/manager/writer, native shuffle, configuration, and Arrow utility suites (Spark 4.1.3 / Scala 2.13).
- 299 Rust unit tests passed across the core, shuffle, and JNI bridge crates. Four existing HDFS-dependent tests remain ignored.
- Added regressions using real native frames and Arrow IPC for reducer boundaries, corrupt offsets, schema mismatches, JNI fetch-failure propagation, cancellation, and cleanup.
- Seven isolated compatibility probes passed using published Celeborn 0.6.3/0.7.0 binaries. Reader/decoder source was also type-checked against Spark 3.4.3/3.5.9 and Scala 2.12 APIs; these probes are not complete legacy Maven builds or live Celeborn cluster tests.
- Spotless, Scalastyle, Rust formatting, workspace/all-target Clippy with warnings denied, and Scala 2.12 parsing passed.
- Three independent review passes completed with no outstanding P1/P2 findings.

### DIFF
--- a/native/core/src/execution/operators/shuffle_scan.rs
+++ b/native/core/src/execution/operators/shuffle_scan.rs
@@ -18,11 +18,13 @@
 use crate::{
     errors::CometError,
     execution::{
-        operators::ExecutionError, planner::TEST_EXEC_CONTEXT_ID, shuffle::ipc::read_ipc_compressed,
+        operators::ExecutionError,
+        planner::TEST_EXEC_CONTEXT_ID,
+        shuffle::{read_ipc_compressed, read_ipc_compressed_validated, validate_remote_schema},
     },
     jvm_bridge::{jni_call, JVMClasses},
 };
-use arrow::array::ArrayRef;
+use arrow::array::{ArrayRef, RecordBatch};
 use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
 use datafusion::common::Result as DataFusionResult;
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
@@ -68,6 +70,8 @@ pub struct ShuffleScanExec {
     baseline_metrics: BaselineMetrics,
     /// Time spent decoding compressed shuffle blocks.
     decode_time: Time,
+    /// Remote inputs require Arrow array and logical schema validation; queried once at construction.
+    requires_validation: bool,
 }
 
 impl ShuffleScanExec {
@@ -76,6 +80,16 @@ impl ShuffleScanExec {
         input_source: Option<Arc<Global<JObject<'static>>>>,
         data_types: Vec<DataType>,
     ) -> Result<Self, CometError> {
+        let requires_validation = if exec_context_id == TEST_EXEC_CONTEXT_ID {
+            false
+        } else if let Some(input) = &input_source {
+            JVMClasses::with_env(|env| unsafe {
+                jni_call!(env,
+                    comet_shuffle_block_iterator(input.as_obj()).requires_validation() -> bool)
+            })?
+        } else {
+            false
+        };
         let metrics_set = ExecutionPlanMetricsSet::default();
         let baseline_metrics = BaselineMetrics::new(&metrics_set, 0);
         let decode_time = MetricBuilder::new(&metrics_set).subset_time("decode_time", 0);
@@ -99,6 +113,7 @@ impl ShuffleScanExec {
             baseline_metrics,
             schema,
             decode_time,
+            requires_validation,
         })
     }
 
@@ -123,6 +138,7 @@ impl ShuffleScanExec {
                 self.input_source.as_ref().unwrap().as_obj(),
                 &self.data_types,
                 &self.decode_time,
+                self.requires_validation,
             )?;
             *current_batch = Some(next_batch);
         }
@@ -138,6 +154,7 @@ impl ShuffleScanExec {
         iter: &JObject,
         data_types: &[DataType],
         decode_time: &Time,
+        requires_validation: bool,
     ) -> Result<InputBatch, CometError> {
         if exec_context_id == TEST_EXEC_CONTEXT_ID {
             return Ok(InputBatch::EOF);
@@ -173,7 +190,21 @@ impl ShuffleScanExec {
 
             // Decode the compressed IPC data
             let mut timer = decode_time.timer();
-            let batch = read_ipc_compressed(slice)?;
+            let batch = match decode_shuffle_batch(slice, data_types, requires_validation) {
+                Ok(batch) => batch,
+                Err(failure) => {
+                    // Remote inputs must invalidate the failed shuffle generation even when
+                    // fetching bytes succeeded and only IPC/codec decoding or logical schema
+                    // validation detects the corruption. JNI preserves FetchFailedException.
+                    // Local inputs leave this callback as a no-op and keep the native error.
+                    let message = env.new_string(failure.to_string())?;
+                    unsafe {
+                        jni_call!(env,
+                            comet_shuffle_block_iterator(iter).on_decode_failure(&message) -> ())?;
+                    }
+                    return Err(failure.into());
+                }
+            };
             timer.stop();
 
             let num_rows = batch.num_rows();
@@ -188,17 +219,35 @@ impl ShuffleScanExec {
                 .map(|col| unpack_dictionary(col))
                 .collect();
 
-            debug_assert_eq!(
-                columns.len(),
-                data_types.len(),
-                "Shuffle block column count mismatch: got {} but expected {}",
-                columns.len(),
-                data_types.len()
-            );
-
             Ok(InputBatch::new(columns, Some(num_rows)))
         })
     }
+}
+
+fn decode_shuffle_batch(
+    bytes: &[u8],
+    expected_types: &[DataType],
+    requires_validation: bool,
+) -> DataFusionResult<RecordBatch> {
+    if requires_validation {
+        let batch = read_ipc_compressed_validated(bytes)?;
+        // Validate before unpack_dictionary or cast_and_stamp_schema can hide an incompatible
+        // wire type by changing values. Keep failures inside get_next's recovery callback.
+        validate_remote_schema(&batch, expected_types)?;
+        Ok(batch)
+    } else {
+        check_column_count(read_ipc_compressed(bytes)?, expected_types.len())
+    }
+}
+
+fn check_column_count(batch: RecordBatch, expected: usize) -> DataFusionResult<RecordBatch> {
+    if batch.num_columns() != expected {
+        return Err(datafusion::common::DataFusionError::Execution(format!(
+            "Shuffle block column count mismatch: got {} but expected {expected}",
+            batch.num_columns()
+        )));
+    }
+    Ok(batch)
 }
 
 /// If `array` is dictionary-encoded, cast it to the value type. Otherwise return as-is.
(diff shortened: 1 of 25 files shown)


native/core/src/execution/operators/shuffle_scan.rs
```diff
@@ -173,7 +190,27 @@ impl ShuffleScanExec {
 
             // Decode the compressed IPC data
             let mut timer = decode_time.timer();
-            let batch = read_ipc_compressed(slice)?;
+            let decoded = if requires_validation {
+                read_ipc_compressed_validated(slice)
+            } else {
+                read_ipc_compressed(slice)
+            };
+            let batch = match decoded.and_then(|batch| check_column_count(batch, data_types.len()))
```
sunchao: [P2] Validate remote logical types before schema casting

The remote validation path checks Arrow buffer validity and column count, but does not compare the decoded logical types with `data_types`. A corrupt IPC schema can therefore reach `cast_and_stamp_schema`, which can silently change values instead of reporting a fetch failure.

Reproduced at `7467fb296f1ac229ba499cea663d1e1cec1ca1c5` with Spark 3.5.9/JDK 17 and the published Celeborn 0.6.3 and 0.7.0 clients, using a native Celeborn handle with stage reruns enabled, `spark.shuffle.compress=false`, no Spark I/O encryption, and Celeborn integrity checking disabled (`celeborn.client.shuffle.integrityCheck.enabled=false`, also its default):

1. Generate a native `NONE` frame using `ShuffleBlockWriter` for `Field("value", Int32, false)` and rows `[-1, 0, 1]`.
2. Flip only the IPC schema's `Int.is_signed` bit from true to false (`frame[135]: 1 -> 0` in this 540-byte frame). The native length, field count, and array buffers remain unchanged.
3. Read that committed mapper batch through the new manager/reader route, stock `ShuffleClientImpl.readPartition(..., needDecompress=false)`, and native `ShuffleScan` with the original Spark `IntegerType` declaration.

Arrow accepts the resulting `UInt32` array, this count check succeeds, and the later cast turns `4294967295` into null. The corrupted frame produces `[null, 0, 1]` with no exception and zero `reportShuffleFetchFailure` calls. The JVM `read()` route instead throws `UnsupportedOperationException: Unsupported data type: Int(32, false)` during import, also without a `FetchFailedException` or failure report. Both routes should reject this incompatible logical schema and invoke the remote failure handler before casting/importing it.

The pristine-frame and corrupt-stale-attempt controls both preserve `[-1, 0, 1]`; the existing bad-codec control correctly reports a fetch failure. The exact-base (`a223ba14e9f1445b5d315962ae9005f7ec45b187`) manager rejects native Celeborn reads before fetching. The local cast itself is unchanged: the introduced exposure is the new remote reader reaching it with `requires_validation=true` but no logical-type check. Compatible dictionary representations and nested-nullability normalization still need to remain supported.

These controls used actual Comet native/JNI execution and stock client reads over loopback transport, with controlled worker/lifecycle responses; they are not deployed-cluster or planner-enablement tests.

sunchao (approved):
Re-reviewed `16d1bb86b61a44d6fc29239445fa1b28ef27cc27` against base `a223ba14e9f1445b5d315962ae9005f7ec45b187`.

Independently verified the [schema-validation P2](https://github.com/apache/datafusion-comet/pull/5531#discussion_r3883250776) is fixed. Actual native/JNI and published Celeborn 0.6.3/0.7.0 paths now reject the signedness-corrupted frame before casting/importing and report the correct shuffle generation once. Pristine/stale-attempt controls and compatible dictionaries and nested schemas preserve their values across all four codecs.

Fresh validation: 794 focused JVM tests across Spark 3.4/3.5/4.0/4.1/4.2, 309 native tests passed (four existing HDFS tests ignored), and 364 current stock-client controls with prior-head comparisons. No remaining introduced or materially worsened P1/P2 found.

Validation used local builds and controlled loopback worker/lifecycle responses, not a deployed Celeborn cluster or planner integration. CI remains in progress.

sunchao:
Merge, thanks @pingzh !!

andygrove:
I ran a post-merge pass over this with a local build, so everything below is retrospective rather than a request to change anything here. Three follow-ups filed: #5534, #5535 and #5536.

Starting with what held up, because I went looking for problems and mostly did not find them. I checked every Celeborn member this reflects against the real `apache/celeborn` sources, and they all line up. The 15-argument `readPartition` matches positionally in 0.6.0, 0.6.3 and 0.7.0, including `shuffleId` before `appShuffleId`, and `ReduceFileGroups.{partitionGroups, mapAttempts, pushFailedBatches}` are public fields in all three. On 0.6.3 and 0.7.0 `CelebornShuffleHandle.stageRerunEnabled` exists on a class that really does extend `BaseShuffleHandle`, and `object CelebornShuffleReader` really does register the broadcast `GetReducerFileGroupResponse` deserializer in its static initialiser, so the comment justifying that `loadClass` is accurate. Passing `null` for `exceptionMaker` is right as well, since Celeborn's own `reportShuffleFetchFailure` sits inside an `exceptionMaker != null` guard, which makes `reportFetchFailure` here the single reporter rather than a double one. Locally I ran 92 shuffle and 192 core Rust tests, 58 in `CometCelebornShuffleReaderSuite` and 113 across the local shuffle suites on Spark 4.1 / Scala 2.13 / JDK 17, all green.

On #5534, `getNextBatch` now releases the Arrow structs when `func` throws and when native returns EOF, but the `importVector` call is still outside the guard. If `importer.importVector` throws part-way through the column loop, the structs from the failing column onward keep their populated C data with the release callback never invoked, and the wrapper buffers leak from the allocator too. I reproduced it with the suite's own `withIsolatedStructAllocator`: export column 0, leave column 1 as a freshly allocated struct, and after the resulting `IllegalStateException: Cannot import released ArrowSchema` the allocator still reports 176 bytes outstanding. The success path is clean by comparison, 48 bytes before `batch.close()` and 0 after, so this is specific to the import failure. The comment at `NativeBatchDecoderIterator.scala:105` already treats Arrow import failures as a first-class case, so I think this one just got missed. It is worth fixing beyond shuffle too, since `getNextBatch` backs every native operator and an allocator OOM inside `importVector` is the most likely way to hit it.

On #5535, could a benchmark for the remote read path go into the 8/n description? Dropping `with_skip_validation(true)` for remote frames is the right call, but it is not cheap and the cost is not written down anywhere. Release build, one `ShuffleBlockWriter` frame of 20 columns by 8192 rows, 300 iterations:

| codec    | local (us/blk) | validated (us/blk) | ratio |
| -------- | -------------- | ------------------ | ----- |
| None     | 38.8           | 175.2              | 4.52x |
| Lz4Frame | 225.8          | 356.6              | 1.58x |
| Snappy   | 558.5          | 696.5              | 1.25x |
| Zstd(1)  | 636.9          | 773.1              | 1.21x |

The local path itself did not regress, and measured against the pre-PR implementation it is actually a little faster, from -8.8% on `None` to +1.1% on Snappy, so the table above is purely the price of the new reader. Since the whole reason to read Celeborn frames natively is to beat Celeborn's own row decoder, that seems worth stating so 8/n can be judged against it. Two things on the same path also look hoistable if it matters. `decodeShuffleBlockWithValidation` re-decodes the `ShuffleScan` protobuf and rebuilds `expected_types` on every block, and `env.convert_byte_array` copies the serialized schema into a fresh `Vec` each time, and both are constant per reducer.

On #5536, `validate_remote_schema` unwraps exactly one dictionary shape, top-level `Dictionary(Int32, Utf8 | Binary)`, and treats everything else as a mismatch. The local reader is much more permissive, since `unpack_dictionary` casts any top-level dictionary to its value type and `cast_and_stamp_schema` absorbs nested ones. So the two readers disagree about what a valid frame is, and only the remote one turns the disagreement into a `FetchFailedException` and a map-stage rerun that reproduces the same frame. I checked whether this is live and it is not. I instrumented the decoder to log any wire schema containing a dictionary at any depth and ran 125 tests across `CometFuzzTestSuite`, `CometShuffleSuite` and `CometNativeShuffleSuite` with zero hits, because `SchemaAlignExec` and `ScanExec::unpack_dictionary_type` between them keep dictionaries off the wire entirely today. What worries me is that nothing holds that in place. `ShuffleBlockWriter` genuinely encodes nested dictionaries, and on a frame the writer produced I get `Shuffle block type mismatch at column 0: got List(Dictionary(Int32, Utf8)) but expected List(Utf8)` from the validator while `arrow::compute::cast` handles the same frame fine. Would it be worth either recursing the dictionary unwrap through the nested arms so the validator accepts whatever the encoder can emit, or adding a test that pins "the writer never emits a dictionary the reader would reject" so a drift breaks the build instead of production?

Two smaller things I did not file. There are now three places that parse the same 16-byte frame header with different rules: `NativeFrameInputStream` checks the length floor, the ceiling, the field-count range and the expected field count, `CometShuffleBlockIterator` checks the floor and ceiling and discards the field count, and `NativeBatchDecoderIterator.readNextBlock` checks only the ceiling. The third has no `compressedLength < 12` check, so a length below 8 reaches `dataBuf.limit(negative)` and throws `IllegalArgumentException` rather than a classified corruption error. On the Celeborn path `NativeFrameInputStream` catches it first so nothing is broken, but the local path has no equivalent guard and three validators that are meant to agree tend to stop agreeing. Separately, `NativeUtil.releaseArrowStructs` and `NativeBatchDecoderIterator.close` now contain the same nine-line "collect the first failure and `addSuppressed` the rest" helper, character for character apart from what they release. Both are new here and the series looks likely to want a third, so a small helper in `org.apache.comet.util` next to `ClassLoaders` might be worth it.

Last one is a process note for 8/n rather than anything about this code. Three pieces of this PR change behaviour for every Comet user who will never run Celeborn: the new strictness in `read_ipc_compressed` around empty streams, multiple batches, trailing bytes and truncated LZ4, the `NativeUtil` Arrow-struct release paths, and the `NativeBatchDecoderIterator` switch from an eager constructor fetch to a lazy one plus no longer swallowing `EOFException` as clean EOF. I ran the local shuffle and fuzz suites against all of it and found no regression, so I do not think anything is broken. But that is 335 lines that mention Celeborn nowhere, the `NativeUtil` part is a standalone leak fix with a standalone test, and the `ipc.rs` part tightens a path every query touches. Would you consider peeling shared-path changes like those out of the Celeborn series going forward? They would each get their own bisect point and could be reviewed on their own terms.

andygrove:
@sunchao @pingzh This 4k loc PR was merged just 3-4 hours after being opened. That isn't enough time for me to review, even with AI assistance. I posted my review anyway, and filed some issues.

What do you think about haivng some minimum period of time for allowing for others to review before merging?

sunchao:
@andygrove sorry. I think it is good idea to have a minimum period of time for these large changes. Let's discuss on the Slack for this. Thanks.

PR #5533, opened 2026-08-28
fix: preserve skipped unbase64 rows

### LINKED ISSUE #5532: unbase64 can fail on rows skipped by LIMIT and semi/anti joins
### Describe the bug

Comet can fail while decoding malformed Base64 on rows that Spark never evaluates. A limit may stop after the first row, or a semi/anti join may stop checking candidates after finding a match. Both native `unbase64` and the JVM codegen dispatcher can evaluate a whole batch before that early termination takes effect.

This means enabling Comet can turn a successful Spark query into `Last unit does not have enough valid bits`, even when the malformed value does not contribute to the result.

### Steps to reproduce

Start Spark with the Comet extension, native execution, and the Comet shuffle manager configured. Create the input with Comet disabled, then compare the same query with Comet disabled and enabled:

```sql
SET spark.comet.enabled = false;
SET spark.sql.shuffle.partitions = 1;

CREATE TABLE unbase64_limit_input USING parquet AS
SELECT /*+ COALESCE(1) */ bad
FROM VALUES ('YWJj'), ('A') AS v(bad);

SELECT hex(unbase64(bad)) FROM unbase64_limit_input LIMIT 1;

SET spark.comet.enabled = true;

SELECT hex(unbase64(bad)) FROM unbase64_limit_input LIMIT 1;
```

In the reproduced single-file scan order, `YWJj` is read before `A`. Spark returns `616263` (the bytes for `abc`). Comet instead throws on the second value. This example relies on that physical input order; it does not assume that SQL without `ORDER BY` guarantees ordering.

The failing Comet plan is:

```text
CometCollectLimit 1
+- CometProject [hex(unbase64(bad))]
   +- CometNativeScan parquet
```

The same problem occurs when the decoder is in a filter below `LIMIT`, in semi/anti join residual conditions, or below an already ordered `WindowGroupLimit`. Compound inputs and `to_binary(bad, 'base64')` can reach the JVM dispatcher and still fail for the same reason.

### Expected behavior

Match Spark's evaluation behavior: preserve successful queries when Spark skips the malformed value, while still raising an error when Spark actually decodes it. Ordinary native decoding should remain available when early termination cannot skip the value.

### Additional context

Reproduced on Apache Comet `78defce020ac3c4d2b32ff7f172f3018b7587a48` with Spark 4.1.3, a fresh native/JVM build, and verified Comet physical operators. The relevant implementation is unchanged at `a223ba14e9f1445b5d315962ae9005f7ec45b187`.

Seven query shapes produced 28 Spark-versus-Comet mismatches across all four combinations of ANSI mode and AQE. Sixteen comparisons using valid inputs or deliberately consumed malformed values matched Spark. This is not specific to ANSI mode, and switching to JVM codegen dispatch alone does not preserve row-by-row evaluation.

sunchao:

Closes #5532.

## Rationale for this change

Spark can finish a query without decoding every input row. For example, if a scan produces valid `YWJj` followed by malformed `A` in one partition:

```sql
SELECT hex(unbase64(bad)) FROM t LIMIT 1;
```

Spark returns `616263` without decoding the second row. Comet can instead decode the whole batch and fail on the skipped value. This example depends on physical input order; SQL without `ORDER BY` does not guarantee that order. Semi and anti joins can similarly stop after the first matching candidate. Routing the decoder through JVM dispatch alone does not help, because that path also evaluates batches.

The planner consults an explicit expression policy, `RequiresSparkEvaluationMask`, and keeps affected evaluation paths in Spark. Only `UnBase64` opts in, including subclasses that can be evaluated inside a dispatched expression tree. The native decoder and its validation rules are unchanged.

Protection covers limits, semi/anti join conditions, and top-K or unpartitioned window limits whose inputs already meet their ordering requirements. It survives row/column transitions and reused AQE plans. Partitioned window limits drain their groups and do not independently trigger fallback, but an outer limit still can. Decoding below a blocking sort remains eligible for native execution.

AQE needs one conservative exception: when final aggregate results decode and Spark cannot read the native partial buffers, both aggregates stay in Spark beneath an outer limit. That choice happens before materialization, because AQE can later remove an intervening sort. Compatible partial aggregations can remain native; decoding already consumed by an aggregate's inputs does not count as decoding in its final result. Offset-only collection does not create a new evaluation mask.

The default fallback also applies when data in an affected path happens to be valid; its performance cost has not been benchmarked. This is a reusable opt-in, not a general classification of throwing expressions. Other expressions, aggregate FILTER evaluation, and pre-existing empty-native-join evaluation differences remain outside this fix.

The fallback is enabled by default and can be disabled with
`spark.comet.exec.preserveEvaluationMasks.enabled=false` for known-valid input. This opt-out
can restore native execution under limits and semi/anti joins, including the conservative AQE
aggregate protection, but malformed values in Spark-skipped rows can then raise errors. Existing
expression support restrictions still apply. The wider throwing-expression and conditional-mask
audit is tracked in #6006.

The rule exits early for plans with no enrolled expression or existing protection tag. Aggregate
buffer restoration is shared with the existing Celeborn fallback, with explicit traversal
boundaries: LIMIT protection can cross unmaterialized shuffle exchanges; Celeborn restoration
stays within its current stage. Existing native ancestors are rebuilt with their logical links.

The focused Scala suite retains the execution and AQE regressions, and a minimal
`unbase64_operator_masks.sql` fixture records the skipped-malformed-row bug beside the other
expression fixtures.

Validated on Spark 4.1.3, Scala 2.13.17, and JDK 17. All **110 executed targeted tests passed**: `CometEvaluationMaskSuite` (9), `CometExecRuleSuite` (32), `CometScalarFunctionSuite` (14), `CometCelebornShufflePlanningSuite` (50), and the applicable `unbase64` SQL registrations (5, including both restored-fixture dispatcher modes). One additional Spark-3.5-only SQL registration skipped its body through the version gate and is excluded from that count.

Full JVM test compilation, Spotless, Scalastyle, and `git diff --check` passed. The JVM tests used the native library from the successful [CI run for the exact rebased main commit](https://github.com/apache/datafusion-comet/actions/runs/35275954309) (`d0c10f2d397ae81b97f187c0dcb4d8f042f2cd9f`), after verifying the artifact digest and that native source/build inputs were unchanged. A local Cargo build was blocked because the configured registry lacks DataFusion 55.1.0; no registry or dependency was changed.

Other Spark/Scala versions, macOS, and the full repository suite were not run locally. Hosted PR CI is separate from these results.


spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -714,6 +722,132 @@ case class CometExecRule(session: SparkSession)
     }
   }
 
+  /** Keep operators that can skip malformed unbase64 inputs in Spark's row pipeline. */
+  private def preserveUnbase64EvaluationMasks(plan: SparkPlan): SparkPlan = {
+    val limitReason = "unbase64 requires Spark evaluation below LIMIT"
+    val joinReason = "unbase64 requires Spark evaluation in first-match join conditions"
+
+    def containsUnbase64(expr: Expression): Boolean =
+      expr.exists(_.isInstanceOf[UnBase64])
```
andygrove: This is the only `unbase64`-specific thing in the whole rule. Everything else in `preserveUnbase64EvaluationMasks` is a general mechanism, but the underlying divergence is not specific to `unbase64` either. Any expression that throws on data has it, under these same operators. `SELECT a + b FROM ansi_int_overflow LIMIT 1` with ANSI on fails in Comet and succeeds in Spark for exactly the same reason, and so do ANSI cast, decimal divide-by-zero, and `element_at`. So we would be shipping a fair amount of planner machinery that fixes one member of the family.

Could the trigger come from the serde instead of a hard-coded class check? Something like a marker trait on `CometExpressionSerde` meaning "this expression can raise a data-dependent error", with `CometUnBase64` mixing it in, so `containsUnbase64` becomes a registry lookup. Then ANSI arithmetic can opt in later without anyone touching `CometExecRule` again.

I want to flag the flip side honestly, because I think it decides the shape of this PR. Once the trigger widens to ANSI `Add` or `Cast`, the fallback swallows a large fraction of real queries. That makes me wonder whether the right call is the one we have made for the other divergences in this family, which is to document it under "Known result-value divergences" in `docs/source/user-guide/latest/compatibility/index.md` and file the general problem, rather than special-case it in the planner. I would rather we pick one of those two directions deliberately than land the narrow version by default.

There is also a shape this approach cannot reach even for `unbase64` alone. `WHERE k = 1 AND unbase64(bad) = X'616263'` short-circuits per row in Spark and evaluates the whole batch natively in Comet, with no `LIMIT` involved. Same for `CASE WHEN k = 1 THEN unbase64(bad) END`. Is that worth a tracking issue alongside this, so we are not implicitly claiming the class is closed?

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -714,6 +722,132 @@ case class CometExecRule(session: SparkSession)
     }
   }
 
+  /** Keep operators that can skip malformed unbase64 inputs in Spark's row pipeline. */
+  private def preserveUnbase64EvaluationMasks(plan: SparkPlan): SparkPlan = {
```
andygrove: Related to my other comment, this changes performance for queries that were never broken. The reason propagates to every ancestor up to the nearest exchange, so any projection with `unbase64` under a `LIMIT` leaves native execution for the whole chain, and `CollectLimitExec` shows up on every `df.show()` and every BI-tool `LIMIT`. That gives back the 2.4x to 3.5x the native kernel bought in #5451, for data that is almost always well-formed, and there is no way to turn it off. Could we gate this behind a `CometConf` key so a user who knows their base64 is clean can keep the native path?

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -714,6 +722,132 @@ case class CometExecRule(session: SparkSession)
...
+      val startsLimit = original match {
+        case _: CollectLimitExec | _: LocalLimitExec | _: GlobalLimitExec => true
+        case topK: TakeOrderedAndProjectExec =>
+          SortOrder.orderingSatisfies(node.children.head.outputOrdering, topK.sortOrder)
+        case windowLimit
+            if ShimCometWindowGroupLimit.windowGroupLimitClass.exists(
+              _.isInstance(windowLimit)) =>
+          SortOrder.orderingSatisfies(
+            node.children.head.outputOrdering,
+            windowLimit.requiredChildOrdering.head)
```
andygrove, line 797: This branch looks like it only holds without a `PARTITION BY`. `WindowGroupLimitEvaluatorFactory` picks the plain limit iterator when `partitionSpec.isEmpty`, and that one stops pulling once `rank >= limit`. With a partition spec it uses `GroupedLimitIterator`, whose `skipRemainingRows()` calls `fetchNextRow()` in a loop, so Spark pulls and evaluates every input row anyway. In that case we fall back and gain nothing.

Since `EnsureRequirements` guarantees the child satisfies `requiredChildOrdering`, this check is effectively always true, so the partitioned case is reachable whenever the child happens to be sorted already and no `SortExec` gets inserted to stop the propagation. Should this be restricted to `partitionSpec.isEmpty`? The new fixture only covers the unpartitioned shape, so a partitioned case would be worth adding either way.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
```diff
@@ -714,6 +722,132 @@ case class CometExecRule(session: SparkSession)
...
+          val restored = original.withNewChildren(children)
+          node.getTagValue(SparkPlan.LOGICAL_PLAN_TAG).foreach(restored.setLogicalLink)
+          restored
+        case _ => node.withNewChildren(children)
+      }
+      reason.foreach(prepared.setTagValue(CometExecRule.UNSAFE_UNBASE64_EVALUATION, _))
+      (prepared, reason)
+    }
+
+    protect(plan, belowLimit = false)._1
```
andygrove: This walks every operator's expression tree on every plan on every rule application, and for the overwhelming majority of queries there is no `UnBase64` to find. Given #5199 is open on planner rule cost, could we bail out up front with something like `if (!plan.exists(_.expressions.exists(containsUnbase64))) return plan`? The sticky-tag behaviour stays correct, because a node that carries the tag still contains the decoder that earned it.

spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -733,8 +735,12 @@ object CometUnBase64 extends CometExpressionSerde[UnBase64] with CodegenDispatch
     "unbase64 with a non-trivial child expression uses the JVM codegen dispatcher to preserve" +
       " Spark's short-circuit evaluation (native path is limited to column and literal children)"
 
+  private val maskedEvaluationReason =
+    "unbase64 below LIMIT or in first-match join conditions falls back to Spark so malformed" +
+      " input is not decoded on rows Spark skips"
+
   override def getUnsupportedReasons(): Seq[String] =
-    Seq(failOnErrorReason, nonTrivialChildReason)
+    Seq(failOnErrorReason, nonTrivialChildReason, maskedEvaluationReason)
```
andygrove, line 748: The wording here reads well, and I know the contract in `adding_a_new_expression.md` allows these to differ from what `getSupportLevel` returns, so I am not asking you to move it. Worth noting for anyone reading later that this reason is never produced by the serde: the decision is made by `CometExecRule` at the operator level.

The part I would like changed is the committed docs, which `GenerateDocs` will not touch. `docs/source/user-guide/latest/expressions.md:618` still has an empty Notes cell for `unbase64`, and the audit entry at `docs/source/contributor-guide/expression-audits/string_funcs.md:248` still lists `failOnError` and non-trivial children as the only restrictions. Could you mirror this reason into both?

spark/src/test/resources/sql-tests/expressions/misc/unbase64_operator_masks.sql
```diff
@@ -0,0 +1,118 @@
...
+ON l.k = r.k AND unbase64(r.bad) > l.expected
+
+-- Semi/anti joins without a throwing decoder remain native.
+query
+SELECT /*+ BROADCAST(r) */ l.* FROM test_unbase64_operator_left l
+LEFT SEMI JOIN test_unbase64_operator_right r ON l.k = r.k
+
+query
+SELECT /*+ BROADCAST(r) */ l.* FROM test_unbase64_operator_left l
+LEFT ANTI JOIN test_unbase64_operator_right r ON l.k = r.k
```
andygrove: The fixtures cover the limit and broadcast-join shapes nicely, and I like that the native-execution controls are here rather than only in the rule suite.

Two shapes only exist as hand-built plans in `CometExecRuleSuite` and never run end to end. One is `TakeOrderedAndProject` with pre-satisfied child ordering, which is the branch whose guard depends on what `outputOrdering` actually reports after conversion. The other is the shuffled-hash and sort-merge semi/anti strategies, since this file only exercises broadcast. Would you mind adding an `ORDER BY ... LIMIT n` case and a case with `spark.sql.autoBroadcastJoinThreshold=-1`? Those are the two places where a synthetic plan and a real one are most likely to disagree.

andygrove (commented):
Thanks for digging into this. The operator analysis is careful, and the negative controls in the fixtures (reversed build order, `WHERE bad = 'A'`, `rn <= 2`) are what convinced me the model is right rather than just conservative.

I checked each operator the rule treats as an early-stop point against Spark master. `BaseLimitExec` codegen guards `consume` behind the count check and the producer loop tests `stopEarly()`, so rows past the limit really are never projected. `TakeOrderedAndProjectExec` uses `take(limit)` only when `SortOrder.orderingSatisfies(child.outputOrdering, sortOrder)`, and the guard here mirrors Spark's own condition exactly. `HashJoin.semiJoin` / `antiJoin` and `BroadcastNestedLoopJoinExec.leftExistenceJoin` all short-circuit through `.exists`. `ExistenceJoin` short-circuits too, but Comet rejects it in all three join serdes, so its condition is already evaluated per row in Spark and there is no gap there. The `materializesInput` list also looks sound: anything missing from it produces more fallback rather than less, and everything in it does drain its input.

The one I think is wrong is `WindowGroupLimit`, and I have left a comment inline.

My main concern is scope, and it is the thing I would like settled before this lands. Nothing about the bug is specific to `unbase64`. Comet evaluates eagerly over a batch while Spark evaluates lazily per row, so every expression that can throw on data has this exact problem under these exact operators. `arithmetic_ansi.sql` shows Comet raising `ARITHMETIC_OVERFLOW` natively, so `SELECT a + b FROM ansi_int_overflow LIMIT 1` with ANSI on fails in Comet and succeeds in Spark for the same reason, and so do ANSI cast, decimal divide-by-zero, and `element_at`. The machinery here is general; only the trigger is hard-coded. I have written that up inline along with the flip side, because I think it decides the shape of the PR.

The rest of my comments are independent of how that question lands.

One documentation note that I could not attach inline because the file is not in the diff. `docs/source/user-guide/latest/expressions.md:618` still has an empty Notes cell for `unbase64`, and the audit entry at `docs/source/contributor-guide/expression-audits/string_funcs.md:248` still lists `failOnError` and non-trivial children as the only restrictions. Both are committed files, so `GenerateDocs` will not pick this up. Could you add the new operator-level restriction to both? The `getUnsupportedReasons()` string reads well, so it is mostly a matter of mirroring it.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
sunchao: Fixed in 28074b67. A known partitioned `WindowGroupLimit` no longer starts the fallback itself. It drains its groups when fully consumed, but an outer `LIMIT` can still stop that input, so the outer protection continues through it. Unknown shim extractions retain the conservative behavior.

I expanded the planner matrix across partitioned/unpartitioned input, all three ranks, partial/final modes, outer limits, dispatcher settings, native child reuse, and repeated planning. The new execution test keeps valid input native, requires the consumed malformed suffix to throw, and checks that `LIMIT 1` can skip that suffix. These pass on Spark 3.5.9 and 4.1.3.

The updated description also records a separate existing empty-build join difference that the earlier broad fallback happened to hide. The partitioned-window change does not claim complete error-evaluation parity.

spark/src/main/scala/org/apache/comet/serde/strings.scala
sunchao: Updated both in 28074b67: the `unbase64` Notes cell in `expressions.md` and the entry in `expression-audits/string_funcs.md` now describe the operator-level fallback. The audit also distinguishes partitioned windows, inherited outer-limit protection, and materialization barriers. I left the generated Implementation cell unchanged.

spark/src/test/resources/sql-tests/expressions/misc/unbase64_operator_masks.sql
sunchao: Added executed coverage in 28074b67. The top-K tests use two ordered partitions, assert there is no intervening sort or exchange, and exercise both `collect()` and RDD collection with the decoder in a child projection, child filter, or output projection.

The join tests disable both static and adaptive broadcast thresholds, force `SHUFFLE_HASH` or `MERGE` for both semi and anti joins, and inspect the executed plans to confirm the strategy and decoder residual. Both groups run with AQE and JVM dispatch on/off, require errors when malformed input is actually consumed, and include native-execution controls. All seven new top-K/join tests pass on Spark 3.5.9 and 4.1.3.

andygrove (commented):
The window group limit narrowing, the two doc files, and the executed top-K and shuffled-hash/sort-merge coverage all look resolved to me, and `RequiresSparkEvaluationMask` is the shape I was hoping for on the trigger.

The performance point from my earlier comment is still open. There is still no way to turn the fallback off, and the description says the cost has not been benchmarked. The reach also grew since I raised it. `hasLimitAncestor` at `CometExecRule.scala:810` is passed down unchanged, so it crosses exchanges and query stages, and with AQE on a `LIMIT` anywhere above a final aggregate whose result expressions decode and whose buffers are not mixed-execution safe now pushes both halves of that aggregate back to Spark. On top of that, every `df.show()` and every BI-tool `LIMIT` over a projection containing `unbase64` gives back what the native kernel in #5451 bought, on data that is almost always well formed. Could this be gated, either on a new `CometConf` key or on the existing `COMET_EXPR_ALLOW_INCOMPATIBLE`, so a user who knows their base64 is clean can keep the native path?

I also still would like a tracking issue for the rest of the family. `CometEvaluationMaskSuite` now asserts that ANSI `Add` and `Cast` return `None` from `evaluationMaskName`, which pins the divergence I described rather than recording it, and `WHERE k = 1 AND unbase64(bad) = X'616263'` and `CASE WHEN k = 1 THEN unbase64(bad) END` are still out of reach for this mechanism with no `LIMIT` involved. Could you file that and link it here?

Two other things worth settling. `spark/src/test/resources/sql-tests/expressions/misc/unbase64_operator_masks.sql` was dropped in the consolidation. The Scala suite runs the same queries, so coverage is not lost, but that fixture was the readable statement of #5532 and it lived where the project asks expression tests to live. Would you keep a minimal version of it? Separately, main has grown `preserveSparkAggregateBuffers` at `CometExecRule.scala:215` from #5537, which un-converts a native Partial back to Spark for the same buffer-compatibility reason as `restoreNativeAggregateBuffers`. On rebase those two should probably become one helper rather than two near-identical walks in the same file.

andygrove (commented):
No change here since my last comment, so the three things I raised are still open: the fallback has no config gate, there is no tracking issue for the rest of the throwing-expression family, and `unbase64_operator_masks.sql` is still gone.

One thing has got worse in the meantime. The consolidation point is now three-way rather than two-way. Main has `preserveSparkAggregateBuffers` from #5537, and #5421 adds `revertUnsafePartialAggregates` with a shared `restoreSparkPartial` helper, so on rebase this PR's `restoreNativeAggregateBuffers` becomes a third near-identical walk over the same aggregate and exchange chain in `CometExecRule.scala`. Worth folding into the shared helper rather than resolving the conflict three ways. Both `CometExecRule.scala` and `CometExpressionSerde.scala` conflict with main now, and the branch is 190 commits behind.

The red check is not yours. The `Spark 3.5, JDK 17, Scala 2.13 [expressions]` job failed on the `Download native library` step about 60 seconds in, back on 31 August. A re-run will clear it.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
sunchao: Addressed in `7df125174`.

The trigger now comes from the serde policy, `RequiresSparkEvaluationMask`, and only `UnBase64` is enrolled, including subclasses inside dispatched expression trees. ANSI Add/Cast returning `None` records the current enrollment scope; I added an explicit link beside those assertions to the broader audit in #6006.

Filed #6006 for the rest of the throwing-expression family, per-row AND/CASE masks, and aggregate FILTER. It calls for skipped-invalid and consumed-invalid controls plus execution-path assertions, rather than assuming every conditional shape currently diverges. The compatibility guide now records this scope as well.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
sunchao: Addressed in `7df125174`.

Added `spark.comet.exec.preserveEvaluationMasks.enabled` (default `true`). Setting it to `false` disables this operator-level fallback, including the conservative AQE aggregate-buffer protection, for users who know their Base64 input is valid. Existing expression support/dispatch restrictions still apply. The config, expression docs, audit, and compatibility guide explain that opting out can raise errors on malformed rows Spark would skip.

The new regression checks native decoding under LIMIT, JVM dispatch for compound inputs, native semi/anti joins, and native aggregate planning under AQE with the gate disabled; enabled controls retain the fallback.

spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
sunchao: Addressed in `7df125174`.

Added the early bailout before the recursive protection pass. It returns the input plan when there is neither an enrolled expression nor a sticky protection tag. The check inspects native operators' original Spark expressions too, so reused AQE/native subtrees still receive protection. The disabled configuration returns immediately before that scan. Existing native-reuse and repeated-planning regressions pass.

sunchao:
Rebased onto current `main` (`d0c10f2d`) and addressed the remaining requests in `7df125174`:

- Added `spark.comet.exec.preserveEvaluationMasks.enabled` (default `true`) with a documented opt-out and native-execution controls.
- Filed #6006 for the broader throwing-expression, conditional, and aggregate FILTER work; linked it from the docs and enrollment assertions.
- Restored a minimal `unbase64_operator_masks.sql` with skipped-malformed, consumed-malformed, dispatcher, and native positive controls.
- Combined this PR's aggregate restoration with `preserveSparkAggregateBuffers` through one helper. It preserves the different boundaries: Celeborn stays within its stage; a falling-back Final may cross unmaterialized shuffle exchanges. #5421 is still open rather than part of the rebased main, so its separate helper is not present in this branch.
- Added the requested early bailout for plans without an enrolled expression or protection tag.

Validated on Spark 4.1.3, Scala 2.13.17, and JDK 17. All **110 executed targeted tests passed**: `CometEvaluationMaskSuite` (9), `CometExecRuleSuite` (32), `CometScalarFunctionSuite` (14), `CometCelebornShufflePlanningSuite` (50), and the applicable `unbase64` SQL registrations (5, including both restored-fixture dispatcher modes). One additional Spark-3.5-only SQL registration skipped its body through the version gate and is excluded from that count.

Full JVM test compilation, Spotless, Scalastyle, and `git diff --check` passed. The JVM tests used the native library from the successful [CI run for the exact rebased main commit](https://github.com/apache/datafusion-comet/actions/runs/35275954309) (`d0c10f2d397ae81b97f187c0dcb4d8f042f2cd9f`), after verifying the artifact digest and that native source/build inputs were unchanged. A local Cargo build was blocked because the configured registry lacks DataFusion 55.1.0; no registry or dependency was changed.

Other Spark/Scala versions, macOS, and the full repository suite were not run locally. Hosted PR CI is separate from these results.

andygrove (commented):
Thanks for adding the config, filing #6006 and restoring the fixture. Those were the three things left from my last round, and they all look right to me.

The thing that has changed since is #5421, which landed on the 20th. This branch now conflicts with `main` in `CometExecRule.scala`, and #5421 renamed `allAggsSupportMixedExecution` to `allAggsSupportNativePartialToSparkFinal`, which three of the new call sites still use. It also added `revertUnsafePartialAggregates`, which runs after `transform` and restores the native Partial chain under any Spark Final whose buffers are unsafe. Once `preserveEvaluationMasks` has tagged a Final, that pass looks like it does what the `throughExchanges = true` path of `restoreNativeAggregateBuffers` does before conversion. When you rebase, could this PR keep only the early tagging of the Final and leave the restoring to `revertUnsafePartialAggregates` and `restoreSparkPartial`, so there's one walk over that chain rather than two? The two AQE aggregate tests in `CometEvaluationMaskSuite` should show quickly whether anything is lost. I'm happy to approve once that's in.

sunchao:
@andygrove Addressed in c7c70faf46d7fed8131088e2f999166c7306ffdf, rebased onto `main` at dd68a531c.

Removed `restoreNativeAggregateBuffers` and the extra restoration walk from `preserveEvaluationMasks`. The rule keeps early Final tagging so fresh partials choose compatible buffers before AQE materialization, then relies on the existing `revertUnsafePartialAggregates` / `restoreSparkPartial` path from #5421 for restoration. Those upstream helpers are unchanged, and the remaining predicate uses `allAggsSupportNativePartialToSparkFinal`.

Both AQE aggregate regressions passed, including the fresh/reused native-plan matrix and the actual join change. The focused evaluation-mask, rule, Celeborn, and SQL tests passed: 106 executed tests on Spark 4.1.3 (107 successful registrations including one Spark-3.5-gated body). `cargo build --locked`, JVM compilation, Spotless, Scalastyle, and diff checks also passed. JVM execution used the verified native artifact for the exact new base, whose native inputs are unchanged by this PR.

Updated the description and requested Spark SQL 4.1 CI coverage with `run-spark-4.1-tests`; hosted checks are pending.

PR #3731, opened 2026-03-18
perf: stop using FFI in native shuffle read path

andygrove:
## Which issue does this PR close?

Performance improvements for native shuffle read. Shows 13% improvement in TPC-H @ 1TB.

## Rationale for this change

Simplifies the shuffle direct read code path, removing unnecessary FFI transfers.

PR #5538, opened 2026-08-29
perf: add direct native reads for broadcast exchange

### LINKED ISSUE #3762: perf: bypass Arrow FFI for broadcast exchange reads
## Summary

When a native operator reads from a broadcast exchange, the data currently passes through Arrow FFI via `ScanExec`. This is the same path that shuffle reads used before `ShuffleScanExec` was introduced in #3731.

## Current Path

1. `CometBroadcastExchangeExec` serializes data as coalesced, compressed Arrow IPC on the driver
2. On each executor, `CometBatchRDD.compute()` deserializes the IPC stream → `ColumnarBatch` objects on the JVM
3. A `CometBatchIterator` exposes those batches to native code
4. Native `ScanExec` reads via Arrow FFI (`CometBatchIterator.next()` exports via Arrow C Data Interface)

## Proposed Path

Introduce a `BroadcastScanExec` native operator analogous to `ShuffleScanExec`:

1. Pass raw compressed Arrow IPC bytes from the JVM to native code (skip JVM-side deserialization)
2. Decompress and decode natively using `read_ipc_compressed()`
3. This avoids the JVM-side decompression + `ArrowStreamReader` deserialization + Arrow FFI export overhead

## Work Required

- New JVM iterator class (e.g. `CometBroadcastBlockIterator`) that yields raw compressed IPC bytes, similar to `CometShuffleBlockIterator`
- New protobuf op `BroadcastScan` alongside existing `ShuffleScan`
- New Rust operator `BroadcastScanExec` in `native/core/src/execution/operators/`
- Planner changes in `CometSink.scala` to emit `BroadcastScan` instead of `Scan` for broadcast exchange inputs (behind a config flag like `spark.comet.exec.broadcast.directRead.enabled`)

## Notes

- Broadcast data is already marked `arrow_ffi_safe=true` since it comes from `ArrowStreamReader` with no mutable buffers, so the current FFI path is at least safe
- Broadcast volumes are typically smaller than shuffle, so the impact may be less dramatic than #3731, but the overhead of JVM-side decompression and FFI export is still real

_No response_

### Additional context

_No response_

goutamadwant:

Closes #3762.

## Rationale for this change

Broadcast exchange input currently crosses the JVM twice before native execution: the JVM decompresses and deserializes Arrow IPC, then exports the resulting arrays through Arrow FFI. Comet can read the codec-prefixed IPC blocks directly and avoid that work, as it already does for direct shuffle reads.

- Add the experimental `spark.comet.exec.broadcast.directRead.enabled` configuration, disabled by default.
- Serialize direct-read broadcast batches as codec-prefixed LZ4 Arrow IPC while keeping them decodable by JVM consumers.
- Add a `BroadcastScan` protobuf operator and native `BroadcastScanExec` planning path that reuse the existing direct block protocol.
- Wire broadcast blocks through AQE and exchange-reuse boundaries without creating Arrow stream inputs.
- Keep direct-read mode in broadcast exchange equality and canonicalization so exchanges using incompatible serialization modes cannot be reused together.
- Preserve independently serialized blocks instead of coalescing a native broadcast beyond the JNI protocol's per-block `Integer.MAX_VALUE` limit.
- Preserve the legacy broadcast serialization and Arrow FFI path when the configuration is disabled.
- Close any already resolved streams and block iterators when input-slot resolution or native plan construction fails.

- `make core`
- `cargo check -p datafusion-comet`
- `cargo clippy -p datafusion-comet --all-targets -- -D warnings`
- `cargo test -p datafusion-comet test_broadcast_scan_has_distinct_plan_name`
- `./mvnw test -Dtest=none -Dsuites=org.apache.spark.sql.comet.util.UtilsSuite -DfailIfNoTests=false`
- `./mvnw test -Dtest=none -Dsuites=org.apache.comet.exec.CometNativeShuffleSuite -DfailIfNoTests=false`
- `./mvnw test -Dtest=none -Dsuites=org.apache.comet.exec.CometExecSuite -DfailIfNoTests=false`

The regression coverage checks legacy serialization with the feature disabled, native codec framing with the feature enabled, JVM decoding compatibility, AQE enabled and disabled, exchange identity across serialization modes, the direct block-size boundary without a large allocation, native-plan failure cleanup, empty build-side handling, and Spark-equivalent join answers.

### DIFF
--- a/native/core/src/execution/operators/shuffle_scan.rs
+++ b/native/core/src/execution/operators/shuffle_scan.rs
@@ -47,11 +47,35 @@ use std::{
 
 use super::scan::InputBatch;
 
-/// ShuffleScanExec reads compressed shuffle blocks from JVM via JNI and decodes them natively.
-/// Unlike ScanExec which receives Arrow arrays via FFI, ShuffleScanExec receives raw compressed
-/// bytes from CometShuffleBlockIterator and decodes them using read_ipc_compressed().
+/// Identifies the JVM source using the shared direct-block protocol.
+#[derive(Debug, Clone, Copy)]
+enum BlockScanKind {
+    Shuffle,
+    Broadcast,
+}
+
+impl BlockScanKind {
+    fn execution_plan_name(self) -> &'static str {
+        match self {
+            Self::Shuffle => "ShuffleScanExec",
+            Self::Broadcast => "BroadcastScanExec",
+        }
+    }
+
+    fn input_name(self) -> &'static str {
+        match self {
+            Self::Shuffle => "shuffle",
+            Self::Broadcast => "broadcast",
+        }
+    }
+}
+
+/// Reads codec-prefixed compressed Arrow IPC blocks from JVM via JNI and decodes them natively.
+/// Shuffle and broadcast scans share the same iterator protocol while retaining distinct plan
+/// names for explain output and diagnostics.
 #[derive(Debug, Clone)]
 pub struct ShuffleScanExec {
+    kind: BlockScanKind,
     /// The ID of the execution context that owns this subquery.
     pub exec_context_id: i64,
     /// The input source: a global reference to a JVM CometShuffleBlockIterator object.
@@ -80,6 +104,33 @@ impl ShuffleScanExec {
         input_source: Option<Arc<Global<JObject<'static>>>>,
         data_types: Vec<DataType>,
     ) -> Result<Self, CometError> {
+        Self::new_with_kind(
+            exec_context_id,
+            input_source,
+            data_types,
+            BlockScanKind::Shuffle,
+        )
+    }
+
+    pub fn new_broadcast(
+        exec_context_id: i64,
+        input_source: Option<Arc<Global<JObject<'static>>>>,
+        data_types: Vec<DataType>,
+    ) -> Result<Self, CometError> {
+        Self::new_with_kind(
+            exec_context_id,
+            input_source,
+            data_types,
+            BlockScanKind::Broadcast,
+        )
+    }
+
+    fn new_with_kind(
+        exec_context_id: i64,
+        input_source: Option<Arc<Global<JObject<'static>>>>,
+        data_types: Vec<DataType>,
+        kind: BlockScanKind,
+    ) -> Result<Self, CometError> {
         let requires_validation = if exec_context_id == TEST_EXEC_CONTEXT_ID {
             false
         } else if let Some(input) = &input_source {
@@ -90,6 +141,7 @@ impl ShuffleScanExec {
         } else {
             false
         };
+
         let metrics_set = ExecutionPlanMetricsSet::default();
         let baseline_metrics = BaselineMetrics::new(&metrics_set, 0);
         let decode_time = MetricBuilder::new(&metrics_set).subset_time("decode_time", 0);
@@ -104,6 +156,7 @@ impl ShuffleScanExec {
         ));
 
         Ok(Self {
+            kind,
             exec_context_id,
             input_source,
             data_types,
@@ -139,6 +192,7 @@ impl ShuffleScanExec {
                 &self.data_types,
                 &self.decode_time,
                 self.requires_validation,
+                self.kind,
             )?;
             *current_batch = Some(next_batch);
         }
@@ -155,6 +209,7 @@ impl ShuffleScanExec {
         data_types: &[DataType],
         decode_time: &Time,
         requires_validation: bool,
+        kind: BlockScanKind,
     ) -> Result<InputBatch, CometError> {
         if exec_context_id == TEST_EXEC_CONTEXT_ID {
             return Ok(InputBatch::EOF);
@@ -162,7 +217,8 @@ impl ShuffleScanExec {
 
         if iter.is_null() {
             return Err(CometError::from(ExecutionError::GeneralError(format!(
-                "Null shuffle block iterator object. Plan id: {exec_context_id}"
+                "Null {} block iterator object. Plan id: {exec_context_id}",
+                kind.input_name()
             ))));
         }
 
@@ -219,6 +275,14 @@ impl ShuffleScanExec {
                 .map(|col| unpack_dictionary(col))
                 .collect();
 
+            debug_assert_eq!(
+                columns.len(),
+                data_types.len(),
+                "{} block column count mismatch: got {} but expected {}",
+                kind.input_name(),
+                columns.len(),
+                data_types.len()
+            );
             Ok(InputBatch::new(columns, Some(num_rows)))
         })
     }
@@ -302,7 +366,7 @@ impl ExecutionPlan for ShuffleScanExec {
     }
 
     fn name(&self) -> &str {
-        "ShuffleScanExec"
+        self.kind.execution_plan_name()
     }
 
     fn metrics(&self) -> Option<MetricsSet> {
@@ -320,7 +384,7 @@ impl DisplayAs for ShuffleScanExec {
                     .enumerate()
                     .map(|(idx, dt)| format!("col_{idx}: {dt}"))
                     .collect();
-                write!(f, "ShuffleScanExec: schema=[{}]", fields.join(", "))?;
+                write!(f, "{}: schema=[{}]", self.name(), fields.join(", "))?;
             }
             DisplayFormatType::TreeRender => unimplemented!(),
         }
@@ -508,6 +572,20 @@ mod tests {
                 .to_string();
             assert!(error.contains("column count mismatch"), "{error}");
         }
+    }
+
+    #[test]
+    fn test_broadcast_scan_has_distinct_plan_name() {
+        use datafusion::physical_plan::ExecutionPlan;
+
+        let scan = super::ShuffleScanExec::new_broadcast(
+            super::super::super::planner::TEST_EXEC_CONTEXT_ID,
+            None,
+            vec![DataType::Int32],
+        )
+        .unwrap();
+
+        assert_eq!(scan.name(), "BroadcastScanExec");
     }
 
     #[test]
--- a/spark/src/main/java/org/apache/comet/CometBroadcastBlockIterator.java
+++ b/spark/src/main/java/org/apache/comet/CometBroadcastBlockIterator.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.comet;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Iterator;
+
+/** Provides codec-prefixed Arrow IPC broadcast blocks to native code via JNI. */
+public final class CometBroadcastBlockIterator extends CometShuffleBlockIterator {
+
+  private static final int INITIAL_BUFFER_SIZE = 128 * 1024;
+
+  private Iterator<ByteBuffer[]> blocks;
+  private ByteBuffer dataBuf = ByteBuffer.allocateDirect(INITIAL_BUFFER_SIZE);
+  private boolean closed = false;
+  private int currentBlockLength = 0;
+
+  public CometBroadcastBlockIterator(Iterator<ByteBuffer[]> blocks) {
+    // Native uses the same block-iterator JNI protocol for shuffle and broadcast inputs. The
+    // superclass stream is never read because every protocol method is overridden here.
+    super(new ByteArrayInputStream(new byte[0]));
+    this.blocks = blocks;
+  }
+
+  @Override
+  public int hasNext() throws IOException {
+    if (closed) {
+      return -1;
+    }
+
+    ByteBuffer[] block = null;
+    long blockSize = 0;
+    while (blocks.hasNext() && block == null) {
+      ByteBuffer[] candidate = blocks.next();
+      long candidateSize = 0;
+      for (ByteBuffer chunk : candidate) {
+        candidateSize += chunk.remaining();
+      }
+      if (candidateSize > 0) {
+        block = candidate;
+        blockSize = candidateSize;
+      }
+    }
+    if (block == null) {
+      close();
+      return -1;
+    }
+
+    if (blockSize > Integer.MAX_VALUE) {
+      throw new IllegalStateException(
+          "Native broadcast block size of "
+              + blockSize
+              + " exceeds the direct-read maximum of "
+              + Integer.MAX_VALUE
+              + " bytes");
+    }
+
+    currentBlockLength = (int) blockSize;
+    if (dataBuf.capacity() < currentBlockLength) {
+      long doubled = Math.max((long) dataBuf.capacity() * 2L, blockSize);
+      dataBuf = ByteBuffer.allocateDirect((int) Math.min(doubled, Integer.MAX_VALUE));
+    }
+
+    dataBuf.clear();
+    dataBuf.limit(currentBlockLength);
+    for (ByteBuffer chunk : block) {
+      dataBuf.put(chunk.duplicate());
+    }
+    return currentBlockLength;
+  }
+
+  @Override
+  public ByteBuffer getBuffer() {
+    return dataBuf;
+  }
+
+  @Override
+  public int getCurrentBlockLength() {
+    return currentBlockLength;
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (!closed) {
+      closed = true;
+      blocks = null;
+      super.close();
+    }
+  }
+}
--- a/spark/src/main/scala/org/apache/comet/CometConf.scala
+++ b/spark/src/main/scala/org/apache/comet/CometConf.scala
@@ -229,6 +229,16 @@ object CometConf extends ShimCometConf {
     createExecEnabledConfig("broadcastHashJoin", defaultValue = true)
   val COMET_EXEC_BROADCAST_EXCHANGE_ENABLED: ConfigEntry[Boolean] =
     createExecEnabledConfig("broadcastExchange", defaultValue = true)
+  val COMET_EXEC_BROADCAST_DIRECT_READ_ENABLED: ConfigEntry[Boolean] =
+    conf(s"$COMET_EXEC_CONFIG_PREFIX.broadcast.directRead.enabled")
+      .category(CATEGORY_EXEC)
+      .doc(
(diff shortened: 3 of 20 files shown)


spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala
```diff
@@ -332,12 +348,38 @@ object Utils extends CometTypeShim with Logging {
...
-    new ArrowReaderIterator(Channels.newChannel(ins), source)
+    new ArrowReaderIterator(Channels.newChannel(compressedIpcInputStream(bytes)), source)
+  }
+
+  private def compressedIpcOutputStream(
+      output: ChunkedByteBufferOutputStream,
+      nativeIpc: Boolean): DataOutputStream = {
+    if (nativeIpc) {
+      output.write(NATIVE_IPC_LZ4_PREFIX)
+      new DataOutputStream(new net.jpountz.lz4.LZ4FrameOutputStream(output))
```
andygrove, line 360: `LZ4FrameOutputStream(output)` defaults to `BLOCKSIZE.SIZE_4MB`. I measured allocation against lz4-java 1.8.0 and it comes to about 12.6MB of heap per instance, versus about 135KB with `SIZE_64KB`. `serializeBatches` builds one of these per `ColumnarBatch`, and Spark's `LZ4CompressionCodec` on the path this replaces uses a 32KB block by default, so this is roughly a 90x jump in per-batch allocation.

The scaladoc on `coalesceBroadcastBatches` just below describes 200K tiny batches arriving from a 400-task, 500-partition shuffle, which is exactly the shape that would hurt. Could we pass `LZ4FrameOutputStream.BLOCKSIZE.SIZE_64KB` here? That fixes the read side at the same time, since `LZ4FrameInputStream` sizes its buffers from the frame's `BD` byte.

spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala
```diff
@@ -49,6 +50,8 @@ import org.apache.comet.shims.CometTypeShim
 import org.apache.comet.vector.CometVector
 
 object Utils extends CometTypeShim with Logging {
+  private val NATIVE_IPC_LZ4_PREFIX = Array[Byte](0x4c, 0x5a, 0x34, 0x5f)
```
andygrove, line 53: Should the direct-read format follow the configured codec rather than always writing LZ4? Native `read_ipc_compressed` handles `SNAP`, `ZSTD` and `NONE` as well as `LZ4_`, and `spark.comet.exec.shuffle.compression.codec` already exists with a `zstd`/`lz4`/`snappy` check. As it stands someone who configured zstd silently gets LZ4 for broadcast.

spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala
```diff
@@ -332,12 +348,38 @@ object Utils extends CometTypeShim with Logging {
...
+  }
+
+  /**
+   * Opens either the existing Spark-codec stream or the codec-prefixed direct-read stream. This
+   * keeps non-native broadcast consumers compatible when direct read is enabled.
+   */
+  private def compressedIpcInputStream(bytes: ChunkedByteBuffer): InputStream = {
+    val input = new PushbackInputStream(bytes.toInputStream(), NATIVE_IPC_LZ4_PREFIX.length)
+    val prefix = new Array[Byte](NATIVE_IPC_LZ4_PREFIX.length)
+    val read = input.read(prefix)
```
andygrove, line 374: `input.read(prefix)` can come back with fewer than 4 bytes, and a short read here falls through to the Spark codec branch on what is actually a native-IPC stream. I do not think it can happen in practice because the chunks are 1MB, but `input.readNBytes(4)` would take the question off the table.

spark/src/main/scala/org/apache/spark/sql/comet/CometBroadcastExchangeExec.scala
```diff
@@ -111,7 +113,11 @@ case class CometBroadcastExchangeExec(
 
   private def getByteArrayRdd(plan: SparkPlan): RDD[(Long, ChunkedByteBuffer)] = {
     plan.executeColumnar().mapPartitionsInternal { iter =>
-      Utils.serializeBatches(iter)
+      if (directRead) {
+        Utils.serializeBroadcastBatches(iter)
```
andygrove, line 117: Reading `directRead` inside the lambda makes it capture the plan node. `mapPartitionsInternal` does not run `sc.clean`, and I confirmed against the compiled class that the closure is now `$anonfun$getByteArrayRdd$1(CometBroadcastExchangeExec, Iterator)` where before it captured nothing. That ships the whole exchange, including `originalPlan` and the child subtree, with every build-side task.

Could you hoist `val useNativeIpc = directRead` above the closure and branch on that instead?

spark/src/main/scala/org/apache/spark/sql/comet/CometBroadcastExchangeExec.scala
```diff
@@ -64,7 +65,8 @@ case class CometBroadcastExchangeExec(
     originalPlan: SparkPlan,
     override val output: Seq[Attribute],
     mode: BroadcastMode,
-    override val child: SparkPlan)
+    override val child: SparkPlan,
+    directRead: Boolean = CometConf.COMET_EXEC_BROADCAST_DIRECT_READ_ENABLED.get())
```
andygrove, line 69: Two questions about the conf-reading default here.

There are two existing construction sites that do not pass this parameter and will silently inherit whatever the conf says at that moment: `CometExecRule.scala:527`, which builds the exchange for `CometSubqueryBroadcastExec`, and `CometPlanAdaptiveDynamicPruningFilters.scala:279`, whose comment says the fresh exchange has to have the same canonical form as the join's exchange so AQE's stage cache produces a `ReusedExchangeExec`. Now that `doCanonicalize` includes `directRead`, that reuse depends on a session conf resolving identically at two different points in planning, one of which is on an AQE thread. Dropping the default so the compiler forces each site to state its value would make the coupling explicit. My DPP test passes today so this is latent rather than live, but it is cheap to close off.

Separately, what does including `directRead` in `doCanonicalize` protect against? The serialized format is self-describing and `decodeBatches` sniffs the prefix, so two exchanges with different modes look reusable to me anyway.

spark/src/main/scala/org/apache/comet/CometConf.scala
```diff
@@ -229,6 +229,16 @@ object CometConf extends ShimCometConf {
     createExecEnabledConfig("broadcastHashJoin", defaultValue = true)
   val COMET_EXEC_BROADCAST_EXCHANGE_ENABLED: ConfigEntry[Boolean] =
     createExecEnabledConfig("broadcastExchange", defaultValue = true)
+  val COMET_EXEC_BROADCAST_DIRECT_READ_ENABLED: ConfigEntry[Boolean] =
+    conf(s"$COMET_EXEC_CONFIG_PREFIX.broadcast.directRead.enabled")
```
andygrove, line 233: The shuffle equivalent is `spark.comet.shuffle.directRead.enabled`. #4986 deliberately moved these out of `spark.comet.exec.*` and kept the old name only as an alternative, so this lands back under the prefix we are moving away from. Should it be `spark.comet.broadcast.directRead.enabled`? Easier to settle now than to add a deprecation alias once it has shipped.

native/core/src/execution/operators/shuffle_scan.rs
```diff
@@ -219,6 +275,14 @@ impl ShuffleScanExec {
                 .map(|col| unpack_dictionary(col))
                 .collect();
 
+            debug_assert_eq!(
```
andygrove, line 278: This looks redundant. `check_column_count` already returns a real error on the non-validating path, `validate_remote_schema` covers the validating one, and `cast_and_stamp_schema` checks again in `poll_next`. In release builds the assert compiles away entirely, and it is the only reason `kind` has to be threaded down into `get_next`. Can it come out?

spark/src/main/java/org/apache/comet/CometBroadcastBlockIterator.java
```diff
@@ -0,0 +1,109 @@
...
+
+package org.apache.comet;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Iterator;
+
+/** Provides codec-prefixed Arrow IPC broadcast blocks to native code via JNI. */
+public final class CometBroadcastBlockIterator extends CometShuffleBlockIterator {
```
andygrove, line 28: Extending `CometShuffleBlockIterator` and handing it a throwaway `ByteArrayInputStream` means the superclass constructor still allocates a 128KB direct `ByteBuffer` and a 16-byte header buffer per instance, neither of which this class ever touches, and there is one iterator per consumer partition.

Since every protocol method is overridden anyway, would it be cleaner to extract a small interface for the JNI block protocol and have both classes implement it? The native `comet_shuffle_block_iterator` bridge would keep working, and the inheritance would stop implying shared state that is not there.

andygrove (commented):
Thanks for picking this up @goutamadwant, and for the care in the cleanup paths. Wrapping `createPlan` so the Arrow streams and block iterators get closed when it throws is a real leak fix that stands on its own, and the same goes for the rollback in `resolveInputObjects`.

I checked the wire format end to end locally rather than just reading it. I fed real `net.jpountz.lz4.LZ4FrameOutputStream` output wrapping an arrow-rs IPC stream into `read_ipc_compressed`, at 60KB (one LZ4 block) and at 26MB (seven blocks), and it decodes correctly under the strict `RequireLz4EndMark` and trailing-data checks that #5531 added. lz4-java writes `FLG=0x60`, so block-independent with no block or content checksum, which is exactly what `lz4_flex` handles. That was my main worry going in and it is fine. The prefix sniffing in `decodeBatches` keeping JVM consumers working is a nice touch.

I also ran the feature against a set of cases the PR does not cover, each asserting that a `BroadcastScan` really shows up in the native plan rather than silently falling back: nested and complex build-side types (struct, array, map, decimal, timestamp, date, binary), non-UTC session timezones with both `timestamp` and `timestamp_ntz`, a reused broadcast exchange, a broadcast nested loop join, DPP, `spark.comet.shuffle.directRead.enabled` on at the same time so `findBlockScanIndices` sees mixed shuffle and broadcast slots, and direct tests of `CometBroadcastBlockIterator` over multiple blocks including the buffer-growth branch. They all pass, so I did not find a correctness problem. Since they pass as written they should be cheap to add and I think they are worth having. The multi-block one especially, because every test in the PR produces a single coalesced block, so the iterator never advances past its first block and the growth branch never runs.

The main thing I would like addressed is the allocation cost on the write side, which I left a comment on in `Utils.scala`. Related to that, this is a `perf:` PR with no numbers in it. Could you add a microbenchmark under `spark/src/test/scala/org/apache/spark/sql/benchmark/` and post before and after with the flag off versus on? I would want to see the write side in particular, given the block-size point.

One more that does not fit on a line in the diff. `Utils.scala` now imports `net.jpountz.lz4` directly, but `grep -rn "lz4" --include=pom.xml .` finds nothing. It only compiles because Spark pulls lz4-java in transitively under `provided` scope. Can we declare it explicitly so a Spark dependency change cannot break our build without warning?

CI has not run on this yet, so none of the above is informed by it.

native/core/src/execution/operators/shuffle_scan.rs
```diff
@@ -47,11 +47,35 @@ use std::{
...
+            Self::Broadcast => "broadcast",
+        }
+    }
+}
+
+/// Reads codec-prefixed compressed Arrow IPC blocks from JVM via JNI and decodes them natively.
+/// Shuffle and broadcast scans share the same iterator protocol while retaining distinct plan
+/// names for explain output and diagnostics.
 #[derive(Debug, Clone)]
 pub struct ShuffleScanExec {
```
peterxcli, line 77:
```suggestion
pub struct BlockScanExec {
```
rename to `BlockScanExec` as it now reads codec-prefixed IPC blocks from either shuffle or broadcast.

native/core/src/execution/operators/shuffle_scan.rs
peterxcli, line 102:
```suggestion
    pub fn new_with_shuffle(
```

native/core/src/execution/operators/shuffle_scan.rs
```diff
@@ -79,6 +103,33 @@ impl ShuffleScanExec {
...
+            data_types,
+            BlockScanKind::Broadcast,
+        )
+    }
+
+    fn new_with_kind(
+        exec_context_id: i64,
+        input_source: Option<Arc<Global<JObject<'static>>>>,
+        data_types: Vec<DataType>,
+        kind: BlockScanKind,
```
peterxcli, line 132:
```suggestion
    pub fn new(
        exec_context_id: i64,
        input_source: Option<Arc<Global<JObject<'static>>>>,
        data_types: Vec<DataType>,
        kind: BlockScanKind,
```
remove the selected constructor and only keep new\_with_kind and rename new_with_kind to new?

peterxcli (commented):
do you have any tpch benchmark result
FYI: https://github.com/apache/datafusion-comet/pull/3731 run tpch @ 1TB

spark/src/main/scala/org/apache/spark/sql/comet/CometBroadcastExchangeExec.scala
```diff
@@ -85,7 +87,7 @@ case class CometBroadcastExchangeExec(
       "number of coalesced rows for broadcast"))
 
   override def doCanonicalize(): SparkPlan = {
-    CometBroadcastExchangeExec(null, null, mode, child.canonicalized)
+    CometBroadcastExchangeExec(null, null, mode, child.canonicalized, directRead)
```
peterxcli, line 90: Keeping `directRead` in canonicalization seems necessary: `decodeBatches` accepts both formats, but the native block reader requires codec-prefixed IPC. In non-AQE execution, exchange reuse runs after columnar conversion, so the consumer can already contain a `BroadcastScan` when its exchange is replaced.

Could we document this distinction and add a reuse regression covering a native join and JVM DPP consumer? The DPP construction sites should also carry the format choice explicitly so they match the join's exchange.

spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala
```diff
@@ -358,26 +400,36 @@ object Utils extends CometTypeShim with Logging {
    * a single Arrow IPC stream.
    */
   def coalesceBroadcastBatches(
-      input: Iterator[ChunkedByteBuffer]): (Array[ChunkedByteBuffer], Long, Long) = {
+      input: Iterator[ChunkedByteBuffer],
+      nativeIpc: Boolean = false): (Array[ChunkedByteBuffer], Long, Long) = {
```
peterxcli, line 404: Alongside TPC-H, could the benchmark include many small build-side batches and report producer serialization, driver coalescing, allocation/GC, and total query time with direct read off/on?

Coalescing still decodes each input stream on the JVM and serializes the combined batch again, so consumer-only timings would miss the allocation cost raised in the LZ4 thread. These measurements would also help decide whether coalescing needs a separate optimization.

peterxcli (commented):
second round

sunchao (commented):
Reviewed `af383e226f549938ec198f9f79542c2ee2ffe7d5` against `98cd8c967995906500b12a709c25724b4cfba634`, covering native/JNI ownership, Spark planning and AQE/reuse, cleanup/error paths, codec compatibility, and boundary cases.

I found two additional P2 issues, detailed inline: coalescing can produce a compressed block larger than the JNI limit even when the input-size check passes, and closed broadcast iterators retain the full direct staging buffer throughout the remaining consumer lifetime.

Validation:

- Fresh local native build passed. The native broadcast scan test passed (1/1). Root-reactor Spark 4.1/JDK 17 tests passed: `UtilsSuite` (10/10), direct broadcast with AQE on/off and an empty build side (2/2), and native-plan failure cleanup (1/1).
- Independently reproduced the buffer retention using the unchanged production iterator classes, including EOF, explicit close, GC, and a control that drops only the buffer reference. Separately reproduced compressed-size expansion with the same Arrow/LZ4 appender and writer operations, and confirmed the production iterator rejects the measured output size.
- Current GitHub check runs show 63 successful and 9 skipped, with no failures or running checks. The inspected CI merge commit has the identical full source tree as this head. [Native CI](https://github.com/apache/datafusion-comet/actions/runs/33234520474/job/99594564250) passed 1,018 tests (4 skipped); the [Spark 4.1 execution shard](https://github.com/apache/datafusion-comet/actions/runs/33234520474/job/99603180590) passed 747 tests (1 canceled, 3 ignored), including the new broadcast cases.

The large-block evidence is a component reproduction, not a full large Spark query. The retention probe does not establish an end-to-end OOM, and these checks do not establish a query-performance improvement. Existing discussion already covers the other allocation, closure-capture, and configuration concerns.

spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala
```diff
@@ -358,26 +400,36 @@ object Utils extends CometTypeShim with Logging {
...
     val buffers = input.filterNot(_.size == 0).toArray
     if (buffers.isEmpty) {
       return (Array.empty, 0L, 0L)
     }
 
+    val totalInputBytes = buffers.foldLeft(0L) { (total, buffer) =>
+      if (Long.MaxValue - total < buffer.size) Long.MaxValue else total + buffer.size
+    }
+    if (shouldSkipDirectBroadcastCoalesce(nativeIpc, totalInputBytes)) {
```
sunchao, line 413: [P2] Check the coalesced output against the JNI block limit

The sum of the compressed inputs is not an upper bound on the compressed output: concatenating Arrow columns can remove the short-distance matches LZ4 used within each input batch. With direct read enabled, a large hinted broadcast can therefore pass this guard and still produce a single block that `CometBroadcastBlockIterator.hasNext()` rejects.

I reproduced this with the same Arrow 18.3 / LZ4 appender and writer operations on correlated BIGINT columns: the separate IPC streams total **551,523,846 bytes**, but the coalesced stream is **2,147,620,180 bytes**, above `Integer.MAX_VALUE`. The row count and resulting broadcast are below the existing row and default 8 GiB broadcast limits. The unchanged iterator rejects that measured size. This is a component reproduction, not a full Spark query run.

Please also validate the actual compressed output size and return the original blocks when it exceeds the JNI limit, or emit bounded independent IPC batches. The current test only exercises the input-size predicate, so it misses this case.

spark/src/main/java/org/apache/comet/CometBroadcastBlockIterator.java
```diff
@@ -0,0 +1,109 @@
...
+  @Override
+  public int getCurrentBlockLength() {
+    return currentBlockLength;
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (!closed) {
+      closed = true;
+      blocks = null;
```
sunchao, line 105: [P2] Drop the direct broadcast buffer when the iterator closes

`close()` clears the source blocks but retains this class's `dataBuf`. After a coalesced broadcast is consumed, that buffer holds a complete compressed copy of the build relation per consumer task. `CometExecIterator` retains the block iterator while the join continues probing, so this direct allocation stays live alongside the decoded build data even after EOF; `super.close()` cannot clear the subclass buffer.

Compiling the unchanged iterator classes and feeding a 32 MiB block leaves **33,685,504 bytes** of direct memory after EOF and forced GC. Explicit close has the same result. Dropping only the subclass's `dataBuf` reference, while keeping the closed iterator alive, reclaims exactly **33,554,432 bytes**. This verifies retention, not an end-to-end OOM.

Please drop the staging-buffer reference on close/EOF and cover buffer growth followed by close. Native decoding finishes consuming the compressed bytes before the next `hasNext()` call, so the buffer is no longer needed at EOF.

PR #5539, opened 2026-08-29
fix: prevent memory leak after failed Arrow vector import

### LINKED ISSUE #5534: NativeUtil.getNextBatch leaks Arrow structs when importVector fails

**Component**: `spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala:209`

#5531 added `releaseArrowStructs` to two of the three exits from `getNextBatch`: the `func` failure
path (line 197) and the EOF path (line 206). The third exit, `importVector(arrays, schemas)` on
line 209, is not guarded.

When `importer.importVector` throws part-way through the column loop, every struct from the failing
column onward keeps its populated C data with the release callback never invoked, and all
`2 * numCols` `ArrowArray` / `ArrowSchema` wrapper buffers leak from the allocator.

This is on every native operator path, not just shuffle. The most reachable trigger is an allocator
OOM inside `importVector` under memory pressure, which is exactly when leaking is worst.

## Reproduction

Add to `NativeUtilSuite`, reusing the harness #5531 introduced:

```scala
test("getNextBatch leaks Arrow structs when importVector fails") {
  withIsolatedStructAllocator { (nativeUtil, allocator, _) =>
    val vector = new IntVector("value", allocator)
    vector.allocateNew(4)
    vector.setSafe(0, 42)
    vector.setValueCount(1)
    intercept[Throwable] {
      nativeUtil.getNextBatch(2, (arrays, schemas) => {
        // Column 0 exported; column 1 left as a freshly allocated struct, which is what native
        // leaves behind if it populates some columns and the importer then rejects one.
        Data.exportVector(allocator, vector, null,
          ArrowArray.wrap(arrays(0)), ArrowSchema.wrap(schemas(0)))
        vector.close()
        1L
      })
    }
    assert(allocator.getAllocatedMemory == 0) // fails: 176
  }
}
```

Observed:

```
importVector failure          = java.lang.IllegalStateException: Cannot import released ArrowSchema
allocated after failure       = 176
```

The success path is clean by comparison, so this is specific to the import failure:

```
success allocated before close = 48
success allocated after close  = 0
```

## Suggested fix

Wrap the `importVector` call in the same `try` / `releaseArrowStructs(arrays, schemas, failure)`
pattern the other two exits use. The comment at `NativeBatchDecoderIterator.scala:105` already
treats "Arrow import failures" as a first-class case, so this looks like an oversight rather than a
decision.

Reproduced on Spark 4.1.3 / Scala 2.13 / JDK 17 at merge commit `98cd8c967`.

1fanwang:

Closes https://github.com/apache/datafusion-comet/issues/5534.

## Rationale for this change

Failed Arrow imports retain native buffers in Comet executors. The allocator
then reports errors such as `Memory was leaked by query. Memory leaked: (176)`,
which adds pressure while the query is already failing.

`ArrowImporter` keeps each new `FieldVector` rollback-owned until array import
succeeds. `NativeUtil.importVector` keeps that ownership through Comet wrapping
and publication, then closes earlier vectors and untouched C Data structs when
a later column fails.

Dictionary values live in the provider rather than in the failed column vector.
The rollback now closes those vectors too, so their native buffers are released
at the failure instead of task shutdown. Cleanup failures remain suppressed on
the original exception.

Current upstream leaves allocator memory charged on every covered failure path.
The branch preserves each original error and returns the allocator to zero under
Spark 3.5 and Spark 4.1.

<details>
<summary>Raw test output</summary>

```console
$ BASE=/tmp/comet-base
$ TEST=spark/src/test/scala/org/apache/comet/vector/NativeUtilSuite.scala
$ git worktree add --detach "$BASE" e0fa3ccb16d4c9b3a03a38b09af6ed0fb242e027
$ git archive 21ae48db7da56d1a0c172205a02d6f4c8722e97e "$TEST" | tar -x -C "$BASE"
$ cd "$BASE"
$ JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -Prelease -Pspark-4.1 test \
    -Dtest=none -Dsuites=org.apache.comet.vector.NativeUtilSuite
Memory was leaked by query. Memory leaked: (176)
Memory was leaked by query. Memory leaked: (288)
Memory was leaked by query. Memory leaked: (33592)
Memory was leaked by query. Memory leaked: (66536)
Memory was leaked by query. Memory leaked: (192)
Tests: succeeded 4, failed 5, canceled 0, ignored 0, pending 0
BUILD FAILURE

$ GREEN=/tmp/comet-green
$ git worktree add --detach "$GREEN" 21ae48db7da56d1a0c172205a02d6f4c8722e97e
$ cd "$GREEN"
$ JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -Prelease -Pspark-4.1 test \
    -Dtest=none -Dsuites=org.apache.comet.vector.NativeUtilSuite
Tests: succeeded 9, failed 0, canceled 0, ignored 0, pending 0
All tests passed.
BUILD SUCCESS
```

</details>

### DIFF
--- a/spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
+++ b/spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
@@ -50,10 +50,32 @@ public class ArrowImporter {
 
   public FieldVector importVector(
       ArrowArray array, ArrowSchema schema, CDataDictionaryProvider provider) {
-    Field field = importField(schema, provider);
-    FieldVector vector = field.createVector(allocator);
-    ArrayImporter importer = new ArrayImporter(allocator, vector, provider);
-    importer.importArray(array);
-    return vector;
+    FieldVector vector = null;
+    try {
+      Field field = importField(schema, provider);
+      vector = field.createVector(allocator);
+      ArrayImporter importer = new ArrayImporter(allocator, vector, provider);
+      importer.importArray(array);
+      return vector;
+    } catch (RuntimeException | Error failure) {
+      if (vector != null) {
+        runCleanup(failure, vector::close);
+      }
+      if (!array.isClosed()) {
+        runCleanup(failure, array::release);
+        runCleanup(failure, array::close);
+      }
+      throw failure;
+    }
+  }
+
+  private static void runCleanup(Throwable failure, Runnable cleanup) {
+    try {
+      cleanup.run();
+    } catch (Throwable cleanupFailure) {
+      if (cleanupFailure != failure) {
+        failure.addSuppressed(cleanupFailure);
+      }
+    }
   }
 }
--- a/spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
+++ b/spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
@@ -22,6 +22,8 @@ package org.apache.arrow.c;
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.util.AutoCloseables;
 import org.apache.arrow.vector.FieldVector;
+import org.apache.arrow.vector.dictionary.Dictionary;
+import org.apache.arrow.vector.types.pojo.DictionaryEncoding;
 import org.apache.arrow.vector.types.pojo.Field;
 
 /**
@@ -51,9 +53,10 @@ public class ArrowImporter {
 
   public FieldVector importVector(
       ArrowArray array, ArrowSchema schema, CDataDictionaryProvider provider) {
+    Field field = null;
     FieldVector vector = null;
     try {
-      Field field = importField(schema, provider);
+      field = importField(schema, provider);
       vector = field.createVector(allocator);
       ArrayImporter importer = new ArrayImporter(allocator, vector, provider);
       importer.importArray(array);
@@ -62,10 +65,40 @@ public class ArrowImporter {
       if (vector != null) {
         AutoCloseables.close(failure, vector);
       }
+      if (field != null) {
+        closeDictionaries(field, provider, failure);
+      }
+      // ArrayImporter.importArray moves the array into a copy it owns and closes the source
+      // before doImport can fail, so a closed array means the importer already took it and
+      // releasing it again would be a double release.
       if (!array.isClosed()) {
         AutoCloseables.close(failure, array::release, array);
       }
       throw failure;
+    }
+  }
+
+  /**
+   * Closes the dictionary vectors registered in {@code provider} for {@code field} and its
+   * children.
+   *
+   * <p>{@code ArrayImporter.doImport} loads a column's dictionary values before its main data, and
+   * both sets of buffers hold references on the same imported C array. Those values live in the
+   * provider rather than in the column's own vector, so closing the column alone leaves the C array
+   * alive until the provider is closed at the end of the task. Each import assigns a fresh
+   * dictionary id, so no other column can be using the entries this one created.
+   */
+  public static void closeDictionaries(
+      Field field, CDataDictionaryProvider provider, Throwable failure) {
+    DictionaryEncoding encoding = field.getDictionary();
+    if (encoding != null) {
+      Dictionary dictionary = provider.lookup(encoding.getId());
+      if (dictionary != null) {
+        AutoCloseables.close(failure, dictionary.getVector());
+      }
+    }
+    for (Field child : field.getChildren()) {
+      closeDictionaries(child, provider, failure);
     }
   }
 }
--- a/spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
+++ b/spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
@@ -20,6 +20,7 @@
 package org.apache.arrow.c;
 
 import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.util.AutoCloseables;
 import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.types.pojo.Field;
 
@@ -59,23 +60,12 @@ public class ArrowImporter {
       return vector;
     } catch (RuntimeException | Error failure) {
       if (vector != null) {
-        runCleanup(failure, vector::close);
+        AutoCloseables.close(failure, vector);
       }
       if (!array.isClosed()) {
-        runCleanup(failure, array::release);
-        runCleanup(failure, array::close);
+        AutoCloseables.close(failure, array::release, array);
       }
       throw failure;
-    }
-  }
-
-  private static void runCleanup(Throwable failure, Runnable cleanup) {
-    try {
-      cleanup.run();
-    } catch (Throwable cleanupFailure) {
-      if (cleanupFailure != failure) {
-        failure.addSuppressed(cleanupFailure);
-      }
     }
   }
 }
--- a/spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
+++ b/spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
@@ -245,6 +245,10 @@ class NativeUtil extends AutoCloseable {
   /**
    * Imports a list of Arrow addresses from native execution, and return a list of Comet vectors.
    *
+   * On failure this releases everything it was given, both the arrays and schemas it has not
+   * imported yet and the vectors it has already imported, so callers must not add cleanup of
+   * their own.
+   *
    * @param arrays
    *   a list of Arrow array
    * @param schemas
@@ -261,18 +265,23 @@ class NativeUtil extends AutoCloseable {
         val arrowSchema = schemas(i)
         val arrowArray = arrays(i)
 
+        // importField's finally consumes the schema. ArrayImporter takes the array normally, while
+        // ArrowImporter's catch releases it if the import fails before that transfer.
         firstUnconsumed = i + 1
         val arrowVector = importer.importVector(arrowArray, arrowSchema, dictionaryProvider)
-        var cometVector: CometVector = null
-        try {
-          cometVector = CometVector.getVector(arrowVector, dictionaryProvider)
-          arrayVectors += cometVector
-        } catch {
-          case failure: Throwable =>
-            val rollback = if (cometVector == null) arrowVector else cometVector
-            AutoCloseables.close(failure, rollback)
-            throw failure
-        }
+        val cometVector =
+          try CometVector.getVector(arrowVector, dictionaryProvider)
+          catch {
+            case failure: Throwable =>
+              // Nothing took ownership of the column, so release both halves: the vector and the
+              // dictionary values the import left in the provider. Read the field before closing,
+              // so the walk never depends on metadata read back from a closed vector.
+              val field = arrowVector.getField
+              AutoCloseables.close(failure, arrowVector: AutoCloseable)
+              ArrowImporter.closeDictionaries(field, dictionaryProvider, failure)
+              throw failure
+          }
+        arrayVectors += cometVector
       }
       arrayVectors.toSeq
     } catch {
--- a/spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
+++ b/spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
@@ -254,15 +254,29 @@ class NativeUtil {
   def importVector(arrays: Array[ArrowArray], schemas: Array[ArrowSchema]): Seq[CometVector] = {
     val arrayVectors = mutable.ArrayBuffer.empty[CometVector]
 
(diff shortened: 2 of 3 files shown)


spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
```diff
@@ -254,15 +254,29 @@ class NativeUtil {
...
-      val arrowArray = arrays(i)
+    try {
+      (0 until arrays.length).foreach { i =>
+        val arrowSchema = schemas(i)
+        val arrowArray = arrays(i)
 
-      arrayVectors += CometVector.getVector(
-        importer.importVector(arrowArray, arrowSchema, dictionaryProvider),
-        dictionaryProvider)
+        arrayVectors += CometVector.getVector(
```
peterxcli: [P1] Keep the current `FieldVector` rollback-owned until publication

`arrayVectors` only owns earlier columns here. If `importer.importVector` fails after `field.createVector` / `importArray` has attached buffers, or if it succeeds and `CometVector.getVector` throws, the current `FieldVector` is unreachable. The C structs are already consumed, so `releaseArrowStructs` cannot recover it. I reproduced the latter with a valid `UInt4Vector`: allocator close reports 48 leaked bytes.

Please close the raw `FieldVector` on `ArrowImporter.importVector` failure and retain rollback ownership across wrapping and `+=`; commit only after it is appended. Add a failure-after-import regression—the current test fails during schema import before a `FieldVector` exists.

spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
```diff
@@ -254,15 +254,29 @@ class NativeUtil {
...
+          importer.importVector(arrowArray, arrowSchema, dictionaryProvider),
+          dictionaryProvider)
+      }
+      arrayVectors.toSeq
+    } catch {
+      case failure: Throwable =>
+        val firstUnconsumed = arrayVectors.length
+        arrayVectors.foreach { vector =>
```
peterxcli: Non-blocking: Arrow 18.3 already provides `org.apache.arrow.util.AutoCloseables`; since `CometVector` is `AutoCloseable`, this loop can be `AutoCloseables.close(failure, arrayVectors.toSeq: _*)`. It closes every resource and attaches cleanup failures to the original. Caveat: it catches `Exception` rather than every `Throwable`, so keep the manual loop if continuing after a close-time `Error` is intentional. No new Comet helper is needed.

peterxcli (changes requested):
I reproduced one remaining ownership leak on `79d97e1`, so I am requesting changes.

A valid `UInt4Vector` imports successfully, then `CometVector.getVector` rejects its unsigned logical type. The current `FieldVector` has not been appended, while its Arrow structs have already been consumed. A focused Spark 4.1 `NativeUtilSuite` regression leaves 48 bytes allocated; the other five tests pass. The same missing rollback exists when `ArrowImporter.importArray` fails after `field.createVector`.

Please keep the current `FieldVector` rollback-owned until import, wrapping, and publication complete, and add a post-import failure regression. Arrow 18.3's `AutoCloseables` is already available; no new cleanup abstraction is needed.

Non-blocking: `NativeUtil` documents that it must be closed but does not extend `AutoCloseable`; extending it would make that lifecycle contract explicit. The test-local `IntVector` can use `Using.resource`.

CI has not run: the workflow runs stopped as `action_required`. Beyond this ownership gap, I found no other correctness or concurrency issue in this two-file diff.

spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
1fanwang: Thanks!
Done in https://github.com/apache/datafusion-comet/commit/926be0dabc153a0c4fafc2c23e6f8c56dcfbf5e9. The current  FieldVector  stays rollback-owned through import, wrapping, and publication, with regressions for both failure paths.

spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
1fanwang: Done in https://github.com/apache/datafusion-comet/commit/1f098bb49c90c63f21a94adafdaab482f25d5bc6. Cleanup now uses Arrow's  AutoCloseables .  NativeUtil  implements  AutoCloseable , and the test vectors use  Using.resource .

andygrove (commented):
Thanks for taking this on, and thanks for the detailed raw test output in the description. The core fix looks right to me. I walked the Arrow 18.3.0 `ArrayImporter` source and the ownership reasoning holds on every exit I could find. `importArray` moves the caller's `ArrowArray` into an internal `ownedArray` and closes the source before `doImport` can fail, so the `array.isClosed()` guard correctly separates "the importer took it" from "it never got that far". Calling `release()` on a freshly allocated struct whose release callback is NULL is safe, which the EOF path merged in #5531 already depends on.

CI has not run here. The workflow runs are sitting in `action_required` and need a maintainer to approve them. Since this touches every native operator path I ran the relevant pieces locally against `1f098bb4`:

| Check | Result |
| --- | --- |
| `test-compile`, Spark 3.5 / Scala 2.12 | pass |
| `NativeUtilSuite`, Spark 3.5 / Scala 2.12 | 7/7 pass |
| `NativeUtilSuite`, Spark 4.1 / Scala 2.13 | 7/7 pass |
| `spotless:check` | pass |

I also checked the regressions are not vacuous. Reverting just `NativeUtil.scala` and `ArrowImporter.java` to `apache/main` while keeping the new tests gives three failures, all `Memory was leaked by query`, at 176, 288 and 33592 bytes. Each passes with the fix in place. `NativeUtilSuite` is already registered in both `pr_build_linux.yml` and `pr_build_macos.yml`, so no workflow change is needed.

One thing worth recording for whoever touches the harness next. These tests catch the leaks indirectly. `withIsolatedStructAllocator` only isolates the C struct allocations, and the imported vector's buffers come from the global `CometArrowAllocator`. An unreleased import keeps the exported source buffers pinned on the isolated allocator, and that is what trips the assertion. I would not change it, since asserting on the shared global allocator would be flaky. It is just worth a note somewhere so nobody later simplifies the harness and quietly makes the assertions blind.

I have left a few comments inline. The dictionary-encoded case is the one I would most like your read on.

spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
```diff
@@ -50,10 +51,21 @@ Field importField(ArrowSchema schema, CDataDictionaryProvider provider) {
...
+    FieldVector vector = null;
+    try {
+      Field field = importField(schema, provider);
+      vector = field.createVector(allocator);
+      ArrayImporter importer = new ArrayImporter(allocator, vector, provider);
+      importer.importArray(array);
+      return vector;
+    } catch (RuntimeException | Error failure) {
+      if (vector != null) {
+        AutoCloseables.close(failure, vector);
```
andygrove, line 66: For a dictionary-encoded column, `ArrayImporter.doImport` loads the dictionary values into the provider's vector before it loads the main data, and those buffers reference the same `ReferenceCountedArrowArray` as the column itself. If the main data import then fails, closing `vector` here does not drop those references, so the C release callback only fires when `NativeUtil.close()` closes the provider. The same applies when `CometVector.getVector` throws on a dictionary column, since the rollback there is only the indices vector.

Does that match your reading? It is a deferred release rather than a permanent leak, and it is bounded by task lifetime. But the trigger this is aimed at is allocator pressure, which is exactly when holding a native batch until end of task hurts most. Would clearing the failing column's dictionary vector be enough? A test with a dictionary-encoded column would pin the behavior down either way.

spark/src/main/java/org/apache/arrow/c/ArrowImporter.java
```diff
@@ -50,10 +51,21 @@ Field importField(ArrowSchema schema, CDataDictionaryProvider provider) {
...
+      }
+      if (!array.isClosed()) {
```
andygrove, line 74: `isClosed()` is package-private and annotated `@VisibleForTesting` in Arrow. It works here because this class shares Arrow's package, and I do not think there is a better signal available.

Could you add a short comment saying what it distinguishes, namely that `ArrayImporter.importArray` closes the source array before it can fail in `doImport`? Without that, a future Arrow upgrade that drops the method is likely to get fixed by deleting the guard, and that turns this into a double release rather than a leak.

spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
```diff
@@ -253,16 +254,33 @@ class NativeUtil {
...
-
-      arrayVectors += CometVector.getVector(
-        importer.importVector(arrowArray, arrowSchema, dictionaryProvider),
-        dictionaryProvider)
+    try {
+      (0 until arrays.length).foreach { i =>
+        val arrowSchema = schemas(i)
+        val arrowArray = arrays(i)
+
+        firstUnconsumed = i + 1
```
andygrove, line 270: Setting this before the import is the right call, but it is worth a comment saying why. It relies on `importer.importVector` consuming both the schema and the array on every exit, success or failure. The schema goes through `importField`'s `finally` and the array through the new catch in `ArrowImporter`. That is what makes it safe to exclude column `i` from `releaseArrowStructs`.

Right now that invariant lives entirely in `ArrowImporter` and there is nothing here pointing at it. Getting it wrong in either direction gives you a leak or a double release.

spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
```diff
@@ -253,16 +254,33 @@ class NativeUtil {
    */
   def importVector(arrays: Array[ArrowArray], schemas: Array[ArrowSchema]): Seq[CometVector] = {
```
andygrove, line 259: Could you extend the scaladoc to say that on failure this method releases everything it was given? `getNextBatch` does its own cleanup on the other two exits, so the next person reading it could reasonably wrap this call in cleanup too and cause a double release.

spark/src/main/scala/org/apache/comet/vector/NativeUtil.scala
```diff
@@ -253,16 +254,33 @@ class NativeUtil {
...
+
+        firstUnconsumed = i + 1
+        val arrowVector = importer.importVector(arrowArray, arrowSchema, dictionaryProvider)
+        var cometVector: CometVector = null
+        try {
+          cometVector = CometVector.getVector(arrowVector, dictionaryProvider)
+          arrayVectors += cometVector
+        } catch {
+          case failure: Throwable =>
+            val rollback = if (cometVector == null) arrowVector else cometVector
```
andygrove: The `var` plus null check plus `rollback` makes the reader reason about a case that cannot happen, since `arrayVectors += cometVector` will not throw. This is equivalent and drops both:

```scala
val cometVector =
  try CometVector.getVector(arrowVector, dictionaryProvider)
  catch {
    case failure: Throwable =>
      AutoCloseables.close(failure, arrowVector)
      throw failure
  }
arrayVectors += cometVector
```

In code whose whole job is memory ownership, the fewer branches a reader has to hold in their head the better.

spark/src/test/scala/org/apache/comet/vector/NativeUtilSuite.scala
```diff
@@ -99,6 +103,99 @@ class NativeUtilSuite extends CometTestBase {
...
+      stringStruct.addOrGet("first", intType, classOf[IntVector])
+      stringStruct.addOrGet("second", stringType, classOf[VarCharVector])
+      stringStruct.allocateNew()
+      stringStruct.setValueCount(1)
+
+      val (arrays, schemas) = nativeUtil.allocateArrowStructs(2)
+      Data.exportVector(allocator, intStruct, null, arrays(0), schemas(0))
+      Data.exportVector(allocator, stringStruct, null, arrays(1), schemas(1))
+      intStruct.close()
+      stringStruct.close()
```
andygrove: `intStruct` and `stringStruct` are closed unconditionally before the `try`, so if either `Data.exportVector` throws they leak and the failure surfaces as an allocator complaint rather than the real error. You use `Using.resource` in the two tests above. Could this one do the same?

peterxcli (approved):
lgtm now, thanks for the update!

sunchao (approved):

sunchao (approved):

sunchao:
Merged, thanks @1fanwang for the contribution and @andygrove @peterxcli for the review!

One nit in the above comments

> These tests catch the leaks indirectly. `withIsolatedStructAllocator` only isolates the C struct allocations, and the imported vector's buffers come from the global `CometArrowAllocator`.

Might worth adding a short code comment to explain `withIsolatedStructAllocator`.

Issue #5487, opened 2026-08-26
Adopt techniques from Spark's ArrowCachedBatchSerializer (SPARK-57268) in Comet's cache format


andygrove:
### What is the problem the feature request solves?

Spark added its own Arrow in-memory cache format in [SPARK-57268](https://github.com/apache/spark/pull/56334) (`org.apache.spark.sql.execution.columnar.ArrowCachedBatchSerializer`), solving the same problem as the serializer added in #5051. It is present in `branch-4.3` and `master` only, so it is not available in any Spark version Comet supports today (3.4 through 4.2) and does not remove the need for Comet's own serializer. It is, however, a more mature implementation of the same idea, and several of its decisions are worth adopting.

Recorded here so the comparison is not lost. Raised on #5051 by @sunchao, who pointed at the Spark work; @viirya wrote it.

Where the two implementations already agree, for the record: empty projections emit row counts without touching the payload, statistics use the same five-field `SimpleMetricsCachedBatch` layout, the payload is schema-agnostic with respect to session timezone, cleanup on early termination goes through a `TaskCompletionListener`, and the format is opt-in via `spark.sql.cache.serializer` with the default unchanged. Spark reached the empty-projection optimization in [SPARK-58390](https://github.com/apache/spark/commit/073c8d8f4), a follow-up eight days after the main PR, which is the same order we found it in.

### Describe the potential solution

Roughly in order of value for effort.

- [ ] **Drop the schema message from each cached stream.** Spark's `ArrowCachedBatch` deliberately stores an encapsulated Arrow RecordBatch message with no Schema message and no end-of-stream marker, reconstructing the schema from the relation's attributes on read, explicitly to avoid repeating the schema bytes in every cached batch. Comet writes a full `ArrowStreamWriter` stream per column per batch, so a 60-column relation cached in 500 batches writes about 30,000 schema messages. That is a direct contributor to the framing overhead measured in #5484 (2.5% more footprint at 6 columns, 32% at 60). Rough arithmetic puts the schema bytes at about a third of that. Contained change: `MessageSerializer.serialize` and `deserializeRecordBatch` plus `VectorLoader`, instead of `ArrowStreamWriter` and `ArrowStreamReader`.

[...]

Issue #5485, opened 2026-08-26
Cached reads that feed Spark operators are slower than Spark's own cache format


andygrove:
### What is the problem the feature request solves?

With Comet's cache serializer installed (`spark.comet.exec.inMemoryCache.enabled=true`), a query that reads a cached relation through Spark operators rather than Comet ones is slower than the same query against Spark's `DefaultCachedBatch`, by roughly 1.7x to 2.5x depending on projection width.

Measured on 5M rows and 6 columns, min of 5 runs after a discarded warm-up, with Comet execution disabled so the cached data feeds Spark operators:

| Read shape | Spark cache | Comet cache |
|------------|------------:|------------:|
| `count(*)` | 54 ms | 190 ms |
| 1 of 6 columns | 103 ms | 260 ms |
| 3 of 6 columns | 308 ms | 542 ms |
| 6 of 6 columns | 358 ms | 595 ms |

This is what remains after #5484, which made decode proportional to the projection. With Comet execution enabled the same reads are now 1.7x to 1.8x *faster* than Spark's cache, so the gap is specific to Spark consuming the cached data.

It matters because the format is chosen when the cache is materialized and cannot be changed afterwards. A pipeline with many fallbacks to Spark operators pays this on every cached read, which is the shape reported in https://github.com/apache/datafusion-comet/pull/5051#issuecomment-5208500029.

### Describe the potential solution

**Diagnose before choosing a fix.** The cause is not yet established. `supportsColumnarOutput` returns true for any Comet-writable schema, so Spark's `InMemoryTableScanExec` takes a columnar path over `CometVector`s and converts to rows above it, where `DefaultCachedBatch` feeds `OnHeapColumnVector`s to the same conversion. Whether the cost is that conversion, the Arrow decode itself, or the loss of Spark's per-column decompression is a guess at this point and should be measured first. `supportsColumnarOutput` returning false for this case, sending the read through `convertCachedBatchToInternalRow` instead, is worth measuring as part of that.

**Detecting the losing case at write time is not sound**, and is worth writing down so it is not re-proposed. The payload format is fixed when the relation materializes, and the readers do not exist yet: one cached relation serves arbitrarily many later queries with different plan shapes, so there is no single correct answer for it. The serializer's inputs are the schema and the input RDD, neither of which carries consumer information.

Three variants that are workable, in descending order of soundness:

[...]

PR #5543, opened 2026-08-29
perf: project cached batches by buffer selection, prune on collated strings

andygrove:

Part of #5487. Ticks these boxes:

- **Drop the schema message from each cached stream** — subsumed by the buffer-selection change below, which removes per-column framing entirely.
- **Evaluate projection by buffer selection instead of per-column streams**
- **Prune on collated string columns**
- **Documentation** — the committed-benchmark-results half is not done; see the note at the end.

Not addressed here: typed readers for the row path (#5485 has no established cause yet, so this would be optimising ahead of a diagnosis) and background prefetch.

## Rationale for this change

#5051 stores each cached column as its own compressed Arrow IPC stream, so a scan decodes only the columns it projects. That works, but it pays an Arrow schema block and compression framing per column per batch, and gives up cross-column compression: footprint grows 2.5% at 6 columns and 32% at 60.

Spark's `ArrowCachedBatchSerializer` (SPARK-57268) reaches the same projection-proportional decode with no per-column framing at all. It keeps one RecordBatch per cached batch and parses the IPC message flatbuffer, which lists every buffer's offset and length within the body, to copy out only the byte ranges belonging to the selected columns. It depends on Arrow's native per-buffer compression rather than wrapping the whole payload in a Spark `CompressionCodec`.

That approach dominates the current design on footprint while keeping the projection win, so this PR adopts it.

Separately, `tracksBounds` matches `case StringType`, which a collated `StringType` does not equal. Collated columns therefore get null bounds and `buildFilter` declines to push predicates on them. That is correct but loses pruning that Spark manages.

**Cached batch format.** `CometCachedBatch.columns: Array[ChunkedByteBuffer]` becomes `bytes: Array[Byte]`: one encapsulated Arrow IPC record batch message and its body, with no Schema message and no end-of-stream marker. The reader rebuilds the schema from the cached relation's attributes via `Utils.toArrowSchema`, so a wide relation no longer repeats the same schema bytes once per cached batch. New `CachedBatchIpc` owns the format — the field-node, buffer-span and variadic-count arithmetic, and the projected read that copies only the selected columns' buffers into a single off-heap allocation.

**Compression** moves from a whole-payload Spark codec to Arrow's per-buffer IPC compression, which is what lets a projected read decompress only what it selected. New `spark.comet.exec.inMemoryCache.compression.codec` (`zstd` default, `none` available) and `...compression.zstd.level`.

Arrow's lz4 is deliberately not offered. It is commons-compress's pure-Java implementation, unrelated to the JNI-accelerated lz4-java behind `spark.io.compression.codec`. Over a 200k-row six-column relation:

| Codec | Materialize | Footprint | Read 1 of 6 | Read 6 of 6 |
|-------|------------:|----------:|------------:|------------:|
| `zstd` | 363 ms | 2 MiB | 56 ms | 62 ms |
| `none` | 1776 ms | 13 MiB | 78 ms | 81 ms |
| ~~`lz4`~~ | 205373 ms | 6 MiB | 51 ms | 107 ms |

lz4 is dominated on both speed and size, so nothing prefers it. zstd also beats storing batches uncompressed on both axes, because the bytes it saves cost more to copy and store than compressing them costs. Reads still accept any codec a batch records.

**Dictionary-encoded columns are decoded before being stored.** A payload with no schema message has nowhere to record either the index type or the dictionary. Comet's native scans do produce such columns, so this is a real path rather than a defensive one.

**Decompression is done in `CachedBatchIpc` rather than left to `VectorLoader`.** arrow-java 18.3.0 leaks on the failure path: `VectorLoader.loadBuffers` collects a field's decompressed buffers into a local list and releases them only after the whole field has loaded, so a buffer that fails to decompress strands every buffer of that field decompressed before it. A string column reaches this — its offsets buffer decompresses, then its data buffer throws — so a single corrupt cached batch leaks off-heap for the life of the executor.

**Collated string pruning.** `tracksBounds` widens to any `StringType`, and bounds are compared with `TypeUtils.getInterpretedOrdering(dataType)` — Spark's own interpreted ordering for the type, which on Spark 4 resolves a `StringType` through `CollationFactory.fetchCollation(collationId).comparator`. So bounds are recorded with the same ordering the partition filter Spark generates over that column uses, the collation awareness comes from Spark at every supported version with no shim, and the ordering is resolved once per column instead of re-dispatching on the `DataType` per row. This also replaces the hand-rolled per-type `compare`.

**Reading** builds one `CachedBatchIpc.Projection` per partition, holding the projected schema and the node/buffer/variadic index layout. That arithmetic walks every field of the cached relation, so computing it per batch would make the bookkeeping O(total columns) against O(selected columns) of useful work — worst in exactly the wide-relation, narrow-projection case this format exists for.

**Dependency.** Adds `org.apache.arrow:arrow-compression`. Already covered by the existing `org.apache.arrow:*` shade include, so it relocates with the rest of Arrow; its `commons-compress` and `zstd-jni` are excluded and come from Spark, which ships both on every supported version.

`CometInMemoryCacheSuite` keeps its existing coverage, with the format-dependent tests rewritten against the new layout:

- The projection tests no longer corrupt per-column streams. They scramble the compressed bytes of the columns a read must not touch, leaving every other byte of the payload identical, and assert the read still succeeds — which it only can if those buffers were never copied out of the payload. Each asserts as a precondition that the column it corrupts is genuinely stored compressed, since Arrow stores a buffer verbatim when compressing would not shrink it.
- A new test asserts the payload begins with its record batch rather than a schema message, so a regression to a self-describing stream shows up directly rather than only as a footprint number.
- Per-column sizes in the statistics row are checked against the message's own buffer layout.
- A new test round-trips every codec the config accepts through a full read, a projected read, a row-count-only read and stats pruning. `none` takes a different path on read and was broken until this test was written.
- The collation test is inverted: it now asserts pruning happens, and adds a UTF8_LCASE case that returns no rows if bounds are recorded with byte-order comparison. The no-bounds case moves to `BinaryType`.
- Two leak tests cover a failure at a column's first buffer and a failure part way through a column, the latter reaching the `VectorLoader` behaviour above. Both assert the decode error surfaces as itself rather than as an Arrow reference-count error, which is what catches a cleanup path releasing the shared body twice.
- The dictionary tests assert values read back correctly, which is what proves the writer decoded them — a row count would not.

`CometCachedBatchHelper` re-derives the IPC buffer arithmetic independently rather than calling into `CachedBatchIpc`, so the assertions built on it cannot pass by inheriting a bug from the code under test.

Nested columns get their own projection coverage, over a six-column relation whose middle four are a struct, an array, a map and a struct wrapping an array. They were previously round-tripped only under `SELECT *`, which cannot exercise the span arithmetic: a flat column always owns one field node and two or three buffers, a nested one owns a run as long as its subtree, and a full projection covers the whole buffer sequence however it is partitioned. So one test gives each column a turn as the sole projection with the other five corrupted, and another compares values against the uncached query across single-column, paired and out-of-order projections — a row count comes from the record batch header, so it cannot catch a window that is misaligned but still decompresses. The per-column statistics test runs over the nested relation too. Both new tests fail if `fieldNodeCount` stops recursing into children.

Also run: `CometInMemoryCacheKryoSuite`, `CometExecSuite`, `UtilsSuite`. Compiles clean against Spark 3.5, 4.0 and 4.1. The shaded jar was checked to confirm arrow-compression relocates and that commons-compress and zstd-jni are not bundled.

`CometInMemoryCacheBenchmark` (Apple M3 Max, JDK 17, Spark 4.1, release build), over a 5M-row relation of six flat columns:

| Query shape | Spark cache scan + convert | `CometInMemoryTableScan` | Relative |
|-------------|---------------------------:|-------------------------:|---------:|
| Repeated scan (3 of 6 columns) | 201 ms | 167 ms | 1.2x |
| Selective filter | 69 ms | 61 ms | 1.1x |
| Row count only (0 of 6) | 45 ms | 47 ms | 1.0x |
| Narrow projection (1 of 6) | 70 ms | 57 ms | 1.2x |
| Full projection (6 of 6) | 556 ms | 290 ms | 1.9x |

And over a 1M-row relation of six columns whose middle three are structs, one nested two levels deep:

| Query shape | Spark cache scan + convert | `CometInMemoryTableScan` | Relative |
|-------------|---------------------------:|-------------------------:|---------:|
| Row count only (0 of 6) | 39 ms | 35 ms | 1.1x |
| Narrow projection (1 of 6) | 109 ms | 61 ms | 1.8x |
| Full projection (6 of 6) | 282 ms | 126 ms | 2.2x |

`verifyPlan` now asserts the projection width each case claims, which is what corrected the flat table above: the previous `full projection (6 of 6 columns)` row was reading three columns, because `count()` over a non-nullable column is rewritten to `count(1)` by `NullPropagation` and the column is then pruned out of the scan. Only `k`, `s1` and `s2` were nullable, and only incidentally, because `Remainder` can divide by zero. Every column of both relations is nullable now so `count(c)` genuinely reads `c`. A real six-column read is 1.9x, not the 2.2x previously published for a three-column one.

Arrays and maps are deliberately not in the benchmark. The baseline arm needs Spark's cache scan to bridge into Comet operators, and `CometSparkToColumnarExec` declines `ArrayType` and `MapType`, so for a query projecting one of those that arm does not exist and the two cases stop measuring the same boundary. They are covered in the suite instead.

As with #5051, both columns read the same Comet-written `CometCachedBatch` and Comet execution is on in both, so this measures keeping the cached scan native against falling back to a Spark cache scan and converting — not Comet against Spark execution, and not a comparison with Spark's own cache format.

## Notes for reviewers

- **The feature is still disabled by default.** `spark.comet.exec.inMemoryCache.enabled` stays `false`, as it has been since #5051, so none of this reaches a default configuration. Everything below the config, including the change of cached format, only affects users who have opted in to the experimental native cache. The config is static: its value at startup is what decides whether Comet's cache serializer is installed at all.
- **The committed benchmark results half of the documentation item is not done.** `spark/benchmarks` is in `.gitignore`, so Comet does not currently commit results files the way Spark does. Doing it properly needs that directory un-ignored plus a workflow to regenerate them, or the numbers rot — worth its own decision rather than being slipped in here. The measured tables live in the new docs page instead.
- Cache materialization retains about 800 bytes per partition, independent of row count. This reproduces identically on 492dd6f2a, so it predates this PR and is untouched here; happy to file it separately.


comphead (commented):
Thanks @andygrove should we also test/benchmark nested data?

andygrove:
Good call — this turned out to be worth doing, because the nested coverage that existed was in the one shape that can't fail.

Nested columns were already round-tripped, but only under `SELECT *`. That can't exercise the part of this format that is actually nontrivial for them: a flat column always owns one field node and two or three buffers, whereas a nested one owns a run as long as its whole subtree, and a full projection covers the entire buffer sequence however it happens to be partitioned. So the span arithmetic was only ever tested where getting it wrong doesn't show.

There are now two tests over a six-column relation whose middle four are a struct, an array, a map and a struct wrapping an array. The first gives each column a turn as the sole projection with the other five corrupted, so a run computed short or long by a buffer gets caught reaching into a corrupted neighbour. The second compares values against the uncached query across single-column, paired and out-of-order projections — row counts come from the record batch header, so they can't catch a window that is misaligned but still decompresses, and an out-of-order projection is the case a full one genuinely cannot stand in for. The per-column statistics test now runs over the nested relation too, since a nested column's recorded size is the sum of its subtree. Both new tests fail if `fieldNodeCount` stops recursing into children, which is how I checked they aren't passing for free.

The format itself needed no changes, so this is coverage rather than a fix.

On the benchmark, I added the three projection widths over a relation of struct columns:

| Query shape | Spark cache scan + convert | `CometInMemoryTableScan` | Relative |
|-------------|---------------------------:|-------------------------:|---------:|
| Row count only (0 of 6) | 39 ms | 35 ms | 1.1x |
| Narrow projection (1 of 6) | 109 ms | 61 ms | 1.8x |
| Full projection (6 of 6) | 282 ms | 126 ms | 2.2x |

The gap is wider than the flat relation's at every width, which is what you'd expect — the conversion the left column pays scales with values per row, not with columns.

I left arrays and maps out of the benchmark on purpose, and said so in the docs. The left column needs Spark's cache scan to bridge into Comet operators, and `CometSparkToColumnarExec` declines `ArrayType` and `MapType` outright, so for a query projecting one of those that arm doesn't exist — the partial aggregate stays on Spark and the two cases stop measuring the same boundary. Tests are the right place for those, and they're covered there.

One thing your question shook out that I should flag: while adding the nested cases I made `verifyPlan` assert the projection width each case claims, and the existing `full projection (6 of 6 columns)` case was reading three. `count()` over a non-nullable column gets rewritten to `count(1)` by `NullPropagation`, which then prunes the column out of the scan, and only `k`, `s1` and `s2` were nullable — incidentally at that, because `Remainder` can divide by zero. Every column of both relations is nullable now so `count(c)` really reads `c`, and I've regenerated the numbers in the docs. The genuine 6-of-6 read is 1.9x rather than the 2.2x that was published for what was really a 3-column read.

comphead (commented):
Reviewed for Spark semantics, the Arrow/JVM boundary, memory ownership, and scope.

The format change reads as sound. I checked the reference counting in `load` and `decompressed` against arrow-java 18.3.0: the 6-arg `ArrowRecordBatch` constructor is `(.., variadicBufferCounts, alignBuffers)` with `retainBuffers` defaulting to true, `extractUncompressedBuffer` returns a non-retained slice without closing its input, and `ZstdCompressionCodec.doDecompress` closes its allocation on both error paths. The retain-before-decompress ordering is correct for all four cases (compressed, stored-verbatim, zero-length, no codec). The collated-string change also matches Spark: `TypeUtils.getInterpretedOrdering` resolves through `CollationFactory` on 4.x, which is the same ordering the generated partition filter uses, so bounds and predicate agree with no shim.

One finding worth acting on before merge. The rest are cleanups.

**Major.** The read path derives every buffer window from `Utils.toArrowSchema(cacheAttributes)`, but nothing checks the writer produced that layout, and `isArrowBacked` admits at least one vector whose buffer count differs. Details inline on `CachedBatchIpc.scala`. The fix is a length `require` in `load` using prefix sums `selectedRange` already computes.

**Scope.** Appropriately scoped. The `getInterpretedOrdering` switch replaces a hand-rolled per-type comparison rather than adding one, so it is not a drive-by.

**Tests.** The corruption-based projection tests are the right assertion and do distinguish the new behaviour from the old. Two gaps: no coverage of a `BinaryType` column backed by a fixed-size vector (see the Major), and the new tests cannot move to `sql-tests/` because the serializer comes from a suite-level `SparkConf` and `spark.sql.cache.serializer` is static.

**Performance.** Benchmark evidence is present, and the `verifyPlan` projection-width assertion is a real improvement over the previously published numbers.

Two nits not worth their own thread. `Projection(arrowFields: Seq[Field], ...)` is indexed positionally and both call sites already pass `.toIndexedSeq`, so `IndexedSeq[Field]` would make that a compile-time guarantee. `hydrateDictionaries` reads oddly beside a scaladoc and a PR description that both say "decode".

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,456 @@
...
+   */
+  final class Projection(arrowFields: Seq[Field], selectedIndices: Array[Int]) {
+
+    private val schema = new Schema(selectedIndices.map(arrowFields).toSeq.asJava)
+
+    // A record batch body is a flat, depth-first sequence of buffers in schema order, so each
+    // top-level column owns a contiguous run of it; field nodes and variadic buffer counts run in
+    // the same order.
+    private val nodeIndices = selectedRange(arrowFields, selectedIndices, fieldNodeCount)
+    private val bufferIndices = selectedRange(arrowFields, selectedIndices, fieldBufferCount)
```
comphead: **Major: the read path trusts an unchecked layout invariant.**

`Projection` derives every node and buffer window from `Utils.toArrowSchema(cacheAttributes)`, but nothing verifies the writer produced that layout. `encodeBatches` takes the `Utils.isArrowBacked(batch)` fast path without converting, and `Utils.isSupportedFieldVector` admits `FixedSizeBinaryVector`, which `Utils.fromArrowType` maps to Spark `BinaryType` while `Utils.toArrowType(BinaryType)` is `ArrowType.Binary`. `TypeLayout.getTypeBufferCount` is 2 for the former and 3 for the latter.

So a cached `BinaryType` column backed by a fixed-size vector (mapInArrow output, an Iceberg `fixed[N]` read) is stored one buffer short. Every buffer index from that column onward shifts, and `batch.buffers(j)` is an unchecked flatbuffer accessor, so the result is wrong values or an `ArrayIndexOutOfBoundsException` inside `setBytes` rather than an error naming the cause. The previous per-column streams were self-describing, so they could not drift this way.

`selectedRange` already computes the full prefix sum. Keeping `starts.last` for each of the three sequences and checking it in `load` is O(1) per batch:

```scala
require(
  batch.nodesLength() == totalNodes && batch.buffersLength() == totalBuffers,
  s"cached batch layout does not match the cached schema: ...")
```

Dropping `FixedSizeBinaryVector` from the `isArrowBacked` fast path, the way `LargeVarCharVector` already is, would close the known case too.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,456 @@
...
+  private def fieldNodeCount(field: Field): Int =
+    1 + field.getChildren.asScala.map(fieldNodeCount).sum
+
+  /**
+   * Number of variadic buffer counts a field contributes, one per view-type buffer, recursively.
+   *
+   * Only Utf8View and BinaryView carry one. Comet's cache never writes view vectors today, but
+   * the span arithmetic above has to stay correct if that changes.
+   */
+  private def fieldVariadicCount(field: Field): Int = {
```
comphead: `fieldVariadicCount` has no effect today, and the stated reason for keeping it ("the span arithmetic above has to stay correct if that changes") does not hold. `fieldBufferCount` goes through `TypeLayout.getTypeBufferCount`, which returns 2 for `Utf8View` and `BinaryView` and ignores the variadic data buffers entirely, so if view vectors ever reach here the buffer spans misalign whether or not the variadic counts are tracked.

Suggest dropping this and the `variadicCounts` plumbing in `load`, and letting the length check above fail loudly instead.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,456 @@
...
+  }
+
+  /**
+   * The on-body compressed size of each top-level column.
+   *
+   * Each column owns the run of buffers its subtree occupies, so its stored size is the sum of
+   * those buffers' recorded lengths. With one payload per batch these are the only per-column
+   * sizes available -- there is no separate stream to measure -- and they are exact.
+   */
+  private def columnSizes(fields: Seq[Field], recordBatch: ArrowRecordBatch): Array[Long] = {
```
comphead, line 523: `fieldBufferCount(fields(i))` re-walks the column's subtree even though `scanLeft` already accumulated it. `starts(i) until starts(i + 1)` is the same range. This runs per batch on the write path, and the `.map(...).sum` allocates an intermediate per column, so a `while` over `buffers` between the two bounds does the whole thing in one pass.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,456 @@
...
+ * A cached batch is one encapsulated Arrow IPC RecordBatch message followed by its body, with no
+ * Schema message and no end-of-stream marker. The schema is not stored because the reader already
+ * has it: `InMemoryRelation` knows the cached relation's attributes, and `Utils.toArrowSchema`
+ * maps them to exactly the fields the writer unloaded. Leaving it out saves a schema message per
+ * cached batch, which for a wide relation cached in many batches is a large share of the payload
+ * that is not data.
+ *
+ * Compression is applied by Arrow per buffer rather than by wrapping the whole payload in a Spark
+ * `CompressionCodec`. That is what makes projection cheap: the message metadata records every
+ * buffer's offset and length within the body, so [[readProjected]] can copy out only the buffers
```
comphead: `readProjected` does not exist. The method is `Projection.load`, so the `[[readProjected]]` link will not resolve. Same name appears in the `alignBuffers` comment at line 131 and in the `CometCachedBatch` scaladoc in `ArrowCachedBatchSerializer.scala`.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,456 @@
...
+        val fields = vectors.map(_.getField)
+        // Serializing consumes the batch, as it does in Utils.serializeBatches. The record batch
+        // holds its own buffers by now -- compressed copies, or retained references when the codec
+        // is none -- so releasing the vectors here does not touch it. getField still answers
+        // afterwards: clearing releases buffers, not the schema.
+        //
+        // Not load bearing for memory: the plan that produced the batch releases its vectors
+        // either way, and dropping this line leaks nothing. It is here because serializeBatches
+        // does the same, so both writers leave a batch they were handed in the same state.
+        root.clear()
```
comphead, line 225: Five lines of comment to say the line does nothing ("dropping this line leaks nothing"). Either drop both, or keep one line noting it matches what `serializeBatches` leaves behind.

Same pattern elsewhere in this file. `decompressed`, `Projection` and `load` each carry multi-paragraph rationale where the invariant is a sentence. The reference-counting notes earn their space. The reconstructions of what the code does do not.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,456 @@
...
+  // per batch on every scan, and the enum lookup walks the CodecType values each time.
+  private val readCodecs: Map[CompressionUtil.CodecType, CompressionCodec] =
+    CompressionUtil.CodecType
+      .values()
+      .filter(_ != CompressionUtil.CodecType.NO_COMPRESSION)
+      .map(t => t -> CommonsCompressionFactory.INSTANCE.createCodec(t))
+      .toMap
+
+  /** The decompressor for a body-compression byte, or None when the batch is stored plain. */
+  private def readCodec(compressionType: Byte): Option[CompressionCodec] =
```
comphead, line 107: Question: `CompressionUtil.CodecType.fromCompressionType` falls back to `NO_COMPRESSION` for an unrecognized byte rather than throwing, so this returns `None` and the payload is read as plain bytes. Is that intended? A batch written by a codec this build does not know would decode to garbage rather than fail.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowCachedBatchSerializer.scala
```diff
@@ -462,24 +509,36 @@ class ArrowCachedBatchSerializer extends SimpleMetricsCachedBatchSerializer {
...
 
     input.mapPartitions { it =>
-      // A ColumnReaders closes its readers (releasing the vectors they are holding) only when the
-      // batch it produced has been consumed. A consumer that stops early -- LIMIT, take(), or a
-      // cancelled task -- leaves the readers for the batch in flight open, so close them on task
-      // completion. Spark's own ArrowCachedBatchSerializer registers a listener for the same
-      // reason.
+      // Built once per partition: resolving the Arrow schema and the projection's buffer layout
+      // walks every field of the cached relation, which would otherwise be paid per batch.
+      val projection = new CachedBatchIpc.Projection(
```
comphead: For a `SELECT count(*)` read, `indices` is empty and `projection` is never used, but a wide relation still pays `toArrowSchema` plus three full field-tree walks per partition. Hoisting the `indices.isEmpty` branch above this, or making `projection` a `lazy val`, keeps the cheapest read cheap. Worth it given this is exactly the wide-relation case the format targets.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowCachedBatchSerializer.scala
```diff
@@ -339,34 +370,47 @@ class ArrowCachedBatchSerializer extends SimpleMetricsCachedBatchSerializer {
...
   // why. Batches that are not get copied into Arrow first, since Utils.serializeBatches only
   // writes CometVector columns.
+  /**
+   * The configured write codec, read on the driver.
+   *
+   * Both write paths resolve this here rather than inside their `mapPartitions` closure: the
+   * closure ships to the executors, where `CometConf` would resolve against whatever `SQLConf`
+   * happens to be current on that thread rather than against this session's.
+   */
+  private def codecSettings(conf: SQLConf): (String, Int) =
```
comphead: A bare `(String, Int)` read back as `codecSetting._1` and `._2` in `encodeBatches`. Two parameters, or a small case class, costs nothing and is equally serializable into the closure.

spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala
```diff
@@ -1315,70 +1394,241 @@ class CometInMemoryCacheSuite extends CometTestBase {
...
       }
     }
   }
 
+  test("Comet in-memory cache decodes only the projected columns of a nested relation") {
+    // The flat case above pins one column and corrupts the rest. Here every column takes its turn,
+    // because a nested column's run of buffers is as long as its subtree rather than a fixed two or
+    // three: a run computed short or long shifts every column after it, so which column is selected
+    // decides whether the misalignment reaches into a corrupted neighbour.
+    nestedProjectionColumns.indices.foreach { selectedIdx =>
```
comphead: `withNestedProjectionCache` is inside the loop, so this materializes the whole relation six times. Caching once and restoring each batch's `bytes` from a `clone()` between iterations does the same work in a sixth of the time.

Also, the flat "decodes only the projected columns" above is the `selectedIdx = 1` special case of this loop. One per-column loop parameterized over both relations would drop a test without dropping coverage.

spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala
```diff
@@ -1584,52 +1838,64 @@ class CometInMemoryCacheSuite extends CometTestBase {
...
     }
   }
 
-  test("Comet in-memory cache broadcasts a batch whose columns have separate dictionaries") {
-    // A broadcast of a cache scan re-serializes each decoded batch as one stream covering every
-    // column, and the writer resolves all of their dictionary IDs against the single provider it
-    // is handed. The columns were decoded from separate streams, so they arrive carrying separate
-    // providers: passing any one of them cannot resolve the others.
-    withDictionaryCache { (relation, batches) =>
+  test("Comet in-memory cache releases its vectors when a column fails after a partial decode") {
```
comphead: This and "releases its vectors when a column fails to decode" assert the same thing at two corruption points, and the comment here says this is "the one that actually catches a leak". Worth merging into one test with two corruption modes rather than two fixtures and two allocator snapshots.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: You're right, and the `FixedSizeBinaryVector` case is real: `toArrowType(BinaryType)` is `Binary` at three buffers against that vector's two, so every buffer index from such a column on shifts.

The length check is in, built the way you suggested. `selectedRange` now returns `starts.last` alongside the indices, and `load` compares both totals against `nodesLength()` and `buffersLength()` before it touches `batch.buffers(j)`. I mutated the check away to confirm it earns its place: the new test for it then passes with no exception at all, which is exactly the silent-wrong-answer failure you described.

I went wider than dropping `FixedSizeBinaryVector` from `isArrowBacked`, because that only closes the top-level case. `isArrowBacked` answers for the top-level vector and never looks at children, so a struct whose child is a `LargeVarCharVector` passes it today and is stored with 64-bit offsets and read back with 32-bit. The length check can't catch that one either, since `LargeUtf8` and `Utf8` are both three buffers. So the write path now asks the direct question: do this batch's vectors already carry the Arrow types the reader will rebuild, recursively? That's `CachedBatchIpc.matchesReaderLayout`, and a batch that disagrees takes the conversion path it was already taking for non-Arrow input rather than being written unreadable. Names, nullability and a timestamp's timezone are excluded from the comparison — the last because a Comet scan labels with the session zone where the reader rebuilds UTC, and that's a label rather than a layout, so comparing it would send every timestamp column down the conversion path for nothing.

One wrinkle: Arrow-Java puts the index type on a dictionary-encoded vector's own field, so the field to compare is the dictionary's, resolved through the same `Utils.lookupDictionary` the writer uses.

On coverage, a unit test builds a `FixedSizeBinaryVector`-backed batch and asserts `isArrowBacked` accepts it while the write path declines it, with a `VarBinaryVector` beside it as the control so the predicate can't pass by refusing everything. The reader's check gets its own test that hands the reader one more attribute than the writer stored.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Agreed, dropped — both `fieldVariadicCount` and the `variadicCounts` plumbing in `load`. Your reasoning holds: `getTypeBufferCount` is 2 for `Utf8View` and ignores the data buffers entirely, so the spans misalign whether or not the counts are tracked. With the length check in place a view vector now fails there loudly instead, which is the right outcome given nothing on the write path can produce one.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Done — a single `while` over `buffers` between `starts(i)` and `starts(i + 1)`, so no re-walk and no intermediate collection per column.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Fixed, all three: the class scaladoc and the `alignBuffers` comment now say `Projection.load`, and so does the `CometCachedBatch` scaladoc in `ArrowCachedBatchSerializer.scala`.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Fair. That one is three lines now and says only what it's for. I also cut the first paragraph of `load`'s doc, which restated the class doc, and tightened `Projection`'s. I left `decompressed` alone — both of its paragraphs are the reference-counting rationale you said earns its space.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Not intended, no. `fromCompressionType` falls through to `NO_COMPRESSION` for anything outside its enum, so an unrecognized byte read the body as plain bytes and returned garbage rather than failing. `readCodec` now checks `NoCompressionCodec.COMPRESSION_TYPE` directly for the genuine uncompressed case and throws for anything else.

The realistic way to reach it is a corrupt payload rather than a writer from a newer build, since a cache never outlives the application that wrote it. Either way it's the same class of thing as the Major above: better to fail than to answer wrongly.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowCachedBatchSerializer.scala
andygrove: Made it a `lazy val`. The `indices.isEmpty` branch below already short-circuits the decode, so that alone is enough to keep a `count(*)` read from paying `toArrowSchema` plus the three field-tree walks.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowCachedBatchSerializer.scala
andygrove: Replaced with a `CacheCodecSettings(name, zstdLevel)` case class, with the "resolved on the driver" note moved onto it. Doing that shook out something else: the comment block above `codecSettings` actually describes `encodeBatches` and had been orphaned when that scaladoc went in, so I've moved it back and corrected its last sentence, which still named `Utils.serializeBatches` as the writer.

spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala
andygrove: Both done. The fixture is out of the loop, with each batch's payload snapshotted once up front and restored between turns — the nested case runs in 213 ms where it was materializing the relation six times. And the flat test is folded into the same loop, so there's now one test parameterized over both relations rather than a general case and its special case.

spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala
andygrove: Merged — one test, two corruption points, one fixture and one allocator snapshot per point. The commentary about what each point catches moved into the test's own comment.

andygrove:
Thanks @comphead — the Major is real and the fix is in, along with everything else. Replies are on the individual threads; the two nits from your summary are done too (`Projection` takes an `IndexedSeq[Field]`, and `hydrateDictionaries` is now `decodeDictionaries`).

Two things worth pulling up here rather than leaving buried in a thread.

The first is that I went wider than dropping `FixedSizeBinaryVector` from `isArrowBacked`. That closes the case you found, but not the class it belongs to: `isArrowBacked` answers for the top-level vector and never looks at children, so a struct whose child is a `LargeVarCharVector` passes it today and is stored with 64-bit offsets and read back with 32-bit. The length check you proposed can't catch that one either, because `LargeUtf8` and `Utf8` are both three buffers, so the totals agree and the values are still wrong. The write path now asks the question directly — do this batch's vectors already carry the Arrow types the reader will rebuild, recursively — and converts the batch when they don't, which is the path it already takes for non-Arrow input. That makes the fast path safe by construction rather than safe for the vectors we happened to enumerate.

The second is that the length check does carry its weight. I mutated it away and re-ran the test written for it: it fails with "no exception was thrown", which is exactly the silent-wrong-answer mode rather than a crash. So the read path really was returning an answer from a layout it had never checked.

Both checks also make the failure modes distinct in a way that should help later: a writer that produces an unexpected layout is now converted at cache time, and a payload that somehow still disagrees at read time fails with an error naming the mismatch rather than an `ArrayIndexOutOfBoundsException` from inside a buffer copy.

Also changed while I was in here: an unrecognized body-compression byte is now rejected rather than read as plain bytes — `fromCompressionType` falls back to `NO_COMPRESSION`, so a corrupt payload was decoding to garbage rather than failing.

Verified on the default profile: `CometInMemoryCacheSuite` 44/44, `CometInMemoryCacheKryoSuite` and `UtilsSuite` green, `spotless:check` and `scalastyle:check` clean, and the whole thing test-compiles against Spark 3.5 and 4.0 as well as 4.1.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,525 @@
...
+      // A batch of zero columns carries only a row count, which a VectorSchemaRoot cannot infer
+      // without vectors to measure.
+      if (vectors.isEmpty) {
+        root.setRowCount(batch.numRows())
+      }
+
+      // alignBuffers=true matches the 8-byte buffer alignment Projection.load reproduces when it
+      // repacks the selected buffers.
+      val unloader = new VectorUnloader(root, true, codec, true)
+      val recordBatch = unloader.getRecordBatch
```
sunchao: **[P2] Release partial Arrow allocations when compression fails**

`VectorUnloader.getRecordBatch` retains each input buffer and accumulates compressed buffers without cleaning them up if a later compression fails. That failure happens before this method's `recordBatch.close()` finally becomes active. Closing the input batch afterwards does not release those extra references or the earlier compressed buffers, so a failed cache materialization can leave batch-sized off-heap allocations behind.

I reproduced this against the current `CachedBatchIpc.serialize` and Arrow 18.3.0 with 4,194,304 integer values, an unlimited `RootAllocator`, and valid zstd level 22. After warming the child JVM, limiting its virtual address space to its current size plus 64 MiB caused zstd's native workspace allocation to fail with an ordinary `RuntimeException: Error compressing: Allocation error : not enough memory`. After closing the input batch, **18,350,080 bytes (17.5 MiB) remained allocated**. The JVM remained usable. Under equivalent pressure, a component using the previous single-column stream-writing sequence threw `ZstdIOException` and returned Arrow allocation to zero after input close.

Please make write-side compression explicitly own and release partial buffers on failure, similar to the guarded decompression path, and add a failure-path regression test. Merely closing the input/root in a finally will not undo the unloader's retained references.

docs/source/user-guide/latest/in-memory-cache.md
```diff
@@ -0,0 +1,170 @@
...
+
+Comet can store Spark's in-memory cache (`CACHE TABLE`, `df.cache()`, `df.persist()`) in an Arrow
+format that Comet operators read directly. Without it, a cached table is stored in Spark's own
+format and every scan of it has to convert each batch before Comet can continue, which shows up in
+the plan as a `CometSparkColumnarToColumnar` above the cache scan.
+
+This feature is **experimental and disabled by default**.
+
+```scala
+spark.conf.set("spark.comet.exec.inMemoryCache.enabled", "true")
```
sunchao: **[P3] Configure the cache before SparkContext startup**

Starting from the documented default of `false`, this runtime setting cannot install Comet's cache serializer. `CometDriverPlugin` chooses `spark.sql.cache.serializer` once during SparkContext initialization, so subsequent cached relations still use Spark's default format and cannot take the native cache path.

I checked this with the current plugin on Spark 4.1.3: startup `false` followed by this runtime setting kept `DefaultCachedBatchSerializer`; startup `true` installed `ArrowCachedBatchSerializer`.

Please replace this activation example with `--conf spark.comet.exec.inMemoryCache.enabled=true` on startup, or builder configuration before creating a fresh SparkContext.

sunchao (commented):
Reviewed `3e74e9dcde021a043a9ad6ccd0322ade2279151c` across the IPC/vector boundary, Spark semantics, memory ownership, compatibility, and tests. The recursive writer-layout check and reader length guard address the earlier layout concern. I found two remaining issues, detailed inline: a write-side Arrow memory leak on recoverable compression failure, and an activation example that configures the cache after the serializer has already been selected.

**Validation:** JDK 17 root-reactor JVM test compilation passed. Exact-head IPC component checks passed 73,800 row comparisons across 19 types and 24 projections, including nested arrays/maps/structs, reordered/repeated/empty projections, nulls, timestamps, and uncompressed/zstd payloads. Normal round trips returned allocator usage to zero. I separately reproduced the compression-failure leak and checked the documentation example with actual Spark 4.1.3 contexts.

**Limits:** These were component and configuration probes, not full native-dependent Spark suites. Local native compilation was blocked because the configured dependency mirror lacks locked DataFusion 55.1.0. Benchmark performance was not remeasured. The previous-write-path comparison mentioned inline uses its single-column method sequence with an explicit SparkConf, not a full base-revision integration run.

**Current CI:** [Required Checks](https://github.com/apache/datafusion-comet/actions/runs/34999395851/job/104495793708) is failing. Five lint checks failed; inspected logs request removing the redundant `s` prefix from the error string at `CachedBatchIpc.scala:295`. The Rust test check and Spark 4.1/JDK 17 build passed, while native-build and downstream jobs were cancelled and several optional jobs were skipped. CI's merge commit `ca15f7da281b5a301e7498e9295d942537ec3d17` contains unrelated changes relative to this head, so those results are not exact-head validation.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: You're right, and the repro made this quick to confirm. `appendNodes` retains each input buffer and then does `buffers.add(codec.compress(...))` into a list that lives inside `getRecordBatch`, so when a compress throws, that retain is stranded and every buffer compressed before it is unreachable from anywhere a caller can get to. Closing the input batch afterwards releases the input's own reference, not the extra retain, and it never sees the compressed buffers at all.

Fixed by not handing the codec to the unloader. `serialize` now unloads with `NoCompressionCodec`, which retains and hands each buffer straight back without allocating, and a new `CachedBatchIpc.compressed` does the compression in the same shape `decompressed` already uses on the read side: retain, compress inside a try that releases on a throw, and an outer catch that closes whatever is already in the list. Peak memory is unchanged, since the unloader was doing the same thing one buffer at a time.

The retain is worth a note because it earns its keep differently in the two cases. A codec that allocates consumes the retained reference and returns a buffer of its own; `NoCompressionCodec` returns the input, and the retain is then the reference the result batch ends up owning. Either way the input buffer is back at its original count once the result is closed.

There is a regression test, and I mutated the fix away to check it isn't passing for free: it then fails with 2,176 bytes still allocated after the input is closed, at 256 rows, which is the same thing you measured at 17.5 MiB with a real workload. The codec it uses is a real zstd codec that throws on the third buffer, so the int column's two are genuine allocations by the time it does. Failing at the first buffer would have passed with no cleanup at all.

docs/source/user-guide/latest/in-memory-cache.md
andygrove: Right, and the page contradicted itself: the paragraph immediately below the example said the config is read at startup. It is now a startup `--conf` on `spark-shell`, with a sentence saying why -- the driver plugin picks `spark.sql.cache.serializer` while the `SparkContext` is initializing, so a session that started with the default goes on using Spark's format however the config is set afterwards. The "is read at startup" sentence in the next section came out, since it was then saying the same thing twice.

andygrove:
Thanks @sunchao — both are fixed in b978e5447, with the details on the individual threads. Two things worth pulling up here.

The leak is real, and the fix is not local to the failure point: `VectorUnloader` accumulates the compressed buffers into a list inside `getRecordBatch`, so nothing a caller holds can reach them once it throws. `serialize` now unloads plain and compresses in a new `CachedBatchIpc.compressed`, the same shape `decompressed` already uses on the read side. I mutated the fix away to check the new test earns its place: it then fails with 2,176 bytes still allocated after the input batch is closed, which is your 17.5 MiB at a test-sized batch rather than a 4M-value one.

The lint failure was the redundant `s` prefix you found. All five failing checks were the same scalafix run across profiles, so that one line covers them; I scanned the rest of the PR's files for a second instance and there isn't one.

Verified on the default profile: `CometInMemoryCacheSuite` 45/45, `CometInMemoryCacheKryoSuite` and `UtilsSuite` green, `spotless:check` and `scalastyle:check` clean, and test-compiles against Spark 3.5 and 4.0 as well as 4.1.

andygrove:
@comphead  @sunchao could you take another look at this PR?

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,586 @@
...
+        // own buffers by now, so this does not touch it, and getField still answers afterwards:
+        // clearing releases buffers, not the schema.
+        root.clear()
+
+        // Sized up front from the body length the record batch already knows, plus room for the
+        // metadata message. An unsized ByteArrayOutputStream starts at 32 bytes and doubles, so a
+        // multi-MiB payload would be reallocated and recopied a dozen-odd times per batch.
+        val sizeHint = recordBatch.computeBodyLength() + METADATA_SIZE_HINT
+        val out = new ByteArrayOutputStream(
+          math.min(math.max(sizeHint, METADATA_SIZE_HINT), Int.MaxValue.toLong).toInt)
```
sunchao: **[P2] Preserve support for batches larger than a single JVM array**

Could we retain chunked storage or split oversized batches by bytes before writing this payload? The row converter limits `conf.columnBatchSize` records but has no aggregate byte limit, so valid binary/string-heavy batches can exceed 2 GiB while every individual column still fits its Arrow buffers. The previous per-column `ChunkedByteBuffer` representation supported that case.

I checked this using the production row-to-Arrow converter and the exact-head IPC writer. A valid multi-column batch above that aggregate limit failed here with `OutOfMemoryError: Requested array size exceeds VM limit` under both default zstd and `none`. Replaying the previous per-column writer sequence successfully stored the same input in chunks. This was a component comparison, not a full base-revision Spark query.

Clamping the capacity to `Int.MaxValue` still requests an array the JVM rejects, regardless of available heap. Reducing only the initial capacity would also leave `toByteArray` unable to represent the payload. Please handle the size boundary before constructing the single-array payload and add regression coverage.

sunchao (commented):
Re-reviewed `6b8ded1dff647a0554aff007a74cc12123a6c6d0` across IPC layout, Spark integration, memory/error cleanup, compatibility, and tests. Both findings from my previous review are fixed. The real zstd allocation-failure replay now returns Arrow allocation to zero after input close, and the activation example correctly sets the cache option before SparkContext startup.

One new P2 remains: storing the entire batch in one `Array[Byte]` introduces an aggregate JVM array-size limit that the previous chunked per-column writer avoided. Details inline.

Validation: native build and JDK 17 root-reactor JVM compilation passed. All 55 focused tests passed (`CometInMemoryCacheSuite` 45, `CometInMemoryCacheKryoSuite` 4, `UtilsSuite` 6), using freshly compiled JVM classes and the CI native binary from an identical source tree. IPC component checks passed 73,800 value comparisons across 19 types and 24 projections, plus 48 allocation-pressure cases with no retained allocations after cleanup.

Current CI: 23 successful checks, including [Required Checks](https://github.com/apache/datafusion-comet/actions/runs/35641942481/job/106488819834), and 14 skipped checks. CI tested merge commit `ce9a15a3fdd66d99eab3672c77f6ae7f89419ed6`, whose tree is identical to the reviewed head. The large-batch comparison is a production row-converter/IPC component probe and a replay of the previous writer sequence, not a full base-revision Spark integration run. The generated jar has the correctly relocated Arrow compression service registration, but the package goal later stopped on a local Maven DNS failure resolving provided `protobuf-java:4.33.0`. Benchmark performance was not remeasured.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
sunchao: Confirmed fixed at `6b8ded1dff647a0554aff007a74cc12123a6c6d0`. I reran the real zstd workspace-allocation failure from the earlier review. The failed write now retains only the original input allocation, and closing the input returns Arrow allocation to zero. The new compression-failure regression test also passed in the focused suite.

docs/source/user-guide/latest/in-memory-cache.md
sunchao: Confirmed fixed at `6b8ded1dff647a0554aff007a74cc12123a6c6d0`. The example now enables the cache with a startup `--conf`, and the explanation matches the serializer selection in `CometDriverPlugin` during SparkContext initialization.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Agreed, this was a real regression. Fixed in 14bf8455f by storing the payload in chunks again.

- `CometCachedBatch.bytes` is now a `ChunkedByteBuffer`. `CachedBatchIpc.serialize` writes through `ChunkedByteBufferOutputStream` with 1 MiB heap chunks, the same size main uses, so a single cached batch is no longer capped at 2 GiB. As a side effect the write no longer holds a second copy of the whole payload.
- `Projection.load` reads the metadata from the chunked input stream and copies each selected buffer's window chunk by chunk, so a window can span any number of chunks. A window that falls outside the payload is reported as a corrupt batch rather than an index error.
- Kryo needed no change: `Utils.arrowBytesKryoClasses` already registers `ChunkedByteBuffer` and its heap chunks for the broadcast path, and `CometInMemoryCacheKryoSuite` passes.

Tests: an internal chunk-size config lets the projection-corruption and leak tests run a second time with 61-byte chunks, so every buffer window and the metadata message cross chunk boundaries. A new test compares values against the uncached query over such a payload and asserts it really is stored in several chunks. Breaking the chunk lookup makes all five chunk tests fail.

The 2 GiB boundary itself isn't unit-testable without allocating that much. What is covered is the cross-chunk read path it depends on.

docs/source/user-guide/latest/in-memory-cache.md
```diff
@@ -0,0 +1,176 @@
...
+it.
+
+Compression defaults to `zstd`, which is faster than storing cached batches uncompressed: the
+bytes it saves cost more to copy and store than compressing them costs. Measured over a 200k-row,
+six-column relation:
+
+| Codec  | Materialize | Footprint | Read 1 of 6 | Read 6 of 6 |
+| ------ | ----------: | --------: | ----------: | ----------: |
+| `zstd` |      363 ms |     2 MiB |       56 ms |       62 ms |
+| `none` |     1776 ms |    13 MiB |       78 ms |       81 ms |
```
mbutrovich, line 80: How was this table measured? None of the committed benchmarks vary the codec, so I couldn't reproduce it. The `none` materialize time is the number I'd like to understand. With `NoCompressionCodec`, `CachedBatchIpc.compressed` hands the input buffers straight back ([CachedBatchIpc.scala#L459-L498](https://github.com/apache/datafusion-comet/blob/14bf8455f91c59133aceba0903fc099830b9ab8c/spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala#L459-L498)), so the write does strictly less work than `zstd` apart from copying about 11 MiB more into heap chunks. 1413 ms for 11 MiB of extra copying is about 8 MB/s, which is far slower than a memcpy. Was `none` measured first in a fresh JVM, before JIT warmup? If so, the order would explain most of the gap.

This table is the justification for the default and for the claim in the `compression.codec` doc string, so could you add a codec axis to `CometInMemoryCacheBenchmark` so the table can be regenerated the same way as the ones below it?

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,633 @@
...
+    if (compressionType == NoCompressionCodec.COMPRESSION_TYPE) {
+      None
+    } else {
+      val codecType = CompressionUtil.CodecType.fromCompressionType(compressionType)
+      if (codecType == CompressionUtil.CodecType.NO_COMPRESSION) {
+        throw new SparkException(
+          s"Comet cached batch records an unknown Arrow compression codec: $compressionType")
+      }
+      Some(readCodecs(codecType))
+    }
```
mbutrovich, line 113: Could you add a test for the unknown-codec branch? It's the only new failure path in the read path that the suite doesn't reach. One way is to build an `ArrowRecordBatch` with `new ArrowBodyCompression(<unused codec byte>, BodyCompressionMethod.BUFFER)`, write it with `MessageSerializer.serialize` into a `ChunkedByteBuffer`, and assert that `Projection.load` throws a `SparkException` naming the codec. If that branch regressed to trusting `fromCompressionType`, the read would return garbage values rather than fail, and no current test would notice.

spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala
```diff
@@ -1035,12 +1042,57 @@ class CometInMemoryCacheSuite extends CometTestBase {
...
+          "SELECT id, CAST(concat('X', cast(id as string)) AS STRING) COLLATE UTF8_LCASE AS s " +
+            "FROM range(100)")
+        .createOrReplaceTempView("collated_case_cache")
+      spark.catalog.cacheTable("collated_case_cache")
+      spark.table("collated_case_cache").count()
+      assert(
+        spark.sql("SELECT id FROM collated_case_cache WHERE s = 'x1'").collect().length == 1,
+        "a case-insensitive match must survive pruning")
+
       // Null-count based pruning stays available for columns without bounds.
```
mbutrovich, line 1059: With this PR the collated column has bounds, so this comment no longer describes the assertion under it. The same sentence at line 1091 is still correct for the binary column.

```suggestion
      // IsNotNull is pushed down through the null count.
```

mbutrovich (commented):

docs/source/user-guide/latest/in-memory-cache.md
andygrove: You were right to doubt it. That table didn't come from a committed benchmark, and it doesn't survive one: with each case warmed up, `none` beats zstd on everything except footprint, so it was supporting the default with the wrong argument.

`CometInMemoryCacheBenchmark` now has a codec axis over the 5M-row flat relation: materialize time with the timer around the caching alone, footprint from the relation's size accumulator, and one-column and six-column reads. zstd takes 1503 ms to materialize against 1091 ms for `none`, reads one column in 47 ms against 36 ms and all six in 296 ms against 63 ms, and holds 55 MiB against 315 MiB. The ~410 ms materialize gap also matches the write path on its own: timing `CachedBatchIpc.serialize` directly puts zstd about 87 ns/row above `none`, ~435 ms over 5M rows. Every table on the page is now regenerated from one run.

So zstd is a footprint default, not a speed one, and the docs and the `compression.codec` doc string now say that, including when `none` is the better choice. I've kept zstd as the default, since a cache that doesn't fit costs more than one that reads slower and Spark's own format compresses by default too, but I'm open to flipping it.

The LZ4 claim came from the same kind of one-off, so I checked it with the same direct timing, since the config rejects that codec and the benchmark can't reach it. It holds: 7.9 s to serialize one 10k-row batch against 0.96 ms for zstd, with 2.5x larger output.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Added, built the way you suggested. `CometCachedBatchHelper.cachedBatchWithBodyCompression` unloads a batch plain, tags it with `new ArrowBodyCompression(99, BodyCompressionMethod.BUFFER)` and writes it with `MessageSerializer.serialize` into a `ChunkedByteBuffer`. The layout is exactly what the writer produces, so the read gets past the layout check to the codec lookup. The test reads it through `convertCachedBatchToColumnarBatch` and asserts that the failure names the codec and surfaces as itself rather than as a reference-count error from cleanup. It checks the message rather than the type, because the job abort wraps it in a `SparkException` either way. With `readCodec` put back to trusting `fromCompressionType`, it fails with "no exception was thrown".

spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala
andygrove: Applied, thanks.

andygrove:
Thanks @mbutrovich, all three are addressed in 20dfd0838, with details on the threads. The codec table is worth pulling up here, because the answer changed the docs and not just the benchmark. The old numbers had zstd winning on speed as well as size. With the codec now varied by the benchmark, `none` is faster at everything but footprint. zstd stays the default, but on footprint grounds, and the docs and the config doc now say so.

Regenerating the numbers also turned up a build break this PR introduced. Any build without `clean` in which no class has changed, such as a second `make release`, fails in shade with a duplicate `META-INF/services` entry. The jar plugin leaves the already-shaded jar in place, and the services transformer then writes the relocated name twice. The fix is an exclude in `spark/pom.xml`. I reproduced the failure without it, and both clean and non-clean builds pass with it.

Verified on the default profile: `CometInMemoryCacheSuite` 49/49, with `CometInMemoryCacheKryoSuite` and `UtilsSuite` green. It compiles against Spark 3.4, 3.5 and 4.0 as well as 4.1, and scalafix, scalastyle and spotless are clean.

sunchao:
Found **three P2 issues** at head `20dfd083`, using five review agents and independent reproduction.

1. **Collated prefix pruning can discard matching rows.**  
   [ArrowCachedBatchSerializer.scala:374](https://github.com/apache/datafusion-comet/blob/20dfd08384e0b26bb23f92580d31740c0e337c43/spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/ArrowCachedBatchSerializer.scala#L374) now admits collated strings into Spark’s prefix-pruning rewrite. That rewrite assumes matching prefixes preserve character counts, which collations can violate. With `UTF8_LCASE`, values beginning with `İ` match prefix `i\u0307`, but caching changed the result from five rows to zero. Disabling pruning restores them; base Comet also returns them correctly. Spark’s default cache shares this latent bug. Exclude unsafe collated `StartsWith` predicates while retaining equality/range pruning.

2. **A tiny raw buffer can retain the entire compressed allocation.**  
   [CachedBatchIpc.scala:345](https://github.com/apache/datafusion-comet/blob/20dfd08384e0b26bb23f92580d31740c0e337c43/spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala#L345) makes every selected buffer a slice of one allocation. Arrow’s raw fallback preserves that shared ownership after other buffers decompress. In a 4,096-row Boolean/string batch, a **512-byte bitmap pinned an extra 2 MiB**, raising live Arrow allocation from approximately **2.00 MiB to 4.03 MiB** versus the base format. It releases on batch close, so this is excess live memory, not a permanent leak. Isolate raw fallback buffers when mixed with compressed buffers.

3. **Two new value tests compare the cache against itself.**  
   At [CometInMemoryCacheSuite.scala:1709](https://github.com/apache/datafusion-comet/blob/20dfd08384e0b26bb23f92580d31740c0e337c43/spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala#L1709), also line 1744, Spark substitutes the cached relation into the supposedly uncached reference query. Changing a stored nested value to `777777` still passed both tests’ comparisons; a reference collected before caching caught it. Collect expected rows before cache registration. This demonstrates a test weakness, not independent production corruption.

The overall design is reasonable: `CachedBatchIpc` separates format handling from lifecycle management, and per-partition `Projection` avoids repeated schema traversal. Explicit codec cleanup is justified. Performance remains workload-dependent: a warmed component comparison showed smaller payloads and faster narrow reads, but slower writes and full-width reads. Those measurements are not whole-query results.

Validation passed **59 focused tests**, **147,600 independent value comparisons**, and **2.25 GB roundtrips with both codecs**, confirming the previous array-size issue is fixed. [CI](https://github.com/apache/datafusion-comet/actions/runs/35872317193) has 24 successful and 14 skipped checks; it tested a merge commit. Full upstream Spark SQL and Iceberg suites remain unverified.

docs/source/user-guide/latest/in-memory-cache.md
```diff
@@ -0,0 +1,184 @@
...
+- Cached tables are scanned by `CometInMemoryTableScan`, which feeds Comet operators directly.
+- Per-batch column statistics are recorded in the layout Spark's `SimpleMetricsCachedBatchSerializer`
+  expects, so Spark can prune whole cached batches on a predicate before any of them is decoded.
+
+Relations whose schema Comet's Arrow writer cannot store — interval types, most notably — are
+delegated in full to Spark's default cache format, per relation. Nothing about the format depends
+on a runtime config, because `spark.sql.cache.serializer` is a static setting and a relation whose
+format could change mid-session could not be read back reliably. Turning
+`spark.comet.exec.inMemoryCache.enabled` off at runtime only sends cached scans back to Spark's
+execution path; the cached data stays readable either way.
```
mbutrovich, line 56: "Nothing about the format depends on a runtime config" isn't true anymore. `spark.comet.exec.inMemoryCache.compression.codec` is a runtime config, and it decides the codec byte each batch is written with. The data stays readable because the reader takes the codec from the batch, so how about saying that instead?

```suggestion
Relations whose schema Comet's Arrow writer cannot store — interval types, most notably — are
delegated in full to Spark's default cache format, per relation. Which format a relation uses does
not depend on a runtime config, because `spark.sql.cache.serializer` is a static setting and a
relation whose format could change mid-session could not be read back reliably. The compression
codec is a runtime config, but each batch records the codec it was written with, so data cached
under one setting stays readable after the setting changes. Turning
`spark.comet.exec.inMemoryCache.enabled` off at runtime only sends cached scans back to Spark's
execution path; the cached data stays readable either way.
```

docs/source/user-guide/latest/in-memory-cache.md
```diff
@@ -0,0 +1,184 @@
...
+runs. Without it, caching fails with a "Class is not registered" error that does not name this
+feature. Comet's driver plugin warns at startup when it sees Kryo, `registrationRequired`, and no
+registrator.
+
+## Limitations
+
+Reads that feed **Spark** operators rather than Comet ones are still slower than Spark's own cache
+format, by roughly 1.7x to 2.5x depending on how wide the projection is. Those reads pay a row
+conversion that Spark's format avoids with generated code over its own layout. This is why the
+feature is off by default.
```
mbutrovich: Where does "roughly 1.7x to 2.5x" come from? It matches the table in #5485, which was measured on the per-column format with Spark's whole-stream codec. This PR changes both, and the new codec table shows a full-width zstd read at 296 ms against 63 ms for `none`, so the Spark-operator read path has probably moved too. Could you re-measure it on this format, or link #5485 and say which format the number describes? This is the reason the page gives for the feature being off by default, so it should describe the format that ships.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
```diff
@@ -0,0 +1,633 @@
...
+      while (k < bufferIndices.length) {
+        val buffer = batch.buffers(bufferIndices(k))
+        offsets(k) = buffer.offset()
+        lengths(k) = buffer.length()
+        total += DataSizeRoundingUtil.roundUpTo8Multiple(lengths(k))
+        k += 1
+      }
+
+      // allocator.buffer(0) is legal but yields a buffer no window can be sliced from, and an
+      // all-empty projection (every selected column a NullVector, say) would ask for exactly that.
```
mbutrovich: Could you add a test for the case this comment describes? I couldn't find a test that caches a `NullType` column, so a projection of only NullVector columns (`total == 0`) never reaches `load`. A zero-row batch is similar: every buffer is empty, so it goes through zstd's empty-buffer framing on write and the zero-length windows on read. Something like caching `SELECT id, NULL AS n` and reading `SELECT n` under both codecs, plus a zero-row batch through `CometCachedBatchHelper.serialize` and `Projection.load`, would cover both paths.

mbutrovich (commented):
Thanks @andygrove, the codec benchmark and the unknown-codec test cover what I asked about last round. This pass has two docs corrections and one test gap on the read path.

The PR description mentions that cache materialization retains about 800 bytes per partition, and that it reproduces on 492dd6f2a. Could you file an issue for it and link it here? It's easy to lose once this merges.

The description also says the committed-benchmark-results half of the Documentation item in #5487 isn't done. Could you split that half into its own checkbox on #5487 or a new issue, so the box doesn't get ticked when this merges?

docs/source/user-guide/latest/in-memory-cache.md
andygrove: Applied, thanks.

docs/source/user-guide/latest/in-memory-cache.md
andygrove: It described the old format. The range came from #5485, which measured the per-column format with Spark's whole-stream codec, and the page quoted it without saying so.

I've re-measured it on this format, and this time in `CometInMemoryCacheBenchmark` itself: a new axis reads the same 5M-row relation with Comet off, cached once in each format. Both formats can come out of one session because `InMemoryRelation` looks the serializer up in the active conf when a relation is cached, and each case asserts which serializer cached what it reads. The gap is wider than the page said and widest for the narrowest reads: 5.2x for a row count (183 ms against 35 ms), 4.8x for one column, 3.4x for three, and 1.5x for all six. The page now shows that table and links #5485, and no longer gives a cause, since #5485 says it isn't established.

spark/src/main/scala/org/apache/spark/sql/comet/execution/arrow/CachedBatchIpc.scala
andygrove: Added both. The NullType test caches `SELECT id, NULL AS n` under both codecs and reads `SELECT n` and `SELECT id, n` against rows collected before caching. The zero-row test serializes an int and a string column with no rows under both codecs, loads both columns together and each on its own through `Projection.load`, and checks the allocator is back to zero afterwards. Both paths already worked, so these are coverage rather than a fix. The comment you pointed at is gone, though: empty windows now take Arrow's shared empty buffer, which is what its own codec returns for them, so an all-empty projection allocates nothing.

andygrove:
Thanks @sunchao, all three were real and are fixed in a710d5fd7.

The collated prefix pruning: `buildFilter` now leaves out any predicate holding a `StartsWith` over a string that isn't the default collation, and keeps equality and range pruning on those columns, since those compare with the collation itself. The new test caches your U+0130 values under UTF8_LCASE and gets 0 of 5 rows back without the fix.

The pinned allocation is fixed without adding a copy. `Projection.load` reads each window's uncompressed-length prefix and puts the ones Arrow stored raw in a second allocation, so decompressing frees the one holding the compressed windows. Empty windows take Arrow's shared empty buffer instead of a slice of either. The new test builds your Boolean-and-string shape at 4,096 rows and checks what each bitmap is backed by: before the fix, a 512-byte bitmap kept a 1 MiB allocation alive.

On the value tests, both now collect their reference rows before the relation is cached. The codec round-trip test had the same flaw, since `checkSparkAnswer` reads the one cached payload on both sides, so it does the same now. With the writer planted to store 777777 as the first `id` of each batch, all three fail.

andygrove:
@mbutrovich, on your two top-level points. I've split committed benchmark results out of #5487's documentation item into a checkbox of their own, so the documentation box can be ticked on merge without them.

The ~800 bytes per partition I couldn't reproduce when writing the issue up, so I haven't filed one. Caching relations of 1 to 64 partitions at 100 and 10,000 rows each, through both write paths and from a native Parquet scan, `CometArrowAllocator` is back to where it started while the relation is cached, after a read, and after uncaching, and every allocator the cache path uses is a child of it. Rather than file something I can't substantiate, I've taken the note out of the description.

Verified on the default profile: `CometInMemoryCacheSuite` 53/53, and it also passes on Spark 3.5 and 4.0. It compiles against 3.4, 3.5, 4.0 and 4.1, and scalafix, scalastyle and spotless are clean.

mbutrovich (approved):

Issue #5551, opened 2026-08-29
[Variant] Enable direct top-level Variant projection in ordinary native Parquet scans


peterxcli:
### Problem

After the schema, normalization, and physical-reader layers are ready, Comet still needs a deliberately narrow JVM admission path for direct top-level Variant projection and Spark-compatible behavior for missing columns and unsupported consumers.

Parent: #5546
Depends on: #5550

### Solution

- Admit only direct top-level `VariantType` fields in ordinary native Parquet scans.
- Reuse Unicode Parquet name matching from #5602 (#5495 is complete); do not port #5407's non-ASCII-name fallback.
- Build the native data schema from Spark's pruned required schema so unread Variant roots remain removable.
- Preserve existence-default value/index pairing and transport a Variant default as a scan-only constant `[value, metadata]` struct.
- Apply a default only when the physical column is absent.
- Export the marked Field and reuse `CometStructVector` so inherited `ColumnVector.getVariant` works.
- Keep nested Variant, pushed `VariantStruct` (#5519), strict malformed reads, expressions, writes, C2R, shuffle/spill, Python operators, encryption, and Iceberg on explicit Spark fallback where applicable.
- Document the supported surface and remaining fallbacks.

### Tests

- `SELECT v` and `SELECT id, v, tail` remain native and match Spark.
- Objects, arrays, scalars, JSON null, SQL NULL, nullable parents, and entirely-null columns round-trip.
- Spark sees logical `VariantType` with exact `[value, metadata]` Binary children.
- A missing Variant column receives its existence default without shifting later defaults.
- Unread Variant pruning from #5377 remains native.
- Case-insensitive Unicode Variant projection remains native and matches Spark, covering a non-ASCII Variant root, an ordinary sibling beside Variant, and physical Kelvin sign `K` matched to logical `k`. Preserve Spark's missing-column behavior for physical `ſ` read as logical `s`.
- Unsupported consumers above have focused fallback assertions.
- Spark 4.0/4.1 profiles pass and Spark 3.5 still compiles unchanged.

Completing this issue closes the ordinary-Parquet phase of #4295. Iceberg and the broader shredded optimization scope in #3983 remain separate.

peterxcli:
take

Issue #5546, opened 2026-08-29
[Variant] Native top-level Variant projection from ordinary Parquet


peterxcli:
### What problem does this solve?

#5407 implements the ordinary-Parquet phase of #4295, but it now combines several independently reviewable concerns: shared Arrow FFI schema export, Variant logical-type transport, native whole-value normalization, Spark-compatible Parquet physical semantics, and scan admission/schema-evolution defaults.

This issue tracks re-landing that work as five atomic pull requests under the broader Variant roadmap in #5438, following the maintainer request in https://github.com/apache/datafusion-comet/pull/5407#issuecomment-5463550828.

### Scope

Support direct, top-level, whole-value `VariantType` projection from ordinary Parquet:

```sql
SELECT v FROM parquet_table;
SELECT id, v, tail FROM parquet_table;
```

The completed series preserves Variant identity through:

```text
Spark VariantType
  -> Comet protobuf
  -> marked Arrow Field
  -> ordinary native Parquet scan
  -> whole-value normalization
  -> Arrow C Data Interface
  -> CometStructVector
  -> ColumnVector.getVariant
```

This tracker does not include Iceberg, nested Variant, Variant expressions, subfield pruning, predicate pushdown, writes, columnar-to-row, shuffle/spill, or Python transport.

### Ordered subtasks

- [x] #5547 — Export complete Arrow Fields through the native FFI boundary (#5552)
- [x] #5548 — Carry Spark VariantType identity through Comet schema serialization (#5631)
- [x] #5549 — Normalize marked Variant arrays at the native Parquet boundary (#5715, in review)
- [ ] #5550 — Match Spark physical Parquet semantics for projected Variant columns
- [ ] #5551 — Enable direct top-level Variant scan admission, defaults, and fallbacks

Each pull request must include focused tests and retain explicit Spark fallback until the complete scan boundary is ready.

### Definition of done

- Direct top-level Variant projection remains a Comet native Parquet scan and matches Spark.
- Fields before and after Variant retain their order.
- Objects, arrays, scalars, Variant JSON null, SQL NULL, and nullable parents round-trip.
- Canonical unshredded and shredded inputs reconstruct whole values.
- Spark receives logical `VariantType` with exactly `[value, metadata]` children.
- Unread Variant pruning from #5377 remains native.
- Unsupported expressions, nested Variant, pushed `VariantStruct`, native writes, C2R, shuffle/spill, Python operators, and Iceberg retain explicit fallback.
[...]

PR #5407, opened 2026-08-21
feat: project Spark 4 VARIANT columns in native Parquet scans [Splitting]

peterxcli:
## Split series

The implementation is being landed through #5546:

- #5552 — complete Arrow Field export (merged; #5547).
- #5631 — Variant type identity (merged; #5548).
- #5715 — marked Variant normalization (in review; #5549).
- #5550 — physical Parquet reader compatibility (next).
- #5551 — JVM scan admission, defaults, and fallbacks (final).

#5495 is complete through #5602. The split PRs reuse its shared Unicode Parquet field matcher; the original branch's non-ASCII fallback and native Unicode fail-closed guard must not be carried into #5550/#5551. Encoded Variant object-key ordering remains tracked separately by #5474, and remaining field-ID shadowing remains in #5550.

The reference implementation and benchmark below describe this original combined PR and its historical dependency baseline. The linked split PRs define the changes being reviewed for merge.

## Which issue does this PR close?

Implements the ordinary-Parquet phase of #4295.

Related work:

- #3983 covers shredded Parquet writing, subfield pruning, and predicate pushdown.
- #5438 tracks the broader native Variant roadmap.
- #5474 tracks removal of the Spark UTF-16 compatibility rewrite after every Spark profile supported by Comet includes the upstream UTF-8 ordering fix.
- #5477 tracks removal or reassessment of isolated Arrow-rs/DataFusion reader compatibility paths, including encoded children, physical ENUM, wide decimals, and per-scan Arrow-schema policy.
- #5495 is complete through #5602; the split series reuses its shared matcher and omits this reference branch's non-ASCII-name fallback.
- #5519 tracks native support for Spark's pushed one-field `VariantStruct`, which is required for this path to engage by default on Spark 4.1+ (see below).

This PR enables whole-value projection of direct, top-level Spark 4 `VariantType` columns through Comet's ordinary native Parquet scan:

```sql
SELECT v FROM parquet_table;
SELECT id, v, tail FROM parquet_table;
```

Iceberg Variant projection and general native Variant expressions remain outside this phase.

[...]

PR #5552, opened 2026-08-29
fix: preserve Arrow Field metadata across C Data exports

### LINKED ISSUE #5547: [Variant] Export complete Arrow Fields through the native FFI boundary
### Problem

The native Arrow C Data exporter constructs `FFI_ArrowSchema` from `ArrayData.data_type()`. That preserves the physical datatype but drops the top-level Field name, nullability, and metadata. Logical types represented by Field metadata cannot survive that boundary reliably.

Parent: #5546

### Solution

- Pass the corresponding RecordBatch Field to every native array-export call.
- Construct `FFI_ArrowSchema` from that Field.
- Keep `FFI_ArrowArray` export unchanged.
- Ensure offset-normalized arrays still use the original output Field.
- Substitute embedded NUL only in the exported C-string name while preserving datatype, nullability, and metadata.

This task must not add Variant scan admission or Variant-specific schema inference.

### Tests

- Top-level Field name, nullability, and metadata survive schema export.
- Offset-normalized arrays retain the original Field.
- A top-level name containing NUL does not panic or lose metadata.
- Existing non-Variant array exports remain unchanged.

peterxcli:

Closes #5547.

First Split of 
- #5552 <- you are here 
- #5631
- #5715
- #5794 
- #5551

## Rationale for this change

Native output currently exports each Arrow array with an `FFI_ArrowSchema` reconstructed from `ArrayData.data_type()`. A DataType describes the physical layout, but it does not carry the enclosing Field name, nullability, or metadata. Logical types identified by Field metadata therefore lose their identity at the Arrow C Data boundary.

The RecordBatch already owns the authoritative Field for each output column. Exporting that Field fixes the shared boundary once for query output and the native Parquet reader.

ArrowSchema names are NUL-terminated C strings. Passing the Field also makes embedded NUL names visible to the exporter, so this change substitutes U+FFFD only in the exported name while retaining the Field datatype, nullability, and metadata. Spark's logical output name remains owned by its plan.

- Pass the corresponding RecordBatch Field to `SparkArrowConvert::move_to_spark`.
- Construct `FFI_ArrowSchema` from the Field while leaving `FFI_ArrowArray` export unchanged.
- Use the original output Field for offset-normalized arrays.
- Apply the same Field-aware export in `Native.currentColumnBatch`.
- Add a focused round-trip test for name sanitization, datatype, nullability, and metadata.

This PR does not add a Variant protobuf type, Arrow extension mapping, Parquet normalization, or scan admission. Those remain in the later #5546 subtasks.

```shell
cd native
cargo fmt --all -- --check
cargo test -p datafusion-comet test_ffi_schema_preserves_field_and_sanitizes_nul_name
cd ..
git diff --check upstream/main...HEAD
```

The focused Rust test passed: 1 passed, 0 failed, 199 filtered out. Building that test also compiled both updated native export callers.


peterxcli:
cc @sunchao

sunchao (commented):
Completed five independent review passes covering native FFI correctness, Spark imports/planning, ownership and error handling, compatibility, and test coverage at a0c8f8978b8f51db4a61746efa20ad4d27217a23.

Found one P2 regression in broadcast coalescing, detailed inline. The Spark 4.1.3/JDK 17 reproduction fails on this head and passes on the exact base 98cd8c967995906500b12a709c25724b4cfba634. The focused Rust test, all four NativeUtilSuite tests, and standalone native/JNI compatibility probes pass. The full Spark version/CI matrix was not run locally.

native/core/src/execution/utils.rs
```diff
@@ -19,28 +19,45 @@
...
+fn ffi_schema_for_field(field: &Field) -> Result<FFI_ArrowSchema, ArrowError> {
+    if field.name().contains('\0') {
+        // ArrowSchema names are NUL-terminated C strings. Spark owns the logical output name, so
+        // substitute only the exported name while retaining the Field's type and metadata.
+        let field = field
+            .clone()
+            .with_name(field.name().replace('\0', "\u{fffd}"));
+        FFI_ArrowSchema::try_from(&field)
+    } else {
+        FFI_ArrowSchema::try_from(field)
```
sunchao: [P2] Reconcile union nullability before broadcast coalescing

Exporting the actual Field exposes different top-level nullability across `CometUnionExec` branches, but the union forwards those batches unchanged to `Utils.coalesceBroadcastBatches`. Its `VectorSchemaRootAppender` requires matching nullability and throws `IllegalArgumentException: Vector schema roots have different schemas` for an ordinary integer payload that is nonnullable in one branch and nullable in the other.

Reproduced with AQE disabled and a Parquet `t(_1 INT, _2 INT)` containing `(1,10), (2,20), (3,30)`:

```sql
SELECT /*+ BROADCAST(b) */ p._1, b.v
FROM t p JOIN (
  SELECT _1 AS k, 99 AS v FROM t
  UNION ALL
  SELECT _1 AS k, _2 + 1 AS v FROM t
) b ON p._1 = b.k;
```

On Spark 4.1.3/JDK 17, this fails at this head and returns all six expected rows at the exact base `98cd8c967`; the Spark-only control also passes. The old DataType export gave both payloads identical top-level flags. Please reconcile compatible nullability differences during broadcast coalescing (or leave these batches uncoalesced), retain complete Field export, and add this regression case.

peterxcli: Thanks, fixed in 7ef3bbd15. I kept complete Field export and added a strict schema-equality guard before VectorSchemaRootAppender.append. If a decoded broadcast batch has a different schema, including top-level nullability, coalescing now closes the partial target and returns the original buffers instead of weakening or rewriting the Field. I also added the exact AQE-disabled Parquet/UNION/BROADCAST regression; it requires CometUnionExec and CometBroadcastExchangeExec and asserts zero coalescing metrics. The focused Spark 4.1 test passed (1 test), along with the Rust FFI test, Spotless, and git diff --check.

sunchao (commented):
Re-reviewed 7ef3bbd15261d6af6c19486e35d99e34930567ec across five independent scopes. No new actionable findings.

The earlier broadcast-nullability regression is fixed. All 38 JVM tests passed on Spark 4.1.3/JDK 17: the full CometJoinSuite (30), NativeUtilSuite (4), and four independent UNION/BROADCAST cases with AQE on/off and actual null values. The existing large homogeneous-broadcast test still confirms successful coalescing.

The committed regression also passed a sensitivity check: removing only the guard restores the schema-mismatch failure; simulating the old primitive export schema fails its zero-coalescing-metrics assertion. Additional FFI, buffer reuse, and cleanup probes passed.

Native sources are unchanged from the previously reviewed head; I verified that identity and reused the tested native library. The full Spark version/CI matrix was not run locally.

sunchao (approved):
The previously reported broadcast issue is fixed. Re-review and focused validation at 7ef3bbd15261d6af6c19486e35d99e34930567ec found no remaining actionable issues. CI is still running.

native/core/src/execution/utils.rs
```diff
@@ -19,28 +19,45 @@
 use crate::execution::operators::ExecutionError;
 use arrow::{
     array::ArrayData,
+    datatypes::Field,
+    error::ArrowError,
     ffi::{FFI_ArrowArray, FFI_ArrowSchema},
 };
 
+fn ffi_schema_for_field(field: &Field) -> Result<FFI_ArrowSchema, ArrowError> {
```
andygrove: The PR body says Spark owns the logical output name, and what #5546 needs from this boundary is the Arrow extension metadata. Given that, is there a reason to export the whole `Field` rather than just the metadata? Something like:

```rust
Field::new("", field.data_type().clone(), true).with_metadata(field.metadata().clone())
```

would carry what the Variant work needs while leaving the observable shape of the boundary identical to today.

That matters because the name and the nullability are the two parts that cost something here. The name is why the NUL sanitizer had to be invented at all. Before this PR the exported name was always `""` and could never contain a NUL, so this change creates the failure mode and then patches it. The nullability is what broke broadcast coalescing in the earlier review round, and now needs a guard in `Utils.scala` plus a regression test that asserts the optimization is off.

If a later subtask in #5546 genuinely needs the real name or nullability, could you say which one in the description? Then the extra surface is clearly paying for something. If not, narrowing to metadata would let the NUL helper and the Scala guard both go away.

One more thing if the name export does stay. `FFI_ArrowSchema::try_from` recurses into children and calls `CString::new` on every nested field name too, so a struct or list whose *child* name contains a NUL still fails with `ArrowError::CDataInterface`. If NUL names are reachable enough to handle at the top level, should the sanitization walk the whole field tree so both cases behave the same way?

native/core/src/execution/utils.rs
```diff
@@ -65,3 +82,26 @@ impl SparkArrowConvert for ArrayData {
...
 pub use datafusion_comet_common::bytes_to_i128;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use arrow::datatypes::DataType;
+    use std::collections::HashMap;
+
+    #[test]
+    fn test_ffi_schema_preserves_field_and_sanitizes_nul_name() {
```
andygrove: This test mostly exercises `FFI_ArrowSchema::try_from` and the three-line NUL replacement. It does not go through `move_to_spark`, either alignment branch, or the actual JVM boundary.

The behavior the PR exists to establish, that field metadata survives from a native `RecordBatch` into a Spark-side `CometVector`, has no coverage yet. Would it be worth adding a round trip in `NativeUtilSuite` that asserts `getValueVector.getField.getMetadata` on the imported vector? Without one, a future change back to a `DataType`-derived schema would leave every test here green.

spark/src/main/scala/org/apache/spark/sql/comet/util/Utils.scala
```diff
@@ -396,6 +396,13 @@ object Utils extends CometTypeShim with Logging {
             }
             while (reader.loadNextBatch()) {
               val sourceRoot = reader.getVectorSchemaRoot
+              if (targetRoot != null && targetRoot.getSchema != sourceRoot.getSchema) {
```
andygrove: Two things here.

The check runs once per batch, but the schema is fixed for the life of an `ArrowStreamReader`. `getVectorSchemaRoot` reads the schema message when the stream is initialized, so this could move up to just after the reader is constructed and run once per buffer instead. `Schema.equals` walks every field and every metadata map, so on a wide relation with many broadcast batches this is repeated work for no added coverage.

On the guard itself, this compares full `Schema` equality, which now covers names and field metadata as well as nullability. So it fires on more than the union nullability case that motivated it. A broadcast over a union whose branches project differently named source columns would trip it, since Spark takes output names from the first child but each native child plan names its own output fields. Same for a union where one branch carries Variant metadata and the other does not, which is exactly where #5546 is heading.

How common do you expect that to be on real plans? Right now losing coalescing surfaces only as a `logWarning`, with no metric or `EXPLAIN` signal a user could act on.

native/core/src/parquet/mod.rs
```diff
@@ -308,8 +308,10 @@ pub extern "system" fn Java_org_apache_comet_parquet_Native_currentColumnBatch(
             .ok_or_else(|| CometError::Execution {
                 source: ExecutionError::GeneralError("There is no more data to read".to_string()),
             });
-        let data = batch_reader?.column(column_idx as usize).into_data();
-        data.move_to_spark(array_addr, schema_addr)
+        let batch = batch_reader?;
+        let field = batch.schema().field(column_idx as usize).clone();
```
andygrove: `.field(idx).clone()` deep-clones the metadata `HashMap` for every column of every batch on the scan path. Binding the schema first lets you pass a reference and skip the clone entirely:

```rust
let batch = batch_reader?;
let schema = batch.schema();
let field = schema.field(column_idx as usize);
let data = batch.column(column_idx as usize).into_data();
data.move_to_spark(field, array_addr, schema_addr)
    .map_err(|e| e.into())
```

spark/src/test/scala/org/apache/comet/exec/CometJoinSuite.scala
```diff
@@ -657,6 +657,30 @@ class CometJoinSuite extends CometTestBase {
...
+              |  UNION ALL
+              |  SELECT _1 AS k, _2 + 1 AS v FROM t
+              |) b ON p._1 = b.k
+              |""".stripMargin),
+          Seq(
+            classOf[CometBroadcastExchangeExec],
+            classOf[CometBroadcastHashJoinExec],
+            classOf[CometUnionExec]))
+
+        val broadcast = collect(cometPlan) { case b: CometBroadcastExchangeExec => b }.head
```
andygrove, line 678: Good regression coverage for the crash, and I like that it pins the exact shape from the earlier review round.

Worth flagging that asserting zero here pins the optimization *off* for this query. If the export gets narrowed to metadata only as suggested on the Rust side, this case should coalesce normally again and the assertion would want to flip.

andygrove (commented):
Thanks for splitting this out of #5407, the atomic scope makes it much easier to reason about. Hoisting `FFI_ArrowArray::new` and the schema conversion above the alignment branch is a real improvement on its own. The old code could write the array through `array_ptr` and then fail on the schema conversion, leaving the JVM holding a half-initialized pair. Now both are built before either write.

I also checked that `RecordBatch::try_new` validates column data types against the schema and rejects nulls in non-nullable fields, so sourcing the exported type from the schema field rather than `ArrayData::data_type()` cannot introduce a schema/buffer disagreement. That part is safe.

My main question is about scope, left inline on `ffi_schema_for_field`. The rest are smaller points on the coalescing guard, the Parquet path, and test coverage.

sunchao (commented):
[P2] Keep broadcast coalescing safe when array field metadata differs

At 754cc2f0, two native V1 Parquet scans of `(k INT, v ARRAY<INT>)` with explicit read schemas carrying different top-level `parquet.field.id` values on `v` can fail when their `UNION ALL` is the broadcast side of an inner join on `k` and `v` remains in the result. The list element fields can be identical. The [metadata export](https://github.com/apache/datafusion-comet/blob/754cc2f016017fa21d4f0ac79425761108ac21bb/native/core/src/execution/utils.rs#L43) and IPC paths preserve each array field's `PARQUET:field_id`; `CometUnionExec` forwards the child batches without normalizing those fields.

Arrow Java 18.3.0's outer `VectorSchemaRootAppender` check ignores metadata, but its [ListVector append path](https://github.com/apache/arrow-java/blob/8e84e4c8bbe041f362690e4ea54280ec682dbb1f/vector/src/main/java/org/apache/arrow/vector/util/VectorAppender.java#L314) uses `TypeEqualsVisitor(targetVector, false, true)`. The differing metadata therefore throws `IllegalArgumentException` and aborts the broadcast. BASE omitted these outer IDs; the previous reviewed head retained them but returned the original buffers when schemas differed. Please retain a compatible fallback or reconcile safe metadata differences while preserving the intended metadata export.

This is a source-derived finding for small, nonempty, non-dictionary batches, with Parquet columns read by name, native scan/execution/union/broadcast enabled, and AQE disabled. I did not execute the query. Primitive-only metadata differences do not trigger this list-specific check.

sunchao (commented):
Re-reviewed `ef4b5993`. The new fallback addresses the P2 array-metadata coalescing failure reported in [the previous review](https://github.com/apache/datafusion-comet/pull/5552#pullrequestreview-5059885746).

I’m withholding approval because the [macOS scan job](https://github.com/apache/datafusion-comet/actions/runs/33321907414/job/99288923061) ended in a native-thread SIGBUS and its relationship to this change remains unresolved.

sunchao (approved):
Re-reviewed `ef4b599373c5f9002b142cd9bb7d526d46d40d95` with five independent scopes using the repository's review checklist. No new actionable findings. The metadata-only export keeps the previous top-level names and nullability. Ordinary unions coalesce again, and incompatible array metadata safely falls back to the original buffers.

Fresh local validation passed: 40 JVM tests on Spark 4.1.3/JDK 17, including the full `CometJoinSuite`, `NativeUtilSuite`, four additional array/null/AQE cases, and the field-ID metadata assertion. That assertion fails with `Set(null)` using the exact-base native library and passes with this head, confirming that it protects the native-to-JVM metadata boundary. The focused Rust test and independent native, JNI/IPC, and coalescing component probes also passed. I did not run the full Spark/platform matrix locally.

CI currently has 60 passing, 9 skipped, and 2 failed checks. The [Linux exec job](https://github.com/apache/datafusion-comet/actions/runs/33321907414/job/99288650192) failed before its tests when Maven Central returned HTTP 429 for a formatting dependency.

I also investigated the [macOS scan crash](https://github.com/apache/datafusion-comet/actions/runs/33321907414/job/99288923061) that prompted the earlier concern. The crash dump and that run's exact native artifact resolve the return address to `hdfsThreadDestructor + 80`, immediately after its indirect `GetJavaVM` call. This strongly matches the existing HDFS teardown failure in [#5023](https://github.com/apache/datafusion-comet/issues/5023). The exact base and previous head both passed the matching macOS scan suite, and the previous head has identical native sources. I found no new PR defect behind these failures. Please rerun the two failed jobs before merging.

sunchao:
@peterxcli can you check the CI failure and see whether it is related? I've retried a few times
```
Error: Unable to download artifact(s): Artifact not found for name: native-lib-macos
        Please ensure that your artifact is not expired and the artifact was uploaded using a compatible version of toolkit/upload-artifact.
        For more information, visit the GitHub Artifacts FAQ: https://github.com/actions/toolkit/blob/main/packages/artifact/docs/faq.md
```

peterxcli:
> @peterxcli can you check the CI failure and see whether it is related? I've retried a few times
> 
> ```
> Error: Unable to download artifact(s): Artifact not found for name: native-lib-macos
>         Please ensure that your artifact is not expired and the artifact was uploaded using a compatible version of toolkit/upload-artifact.
>         For more information, visit the GitHub Artifacts FAQ: https://github.com/actions/toolkit/blob/main/packages/artifact/docs/faq.md
> ```

@sunchao Seems like it's a flaky? I merge upstream main again and this error is gone, however it looks like other flaky failure just appeared.

peterxcli:
@sunchao I think remaining failure are all intermittent, do you think this is ok to merge? Thanks!

sunchao:
@peterxcli From the code-review side, my approval still stands for `979cd3ae06b13c3037eab379f3bf9343ce3fed7a`, with no new P1/P2 findings in the current version. I can't confirm that the remaining failure is intermittent, though. The [current macOS scan job](https://github.com/apache/datafusion-comet/actions/runs/33488945876/job/99811695825) still has a native SIGBUS. The diagnostics identify the HDFS thread-cleanup location, but do not by themselves prove a flake or rule out a PR interaction. I have not rerun that job, so I would keep the unresolved CI result separate from my code approval.

peterxcli:
@sunchao looks like all test pass now

sunchao:
@peterxcli Confirmed: the current checks for `979cd3ae06b13c3037eab379f3bf9343ce3fed7a` now show 62 successful and 9 skipped, with no failing or pending check runs. The CI rerun, including the macOS scan job, is green. My code approval still stands. Thanks for following up.

sunchao:
Merged, thanks @peterxcli for the contribution!

PR #5556, opened 2026-08-29
fix: accept UTC timezone aliases in Python Arrow input

sunchao:

Follow-up to #5368. No separate issue.

## Rationale for this change

The schema check added in #5368 rejects valid timestamp batches when native producers use the equivalent timezone labels `UTC` and `Etc/UTC`.

With `spark.sql.session.timeZone=Etc/UTC`, a native Parquet scan produces `Timestamp(MICROSECOND, UTC)`, while native `date_trunc` preserves `Etc/UTC`. Unioning those branches and coalescing them into one partition makes `mapInArrow` and `mapInPandas` fail with `Arrow input schema changed between batches` when the second schema reaches the same Python input stream.

Confirmed on the merged commit `e0fa3ccb16d4c9b3a03a38b09af6ed0fb242e027`, in both branch orders.

- Accept `UTC` and `Etc/UTC` timestamp types interchangeably only when their units match, including in nested fields.
- Preserve the first batch's stream schema and existing buffers. Keep all other type, dictionary, extension-metadata, and child-layout checks unchanged.
- Add positive and negative Arrow compatibility tests and a 16-case Python regression covering both APIs, timezone labels, branch orders, and accelerated/fallback modes, with nulls and fractional timestamps.

- Merged-code control: the 16 new Python cases produced **4 failures and 12 passes**. The failures were exactly the accelerated `Etc/UTC` cases for both APIs and branch orders.
- Fixed code: the full `test_pyarrow_udf.py` suite passed **133/133**, including all 16 new cases, with no skips. This used Spark 4.1.3, real native execution, and real Python workers; accelerated cases assert the native Python, union, coalesce, and projection plan nodes.
- `CometArrowPythonRunnerSuite`: **11/11 passed**, including Arrow C Data JNI ownership/cleanup checks. Running the same tests on the merged control gives the expected single alias-compatibility failure.
- Full Spark 4.1 Spotless check and `git diff --check` passed.

Local execution used all production JVM sources compiled into an isolated unshaded test jar and a native library built from the exact merged commit. Standard Maven packaging could not finish because the inherited Maven mirror timed out; shaded packaging and the Spark 4.0 matrix remain for CI.

### DIFF
--- a/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
+++ b/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
@@ -334,7 +334,7 @@ private[python] object CometArrowPythonRunnerBase {
     ArrowType.ExtensionType.EXTENSION_METADATA_KEY_NAME,
     ArrowType.ExtensionType.EXTENSION_METADATA_KEY_METADATA)
 
-  private def hasCompatibleType(expected: ArrowType, actual: ArrowType): Boolean = {
+  private def areArrowTypesCompatible(expected: ArrowType, actual: ArrowType): Boolean = {
     if (expected == actual) {
       true
     } else {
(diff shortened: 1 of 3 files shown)


spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
```diff
@@ -334,10 +334,26 @@ private[python] object CometArrowPythonRunnerBase {
     ArrowType.ExtensionType.EXTENSION_METADATA_KEY_NAME,
     ArrowType.ExtensionType.EXTENSION_METADATA_KEY_METADATA)
 
+  private def hasCompatibleType(expected: ArrowType, actual: ArrowType): Boolean = {
```
comphead: maybe we can find better name?

comphead (approved):
Thanks @sunchao lgtm

Issue #5488, opened 2026-08-27
Large-offset Arrow vectors from PyArrow UDFs cannot be serialized for broadcast or collect


andygrove:
### Describe the bug

`Utils.getFieldVector` accepts a fixed list of Arrow vector types that excludes `LargeVarCharVector` and `LargeVarBinaryVector`, throwing `Unsupported Arrow Vector for serialize` for either. Comet otherwise supports those representations deliberately:

- `CometPlainVector` tracks whether the variable-width offsets are 64-bit, with the comment "PyArrow UDFs can hand back large_string / large_binary columns".
- `Utils.toArrowType` maps `ArrowType.LargeUtf8` to `StringType`, again noting PyArrow UDF output.
- `ArrowWriters` has `LargeStringWriter` and `LargeBinaryWriter`.

So a batch carrying a large-offset column can be produced and read, but not serialized. On `main` the reachable consumers are the two `Utils.serializeBatches` call sites:

- `org.apache.spark.sql.comet.operators.getByteArrayRdd` (`operators.scala`), used to collect a native plan's output as bytes.
- `CometBroadcastExchangeExec.getByteArrayRdd`.

The producer is `CometMapInBatchExec`, whose runner preserves the Arrow vectors a Python worker returns. A `mapInArrow` or PyArrow UDF returning `pa.large_string()` or `pa.large_binary()` whose result is then broadcast or collected should therefore fail at serialization.

### Steps to reproduce

Not run end to end. A `mapInArrow` returning `pa.large_string()`, with the result on the build side of a broadcast join (or collected), is the shape expected to hit it.

The rejection itself is trivially reproducible at the `Utils` level: build a `CometPlainVector` over a `LargeVarCharVector` holding two short strings, put it in a `ColumnarBatch`, and call `Utils.serializeBatches`. An ordinary `VarCharVector` succeeds on the same path.

### Expected behavior

A batch Comet can produce and read should be serializable, or should be normalized to a representation that is, rather than failing at the serialization boundary.

### Additional context

Raised during review of #5051 by @sunchao (https://github.com/apache/datafusion-comet/pull/5051#discussion_r3867601991), where it surfaced through the cache serializer. Filed separately because it is not specific to that PR and outlives it: `getFieldVector` and both `serializeBatches` consumers are on `main` today, while the cache serializer is not.

[...]

PR #5557, opened 2026-08-29
feat: support large Arrow inputs for accelerated Python UDFs

### LINKED ISSUE #5555: Support useLargeVarTypes in accelerated mapInArrow/mapInPandas

Spark's `spark.sql.execution.arrow.useLargeVarTypes=true` requests Arrow `large_string` and `large_binary` input columns with 64-bit offsets. Comet's accelerated `mapInArrow` / `mapInPandas` path currently falls back whenever this setting is enabled, even when `spark.comet.exec.pyarrowUDF.enabled=true` and the input is already a Comet columnar batch.

The [current eligibility check on upstream main](https://github.com/apache/datafusion-comet/blob/e13f9d265af0addb81385305d9c6553cdb6f9dfc/spark/src/main/scala/org/apache/comet/rules/EliminateRedundantTransitions.scala#L204) returns `None` unconditionally for this configuration. Users therefore lose the columnar Python execution path and pay the Arrow-to-row-to-Arrow conversion for that stage. Other operators may still use Comet. The fallback applies even to small batches; it is not conditional on a column actually exceeding the ordinary Arrow size limit.

Native Comet input vectors use ordinary string/binary layouts with 32-bit offsets. Serializing those buffers under their matching schema produces valid IPC, but does not supply the large input types requested by Spark. Removing the fallback or changing only the advertised field types would not implement support: the offset buffers must actually match the large types.

A small reproduction was verified on Spark 4.0.4 / JDK 17 / PyArrow 25.0.1 using the Comet jar from PR #5368 at `5fd48b3e`. With the Python acceleration flag enabled and three Parquet rows containing strings, binary values, nulls, and empty values:

| `useLargeVarTypes` | Python operator | Types observed inside the worker |
| --- | --- | --- |
| `false` | `CometMapInBatch` | `string`, `binary` |
| `true` | Spark `MapInArrow`, preceded by `CometColumnarToRow` | `large_string`, `large_binary` |

Both executions returned the same values. This is a missing acceleration capability, not a query correctness failure.

For an existing Comet-enabled Spark 4.x session with a working native Parquet scan:

```python
import tempfile
import pyarrow as pa

spark.conf.set("spark.comet.exec.pyarrowUDF.enabled", "true")
spark.conf.set("spark.sql.execution.arrow.useLargeVarTypes", "true")

def passthrough(batches):
    for batch in batches:
        assert pa.types.is_large_string(batch.schema.field("s").type)
        assert pa.types.is_large_binary(batch.schema.field("b").type)
        yield batch

with tempfile.TemporaryDirectory() as directory:
    path = directory + "/input"
    spark.createDataFrame(
        [(1, "a", bytearray(b"x")), (2, None, None), (3, "", bytearray())],
        "id int, s string, b binary",
    ).coalesce(1).write.parquet(path)
    source = spark.read.parquet(path)
    result = source.mapInArrow(passthrough, source.schema)
    result.explain()
    print(result.collect())
```

Allow eligible `mapInArrow` and `mapInPandas` operations to use `CometMapInBatch` while preserving Spark's requested large Arrow input types. This could widen offsets at the Comet-to-Python boundary or produce suitable vectors upstream; the implementation should preserve source ownership and reuse payload buffers where practical.

Acceptance criteria:

- With both settings enabled, an otherwise eligible plan remains accelerated and Python receives the same large string/binary input types as vanilla Spark.
- Preserve values, nulls, nested strings/binary values, empty batches, and behavior across multiple batches and chained UDFs for both APIs.
- Account for and release conversion allocations on success and failure, without changing the ordinary `useLargeVarTypes=false` fast path.
- Add worker-visible type assertions and plan assertions, measure the conversion cost, and update the documented limitation. Document any remaining size limits in native producers; small-batch type tests alone do not demonstrate support beyond 2 GiB.

### Additional context

This limitation predates #5368, which removes the intermediate input-buffer copy but deliberately retains the fallback.

Related issues have different scopes:

- #5488 concerns serializing large-offset vectors returned by Python for broadcast or collect. That output-side issue does not cover supplying large input vectors to Python.
- #5386 extends acceleration to scalar Python UDFs and mentions this configuration as an existing limitation; it does not track implementing large input types.

sunchao:
> **Draft / merge order:** this feature depends on #5560 (dictionary shuffle input) and #5561 (Spark 4.2 worker framing). After both fixes merge, this branch will be rebased onto `main`, the integrated validation will be rerun, and the PR can return to ready-for-review state.

Closes #5555.

## Rationale for this change

Setting `spark.sql.execution.arrow.useLargeVarTypes=true` currently disables Comet's accelerated `mapInArrow` / `mapInPandas` path, even for a small batch that is already columnar. For example, a Parquet batch containing strings and binary values falls back through `CometColumnarToRow` and Spark's Arrow writer so that Python receives `large_string` and `large_binary`. This preserves Spark's requested types but reintroduces the row conversion that the accelerated path avoids.

This PR keeps eligible operations accelerated while supplying real 64-bit offset buffers. Changing the IPC schema alone would be incorrect because ordinary Comet string and binary vectors have 32-bit offsets.

The Python runner reads the setting from the driver-captured worker configuration and recursively advertises large string and binary fields, including fields inside structs, arrays, and maps. Serialization widens only ordinary string and binary offset buffers; value and validity buffers remain borrowed, already-large offsets are reused, and temporary offsets are released after the synchronous write. Empty inputs also advertise the configured types, and compatible batches with different field names, nullability, or descriptive metadata can share the stream schema.

The planner's unconditional large-type fallback and its now-unused SQLConf shim accessors are removed. Spark 3.x retains its existing unsupported-API fallback. The three shim copies remain separate, with the stale blank lines left by the removed accessor cleaned up.

The widening cleanup always owns temporary offset buffers through one list instead of using a null sentinel, and `serializeBatch` now requires callers to choose the offset mode explicitly. A mixed-layout JVM regression test drives the depth-first buffer walk through `struct<string,int,binary>`, `list<string>`, `map<string,string>`, `list<struct<string>>`, a zero-buffer null vector, fixed-size binary, and a trailing string in both offset modes. It checks values, types, source reference counts, and writer allocator cleanup.

The documentation describes the remaining limit: widening at the Python boundary does **not** remove the 32-bit size limits of native producers that first construct ordinary arrays. This is input support; it does not claim to resolve the separate output serialization issue #5488. The end-to-end benchmark supports `BENCHMARK_LARGE_VAR_TYPES=true` to exercise the new path.

Dictionary decoding and Spark 4.2 command framing are intentionally excluded from this diff and live in #5560 and #5561.

Native code was built before the JVM and Python checks; this PR does not change native code.

- Spark 4.0 / Scala 2.13 root-reactor package build: **21/21 focused JVM tests passed** (15 Arrow runner tests and 6 planner tests). Coverage includes regular and large offsets, nested buffer traversal, already-large offsets, empty batches, cross-allocator FFI ownership, source refcounts, writer cleanup, and partial allocation failure.
- PySpark 4.0.4 / JDK 17 / PyArrow 25.0.1 / pandas 3.0.5: **all 128 real-worker integration tests passed** on this narrow head.
- Spark 3.5 compatibility validation from the feature commit passed **28 tests**, with one expected Spark-4-only case canceled.
- Scala formatting/style, Python syntax/lint for the changed code, suite-registration checks, and `git diff --check` passed.

An isolated JVM harness called the production serializer with 8,192 rows, alternating string and binary columns containing 32-byte values and one null per seven rows. After warmup, it took the median of seven samples while alternating measurement order. The output channel discarded bytes, so these measurements isolate serializer work and exclude Python, transport copies, dictionary decoding, and end-to-end query costs:

| Variable-width columns | Default offsets | Large offsets | Added time per batch | Peak writer allocation, default / large |
| 2 | 3.0 us | 51.8 us | 48.7 us | 1 KiB / 257 KiB |
| 16 | 12.8 us | 398.8 us | 386.0 us | 1 KiB / 2,049 KiB |

Writer allocation returned to zero after serialization. These local measurements describe the cost of offset widening for plain vectors, not a claimed query speedup.

### DIFF
--- a/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
+++ b/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
@@ -112,6 +112,13 @@ private[python] trait CometArrowPythonRunnerBase
       private var arrowWriter: ArrowStreamWriter = _
       private var writeRoot: VectorSchemaRoot = _
       private var streamFields: Seq[Field] = _
+      // This map is captured on the driver; do not resolve SQLConf from the task closure.
+      private val useLargeVarTypes = workerConf
+        .getOrElse(SQLConf.ARROW_EXECUTION_USE_LARGE_VAR_TYPES.key, "false")
+        .toBoolean
+
+      private def inputField(field: Field): Field =
+        if (useLargeVarTypes) CometArrowPythonRunnerBase.withLargeVarTypes(field) else field
 
       // The runner's input schema is a single struct column ("struct") whose children are the
       // user's input columns (see `schema` above). Cast once here rather than at each use site.
@@ -158,7 +165,7 @@ private[python] trait CometArrowPythonRunnerBase
               // no sample batch, so derive the schema from the Spark input schema. The timezone is
               // irrelevant here because no rows are exchanged.
               val childFields = inputStructType.fields.toSeq.map(f =>
-                Utils.toArrowField(f.name, f.dataType, nullable = true, "UTC"))
+                inputField(Utils.toArrowField(f.name, f.dataType, nullable = true, "UTC")))
               startWriter(childFields, dataOut)
             }
             arrowWriter.end()
@@ -176,15 +183,15 @@ private[python] trait CometArrowPythonRunnerBase
             .getValueVector
             .asInstanceOf[FieldVector]
         }
-        val batchFields = sourceVectors.map(_.getField)
+        val batchFields = sourceVectors.map(vector => inputField(vector.getField))
 
         if (arrowWriter == null) {
           // Build the schema-only struct root once from the first batch's child fields.
           // mapInArrow/mapInPandas exchange the columns under a single non-nullable struct.
           // Comet's FFI-imported vectors leave the Arrow Field name null, so restore the real
           // column names from the input schema (the worker reads columns by name, and shaded
-          // Arrow rejects a null field name). Keep the field types and child structure as-is so
-          // the advertised schema matches the source buffers. Keeping the type as-is also means
+          // Arrow rejects a null field name). Apart from optional string/binary offset widening,
+          // keep the field types and child structure consistent with the source buffers. This means
           // a TimestampType reaches the worker with Comet's UTC time zone
           // rather than the session zone vanilla Spark would label it with; this is a documented
           // limitation (see pyarrow-udfs.md), not a value difference, since the stored instant is
@@ -206,7 +213,8 @@ private[python] trait CometArrowPythonRunnerBase
           new WriteChannel(Channels.newChannel(dataOut)),
           sourceVectors,
           cometBatch.numRows(),
-          allocator)
+          allocator,
+          useLargeVarTypes)
 
         pythonMetrics("pythonDataSent") += dataOut.size() - startData
         true
@@ -329,6 +337,24 @@ private[python] trait CometArrowPythonRunnerBase
 
 private[python] object CometArrowPythonRunnerBase {
 
+  /** Match Spark's large string/binary input types, including fields nested in containers. */
+  private[python] def withLargeVarTypes(field: Field): Field = {
+    val fieldType = field.getFieldType
+    val dataType = fieldType.getType match {
+      case ArrowType.Utf8.INSTANCE => ArrowType.LargeUtf8.INSTANCE
+      case ArrowType.Binary.INSTANCE => ArrowType.LargeBinary.INSTANCE
+      case other => other
+    }
+    new Field(
+      field.getName,
+      new FieldType(
+        fieldType.isNullable,
+        dataType,
+        fieldType.getDictionary,
+        fieldType.getMetadata),
+      field.getChildren.asScala.map(withLargeVarTypes).asJava)
+  }
+
   // Extensions can change interpretation even when their underlying storage types match.
   private val extensionMetadataKeys = Seq(
     ArrowType.ExtensionType.EXTENSION_METADATA_KEY_NAME,
@@ -350,7 +376,8 @@ private[python] object CometArrowPythonRunnerBase {
    *
    * VectorUnloader recursively retains the source buffers without moving them between allocators.
    * The wrapping record batch takes its own references, so closing both temporary batches
-   * restores the original reference counts after the synchronous pipe write. The borrowed
+   * restores the original reference counts after the synchronous pipe write. Large variable types
+   * replace only the offset buffers and release those allocations after the write. The borrowed
    * VectorSchemaRoot must never be closed because its vectors are owned by the input
    * ColumnarBatch.
    */
@@ -358,7 +385,8 @@ private[python] object CometArrowPythonRunnerBase {
       writeChannel: WriteChannel,
       sourceVectors: Seq[FieldVector],
       numRows: Int,
-      allocator: BufferAllocator): Unit = {
+      allocator: BufferAllocator,
+      useLargeVarTypes: Boolean = false): Unit = {
     val sourceRoot =
       new VectorSchemaRoot(sourceVectors.map(_.getField).asJava, sourceVectors.asJava, numRows)
     val sourceBatch = new VectorUnloader(sourceRoot).getRecordBatch
@@ -379,17 +407,27 @@ private[python] object CometArrowPythonRunnerBase {
         buffers.add(structValidity)
         buffers.addAll(sourceBatch.getBuffers)
 
-        val wrappedBatch = new ArrowRecordBatch(
-          numRows,
-          nodes,
-          buffers,
-          sourceBatch.getBodyCompression,
-          sourceBatch.getVariadicBufferCounts,
-          true)
+        val widenedOffsets = if (useLargeVarTypes) new ArrayList[ArrowBuf]() else null
         try {
-          MessageSerializer.serialize(writeChannel, wrappedBatch)
+          if (useLargeVarTypes) {
+            widenOffsets(sourceVectors, buffers, widenedOffsets, allocator)
+          }
+          val wrappedBatch = new ArrowRecordBatch(
+            numRows,
+            nodes,
+            buffers,
+            sourceBatch.getBodyCompression,
+            sourceBatch.getVariadicBufferCounts,
+            true)
+          try {
+            MessageSerializer.serialize(writeChannel, wrappedBatch)
+          } finally {
+            wrappedBatch.close()
+          }
         } finally {
-          wrappedBatch.close()
+          if (widenedOffsets != null) {
+            widenedOffsets.asScala.foreach(_.close())
+          }
         }
       } finally {
         structValidity.close()
@@ -397,5 +435,42 @@ private[python] object CometArrowPythonRunnerBase {
     } finally {
       sourceBatch.close()
     }
+  }
+
+  /**
+   * Replace only 32-bit offsets; validity and payload buffers remain borrowed from the source.
+   */
+  private def widenOffsets(
+      sourceVectors: Seq[FieldVector],
+      buffers: ArrayList[ArrowBuf],
+      widenedOffsets: ArrayList[ArrowBuf],
+      allocator: BufferAllocator): Unit = {
+    // Skip the wrapping struct's validity buffer, then follow VectorUnloader's depth-first order.
+    var bufferIndex = 1
+    def visit(vector: FieldVector): Unit = {
+      vector.getField.getType match {
+        case ArrowType.Utf8.INSTANCE | ArrowType.Binary.INSTANCE =>
+          val source = buffers.get(bufferIndex + 1)
+          val count = vector.getValueCount.toLong + 1L
+          val offsets = allocator.buffer(count * 8L)
+          // Register each allocation before populating it so partial conversion failures clean up.
+          widenedOffsets.add(offsets)
+          if (vector.getValueCount == 0 && source.readableBytes() == 0L) {
+            offsets.setLong(0L, 0L)
+          } else {
+            var i = 0L
+            while (i < count) {
+              offsets.setLong(i * 8L, source.getInt(i * 4L).toLong)
+              i += 1L
+            }
+          }
+          offsets.writerIndex(count * 8L)
+          buffers.set(bufferIndex + 1, offsets)
+        case _ =>
+      }
+      bufferIndex += vector.getFieldBuffers.size()
+      vector.getChildrenFromFields.asScala.foreach(visit)
+    }
+    sourceVectors.foreach(visit)
   }
 }
--- a/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
+++ b/spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
@@ -386,7 +386,7 @@ private[python] object CometArrowPythonRunnerBase {
       sourceVectors: Seq[FieldVector],
       numRows: Int,
       allocator: BufferAllocator,
-      useLargeVarTypes: Boolean = false): Unit = {
+      useLargeVarTypes: Boolean): Unit = {
     val sourceRoot =
       new VectorSchemaRoot(sourceVectors.map(_.getField).asJava, sourceVectors.asJava, numRows)
     val sourceBatch = new VectorUnloader(sourceRoot).getRecordBatch
@@ -407,7 +407,7 @@ private[python] object CometArrowPythonRunnerBase {
         buffers.add(structValidity)
         buffers.addAll(sourceBatch.getBuffers)
 
-        val widenedOffsets = if (useLargeVarTypes) new ArrayList[ArrowBuf]() else null
+        val widenedOffsets = new ArrayList[ArrowBuf]()
         try {
           if (useLargeVarTypes) {
             widenOffsets(sourceVectors, buffers, widenedOffsets, allocator)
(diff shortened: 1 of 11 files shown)


spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
```diff
@@ -329,6 +348,55 @@ private[python] trait CometArrowPythonRunnerBase
...
+   * so materialize those columns first. The temporary decoded vectors own their buffers and are
+   * closed after the synchronous write, including schema and serialization failures.
+   */
+  private[python] def withMaterializedInputVectors[T](
+      columns: Seq[CometDecodedVector],
+      allocator: BufferAllocator)(body: Seq[FieldVector] => T): T = {
+    val materialized = new ArrayList[FieldVector]()
+    try {
+      val vectors = columns.map {
+        case dictionaryVector: CometDictionaryVector =>
```
andygrove: This matches on the top-level column, so a dictionary-encoded field nested inside a struct, list or map keeps its `DictionaryEncoding` in the advertised schema while `widenOffsets` skips it as `Int32`. I built that shape by hand and it fails with the same `NullPointerException: ... "provider" is null` you are fixing for the top-level case.

I do not think Comet can produce it today. `make_builders` in `native/shuffle/src/spark_unsafe/row.rs` passes `1.0` for list elements, struct fields and map key/value builders, and `CometVector.getVector` dispatches on the container types before it looks at `getDictionary()`. The one route I can see is a chained UDF whose first worker returns a struct with a dictionary-encoded child, which then becomes the second runner's input, which is the shape your chained-UDF test already exercises with plain types.

Would it be worth either recursing into children here, or adding a `require` that rejects a `getDictionary() != null` field anywhere in the tree, so that it fails with a Comet message instead of an Arrow NPE?

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
```diff
@@ -379,17 +449,27 @@ private[python] object CometArrowPythonRunnerBase {
...
         buffers.addAll(sourceBatch.getBuffers)
 
-        val wrappedBatch = new ArrowRecordBatch(
-          numRows,
-          nodes,
-          buffers,
-          sourceBatch.getBodyCompression,
-          sourceBatch.getVariadicBufferCounts,
-          true)
+        val widenedOffsets = if (useLargeVarTypes) new ArrayList[ArrowBuf]() else null
```
andygrove: Small one: `widenedOffsets` is `null` when `useLargeVarTypes` is false, so the same condition gets tested twice in two different ways, once as `if (useLargeVarTypes)` to call `widenOffsets` and once as `if (widenedOffsets != null)` in the `finally`. Always allocating the list would drop the sentinel and the null check for the cost of one empty `ArrayList` per batch, on a path that already allocates an offset buffer per variable-width column when the flag is on.

Related, the `useLargeVarTypes: Boolean = false` default on `serializeBatch` is only reached from tests, since production always passes it explicitly.

spark/src/test/resources/pyspark/benchmark_pyarrow_udf.py
```diff
@@ -84,6 +86,10 @@ def _build_spark() -> SparkSession:
         .config("spark.plugins", "org.apache.spark.CometPlugin")
         .config("spark.comet.enabled", "true")
         .config("spark.comet.exec.enabled", "true")
+        .config(
+            "spark.sql.execution.arrow.useLargeVarTypes",
+            os.environ.get("BENCHMARK_LARGE_VAR_TYPES", "false"),
+        )
```
andygrove, line 92: `BENCHMARK_LARGE_VAR_TYPES` covers the widening path nicely. Could the dictionary path get a number too? It looks like the more expensive of the two new costs, and unlike widening it applies in the default `useLargeVarTypes=false` mode as well.

On a single 8192-row string column with 64 distinct 32-byte values I measured the decode at roughly 350 us per batch against roughly 13 us for the whole plain serialization step, so about 25x to 30x. I also tried replacing `DictionaryEncoder.decode` with a direct offset-driven copy to see whether Arrow's per-row `getObject` boxing was the problem, and it came out slightly slower (398 us versus 356 us). So the cost looks inherent to materializing the values rather than something to micro-optimize, which seems worth knowing before anyone tries.

The gap I would most like closed is that nothing in the repo would catch a regression here. The three `WORKLOADS` all read straight from Parquet with no shuffle in the plan, so none of them ever sees a `CometDictionaryVector`. Would you add a fourth that repartitions on a low-cardinality string column with `spark.comet.shuffle.mode=jvm` and `preferDictionary.ratio` set, and quote the accelerated-versus-vanilla number for it? If the accelerated path still wins end to end for dictionary input, that is the number that justifies removing the fallback, and it belongs in the description next to the widening table. Note `_mixed_with_strings` builds `concat('row_', id)`, which is all-distinct, so it would need a low-cardinality column to be a useful dictionary case even with a shuffle added.

spark/src/test/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerSuite.scala
```diff
@@ -265,6 +285,317 @@ class CometArrowPythonRunnerSuite extends AnyFunSuite with Matchers {
     }
   }
 
+  for {
+    failSerialization <- Seq(false, true)
+    useLargeVarTypes <- Seq(false, true)
+  } {
+    test(
+      "dictionary inputs materialize logical values " +
```
andygrove: Two coverage notes on this suite. Both cases pass for me, so this is about where the guard lives rather than a bug.

These dictionary tests build their vectors from a plain `RootAllocator`, so the reference manager under test is not the `ReferenceCountedArrowArray` that real shuffle input carries. That is the same point that came up on #5368 for the plain path, which you then covered with the `Data.exportVector` cases. I drove real Comet JVM-shuffle batches through it (`shuffle.mode=jvm`, `preferDictionary.ratio=1.01`, `repartition(2, "id")`), saw four genuine `CometDictionaryVector` columns, and the decoded values matched `df.collect()` in both offset modes. Worth one case, since that is the shape production actually produces.

Separately, the four new large-type tests all use flat top-level vectors, and `direct batches preserve nested list, struct, map, and null field layouts` further down is not parameterized on `useLargeVarTypes`. The `bufferIndex` walk in `widenOffsets` is the part of this change where a slip would silently corrupt a later column instead of failing, and the pytest suite that does cover nesting is not run by the per-Spark matrix jobs. Would you parameterize that nested test the way you parameterized these? The arrangement that would catch drift is a single batch holding `struct<utf8,int32,binary>`, `list<utf8>`, `map<utf8,utf8>`, `list<struct<utf8>>`, a `NullVector` (zero field buffers), a `FixedSizeBinaryVector` (two buffers, must not widen) and a trailing top-level `utf8`. That round-trips exactly for me today.

andygrove (commented):
Thanks for this. I checked the PR out locally, built it, ran its suites (24/24 pass on Spark 4.0 / Scala 2.13, matching your 18 + 6), and then wrote eight probes of my own to test the parts I could not run here. PyPI is blocked on my machine so the pytest suite could not run, and I substituted JVM-level probes that drive real Comet batches instead.

The widening implementation holds up. I verified the `bufferIndex` walk in `widenOffsets` against `VectorUnloader`'s depth-first order and Arrow's per-type buffer counts, then tested it with a batch holding `struct<utf8,int32,binary>`, `list<utf8>`, `map<utf8,utf8>`, `list<struct<utf8>>`, a `NullVector` (zero field buffers), a `FixedSizeBinaryVector` (two buffers, correctly skipped) and a trailing top-level `utf8`, which is the arrangement where any index drift would corrupt a later column. Exact round-trip, writer allocator back to zero, source refcounts balanced, source vectors unmutated. I also drove real FFI-imported Comet scan batches at `spark.comet.batchSize=7` and real Comet JVM-shuffle dictionary batches through it, and both match `df.collect()` byte for byte in either offset mode. And I confirmed the Spark 4.2 change against `spark-core_2.13-4.2.0` bytecode: `BasePythonRunner$Writer.open` writes `runnerConf` then `evalConf` via `PythonWorkerUtils.writeConf` and only then calls `writeCommand`, neither method exists on 4.0.4 or 4.1.3, and your override is byte-for-byte what upstream 4.2's own `ArrowPythonRunner` does.

My main comment is about the shape of the PR rather than the code.

This reads as three changes rather than one, and I think two of them would land better on their own.

The dictionary decode is a fix for a crash that already exists on `main`, independent of `useLargeVarTypes`. I reproduced it: `startWriter` advertises the indices field with its `DictionaryEncoding` intact but passes `null` as the `DictionaryProvider`, so `ArrowStreamWriter.start()` throws `NullPointerException: Cannot invoke "DictionaryProvider.lookup(long)" because "provider" is null` from `DictionaryUtility.toMessageFormat`. It reproduces with `CometShuffleManager` plus `spark.comet.shuffle.mode=jvm` plus `preferDictionary.ratio=1.01`, which is the config your new `test_pyarrow_udf_dictionary_shuffle.py` uses. `branch-1.0` looks affected the same way, since it has the same `ArrowStreamWriter(writeRoot, null, ...)` and derives its fields from `getValueVector.getField` too.

The Spark 4.2 command framing is also a standalone fix. On `main` the 4.2 worker gets an extra length-prefixed map after the base runner has already written both conf maps, and it reads that as the UDF count. This went unnoticed because the PyArrow matrix had no 4.2 real-worker job until this PR adds one, which is a good catch on its own.

Would you consider landing those two as separate fix PRs first, each with its regression test, and rebasing the feature on top? It would shrink this one to the widening work plus its tests and docs, and it would give the two crashes their own bisect points and backport candidacy. The feature depends on the dictionary fix anyway, since removing the planner fallback would otherwise expose the same NPE to `useLargeVarTypes=true` users. I realise both are opt-in behind an experimental flag, so this is about reviewability and bisectability rather than urgency.

One small thing that does not deserve its own thread: with `arrowUseLargeVarTypes` gone, the `spark-3.5` and `spark-4.x` copies of `ShimSQLConf` are now byte-identical and the `spark-3.4` one differs only in the `LegacyBehaviorPolicy` import path. Genuinely collapsing them looks like more trouble than it is worth given `shims.minorPlusVerSrc` has a single slot already used by `spark-4.1+`, so I would leave the structure alone. All three do keep a blank line between the last `val` and the closing brace where the deleted method used to be, which is worth removing while you are in there.

Nothing here is blocking from my side. The correctness work is solid and better tested than most changes this size.

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
sunchao: I split the dictionary-input fix into #5560. I kept that fix scoped to the top-level `CometDictionaryVector` shape that the JVM shuffle produces today: nested list, struct, and map builders explicitly disable dictionary encoding, and the container import paths do not surface a dictionary-backed child to this runner.

Given that there is no production producer for the hand-built nested shape, adding recursive materialization here would introduce ownership logic with no end-to-end path to test. #5560 covers the reachable JVM-shuffle shape with real-worker tests and allocator/refcount checks. If nested dictionaries become producible, I agree they should be handled recursively with a provider-aware test at that boundary rather than allowed to reach Arrow's null-provider failure. The dictionary code is now outside this feature PR.

spark/src/main/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerBase.scala
sunchao: Done on the narrowed feature branch in `1876303b9`. `widenedOffsets` is now always an `ArrayList`, cleanup no longer uses a null sentinel, and `serializeBatch` requires the Boolean at every call site. The final narrow head passed 21/21 focused JVM tests and all 128 real-worker cases.

spark/src/test/resources/pyspark/benchmark_pyarrow_udf.py
sunchao: Moved with the dictionary fix to #5560. That PR adds a `dictionary JVM shuffle` workload with low-cardinality strings, forces the JVM shuffle and dictionary ratio, requires `CometColumnarExchange`, and supports selecting just that workload.

On my local Spark 4.0.4 `local[2]` debug build with 2,000,000 rows, two warmups, and five measured iterations, median end-to-end time was 2.453 s vanilla / 1.794 s accelerated (1.37x) for `mapInArrow`, and 2.415 s / 1.797 s (1.34x) for `mapInPandas`. I put the exact scope and numbers in #5560's description rather than carrying dictionary performance material in this feature PR.

spark/src/test/spark-4.x/org/apache/spark/sql/execution/python/CometArrowPythonRunnerSuite.scala
sunchao: Both guards now live with the code they cover. #5560 owns a real JVM-shuffle regression module for `mapInArrow` and `mapInPandas`; the workflow runs it in a separate process for each Spark 4.0/4.1 worker so it gets the Comet shuffle manager instead of inheriting the main module's shuffle-disabled session.

This feature PR now has the mixed-layout JVM test on its narrow head. It runs regular and large modes through `struct<string,int,binary>`, `list<string>`, `map<string,string>`, `list<struct<string>>`, `NullVector`, `FixedSizeBinaryVector`, and a trailing string, with exact values/types, recursive source refcounts, and writer-allocation cleanup. The focused feature suites pass 21/21.

sunchao:
Thanks for the careful review. I completed the requested split:

- #5560 contains the existing-main dictionary shuffle crash fix, its ownership and real-worker regressions, and the low-cardinality shuffle benchmark.
- #5561 contains only the Spark 4.2 worker framing fix and the 4.2 real-worker CI row.
- This PR is now an 11-file large-offset-only diff. It also includes the small offset-list cleanup, explicit serializer mode, mixed nested-buffer regression, and shim whitespace cleanup from this review.

I converted this PR to draft and marked both fixes as prerequisites. Once #5560 and #5561 merge, I will rebase this branch onto `main`, rerun the integrated validation, and mark it ready again. The current narrow head passes 21/21 focused JVM tests and all 128 Spark 4.0 real-worker cases.

PR #5562, opened 2026-08-30
chore: remove redundant condition and stray println in shuffle code

viirya:

N/A -- trivial cleanup, no issue filed.

## Rationale for this change

Two small pieces of dead/leftover code noticed while reading through the shuffle
type gates. Neither has any behavioral effect; removing them makes the
surrounding code slightly easier to read.

- `CometShuffleExchangeExec`: in the columnar shuffle's
  `supportedSerializableDataType`, the `StructType` case tested
  `fields.nonEmpty` twice -- once at the start of the conjunction and again as
  the final term. The trailing check is dead:

  ```scala
  case StructType(fields) =>
    fields.nonEmpty && fields.forall(f => supportedSerializableDataType(f.dataType)) &&
    // Java Arrow stream reader cannot work on duplicate field name
    fields.map(f => f.name).distinct.length == fields.length &&
    fields.nonEmpty   // <-- removed
  ```

  It appears to have been introduced accidentally in bf18cec52 ("feat:
  Implement TimeType support - Infrastructure - shuffle").

- `CometNativeShuffleSuite`: remove a leftover debugging
  `println(shuffled.queryExecution.executedPlan)` from the
  "native shuffle with Map[_, NullType] column" test.

Net change is `1 insertion(+), 3 deletions(-)`.

Covered by existing tests -- no behavior changes, so no new tests are added.

The first change is a boolean identity (`A && B && C && A` where `A` is the
leading term, so the trailing `A` cannot change the result); the second only
removes output to stdout from a test.

Verified locally that `mvn spotless:apply` reports no formatting changes and
that `mvn test-compile -pl spark -am` succeeds. I did not run
`CometNativeShuffleSuite` to a green state locally, since my working copy has
no `libcomet` built and the whole suite falls back without it (the same 11
passed / 25 failed before and after this change, so the failures are unrelated
to it). Relying on CI for the full suite run.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometShuffleExchangeExec.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometShuffleExchangeExec.scala
@@ -551,8 +551,7 @@ object CometShuffleExchangeExec
       case StructType(fields) =>
         fields.nonEmpty && fields.forall(f => supportedSerializableDataType(f.dataType)) &&
         // Java Arrow stream reader cannot work on duplicate field name
-        fields.map(f => f.name).distinct.length == fields.length &&
-        fields.nonEmpty
+        fields.map(f => f.name).distinct.length == fields.length
       case ArrayType(elementType, _) =>
         supportedSerializableDataType(elementType)
       case MapType(keyType, valueType, _) =>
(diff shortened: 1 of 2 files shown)


sunchao (approved):
Reviewed 5c1202ce5e83. No blocking findings.

andygrove (approved):

viirya:
Thanks @sunchao @andygrove

PR #5564, opened 2026-08-30
test: cover struct data columns in native shuffle

### LINKED ISSUE #5563: Native shuffle has no test coverage for struct data columns

`CometShuffleExchangeExec.supportedSerializableDataType` (the native shuffle
data-column type gate) allows struct data columns, recursing through
`StructType` / `ArrayType` / `MapType`:

```scala
case StructType(fields) =>
  fields.nonEmpty && fields.forall(f => supportedSerializableDataType(f.dataType))
case ArrayType(elementType, _) =>
  supportedSerializableDataType(elementType)
case MapType(keyType, valueType, _) =>
  supportedSerializableDataType(keyType) && supportedSerializableDataType(valueType)
```

But `CometNativeShuffleSuite` never exercised a struct data column. The only
nested coverage in that suite is:

- `native shuffle on nested array` -- `array<array<int>>`
- `native shuffle with Map[_, NullType] column` -- `map<int, null>`

The word `struct` does not appear anywhere in the suite. By contrast,
`CometColumnarShuffleSuite` has extensive nested coverage (struct with nulls,
array/struct as map key/value, `array<map>`, many map value types), so the gap
is specific to native shuffle.

This matters because nested types in shuffle have been a real source of bugs --
e.g. cd4d0e25d ("fix: normalize nested field nullability in ShuffleScanExec and
ExpandExec", #5137), where Arrow treats nested field nullability as part of
`DataType` identity and the decoded shuffle block had to be reconciled against
the catalyst-declared schema. A behavior the gate explicitly claims to support,
with no end-to-end test, can regress silently.

Note that struct columns can only reach native shuffle as *data* columns:
`supportedHashPartitioningDataType` rejects nested types as a hash key, so a
query partitioning on a struct falls back to Spark. Tests therefore need to
partition on a primitive and carry the struct along.

## Describe the solution you'd like

Add native shuffle tests covering struct data columns, including nulls, the
recursive nesting branches of the type gate (`struct<array<...>>`,
`array<struct<...>>`), and a struct containing a map field.

## Describe alternatives you've considered

None -- this is purely additional test coverage for existing, already-supported
behavior. No product code change is proposed.

## Additional context

Verified by mutation testing: temporarily changing the native gate's
`case StructType(...)` to `false` makes exactly the new tests fail and leaves
the other 36 tests in the suite passing, confirming they assert native struct
shuffle rather than passing vacuously via a silent fallback to Spark.

viirya:

Closes #5563.

## Rationale for this change

`CometShuffleExchangeExec.supportedSerializableDataType` allows struct data
columns in native shuffle, but `CometNativeShuffleSuite` never exercised one --
the only nested coverage was `array<array<int>>` and `map<int, null>`, and the
word `struct` did not appear in the suite at all. `CometColumnarShuffleSuite`
has extensive nested coverage, so the gap was specific to native shuffle.

Nested types in shuffle have been a real source of bugs (e.g. cd4d0e25d, #5137,
on nested field nullability in `ShuffleScanExec`), so a behavior the type gate
explicitly claims to support deserves an end-to-end test rather than relying on
it not regressing.

Test-only. Four tests added to `CometNativeShuffleSuite`:

- `native shuffle on struct data column` -- a plain `struct<int, string>`
- `native shuffle on struct data column including nulls` -- a null struct and a
  null field inside a struct
- `native shuffle on deeply nested data columns` -- `struct<array<int>, string>`
  and `array<struct<int, string>>`, covering the recursive branches of the gate
- `native shuffle on struct data column with map field`

All of them partition on a primitive column and carry the struct as a data
column, because `supportedHashPartitioningDataType` rejects nested types as a
hash key -- a struct can only reach native shuffle as data. The first test
carries a comment explaining this, so the arrangement is not mistaken for an
oversight.

No product code is changed.

This PR is the tests. `CometNativeShuffleSuite` passes locally: 40 tests,
40 succeeded, 0 failed (36 pre-existing + 4 new).

I also mutation-tested them to confirm they are not vacuous: temporarily
changing the native gate's `case StructType(...)` to `false` makes exactly these
4 tests fail while the other 36 keep passing, so they really do assert that
struct data columns go through native shuffle, rather than passing because the
plan silently fell back to Spark.

### DIFF
--- a/spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
@@ -616,6 +616,63 @@ class CometNativeShuffleSuite extends CometTestBase with AdaptiveSparkPlanHelper
     }
   }
 
+  test("native shuffle on struct data column") {
+    // The native shuffle type gate (CometShuffleExchangeExec.supportedSerializableDataType)
+    // allows struct data columns, but nothing exercised them: struct columns can only reach
+    // native shuffle as DATA, since supportedHashPartitioningDataType rejects them as a
+    // hash key. So partition on a primitive and carry the struct along.
+    Seq(10, 201).foreach { numPartitions =>
+      withParquetTable((0 until 50).map(i => (i, (i + 1, (i + 2).toString), i + 3)), "tbl") {
+        val df = sql("SELECT * FROM tbl")
+          .filter($"_3" > 10)
+          .repartition(numPartitions, $"_1")
+          .sortWithinPartitions($"_1")
+
+        checkShuffleAnswer(df, 1)
+      }
+    }
+  }
+
+  test("native shuffle on struct data column including nulls") {
+    Seq(10, 201).foreach { numPartitions =>
+      val data: Seq[(Int, (Int, String))] =
+        Seq((1, (0, "1")), (2, (3, "3")), (3, null), (4, (5, null)))
+      withParquetTable(data, "tbl") {
+        val df = sql("SELECT * FROM tbl")
+          .repartition(numPartitions, $"_1")
+          .sortWithinPartitions($"_1")
+
+        checkShuffleAnswer(df, 1)
+      }
+    }
+  }
+
+  test("native shuffle on deeply nested data columns") {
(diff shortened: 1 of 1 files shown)


sunchao (approved):

viirya:
Thanks @sunchao

Issue #5446, opened 2026-08-24
perf: verify shuffle IPC context-reuse allocation savings with Arrow 59.1+


peterxcli:
### What is the problem the feature request solves?

PR #5038 reuses the Arrow IPC write context across shuffle blocks. On the original Arrow 58.4.0 baseline, `CompressionContext` owned only an optional zstd compressor; the `FlatBufferBuilder` was still allocated separately for every batch. Since Comet did not enable Arrow IPC buffer compression, that lifetime change alone did not provide metadata-allocation savings.

Upstream `main` uses **DataFusion 55.0.0 and Arrow/Parquet 59.2.0** through merged [#5262](https://github.com/apache/datafusion-comet/pull/5262). The dependency upgrade is no longer a blocker. The upgrade also migrates the shuffle writer paths to `IpcWriteContext`. Arrow 59.2.0's [write context owns a reusable FlatBufferBuilder](https://github.com/apache/arrow-rs/blob/59.2.0/arrow-ipc/src/compression.rs#L29-L43), so the deferred optimization can now be measured on main.

This issue tracks validating the shuffle performance benefit, not another dependency upgrade.

### Describe the potential solution

Completion checklist:

- [x] Comet uses Arrow 59.1 or newer through a compatible DataFusion upgrade (#5262).
- [x] Shuffle writer paths use `IpcWriteContext`, retaining the reusable context lifetimes introduced by #5038.
- [ ] Record focused, repeated same-machine measurements demonstrating fewer IPC metadata allocations on the production shuffle paths.
- [ ] Validate the measured implementation with the shuffle IPC round-trip and writer tests.

Comet's outer per-block compression encoders are independent and outside this issue's scope.

### Additional context

- Follow-up comment: https://github.com/apache/datafusion-comet/pull/5038#issuecomment-5387092772
- Context-reuse implementation: #5038
- Historical benchmark evidence: https://github.com/apache/datafusion-comet/pull/5038#issuecomment-5082906359
- Original implementation issue: #5008
- DataFusion 55 tracker: #4865
- Merged dependency upgrade: #5262

peterxcli:
take

comphead:
Thanks @peterxcli love to investigate the results

PR #5565, opened 2026-08-31
perf: reuse zstd compression contexts across shuffle blocks

dwsmith1983:

Part of #5002 (the compression-context reuse item; the issue stays open for its remaining items).

## Rationale for this change

Every shuffle block written through the local shuffle path creates and destroys its own zstd context: a fresh `CCtx` per encoded block in `ShuffleBlockWriter`. Context setup is pure overhead that scales with block count, so high-partition shuffles with small blocks pay the most. The saving is a few microseconds per block, which is noise at Spark's default 200 partitions and a measurable few percent of encode time at 10,000.

- New `ShuffleCodecContext` (`native/shuffle/src/codec_context.rs`) wrapping a lazily created `zstd_safe::CCtx`, reused via `Encoder::with_context`. The session is reset and the configured level applied on every frame, since `with_context` does not initialize the level and a failed encode must not poison the next block.
- Contexts are task-scoped, never per output partition: `LocalPartitionWriter` owns the context and the per-partition `BufBatchWriter`s and `SpillWriter` borrow it; `RssPartitionWriter` is already one per task. `write_burst_complete` releases the workspace at spill and finish boundaries.
- A retained-size cap (8 MiB, checked against the measured `CCtx` sizes per level, with a test that fails if a zstd bump moves level 8 across it) bounds what a task keeps between bursts: zstd's session reset preserves the allocated window, so a context that once saw a wide-window frame would otherwise stay that large.
- The RSS path frees the zstd workspace at the end of each admitted encode, success or error: its memory accounting charges the workspace per admitted invocation and releases it afterward, so the context must not outlive that window. The remote path therefore keeps per-block context creation as on main; extending reuse there needs the pusher's reserve/push/release contract changed on both sides.
- The decode side is unchanged from main. An earlier revision reused the decoder context too; it measured throughput neutral and introduced retention main does not have, so it was dropped after review.
- Wire format is unchanged. lz4 and snappy keep per-block encoders; the pinned crates expose no reset for them and their setup cost is far below zstd's workspace.

Benchmarks with `shuffle_bench` (4M-row hash shuffle, single task, M-series macOS, release builds). The first table is the earlier three-iteration run at the head of that revision against its merge base; the second is a five-iteration rerun of the trimmed head against current main on a different generated input, so only base-versus-head within a table is meaningful.

| Shape | base | this PR | encode base -> PR |
|---|---|---|---|
| 2,000 partitions, zstd level 1 | 0.323s | 0.328s | 0.188s -> 0.187s |
| 2,000 partitions, zstd level 6 | 0.743s | 0.746s | 0.606s -> 0.610s |
| 10,000 partitions, zstd level 3 | 0.581s | 0.557s | 0.326s -> 0.297s |

| 10,000 partitions, zstd level 3, 5 iterations | base | this PR | delta |
|---|---|---|---|
| wall, average | 1.202s | 1.135s | -5.6% |
| wall, min / max | 1.179s / 1.228s | 1.113s / 1.171s | ranges do not overlap |
| encode, last iteration | 0.653s | 0.610s | -6.6% |

The wall delta is about the size of the run-to-run spread, so encode time is the more direct signal. Large-block shapes are compression-bound and within noise.

On Linux (arm64, Docker, rust 1.94, four CPUs pinned), same shape and input, ten timed iterations per run with base and head alternated over two rounds: round one gives 1.349s to 1.309s wall (3.0 percent) and 0.713s to 0.672s encode (6 percent), with the min/max ranges not overlapping. The second round's base run drifted slower than every other run, so only the direction is taken from it.

Tests in the shuffle crate (130 passing) and the core crate (266 passing, 4 ignored):

- reuse across blocks, writers, and codecs round-trips and every block decodes independently
- two writers with different zstd levels sharing one context each keep their own level
- a mid-frame write failure does not poison the context for the next block
- the retained-size cap releases an oversized context after an encode, with the level-8 boundary sentinel
- the RSS release-versus-local-retain contract is pinned via a test accessor
- spill and finish boundaries release the workspace, including on the error path

Scala on Spark 3.5: CometNativeShuffleSuite, CometShuffleSuite, and CometCelebornShuffleReaderSuite, 150 passing. `cargo clippy --all-targets -- -D warnings` and `cargo fmt` clean.


native/shuffle/src/ipc.rs
```diff
@@ -15,23 +15,52 @@
...
 use arrow::ipc::reader::StreamReader;
 use datafusion::common::DataFusionError;
 use datafusion::error::Result;
+use std::cell::RefCell;
 use std::io::{Error, ErrorKind, Read};
 
+thread_local! {
+    /// Backs the entry points below. They're called from many JVM task threads; a
+    /// thread-local gets each thread context reuse without changing any caller.
+    static DECODE_CONTEXT: RefCell<ShuffleDecodeContext> =
```
sunchao: **[P2]** Could we move this decode context under reader or task ownership, or release it when it exceeds a bounded size? `ResetDirective::SessionOnly` preserves zstd's allocated window. With the locked zstd 1.5.7 build and the same context sequence used here, a valid 17-byte level-22 frame made `DCtx::sizeof()` grow from 95,992 to 134,707,000 bytes, and reset left it at that size. Both production entry points use this thread-local context, so executor worker threads retain the native allocation across tasks. At 32 threads that is about 4 GiB. The base path dropped the decoder per frame. It might be worth adding a regression that verifies the workspace is released when a reader closes and after a decode failure.

native/shuffle/src/writers/local/local_partition_writer.rs
```diff
@@ -64,6 +65,9 @@ enum DataOutput {
 pub(crate) struct LocalPartitionWriter {
     output_index_file: String,
     data_output: DataOutput,
+    /// Compression state shared by every block this task writes; the per-partition
+    /// `BufBatchWriter`s borrow it (see [`ShuffleCodecContext`]).
+    codec_context: ShuffleCodecContext,
```
sunchao, line 82: **[P2]** Could we account for this context's retained workspace for the full local-writer lifetime, or release it at a spill or idle boundary? `LocalPartitionWriter` has no memory reservation for the `CCtx`, while the repartitioner frees its tracked reservation after spilling. With the locked zstd build, the same `Encoder::with_context` path retained 72,082,969 bytes at level 15 and 874,070,679 bytes at level 22 after `SessionOnly` reset. Those levels are accepted by the current configuration. Concurrent tasks can therefore keep large native allocations after the pool reports their buffered memory as released. A size-based regression would also catch this because the current test only checks that the context is present.

sunchao (commented):
Reviewed exact head `8da6e6b445d8801b9e915c59c778cb7034851a70` against base `918bc7d7ac4123a02ef4bba8ddae920f76db9cbc` with five independent specialist scopes.

I found two P2 native-memory lifetime regressions that should be addressed before merge:

- The thread-local decode context can retain a 128 MiB zstd workspace per executor worker thread across tasks.
- The local writer can retain up to an 834 MiB zstd compression workspace per active task outside DataFusion memory accounting.

The RSS release path is correct on success and checked error paths. I found no wire-format, row-correctness, or Rust API compatibility defect.

CI snapshot during this review: 36 passed, 27 pending, 7 skipped, and no failures. `PR Benchmark Check` is skipped.

Local validation:

- `datafusion-comet-shuffle` library: 98 passed
- `datafusion-comet` core library: 201 passed, 4 ignored
- Focused IPC tests: 9 passed
- Focused shuffle-scan tests: 7 passed
- Codec-context and multi-partition spill tests passed
- Exact-version zstd workspace reproducer passed
- `git diff --check` passed and the worktree remained clean

Validation limits: I did not run a Spark/Celeborn end-to-end workload or reproduce the author's M-series performance numbers. The remaining CI jobs were still running when this review was submitted.

native/shuffle/src/ipc.rs
dwsmith1983: Good catch — I had not realized SessionOnly keeps the window allocation. Went with the bounded-size option since the thread-local has no close hook: after every decode, error paths included, the context is dropped if its measured size exceeds 8 MiB (common levels sit at ~1-5 MiB, so they keep reuse; a wide-window frame pays per-frame creation like before). Added the regressions you suggested — one decodes a wide-window level-22 frame and asserts the workspace is released, one does the same through a decode failure.

native/shuffle/src/writers/local/local_partition_writer.rs
dwsmith1983: Same treatment here, plus boundary releases: the writer drops its context when each spill event completes and after the final flush, and any block that leaves the context above 8 MiB drops it immediately. So retained memory between phases is zero and the worst case anywhere is the cap, matching the base path profile rather than adding a new reservation surface. The size-based regression drives a real repartitioner through spill and finish and checks the context is gone at both points; level 22 locally measures ~834 MB retained without the cap, which lines up with your numbers.

native/shuffle/src/ipc.rs
```diff
@@ -15,23 +15,65 @@
...
 
+use crate::codec_context::ShuffleDecodeContext;
 use arrow::array::RecordBatch;
 use arrow::ipc::reader::StreamReader;
 use datafusion::common::DataFusionError;
 use datafusion::error::Result;
+use std::cell::RefCell;
 use std::io::{Error, ErrorKind, Read};
 
+thread_local! {
```
andygrove: The 8 MiB cap turns the unbounded retention into a bounded one, which is a good fix, but the memory is still held for the life of the thread and still is not visible to any reservation. Both production callers go through the thread-local: the static JNI `decodeShuffleBlock` and `ShuffleScanStream`. Those run on JVM task threads and tokio workers, all of which live as long as the executor. A 16 core executor that decodes one zstd shuffle block ends up sitting on roughly 128 MiB of native memory for the rest of its life, including during stages that never shuffle.

`ShuffleScanStream` looks like a natural owner here. Could `decode_shuffle_batch` take a `&mut ShuffleDecodeContext` held by the stream and go through `read_ipc_compressed_with`? That would bound retention to the operator rather than the thread, and it would leave the thread-local for `Java_org_apache_comet_Native_decodeShuffleBlock`, which really has no handle to hang a context off. Right now `read_ipc_compressed_with` and `read_ipc_compressed_validated_with` are exported from `lib.rs` but only ever called from tests, so this would also give them a real caller.

native/shuffle/src/codec_context.rs
```diff
@@ -0,0 +1,140 @@
...
+// under the License.
+
+use arrow::ipc::writer::CompressionContext;
+use std::io;
+use zstd::zstd_safe::{CCtx, CParameter, DCtx, ResetDirective};
+
+/// Largest zstd workspace worth caching between frames. Covers the commonly configured
+/// levels; higher levels (tens to hundreds of MiB of window) fall back to a fresh context
+/// per frame, which is what per-block encoding paid anyway.
+const MAX_RETAINED_ZSTD_CONTEXT_BYTES: usize = 8 * 1024 * 1024;
```
andygrove, line 42: I measured `CCtx::sizeof()` against the pinned zstd 1.5.7 build after one streaming frame and the cap is closer to the edge than the comment suggests. Levels 7 and 8 come in at 8,119,825 bytes against a cap of 8,388,608, so about 3% of headroom. Levels 1 through 6 are 1.31 to 5.24 MiB and level 9 jumps to 14.74 MiB, so anything at 9 or above never reuses at all. On the decode side a level 19 frame leaves the `DCtx` at 8.47 MiB, which also misses the cap.

Two things that would help. Could the measured level to size table go in the comment next to the constant, so the choice of 8 MiB is traceable and someone bumping `zstd-sys` can see what they are moving? And could a test pin where the boundary actually falls, say level 6 retains and level 9 does not? As it stands a routine dependency bump could push levels 7 and 8 over the line and silently disable the optimization for those users with every test still passing.

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -269,9 +273,52 @@ impl ShuffleBlockWriter {
         // write header
         output.write_all(&self.header_bytes)?;
 
+        let encode_result =
```
andygrove, line 276: This is right given that `rss_codec_workspace` is charged per admitted invocation, and I checked that it fires on the error paths as well as the success path. The consequence though is that RSS allocates and frees a `CCtx` per block exactly as it does on main, so the remote path gets none of the benefit this PR is after. Small frames pushed to Celeborn are arguably the shape where per-block context setup hurts most.

Was reserving the workspace once for the lifetime of the `RssPartitionWriter` rather than per invocation considered? There is already one writer per task, so the accounting would be a single up-front charge instead of a repeated one. If that turns out to be awkward against the pusher's admission model it would be worth saying so in the description, which currently reads as though both paths benefit.

native/shuffle/src/shuffle_writer.rs
```diff
@@ -577,6 +566,56 @@ mod test {
         repartitioner.insert_batch(batch.clone()).await.unwrap();
     }
 
+    /// The zstd context is reused within one encode burst but must not survive past it: a
+    /// spill event and the final shuffle write each end with the context released.
+    #[tokio::test]
+    #[cfg_attr(miri, ignore)] // miri can't call foreign function `ZSTD_createCCtx`
+    async fn local_writer_releases_zstd_context_at_burst_boundaries() {
```
andygrove, line 589: This test only ever asserts `!holds_zstd_cctx()`, so it would pass just as happily if the context were never created in the first place. That is true of the suite generally. The tests establish that blocks round-trip correctly under a shared context and that release happens at the right boundaries, but nothing observes that N blocks produce fewer than N context creations, which is the actual claim of the PR.

Would a test-only creation counter on `ShuffleCodecContext` work? Asserting that one spill burst over two partitions creates exactly one context would pin the reuse behaviour directly, and the same counter would let you pin the level boundary from my comment on `codec_context.rs`.

andygrove (commented):
Thanks for taking sunchao's feedback on board. The release-at-boundary handling reads carefully and I like that the error paths are covered.

I checked the pinned zstd 0.13.3 / zstd-safe 7.2.4 / zstd-sys 2.0.16+zstd.1.5.7 sources and `Encoder::with_context` / `Decoder::with_context` really are drop-ins for `Encoder::new` / `Decoder::with_buffer`. Both land in the same raw + zio path, and the only extra work the owned constructors do is setting the compression level (which `zstd_cctx` replicates) plus `DCtx::init` and `load_dictionary(&[])`, which are no-ops when no dictionary is ever set. So I have no concerns about the wire format claim.

I also measured `CCtx::sizeof()` and `DCtx::sizeof()` against that exact zstd build after one streaming frame, since most of my comments turn on those numbers:

| encode level | `CCtx::sizeof()` |
| --- | --- |
| 1 / 3 / 6 | 1.31 / 3.49 / 5.24 MiB |
| 7 / 8 | 7.74 MiB |
| 9 / 12 / 15 | 14.74 / 44.74 / 68.74 MiB |
| 22 | 833.58 MiB |

On the decode side a level 19 frame leaves the `DCtx` at 8.47 MiB and level 22 at 128.47 MiB, while everything up to level 15 stays at or under 4.47 MiB.

One point that is not tied to a line. The benchmark table in the description was measured at 8da6e6b, before abf5807 added the per-block `sizeof()` check and the release at every spill boundary. Levels 1, 3 and 6 all sit well under the 8 MiB cap so I would expect the gains to survive, but the 10,000 partition level 3 case is both the headline result and the one where the new per-spill release actually lands. Could you re-run against 9afe4eb and update the table? It would be good for whoever merges this to be judging the numbers the code actually produces.

native/shuffle/src/ipc.rs
dwsmith1983: Done — the scan operator owns the decode context now (on the exec rather than the stream since the decode loop runs exec-side, where JNI calls are allowed; retention still dies with the operator). The thread-local is down to one production caller, the static decodeShuffleBlock entry, and the _with variants have a real caller.

native/shuffle/src/codec_context.rs
dwsmith1983: Measured the full table against the pinned build and got the same numbers you did — it is in the comment next to the constant now, with the zstd-sys version. Boundary tests pin levels 6 and 8 retained (8 being the ~3% edge, so a bump that crosses it fails loudly) and level 9 recreated per block; the decode side pins level 1 retained and level 19 released.

native/shuffle/src/writers/shuffle_block_writer.rs
dwsmith1983: I looked at a lifetime reservation: the pusher contract treats reserve, push, and release as one synchronous invocation with a single outstanding reservation, and release takes no amount, so a standing charge would need the JNI contract extended on both sides. Kept RSS per-invocation (same cost as main) and updated the description so it no longer reads as if both paths benefit. Happy to look at extending the pusher contract as a follow-up if there is appetite.

native/shuffle/src/shuffle_writer.rs
dwsmith1983: Added the counter, test-gated on both context types. The burst test asserts exactly one creation across a two-partition spill burst and two after the finish burst, and the boundary tests from your other comment use the same counter.

sunchao (commented):
Reviewed `7c907a84` against `199a910b`. I found no new P1/P2 correctness finding in the current source. The [writer benchmark rerun](https://github.com/apache/datafusion-comet/pull/5565#pullrequestreview-5067640978) remains unanswered, and the current-head workflows are still `action_required`, so the performance and execution evidence is not yet sufficient for approval.

Could you also add a focused decoder-reuse microbenchmark against the PR base? Using identical prebuilt frames and fixed total decoded bytes, compare repeated small zstd frames with a large-frame control and a sequence where the measured `DCtx` exceeds 8 MiB before returning to small frames. Representative numeric and nullable-string data, plus a `NONE` control, would help distinguish reuse savings from reset/locking overhead. Please report decode time, allocation/context-creation counts, and peak versus retained native memory, including after scan-owner release and static JNI worker reuse. Matching builds/dependencies, repeated warmups, identical decoded results and confirmation of the native reader path would make the comparison useful. This is a request to validate the tradeoff, not a claim that a regression has been measured.

dwsmith1983:
Re-ran the writer benchmark at the current head against the merge base, per @andygrove's request, same input and shapes; description table updated:

| shape | base avg | head avg | encode base -> head |
|---|---|---|---|
| 2,000 parts, zstd-1 | 0.323s | 0.328s | 0.188 -> 0.187s |
| 2,000 parts, zstd-6 | 0.743s | 0.746s | 0.606 -> 0.610s |
| 10,000 parts, zstd-3 | 0.581s | 0.557s | 0.326 -> 0.297s |

The retained-size checks and boundary releases cost nothing measurable at 2,000 partitions, and the 10,000-partition result holds at the head the code actually produces (~4% wall, ~9% encode).

@sunchao added `benches/ipc_decode.rs`: frames prebuilt once via ShuffleBlockWriter (Int64 + nullable Utf8), decoded results asserted identical across variants before measuring, four scenarios, repeated small zstd frames, a large-frame control, small frames with a wide-window level-19 frame every 16 (retained workspace passes the cap, context drops and recreates), and a NONE control, each with one reused context vs a fresh context per frame (the per-frame-creation profile of the base path; head-fresh matches a trimmed copy of the bench run on the base tree within noise: 7.289 vs 7.320 ms on the small-frames scenario).

| scenario | reused | fresh per frame |
|---|---|---|
| 64 small zstd-3 frames | 7.285 ms | 7.289 ms |
| 1 large zstd-3 frame | 8.945 ms | 8.918 ms |
| over-cap recovery | 7.772 ms | 7.779 ms |
| 64 small NONE frames | 337 us | 342 us |

Decode reuse is throughput-neutral here, about 56 ns/frame of context setup against ~114 us/frame of decode work, and the over-cap drop/recreate adds nothing measurable, so no regression from the session reset or the operator-owned locking either. The decode-side case for the context API is the bounded retained workspace, not speed. Creation counts are pinned by the test counters and retained-vs-released sizes by the measured table in codec_context.rs; I did not instrument peak RSS beyond sizeof. I can add if you would like.

sunchao (approved):
Reviewed the update at `6c8bae4f`; no new findings.

sunchao (commented):
Re-reviewed `d8fa1615c9f82555df8e31a85992685a2488dd59` against `8729f6e6adf7091e18a48670e790d4ba8fd41e51`. The PR-only patch is unchanged from the previously reviewed `6c8bae4f` update, and I found no new P1/P2 issues. The existing approval stands. Current-head workflows report `action_required`, and no local tests were run in this re-review.

sunchao (commented):
Thanks for syncing the base. I checked the upstream Arrow-export changes against the unchanged codec patch at `dc8bb14e` and found no new issue. The existing approval stands. No tests were rerun; current-head workflows require action.

sunchao (commented):
Rechecked `7a120d1e` after the base update. The codec patch is byte-for-byte unchanged. I checked the imported Celeborn reservation/transport and shared codec-enum interfaces and found no new P1/P2. The existing approval stands.

No tests were rerun. This head has no check runs and three workflows awaiting approval.

sunchao (approved):
Re-reviewed `1372d786575c2512b08ce0cc1ea0eea9fa0bfac8` against `ef62b463`, including the merge adaptation that passes the operator-owned decode context through the shared remote dictionary-normalization path. I found no remaining P1/P2.

Four focused decoder-component tests passed, covering reused/fresh/static entry points, codec transitions, dictionary values and nulls, error recovery, and oversized-context release. This is not full Comet/JNI/Spark validation or a benchmark. The three current-head workflows still require authorization, with no executed check results reported.

sunchao:
Reviewed **`1372d786`****&#x20;against&#x20;****`ef62b463`**. The overall design is sound, and I found no new blocking defect. I would keep context reuse, with these refinements:

1. **Release decoder memory at EOF.**\
   The EOF branch leaves the decoder context populated. Plan and stream clones share it, so a finished input can retain its workspace while other parts of the task continue. Clearing the context at definitive EOF preserves all useful reuse and shortens memory retention. The 8 MiB cap bounds this cost but does not eliminate it. [EOF handling (line 185)](/home/dev-user/code/reviews/comet-pr5565-design-20260904/native/core/src/execution/operators/shuffle_scan.rs:185)
2. **Combine the batch and decoder under one mutex.**\
   The new decoder mutex is only acquired while the existing batch mutex is already held. It provides no additional concurrency in the current call graph. A single `Arc<Mutex<ScanInputState>>`, containing the batch and decoder, would preserve sharing across clones while removing an allocation per operator, an extra lock per fetched block, and a second synchronization invariant. This is a concrete simplification; its performance magnitude remains unmeasured. [State and locking (line 141)](/home/dev-user/code/reviews/comet-pr5565-design-20260904/native/core/src/execution/operators/shuffle_scan.rs:141)
3. **Improve the benchmark’s coverage of the actual design.**\
   Its “small” frames contain 8,192 rows, and it calls the decoder directly, bypassing the operator mutex. The reusable context also survives across benchmark iterations. That measures steady-state decoding, but cannot establish reader startup/cleanup costs or operator overhead. Add the previously measured 400-row case, zstd levels 1/3, an LZ4 control, and a finite reader lifetime. An operator-path benchmark would address the remaining overhead question. [Benchmark (line 143)](/home/dev-user/code/reviews/comet-pr5565-design-20260904/native/shuffle/benches/ipc_decode.rs:143)
4. **Correct the supposedly fresh decoder in the reuse test.**\
   Its `fresh` result comes from `read_ipc_compressed`, which now uses the thread-local cached context. The test therefore compares two reused contexts. Create a new `ShuffleDecodeContext` for each reference decode so the test exercises its stated comparison. [Test (line 331)](/home/dev-user/code/reviews/comet-pr5565-design-20260904/native/shuffle/src/ipc.rs:331)

The remaining abstractions mostly earn their place:

- **Task-owned encoder state with mutable borrowing is appropriate.** Putting contexts inside per-partition writers would multiply retained memory and lose reuse across small partitions.
- **The spill-completion hook is reasonable.** It expresses a lifecycle boundary without teaching the partitioner about zstd, and current callers invoke it on both success and error.
- **Resetting the session and applying the level are justified.** The zstd constructor does neither. Caching the last level would add bookkeeping for an unmeasured, small saving.
- **RSS’s per-invocation release should remain.** Longer reuse requires changing its memory-reservation contract. The existing context types and thin decoder wrappers are preferable to introducing a general codec pool or framework.

One optional performance follow-up: zstd’s Rust wrapper still allocates a **32 KiB output buffer per frame**. This predates the PR. Profile that allocation before adding lower-level streaming machinery to reuse it. [Allocation (line 49)](/home/dev-user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zstd-0.13.3/src/stream/zio/writer.rs:49)

This was a fresh source/design review. The measured codec paths and lockfile are unchanged from our earlier benchmarked revision; I did not rerun benchmarks or tests. Current CI has **64 successful checks**, with the performance check skipped.

dwsmith1983:
Thanks for the careful pass. All four are in d21b24eb9:

1. EOF now releases the decoder. Every store of an input batch goes through one `receive` point on the scan input state, and `InputBatch::EOF` resets the decode context there, so a finished input no longer keeps its zstd workspace while the task continues. Pinned by `eof_releases_retained_decoder_workspace`.
2. Batch and decoder live under a single `Arc<Mutex<ScanInputState>>`. The separate decoder mutex is gone, and the `batch` field is no longer `pub` (nothing outside the module used it). Sharing across `execute()` clones is unchanged.
3. The decode benchmark now covers {400, 8192} rows x {zstd 1, zstd 3, lz4, none}, each with a context that outlives iterations, one locked per frame behind the same `Arc<Mutex<_>>` shape the operator uses, one created per reader lifetime of 64 frames and dropped (so startup and cleanup are measured), and a fresh context per frame. On this machine the per-frame context create and drop is about 2 percent of decode time at 400 rows and within noise at 8192 rows, and the lock is not measurable. That matches the earlier framing: the decode side is about retention bounds, not throughput. A benchmark through `ShuffleScanExec` itself is not possible without a JVM, since its decode path runs inside `with_env` and fetches blocks through JNI; the locked variant is the closest stand-in and is labeled as such.
4. The reuse test now builds a fresh `ShuffleDecodeContext` for every reference decode and asserts the reused side created exactly one context across all frames, so it compares what it says it compares.

The 32 KiB per-frame output buffer in the zstd writer is noted as a separate follow-up; I did not touch it here.

dwsmith1983:
@sunchao the four refinements from your design pass are in d21b24eb9, details in the comment above. Ready for another look.

sunchao (approved):

sunchao (approved):
## Correctness

Re-reviewed `f2df122c` against `2da32915`, including the merge with main's scratch-buffer recycling. Fifteen of the 19 feature patches are unchanged after removing hunk positions. I traced the four adaptations that now pass both scratch and codec context through the writer APIs.

Single-partition writes keep the same scratch/context through final flush. Multi-partition spill and finish paths drain or clear scratch before reuse. The spill-completion hook still releases zstd state before freeing the input reservation, including on write failure. Codec reset, compression-level assignment, independent-frame completion, RSS release and decoder EOF handling are unchanged. I found no new or remaining P1/P2.

The maintained Spark 3.5/4.0 compression sources were checked for frame completion and level handling. This preserves Comet's existing IPC framing, not interchangeability with Spark's serializer. No expression, null or ANSI behavior changes here. Maintained Spark 3.4/4.1 source coverage remains unavailable.

Formatting checks passed for all 18 changed Rust files. No native compilation, product tests, JVM execution or benchmarks were run in this follow-up. Earlier remote results were not treated as current validation. At 03:42 UTC, all three exact-head workflows were `action_required`, with zero check runs. The synthetic merge also uses a newer first parent than the API base, so it is not evidence of an executed test pass.

## Performance

The scratch adaptation adds borrowing, not another payload copy or a codec allocation per partition. Task-level local context reuse and scratch capacity reduction remain intact. RSS still releases the workspace per admitted invocation. The unchanged zstd wrapper retains a 32 KiB output allocation per frame, and the 8 MiB context threshold bounds between-frame retention rather than peak allocation.

The existing decode benchmark covers fresh and reused contexts plus codec controls, but not JNI fetching. Its earlier figures were not reproduced against this merged base. I make no new throughput claim.

## Design

Caller-owned scratch and task-owned codec state remain separate resources with explicit lifetimes. Passing both through the existing APIs preserves main's recycling and the established reservation boundaries. Extending RSS reuse would still require changing its reservation contract.

## Abstraction & complexity

The update adds no pool, cache hierarchy or synchronization layer. It keeps the small context types and existing writer interfaces. The shared scan-state mutex is unchanged, and I found no additional abstraction needed for this adaptation.

sunchao (commented):
## Correctness

Re-reviewed `89aed721` against `719cba11`. The complete feature patch and all 19 feature-file blobs are identical to the previously reviewed `f2df122c` pair. The entire native tree is unchanged. The only head update is `CometExplodeBenchmark.scala`, copied exactly from the new base. No new or remaining P1/P2 was found, and the existing approval stands.

The unchanged framing and resource-lifetime paths were rechecked against the maintained Spark 3.5/4.0 sources. This update changes no expression, null, ANSI, error or fallback behavior. Maintained Spark 3.4/4.1 source coverage remains unavailable.

Source/dependency identity and three diff checks passed. No product tests, JVM/native execution, formatting or benchmarks were rerun. At 05:24 UTC, all three current-head workflows were `action_required`, with zero check runs. The current synthetic merge matches this head's tree, which establishes source identity but not executed validation.

## Performance

There is no shuffle hot-path, allocation, locking or dependency change in this update. Existing decoder benchmark coverage is unchanged, and the inherited explode benchmark does not measure codec reuse. No new performance measurement is claimed.

## Design

The merge preserves the already-reviewed ownership and reservation boundaries. The imported benchmark adds no production registration or configuration, so I found no new design interaction.

## Abstraction & complexity

No production helper or abstraction changes here. The synchronization introduces no additional complexity into the codec feature.

andygrove:
This seems lot a lot of additional complexity for little performance gain?

andygrove:
Here's my AI review:

> This seems lot a lot of additional complexity for little performance gain?

Digging into the numbers, I think that instinct is right, and I'd narrow the scope rather than land this as is.

**The measured gain is about 3 µs per shuffle block, and the defaults never reach it.**

Two things from `CometConf.scala` matter here: the default shuffle codec is `lz4` (line 510) and the default zstd level is `1` (line 518). This PR does nothing for lz4 or snappy, so out of the box it's inert — you need a non-default codec *and* a non-default level *and* a high partition count before it does anything.

Working backwards from the benchmark table gives a consistent mechanism:

| shape | encode base → PR | blocks | implied per-block |
|---|---|---|---|
| 10,000 parts, zstd-3 | 0.326 → 0.297s | ~10,000 | **~2.9 µs** |
| 2,000 parts, zstd-1 | 0.188 → 0.187s | ~2,000 | predicted ~6 ms, below noise |
| 2,000 parts, zstd-6 | 0.606 → 0.610s | ~2,000 | predicted ~6 ms, below noise |

(4M rows / 10k partitions is ~400 rows each, under the 8192 batch size, so blocks ≈ partitions.) Worth noting that level 6 has a *larger* workspace than level 3 and still showed nothing, which confirms the variable is block count, not level. So the honest claim is ~3 µs saved per block: 29 ms at 10,000 blocks, 0.6 ms at Spark's default 200 partitions. Given how sensitive these shuffle benchmarks are to machine load, I'd also want the 10k result reproduced on Linux with more than 3 iterations before treating the 4% as real.

**The decode half has no measured benefit, and it introduces a retention hazard main doesn't have.** The comment above puts it at 56 ns/frame of setup against ~114 µs/frame of work. The stated justification — "the decode-side value is the bounded retained workspace, not speed" — is circular: on main, `zstd::Decoder::with_buffer` builds a fresh DCtx per frame, so retention between frames is already zero. The PR introduces retention and then spends the 8 MiB cap, the `sizeof()` checks, the error-path releases, and the EOF hook bounding retention it created.

The thread-local in `ipc.rs` also backs the live production path for the JVM shuffle reader (`NativeBatchDecoderIterator` → `Native.decodeShuffleBlock`), so every Spark task thread now holds roughly 1-2.6 MB of native zstd workspace that Comet's memory accounting doesn't track. That's exactly the property used to justify releasing per-invocation on the RSS encode path — same hazard, opposite decision, for a 0.05% saving.

**Where the complexity actually lands.** It's ~290 net production lines and ~1,080 test/bench lines, but the part that costs us long term is four things:

1. `MAX_RETAINED_ZSTD_CONTEXT_BYTES` is a magic constant backed by a hand-measured table of zstd workspace sizes, with a test whose only job is to fail when a zstd bump silently disables the optimization (level 8 sits 3% under the cap). That's a tax on every dependency bump.
2. Two divergent lifetime contracts for one object — RSS releases, local retains — enforced by a bool threaded through `write_batch_with_codec_limits`.
3. A new `write_burst_complete()` on the `PartitionWriter` trait that exists only to drop one codec's allocator, i.e. a lifecycle hook in the partitioner API serving zstd specifically.
4. The thread-local above.

**What I'd suggest instead:** keep the encode-side reuse for local shuffle, and drop the decode half entirely — `ShuffleDecodeContext`, the four `_with` entry points, `decode_remote_shuffle_batch_with`, the thread-local, the `ScanInputState` restructure, `benches/ipc_decode.rs`, and the decode tests. That's roughly half the production diff, removed for something already measured as neutral.

Then, could the `sizeof()`-based cap be replaced with a retain/don't-retain decision made once in `ShuffleBlockWriter::try_new` from the configured level? Codec and level are uniform per query, so the alternating-level test is defending a case that can't occur in production, and `reset(SessionOnly)` preserves parameters anyway, which makes the per-block `set_parameter` redundant. That would drop the constant, the measured table, and both `release_zstd_if_oversized` sites, leaving the mechanical `&mut` threading as the only real cost.

One hypothesis I checked and discarded, in case anyone else wonders the same: hoisting `CompressionContext` out of `BufBatchWriter` is *not* a hidden memory win at high partition counts. `IpcWriteOptions::try_new(64, false, V5)` leaves Arrow's inner buffer compression off, so those per-partition contexts were an empty `Vec` plus a `None`, about 40 bytes each. Good hygiene, not a saving.

dwsmith1983:
> This seems lot a lot of additional complexity for little performance gain?

Fair question. the gain is on the encode side, 7 percent wall and 11 percent encode time at 10k partitions with zstd level 3 on the writer benchmark, and flat at large blocks. The decode side measured throughput neutral; its value is bounding retained memory. About half of the added lines are tests, and a good part of the surface came from review asks (the exec-side decode context, the retained-size cap, the spill boundary hook, RSS releasing per invocation).

If the trade off looks wrong to you, the cleanest is to drop the decode side entirely and keep only writer-side context reuse. That removes the scan operator changes, the decode context type, and the thread-local, roughly half the diff, while keeping the measured win.

peterxcli (commented):
Haven't looked through the code yet, but this might relate to https://github.com/apache/datafusion-comet/issues/5446 and https://github.com/apache/datafusion-comet/issues/5008

sunchao (commented):
## Correctness

Re-reviewed `a23d1ed6278dd753f32dc386a000bb0f9ece65a0` against `7190df631afe3795914839203c7afe57ea23903c`, since accepted `89aed721`. This is not a byte-identical synchronization. Seventeen of the 19 feature files are identical, while two RSS files integrate main's typed size-limit errors and tests. The PR's added and removed lines remain identical after excluding hunk positions and unchanged context.

I traced the affected admission, encode, error, growth-retry and split paths. `write_rss_batch` still releases its zstd context before returning, so output/compaction cleanup and reservation release occur afterward. Early admission failures allocate no context. Scratch draining, local spill release before `reservation.free()`, configured-level initialization, independent-frame completion, decoder reset and EOF release remain intact. No new P1/P2 was found in this integration.

The maintained Spark 3.5/4.0 compression sources preserve configured levels and distinguish ordinary completed frames from continuous streams. Comet still completes each independent IPC frame. This update changes no expression, null or ANSI semantics and does not imply interchangeability with Spark serializer bytes. Maintained Spark 3.4/4.1 sources remain unavailable. Locked Arrow 58.4 and zstd dependencies are unchanged. The newer Arrow 59.2 work described in #5446 is outside this exact pair.

Three diff checks and fresh source/dependency/archive verification passed. No product tests, native/JNI/Spark execution, formatting or benchmarks ran. At `2026-09-05T21:35:07.916Z`, all three workflows were `action_required`, with zero head or merge checks. GitHub reported `dirty` and no synthetic merge SHA, although the assigned base is a parent of this head. There is no executed CI checkout to qualify as a pass. An existing approval is associated with this head, so I am leaving a follow-up comment rather than another approval.

## Performance

The [published writer table](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5487809068) implies about 4.1% wall-time and 8.9% encode-time improvement for 10,000 partitions at zstd level 3, rather than the later reply's 7%/11%. These are historical author measurements, not a fresh result. The defaults are LZ4 and zstd level 1. This does not establish that a non-default zstd level is necessary for a gain, or that the added context bookkeeping has zero default-path cost.

The decoder reports range from roughly 2% setup/drop cost at 400 rows to neutral at 8,192 rows. The benchmark includes finite-reader and mutex-proxy controls but not the production JNI path or measured peak memory. The [existing request](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5552528334) for a repeated Linux writer comparison is appropriate. Any retained decoder optimization also needs evidence that its benefit justifies the retained memory and lifecycle overhead. Source equality does not supply that measurement.

## Design

The encoder-only option offered by the author is a concrete simplification. I agree that decoder retention is not a memory improvement over this base: the base drops an owned decoder after every frame. The PR instead adds operator-owned and thread-local retention, then bounds it. EOF release fixes the operator lifetime, but the JNI thread-local can still retain up to the cap across tasks. Please resolve the existing decoder-scope discussion before merge. The merge update has not removed that part of the implementation.

## Abstraction & complexity

Task-owned encoder state and mutable borrowing avoid multiplying workspaces by output partition. The spill hook and RSS per-invocation release express real accounting boundaries. Dropping decoder reuse would remove a separate context type, wrapper entry points, thread-local state and owner lifecycle work.

I would not remove the measured-size guard or level initialization solely from the discussion's uniform-level argument. `with_context` does not initialize the configured level, and every replacement context needs that initialization. A level-based retention policy would also need to preserve the actual memory bound across supported configurations and dependency updates. The current 8 MiB check limits retention between frames, not peak memory.

dwsmith1983:
Trimmed to the write path in 27c2033c1: the decoder context, the thread-local, the scan operator changes and the decode bench are gone, and ipc.rs is back to main. The retained-size cap and the level initialization stay, for the reasons sunchao gave. Remaining diff is about 260 production lines.

One correction to my earlier reply: the posted table implies 4.1 percent wall and 8.9 percent encode at 10,000 partitions, not 7 and 11.

Linux run, since you asked: rust 1.94 on linux/arm64 in Docker, 10,000 partitions, zstd level 3, ten timed iterations, base and head alternated. Round one gives 3.0 percent wall and 6 percent encode with the min/max ranges not overlapping. A second round showed more, but the base run drifted, so I would not read anything into it beyond the direction. That is in line with your per-block estimate. On macOS the same shape over five iterations gives 5.6 percent wall and 6.6 percent encode.

dwsmith1983:
> Haven't looked through the code yet, but this might relate to #5446 and #5008

#5038 moved Arrow's `IpcWriteContext` from per batch to per partition writer; this PR moves it once more, to one per task (`ShuffleCodecContext.arrow_ipc`, borrowed by each `BufBatchWriter` and the spill writer), and puts the outer zstd `CCtx` beside it, which #5446 lists as out of scope. Under Arrow 58.4 the task-level hoist was cosmetic, the context held an empty Vec and a None. Under 59.2 the context owns the `FlatBufferBuilder` and keeps its buffer across writes, so per partition it is one retained builder per partition, ten thousand at the shape in the benchmark, and per task it is one. So the allocation measurement in #5446 would differ between main and this branch; running it on both would show whether the task scope is worth anything on its own.

sunchao (approved):
## Correctness

Re-reviewed `5a1ead6e8914ceaf18d95cb53b597ebe19013032` against `7f1e00189b1ed86f1cb5acd872d97fce694482b1`, following review 5123119243. The decoder-scope concern is addressed: `ipc.rs` and remote decoding match the base, the production scan code matches the base, and the decoder context, thread-local cache and decode benchmark are removed. The remaining scan changes adapt tests to the writer API. I found no new or remaining P1/P2.

I rechecked the encoder paths through the merged source. Each zstd frame resets its session, applies the configured level and finishes independently. The size check runs after successful and failed compression; spill completion releases zstd before the input reservation is freed. RSS releases it before returning to the caller's reservation cleanup. Successful local finish releases it explicitly; production finish errors unwind the task-owned repartitioner. Mutable borrowing keeps the context exclusive while a frame is encoded. Dictionary-bearing schemas still use a fresh Arrow stream writer per frame.

The maintained Spark 3.5/4.0 compression sources confirm configured-level handling and the distinction between completed frames and continuous streams. Comet preserves its existing independent IPC framing; this does not imply interchangeability with Spark serializer bytes. There are no new expression, null, ANSI or fallback semantics. Maintained Spark 3.4/4.1 source coverage remains unavailable.

This pair locks Arrow **59.3.0** and DataFusion **55.0.0**, with identical dependencies in head/base; zstd remains 0.13.3 / zstd-safe 7.2.4 / zstd 1.5.7. I verified the relevant library source against the locked crate checksums, refreshed the maintained Spark source checks, and passed both diff checks and formatting checks for all 15 changed Rust files. No native build, product tests, JNI/Spark execution or benchmarks ran in this follow-up. At `2026-09-08T02:57:34.298Z`, all three current-head workflows were `action_required`, with zero head or merge check results. The test counts in the description are author reports, not current-head CI validation.

## Performance

The [September 6 report](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5559185617) addresses the request for a repeated Linux writer comparison: Linux/arm64 Docker, ten timed iterations with base/head alternation, reporting 3.0% wall and 6% encode improvement in the first round. The author appropriately discounts the drifting second round and corrects the earlier 7%/11% claim to the table's 4.1%/8.9%. These remain author measurements of the trimmed revision, without raw iterations or complete commit/build provenance in the discussion; they do not establish performance at this merged head. The description should also be reconciled with that report because it still says no Linux run was available.

Arrow's shared context now contains a reusable FlatBuffer builder, so hoisting it can avoid repeated builder allocations even when outer compression is LZ4 or NONE. However, on this base the multi-partition spill/finish loops create temporary `BufBatchWriter`s sequentially. The source does **not** establish the latest comment's implication of 10,000 simultaneously retained builders. A paired allocation measurement through that local multi-partition path would substantiate the additional benefit; the existing fresh/reused context benchmark alone does not measure the ownership change. RSS still creates a zstd context per admitted block, as documented.

## Design

The update takes the concrete simplification discussed previously: retain writer-side reuse and restore ordinary per-frame decoder ownership. This removes the decoder retention introduced by the earlier revision. The remaining task-owned encoder context fits the sequential local write path, and the spill hook expresses a real lifetime boundary. RSS's shorter lifetime remains justified by its existing admission contract; extending reuse there would require a separate reservation change.

## Abstraction & complexity

Removing the decoder type, wrapper APIs, thread-local cache and scan-state lifecycle work substantially narrows the maintenance surface. The remaining context groups state borrowed by the same encode operation without adding a pool or synchronization layer. Arrow copies the finished metadata into an owned buffer before resetting its builder, so sequential reuse does not leave prior frames borrowing mutable builder storage.

The configured-level initialization and measured-size guard remain justified. The 8 MiB limit bounds retained **zstd** workspace between blocks, not peak encode memory or the Arrow builder. It should not be described as a total task-memory cap. The level-8 boundary test remains useful protection against a dependency update silently disabling reuse.

dwsmith1983:
The multi-partition path builds its writers one at a time, so "ten thousand retained builders" was wrong; the effect is per flush, not simultaneous retention. Measured with the allocation counter #5727 added, driving `LocalPartitionWriter` in multi mode over 10,000 partitions with a 300-row batch, finish pass and spill pass, warm-up first, main at bb9e74020 versus this head:

| codec | main allocs / bytes per flushed partition | head | delta |
|---|---|---|---|
| none | 48.0 / 36,648 | 40.0 / 36,096 | -8 / -552 |
| lz4 | 51.0 / 190,677 | 43.0 / 190,125 | -8 / -552 |
| zstd 3 | 49.0 / 69,416 | 41.0 / 68,864 | -8 / -552 |

The 8 allocations and 552 bytes are the `FlatBufferBuilder` growing during its first message, once per fresh context, and the same across codecs, so that part is the Arrow context and nothing else; it grows with message width (12 / 8,232 at 50 columns). Empty partitions cost nothing on either tree. The counter cannot see the zstd C side: main creates a `CCtx` per flushed partition, about 3.7 MB at level 3 by `sizeof`, where the head creates one per task, which is the larger effect and also a count effect. Probe sources and raw output are in the discussion if wanted; the description now carries the Linux run.

sunchao (commented):
## Correctness

Rechecked `d4fec191` against `bb9e7402`, following review 5136951863. All 15 files in the PR contribution are byte-identical to the previously reviewed `5a1ead6e` head. The entire incremental source diff equals the base update, and the lockfile is unchanged across the prior head, current head and current base. The decoder files and production scan code still match the base. I found no new or remaining P1/P2, and the existing approval stands.

The unchanged callers preserve the reviewed lifetime boundaries. Spill completion releases zstd before freeing the input reservation, including after a write error. RSS releases it before returning to the caller's reservation cleanup. Successful local finish releases it explicitly, while production finish errors drop the task-owned repartitioner. Session reset, configured-level initialization, independent frame completion and fresh dictionary tracking remain intact. The maintained Spark 3.5/4.0 compression sources still agree on configured-level and completed-frame behavior. This update adds no expression, null, ANSI or fallback behavior, and does not imply that Comet IPC bytes are interchangeable with Spark serializer bytes. Maintained Spark 3.4/4.1 source coverage remains unavailable.

Both diff checks and fresh source/dependency verification passed. No native build, product tests, JNI/Spark execution or benchmarks ran in this follow-up. At `2026-09-08T09:25:23.897Z`, all three workflows required action, with zero head or merge checks. The synthetic merge has the assigned parents and the head's tree, which establishes source identity but supplies no executed CI result. The test counts in the description remain author reports.

## Performance

The description now reconciles the [Linux benchmark report](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5559185617). It reports ten timed iterations on Linux/arm64, a first-round change from 1.349s to 1.309s wall time and 0.713s to 0.672s encode time, and explicitly discounts the drifting second round. These are historical author measurements of the trimmed revision, with incomplete exact commit/build and raw-iteration provenance. They do not establish performance at this rebased head.

The new [local-writer allocation report](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5582668040) also corrects the earlier simultaneous-retention claim. The multi-partition writers are sequential. The reported probe exercises finish and spill passes over 10,000 partitions after warm-up, and reports eight fewer allocations and 552 fewer allocated bytes per flushed partition for NONE, LZ4 and zstd. That targets the production ownership change rather than only the existing fresh/reused-context microbenchmark. These remain author-reported results: the probe sources and raw output are offered but not included in the discussion, so I have not independently reproduced the counts. The counter excludes zstd's C allocations. The Arrow allocation result and zstd workspace retention should therefore remain separate from each other and from an end-to-end throughput claim.

## Design

The base update preserves the writer-only scope and the existing reservation interfaces. It introduces no new interaction with decoder ownership. Task-scoped local state and per-invocation RSS release remain appropriate for their respective callers, so I found no new design change required by the rebase.

## Abstraction & complexity

No helper, pool or synchronization layer was added in this update. The measured-size guard still bounds retained zstd workspace between blocks, not peak encode memory or Arrow builder capacity. The level-8 boundary test and explicit level initialization retain their existing purpose without another abstraction.

sunchao (commented):
## Correctness

Rechecked `c314c320` against `8e684685`, following published review 5139980456 at `d4fec191`. All 15 authored files and the complete contribution diff are unchanged. The 43-file increment from that published head exactly equals the base update. The further 17-file increment since `3298b72c` also equals its base update. All 27 previously audited source units, including writer, decoder, JNI, iterator and lockfile files, are unchanged. I found no new or remaining P1/P2. The existing approval stands.

I inspected the newly inherited Variant normalization and `concat_ws` integration boundaries. They do not modify the shuffle interfaces, frame completion, writer ownership or release paths. The previous lifetime and error-cleanup conclusions remain applicable by exact source identity. The maintained Spark 3.5/4.0 compression-source refs and hashes were rechecked. Configured-level and completed-frame behavior remains unchanged, with no new expression, null, ANSI or fallback semantics in the compression contribution. Maintained Spark 3.4/4.1 coverage remains unavailable. This is not a separate qualification of the inherited Variant or expression PRs.

Three diff checks passed. No build, product tests, JNI/Spark execution or benchmarks ran. The lockfile remains identical across the published head, intervening head, current head and base. The complete public snapshot ends at `2026-09-08T16:31:17.975Z`. All three workflows require action and the fresh `16:45:21 UTC` check found zero jobs. The synthetic merge has the assigned parents and head tree, without an executed CI result. Prior test results are not current-head validation.

## Performance

The description and every discussion body are unchanged since the preceding audit. There is no new writer measurement. The published review's Linux and allocation-result qualifications still apply. Main's added hash benchmark and Variant/string work do not measure this PR's unchanged compression reuse. No current-head speedup is inferred from source equality.

## Design

The base merges preserve the existing writer-only contribution and reservation interfaces. Local task ownership and RSS per-invocation release are unchanged. I found no new design interaction requiring a change to this PR.

## Abstraction & complexity

The compression contribution adds no helper, pool or synchronization layer in this update. The existing level initialization and measured-size guard retain their reviewed purposes. The 8 MiB guard bounds retained zstd workspace between blocks, not peak encode memory or total task memory.

dwsmith1983:
@andygrove the decoder half is out as you suggested, and the description reflects the trimmed scope. Ready for your look.

sunchao (commented):
## Correctness

Rechecked `b1851047` against `fefee03d`, following [review 5144578275](https://github.com/apache/datafusion-comet/pull/5565#pullrequestreview-5144578275). All 15 authored files and the complete compression contribution are unchanged. The 64-file increment exactly equals the main update. Of 27 previously audited source units, 24 remain identical. I inspected the changed planner, JNI runtime and lockfile boundaries and found no new or remaining P1/P2. The existing approval stands.

The inherited daemon-thread attachment and dynamic-filter planning do not change the task-owned encoder, frame completion, decoder or reservation interfaces. Session reset and level initialization still occur before each zstd frame. Error paths retain the reviewed size check and spill/RSS releases. Mutable borrowing remains exclusive, and dictionary-bearing Arrow streams still use a fresh stream writer per block. The relevant Arrow and zstd package entries are unchanged despite the added base64 dependency in core.

Maintained Spark 3.5/4.0 compression sources were rechecked for configured levels and completed-frame versus continuous-stream behavior. Comet's existing IPC framing remains unchanged. This does not establish interchangeability with Spark serializer bytes or introduce expression, null, ANSI or fallback semantics. Maintained Spark 3.4/4.1 source remains unavailable.

Three diff checks passed. At 04:22 UTC on September 9, the only successful current-head check is PR labeling. The [CI workflow](https://github.com/apache/datafusion-comet/actions/runs/34293233403), CodeQL and Delta build gate require action. The complete jobs read at 04:05:43 UTC found zero jobs in all three, and the 04:22 UTC workflow refresh still reports them as `action_required`. Synthetic merge `ed9c1c05` has the assigned parents and the head's tree, but no product build/test execution is credited to it. No local build, native/JNI/Spark test or benchmark ran.

## Performance

There is no new benchmark in this update. The existing [Linux writer timings](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5559185617) and [paired allocation report](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5582668040) retain their earlier qualifications: author measurements of older revisions, without complete raw/build provenance for this head. The allocation counter measures Arrow-side allocations, not zstd's C workspace. Neither result establishes current-head end-to-end Spark shuffle benefit, and RSS still recreates zstd per admitted block.

## Design

The main integration preserves local task ownership and RSS's shorter reservation lifetime. The new author comment confirms the previously reviewed removal of decoder reuse. No new ownership or concurrency mechanism needs review in the compression contribution.

## Abstraction & complexity

No compression helper, pool or synchronization layer was added in this update. The existing reset, level initialization, size guard and creation-counter tests retain their purposes. The 8 MiB guard bounds retained zstd workspace between blocks, not peak encoding memory, Arrow builder capacity or total task memory.

andygrove (approved):
The decoder half is out, which resolves the retention concern from my last pass, so what's left is the write path only. `ShuffleCodecContext::zstd_cctx` resets the session and reapplies the compression level on every call. The alternating-level test and the post-write-failure test both show a shared context produces an independent frame each time rather than carrying state from one block into the next. The 8 MiB cap in `codec_context.rs` now has the measured table in the comment, and the level 8 versus level 9 tests pin the boundary directly instead of only asserting retain-or-release.

Keeping RSS on per-invocation release given the pusher's reserve, push, and release contract reads as the right tradeoff rather than reworking that protocol here, and the description no longer implies both paths benefit. Ownership on the local path stays exclusive per task through plain `&mut` borrows rather than a shared handle, so there's no concurrency question left on the writer side either.

sunchao (commented):
## Correctness

Rechecked `abb59d5c` against `424c31aa`, following [review 5149703690](https://github.com/apache/datafusion-comet/pull/5565#pullrequestreview-5149703690) and the [new maintainer approval](https://github.com/apache/datafusion-comet/pull/5565#pullrequestreview-5154027057). All 15 authored files and the complete compression diff are unchanged from `b1851047`. All 28 previously audited source units are identical. The 22-file increment exactly equals the base update. I reread the full contribution and checked the relevant integration boundaries, finding no new or remaining P1/P2. The existing approval stands.

The inherited JNI change generalizes typed exception handling while preserving `ShuffleSizeLimit` classification and its message through DataFusion wrappers. It does not change encoder ownership, frame completion or reservation cleanup. The earlier decoder-retention findings remain addressed because decoder reuse is still absent. Local spill releases zstd before freeing the input reservation, and RSS releases it before the caller's per-invocation reservation cleanup. Production finish errors still unwind the task-owned writer. The new cache-statistics and aggregate-planning changes do not alter these interfaces.

Maintained Spark 3.5/4.0 compression sources still confirm configured-level handling and the distinction between completed frames and continuous streams. This PR preserves Comet's existing independent IPC framing, without claiming interchangeability with Spark serializer bytes. No expression, null, ANSI or fallback semantics changed. Maintained Spark 3.4/4.1 sources remain unavailable. Arrow 59.3.0, DataFusion 55.0.0 and the pinned zstd packages are unchanged, as is the full lockfile.

Three diff checks passed. At 16:30 UTC on September 9, CI, CodeQL and the Delta build gate remain `action_required`, each with zero jobs. Only labeling succeeded, with its log confirming base `424c31aa` checkout. Synthetic merge `a58884bc` has the assigned parents and the head's tree, but no product build/test execution is established. No local build, native/JNI/Spark test, size probe or benchmark ran.

## Performance

There is no new benchmark or change to the encode hot path. The [Linux writer timings](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5559185617) and [paired allocation report](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5582668040) remain author measurements of older revisions. The allocation counter measures Arrow-side allocations, not zstd's C workspace. These reports do not establish end-to-end Spark shuffle benefit at this head. RSS still recreates zstd per admitted block, as the description and maintainer review acknowledge.

## Design

The maintainer's new approval agrees with the already-reviewed writer-only scope and RSS reservation tradeoff. The base integration preserves exclusive local task ownership and the shorter RSS lifetime. No new codec owner, concurrency mechanism or reservation protocol was introduced.

## Abstraction & complexity

The update adds no compression helper, pool or synchronization layer. The reset, configured-level initialization, measured-size guard and creation-counter tests retain their existing purposes. The 8 MiB guard bounds retained zstd workspace between blocks, not peak encoding memory, Arrow builder capacity or total task memory.

sunchao (commented):
Re-reviewed `88b0e85f` against `5e302d99` following `abb59d5c`. All 15 files in the PR contribution are unchanged, and the intervening changes exactly match the base advance. The previously addressed decoder-retention and encoder-release concerns remain fixed. I found no new P1/P2 issue, and my approval stands.

CI now supplies execution evidence that was unavailable in the previous review. [Rust CI](https://github.com/apache/datafusion-comet/actions/runs/34545897014/job/103117651535) passed 1,305 tests, with 5 skipped, including reuse counts, compression-level boundaries, release boundaries and recovery after a write error. The [Linux Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/34545897014/job/103123410208), [Linux Spark 4.0](https://github.com/apache/datafusion-comet/actions/runs/34545897014/job/103123410281) and [macOS Spark 4.0](https://github.com/apache/datafusion-comet/actions/runs/34545897014/job/103125274118) shuffle jobs also passed. I verified their checkout was the exact head/base merge and matched the native artifact digests to its producer jobs. I did not run local tests or benchmarks. The performance check remains skipped, so the earlier timing and allocation figures remain historical measurements.

sunchao (commented):
Re-reviewed `98a7c191` against `8320ae48` after `88b0e85f`. The 15-file PR contribution and dependency lock are unchanged. The 35-file increment exactly matches the base advance. I rechecked the integration and encoder lifetimes: the earlier retention and release concerns remain addressed, with no new P1/P2 findings. My approval stands.

Current-head [CI](https://github.com/apache/datafusion-comet/actions/runs/34603298237), [CodeQL](https://github.com/apache/datafusion-comet/actions/runs/34603297730) and the [Delta build gate](https://github.com/apache/datafusion-comet/actions/runs/34603297768) require workflow approval and have no jobs yet. Only the label check has passed. The previous 1,305 Rust tests and shuffle-suite passes remain evidence for the earlier merged tree, not a current-head CI pass. This follow-up used source equivalence and lifecycle checks. I ran no new local tests or benchmarks.

sunchao (commented):
Re-reviewed `d24ebb9d` against `db790673`. The complete 15-file shuffle patch is unchanged from `98a7c191`. The prior decoder and encoder retention findings remain addressed. I checked the inherited DataFusion 55.1 and zstd dependency updates: the reservation/stream interfaces, zstd context reset/release operations, and vendored compression source are unchanged. No new correctness, performance, design, or complexity findings.

Validation is source-level. There was no local Spark/JNI/native execution or new benchmark, and the workspace-size table was not remeasured with the updated zstd-sys build. At the September 12, 19:13 UTC snapshot, [CI](https://github.com/apache/datafusion-comet/actions/runs/34667588465), [CodeQL](https://github.com/apache/datafusion-comet/actions/runs/34667588223), and the [Delta gate](https://github.com/apache/datafusion-comet/actions/runs/34667588268) remained `action_required`. Only labeling passed. The current synthetic merge was not tested. My existing approval stands.

andygrove (commented):
My approval still stands. The eight commits since are all merges from `main`, and the PR-only patch is byte-for-byte identical to what I approved modulo hunk offsets. `main` did not touch `native/shuffle/` in that window.

Two things to pick up on the next push. The workspace table in `codec_context.rs` is pinned to `zstd-sys 2.0.16+zstd.1.5.7`, and the two boundary tests in `shuffle_block_writer.rs` quote the same build. Since I approved, #5865 bumped us to datafusion 55.1.0, which moved `zstd-safe` to 7.3.0 and `zstd-sys` to `2.1.0+zstd.1.5.7`. The numbers do still hold as far as I can tell, since both `zstd-sys` releases vendor libzstd 1.5.7 and the only source change in `zstd-safe` 7.3.0 is marking `compress_block`, `decompress_block` and `insert_block` unsafe, none of which this path calls. `zstd_context_retained_at_level_eight_near_cap` also passed on `5368009`, which already had the bump. Could you refresh the version string in all three places? That comment is the only record of where the 8 MiB cap came from.

The other is a rebase hazard. `finish_all` ends with `self.codec_context.release_zstd()` right after `write_timer.stop()`, and #5807 deletes that whole index-file block and replaces it with `self.partition_offsets.set(offsets)?`. Whichever lands second gets a conflict in exactly the hunk that release line lives in, and the obvious resolution drops it silently. #5807 also changes the second parameter of `LocalPartitionWriter::try_new`, which your new `local_writer_releases_zstd_context_at_burst_boundaries` test passes an index-file path to. Worth watching on the rebase that the release survives and the burst test still asserts what it does today.

Also worth noting the workflows need authorizing on the current head. Only `label` has run on `d24ebb9`. The previous commit carried a byte-identical patch and was green across 59 checks, so this is an approval gate rather than a signal.

For what it is worth I checked the no-history property directly rather than relying on the round trip. `codec_context_serves_alternating_zstd_levels` asserting `sizes[0] == sizes[2]` compares a first frame on a fresh context against a third frame after an intervening level-19 frame, which is a real assertion. And `SessionOnly` does preserve a loaded dictionary, but none is ever loaded here, since `with_context` skips the `load_dictionary` that `Encoder::new` does and the context is a bare `CCtx::try_create()`.

dwsmith1983:
> Could you refresh the version string in all three places?

Done. All three now name zstd-sys 2.1.0+zstd.1.5.7 with zstd-safe 7.3.0, and the two boundary tests and the burst test pass on that build. On #5807, the `release_zstd` call in `finish_all` and the burst test's writer arguments are the two things I will check on whichever side rebases second.

native/shuffle/src/writers/shuffle_block_writer.rs
```diff
@@ -233,10 +237,10 @@ impl ShuffleBlockWriter {
         &self,
         batch: &RecordBatch,
         output: &mut W,
-        compression_context: &mut IpcWriteContext,
+        codec_context: &mut ShuffleCodecContext,
```
sunchao, line 240: ### Correctness

[P2] Migrate the reader benchmark to the new context type

The rebase brought in [`benches/shuffle_reader.rs`](https://github.com/apache/datafusion-comet/blob/2ecbbc2c53dee9eb21fafc51fdff250c20b8f13b/native/shuffle/benches/shuffle_reader.rs#L74-L81), where `encode_block` still constructs `IpcWriteContext` and passes `&mut context` to this method. This parameter now requires the distinct `ShuffleCodecContext` type, so the reader benchmark cannot type-check. It is an unconditional `[[bench]]` target in the shuffle crate, which is a default workspace member; this affects `cargo check --benches` and `cargo clippy --all-targets --workspace`. Please update its import and context initialization to `ShuffleCodecContext`, as already done in the writer benchmark.

sunchao (commented):
Re-reviewed `2ecbbc2c` against `1d0ce5fe`. The three zstd version labels now match the locked build. The prior decoder and encoder retention findings remain addressed, and the spill/finish release calls are intact.

One new **P2** surfaced in the rebase: the newly inherited [shuffle reader benchmark](https://github.com/apache/datafusion-comet/blob/2ecbbc2c53dee9eb21fafc51fdff250c20b8f13b/native/shuffle/benches/shuffle_reader.rs#L74-L81) still passes `IpcWriteContext` to `write_batch`, whose parameter this PR changes to `ShuffleCodecContext`. That benchmark must be migrated before the benchmark targets can type-check; details are inline.

Validation was source-level, including the current caller, manifests, exact dependency versions, and merge-tree parity. I did not run the compiler, Spark/JNI tests, or new benchmarks. At September 15, 05:36 UTC, [CI](https://github.com/apache/datafusion-comet/actions/runs/34910021780), [CodeQL](https://github.com/apache/datafusion-comet/actions/runs/34910021521), and the [Delta gate](https://github.com/apache/datafusion-comet/actions/runs/34910021583) remained `action_required` with zero jobs; only labeling passed.

dwsmith1983:
> Please update its import and context initialization to `ShuffleCodecContext`, as already done in the writer benchmark.

Done. `benches/shuffle_reader.rs` now builds its context as `ShuffleCodecContext::default()`; `cargo check --benches` and clippy with all targets are clean on the shuffle crate, and its 134 tests still pass.

sunchao (approved):
Re-reviewed `cc212dc8` against `a8e8157e`, following `2ecbbc2c`. The [benchmark caller P2](https://github.com/apache/datafusion-comet/pull/5565#discussion_r4012356127) is fixed: `shuffle_reader.rs` now imports and constructs `ShuffleCodecContext`, matching `write_batch`. The other 15 files in the PR contribution are byte-identical to the last reviewed head. Per-frame reset and level initialization, the retained-size cap, spill/finish release, and RSS per-invocation release remain intact. I found no new or remaining P1/P2.

Validation was source-level, including the full contribution, the corrected caller, dependency and integration checks, and synthetic merge-tree parity. The [author reports](https://github.com/apache/datafusion-comet/pull/5565#issuecomment-5678144600) clean benchmark checking and clippy plus 134 shuffle tests. I did not rerun those commands or benchmarks. The earlier timing and allocation figures remain historical author measurements.

At September 15, 15:42 UTC, [CI](https://github.com/apache/datafusion-comet/actions/runs/34985569452) and [CodeQL](https://github.com/apache/datafusion-comet/actions/runs/34985569030) still require workflow approval and have zero jobs. Only labeling passed.

sunchao (commented):
Re-reviewed `7e2d2d39` against `3868760f`. The only new commit is a main merge: all 16 files in this PR’s contribution are byte-identical to the previously reviewed head, and the complete 28-file update exactly matches the base update.

The earlier [P2 benchmark-caller fix](https://github.com/apache/datafusion-comet/pull/5565#discussion_r4012356127) remains intact. I rechecked session reset and compression-level reapplication, per-frame dictionary state, the 8 MiB retention limit, spill/finish release, and RSS release before its reservation ends. No new P1/P2 findings. The existing approval still stands.

Validation is source-based for this update. [CI](https://github.com/apache/datafusion-comet/actions/runs/35042297590) and [CodeQL](https://github.com/apache/datafusion-comet/actions/runs/35042297443) require approval and have zero jobs. The successful labeling job checked out the base. I did not rerun local compilation, runtime tests, or benchmarks. The inherited CI policy now uses the default Spark profile for PRs and moves non-default profiles to nightly/opt-in coverage.

sunchao (commented):
Re-reviewed `3fef718f` against `8c229a70`, following `7e2d2d39`. This is a main merge: all 16 PR files are byte-identical to the prior reviewed head, and the full 39-file increment exactly matches the base update.

The [benchmark-caller fix](https://github.com/apache/datafusion-comet/pull/5565#discussion_r4012356127), per-frame reset and level initialization, retention cap, spill/finish release, and RSS reservation lifetime remain intact. No new or remaining P1/P2. The existing approval still stands.

Validation is source-based. I did not rerun compilation, runtime tests, or benchmarks. At September 16, 12:27 UTC, [CI](https://github.com/apache/datafusion-comet/actions/runs/35090857735) and [CodeQL](https://github.com/apache/datafusion-comet/actions/runs/35090856158) require approval and have zero jobs. Only labeling passed, on a checkout of the base.

PR #4035, opened 2026-04-22
fix: fall back to Spark for shuffle/sort/aggregate on non-default collated strings [Spark 4]

andygrove:
## Which issue does this PR close?

Closes #1947.

## Rationale for this change

The `listagg-collations.sql` test is currently disabled for Spark 4.0.1 because Comet returns `aabb` instead of `ab` for:

```sql
SELECT lower(listagg(DISTINCT c1 COLLATE utf8_lcase) WITHIN GROUP (ORDER BY c1 COLLATE utf8_lcase))
FROM (VALUES ('a'), ('B'), ('b'), ('A')) AS t(c1);
```

Root cause: Comet accepts a collated `StringType` as a supported type in shuffle, sort, and aggregate, but hashes/sorts/compares the bytes directly. For this query the plan includes `CometColumnarExchange hashpartitioning(collate(c1, utf8_lcase))`, so rows with `'a'` and `'A'` are routed to different partitions and the downstream DISTINCT dedup (which runs in Spark with collation-aware equality) never sees them together.

Two things were broken:

1. The Spark 4.0 `isStringCollationType` shim checked `dt.isInstanceOf[StringTypeWithCollation]`, but `StringTypeWithCollation` is an `AbstractDataType` used only for expression type matching, not a concrete data type. The check never matched a real collated `StringType`.
2. Even if it had matched, none of the shuffle/sort/aggregate type predicates were consulting it.

## What changes are included in this PR?

- `common/src/main/spark-4.0/.../CometTypeShim.scala`: Detect non-default collations by comparing `collationId` against the default `StringType` object's id.
- `spark/.../CometShuffleExchangeExec.scala`: Reject collated-string keys in hash and range partitioning for both native and columnar shuffle paths.
- `spark/.../QueryPlanSerde.scala` (`supportedScalarSortElementType`): Reject collated strings as a sort key.
- `spark/.../operators.scala` (`CometBaseAggregate`): Reject collated-string grouping expressions.
- `dev/diffs/4.0.1.diff`: Remove the sections that disabled the failing `listagg-collations.sql` query and its golden output; the test now runs unmodified against upstream Spark 4.0.1.

## How are these changes tested?

New suite `spark/src/test/spark-4.0/.../CometCollationSuite.scala` with four tests:

1. The exact repro from #1947.
2. DISTINCT on a `utf8_lcase` collated column.
3. GROUP BY a `utf8_lcase` collated column.
[...]

PR #5567, opened 2026-08-31
feat: support nested types as native shuffle hash partitioning keys

### LINKED ISSUE #5566: Native shuffle rejects nested types as hash partitioning keys although the native hasher supports them

`CometShuffleExchangeExec.supportedHashPartitioningDataType` rejects nested types
as hash partitioning keys, so a query that repartitions on a struct, array or map
column falls back to Spark for the entire shuffle:

```scala
def supportedHashPartitioningDataType(dt: DataType): Boolean = dt match {
  case st: StringType if isStringCollationType(st) => false
  case _: BooleanType | ... | _: DateType => true
  case _: DecimalType => true
  case dt if isTimeType(dt) => true
  case _ => false     // struct / array / map land here
}
```

The comment above it explains the restriction as:

> Native code does not support hashing complex types, see hash_funcs/utils.rs

That is no longer accurate. `native/spark-expr/src/hash_funcs/utils.rs` hashes
nested types recursively -- struct fields, `List`/`LargeList`/`FixedSizeList`
elements, and map keys and values -- and shuffle partitioning calls the same
`create_murmur3_hashes` entry point, with the same seed, that the `hash`
expression uses. Nested hashing is exercised today through `hash()` / `xxhash64()`,
whose gate (`HashUtils.unsupportedReasonFor` in `serde/hash.scala`) recurses
through struct/array/map, with coverage in `CometHashExpressionSuite`.

So a shuffle on a nested key falls back even though the machinery to run it
natively is present and in use elsewhere.

## Describe the solution you'd like

Allow struct/array/map as hash partitioning keys for native shuffle, checked
recursively so that a leaf type which cannot be hashed natively disqualifies the
whole key and the shuffle still falls back to Spark. Two such leaves exist today:

- **collated strings** -- Comet hashes raw bytes. This is a correctness matter,
  not just a difference: in #1947 / #4035, rows equal under the collation reached
  different partitions and a downstream collation-aware `DISTINCT` in Spark then
  produced `aabb` instead of `ab`.
- **CalendarInterval** -- allowed as a shuffle *data* column, but the native
  hasher has no `Interval` branch and fails with "Unsupported data type in
  hasher" (#5059).

Map keys need one extra restriction. Map entry order is not semantically
meaningful, so two equal maps must hash alike. Spark 4.0+ normalizes a map
shuffle key by wrapping it in `mapsort(...)` (#1941); earlier versions insert no
such normalization, so Comet would hash physical entry order and could route
equal maps to different partitions. Map keys should therefore be admitted only on
Spark 4.0+, and only when the `mapsort` is itself convertible -- `CometMapSort`
supports scalar map keys only, and otherwise the existing partitioning-expression
check already forces a fallback.

Worth putting behind a config so the previous behavior remains available.

## Describe alternatives you've considered

Leaving the gate as it is and only correcting the stale comment. That keeps a
supported, tested native capability unreachable from shuffle for no technical
reason.

## Additional context

Notably, Spark rewrites a deeply nested partitioning key into a `transform(...)`
containing a nested `mapsort(...)`, e.g. for
`struct<a: array<struct<m: map<string, array<int>>, s: string>>, i: int>`:

```
hashpartitioning(if (isnull(k)) null else named_struct(
  a, transform(k.a, lambdafunction(
       if (isnull(x)) null else named_struct(m, mapsort(x.m), s, x.s), x, false)),
  i, k.i), 10)
```

so the normalization applies at any depth, not only to a top-level map key.

viirya:

Closes #5566.

## Rationale for this change

`CometShuffleExchangeExec.supportedHashPartitioningDataType` rejected struct,
array and map partitioning keys, so a query repartitioning on a nested column fell
back to Spark for the whole shuffle. The comment justified this with:

> Native code does not support hashing complex types, see hash_funcs/utils.rs

but that file hashes nested types recursively (struct fields,
`List`/`LargeList`/`FixedSizeList` elements, map keys and values), and shuffle
partitioning calls the same `create_murmur3_hashes` entry point, with the same
seed, that the `hash` expression uses.

Note that the two gates are not equivalent, so this is not simply "what `hash`
already allows": `HashUtils.unsupportedReasonFor` rejects `decimal(precision > 18)`
and `TimeType` at any depth, while the shuffle gate admits both. Comet does not
guarantee that native shuffle assigns rows to the same partitions as Spark
(#3079), and for a high-precision decimal it will not in general -- the native
hasher uses 16 little-endian `i128` bytes where Spark uses the minimal big-endian
`BigDecimal` bytes.

- `CometShuffleExchangeExec`: add recursive `StructType` / `ArrayType` / `MapType`
  cases to the hash partitioning gate, and replace the stale comment. Recursion
  goes through the same predicate, so a leaf type that cannot be hashed natively
  disqualifies the whole key and the shuffle still falls back -- this covers
  collated strings (#1947 / #4035) and `CalendarInterval` (#5059) without needing
  separate checks.
- Map keys are admitted only on Spark 4.0+. Map entry order is not semantically
  meaningful, so equal maps must hash alike; Spark 4.0+ normalizes a map shuffle
  key with `mapsort(...)` (#1941) while earlier versions do not, and Comet would
  otherwise hash physical entry order. When the `mapsort` is not convertible
  (`CometMapSort` supports scalar map keys only) the existing
  partitioning-expression check already forces a fallback.
- New config `spark.comet.shuffle.native.partitioning.hash.nested.enabled`,
  **disabled by default**. The native hasher only vectorizes nested shapes whose
  leaves are primitives; `array<struct<...>>` and a map inside a struct fall
  through to a per-element path that re-enters `create_murmur3_hashes` for every
  element. Enabling this by default before measuring could make those shuffles
  slower than letting Spark do them, so the default stays off until there are
  numbers. I plan to do that benchmark as a follow-up issue and PR.
- Corrects the comment in `CometFuzzTestBase` claiming its Parquet file has no
  nested complex types: `generateSchema` adds `struct<array<..>>` and
  `array<struct<..>>` when both array and struct generation are on. Maps are the
  only thing missing.

`CometNativeShuffleSuite` and `CometFuzzTestSuite`, on two profiles:

- Spark 4.1 (Scala 2.13): 52 and 45 tests, all passing
- Spark 3.5 (Scala 2.12): 47 passing + 5 canceled by `assume(isSpark40Plus, ...)`,
  and 45 passing

New tests, all opting in via the config:

- struct, array, and two-level (`struct<array<..>>`, `array<struct<..>>`) keys
- a four-level key mixing all three recursive branches:
  `struct<a: array<struct<m: map<string, array<int>>, s: string>>, i: int>`
- multi-entry maps written in opposite key orders (`map('a',1,'b',2)` vs
  `map('b',2,'a',1)`), asserting both rows share a `spark_partition_id()`, which is
  the property the Spark 4.0+ restriction exists to protect
- null keys at four levels -- top level, inside a struct, as an array element, and
  as a map value -- with pairs of equal keys asserted into the same partition
- partition assignment compared against Spark per row for shallow and deep keys,
  with `checkCometExchange` so the test cannot pass by comparing Spark to Spark

Fallback tests: a map key on Spark 3.x, a map whose own key is nested, a collated
string nested inside a struct, and a `CalendarInterval` leaf (via `make_interval`)
inside a struct. `CometFuzzTestSuite` keeps its existing expectation that these
keys fall back by default, and additionally asserts they are admitted with the
config on.

Two notes on verification beyond the assertions:

- `checkShuffleAnswer` asserts `shuffleType == CometNativeShuffle` at the plan
  level. To confirm a nested key really is hashed in native code, I also inspected
  the executed plan and exchange metrics for a deep key, which report native
  shuffle-writer counters (`repart_time`, `encode_time`, `interleave_time`,
  `input_batches`) that the Spark and columnar paths do not produce.
- I sanity-checked that the partition-assignment test can fail, by perturbing the
  native struct hash and confirming it reports every partition id shifted by one.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometShuffleExchangeExec.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometShuffleExchangeExec.scala
@@ -51,7 +51,7 @@ import com.google.common.base.Objects
 
 import org.apache.comet.{CometConf, CometExplainInfo}
 import org.apache.comet.CometConf.{COMET_SHUFFLE_ENABLED, COMET_SHUFFLE_MODE}
-import org.apache.comet.CometSparkSessionExtensions.{cometCelebornShuffleFallbackReason, hasFallbackReason, isCometCelebornShuffleManagerEnabled, isCometShuffleManagerEnabled, withFallbackReasons}
+import org.apache.comet.CometSparkSessionExtensions.{cometCelebornShuffleFallbackReason, hasFallbackReason, isCometCelebornShuffleManagerEnabled, isCometShuffleManagerEnabled, isSpark40Plus, withFallbackReasons}
 import org.apache.comet.serde.{Compatible, OperatorOuterClass, QueryPlanSerde, SupportLevel, Unsupported}
 import org.apache.comet.serde.operator.CometSink
 import org.apache.comet.shims.{CometTypeShim, ShimCometShuffleExchangeExec}
@@ -410,12 +410,21 @@ object CometShuffleExchangeExec
   private def nativeShuffleFailureReasons(s: ShuffleExchangeExec): Seq[String] = {
     val conf = SQLConf.get
 
+    val nestedHashPartitioningEnabled =
+      CometConf.COMET_SHUFFLE_NATIVE_HASH_PARTITIONING_NESTED_ENABLED.get(conf)
+
     /**
      * Determine which data types are supported as partition columns in native shuffle.
      *
      * For HashPartitioning this defines the key that determines how data should be collocated for
-     * operations like `groupByKey`, `reduceByKey`, or `join`. Native code does not support
-     * hashing complex types, see hash_funcs/utils.rs
+     * operations like `groupByKey`, `reduceByKey`, or `join`.
+     *
+     * Nested types (struct/array/map) are supported when
+     * `spark.comet.shuffle.native.partitioning.hash.nested.enabled` is enabled: the native
+     * Murmur3 kernel in hash_funcs/utils.rs hashes them recursively. Nesting is checked
+     * recursively, so a leaf type that cannot be hashed natively -- a collated string, or an
+     * interval the hasher has no branch for -- disqualifies the whole key and the shuffle falls
+     * back to Spark.
      */
     def supportedHashPartitioningDataType(dt: DataType): Boolean = dt match {
       // Collated strings require collation-aware hashing; Comet only hashes raw bytes,
@@ -433,6 +442,22 @@ object CometShuffleExchangeExec
         true
       case dt if isTimeType(dt) =>
         true
+      case StructType(fields) if nestedHashPartitioningEnabled =>
+        // `fields.nonEmpty` mirrors the guard on the data-column gate below. An empty struct is
+        // not reachable end-to-end anyway: Parquet cannot store an empty group, and an in-memory
+        // relation with one does not survive scan conversion.
+        fields.nonEmpty && fields.forall(f => supportedHashPartitioningDataType(f.dataType))
+      case ArrayType(elementType, _) if nestedHashPartitioningEnabled =>
+        supportedHashPartitioningDataType(elementType)
+      case MapType(keyType, valueType, _) if nestedHashPartitioningEnabled =>
+        // Map entry order is not semantically meaningful, so two equal maps must hash alike.
+        // Spark 4.0+ normalizes a map shuffle key by wrapping it in `mapsort(...)`, which is
+        // gated separately by CometMapSort (scalar map keys only) and, when unsupported, fails
+        // the expression check below. Earlier Spark versions insert no such normalization, so
+        // Comet would hash physical entry order and could route equal maps differently.
+        isSpark40Plus &&
+        supportedHashPartitioningDataType(keyType) &&
+        supportedHashPartitioningDataType(valueType)
       case _ =>
         false
     }
--- a/spark/src/test/scala/org/apache/comet/CometFuzzTestSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometFuzzTestSuite.scala
@@ -175,10 +175,21 @@ class CometFuzzTestSuite extends CometFuzzTestBase {
         case "jvm" =>
           1
         case "native" =>
-          // native shuffle does not support complex types as partitioning keys
+          // Nested hash partitioning keys are off by default, so native shuffle falls back here.
           0
       }
       assert(cometShuffleExchanges.length == expectedNumCometShuffles)
+
+      // With the config enabled these keys do run through native shuffle. This is the widest
+      // nested-type coverage in the repo, so it is worth asserting that they are admitted rather
+      // than only that they fall back.
+      withSQLConf(CometConf.COMET_SHUFFLE_NATIVE_HASH_PARTITIONING_NESTED_ENABLED.key -> "true") {
+        val enabledDf = spark.sql(sql)
+        enabledDf.collect()
+        val enabledPlan =
+          enabledDf.queryExecution.executedPlan.asInstanceOf[AdaptiveSparkPlanExec].executedPlan
+        assert(collectCometShuffleExchanges(enabledPlan).length == 1)
+      }
     }
   }
 
--- a/spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
@@ -35,7 +35,7 @@ import org.apache.arrow.vector.types.pojo.{Field, Schema}
 import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkEnv
 import org.apache.spark.sql.{CometTestBase, DataFrame, Dataset, Row}
-import org.apache.spark.sql.comet.{CometExec, CometMetricNode}
+import org.apache.spark.sql.comet.{CometExec, CometMetricNode, CometTakeOrderedAndProjectExec}
 import org.apache.spark.sql.comet.execution.arrow.CometArrowStream
 import org.apache.spark.sql.comet.execution.shuffle.CometShuffleExchangeExec
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
@@ -44,6 +44,7 @@ import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
 
 import org.apache.comet.{CometConf, CometExecIterator, CometShuffleBlockIterator, CometShuffleSizeLimitException, Native}
+import org.apache.comet.CometSparkSessionExtensions.isSpark40Plus
 import org.apache.comet.serde.{OperatorOuterClass, PartitioningOuterClass}
 import org.apache.comet.shuffle.ShufflePartitionPusher
 
@@ -61,6 +62,13 @@ class CometNativeShuffleSuite extends CometTestBase with AdaptiveSparkPlanHelper
   }
 
   import testImplicits._
+
+  /**
+   * Runs `f` with nested hash partitioning keys enabled. The config is disabled by default, so
+   * every test that expects a nested key to reach native shuffle has to opt in.
+   */
+  private def withNestedHashPartitioning(f: => Unit): Unit =
+    withSQLConf(CometConf.COMET_SHUFFLE_NATIVE_HASH_PARTITIONING_NESTED_ENABLED.key -> "true")(f)
 
   private def rssShufflePlanBytes: Array[Byte] = {
     val scan = OperatorOuterClass.Operator
@@ -447,8 +455,14 @@ class CometNativeShuffleSuite extends CometTestBase with AdaptiveSparkPlanHelper
                 .filter($"_3" > 10)
                 .repartition(numPartitions, $"_2")
 
-              // Partitioning on nested array falls back to Spark
+              // Partitioning on a nested array falls back to Spark by default, and is native
+              // when nested hash partitioning is enabled.
               checkShuffleAnswer(df, 0)
+              withNestedHashPartitioning {
+                checkShuffleAnswer(
+                  sql("SELECT * FROM tbl").filter($"_3" > 10).repartition(numPartitions, $"_2"),
+                  1)
+              }
 
               df = sql("SELECT * FROM tbl")
                 .filter($"_3" > 10)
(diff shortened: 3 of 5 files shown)


spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometShuffleExchangeExec.scala
```diff
@@ -405,6 +414,22 @@ object CometShuffleExchangeExec
         true
       case dt if isTimeType(dt) =>
         true
+      case StructType(fields) if nestedHashPartitioningEnabled =>
```
andygrove, line 445: Do you have any numbers for this? What I am worried about is the shapes that miss the vectorized paths. `array<struct<...>>` and a map nested inside a struct both fall through `hash_list_with_primitive_elements!` into `hash_list_array!`, which slices a one element array and re-enters `create_murmur3_hashes` for every element, plus a `.columns().to_vec()` per struct element on top of that. The four level key in the new tests pays that at every level.

Since the config defaults to true we are opting everyone into this, and if it turns out slower than just letting Spark do the shuffle then the default is wrong. `struct<int, string>` is not the interesting case here.

spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
```diff
@@ -672,6 +680,194 @@ class CometNativeShuffleSuite extends CometTestBase with AdaptiveSparkPlanHelper
...
+            (3, named_struct('a', array(named_struct('m', map('x', array(1, 2)), 's', 'p')), 'i', 1)),
+            (4, named_struct('a', array(), 'i', 4)),
+            (5, null)""")
+      val df = sql("SELECT * FROM tbl").repartition(10, $"k").sortWithinPartitions($"id")
+
+      checkShuffleAnswer(df, 1)
+    }
+  }
+
+  test("native shuffle on map hash partitioning key") {
```
andygrove, line 790: Every map in the new tests has a single entry, so `mapsort` is a no-op and the property this case is gated on never actually gets exercised. The reason we only admit maps on 4.0+ is that two equal maps with different physical entry order have to hash alike, and nothing here would notice if that stopped being true.

Could you add multi-entry maps written in different key orders, something like `map('a', 1, 'b', 2)` and `map('b', 2, 'a', 1)`, and assert the two rows get the same `spark_partition_id()`?

spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
```diff
@@ -672,6 +680,194 @@ class CometNativeShuffleSuite extends CometTestBase with AdaptiveSparkPlanHelper
...
+        withSQLConf(
+          CometConf.COMET_SHUFFLE_NATIVE_HASH_PARTITIONING_NESTED_ENABLED.key -> enabled) {
+          val df = sql("SELECT * FROM tbl").repartition(10, $"_2").sortWithinPartitions($"_1")
+
+          checkShuffleAnswer(df, expectedShuffles)
+        }
+      }
+    }
+  }
+  test("native shuffle nested hash partitioning key matches Spark's partition assignment") {
```
andygrove, line 845: This is the only test that checks routing rather than the answer, so I would like it to be hard to fool, and right now it never asserts that Comet ran the shuffle. If the gate ever stops admitting these keys this quietly becomes Spark compared against Spark and still passes. Can you add a `checkCometExchange(df, 1, true)` alongside the `sparkRows.nonEmpty` check?

The other gap is nulls. None of the 200 shallow rows or the 40 deep rows have a null key at any level, and that is the case I would most want covered. The struct branch in `hash_funcs/utils.rs` is the one nested branch that does not look at its own null mask. `List` and `Map` both guard on `is_null(row_idx)`, but `DataType::Struct(_)` just takes `struct_array.columns()` and recurses, while Spark is `case null => seed`. So the two only agree when the children happen to be null under a null parent. Parquet gives us that, which is why `CometHashExpressionSuite` is green, but a struct built in the plan may not, and Spark's own rewrite for a nested map key is `If(IsNull(k), null, named_struct(...))`, which is exactly that shape.

If a null struct ever hashes off leftover child values then two rows with the same key can land in different partitions, and that breaks grouping and joins whatever Spark would have done. Could you add null keys at each level here?

spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
```diff
@@ -672,6 +680,194 @@ class CometNativeShuffleSuite extends CometTestBase with AdaptiveSparkPlanHelper
...
+    withParquetTable(
+      (0 until 200).map(i => (i, (i % 13, (i % 7).toString), Seq(i % 11, i % 5))),
+      "tbl") {
+      Seq("_2", "_3", "_2, _3").foreach { keys =>
+        val query =
+          "SELECT _1, spark_partition_id() AS pid FROM (" +
+            s"SELECT /*+ REPARTITION(10, $keys) */ * FROM tbl)"
+        val cometRows = sql(query).collect().map(r => (r.getInt(0), r.getInt(1))).sorted
+        // `withSQLConf` returns Unit, so capture the Spark-side rows via a var rather than
+        // relying on the block's value.
```
andygrove: `withSQLConf` returns `T`, not Unit. `SQLTestUtils` declares `protected override def withSQLConf[T](pairs: (String, String)*)(f: => T): T` and `CometTestBase` mixes it in, so both of these can be `val sparkRows = withSQLConf(...) { ... }`. Also missing a blank line before this test and before the interval one below it.

spark/src/test/scala/org/apache/comet/exec/CometNativeShuffleSuite.scala
```diff
@@ -672,6 +680,194 @@ class CometNativeShuffleSuite extends CometTestBase with AdaptiveSparkPlanHelper
...
+        withSQLConf(CometConf.COMET_ENABLED.key -> "false") {
+          sparkRows = sql(query).collect().map(r => (r.getInt(0), r.getInt(1))).sorted
+        }
+        assert(sparkRows.nonEmpty, "Spark produced no rows; the comparison would be vacuous")
+        assert(cometRows === sparkRows, "partition assignment differs from Spark for a deep key")
+      }
+    }
+  }
+  test("native shuffle on nested hash partitioning key with interval leaf falls back") {
+    // CalendarIntervalType is allowed as a shuffle DATA column but the native hasher has no
```
andygrove, line 908: I do not think this is testing `CalendarIntervalType`. `INTERVAL '1' MONTH` goes through `constructMultiUnitsIntervalLiteral` and comes out as `YearMonthIntervalType(MONTH, MONTH)`. The gate rejects both so the assertion still holds, but #5059 and the `CalendarIntervalType` case in `supportedSerializableDataType` are about the calendar type, and that one stays uncovered.

Could you use `make_interval(...)`, or set `spark.sql.legacy.interval.enabled=true`, so the test matches its comment?

spark/src/test/scala/org/apache/comet/CometFuzzTestSuite.scala
```diff
@@ -171,14 +171,9 @@ class CometFuzzTestSuite extends CometFuzzTestBase {
...
-      val expectedNumCometShuffles = CometConf.COMET_SHUFFLE_MODE.get() match {
-        case "jvm" =>
-          1
-        case "native" =>
-          // native shuffle does not support complex types as partitioning keys
-          0
-      }
-      assert(cometShuffleExchanges.length == expectedNumCometShuffles)
+      // Both shuffle modes support the struct and array partitioning keys in this file (it is
+      // generated without maps and without nested complex types), so one Comet shuffle either way.
```
andygrove: This file does have nested complex types. `FuzzDataGenerator.generateSchema` adds `StructType(arraysOfPrimitives)` and `createArrayType(StructType(primitives))` when both `generateArray` and `generateStruct` are set, so `struct<array<...>>` and `array<struct<...>>` are both in there. Maps are the only thing missing. The comment in `CometFuzzTestBase` says the same thing and is equally wrong, so this was inherited, but could you fix both while you are here?

The knock-on is that this now asserts a native shuffle for `array<decimal(36,18)>` and `struct<... decimal(36,18) ...>`, where we hash 16 LE `i128` bytes and Spark hashes the minimal BE `BigDecimal` bytes. I closed #3079 on the grounds that we do not need to allocate the same partitions as Spark, so I am not asking you to change the gate. But the description says this matches the `hash` expression's coverage and it does not, since `HashUtils.unsupportedReasonFor` rejects `decimal(precision > 18)` and `TimeType` at any depth. Worth saying so explicitly.

Last thing, this test only asserts a shuffle count and then calls `collect()`. Now that these keys run natively, is it worth comparing `spark_partition_id()` against Spark here too? This is the widest nested type coverage we have, and it is the one place a leaf type difference would show up across all of them at once.

andygrove (commented):
Thanks @viirya. I checked the two things this rests on and they both hold up. The native shuffle partitioner really does call `create_murmur3_hashes` with seed 42, the same entry point the `hash` expression uses, and `InsertMapSortInRepartitionExpressions` really does rewrite nested map keys recursively on 4.0+. Recursing through the same predicate so that a bad leaf disqualifies the whole key is the right shape for this, and the collation and interval fallbacks fall out of it for free.

Most of what I have is about the tests. The map cases do not exercise the normalization they are gated on, the partition assignment test can pass without Comet running the shuffle at all, and there are no null keys anywhere. Details inline. I would also like to see some numbers before we turn this on by default.

viirya:
Thanks for the detailed review — three of these were straight mistakes on my part.

**`withSQLConf` return type.** You're right that Spark 4.x declares `withSQLConf[T](...)(f: => T): T`, and my comment claiming it returns `Unit` was wrong. But the `val` form does not compile on 3.x: `SQLHelper.withSQLConf` there is `public default void withSQLConf(Seq, Function0<BoxedUnit>)`, and `-Pspark-3.5` fails with `value nonEmpty is not a member of Unit`. Since we build against both, I kept the `var` and fixed the comment to say what is actually going on. Added the missing blank lines.

**The interval test wasn't testing `CalendarIntervalType`.** Correct — `INTERVAL '1' MONTH` comes out as `YearMonthIntervalType`. Switched to `make_interval(...)`, so it now matches its comment and covers the type #5059 is about.

**The fuzz file does have nested complex types.** Also correct: `generateSchema` adds `struct<array<..>>` and `array<struct<..>>` when both flags are set, and maps are the only omission. Fixed my comment and the inherited one in `CometFuzzTestBase`.

**Map entry order.** Added a test with `map('a',1,'b',2)` vs `map('b',2,'a',1)` (and a three-entry pair) asserting the rows share a `spark_partition_id()` and match Spark.

**Parity test could pass without Comet.** Added `checkCometExchange(..., 1, true)` to both the shallow and deep cases.

**Null keys.** Added a test with null keys at four levels — top level, inside a struct, as an array element, and as a map value — with pairs of equal keys asserted into the same partition and compared against Spark.

On the `DataType::Struct(_)` branch: you are right that it recurses into `columns()` without consulting its own null mask while `List` and `Map` guard on `is_null(row_idx)`. I tried to construct a misroute through a plan-built null struct (both `if(true, NULL, named_struct(...))` and nulling a struct that has non-null children) and could not — Comet's null struct came out with null children in both cases, so the two agreed. So I cannot show it biting today, but the asymmetry with the other two branches is real, and the new tests pin the current behavior. Happy to file it separately if you would like it tracked.

**Default and performance.** Agreed, and I have flipped it to `false`. You are right that `struct<int, string>` is not the interesting case: the shapes that miss the vectorized element paths fall into `hash_list_array!`, which slices a one-element array and re-enters `create_murmur3_hashes` per element, and I have no numbers for that. Turning it on by default before measuring was the wrong call. I would like to do the benchmark as a follow-up issue and PR rather than hold this one; all the tests opt in explicitly now.

**On the decimal point.** You are right that the description overstated the parity with the `hash` expression — `HashUtils.unsupportedReasonFor` rejects `decimal(precision > 18)` and `TimeType` at any depth and the shuffle gate does not. I have said so explicitly in the description. Worth noting what I actually measured: with the config on, every complex column in that fuzz file has partition parity with Spark, including `array<decimal(36,18)>` and the structs containing it. `FuzzDataGenerator` builds decimals from `nextDouble()`, so the values stay in the range where the 16-byte LE `i128` and the minimal BE `BigDecimal` encodings agree — the divergence you describe is real in the code, just not reachable with the values that generator produces.

**`spark_partition_id()` in the fuzz test.** I tried this and backed it out. Pairing rows needs a stable identifier and neither attempt held up: `c0` has many duplicate values, so sorting by it does not pair rows deterministically, and `CAST(<key> AS STRING)` is not a faithful key for nested values — that version failed for **jvm** shuffle too, which this change does not touch, so it was my test construction rather than routing. Doing it properly looks like it needs a real row id, i.e. materializing the fuzz table with something like `monotonically_increasing_id()`, which means changing the shared fixture that all three variants (native + 2x jvm) run against. I would rather not do that inside this PR unless you want it here. Happy to add it now or file a follow-up — which would you prefer?

spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/CometShuffleExchangeExec.scala
```diff
@@ -424,6 +433,22 @@ object CometShuffleExchangeExec
...
+        supportedHashPartitioningDataType(elementType)
+      case MapType(keyType, valueType, _) if nestedHashPartitioningEnabled =>
+        // Map entry order is not semantically meaningful, so two equal maps must hash alike.
+        // Spark 4.0+ normalizes a map shuffle key by wrapping it in `mapsort(...)`, which is
+        // gated separately by CometMapSort (scalar map keys only) and, when unsupported, fails
+        // the expression check below. Earlier Spark versions insert no such normalization, so
+        // Comet would hash physical entry order and could route equal maps differently.
+        isSpark40Plus &&
+        supportedHashPartitioningDataType(keyType) &&
+        supportedHashPartitioningDataType(valueType)
```
sunchao, line 460: [P2] Rebase sliced map offsets before enabling map shuffle keys

Could we fix `spark_map_sort` for sliced maps before admitting this case? With the nested-hash config enabled on Spark 4.0, a native OFFSET below a map repartition can pass a MapArray whose first entry offset is nonzero. DataFusion's limit uses `batch.slice(skip, ...)`, and Arrow preserves the map's original entry offsets. The unified native shuffle evaluates `mapsort(m)` on that slice without a JVM round trip.

`map_sort.rs` builds its `take` indices only for the visible maps, but passes the original offsets to `MapArray::try_new`. For two two-entry maps, skipping the first leaves offsets `[2,4]`. Sorting creates two entries, so Arrow rejects the result with `Max offset of 4 exceeds length of entries 2`. The query fails instead of repartitioning. The map-sort code predates this PR, but the base gate rejected map hash keys, so the newly admitted path exposes it.

Please rebase the output offsets, or retain fallback for this case, and cover OFFSET followed by map repartition while asserting both operators run natively. This finding is source-derived. I have not executed the query.

viirya:
@sunchao thanks — confirmed, and it is a query failure rather than a fallback. I executed it.

Unit level, slicing two two-entry maps to drop the first:

```
sliced len=1 offsets=OffsetBuffer(ScalarBuffer([2, 4]))
Err = Arrow error: Invalid argument error: Max offset of 4 exceeds length of entries 2
```

End to end, with the nested-hash config on:

```sql
SELECT * FROM (SELECT * FROM tbl ORDER BY _1 LIMIT 10 OFFSET 5) DISTRIBUTE BY _2
```
```
CometExchange hashpartitioning(mapsort(_2#3383), 10), REPARTITION_BY_COL, CometNativeShuffle
+- CometTakeOrderedAndProjectExec(limit=15, offset=5, orderBy=[_1#3382 ASC NULLS FIRST])
   +- CometNativeScan parquet [_1#3382,_2#3383]

CometNativeException: Error inserting batch: Arrow error:
  Invalid argument error: Max offset of 30 exceeds length of entries 20
```

One correction to the framing, though: this is not only exposed by the newly admitted
path. `mapsort` is not user-callable, but Spark 4.0+ has two rules that insert it —
`InsertMapSortInRepartitionExpressions` and `InsertMapSortInGroupingExpressions`. The
second one needs no repartition at all, so with the nested-hash config **off** (its
default) and no map shuffle key anywhere:

```sql
SELECT _2, count(*) FROM (SELECT * FROM tbl ORDER BY _1 LIMIT 15 OFFSET 5) GROUP BY _2
```
```
HashAggregate(keys=[_groupingmapsort#3400], functions=[count(1)])
+- HashAggregate(keys=[_groupingmapsort#3400], functions=[partial_count(1)])
   +- CometTakeOrderedAndProjectExec(limit=20, offset=5, orderBy=[_1 ASC NULLS FIRST])
      +- CometNativeScan parquet [_1,_2]

CometNativeException: Invalid argument error: Max offset of 40 exceeds length of entries 30
```

So it reproduces on current `main` and is a pre-existing bug rather than a regression
from this PR. Since it stands on its own I have split the fix out into #5630 (issue
#5629) rather than carrying it here — it rebases the output offsets as you suggested,
with a Rust unit test and an end-to-end test on the group-by path, both verified to
fail without the fix.

I have taken the repartition-on-map coverage back out of this PR for now, because on
`main` that path is still gated off and the test would be dead code. Once #5630 merges
I will sync this branch and add it there, asserting both operators run natively as you
asked.

sunchao:
@viirya Agreed: this is a pre-existing `mapsort` defect, and #5567 adds another path that exposes it. The existing map-grouping rule is independent of the nested-hash option. Thanks for isolating the fix in #5630; I’m reviewing that PR now. Once this branch incorporates it, I’ll recheck the OFFSET → map-repartition coverage here, including the native-operator assertions.

sunchao (approved):
Thanks for syncing the mapsort fix and adding the OFFSET-to-map-repartition regression. At `e819c2f9`, the kernel matches the reviewed #5630 fix, and the new test asserts both native operators. I found no remaining P1/P2 in this update. I reused the earlier Arrow-only probe evidence. The new Spark test was not run here, and CI is still running.

viirya:
@andygrove Would you like to take another look before I merge this? Thanks!

viirya:
I will merge this after resolving the conflicts and passing CI.

sunchao (approved):
## Correctness

Re-reviewed `23d28ed751b6f12fbebcb803a7ac17f1d6f54ab0` against `c348f775d5b20d91d1989756b72f62f366d67d32`, following the approval of `e819c2f9`. The authored additions and removals are identical after excluding hunk positions and unchanged context. The nested hasher, Murmur3 kernel, positive-modulo helper, MapSort kernel/serde and native hash-partition branch also remain identical to the approved version. The surrounding checkout has changed, including Arrow 59.3/DataFusion 55 and base-side shuffle/Celeborn changes, so I checked the current integration rather than carrying forward the old runtime evidence.

The sliced-map finding remains addressed: MapSort rebuilds offsets from zero, and the OFFSET-to-map-repartition regression still asserts both native operators. Maintained Spark 3.5/4.0 source confirms ordered recursive hashing, seed preservation for nulls and empty collections, and Spark 4.0's recursive map normalization. The updated gate retains its version restriction and recursive leaf checks. This does not establish universal Spark partition parity: the previously documented high-precision decimal encoding difference remains. Existing null-parent-buffer and fixture-coverage caveats are unchanged, with no new reachable P1/P2 verified in this rebase.

I inspected the [current Spark 4.0 shuffle job](https://github.com/apache/datafusion-comet/actions/runs/33993755372/job/101381246013). All 477 tests passed, with no failures, cancellations or ignored tests. Its log includes the OFFSET, nested partition-assignment, map-order and nested-null cases. Checkout `5d2d2bf8` has the assigned base and head as parents and its complete Git tree equals this head. The final current-head CI refresh at `2026-09-05T22:49:34Z` showed 59 successful checks, seven skipped and five in progress, with no failed check at that cutoff. No local product test or benchmark ran, and the older remote probe was not accessed or reverified. Maintained Spark 3.4/4.1 sources remain unavailable.

## Performance

The rebase adds no new work to the nested hashing path, and the option remains disabled by default. The existing cost concern is unchanged: non-primitive list elements and general map entries allocate slices and recurse per element, with a child-vector allocation for structs. Please link the promised benchmark follow-up and keep the default off until representative deep-key measurements compare allocation cost and whole-shuffle throughput against the fallback path with matched results and plans. The successful correctness tests are not performance measurements.

## Design

The rebase preserves the existing division between recursive type admission and partition-expression conversion. In particular, admitting a map type does not bypass MapSort's scalar-key and compatibility checks. I found no new design issue in their integration.

## Abstraction & complexity

The update introduces no additional hashing abstraction or duplicate algorithm. The small recursive predicate and test opt-in helper remain sufficient for the feature, and the unchanged native hashing branch needs no further indirection.

viirya:
Thanks @andygrove @sunchao !

Issue #5574, opened 2026-08-31
Codegen dispatcher is unreachable from `convert`, so serdes that decline there never get a dispatch attempt


andygrove:
Part of #5572.

The dispatcher is only reachable from `getSupportLevel`. `dispatchIfFallback` is called from exactly two places, the `Unsupported` arm (`spark/src/main/scala/org/apache/comet/serde/QueryPlanSerde.scala:941`) and the non-opt-in `Incompatible` arm (`:968`). The `Compatible` arm just calls `handler.convert(...)` and propagates whatever it returns.

So a serde that reports `Compatible` and *then* declines inside `convert` gets no dispatch attempt, even when it mixes in `CodegenDispatchFallback`. Around 15 serdes decline that way via `withFallbackReason`, including two that already carry the mixin and therefore silently lose it on that path:

- `CometJsonToStructs` (`serde/structs.scala:195`)
- `CometArrayReverse` (`serde/arrays.scala:598`)

and, without the mixin, `CometCreateArray`, `CometGetArrayItem`, `CometGetArrayStructFields`, `CometArrayInsert`, `CometStaticInvoke`, `CometUuid`, `CometShuffle`, `CometAttributeReference`, `CometScalarFunction`.

`CometUnixTimestamp` is the clearest illustration: it makes the identical input-type check twice, once in `getSupportLevel` (`serde/datetime.scala:319`) and once in `convert` (`:329`). Only the first copy would benefit from adding the mixin.

Two ways to close this:

1. Move the checks into `getSupportLevel` serde by serde. Precise, no behavior change anywhere else, but it is a lot of small edits and nothing stops the pattern coming back.
2. Add `.orElse(dispatchIfFallback(handler, expr, inputs, binding).map(_._2))` after `handler.convert` in the `Compatible` arm. One line, and it fixes the pattern permanently.

Option 2 needs care. A `Compatible` serde's `convert` can legitimately return `None` because a *child* failed to serialize, not because of anything wrong with this node, and in that case retrying the whole subtree through the dispatcher is a real behavior change rather than a bug fix — though arguably a desirable one, since the dispatcher can evaluate a subtree containing an expression that has no native serde at all. It also needs the guard in the prerequisite issue first. Worth prototyping option 2 and measuring how often it changes a plan across the TPC-DS and Spark SQL suites before committing to it.

[...]

Issue #5544, opened 2026-08-29
Fix remaining reported issues for arrays of map


comphead:
@comphead, smaller follow-ups are reasonable, provided this PR preserves the previous safe behavior for the affected cases. The remaining **[P2]** concerns come from the actual BASE-to-HEAD changes:

- The TRY/bracket and collation examples contain folded map literals. [BASE declines those non-null complex literals](https://github.com/apache/datafusion-comet/blob/e0ab0a6fe60c05bd679654f0201ebe88319cc3a7/spark/src/main/scala/org/apache/comet/serde/literals.scala#L43-L64); [HEAD newly admits and rebuilds them](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/literals.scala#L43-L86), exposing the consumer limitations. That admission is not restricted to an outer array-of-map literal.
- The stateful example exercises the [new ANSI guard](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L623-L696), which serializes the left operand twice. [BASE has no second serialization](https://github.com/apache/datafusion-comet/blob/e0ab0a6fe60c05bd679654f0201ebe88319cc3a7/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L601-L639). This change affects nullable array lookups beyond literal expansion.
- The Slice example does construct an array of maps. The [new `deepNullable` casts](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L520-L534) change its nested map type, while [Slice still declares the original element type](https://github.com/apache/datafusion-comet/blob/67fbae59727d108dda1a875e5f00c33a545855e4/spark/src/main/scala/org/apache/comet/serde/arrays.scala#L465-L484). BASE passes that single map child through without the new cast.

So a narrow solution here is fine: restrict the newly enabled paths or retain safe fallback until the corresponding consumer fix lands; broader collation/collection improvements can remain separate PRs. The scope distinction above is from pinned source comparison, not an additional executed reproduction.

_Originally posted by @sunchao in https://github.com/apache/datafusion-comet/pull/5452#pullrequestreview-5058500152_

sunchao:
[...]

PR #5593, opened 2026-08-31
chore: warn when a Compatible serde declines in convert

andygrove:

Related to #5574.

That issue is about the dispatcher being unreachable from `convert`. This PR does not fix that — it adds the diagnostic that makes the affected cases visible, so the cleanup in #5574 can be scoped from evidence rather than from a static sweep.

## Rationale for this change

A serde whose `getSupportLevel` returns `Compatible` is promising that `convert` will succeed. When `convert` returns `None` instead, two things happen and neither is obvious:

- the enclosing operator falls back to Spark, and
- the JVM codegen dispatcher never gets a chance at the expression, because `dispatchIfFallback` is only reached from the `Unsupported` and `Incompatible` arms of `exprToProtoInternal` (`QueryPlanSerde.scala:968` and `:995`).

So a check that lives in `convert` instead of `getSupportLevel` silently costs the user a whole-operator fallback that the dispatcher could have absorbed. Today nothing surfaces that.

A warning on the `Compatible` arm when `convert` returns `None`, plus a new `CometSerdeInvariantSuite`.

The one piece of actual logic is the suppression rule. `convert` recurses into children, so a `Compatible` serde legitimately returns `None` when a *child* could not be serialized — `getSupportLevel` inspects the node, not the subtree. Without a guard, a single unsupported leaf would warn once for every ancestor on the way up.

The guard counts conversion failures on the planning thread and compares the counter across the `convert` call: if no nested `exprToProtoInternal` declined, the decline originated in this node. That is O(1). An earlier revision searched the subtree for a descendant carrying a fallback reason, which is O(n) per decline and therefore O(n·depth) along a failing path — a real cost here, since `createBalancedBinaryExpr` exists precisely because `And`/`Or` chains nest deeper than protobuf's 100-level recursion limit. The counter is also strictly more accurate: it catches children the serde synthesised inside `convert` (a `Cast` wrapper, say), which are not in the expression's subtree and so were invisible to the search. Verified equivalent on `CometArrayExpressionSuite` — both approaches flagged the same expressions.

Two categories warn without being a serde bug, and both are documented in the scaladoc:

- `getSupportLevel` is not given `inputs`, so an `inputs`-dependent decline cannot be moved there. `CometAttributeReference` is the only one today.
- Every serde that reports `Compatible` and then routes to `emitJvmCodegenDispatch` warns whenever the dispatcher declines, because the dispatcher's checks live there rather than in `getSupportLevel`. That is the whole `CometCodegenDispatch` class of serdes, and it fires both when `spark.comet.exec.scalaUDF.codegen.enabled=false` turns the dispatcher off wholesale and, under the default configuration, when the dispatcher cannot handle a particular expression. Those are genuine instances of the invariant violation; moving those checks up belongs with #5574, not here.

New `CometSerdeInvariantSuite` (5 tests, registered in both PR workflows): warns on a real `Compatible`-then-`None` path, stays quiet when only a child failed, stays quiet when the parent tags itself *and* a child failed, warns when the node is the only tagged expression, and warns with a placeholder when no reason was recorded. The tests capture WARN output through a log4j2 appender attached to the `QueryPlanSerde` logger.

On noise, which was the main thing I wanted to check before proposing this. Across two of the larger expression suites the warning fires three times in ~205 tests, and all three are true positives.

`CometExpressionSuite` (141 tests) produces one, from a test that turns the dispatcher off:

```
CometInitCap reported Compatible for initcap(_1#10853) but convert() returned None
(initcap: spark.comet.exec.scalaUDF.codegen.enabled=false; expression has no native path
so the plan falls back to Spark), so the enclosing operator falls back to Spark.
```

`CometArrayExpressionSuite` (64 tests) produces two. One is that same disabled-dispatcher category, on `CometScalaUDF`. The other fires under the default configuration, on the fixture added for #5544:

```
CometCreateMap reported Compatible for map(1, [10,20]) but convert() returned None
(map: codegen dispatch: unsupported output type MapType(IntegerType,StructType(
StructField(x,IntegerType,false),StructField(x,IntegerType,false)),false)), so the
enclosing operator falls back to Spark.
```

That last one is the case this PR exists to find. `CometCodegenDispatch.getSupportLevel` returns `Compatible()` without looking at the expression at all, and `convert` then declines on the output type — which is derivable from the node, so that check can move up to `getSupportLevel` and let the fallback path do something better than fail the whole operator. It turned up by running the diagnostic, not by reading code, which is the argument for landing it ahead of #5574.


sunchao (commented):
The counter adds thread-local work even with WARN disabled, while the PR base has neither this counter nor the new warning. Could you add a focused HEAD-versus-base planning comparison for one wide successful projection and one nested-fallback shape? Please match the hardware, build, Spark/JDK profile, warmup and inputs, and report planning time, allocations and warning counts with WARN on and off. Outside the timed work, please verify unchanged results and intended native/dispatcher/fallback routing, including that the successful projection reaches the native path. A synthesized child can be included in the nested case.

sunchao (approved):
## Correctness

Rechecked `68e7b4d8` against `7f1e0018` and the previously reviewed `783cccde`. The update merges main, and the authored patch has the same stable patch ID. The diagnostic logic and five tests are unchanged. I found no new or remaining verified P1/P2 issue in the merged context and approve this head. The planning-performance evidence remains unqualified, as noted below. The counter still suppresses inherited child failures without changing the returned conversion result or the existing native/dispatcher/fallback choices.

The maintained Spark 3.5 and 4.0 binding implementations return the original attribute when `allowFailures=true` cannot resolve it, which still matches the test's expected decline. Their `IsNull` child/null behavior is unchanged by this diagnostic wrapper. Input-dependent binding failures and dispatcher declines remain qualifications on what the warning means. The update does not establish that every warning identifies a faulty native serde or a completed operator fallback.

All five invariant tests passed in the inspected [Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/34132282264/job/101781989000), [4.0](https://github.com/apache/datafusion-comet/actions/runs/34132282264/job/101781988966), [4.1](https://github.com/apache/datafusion-comet/actions/runs/34132282264/job/101781989003) and [4.2](https://github.com/apache/datafusion-comet/actions/runs/34132282264/job/101781988971) expression jobs. They executed merge `d246d45f`, whose parents are the assigned base/head and whose tree equals the head. The snapshot has 62 successful, nine skipped and two failed checks. Both failed Spark 3.5 jobs stopped before checkout because the `amd64/rust` container pull returned HTTP 500, so those test groups did not execute. No local build or benchmark was run. Maintained Spark 3.4/4.1 sources remain unavailable, separate from CI coverage.

## Performance

The planning comparison requested in the [previous review](https://github.com/apache/datafusion-comet/pull/5593#pullrequestreview-5071340225) remains open. The unchanged code still reads the thread-local counter for every `Compatible` conversion with WARN disabled and updates it on declined conversions. On a local decline, serde-name and reason preparation also precede the logging guard. Spark's by-name logging defers the message interpolation, so that part should not be counted as unconditional formatting. The update supplies no matched HEAD/base planning time, allocation or warning-count results. The existing focused comparison is still needed to assess this planning-path cost. CI correctness results do not close that validation gap or demonstrate a regression.

## Design

The merge keeps the diagnostic separate from routing changes and retains the constant-time nested-failure check. It does not reintroduce the earlier subtree search or change the documented binding/dispatcher caveats. I have no new design request beyond resolving the existing validation question.

## Abstraction & complexity

No new abstraction or state was added in this update. The private wrapper, thread-local counter and warning helper are unchanged, and the test suite remains registered in both platform workflows. I found no additional simplification to request from this merge.

sunchao (approved):
## Correctness

Rechecked `9ee3f5130b5db0d7bc78ba208f6debad7f946280` against `424c31aa79d13fddf743ffa29bae3c6f146e6c5e`, including the update since [the prior approval](https://github.com/apache/datafusion-comet/pull/5593#pullrequestreview-5137079879). I found no new or remaining verified P1/P2 findings and approve this revision.

The new authored commit changes only the warning's scaladoc. It correctly removes the claim that warnings are silent with default settings: `CometCodegenDispatch` reports `Compatible` before the dispatcher checks individual expressions, so unsupported output types can decline during conversion even with dispatch enabled. Disabling dispatch globally is another trigger. The existing input-dependent attribute-binding exception remains documented. This update changes neither fallback routing nor the returned conversion result.

The counter, warning implementation and all five invariant tests remain unchanged. The main merge adds `ApplyFunctionExpression` registration and updates several expression handlers. Their nested conversions still use the counted entry point. In particular, rebuilt literals and synthesized array casts remain covered, and Unsupported dispatch fallbacks do not enter the Compatible warning arm. The 121-file base update was checked for these interactions. This is not an independent qualification of every incoming main change.

All five Linux expression jobs in [CI run 34378827828](https://github.com/apache/datafusion-comet/actions/runs/34378827828) executed and passed the five `CometSerdeInvariantSuite` tests, across Spark 3.4, 3.5, 4.0, 4.1 and 4.2. Each log records merge `fd6ad9f7`, with this base and head as parents, and its entire source tree equals head. These tests cover a real binding decline, nested-failure suppression and warning formatting. The stand-in handlers are diagnostic fixtures, not native execution or performance evidence. At `2026-09-09T19:24:33Z`, checks show **59 successful, nine skipped and one queued**, with no failures. The aggregate CI check is still pending. No local product test was run. Canonical binding, null-expression and lazy logging semantics were checked on maintained Spark 3.5/4.0. Maintained 3.4/4.1 source gaps remain.

## Performance

The [existing matched HEAD/base planning comparison](https://github.com/apache/datafusion-comet/pull/5593#pullrequestreview-5071340225) remains unprovided. Thread-local work occurs even when WARN is disabled. On the warning path, the serde name and recorded reason are prepared before the logging guard, although Spark's by-name message interpolation remains lazy. The PR body's warning counts and comparison with an abandoned subtree-search implementation do not quantify overhead against base. No measured regression establishes a new P2, and this approval does not qualify planning performance. The documentation-only increment adds no runtime work.

## Design

The clarification matches the current boundary: support admission lacks some information or checks that conversion can use. The warning is a diagnostic of that mismatch, not proof that every decline is a serializer bug. Before/after counter comparison continues to suppress ancestor warnings after a nested decline without inspecting the tree again. Moving dispatcher admission checks is still separate work. This PR does not implement it.

## Abstraction & complexity

No new abstraction was added. The existing counter remains scoped to each planning thread and compares only the interval around one handler conversion, so an earlier sibling failure does not suppress a later independent warning. The documentation now explains why expected fallbacks can reach this helper. No additional design or complexity P1/P2 is warranted.

andygrove:
Moving this to draft. I merged main and re-ran the noise check, and the result made me doubt the premise rather than confirm it.

The warning now fires three times across `CometExpressionSuite` and `CometArrayExpressionSuite` (~205 tests), and I no longer think any of the three is worth acting on. The message tells the reader to report the case from `getSupportLevel` as `Unsupported` or `Incompatible` so that a `CodegenDispatchFallback` serde can route it through the codegen dispatcher — but that only does something when the handler actually mixes in `CodegenDispatchFallback`, which is the one thing `dispatchIfFallback` tests. `CometCreateMap` is a `CometCodegenDispatch`, `CometScalaUDF` is a plain `CometExpressionSerde`, and `CometInitCap` is a `CometScalarFunction with NativeOptInAvailable`. None of them is one. Take the advice on any of the three and `dispatchIfFallback` no-ops, control falls through to `withFallbackReason(...); None`, and the user gets the identical Spark fallback with no warning attached. Nothing was lost, so there was nothing to report.

`CometCreateMap` is the clearest case and the one that changed my mind. Its `convert` delegates straight to the dispatcher, and the dispatcher is what declined — the output type has two struct fields named `x`, which it cannot write. Telling that serde to report `Unsupported` so the dispatcher can have a go is circular. `CometCodegenDispatch` backs around 62 serdes, so that circular shape is the dominant source of warnings here rather than an edge case.

The rationale in the description is still right as far as it goes: a `Compatible` serde that declines inside `convert` does skip a dispatcher attempt. I had the population wrong. It only costs anything when the serde is a `CodegenDispatchFallback`, and the check I added does not test for that, which is why it fires exclusively on cases where the dispatcher was never reachable to begin with.

The narrow fix is to gate the warning on `handler.isInstanceOf[CodegenDispatchFallback]`. That is one line, and it would have suppressed all three. What stops me merging is what is left afterwards: silence across all 205 tests. The argument for landing this ahead of #5574 was that it would let the cleanup be scoped from evidence, and with the gate in place the evidence is empty. It stops being a diagnostic and becomes a guard against a violation I have not shown exists — and if that is all it is, a `logWarning` in the planner is the wrong home for it, since an assertion in `CometSerdeInvariantSuite` gets the same protection without shipping a warning nobody can act on.

So what decides this is whether a real instance exists: a `CodegenDispatchFallback` serde that reports `Compatible` and then declines in `convert` for a reason depending only on the node. There are 32 serdes mixing that trait in — `CometCast`, `CometLike`, `CometConcatWs`, `CometRound`, `CometSequence`, `CometStringLPad`/`RPad` and the rest. I want to work through those before deciding. If one turns up, I will add the gate and put this back up for review on its original terms. If none does, I would rather close this and fold the invariant into #5574 as a test than merge a warning that has only ever fired on non-problems.

PR #5594, opened 2026-08-31
chore: rename .claude directory to vendor-neutral .ai

### LINKED ISSUE #5592: Rename the .claude directory to a vendor-neutral .ai

Comet's agent skills live under `.claude/skills/`. The skills themselves are vendor-neutral
Markdown -- there is nothing Claude-specific in their content -- but the directory name ties a
shared project asset to one LLM vendor.

Both `apache/datafusion` and `apache/datafusion-python` have already moved to a vendor-neutral
layout:

| | `apache/datafusion` | `apache/datafusion-python` |
|---|---|---|
| Skills | `.ai/skills/<name>/SKILL.md` | `.ai/skills/<name>/SKILL.md` |
| `.claude/skills` | (absent) | symlink to `../.ai/skills` |
| `CLAUDE.md` | symlink to `AGENTS.md` | symlink to `AGENTS.md` |

Adopt the same layout as the sibling repos:

- Move `.claude/skills` to `.ai/skills`.
- Keep `.claude/skills` as a symlink to `../.ai/skills`, so agents that only look in the
  vendor-specific location still discover the skills with no contributor setup.
- Add a `CLAUDE.md` symlink to the existing `AGENTS.md`, which is already the vendor-neutral
  agent guide but does not currently mention skills at all.
- Update the RAT excludes in `pom.xml`.

One migration note for contributors: `CLAUDE.md` is currently in `.gitignore` so that personal
untracked instruction files can live at the repo root. Tracking a `CLAUDE.md` symlink means
dropping that entry, and anyone holding a personal root `CLAUDE.md` will hit a checkout conflict
on pull and will need to move or delete it. Both sibling repos accepted the same tradeoff.

## Additional context

None.

andygrove:

Closes #5592.

## Rationale for this change

Comet's agent skills live under `.claude/skills/`. Their content is vendor-neutral Markdown, so
the directory name ties a shared project asset to one LLM vendor for no reason.

`apache/datafusion` and `apache/datafusion-python` have both already moved to `.ai/skills/`, with
symlinks covering the vendor-specific discovery paths. This adopts the same layout, following
`datafusion-python`, which keeps the `.claude/skills` symlink so the skills stay discoverable
with no contributor setup.

- Move `.claude/skills` to `.ai/skills` (7 `SKILL.md` files, tracked as renames).
- Add `.claude/skills` -> `../.ai/skills`, and `CLAUDE.md` -> `AGENTS.md`.
- Drop the `CLAUDE.md` entry from `.gitignore`, which would otherwise shadow the new symlink.
- Document the layout in a new `## Skills` section in `AGENTS.md`, which previously did not
  mention skills at all, and note that the symlinks are compatibility shims rather than places
  to add content.
- Add `.ai/**` to the RAT excludes in `pom.xml`, keeping `.claude/**` for the symlink.
- Fix three now-stale `CLAUDE.md` references in skill prose. Two in `wire-datafusion-function`
  meant the reader's own agent config rather than a repo file, so they now say "agent
  instructions / project memory"; the one in `suggest-native-expression` pointed at `CLAUDE.md`
  for build commands, which actually live in
  `docs/source/contributor-guide/development.md`.

Skill content is otherwise unchanged, and no documentation needed updating -- the contributor
guide refers to skills by name, never by path.

### Note for contributors

`CLAUDE.md` was in `.gitignore` so contributors could keep a personal untracked instruction file
at the repo root. Tracking a `CLAUDE.md` symlink means dropping that entry, so anyone holding a
personal root `CLAUDE.md` will hit a checkout conflict on pull and will need to move or delete
it. Both sibling repos accepted the same tradeoff.

No tests -- this is a file move plus build-config changes with no runtime behavior. Verified
locally:

- `./mvnw -N apache-rat:check` passes. The report shows `AL CLAUDE.md`, i.e. RAT follows the
  symlink to `AGENTS.md` and finds the ASF header.
- Both symlinks are staged with mode `120000` and resolve correctly; `.claude/skills` lists all
  seven skills.
- `npx prettier "**/*.md" --check` flags exactly the same files before and after this change, so
  the move introduces no new formatting diffs.

I was not able to run `dev/release/run-rat.sh` locally because it downloads the RAT jar from
Maven Central, which is blocked in my environment. That check should be unaffected: the moved
files are unchanged and all carry ASF headers, and neither the old nor the new path appears in
`dev/release/rat_exclude_files.txt`.


viirya (approved):
The move itself looks good — all seven skills came across as clean 100%-similarity renames, both symlinks are staged with the right `120000` mode, and the three stale `CLAUDE.md` references are the complete set (I grepped `main`; those three are the only hits outside the skills dir itself).

One problem, and it's in the one check you flagged as untestable locally: **`dev/release/run-rat.sh` will fail on this branch.**

I had the same Maven Central block, but RAT 0.16.1 was in my local m2 cache, so I was able to assemble a classpath and run the real thing over a `git archive` tarball:

| | `check-rat-report.py` exit |
|---|---|
| `origin/main` | 0 — no unapproved licenses |
| this branch | 1 — 2 unapproved |

```
NOT APPROVED: .claude/skills (comet-test/.claude/skills): false
NOT APPROVED: CLAUDE.md    (comet-test/CLAUDE.md):        false
```

The two RAT paths diverge, which is exactly why the check you ran passed:

- `./mvnw -N apache-rat:check` scans the **working tree**, so it follows the symlink to `AGENTS.md`, finds the ASF header, and reports `AL CLAUDE.md`. Your observation was correct.
- `run-rat.sh` scans a **`git archive` tarball** (`create-tarball.sh:115,118`). There a symlink is a 0-byte symlink entry with no content to follow, so the header comes back empty → `Unknown license` → unapproved.

The new `.ai/**` exclude in `pom.xml` doesn't help here, since that only configures the Maven plugin; the release script reads `dev/release/rat_exclude_files.txt`, where neither symlink is excluded.

Adding these two lines to `dev/release/rat_exclude_files.txt` takes it back to exit 0 (verified):

```
CLAUDE.md
.claude/skills
```

Worth noting that `datafusion-python`, which this PR follows, already carries exactly these two entries as the last two lines of its own `rat_exclude_files.txt` (plus `**/SKILL.md`) — so this is the precedent completing itself rather than a new requirement. `apache/datafusion` doesn't have them, but it also has no `.claude` directory, so it isn't a counterexample.

Since PR CI only exercises the Maven path, this would stay hidden until someone cuts a release candidate — cheaper to fix now.

One non-blocking note: the `.gitignore` removal and its checkout conflict for contributors holding a personal root `CLAUDE.md` is well documented in the PR description, but it'd be worth a line in the release notes too, since the failure mode (pull refusing to overwrite an untracked file) is confusing if you hit it cold.

sunchao (approved):

ziting-openai (approved):
## Summary

Reviewed head `a0ce9bd565629f11d9ede53cc620b6e69626a39d`, covering the directory migration, compatibility paths, and license/release integration. No P1 or significant P2 findings remain. Approve.

## Prior state and problem

The seven shared skills lived under `.claude/skills`. This change gives them a vendor-neutral home while preserving the existing discovery path and sharing the root agent instructions.

## Design approach

The canonical files move to `.ai/skills`. `.claude/skills` points to `../.ai/skills`, and `CLAUDE.md` points to `AGENTS.md`. The Maven RAT configuration excludes the new skill directory, while the release RAT exclusions cover the two symlink entries.

I verified all 13 changed paths against their Git blobs and checked both symlinks' `120000` modes and resolved targets. Both skill paths expose all seven files. Five skills are byte-for-byte unchanged from the base, and the remaining two contain exactly the three intended reference edits. The replacement development-guide path exists.

Both license paths passed locally with RAT 0.16.1: the nonrecursive Maven check reported zero unapproved licenses, and the unchanged release script accepted the exact-head GitHub source archive. Applying the base exclusion list to that same archive report rejects exactly `.claude/skills` and `CLAUDE.md`, confirming that the current-head additions address the earlier review concern. The archive report contains 2,048 resources.

Current-head CI reports 64 successful checks and nine skips, with no failures or pending checks. Local validation did not include a full native/JVM build, agent-client discovery, or a Windows checkout.

## Key design decisions

The symlinks retain one canonical copy of each instruction file. Keeping working-tree and archive RAT handling separate is necessary because archive symlink entries do not contain the target's license header.

## Implementation sketch

1. Move the seven skills and update the three stale prose references.
2. Add the compatibility symlinks and document the canonical locations.
3. Preserve Maven license-check behavior and exclude the two archive symlink entries in the release checker.

A contributor with a personal untracked root `CLAUDE.md` must move it before checking out this change, as documented in the PR. No Spark execution code, runtime configuration, dependency version, or JVM/native interface changes.

## Suggested improvements

No additional changes requested at this head.

comphead (approved):
Thanks @andygrove lets give it a try

PR #5595, opened 2026-08-31
chore: move dead and defensive serde guards out of convert

andygrove:

Part of #5574. First of a few small PRs moving `convert`-side declines into `getSupportLevel`; this one is deliberately the zero-risk slice.

## Rationale for this change

A serde that reports `Compatible` from `getSupportLevel` and then returns `None` from `convert` breaks the serde invariant, and it is also the one shape the codegen dispatcher cannot see — `dispatchIfFallback` is reached only from the `Unsupported` and `Incompatible` arms of `exprToProtoInternal` (`QueryPlanSerde.scala:941` and `:968`), so a decline from inside `convert` always costs a whole-operator Spark fallback.

Auditing all 35 `convert`-side declines, five are node-local *and* had no user-visible effect, because each guard was already dead or unreachable. Those are worth moving first: the change is provably behavior-preserving, and it shrinks the surface the harder follow-ups have to reason about.

- **`CometFromUnixTime`** (`unixtime.scala`) and **`CometUnixTimestamp`** (`datetime.scala`) re-checked in `convert` exactly what `getSupportLevel` had already reported as `Unsupported`. `CometFromUnixTime` additionally mixes in `CodegenDispatchFallback`, so the `Unsupported` result is either dispatched or fails before `convert`; either way the declining input never reached the duplicated check. Removed.
- **`CometShuffle`** (`collectionOperations.scala`) and **`CometUuid`** (`nondetermenistic.scala`) guarded a `randomSeed` that their own comments note is always defined in a resolved plan. Moved to `getSupportLevel`; `convert` now reads the seed directly, matching the existing pattern in `CometKnownFloatingPointNormalized` and `CometRandStr`.
- **`CometScalarSubquery`** screened the data type in `convert`. Moved to `getSupportLevel`. The `serializeDataType` backstop stays — that is a genuinely different predicate from `supportedDataType`, as the scaladoc on `serializeDataType` spells out.

Also drops four `withFallbackReason` imports that became unused.

**Deliberately not included:** `CometLiteral`. Its `convert` guards look like the same pattern but are reachable, not dead — `supportedDataType` admits `CalendarIntervalType` while the literal value match has no arm for it, so a non-null calendar-interval literal really does fall through to `case dt =>`. Left as-is.

No new tests, because there is no new behavior — the point of this slice is that the removed guards were unreachable. Covered by existing suites:

- `CometTemporalExpressionSuite`, `CometUuidExpressionSuite`, `CometArrayExpressionSuite` — 92 tests
- `CometSqlFileTestSuite` (includes `from_unix_time_enabled.sql`, `to_unix_timestamp_time_parser_policy.sql`, `to_unix_timestamp_time_parser_policy_corrected.sql`) and `CometExpressionSuite` — 607 tests

All green on Spark 4.1.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/serde/datetime.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/datetime.scala
@@ -309,13 +309,18 @@ object CometUnixTimestamp extends CometExpressionSerde[UnixTimestamp] {
   }
 
   override def getSupportLevel(expr: UnixTimestamp): SupportLevel = {
-    if (DatetimeCollation.hasNonDefaultCollation(expr)) {
-      Incompatible(Some(collationReason))
-    } else if (isSupportedInputType(expr)) {
-      Compatible()
-    } else {
+    // The input type is screened ahead of the collation check on purpose. A non-date/timestamp
+    // input has no native path at all, so it must report `Unsupported` rather than
+    // `Incompatible`: the latter is waved straight through to `convert` when
+    // `spark.comet.expression.UnixTimestamp.allowIncompatible=true`, and the native kernel then
+    // raises an execution error on the string child instead of falling back to Spark.
+    if (!isSupportedInputType(expr)) {
       val inputType = expr.children.head.dataType
       Unsupported(Some(s"unix_timestamp does not support input type: $inputType"))
+    } else if (DatetimeCollation.hasNonDefaultCollation(expr)) {
+      Incompatible(Some(collationReason))
+    } else {
+      Compatible()
     }
   }
 
--- a/spark/src/main/scala/org/apache/comet/serde/datetime.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/datetime.scala
@@ -27,7 +27,6 @@ import org.apache.spark.sql.types.{CalendarIntervalType, DataType, DateType, Dou
 import org.apache.spark.unsafe.types.UTF8String
 
 import org.apache.comet.CometConf
-import org.apache.comet.CometSparkSessionExtensions.withFallbackReason
 import org.apache.comet.expressions.{CometCast, CometEvalMode}
 import org.apache.comet.serde.CometGetDateField.CometGetDateField
 import org.apache.comet.serde.ExprOuterClass.Expr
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/org/apache/comet/serde/datetime.scala
```diff
@@ -324,12 +323,7 @@ object CometUnixTimestamp extends CometExpressionSerde[UnixTimestamp] {
       expr: UnixTimestamp,
       inputs: Seq[Attribute],
       binding: Boolean): Option[ExprOuterClass.Expr] = {
-    if (!isSupportedInputType(expr)) {
-      val inputType = expr.children.head.dataType
-      withFallbackReason(expr, s"unix_timestamp does not support input type: $inputType")
-      return None
-    }
-
+    // getSupportLevel reports an unsupported input type before reaching here, so no re-check.
```
sunchao, line 331: [P2] Keep string-input rejection ahead of the collation opt-in

On Spark 4.0, `unix_timestamp(s, collate('yyyy-MM-dd', 'UTF8_LCASE'))` with a non-foldable plain `STRING` column `s` makes `getSupportLevel` return `Incompatible` before it checks `isSupportedInputType`. With `spark.comet.expression.UnixTimestamp.allowIncompatible=true`, `exprToProto` therefore calls this method. The deleted guard returned `None` for that input. This now serializes the string child into the native UnixTimestamp expression. The format is not serialized, so no collated scan or Collate serializer is required. Native `SparkUnixTimestamp` only accepts date/timestamp inputs, and its unsupported-string error propagates instead of falling back to Spark. Could we retain the input-type rejection even when collation selects `Incompatible`, and cover this live-string/collated-format case in a regression test? This path is source-traced. The SQL witness was not executed.

andygrove: Good catch, and it reproduces. `getSupportLevel` checked collation first, so `unix_timestamp(s, 'yyyy-MM-dd HH:mm:ss' COLLATE UTF8_LCASE)` over a plain `STRING` column returned `Incompatible` before it ever looked at the input type, and with `allowIncompatible=true` that goes straight to `convert` — which no longer has the guard. Running it on this branch gives `CometNativeException: unix_timestamp does not support input type: Utf8` instead of a fallback.

Fixed in 559c0f9 by screening the input type ahead of the collation check: a non-date/timestamp input has no native path at all, so it should report `Unsupported` regardless of collation, and `allowIncompatible` should never be able to reach it. Collation still reports `Incompatible` for the input types that do have a native path, so the opt-in behavior there is unchanged. Added the live-string/collated-format case to `CometTemporalExpressionSuite` — it fails on the parent commit with the native error above and passes with the fix.

sunchao (approved):
Re-reviewed `559c0f9a9746595567dc5afccb47dfbc93e358c1`. The previously reported string-input/collation guard issue is fixed, with no new P1/P2 findings. Current CI shows 62 successful checks and 9 skipped. No local tests were run.

andygrove:
Merged. Thanks @sunchao

PR #5598, opened 2026-08-31
chore: bench additional scalar functions

coderfender:
Part of #5396

Closes #.

  - benches/rlike.rs (RLike PhysicalExpr via .evaluate)
  - benches/get_struct_field.rs (both PhysicalExpr)
  - json: benches/from_json.rs (FromJson PhysicalExpr)

benches only


native/spark-expr/Cargo.toml
```diff
@@ -244,6 +244,23 @@ name = "abs"
...
+[[bench]]
+name = "create_named_struct"
+harness = false
+
+[[bench]]
+name = "get_struct_field"
+harness = false
+
+[[bench]]
+name = "from_json"
```
sunchao, line 263: [P1] Separate the `from_json` and `modulo` benchmark entries

Could you give `from_json` its own `harness = false` and start a new `[[bench]]` table before `modulo`? Both `name` assignments currently belong to the same table, so Cargo rejects the native workspace manifest before any target can compile. This blocks ordinary native builds as well as these benchmarks.

coderfender: it was a bad residue from merge conflict. Fixed it

sunchao (approved):

sunchao:
Merged, thanks @coderfender !

PR #5600, opened 2026-08-31
feat: route `round` on float/double through the codegen dispatcher

### LINKED ISSUE #5576: `round` on float/double falls back to Spark, while `bround` already uses the codegen dispatcher
Part of #5572.

`round` on a float or double falls the whole projection back to Spark:

```scala
case _: FloatType | DoubleType =>
  // ... Spark uses BigDecimal for rounding float/double ...
  Unsupported(Some("Comet does not support Spark's BigDecimal rounding"))
```

`spark/src/main/scala/org/apache/comet/serde/arithmetic.scala:376-390`

The reasoning in that comment is sound — Spark rounds floating point through `BigDecimal`, which goes via `java.lang.Double.toString()` and its rounding algorithm, and the native path cannot reproduce it. But "we can't do this natively" is precisely the case the dispatcher exists for: running Spark's own `doGenCode` inside the Comet pipeline gives the `BigDecimal` behavior byte for byte.

`bround` already does exactly this — `CometBRound` is a plain `CometCodegenDispatch[BRound]` (`serde/math.scala:269`) — so the two sibling functions behave completely differently today for no principled reason.

`round(double_col, n)` is everyday analytics SQL, which makes this probably the highest-traffic single fallback in the sweep. `docs/source/user-guide/latest/expressions.md` documents it as "Float/double inputs fall back".

Fix: mix `CodegenDispatchFallback` into `CometRound`. `FloatType` and `DoubleType` are both in `CometBatchKernelCodegen.isSupportedDataType`, so `canHandle` admits it.

Leave the negative-scale decimal arm alone for now — it is only reachable with `spark.sql.legacy.allowNegativeScaleOfDecimal=true` and is tracked as deferred on the EPIC.

andygrove:

Closes #5576.

## Rationale for this change

`round` on a float or double fell the whole projection back to Spark:

```scala
case _: FloatType | DoubleType =>
  Unsupported(Some("Comet does not support Spark's BigDecimal rounding"))
```

The reasoning behind that is sound — Spark rounds floating point through a `BigDecimal` built from `java.lang.Double.toString()`, and no native kernel reproduces that. But "we can't do this natively" is precisely what the JVM codegen dispatcher is for: running Spark's own `RoundBase.doGenCode` inside the Comet pipeline gives that behavior byte for byte. `bround` — the sibling function with the *same* `doGenCode` — already dispatches, so the two behaved completely differently for no principled reason.

`round(double_col, n)` is everyday analytics SQL. TPC-DS q78 is a concrete example: it goes from **72/74 eligible operators (97%) to 74/74 (100%)**, and `10 native, 0 codegen dispatch` to `12 native, 1 codegen dispatch` — the projection that used to fall back was also carrying two other expressions that could have run natively.

- `CometRound` mixes in `CodegenDispatchFallback`, adds `getUnsupportedReasons()`, and reuses those reason strings in `getSupportLevel`. Integral and non-negative-scale decimal inputs are unaffected and still take the native path.
- The issue suggested leaving the negative-scale decimal arm alone. It turns out to dispatch cleanly — `canHandle` admits it and the result matches Spark — so both `Unsupported` arms now stay in the Comet pipeline rather than falling back. There is a test pinning that.
- Removes the `Float32`/`Float64` arms of the native `spark_round`. They were unreachable: `CometRound.convert` is the only thing that emits the `round` scalar function and it never emits it for those types. Their unit tests go with them.
- `expressions.md`: `round` is now `Hybrid` rather than `Native`, and the note says where float/double actually run.
- Regenerated the two q78 plan-stability goldens.

`round.sql` previously marked every query `expect_fallback(BigDecimal rounding)`. Those are now plain `query` records, i.e. `checkSparkAnswerAndOperator`, which fails if anything falls back. The fixture is also extended with a float column, NaN / ±Infinity, null scale, the one-arg form, and the two pathological doubles named in the source comment — `-5.81855622136895E8` and `6.1317116247283497E18` — whose shortest decimal representation rounds differently than the exact binary value. Those are the cases a native implementation would get wrong, so they are the ones worth pinning.

New `CometExpressionSuite` test for negative-scale decimals under `spark.sql.legacy.allowNegativeScaleOfDecimal=true` (built via the DataFrame API, since the type cannot be spelled in SQL syntax).

Run locally:

- `CometSqlFileTestSuite` round fixture and the three `CometExpressionSuite` round tests, on Spark 4.1 and 3.4.
- Both plan-stability suites in check mode on Spark 3.4 and 4.2; goldens regenerated across 3.4 / 3.5 / 4.0 / 4.1 / 4.2, and only the two base q78 files changed — the version-specific directories pruned as duplicates, so every supported version produces the same plan.
- `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and the `spark-expr` round unit tests; `spotless:check` and `scalastyle:check`.

### DIFF
--- a/docs/source/user-guide/latest/expressions.md
+++ b/docs/source/user-guide/latest/expressions.md
@@ -465,7 +465,7 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `random` | ✅ | Native | Alias for `rand` (Spark 4.0+); seed must be a literal |
 | `randstr` | ✅ | Native | Random string (Spark 4.0+); length and seed must be literals |
 | `rint` | ✅ | Native |  |
-| `round` | ✅ | Native | Float/double inputs fall back |
+| `round` | ✅ | Hybrid | Float/double inputs route through the JVM codegen dispatcher; other types run natively |
 | `sec` | ✅ | Native |  |
 | `shiftleft` | ✅ | Native |  |
 | `sign` | ✅ | Native |  |
--- a/native/spark-expr/src/math_funcs/round.rs
+++ b/native/spark-expr/src/math_funcs/round.rs
@@ -19,12 +19,9 @@ use crate::arithmetic_overflow_error;
 use crate::math_funcs::utils::{get_precision_scale, make_decimal_array, make_decimal_scalar};
 use arrow::array::{Array, ArrowNativeTypeOp};
 use arrow::array::{Int16Array, Int32Array, Int64Array, Int8Array};
-use arrow::datatypes::{DataType, Field};
+use arrow::datatypes::DataType;
 use arrow::error::ArrowError;
-use datafusion::common::config::ConfigOptions;
 use datafusion::common::{exec_err, internal_err, DataFusionError, ScalarValue};
-use datafusion::functions::math::round::RoundFunc;
-use datafusion::logical_expr::{ScalarFunctionArgs, ScalarUDFImpl};
 use datafusion::physical_plan::ColumnarValue;
 use std::{cmp::min, sync::Arc};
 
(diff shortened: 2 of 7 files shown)


comphead (approved):

andygrove:
Merged. Thanks @comphead

PR #5604, opened 2026-09-01
test: verify Celeborn reflection compatibility

### LINKED ISSUE #5530: Native Celeborn shuffle: add a Celeborn test dependency so the reflection against its internals is verified
Following up on PR #5513 (part of #5352). The native Celeborn shuffle path now carries roughly 2,500 lines that reflect on Celeborn's private internals, and Celeborn is not a dependency of the build in any scope. `grep -rn celeborn` across every `pom.xml` returns nothing, and the only references to Celeborn types in `spark/src/main` are class-name string literals in `CelebornShufflePusherFactory.scala` and one `Class.forName` in `CelebornTransportCallbackTracker.java`.

That means every test fixture is a hand-written Scala stand-in shaped to match what the production code expects. `CelebornShufflePartitionPusherSuite` defines `RecordingCelebornPushClient`, `AsyncRecordingCelebornPushClient`, `TransportRecordingCelebornPushClient`, `RecordingCelebornTransportClientFactory`, `RecordingCelebornTransportClientPool`, `RecordingCelebornTransportClient`, `RecordingCelebornTransportResponseHandler`, `RecordingCelebornTransportRequest`, `RecordingCelebornTransportCallback`, `RecordingCelebornPushState`, `RecordingCelebornInFlightTracker`, and `RecordingCelebornRetryExecutor`. Fixtures and production code come from the same mental model, so a divergence between that model and real Celeborn is structurally invisible to the suite.

To be clear, I checked every reflected member against the real v0.6.0 and v0.7.0 sources and they all match today. The reflection targets in the PR are accurate. `ShuffleClientImpl.pushStates`, `pushDataRetryPool`, `cryptoHandler`, `getPushState`, `pushOrMergeData`, `computeBatchCRC`, `mapperEnd`, `cleanup`, and `getDataClientFactory` all resolve, as do `TransportClientFactory.clientBootstraps` and `connectionPool`, `ClientPool.clients` and `locks`, `TransportClient.channel` and `getHandler`, `TransportResponseHandler.outstandingPushes`, `PushState.exception` and `inFlightRequestTracker`, `InFlightRequestTracker.totalInflightReqs`, and `PushRequestInfo.callback`. The four-argument versus five-argument `mapperEnd` difference between 0.6 and 0.7 is handled correctly too.

My concern is that nothing keeps them matching. A Celeborn point release that renames a private field, or that starts mutating `outstandingPushes` through one of the roughly thirty `ConcurrentHashMap` methods `CallbackTrackingRequests` does not override, breaks this at runtime with no compile error and no failing test. `CallbackTrackingRequests` extends `ConcurrentHashMap` but delegates to a separate backing map, so anything it does not override sees an empty map. It happens to override everything `TransportResponseHandler` uses in 0.6.0, which is why it works, but that is a coincidence rather than a contract.

Would you consider adding a `provided`-scope Celeborn dependency plus one CI job that runs these suites against the real classes for 0.6 and 0.7? Even a reflection-only smoke test that asserts every field name and type resolves would convert a silent runtime break into a build failure, which seems worth a lot here given how much behavior depends on those names.

pingzh:

Closes #5530.

## Rationale for this change

The native Celeborn shuffle path reflects on private Celeborn members. Its ordinary unit tests use hand-written stand-ins, so they cannot detect when a released Celeborn client changes one of those fields or method signatures.

- Add an opt-in `celeborn-reflection-compatibility` Maven profile with a test-scoped shaded Celeborn client dependency and an isolated test source directory.
- Add a reflection-only suite that verifies every Celeborn field and method used by the partition pusher and callback tracker, including the version-specific `mapperEnd`, CRC, and crypto differences.
- Run the suite in a dedicated Linux CI matrix against Celeborn 0.6.0 and 0.7.0.
- Keep the compatibility suite out of normal builds and the generic suite matrix.

- Clean compatibility-suite runs against Celeborn 0.6.0 and 0.7.0 (4 tests each).
- Clean profile-free test compilation with Spark 3.5 and Scala 2.12, confirming the isolated suite is not compiled normally.
- Spotless formatting check.
- `python3 dev/ci/check-suites.py`.
- Workflow YAML parsing and `git diff --check`.

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -176,6 +176,49 @@ jobs:
       - name: Compile (skip tests)
         run: ./mvnw -B install -DskipTests -Dmaven.test.skip=true -Pspark-4.1
 
+  celeborn-reflection-compatibility:
+    needs: lint
+    name: Celeborn ${{ matrix.celeborn_version }} reflection compatibility
+    runs-on: ubuntu-24.04
+    container:
+      image: amd64/rust
+      env:
+        JAVA_TOOL_OPTIONS: --add-exports=java.base/sun.nio.ch=ALL-UNNAMED --add-exports=java.base/sun.util.calendar=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED
+    strategy:
+      fail-fast: false
+      matrix:
+        celeborn_version: ["0.6.0", "0.7.0"]
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Setup Rust & Java toolchain
+        uses: ./.github/actions/setup-builder
+        with:
+          rust-version: ${{ env.RUST_VERSION }}
+          jdk-version: 17
+
+      - name: Cache Maven dependencies
+        uses: actions/cache@v6
+        with:
+          path: |
+            ~/.m2/repository
+            /root/.m2/repository
+          key: ${{ runner.os }}-java-maven-${{ hashFiles('**/pom.xml') }}-celeborn-${{ matrix.celeborn_version }}
+          restore-keys: |
+            ${{ runner.os }}-java-maven-
+
+      - name: Verify reflected Celeborn internals
+        env:
+          SPARK_LOCAL_HOSTNAME: localhost
+          SPARK_LOCAL_IP: 127.0.0.1
+        run: |
+          SPARK_HOME="$GITHUB_WORKSPACE" ./mvnw -B clean test \
+            -Pspark-3.5,scala-2.12,celeborn-reflection-compatibility \
+            -Dceleborn.version="${{ matrix.celeborn_version }}" \
+            -Dtest=none \
+            -Dsuites=org.apache.comet.shuffle.CelebornReflectionCompatibilitySuite \
+            -DfailIfNoTests=false
+
   # Build native library once and share with all test jobs
   build-native:
     needs: lint
--- a/dev/ci/check-suites.py
+++ b/dev/ci/check-suites.py
@@ -36,6 +36,7 @@ if __name__ == "__main__":
         "org.apache.comet.parquet.ParquetReadFromS3Suite", # manual test suite
         "org.apache.comet.IcebergReadFromS3Suite", # manual test suite
         "org.apache.comet.cloud.s3.CometS3CredentialBridgeSuite", # manual test suite
+        "org.apache.comet.shuffle.CelebornReflectionCompatibilitySuite", # dedicated version matrix
         "org.apache.spark.sql.comet.CometPlanStabilitySuite", # abstract
         "org.apache.spark.sql.comet.ParquetDatetimeRebaseSuite", # abstract
         "org.apache.comet.exec.CometColumnarShuffleSuite" # abstract
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,8 @@ under the License.
       property without an error.
     -->
     <delta.version>4.1.0</delta.version>
+    <!-- Used only to verify reflected Celeborn internals; never bundled with Comet. -->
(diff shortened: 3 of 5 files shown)


ziting-openai (approved):
## Summary

Reviewed head `cbaa94dd31ae0fa9bf39241418133597d219e6b4` against `main` at `6dedcfeee8777b578570e6304e514cc359df1de3`, using five independent passes covering reflected contracts, version compatibility, test execution, initialization and ownership boundaries, and regression coverage. **0 P1 / 0 significant P2 findings; approve.** Production shuffle behavior is unchanged.

## Prior state and problem

The partition pusher and callback tracker depend on private Celeborn fields and public method signatures. Tests using stand-ins can validate Comet's behavior while missing a mismatch with the released client jars.

## Design approach

An opt-in Maven profile adds the shaded Celeborn client as a test dependency and registers a separate test source directory. Four reflection-only tests check the released artifact, partition-pusher signatures, push-state counters, and transport callback interfaces. A dedicated Linux matrix runs them against Celeborn 0.6.0 and 0.7.0 with Spark 3.5 / Scala 2.12.

The checked signatures match the partition pusher and callback tracker, including the four- versus five-argument `mapperEnd`, the 0.7 CRC method and crypto field, exact atomic counter types, and shaded transport interfaces. Bytecode inspection of both released jars also confirms the declared retry-executor and bootstrap-list types accept the wrappers used by Comet.

I compiled the exact new suite without modification and ran it in separate JVMs against each released jar: **4/4 tests passed for each version**. The standalone harness used Scala 2.12.18, JDK 21, ScalaTest 3.2.19, and the Spark 3.5 logging API. The repository suite-inventory check and structural POM/workflow checks passed.

The exact-head CI jobs also passed all four tests for [Celeborn 0.6.0](https://github.com/apache/datafusion-comet/actions/runs/33453297017/job/99688326307) and [0.7.0](https://github.com/apache/datafusion-comet/actions/runs/33453297017/job/99688326193) on JDK 17. Other native and Spark jobs were still running when checked. The separate [PyArrow Spark 4.1 failure](https://github.com/apache/datafusion-comet/actions/runs/33453296925/job/99687769755) is an HTTP 429 while downloading Maven 3.9.6, before its build or tests begin. I did not rerun the full Comet native/JVM matrix locally.

## Key design decisions

The profile has no automatic activation, the dependency has test scope, and the source directory is registered only by that profile. Ordinary builds therefore retain their existing dependency and test-source paths. Loading classes without initialization keeps this check independent of a live Celeborn service; it checks reflection compatibility, not end-to-end shuffle semantics.

## Implementation sketch

1. Activate the compatibility profile with a selected Celeborn version.
2. Compile the isolated suite against the released shaded client.
3. Resolve and assert the required fields, methods, and version-specific contracts.
4. Run both versions in the dedicated CI matrix while keeping the suite out of the generic matrix.

Only opt-in test execution and CI coverage change. No production dependency, shuffle implementation, or native code is added or modified.

## Suggested improvements

No actionable P1 or significant P2 findings survived validation. No inline comments are needed.

sunchao (approved):
Reviewed `cbaa94dd31ae0fa9bf39241418133597d219e6b4` against base/merge-base `6dedcfeee8777b578570e6304e514cc359df1de3`. No introduced or materially worsened P1/P2 found.

Fresh validation on Spark 3.5.9 / Scala 2.12.18 / JDK 17: four compatibility tests passed against each released Celeborn 0.6.0 and 0.7.0 client, plus 73 profile-free pusher/manager tests. The suite also passed without project fixtures and failed as expected when the client dependency was removed. Exact-base/head Maven model checks and packaged binary/source JAR inspection confirmed dependency and test-source isolation.

Both dedicated compatibility CI jobs passed. The separate PyArrow Spark 4.1 job failed with HTTP 429 while downloading Maven 3.9.6. Other CI jobs are still running. This verifies reflection contracts and build isolation, not native shuffle transport or deployed-cluster behavior.

sunchao:
Merged, thanks @pingzh !

PR #5606, opened 2026-09-01
fix: route translate through codegen dispatcher by default

### LINKED ISSUE #5585: `translate` falls back to Spark by default instead of using the codegen dispatcher like the other string functions
Part of #5572.

`CometStringTranslate` reports `Incompatible` unconditionally:

```scala
override def getSupportLevel(expr: StringTranslate): SupportLevel = Incompatible(
  Some(incompatReason))
```

`spark/src/main/scala/org/apache/comet/serde/strings.scala:119-120`

The reason is real — DataFusion's `translate` iterates Unicode graphemes where Spark uses code points, and substitutes U+0000 rather than treating it as a deletion sentinel. But with no `CodegenDispatchFallback` mixin, the consequence is that `translate` falls the whole projection back to Spark unless the user sets `spark.comet.expression.StringTranslate.allowIncompatible=true` and accepts wrong answers.

Every comparable string function in the same file already has the better shape: compatible by default through the dispatcher, native as an explicit opt-in. See `CometInitCap` (`:146`), `CometStringReplace` (`:177`), `CometCaseConversionBase` for `upper` / `lower` (`:50`), and the regex family at `:368`, `:470`, `:508`, `:545`, `:598`.

Fix: mix `CodegenDispatchFallback` into `CometStringTranslate`. `StringTranslate` is a plain codegen expression over `StringType`, so `canHandle` admits it.

Once that lands, `docs/source/user-guide/latest/expressions.md` should report `translate` as Hybrid, and the existing note about the grapheme/code-point difference becomes an opt-in caveat rather than a description of default behavior.

adibmbrk:

Closes #5585.

## Rationale for this change

`CometStringTranslate` reported itself as `Incompatible` unconditionally, forcing whole projections to fall back to Spark unless `spark.comet.expression.StringTranslate.allowIncompatible=true` was set. Comparable string functions (`CometInitCap`, `CometStringReplace`) instead run natively by default through the JVM codegen dispatcher and treat their semantic differences as an opt-in caveat.

- `CometStringTranslate` mixes in `NativeOptInAvailable`: default is `Compatible` and routes through the codegen dispatcher (Spark-compatible); the incompatible native DataFusion path is opt-in via `allowIncompatible`.
- Docs: `translate` updated from `Native` to `Hybrid`.

Ran the affected suites against Spark 4.1 (`-Dtest=none -Dsuites=...`):

- `CometFallbackInvarianceSuite` — `pass=25 fail=0 excused=0 vacuous=0` (translate is now a real native-vs-fallback parity pass rather than the previous SKIPPED-VACUOUS case).
- `CometStringExpressionSuite` — 39 tests succeeded, 0 failed.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/serde/strings.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/strings.scala
@@ -109,15 +109,37 @@ object CometOctetLength extends CometScalarFunction[OctetLength]("octet_length")
   }
 }
 
-object CometStringTranslate extends CometScalarFunction[StringTranslate]("translate") {
+object CometStringTranslate
+    extends CometScalarFunction[StringTranslate]("translate")
+    with NativeOptInAvailable {
   private val incompatReason =
     "DataFusion's translate iterates over Unicode graphemes (Spark uses code points) and" +
       " substitutes U+0000 instead of treating it as a deletion sentinel"
 
   override def getIncompatibleReasons(): Seq[String] = Seq(incompatReason)
 
-  override def getSupportLevel(expr: StringTranslate): SupportLevel = Incompatible(
-    Some(incompatReason))
+  override def getSupportLevel(expr: StringTranslate): SupportLevel =
+    if (!CometConf.isExprAllowIncompat(getExprConfigName(expr))) {
+      Compatible(nativeOptIn =
+        Some(NativeOptIn(CometConf.getExprAllowIncompatConfigKey(getExprConfigName(expr)))))
+    } else {
+      Compatible()
+    }
+
+  override def convert(
+      expr: StringTranslate,
+      inputs: Seq[Attribute],
+      binding: Boolean): Option[Expr] = {
+    if (CometConf.isExprAllowIncompat(getExprConfigName(expr))) {
+      // Native path: faster but iterates Unicode graphemes and substitutes U+0000 rather than
+      // treating it as a deletion sentinel, so it is only used when incompatibility is allowed.
+      super.convert(expr, inputs, binding)
+    } else {
+      // Default: run Spark's own generated code inside the Comet pipeline for exact
+      // compatibility. Falls back to Spark when the codegen dispatcher is disabled.
+      CometScalaUDF.emitJvmCodegenDispatch(expr, inputs, binding)
+    }
+  }
 }
 
 object CometLevenshtein extends CometExpressionSerde[Levenshtein] {
--- a/docs/source/user-guide/latest/expressions.md
+++ b/docs/source/user-guide/latest/expressions.md
@@ -610,7 +610,7 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `to_char` | ✅ | Codegen dispatch |  |
 | `to_number` | ✅ | Codegen dispatch |  |
 | `to_varchar` | ✅ | Codegen dispatch |  |
-| `translate` | ✅ | Native | DataFusion's `translate` iterates over Unicode graphemes (Spark uses code points) and substitutes U+0000 instead of treating it as a deletion sentinel, so the native path is opt-in via allowIncompatible |
+| `translate` | ✅ | Hybrid | Routes through the JVM codegen dispatcher by default; the native path (which iterates Unicode graphemes rather than code points and substitutes U+0000 instead of a deletion sentinel) is opt-in via allowIncompatible |
 | `trim` | ✅ | Native |  |
 | `try_to_binary` | ✅ | — | Runs natively (rewrites to `try_eval(to_binary(...))`) |
 | `try_to_number` | ✅ | Codegen dispatch | Routed through the JVM codegen dispatcher |


sunchao (commented):
### PR summary

The default JVM dispatcher path keeps `translate` inside the Comet projection while preserving Spark semantics. Keeping the native implementation behind the existing opt-in also makes the compatibility boundary clear.

### CI and validation

Reviewed `dd198bb3` against base `6dedcfee`. This was a source review, without running builds or tests. The four current-head workflows show `action_required`, so they do not establish passing CI. The Spark 4.1 suite results in the description are author-reported. One P2 finding is attached inline about the existing default SQL fixture.

### Performance validation

Could you add a focused microbenchmark comparing the BASE default Spark fallback with this HEAD's default dispatcher, with the native opt-in disabled? A literal translation map and a representative changing-map column would cover the main cache behaviors. Please consume and compare the results, verify the actual execution paths, and keep the build, Spark/JDK, hardware, data and batch settings matched. Separating first-use compilation from warmed throughput and allocation/GC would show whether avoiding fallback transitions offsets the new dispatch costs. I have not measured a slowdown.

spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -109,15 +109,37 @@ object CometOctetLength extends CometScalarFunction[OctetLength]("octet_length")
...
     "DataFusion's translate iterates over Unicode graphemes (Spark uses code points) and" +
       " substitutes U+0000 instead of treating it as a deletion sentinel"
 
   override def getIncompatibleReasons(): Seq[String] = Seq(incompatReason)
 
-  override def getSupportLevel(expr: StringTranslate): SupportLevel = Incompatible(
-    Some(incompatReason))
+  override def getSupportLevel(expr: StringTranslate): SupportLevel =
+    if (!CometConf.isExprAllowIncompat(getExprConfigName(expr))) {
+      Compatible(nativeOptIn =
```
sunchao, line 123: [P2] Update the default translate SQL fixture

Could you update the default cases in `string_translate.sql` along with this change? It still has four `expect_fallback(is not fully compatible with Spark)` blocks. The SQL-file suite registers these blocks and requires that fallback contract, but this `Compatible` branch now accepts the first ordinary three-string-column query and dispatches it by default. The successful new route therefore cannot satisfy the old fixture expectation, even when its result values are correct. Please check Comet execution and result parity for the default cases, and keep any dispatcher-disabled fallback case separate with its actual reason. This is source-derived, not an observed test failure.

andygrove (commented):
`spark/src/test/resources/sql-tests/expressions/string/string_translate.sql` still wraps all four default-config queries in `expect_fallback(is not fully compatible with Spark)`. `CometStringTranslate.convert` now routes to `CometScalaUDF.emitJvmCodegenDispatch` when `allowIncompatible` is false. That path dispatches successfully and keeps the query in the Comet pipeline. It no longer falls back. `checkSparkAnswerAndFallbackReasons` in `CometTestBase` fails when no fallback reason gets recorded, so these four blocks read like they will fail outright rather than pass vacuously. `string_replace.sql` went through this exact transition for `CometStringReplace` and its default-config queries are plain `query` blocks now, with a comment noting they route through the codegen dispatcher. Could `string_translate.sql` follow that same shape?

`docs/source/contributor-guide/expression-audits/string_funcs.md` under `## translate` still says the expression falls back to Spark by default and runs natively only when incompatible expressions are allowed. That's the previous behavior, not this PR's. The `initcap` and `replace` entries earlier in the same file were updated for the identical transition to describe the `NativeOptIn` codegen-dispatch default. The `translate` entry should get the same update.

sunchao (commented):
## Correctness

Rechecked unchanged head `dd198bb30e57335179659c324343d7c7b37f991f` against `6dedcfeee8777b578570e6304e514cc359df1de3`, including [the new review](https://github.com/apache/datafusion-comet/pull/5606#pullrequestreview-5154033968). The [existing P2](https://github.com/apache/datafusion-comet/pull/5606#discussion_r3901095367) remains applicable and unresolved. I have no additional independent P1/P2 finding to post inline.

The fixture concern is supported by the current harness. `string_translate.sql` still has four `expect_fallback(is not fully compatible with Spark)` blocks and no dispatcher/configuration override or version restriction. The first query has three ordinary string-column inputs that the default dispatcher accepts. `CometStringTranslate` now returns `Compatible`, so that query no longer reaches the serializer branch that produces the requested incompatibility reason. `checkSparkAnswerAndFallbackReasons` checks the result and then requires the reason: successful dispatch cannot satisfy this assertion merely by returning correct values. The native-opt-in message is informational and is stored separately from fallback reasons. Disabling the dispatcher would produce a different reason. The first query is sufficient to establish the mismatch. Literal-only optimizer details do not invalidate it.

The existing request to convert the default cases to Comet-execution/result checks is therefore still appropriate. `string_replace.sql` provides the cited precedent with plain `query` blocks. Any dispatcher-disabled fallback control should have its own configuration and actual reason. This remains a source-derived assertion failure, not a locally observed test failure.

At the September 9, 12:34 UTC final check, there were zero current-head checks and six completed workflows with `action_required`. They establish no passing CI. The Spark 4.1 fallback-invariance and string-expression suite results in the description are author-reported and do not include `CometSqlFileTestSuite`. I ran no product build or runtime test. Maintained Spark 3.5/4.0 source was checked. Maintained 3.4/4.1 branches remain unavailable, with no substitute-source compatibility claim.

## Performance

The [earlier benchmark request](https://github.com/apache/datafusion-comet/pull/5606#pullrequestreview-5074365723) remains unanswered. Spark caches the translation dictionary, making a stable literal map and changing column maps useful separate cases. The dispatcher also serializes the bound expression, initializes a task-local generated kernel on first use, caches that kernel, and allocates output vectors for batches. These source facts explain what to measure. They do not establish a slowdown or a speedup.

The existing string-expression benchmark contains a literal-map `translate` query, but its presence is not a matched BASE/HEAD result. Please retain the requested comparison of BASE's default fallback with HEAD's default dispatcher, with the incompatible native path disabled. Consume and compare results, verify the actual paths, match the build/runtime/data/batch settings, and separate first-use compilation from warmed throughput and allocation/GC. No measured regression is claimed here.

## Design

The dispatcher boundary remains consistent with the maintained Spark implementations inspected: it carries the original bound Spark expression into Spark-generated code, including code-point/deletion behavior, null handling and the Spark 4.0 collation-aware implementation. This supports the proposed compatible default while keeping DataFusion's different grapheme/NUL behavior behind the explicit native opt-in. It is source support for the approach, not full runtime qualification across every Spark version.

`CometStringReplace` already uses the same support-level/convert pattern. The fixture update can follow that implementation's existing default test shape. An unsupported or disabled dispatcher still needs to report its real fallback cause, which is why retaining the previous incompatibility assertion would test the wrong contract.

## Abstraction & complexity

The change reuses `NativeOptInAvailable` and the existing dispatcher. I see no additional abstraction needed for this follow-up. The marker supplies documentation about the optional native path, while the actual support level and conversion select execution.

The contributor-audit request in the new review is also valid: the `translate` entry in `docs/source/contributor-guide/expression-audits/string_funcs.md` still describes the previous default Spark fallback. Its `initcap` and `replace` entries already describe the compatible codegen default. The user-guide table's new `Hybrid` label matches this PR, so updating the stale contributor entry would make the two descriptions consistent. This is an existing external review request. I am not adding a duplicate inline comment.

andygrove (commented):
This landed on `main` as #5032, which routes both `translate` and `to_csv` through the codegen dispatcher by default. `CometStringTranslate` there now mixes in `CodegenDispatchFallback`, `expressions.md` already reports `translate` as `Hybrid`, and `string_translate.sql` is plain `query` blocks with added regression rows for the combining-accent and U+0000 inputs where the native path disagrees with Spark. The opt-in native path stays covered by `string_translate_enabled.sql`. That leaves this branch conflicting in both files it touches.

The hand-rolled `getSupportLevel` and `convert` branching here is also the shape we moved away from during review of #5032, in favour of the mixin, since `QueryPlanSerde` already routes a non-opted-in `Incompatible` result to the dispatcher. Would it make sense to close this as superseded? Sorry for the duplicated effort.

One piece of your change did not get covered upstream. The `## translate` entry in `docs/source/contributor-guide/expression-audits/string_funcs.md` still says it falls back to Spark by default and runs natively only when incompatible expressions are explicitly allowed, which no longer matches the code. The `initcap` and `replace` entries in the same file describe the `NativeOptIn` codegen-dispatch default. Would you be up for a small PR against current `main` to bring that entry in line? #5585 is also still open even though the work has landed, so it could be closed at the same time.

sunchao (commented):
Confirmed the [superseding change in #5032](https://github.com/apache/datafusion-comet/pull/5032): merge commit `5e302d99` is contained in current main `9b63e7da`. Both dispatcher declarations and the default/native-opt-in translate fixtures still match that merged implementation. Main uses `CodegenDispatchFallback`, reports `translate` as Hybrid, and includes the combining-mark/U+0000 regression rows. I agree with [the latest review](https://github.com/apache/datafusion-comet/pull/5606#pullrequestreview-5191075831) that this PR can be closed as superseded.

On this PR's unchanged `dd198bb3`, the [existing P2](https://github.com/apache/datafusion-comet/pull/5606#discussion_r3901095367) still applies: its four fallback assertions were not updated. Main's fix does not change this head. No new P1/P2 finding.

The [contributor audit entry](https://github.com/apache/datafusion-comet/blob/9b63e7dacf70ac5428a7aa3ddb6f18f20f685778/docs/source/contributor-guide/expression-audits/string_funcs.md#L230) still describes default Spark fallback, so the requested small documentation PR against main remains appropriate. [#5585](https://github.com/apache/datafusion-comet/issues/5585) is also still open.

Source/provenance checks only. This head has zero check runs and six `action_required` workflows with zero jobs. No new runtime test or benchmark evidence.

adibmbrk:
Closing as superseded by #5032, which landed the same change on `main`

PR #5607, opened 2026-09-01
fix: route length/bit_length/octet_length binary input through codegen dispatcher

### LINKED ISSUE #5584: `length` / `bit_length` / `octet_length` fall back to Spark on binary input
Part of #5572.

`length`, `bit_length` and `octet_length` all decline `BinaryType` input, and each falls the whole projection back to Spark:

```scala
case _: BinaryType => Unsupported(Some("Length on BinaryType is not supported"))
```

`spark/src/main/scala/org/apache/comet/serde/strings.scala:89` (`CometLength`), `:98` (`CometBitLength`), `:107` (`CometOctetLength`)

All three are trivial in Spark's generated code — `length(binary)` is `.numBytes()`, and the other two are arithmetic on it. `BinaryType` is in `CometBatchKernelCodegen.isSupportedDataType`, so the dispatcher handles them.

Fix: mix `CodegenDispatchFallback` into all three. They are near-identical serdes, so this is one small change.

The generated docs list all three as "✅ Native" with no note today, so they should pick up the Hybrid marker once the mixin is added.

adibmbrk:

Closes #5584.

## Rationale for this change

`length`, `bit_length`, and `octet_length` rejected `BinaryType` input and fell the entire projection back to Spark, even though the operation is trivial (`numBytes()`) and `BinaryType` is already supported by the codegen dispatcher.

- Mix `CodegenDispatchFallback` into `CometLength`, `CometBitLength`, and `CometOctetLength` so binary input routes through the JVM codegen dispatcher (Spark's own `doGenCode`, run inside the Comet pipeline) instead of falling back to Spark.
- Regenerate the affected docs rows (`length`, `len`, `char_length`, `character_length`, `bit_length`, `octet_length`) from Native to Hybrid.
- Add `CometBinaryLengthBenchmark` (see below).

Updated the `length.sql`, `bit_length.sql`, and `octet_length.sql` fixtures: the binary cases now assert native-vs-Spark parity (`checkSparkAnswerAndOperator`) instead of `expect_fallback`. Added binary coverage to `length.sql`.

Ran against Spark 4.1:
- `CometStringExpressionSuite` — 33 succeeded, 0 failed.
- `CometSqlFileTestSuite` — 467 succeeded, 0 failed.

## Benchmark

Added `CometBinaryLengthBenchmark`, per review request. `CometStringExpressionBenchmark` covers these three roots only on `StringType`, which takes the native DataFusion kernel and so never exercises the binary route.

Three arms, all cases of one `Benchmark` so warmup, iteration count, data and SQL settings are matched by construction:

| Arm | Configuration |
| --- | --- |
| `Spark` | `spark.comet.enabled=false` |
| `Comet (Spark fallback)` | Comet on, `spark.comet.exec.scalaUDF.codegen.enabled=false` — the pre-PR path |
| `Comet (codegen dispatch)` | Comet on, dispatcher enabled — the path this PR adds |

Shapes span payload width (8 B / 64 B / 1 KB) and null fraction (0 / 50 / 90%). Each case prints its physical plan and a result digest before the timings; all three arms agreed in all 15 cases.

`length`, best time over 2M rows, Apple M4 / OpenJDK 21.0.8 / Spark 4.1 (`bit_length` and `octet_length` track it closely):

| Shape | Spark | Comet (Spark fallback) | Comet (codegen dispatch) |
| --- | --- | --- | --- |
| 8 B | 169 ms | 101 ms | 135 ms |
| 64 B | 201 ms | 164 ms | 206 ms |
| 1 KB | 754 ms | 1193 ms | 1522 ms |
| 64 B, 50% null | 167 ms | 104 ms | 150 ms |
| 64 B, 90% null | 92 ms | 68 ms | 80 ms |

## Known follow-up

Review surfaced a pre-existing correctness hole in `CometBatchKernelCodegen.canShortCircuitNulls`: its single-ordinal null short-circuit can swallow an ANSI error raised by a foldable-but-throwing subtree between the dispatched root and its single input ordinal (e.g. `length(substring(X'00', CAST(1L DIV 0L AS INT), n))`). This is not introduced by this PR — `upper` on `main` reproduces the same issue — and does not block this PR. Tracked and fixed separately in #5608.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/serde/strings.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/strings.scala
@@ -82,7 +82,10 @@ object CometUpper extends CometCaseConversionBase[Upper]("upper")
 
 object CometLower extends CometCaseConversionBase[Lower]("lower")
 
-object CometLength extends CometScalarFunction[Length]("length") {
+object CometLength extends CometScalarFunction[Length]("length") with CodegenDispatchFallback {
+  // The native `length` UDF has no path for BinaryType. Rather than fall the projection back to
+  // Spark, route the binary case through the JVM codegen dispatcher (Spark's own `doGenCode`, i.e.
+  // `numBytes()`) inside the Comet pipeline so the result stays native and matches Spark exactly.
   override def getUnsupportedReasons(): Seq[String] = Seq("`BinaryType` input is not supported")
 
   override def getSupportLevel(expr: Length): SupportLevel = expr.child.dataType match {
@@ -91,7 +94,11 @@ object CometLength extends CometScalarFunction[Length]("length") {
   }
 }
 
-object CometBitLength extends CometScalarFunction[BitLength]("bit_length") {
+object CometBitLength
+    extends CometScalarFunction[BitLength]("bit_length")
+    with CodegenDispatchFallback {
+  // See CometLength: BinaryType has no native path, so route it through the codegen dispatcher
+  // (Spark's own `doGenCode`, i.e. `numBytes() * 8`) instead of falling back to Spark.
   override def getUnsupportedReasons(): Seq[String] = Seq("`BinaryType` input is not supported")
 
   override def getSupportLevel(expr: BitLength): SupportLevel = expr.child.dataType match {
@@ -100,7 +107,11 @@ object CometBitLength extends CometScalarFunction[BitLength]("bit_length") {
   }
 }
 
-object CometOctetLength extends CometScalarFunction[OctetLength]("octet_length") {
+object CometOctetLength
+    extends CometScalarFunction[OctetLength]("octet_length")
+    with CodegenDispatchFallback {
+  // See CometLength: BinaryType has no native path, so route it through the codegen dispatcher
+  // (Spark's own `doGenCode`, i.e. `numBytes()`) instead of falling back to Spark.
   override def getUnsupportedReasons(): Seq[String] = Seq("`BinaryType` input is not supported")
 
   override def getSupportLevel(expr: OctetLength): SupportLevel = expr.child.dataType match {
--- a/spark/src/main/scala/org/apache/comet/serde/strings.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/strings.scala
@@ -82,10 +82,9 @@ object CometUpper extends CometCaseConversionBase[Upper]("upper")
 
 object CometLower extends CometCaseConversionBase[Lower]("lower")
 
-object CometLength extends CometScalarFunction[Length]("length") with CodegenDispatchFallback {
-  // The native `length` UDF has no path for BinaryType. Rather than fall the projection back to
-  // Spark, route the binary case through the JVM codegen dispatcher (Spark's own `doGenCode`, i.e.
-  // `numBytes()`) inside the Comet pipeline so the result stays native and matches Spark exactly.
+object CometLength extends CometScalarFunction[Length]("length") {
+  // Benchmarked: the codegen dispatcher is slower than Spark fallback on BinaryType (see
+  // CometBinaryLengthBenchmark), so BinaryType stays on plain Spark fallback for now.
   override def getUnsupportedReasons(): Seq[String] = Seq("`BinaryType` input is not supported")
 
   override def getSupportLevel(expr: Length): SupportLevel = expr.child.dataType match {
@@ -94,11 +93,8 @@ object CometLength extends CometScalarFunction[Length]("length") with CodegenDis
   }
 }
 
-object CometBitLength
-    extends CometScalarFunction[BitLength]("bit_length")
-    with CodegenDispatchFallback {
-  // See CometLength: BinaryType has no native path, so route it through the codegen dispatcher
-  // (Spark's own `doGenCode`, i.e. `numBytes() * 8`) instead of falling back to Spark.
+object CometBitLength extends CometScalarFunction[BitLength]("bit_length") {
+  // See CometLength: benchmarked slower via the dispatcher, so BinaryType stays on Spark fallback.
   override def getUnsupportedReasons(): Seq[String] = Seq("`BinaryType` input is not supported")
 
   override def getSupportLevel(expr: BitLength): SupportLevel = expr.child.dataType match {
@@ -107,11 +103,8 @@ object CometBitLength
   }
 }
 
-object CometOctetLength
-    extends CometScalarFunction[OctetLength]("octet_length")
-    with CodegenDispatchFallback {
-  // See CometLength: BinaryType has no native path, so route it through the codegen dispatcher
-  // (Spark's own `doGenCode`, i.e. `numBytes()`) instead of falling back to Spark.
+object CometOctetLength extends CometScalarFunction[OctetLength]("octet_length") {
+  // See CometLength: benchmarked slower via the dispatcher, so BinaryType stays on Spark fallback.
   override def getUnsupportedReasons(): Seq[String] = Seq("`BinaryType` input is not supported")
 
   override def getSupportLevel(expr: OctetLength): SupportLevel = expr.child.dataType match {
--- a/spark/src/test/scala/org/apache/spark/sql/benchmark/CometBinaryLengthBenchmark.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/benchmark/CometBinaryLengthBenchmark.scala
@@ -1,0 +1,248 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.spark.sql.benchmark
+
+import org.apache.spark.benchmark.Benchmark
+import org.apache.spark.sql.catalyst.optimizer.ConstantFolding
+import org.apache.spark.sql.internal.SQLConf
+
+import org.apache.comet.CometConf
+
+/**
+ * One binary-input shape under benchmark. Each shape is materialized as its own column in the
+ * source table so a single write covers the whole grid.
+ *
+ * @param name
+ *   short label for the shape
+ * @param column
+ *   name of the source column holding the shape
+ * @param width
+ *   payload width in bytes for the non-null rows
+ * @param nullPercent
+ *   percentage of rows that are NULL
+ */
+case class BinaryLengthShape(name: String, column: String, width: Int, nullPercent: Int) {
+
+  /** SQL that materializes this shape as a `binary` column. */
+  def sql: String = {
+    // RPAD to an exact width over the row id keeps values distinct (so Parquet does not collapse
+    // the column to a single dictionary entry) while holding the payload width fixed. The digits
+    // and the pad character are ASCII, so the UTF-8 encoding is exactly `width` bytes.
+    val payload = s"CAST(RPAD(CAST(value AS STRING), $width, 'x') AS BINARY)"
+    val expr = if (nullPercent == 0) {
+      payload
+    } else {
+      s"CASE WHEN PMOD(value, 100) < $nullPercent THEN CAST(NULL AS BINARY) ELSE $payload END"
+    }
+    s"$expr AS $column"
+  }
+}
+
+// spotless:off
+/**
+ * Benchmark to measure performance of Comet `length` / `bit_length` / `octet_length` on
+ * `BinaryType` input, across the two paths the serdes can take. Added in response to the review
+ * request on the PR that mixed `CodegenDispatchFallback` into the three serdes.
+ * `CometStringExpressionBenchmark` covers these three roots only on `StringType`, which takes the
+ * native DataFusion kernel and so never exercises the binary route.
+ *
+ * Three arms, all cases of the same `Benchmark` so they share warmup, iteration count, data and
+ * SQL settings by construction:
+ *
+ *   - `Spark` -- Comet off entirely, the reference.
+ *   - `Comet (Spark fallback)` -- Comet on with the codegen dispatcher disabled. Binary input is
+ *     `Unsupported` and, with the dispatcher off, `CodegenDispatchFallback` produces no marker,
+ *     so the enclosing projection falls back to Spark. This is the path this PR replaces, and it
+ *     is what the pre-PR build does unconditionally.
+ *   - `Comet (codegen dispatch)` -- Comet on with the dispatcher enabled, i.e. the path this PR
+ *     adds: Spark's own `doGenCode` (`numBytes()`) compiled into a per-batch kernel that reads
+ *     the Arrow vector directly, with no transition out of the Comet pipeline.
+ *
+ * The shapes span payload width (8 / 64 / 1024 bytes) at a fixed null fraction, and null fraction
+ * (0% / 50% / 90%) at a fixed width. Both axes matter, and the measured result is that the
+ * dispatcher path is the slower of the two Comet arms:
+ *
+ *   - Payload width drives the size of the gap. The kernel's `getBinary` allocates a `byte[]` and
+ *     copies the whole payload out of the Arrow buffer per row (see `emitBinaryBodyUnsafe` in
+ *     `CometBatchKernelCodegenInput`), because that is what Spark's generated `numBytes()` reads.
+ *     So a `length` that should be an offset subtraction pays for the full value, and the cost the
+ *     dispatch arm adds over the Spark-fallback arm tracks the payload: tens of ms at 8-64 B,
+ *     several hundred at 1 KB, where the dispatch arm also runs about twice as slow as Spark.
+ *   - Null fraction pulls the other way. The kernel short-circuits null rows ahead of the
+ *     generated code, so a skipped row costs no copy and the gap shrinks as nulls rise -- at the
+ *     same 64 B width it is several times smaller at 90% nulls than at 0%. It narrows sharply but
+ *     does not close: the dispatch arm stays the slower of the two Comet arms at every shape here.
+ *
+ * Keeping the projection inside the Comet pipeline is therefore not automatically a win for these
+ * roots; on wide binary it is a regression against the Spark fallback it replaces.
+ *
+ * Each case prints its physical plan and a digest of its result set before the timings, so the
+ * report shows which operators each arm actually ran and that all three arms agree on the output.
+ *
+ * To run this benchmark:
+ * {{{
+ *   SPARK_GENERATE_BENCHMARK_FILES=1 make benchmark-org.apache.spark.sql.benchmark.CometBinaryLengthBenchmark
+ * }}}
+ * Results will be written to "spark/benchmarks/CometBinaryLengthBenchmark-**results.txt".
+ */
+// spotless:on
+object CometBinaryLengthBenchmark extends CometBenchmarkBase {
+
+  private val shapes = List(
+    BinaryLengthShape("width_8B", "b_w8", width = 8, nullPercent = 0),
+    BinaryLengthShape("width_64B", "b_w64", width = 64, nullPercent = 0),
+    BinaryLengthShape("width_1KB", "b_w1024", width = 1024, nullPercent = 0),
+    BinaryLengthShape("width_64B_50pct_null", "b_w64_n50", width = 64, nullPercent = 50),
+    BinaryLengthShape("width_64B_90pct_null", "b_w64_n90", width = 64, nullPercent = 90))
+
+  private val roots = List("length", "bit_length", "octet_length")
+
+  /**
+   * Excluding `ConstantFolding` matches the rest of the expression benchmarks. Nothing in these
+   * queries is foldable (every root reads a column), but the exclusion is applied to all three
+   * arms so the optimizer configuration is identical across them.
+   */
+  private val noConstantFolding =
+    SQLConf.OPTIMIZER_EXCLUDED_RULES.key -> ConstantFolding.ruleName
+
+  private val sparkConfigs = Seq(noConstantFolding, CometConf.COMET_ENABLED.key -> "false")
+
+  private def cometConfigs(dispatch: Boolean): Seq[(String, String)] = Seq(
+    noConstantFolding,
+    CometConf.COMET_ENABLED.key -> "true",
+    CometConf.COMET_EXEC_ENABLED.key -> "true",
+    CometConf.COMET_SCALA_UDF_CODEGEN_ENABLED.key -> dispatch.toString)
+
+  override def runCometBenchmark(mainArgs: Array[String]): Unit = {
+    // Twice the 1M default the other expression benchmarks use. `length` on binary is O(1) per
+    // row, so what separates the two Comet arms is only the per-row cost of leaving the pipeline
+    // (`CometColumnarToRow` then a Spark `Project`) versus staying in it (`CometProject`). The row
+    // count has to be large enough that fixed query-submission cost does not drown that out: at
+    // 8192 rows every arm reports hundreds of ns/row, which is overhead rather than expression
+    // work.
+    val values = 2 * 1024 * 1024 // 2M rows
+
+    runBenchmarkWithTable("Binary length expressions", values) { v =>
+      withTempPath { dir =>
+        withTempTable("parquetV1Table") {
+          prepareTable(dir, spark.sql(s"SELECT ${shapes.map(_.sql).mkString(", ")} FROM $tbl"))
+
+          for (root <- roots; shape <- shapes) {
+            val name = s"$root ${shape.name}"
+            val query = s"select $root(${shape.column}) as v from parquetV1Table"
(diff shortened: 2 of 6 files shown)


spark/src/main/scala/org/apache/comet/serde/strings.scala
```diff
@@ -82,7 +82,10 @@ object CometUpper extends CometCaseConversionBase[Upper]("upper")
 
 object CometLower extends CometCaseConversionBase[Lower]("lower")
 
-object CometLength extends CometScalarFunction[Length]("length") {
+object CometLength extends CometScalarFunction[Length]("length") with CodegenDispatchFallback {
```
sunchao: [P2] Preserve child evaluation order for compound binary inputs

This marker dispatches the entire binary-producing child, so the existing kernel null shortcut can now suppress an earlier ANSI error. An unexecuted source-derived diagnostic is `IF(flag, length(substring(X'00', CAST(1L DIV 0L AS INT), n)), 0)` over persisted Parquet rows `(true, NULL)` and `(false, NULL)`, with `flag BOOLEAN`, nullable `n INT`, and ANSI/Comet projection/codegen dispatch enabled. In the inspected Spark 3.5/4.0 source, the conditional keeps the failing constant inside a branch. Spark evaluates Substring's position before its later length argument, so the selected branch must raise division by zero even when `n` is null.

Here `CometScalaUDF` captures the Length tree with only `n` bound. Its nodes pass `allNullIntolerant` and the single-input-ordinal guard in `CometBatchKernelCodegen`, which writes NULL before evaluating the generated child code. At BASE the unsupported binary Length has no dispatcher marker and the enclosing projection falls back to Spark. The new BitLength and OctetLength markers expose the same issue. Please preserve Spark's evaluation order, or retain fallback for these unsafe compound trees, and add a regression asserting the ANSI error for all three roots. This is a source trace, not an executed reproduction.

andygrove: Thanks @sunchao, I ran this down and the trace holds. Reproduced on this branch (Spark 4.1, ANSI on):

```sql
CREATE TABLE t (flag BOOLEAN, n INT) USING parquet;
INSERT INTO t VALUES (true, NULL), (false, NULL);
SELECT IF(flag, length(substring(X'00', CAST(1L DIV 0L AS INT), n)), 0) FROM t;
```

Spark raises `[DIVIDE_BY_ZERO]`, Comet returns a row. Same for `bit_length` and `octet_length`.

For anyone reading later, the three pieces that have to line up:

* `ConstantFolding` refuses to fold `1L DIV 0L` because it sits under an `If` branch (it tags `FAILED_TO_EVALUATE` and leaves the node alone), so the throwing literal survives into the physical plan.
* `TernaryExpression.nullSafeCodeGen` emits `Substring`'s `pos` code before it tests `len`'s null, so Spark evaluates the division even though `n` is NULL.
* `Length`, `Substring`, `Cast` and `IntegralDivide` are all null-intolerant and the dispatched tree reads exactly one ordinal, so `canShortCircuitNulls` takes its single-ordinal branch and the kernel writes NULL before `ev.code` runs.

One correction on scope: this isn't introduced here, it's the residual hole in #5219. The single-ordinal branch assumes "there is nothing left for Spark to evaluate ahead of that ordinal's own null check", and that's false whenever the tree carries a literal-only subtree that throws. `upper` reproduces it on main today, unchanged by this PR:

```sql
SELECT IF(flag, upper(substring('abc', CAST(1L DIV 0L AS INT), n)), NULL) FROM t;
```

I confirmed that one on the same build: Spark raises, Comet doesn't.

So I'd rather fix `canShortCircuitNulls` than special-case the three length serdes, otherwise we paper over three of the ~70 expressions that share the hole. Filed as #5608, with the suggested guard and a regression test covering `upper` plus all three roots from this PR.

@adibmbrk I don't think this needs to block the PR. Please add a link to #5608 in the PR description so the connection isn't lost.

sunchao (commented):
Re-reviewed `37f12bed` after the reproduction discussion. [P2] Thanks for reproducing the evaluation-order issue and tracking the shared fix in #5608. I agree the shared dispatcher is the right place to address it, with these three newly exposed binary callers covered alongside `Upper`. I found no additional P1/P2 issues.

Could you share one focused binary-input microbenchmark comparing this dispatcher path with the previous Spark-fallback path? Please use matched Spark/Comet build settings, data and warmup, with representative payload widths and null fractions, and include the actual execution plans and matching results. The existing string-expression benchmark does not exercise this binary route.

kavix (approved):

sunchao (commented):
Thanks for documenting the separate fix. I verified that [#5608](https://github.com/apache/datafusion-comet/issues/5608) was closed by merged [#5623](https://github.com/apache/datafusion-comet/pull/5623), preserving the agreed separate-fix scope for the existing [P2 evaluation-order issue](https://github.com/apache/datafusion-comet/pull/5607#discussion_r3903164108). That guard is absent from this exact head, and I have not validated the combined changes.

The updated description also says the binary-input benchmark was added. At `37f12bed`, [CometStringExpressionBenchmark](https://github.com/apache/datafusion-comet/blob/37f12bed6ec897b18d344d7d8c409663f642d0b0/spark/src/test/scala/org/apache/spark/sql/benchmark/CometStringExpressionBenchmark.scala#L89) is unchanged from the base and still builds string `c1` and integer `c2`. Could you push or link the benchmark commit and its results for the stated payload widths and null fractions, including dispatcher-on/off plans and matching outputs, so the earlier request can be closed?

No new P1/P2 findings. This pass was source and discussion verification, not a benchmark or Spark/JNI run. Current-head checks are 63 successful and 9 skipped. A separate title-check workflow remains `action_required`.

adibmbrk:
Hi @sunchao @andygrove, I have committed the microbenchmark and the results from it are in the PR description.

sunchao (commented):
Thanks for adding the binary benchmark. [P2] Could we retain Spark fallback for these three binary roots until a narrow implementation improves the comparison, or provide evidence that justifies the tradeoff? Your reported `length` times increase from 164 to 206 ms at 64 B (+25.6%) and from 1193 to 1522 ms at 1 KB (+27.6%). With Comet execution and the default dispatcher setting enabled, the new markers select that route for ordinary binary-column projections. I inspected the source but have not independently reproduced these timings.

A native binary-length implementation or a direct-vector length path that reads offsets seems worth measuring, while preserving null handling, indexing and compound-child evaluation. The generated dispatcher getter allocates and copies the payload, but `CometPlainVector.getBinary` in the fallback also does so. These numbers do not isolate copying as the cause of the extra time. A length-specific optimization should leave the general `getBinary` contents contract intact.

For the comparison, could you capture the actual timed write-command plans and matching results on your Spark 4.1 build? In the maintained Spark 3.5/4.0 sources, `noop()` creates a separate command `QueryExecution`. The SELECT plan printed by `describe` therefore does not establish the timed writer operators. I have not independently verified that boundary on your Spark 4.1 build.

The previously deferred #5608 guard is now in the reported base and inspected merge source, but not raw HEAD. Its accepted separate-fix scope is unchanged. This pass was source review, not a Spark/JNI or benchmark run. All four current-head workflows are awaiting approval (`action_required`), with no check results yet.

sunchao:
Reviewed head **`b517646`** against base `90c1dd9` with five independent scopes. **The change adds little code, but the default execution choice is not justified by the supplied performance results. I also found a new build failure.**

1. **[P1] The benchmark breaks Spark 3.4/3.5 test compilation.**  
   [`describe`](https://github.com/apache/datafusion-comet/blob/b517646ff95d1e9a2b086a1802fddbdd3b805d87/spark/src/test/scala/org/apache/spark/sql/benchmark/CometBinaryLengthBenchmark.scala#L206) returns `String`, but those Spark versions’ `withSQLConf` returns `Unit`. Current CI reports exactly `found: Unit, required: String`. I independently reproduced this with an extracted compilation. Capturing the result inside the configuration block and returning it afterward compiles with both API signatures. [CI failure](https://github.com/apache/datafusion-comet/actions/runs/33795817298/job/100916316855)

2. **[P2] The new default route is slower in every reported benchmark shape.**  
   The author’s `length` results show **164 → 206 ms at 64 B (+25.6%)** and **1193 → 1522 ms at 1 KB (+27.6%)** versus the previous Comet/Spark fallback. These are author-reported measurements; I did not independently reproduce the timings. They argue for retaining fallback until a faster implementation or representative pipeline measurements justify changing it. [Reported results](https://github.com/apache/datafusion-comet/pull/5607)

   The dispatcher adds a JVM callback and Arrow bridging per expression/batch. Its [binary getter](https://github.com/apache/datafusion-comet/blob/b517646ff95d1e9a2b086a1802fddbdd3b805d87/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegenInput.scala#L281) also allocates and copies the payload merely to read its length. **The old fallback copies binary values too**, so copying alone does not establish the cause of the measured regression.

3. **The benchmark should capture the actual timed execution plan.**  
   [`describe`](https://github.com/apache/datafusion-comet/blob/b517646ff95d1e9a2b086a1802fddbdd3b805d87/spark/src/test/scala/org/apache/spark/sql/benchmark/CometBinaryLengthBenchmark.scala#L207) invokes `df.noop()` and then prints the original SELECT’s plan. I verified that Spark 4.1.3’s `noop()` executes a separate write command with its own `QueryExecution`. Capture that command through a query listener and assert the intended operators. Also measure native downstream consumers and a projection computing all three lengths from the same input. Those cases would test the claimed benefit of preserving the Comet pipeline.

**For design and complexity, I favor a small native binary-length implementation.** Reusing `CodegenDispatchFallback` is consistent with existing code and introduces no unnecessary framework. The execution machinery is nevertheless expensive for an operation that only needs adjacent Arrow offsets.

Arrow 58.4 already provides binary length kernels. A narrow adapter could share byte-length handling between `length` and `octet_length`, preserve Spark’s wrapping multiplication for `bit_length`, and avoid payload materialization and JVM bridging. Three local component tests passed for nulls, empty/arbitrary bytes, sliced arrays, and wide values. This validates the underlying kernels, not a completed Comet implementation.

The performance and plan-reporting concerns were already raised in the discussion; the compilation failure is the new finding. At the last CI check: **4 failed, 13 passed, 6 running, 6 skipped**. No full Comet build or end-to-end benchmark was run locally.

adibmbrk:
Thanks for the detailed review, @sunchao — addressed both rounds:

1. Compile break (P1) — Fixed. `describe()` now captures its result inside the `withSQLConf` block into a `var` and returns it after, so it compiles regardless of whether `withSQLConf` returns `Unit` (3.4/3.5) or the block's value. Verified with a clean `test-compile` on both the default (Spark 4.1) and Spark 3.5 profiles.

2. Default route regression (P2) — Reverted. `CometLength`, `CometBitLength`, and `CometOctetLength` no longer mix in `CodegenDispatchFallback`; `BinaryType` input falls straight back to Spark again, same as before the original PR. Docs and SQL fixtures reverted to match (`Hybrid` → `Native`, `expect_fallback` assertions restored).

3. Timed-plan capture — Fixed. The benchmark now registers a `QueryExecutionListener` around `df.noop()` and reports the actual write command's `executedPlan`, instead of the pre-`noop` SELECT plan.

4. Additional benchmark coverage — Added a native downstream-consumer case (`sum(length(...))`) and a combined-projection case (all three roots in one query), each across the three width shapes.

Re-ran the full benchmark (21 scenarios × 3 arms, 2M rows) after the fix. Results: both Comet arms (`Spark fallback` / `codegen dispatch`) now report identical plans and near-identical timings at every shape, confirming the dispatcher no longer affects these three roots on binary input. Separately, the numbers also show Comet trailing plain Spark at 1 KB width regardless of arm (~0.6–0.7x) — that's a pre-existing `CometColumnarToRow` transition cost unrelated to the dispatcher, not something this PR introduces or claims to fix.

adibmbrk:
**Benchmark environment:** Apple M4, OpenJDK 21, 2M rows, `SPARK_GENERATE_BENCHMARK_FILES=1 make benchmark-org.apache.spark.sql.benchmark.CometBinaryLengthBenchmark`

Best Time (ms) / Relative-to-Spark:

| Scenario | Spark | Comet (Spark fallback) | Comet (codegen dispatch) |
|---|---|---|---|
| length width_8B | 147 (1.0x) | 87 (1.7x) | 86 (1.7x) |
| length width_64B | 190 (1.0x) | 153 (1.2x) | 153 (1.2x) |
| length width_1KB | 732 (1.0x) | 1065 (0.7x) | 1279 (0.6x) |
| length width_64B_50pct_null | 158 (1.0x) | 93 (1.7x) | 94 (1.7x) |
| length width_64B_90pct_null | 84 (1.0x) | 65 (1.3x) | 66 (1.3x) |
| bit_length width_8B | 139 (1.0x) | 95 (1.5x) | 86 (1.6x) |
| bit_length width_64B | 186 (1.0x) | 150 (1.2x) | 151 (1.2x) |
| bit_length width_1KB | 757 (1.0x) | 1170 (0.6x) | 1129 (0.7x) |
| bit_length width_64B_50pct_null | 161 (1.0x) | 92 (1.7x) | 93 (1.7x) |
| bit_length width_64B_90pct_null | 83 (1.0x) | 61 (1.4x) | 61 (1.4x) |
| octet_length width_8B | 145 (1.0x) | 102 (1.4x) | 104 (1.4x) |
| octet_length width_64B | 184 (1.0x) | 148 (1.2x) | 147 (1.2x) |
| octet_length width_1KB | 723 (1.0x) | 1094 (0.7x) | 1216 (0.6x) |
| octet_length width_64B_50pct_null | 153 (1.0x) | 90 (1.7x) | 91 (1.7x) |
| octet_length width_64B_90pct_null | 80 (1.0x) | 63 (1.3x) | 63 (1.3x) |
| length downstream-sum width_8B | 107 (1.0x) | 82 (1.3x) | 82 (1.3x) |
| length downstream-sum width_64B | 166 (1.0x) | 160 (1.0x) | 159 (1.0x) |
| length downstream-sum width_1KB | 711 (1.0x) | 1138 (0.6x) | 1067 (0.7x) |
| combined roots width_8B | 134 (1.0x) | 92 (1.5x) | 87 (1.5x) |
| combined roots width_64B | 182 (1.0x) | 147 (1.2x) | 148 (1.2x) |
| combined roots width_1KB | 662 (1.0x) | 1199 (0.6x) | 1009 (0.7x) |

All 21 scenarios: the three arms agree on output (no digest-mismatch warnings). Both Comet arms are now within noise of each other at every shape — confirming the dispatcher no longer changes the plan for `length`/`bit_length`/`octet_length` on `BinaryType`. Comet still trails plain Spark at 1 KB width regardless of arm (~0.6–0.7x); that's a general `CometColumnarToRow`→`Project` transition cost, not specific to the dispatcher this PR is about.

spark/src/test/scala/org/apache/spark/sql/benchmark/CometBinaryLengthBenchmark.scala
```diff
@@ -0,0 +1,301 @@
...
+      override def onSuccess(funcName: String, qe: QueryExecution, durationNs: Long): Unit =
+        captured = qe.executedPlan
+      override def onFailure(funcName: String, qe: QueryExecution, exception: Exception): Unit =
+        ()
+    }
+    spark.listenerManager.register(listener)
+    try {
+      df.noop()
+    } finally {
+      spark.listenerManager.unregister(listener)
```
sunchao, line 265: ### Correctness

[P2] Wait for the write callback before removing the listener

Could you wait for the matching write-command callback before unregistering this listener? In the maintained Spark 3.5/4.0 sources, `SQLExecution` posts `SparkListenerSQLExecutionEnd` to the shared `AsyncEventQueue`. Returning from `df.noop()` does not wait for `QueryExecutionListener.onSuccess`, and `unregister` only removes the listener. If the queue has not delivered the event yet, this `finally` removes it first and the subsequent `require(captured != null)` aborts the benchmark before timing starts. `@volatile` makes a completed assignment visible but does not wait for it. The callback also accepts every query from the session, so a queued earlier action can supply an unrelated plan. Please drain prior events, identify and await the intended write callback with a bounded wait, then unregister in `finally`. Spark's `DataFrameCallbackSuite` explicitly waits for listener delivery before checking results or removing its listener. This is a source-verified race, not a locally executed benchmark reproduction.

sunchao (commented):
## Correctness

The update addresses the earlier binary-dispatch concern: the three serdes again return `Unsupported` for binary input without `CodegenDispatchFallback`. Against the assigned base, their executable code and the shared dispatcher are unchanged. Maintained Spark 3.5/4.0 source and tests confirm binary byte counts for `length`/`octet_length`, 32-bit wrapping multiplication for `bit_length`, and null results for null inputs and zero for empty binary values. String character counting versus UTF-8 byte counting also remains unchanged. The added `length.sql` cases cover binary columns, literals, empty values, nulls and arbitrary bytes through result comparison plus fallback reasons, with constant folding disabled by the harness. The previously separated #5608 guard is now present in both head and base.

The prior `withSQLConf` return-type error is corrected in source: `describe` returns the captured `String` after the configuration block. One new P2 remains in `timedPlan`: its asynchronous query listener can be removed before receiving the write completion, causing the benchmark to fail or capture a queued unrelated query. See the inline comment.

### Validation

This follow-up reviewed the complete authored diff, the rebased contributions and the update from `b517646f`. All three current-head workflows are `action_required`, with zero jobs and no check results. There is no executed CI checkout or current test result to credit. The author's compile and benchmark reports remain author evidence. I ran no local build, Spark/JNI test or benchmark. Maintained Spark 3.4/4.1 sources remain unavailable, and the old remote worker packet was not reopened. Its prior publication was verified through ROOT's checksum-checked local records and current discussion.

## Performance

Removing the three markers addresses the earlier P2 about selecting the reported slower route by default. Both Comet benchmark arms now use the existing binary fallback for these roots. The new downstream aggregate and combined-root scenarios broaden the harness, but their numbers do not establish a dispatcher improvement. The reported results alone also do not isolate the cause of the 1 KB slowdown or establish that differences between runs are noise.

## Design

Retaining the established fallback keeps the production decision consistent with the available evidence. Capturing the write-command `QueryExecution` is the right boundary for this benchmark, but the listener must wait for and identify that command before its plan can support a comparison. The added scenarios need no new production mechanism.

## Abstraction & complexity

The compile fix uses a local result variable and introduces no shared abstraction. The listener helper is small, but its asynchronous lifecycle needs explicit completion and cleanup ordering. No additional production abstraction or general binary-getter change is needed to address this update.

andygrove:
Thanks for working through this, and for measuring rather than assuming.

The thing that stands out at `4feac844` is that the description no longer matches the code. It still says the three serdes mix in `CodegenDispatchFallback`, that six docs rows moved from Native to Hybrid, and that the binary fixtures assert parity instead of `expect_fallback`. After the revert none of that is true, and the fixtures assert the opposite. It also still says `Closes #5584`, so merging as-is would auto-close an issue that nobody has implemented. Could you rewrite the description and unlink the issue, and post the measurement on #5584 instead, so that the next person to pick it up finds the numbers?

I checked out the head and planned all five of the benchmark's query shapes with `spark.comet.exec.scalaUDF.codegen.enabled` on and off. They are identical in every shape apart from the exprId suffixes: `Project` over `CometColumnarToRow` over `CometNativeScan`, and a plain Spark `HashAggregate` for the aggregate case. That is what I would expect once the mixin is gone, since nothing else in those queries is dispatcher-eligible. So the third arm is measuring the same plan as the second in all 21 scenarios, and the digest check cannot notice, because equal plans give equal digests. `CometCodegenDispatchBenchmark` has a guard for exactly this at line 341 that prints "the two Comet cases below are measuring the same plan". Worth knowing too that `benchmarks/micro/run.py` discovers benchmarks from that directory automatically, so this would run on every micro-benchmark run rather than only when someone asks for it.

That leads to the bigger problem. The comments in `strings.scala` cite this benchmark as the evidence for staying on Spark fallback, but the benchmark cannot produce that evidence any more, because the path it compared against was deleted in the same PR. The numbers only exist at `0f579eafb`, which goes away on squash.

Your last run is also a useful accident. Since both Comet arms execute the same plan, the spread between them is this harness's noise floor. At 64 B that is under one percent. At 1 KB it runs from -15.8% to +20.1% (`length` 1065 vs 1279, combined roots 1199 vs 1009). Against that, the +25.6% at 64 B that motivated the revert looks real, but the +27.6% at 1 KB is inside the noise, so I would not lean on the payload-scaling claim.

Where does that leave things? The `length.sql` binary block is worth keeping on its own. It fills a gap that the other two fixtures already covered, and it passes. Beyond that I would drop `CometBinaryLengthBenchmark` rather than try to repair it. If you still want a binary length benchmark, three entries in `CometStringExpressionBenchmark`'s existing `StringExprConfig` table plus one binary corpus column gets the Spark versus Comet comparison in about four lines, and it inherits the "is the Comet plan fully native" warning from `runExprBenchmark`. Would that work for you?

sunchao (commented):
## Correctness

Follow-up on unchanged head `4feac8447fa325102bafffb5e40fec5d9255038b` against `bc74cc79fcf5eeed9610d9a20937375652cb74df`, after the [new discussion](https://github.com/apache/datafusion-comet/pull/5607#issuecomment-5592015623). The binary roots still return `Unsupported` without `CodegenDispatchFallback`. Their executable code and the inspected shared dispatcher match the base, so this revision introduces no binary dispatcher execution. Maintained Spark 3.5/4.0 still supply the binary semantics through fallback: byte counts for `length`/`octet_length`, 32-bit wrapping multiplication for `bit_length`, NULL for null input and zero for empty input. String character versus byte counting is unchanged.

The existing [P2 listener finding](https://github.com/apache/datafusion-comet/pull/5607#discussion_r3953950725) remains. I independently rechecked the maintained listener and writer paths: `noop()` returns without waiting for the asynchronous callback, `unregister` does not drain the queue, and this callback accepts unrelated same-session completions. `@volatile` does not provide the missing completion ordering. Keep the finding open if this benchmark remains. Removing the helper with the redundant benchmark would also remove this issue. No duplicate inline is needed.

The description also needs to match the reverted implementation. It still claims dispatcher enrollment, Hybrid documentation and binary native-execution fixtures, while the source retains fallback. Please update the title/description and remove `Closes #5584`, as already requested in the discussion. [#5584](https://github.com/apache/datafusion-comet/issues/5584) remains open and its requested binary route is not implemented here.

### Validation

I reviewed the full five-file contribution, confirmed the incremental source diff since the prior review is empty, and rechecked maintained Spark source/tests, type dispatch, the null-order guard and benchmark lifecycle. The prior compile fix remains intact in source. At **2026-09-09 05:01 UTC**, all three current-head workflows are `action_required`. Fresh job queries return zero jobs, with no head or merge checks. There is no executed CI checkout or current test result to credit. No local build, Spark/JNI execution or benchmark was run. Maintained Spark 3.4/4.1 sources remain unavailable.

## Performance

The source supports the external review's central comparison concern: toggling the dispatcher does not select a different binary route for these queries. The third arm adds 21 duplicate timed cases. The new object is also discovered by default unfiltered micro-benchmark runs, although suite filters can exclude it. Equal digests check outputs, not whether two distinct implementations were measured.

The reported timings remain external evidence. A single table of best times does not establish a noise floor, statistical significance or the cause of the 1 KB spread. The historical dispatch getter copies the payload, but the existing fallback getter does too. That observation cannot isolate the cost of JNI dispatch or explain the measured difference. The current benchmark cannot regenerate a dispatch-versus-fallback comparison after the markers were removed.

## Design

I support keeping the added binary `length.sql` coverage and dropping the redundant three-arm benchmark. If binary performance coverage is retained, use an honest Spark-versus-Comet-fallback comparison in the existing expression harness, with appropriate binary corpus cases. Its native-plan warning is useful context, although it is not proof of the timed write-command plan. Preserve historical measurements with their exact revision and environment rather than citing the current same-route benchmark as evidence for the removed route.

## Abstraction & complexity

The new 301-line benchmark adds a separate shape model, three-arm orchestration, digest reporting and listener lifecycle without retaining three distinct execution paths. Reusing the existing expression benchmark removes that duplicated machinery and the listener race. No new production adapter or general binary-getter change is needed for the current test-only behavior.

andygrove (commented):
This is still `4feac8447`, the commit I already went through in my last comment, so the points from there haven't been addressed by any new commit. The description still says the three serdes mix in `CodegenDispatchFallback`, that six doc rows moved from Native to Hybrid, and that the binary fixtures assert parity instead of falling back. None of that matches this diff. It also still carries `Closes #5584`, which would auto-close that issue on merge without the binary route it asks for actually landing in the code.

`CometBinaryLengthBenchmark`'s third arm still isn't evidence of anything now that the dispatcher mixin is gone. With no dispatch path left to compare against the fallback arm, the two Comet cases run the identical plan. That's exactly the situation `CometCodegenDispatchBenchmark.checkPlans` warns about when it detects the dispatcher never ran for a query, and this benchmark has no equivalent check. So the comments in `strings.scala` end up citing a comparison whose other side this same PR deletes.

@sunchao's open point about the listener race in `timedPlan` stands too. I don't have anything to add beyond agreeing it needs a fix, or the benchmark needs to go, before this merges.

The `length.sql` binary block is worth keeping regardless of how the benchmark question gets resolved. It closes a real gap against the coverage `bit_length.sql` and `octet_length.sql` already had.

andygrove:
Triage note: #5874 covers `length` on `BinaryType` too, and reaches the opposite conclusion. It registers `datafusion-spark`'s `SparkLengthFunc` under `length` / `char_length` / `character_length` / `len` and deletes the `BinaryType` gate from `CometLength` outright, so binary length runs natively rather than staying on Spark fallback because the dispatcher benchmarked slower.

The two collide fairly directly. Both edit `CometLength` in `strings.scala`, both rewrite the notes in `bit_length.sql` and `octet_length.sql`, both add binary cases to `length.sql` — this PR asserting `expect_fallback(Length on BinaryType is not supported)` and #5874 asserting it runs natively — and both add a file called `CometBinaryLengthBenchmark.scala`.

If the `datafusion-spark` kernel really is Spark-compatible for binary, native should beat both the dispatcher and the fallback, which would leave this PR as `bit_length` / `octet_length` plus the benchmark and the documentation of why the dispatcher is not worth it there. Could you and @dwsmith1983 settle the ordering before either gets more review?

PR #4891, opened 2026-07-10
chore: Codegen fallback explain for plan

comphead:
## Summary

Adds a new opt-in explain config, `spark.comet.explainCodegen.enabled`, that annotates each Comet operator with a single `[COMET-INFO: JVM codegen dispatcher: <name1>, <name2>, ...]` segment naming the expressions on that operator routed through the JVM codegen dispatcher (Spark's `doGenCode` running inside a Comet kernel). This helps distinguish "runs natively in DataFusion" from "runs Spark's generated code inside a Comet kernel" without changing execution behavior.
  
  ## Changes

  - **New config** `spark.comet.explainCodegen.enabled` (default `false`) in `CometConf`.
  - **New tree tag** `CODEGEN_DISPATCH_EXPRS` on `CometExplainInfo`, plus `CometSparkSessionExtensions.withCodegenDispatchExpr` helper.
  - **`CometExecRule.rollUpInfoMessages`** now aggregates dispatch names across an operator's expression trees into a single sorted, deduped info line.
  - **`CometScalaUDF`** tags dispatched expressions (when the new config is on) and now prefixes fallback reasons with the expression's display name — using `ScalaUDF.udfName` when set, falling back to
  `prettyName`, lowercased — so multiple fallback reasons on the same operator are distinguishable. Also converts `getOrElse(return None)` sites into explicit fallback-reason paths.
  - **Docs**: new section in `understanding-comet-plans.md` with an example and note that the operator remains Comet-accelerated (informational only).
  - **Tests**: extends `CometCodegenSuite` to cover the annotation.

PR #5610, opened 2026-09-01
test: add helpers to assert whether an expression ran natively or via codegen dispatch

### LINKED ISSUE #5609: Test helpers to assert whether an expression ran natively or through the codegen dispatcher
### Is your feature request related to a problem or challenge?

Comet evaluates a scalar expression one of three ways: natively (a DataFusion expression), through the JVM codegen dispatcher (Spark's own `doGenCode` compiled into an Arrow batch kernel), or not at all (the enclosing operator falls back to Spark).

Our tests can see the third and not the first two. `checkSparkAnswerAndOperator` and the default `query` mode in the SQL file harness both assert "no fallback", which is a real assertion, but native and codegen-dispatched execution are indistinguishable to them: both produce Spark-matching results by construction.

That leaves a class of change invisible to the test suite. A serde that widens from native to codegen dispatch silently gives up the native kernel; one that narrows from dispatch to native silently gives up Spark-exact semantics. Neither changes a single result, so every existing assertion stays green.

This bites in practice on the growing set of expressions whose mechanism depends on the argument type. `CometRound` (#5600) dispatches on float/double and stays native on decimal/integral. `CometLength` / `CometBitLength` / `CometOctetLength` (#5607) will dispatch on `BinaryType` and stay native on `StringType`. Nothing pins either split. `lower.sql` and `upper.sql` open with a comment saying the fixture exists to exercise the dispatcher route, and nothing checks that it does.

The pieces to assert this already exist. `ExtendedExplainInfo` exposes `getNativeExpressions` and `getCodegenDispatchExpressions`, and `CometCodegenSuite` uses them. There is just no reusable helper, so writing the assertion is enough friction that nobody does.

### Describe the solution you'd like

A helper on `CometTestBase`:

```scala
checkSparkAnswerAndImpl(df, native = Seq("round"), dispatched = Seq.empty)
```

and matching SQL file harness query modes, so the assertion is available at the layer where most expression coverage actually lives:

```sql
query expect_dispatch(bit_length)
SELECT bit_length(b) FROM test_bit_length_binary

query expect_native(bit_length)
SELECT bit_length(s) FROM test_bit_length
```

Naming an expression asserts both that it ran through the expected mechanism and that it did not run through the other one, so the assertion is a claim rather than a hint.

### Describe alternatives you've considered

Making the mechanism assertion mandatory: have the default `query` mode assert that the dispatched-expression set matches a file-level declaration defaulting to empty. That catches the whole regression class across every fixture permanently rather than only where someone remembered to annotate.

It needs a one-time annotation pass over every fixture that already dispatches (`rlike`, `regexp_replace`, `split`, `lower`, `upper`, `round` on float/double, the `mask` family), and support levels differ by Spark version, so some fixtures would need version-conditional declarations. Worth doing, but after the opt-in form is in use and the annotation cost is understood. Tracked as a follow-up rather than built here.

### Additional context

Came out of review discussion on #5607.

andygrove:

Closes #5609.

## Rationale for this change

Comet evaluates a scalar expression one of three ways: natively (a DataFusion expression), through the JVM codegen dispatcher (Spark's own `doGenCode` compiled into an Arrow batch kernel), or not at all (the enclosing operator falls back to Spark).

Our tests can see the third and not the first two. `checkSparkAnswerAndOperator` and the default `query` mode in the SQL file harness both assert "no fallback", which is a real assertion, but native and dispatched execution are indistinguishable to them: both produce Spark-matching results by construction.

So a serde that widens from native to codegen dispatch silently gives up the native kernel, and one that narrows from dispatch to native silently gives up Spark-exact semantics. Neither changes a result, so every existing assertion stays green.

This matters on the growing set of expressions whose mechanism depends on the argument type. `CometRound` (#5600) dispatches on float and double and stays native on decimal and integral. `CometLength` / `CometBitLength` / `CometOctetLength` (#5607) will dispatch on `BinaryType` and stay native on `StringType`. Nothing pinned either split. `lower.sql` and `upper.sql` open with a comment saying the fixture exists to exercise the dispatcher route, and nothing checked that it did.

`ExtendedExplainInfo` already exposes `getNativeExpressions` and `getCodegenDispatchExpressions`, and `CometCodegenSuite` already uses them. There was just no reusable helper, so writing the assertion was enough friction that nobody did.

- `CometTestBase.checkSparkAnswerAndImpl(df, native, dispatched)`, plus the underlying `assertExpressionImpl` split out so callers holding a plan can reuse it. Naming an expression asserts both that it ran through the expected mechanism and that it did not run through the other one, so a name is a claim rather than a hint.
- Two SQL file harness query modes, `expect_dispatch(<names>)` and `expect_native(<names>)`, accepting a comma-separated list. Both check results and coverage like a plain `query` first, then delegate to the same assertion.
- Documentation in the Comet SQL Tests guide (a section per mode, a tip on when to reach for them, and a step in "Adding a new test") and in the new-expression guide (the Scala helper alongside `checkSparkAnswerAndOperator`, and a tip in the SQL test list).
- `round.sql`, `lower.sql` and `upper.sql` annotated as worked examples.

Deliberately not included: making the assertion mandatory. Having the default `query` mode assert against a file-level declaration would catch this class of regression everywhere rather than only where someone annotated, but it needs a one-time pass over every fixture that already dispatches (`rlike`, `regexp_replace`, `split`, `lower`, `upper`, `round` on float and double, the `mask` family) and some would need version-conditional declarations. Worth doing once the opt-in form is in use; the alternatives section of #5609 records it.

New `SqlFileTestParserSuite` covers the two directives: single name, comma-separated list, surrounding whitespace, empty names dropped, and that the new patterns do not shadow `expect_fallback` / `expect_error` / `spark_answer_only` / `tolerance=` / `ignore`. Pure text parsing, so no Spark session. Registered in both `pr_build_linux.yml` and `pr_build_macos.yml`.

`CometCodegenSuite` gains a test that the helper actually fails. Against `SELECT abs(a), hypot(a, b)` (a known native/dispatched pair) it asserts the correct claim passes and that four wrong claims are rejected: each mechanism swapped, and a name the query does not contain, which is what a fixture typo looks like.

End to end via the annotated fixtures. Run against Spark 4.1:

| Suite | Result |
| --- | --- |
| `CometSqlFileTestSuite` | 467 succeeded, 0 failed |
| `CometExpressionSuite` | 141 succeeded, 0 failed, 4 ignored |
| `CometCodegenSuite`, `SqlFileTestParserSuite`, `CometMathExpressionSuite`, `CometStringExpressionSuite` | 134 succeeded, 0 failed |

`dev/ci/check-suites.py` passes with the new suite registered in both workflow files.

The annotations earned their keep on the first run. `SELECT round(123.456, 2), round(2.5, 0), round(3.5, 0), round(-2.5, 0), round(NULL, 0)` reported `round` in both sets, because `round(NULL, 0)` implicitly casts the untyped null to double and so dispatches while the decimal literals beside it stay native. The query is now split, with the reason in a comment.

Only the default Spark 4.1 profile has been exercised locally; the 3.4 / 3.5 / 4.0 profiles are left to CI. The annotated splits do not depend on Spark version (`CometRound` and `CometCaseConversionBase` branch on argument type and config, not on version), so no divergence is expected there.

## Note for reviewers

While writing this I noticed `lower_enabled.sql` and `upper_enabled.sql` set `spark.comet.expression.Lower.allowIncompatible=true` to reach the native path, but `CometCaseConversionBase` reports `Compatible` and branches on `spark.comet.caseConversion.enabled` instead, so that config is a no-op and both fixtures currently exercise the dispatcher despite their names. Left alone here rather than folded in. Happy to file it separately.


spark/src/test/scala/org/apache/spark/sql/CometTestBase.scala
```diff
@@ -342,6 +342,84 @@ abstract class CometTestBase
...
+      native: Seq[String],
+      dispatched: Seq[String]): Unit = {
+    val explainInfo = new ExtendedExplainInfo()
+    val actualNative = explainInfo.getNativeExpressions(cometPlan)
+    val actualDispatched = explainInfo.getCodegenDispatchExpressions(cometPlan)
+    def detail: String =
+      s"native=[${actualNative.mkString(", ")}] " +
+        s"codegen-dispatched=[${actualDispatched.mkString(", ")}]"
+    native.foreach { name =>
+      if (actualDispatched.contains(name)) {
```
sunchao, line 400: [P2] Account for expressions nested inside a dispatched subtree

Could the exclusion check account for descendants of a dispatched expression? With non-null Double columns in a nonempty Parquet table, Comet projection and codegen dispatch enabled, `SELECT abs(a), hypot(abs(b), c) FROM t` lowers the first `abs` natively but serializes the whole `hypot(abs(b), c)` tree to the JVM. Dispatch tags only `hypot`, so `native = Seq("abs"), dispatched = Seq("hypot")` can pass even though the nested `abs` runs in the JVM kernel. The new assertion therefore misses a composed case where the named expression uses both mechanisms. A nested-expression regression case would cover this missing classification. This is source-derived, not an executed reproduction.

spark/src/test/scala/org/apache/comet/CometSqlFileTestSuite.scala
```diff
@@ -155,6 +155,10 @@ class CometSqlFileTestSuite extends CometTestBase with AdaptiveSparkPlanHelper {
                     checkSparkAnswerAndOperatorWithTolerance(sql, tol)
                   case ExpectFallback(reason) =>
                     checkSparkAnswerAndFallbackReason(sql, reason)
+                  case ExpectDispatch(names) =>
```
sunchao, line 163: [P2] Accept the new implementation modes as positive sentinels

Could `requireSentinelForCodegenExpectError` also recognize `ExpectDispatch` and `ExpectNative`? Both branches call `checkSparkAnswerAndImpl`, which first performs the same answer/operator checks as a plain query. For a file with `spark.comet.exec.scalaUDF.codegen.enabled=true` and an `expect_error` record, upgrading its last plain positive query to either new mode makes preflight reject the file as missing a sentinel before any SQL runs. The stronger assertion should be able to serve as that successful control query without requiring a redundant plain query.

manuzhang:
@andygrove please rebase onto latest main branch.

manuzhang:
I agree with @sunchao’s comments. Other than those two points, the PR looks good to me.

sunchao (commented):
## Correctness

The update from `3921c004` to `b2cab4e3` preserves all 12 authored edits. Its 283-file increment matches the base branch's changes. I rechecked both existing P2 comments against the merged source, and both remain applicable: [nested expressions inside a dispatched subtree](https://github.com/apache/datafusion-comet/pull/5610#discussion_r3910116650) are still omitted from the dispatched-name set, and [the positive-sentinel gate](https://github.com/apache/datafusion-comet/pull/5610#discussion_r3910116654) still accepts only `CheckCoverageAndAnswer`. No new inline findings.

On the maintained Spark 3.5 and 4.0 branches, `Hypot` generates code for its children, including a nested `Abs`. The current Comet dispatcher still serializes that whole bound tree while annotating only its outer expression. The merged closure-serialization error guard and null-short-circuit correction do not change that classification. The harness still disables constant folding, and neither new query mode bypasses answer/operator checks or converts an execution failure into success. Maintained Spark 3.4 and 4.1 sources were unavailable, so this source comparison does not claim those versions.

The public snapshot through 2026-09-08 03:38:12 UTC reports 44 successful checks and 10 skipped checks. I verified that the inspected CI jobs checked out `c5230523`, whose parents are the assigned base `7f1e0018` and head `b2cab4e3`, and whose tree equals the head. The [Spark 3.5 expression job](https://github.com/apache/datafusion-comet/actions/runs/34132669013/job/101782341699) passed 1,326 tests with 12 canceled and 12 ignored. The [Spark 4.0 expression job](https://github.com/apache/datafusion-comet/actions/runs/34132669013/job/101782341803) passed 1,334 with 3 canceled and 12 ignored. Both ran the parser suite, the helper's negative controls, and the three annotated fixtures. Those controls do not cover either remaining finding. I did not run a local Spark/JNI reproduction.

## Performance

The update adds no work to the assertion helper beyond the already reviewed implementation. The merged production changes match the base branch, so they are not new performance changes from this PR. The CI results above are correctness evidence, with no new speedup claim or applicable expression microbenchmark.

## Design

The merged dispatcher guards preserve clean fallback on closure-serialization failure, but they do not address either outstanding test contract. The remaining changes still belong at the existing boundaries: account for descendants when classifying dispatch coverage, and let the stronger positive query modes satisfy the sentinel gate. No additional design issue emerged from the update.

## Abstraction & complexity

The update introduces no new helper, parser mode or indirection beyond the prior reviewed contribution. Both SQL modes still delegate to the same assertion helper. The two outstanding corrections can use those existing paths without adding a parallel assertion framework.

spark/src/test/scala/org/apache/spark/sql/CometTestBase.scala
andygrove: Fixed in b3fcf823e. You were right, and it was source-derivable — I confirmed it by writing the case you described and watching it pass when it should not have.

The cause is at the tagging site rather than in the exclusion check. `emitJvmCodegenDispatch` binds and closure-serializes the *whole* subtree into one kernel — its only inputs are the `AttributeReference`s the tree reads — but it named only the root. So for `hypot(abs(b), c)` the dispatched set held `hypot` alone, and the inner `abs` was in neither set.

Rather than teach the exclusion check about descendants, I made the classification true at the source: the dispatch site now also names every expression in the subtree. Attribute references and literals are left out, since they are the kernel's inputs rather than work it performed (and would otherwise show up in the coverage stats as "expressions").

That makes the helper reject both halves of your example, which is the point — `abs` is genuinely on both sides of the fence in that query, so neither claim about it alone is true:

```scala
val query = "SELECT abs(a), hypot(abs(b), c) FROM t"
intercept[TestFailedException] { checkSparkAnswerAndImpl(sql(query), native = Seq("abs")) }
intercept[TestFailedException] { checkSparkAnswerAndImpl(sql(query), dispatched = Seq("abs")) }
checkSparkAnswerAndImpl(sql(query), dispatched = Seq("hypot"))   // still unambiguous
```

That is *an expression nested inside a dispatched subtree is classified as dispatched* in `CometCodegenSuite`. Against the previous head the first `intercept` fails with "Expected exception ... but no exception was thrown", so it does reproduce your case rather than just documenting it.

**Two consequences worth flagging, since this tag is not only used by the test helper.**

The extended-explain line and the coverage stats read the same tag, so both got more accurate: a query like `sequence(a, comet_seq_stopper())` now reports `JVM codegen dispatcher: comet_seq_stopper, sequence` and counts 2 rather than 1. Previously the nested UDF was counted in neither bucket.

That broke two existing assertions in `CometCodegenSuite` that did `explain.contains("JVM codegen dispatcher: sequence")` — a substring match pinned to the name being first in a sorted list. I replaced that pattern everywhere it appeared with a `dispatchedNames(explain)` helper that parses the segment and checks membership, so it does not break again the next time a query dispatches a second expression. Full `CometCodegenSuite` (95) and `CometSqlFileTestSuite` (472) pass.

spark/src/test/scala/org/apache/comet/CometSqlFileTestSuite.scala
andygrove: Done in b3fcf823e — agreed, and the argument that they are *stronger* controls is the right one: both branches call `checkSparkAnswerAndImpl`, which runs the same answer and operator checks a plain `query` does before asserting anything about the mechanism. Refusing them meant the preflight pushed fixtures toward a weaker sentinel.

```scala
val hasSentinel = file.records.exists {
  case SqlQuery(_, CheckCoverageAndAnswer | _: ExpectDispatch | _: ExpectNative, _) => true
  case _ => false
}
```

I also updated the failure message, which still told authors to add a plain `query` — the advice would have been wrong as soon as this changed:

> Add at least one `query`, `expect_dispatch` or `expect_native` over valid input so the operator check fails if the expression did not execute natively.

sunchao (commented):
## Correctness

Re-reviewed `b3fcf823e840899aadf72b37f31951ae4de9ea83` against `7f1e00189b1ed86f1cb5acd872d97fce694482b1`, including the full 13-file contribution and the three-file update since [review 5137168012](https://github.com/apache/datafusion-comet/pull/5610#pullrequestreview-5137168012). Both previous P2 findings are addressed. One new P2 remains: the expanded dispatcher names change Q78's extended plan, but its approved plan has not been updated. The inline comment gives the failing CI evidence.

The [nested-expression finding](https://github.com/apache/datafusion-comet/pull/5610#discussion_r3910116650) is fixed by recording descendants of the subtree serialized into the JVM kernel. The new `hypot(abs(b), c)` regression rejects either exclusive assertion for `abs` when another `abs` runs natively. Attribute references and literals remain excluded. The [positive-sentinel finding](https://github.com/apache/datafusion-comet/pull/5610#discussion_r3910116654) is fixed by accepting both implementation modes. They still compare answers with Spark and check operator coverage before examining expression names. The sentinel branch was checked directly. No dedicated sentinel regression was added.

On the maintained Spark 3.5/4.0 branches, `Hypot` evaluates Double children with the normal binary null handling, while `Abs` retains its child's type and the existing ANSI/overflow branches. Binding substitutes input references while preserving the composed expression. The update changes coverage tags after successful dispatch preparation, without changing those computations, return-type checks, null handling, exceptions or fallback gates. The row-backed regression avoids constant folding. Existing round fixtures separate decimal/integer native cases from floating-point dispatch cases. Maintained Spark 3.4/4.1 sources are unavailable, so I make no canonical-source compatibility claim for those versions.

### Validation and CI

At `2026-09-09T18:43:24Z`, the current checks show 54 successes, five failures, nine skips and one queued check. All five Linux exec profiles (Spark 3.4, 3.5, 4.0, 4.1 and 4.2) fail only Q78's expected-plan comparison. The [Spark 3.5 expression job](https://github.com/apache/datafusion-comet/actions/runs/34372705635/job/102547005725) passed 1,346 tests with 13 canceled and 12 ignored. The [Spark 4.0 expression job](https://github.com/apache/datafusion-comet/actions/runs/34372705635/job/102547005677) passed 1,355 with three canceled and 12 ignored. Both passed the new nested-expression regression and the parser/round/lower/upper tests.

These jobs executed merge `e8f7f5aac9562c9e2408ec687593c81707d1d099`, combining this head with newer base `424c31aa79d13fddf743ffa29bae3c6f146e6c5e`. All 11 non-workflow contribution files, the golden file, the plan-stability suite and the explain/tag helpers match this head. The workflow files differ. `CometExecRule` and `QueryPlanSerde` have inherited changes, but their coverage rollup, lifting and self-dispatch sections match exactly. Builder/upload/download log identities also match for the native artifact. This is CI merge execution, not execution of the assigned base/head tree. Source/diff checks passed. No local build, runtime test or benchmark was run.

## Performance

The update adds a traversal of the dispatched expression subtree and set accumulation during planning. It does not add work to the per-row kernel or change argument serialization. The overhead grows with subtree size. There is no new performance claim or measurement. I found no material performance issue in this update.

## Design

Keeping descendant names on the dispatched root fits the existing coverage rollup and closes the mixed native/JVM blind spot. The separate marker for the root's own dispatch status remains necessary for native tagging. Recognizing the two stronger modes as positive sentinels preserves the runner's requirement for a successful comparison. The required follow-up is to regenerate the affected approved plan using the suite's existing workflow.

## Abstraction & complexity

The update reuses the existing tag sets and test modes without introducing another execution path. Parsing the dispatcher names as a set removes the old root-first substring assumption. The traversal and sentinel alternatives are localized. No additional abstraction or simplification is required for this fix.

spark/src/main/scala/org/apache/comet/serde/CometScalaUDF.scala
```diff
@@ -172,6 +172,18 @@ object CometScalaUDF extends CometExpressionSerde[ScalaUDF] {
...
+    // The whole subtree under `expr` was bound and closure-serialized into this one kernel, so
+    // every expression in it ran in the JVM, not just the root. Naming only the root understates
+    // that: for `hypot(abs(b), c)` the dispatched set would hold `hypot` alone, and a test could
+    // assert `abs` was native while an `abs` was in fact running inside the kernel. Attribute
+    // references and literals are the kernel's inputs rather than work it performed, so they are
+    // left out - which also keeps them from appearing in the coverage stats as "expressions".
+    target.foreach {
+      case _: AttributeReference | _: Literal =>
+      case node if !(node eq target) =>
+        withCodegenDispatchExpr(expr, CometExplainInfo.exprDisplayName(node))
```
sunchao, line 184: ### Correctness

[P2] Regenerate Q78's approved plan for the expanded dispatch names

Could you regenerate `tpcds-plan-stability/approved-plans-v2_7/q78/extended.txt` alongside this change? Recording descendants here changes Q78's dispatcher list from `round` to `add, cast, coalesce, divide, knownfloatingpointnormalized, normalizenanandzero, round`, and its dispatch count from 1 to 7, while the approved plan still contains the old values. All five Linux exec profiles fail this exact comparison at `CometPlanStabilitySuite.scala:172`. See the [Spark 3.5 job](https://github.com/apache/datafusion-comet/actions/runs/34372705635/job/102547005711) and [Spark 4.0 job](https://github.com/apache/datafusion-comet/actions/runs/34372705635/job/102547005707). I checked that this source and the approved plan are identical between the reviewed head and CI's merge. Updating the golden through the suite's regeneration workflow preserves the intended expanded coverage and restores the plan-stability test.

andygrove:
Good catch, thanks. Regenerated in a66c196bc.

I ran the base tier first, `SPARK_GENERATE_GOLDEN_FILES=1` against `-Pspark-3.4` for both `CometTPCDSV1_4_PlanStabilitySuite` and `CometTPCDSV2_7_PlanStabilitySuite`, since 3.4 is the only profile whose `planNameChain` resolves straight to `approved-plans-vX_Y` and therefore writes the shared base rather than a version-suffixed directory. v2_7's q78 came out with exactly the list you predicted:

```
-   +-  CometProject [COMET-INFO: JVM codegen dispatcher: round]
+   +-  CometProject [COMET-INFO: JVM codegen dispatcher: add, cast, coalesce, divide, knownfloatingpointnormalized, normalizenanandzero, round]
-...Accelerated expressions: 12 native, 1 codegen dispatch.
+...Accelerated expressions: 12 native, 7 codegen dispatch.
```

v1_4's q78 needed the same update and was not in your report. It records the dispatcher on two operators rather than one, and `CometTakeOrderedAndProject` picks up the shorter `add, cast, coalesce, divide, round` because the normalization pair only appears under the project. Worth knowing that the V1_4 suite fails on this too.

Then I re-ran both suites with regeneration on 3.5, 4.0, 4.1 and 4.2 to check whether any of them now diverges from the base and needs its own `approved-plans-*-sparkX_Y/q78` directory. All eight runs passed and the working tree came back with only those two base files modified, so `pruneDuplicateQueryDirs` removed nothing and added nothing: the expanded list is version-independent, as your five identical failures suggested. No other query moved, which is also the answer to whether recording descendants changes any golden beyond q78 — q24, q24a and q24b are the only other plans carrying a dispatcher line and none of them shifted.

comphead (approved):
Thanks @andygrove I had a quick view and it is a good follow up for https://github.com/apache/datafusion-comet/pull/4891

PR #5620, opened 2026-09-01
chore: add benches datetime

coderfender:

Part of https://github.com/apache/datafusion-comet/issues/5396

to / from etc timestamp. to / from timestamp with tz benches dded (per comet kernel implementation)

### DIFF
--- a/native/spark-expr/benches/timestamp_trunc.rs
+++ b/native/spark-expr/benches/timestamp_trunc.rs
@@ -1,0 +1,65 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use arrow::array::RecordBatch;
+use arrow::datatypes::{DataType, Field, Schema, TimeUnit};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use datafusion::physical_expr::expressions::{lit, Column};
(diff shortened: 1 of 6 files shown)


sunchao (approved):
Reviewed `f1b1b462ef69efbf08c7f596c338569d4edd0d69`. No P1/P2 findings in the four datetime benchmarks, shared timestamp helper, or target registrations. The timed loops invoke the intended functions over the input arrays and consume their outputs.

These measure native time per batch, including output destruction and UDF argument setup. They do not measure the complete composed `convert_timezone` expression or Spark-versus-Comet query performance. This was a source review with input-range and registration checks. The benchmarks were not built or run locally. Current GitHub checks show 66 successful and 8 skipped.

native/spark-expr/benches/timestamp_trunc.rs
```diff
@@ -0,0 +1,65 @@
...
+    )]));
+
+    let mut group = c.benchmark_group("timestamp_trunc");
+    for format in ["YEAR", "MONTH", "DAY", "HOUR"] {
+        let expr =
+            TimestampTruncExpr::new(Arc::new(Column::new("a", 0)), lit(format), TZ.to_string());
+        for rows in ROW_COUNTS {
+            for (null_ratio, tag) in NULL_RATIOS {
+                let ts = timestamp_micros_array(rows, null_ratio, Some(TZ), |i| {
+                    (i as i64) * MICROS_PER_DAY
```
manuzhang, line 50: is it possible to generate timestamps with varying hours/minutes/seconds

coderfender: Sure but this is more of a bench case than a unit test . Do you still think that it would make sense to add more inputs? @manuzhang

manuzhang: Not a blocking issue. I'm thinking the input could be more representative.

native/spark-expr/benches/unix_timestamp.rs
```diff
@@ -0,0 +1,76 @@
...
+    // through `array_with_timezone` first. Bench both so a regression in either path shows up.
+    for (variant, tz_input, udf_tz) in [
+        ("ntz", None, "UTC"),
+        (
+            "tz_aware",
+            Some("America/Los_Angeles"),
+            "America/Los_Angeles",
+        ),
+    ] {
+        let udf = SparkUnixTimestamp::new(udf_tz.to_string());
```
manuzhang, line 47: Not a blocking issue either, but `SparkUnixTimestamp` has a `Date32` path, which isn't covered here.

sunchao:
Merged, @coderfender please address @manuzhang 's comments in a follow-up if you consider necessary, thanks both!

PR #5622, opened 2026-09-01
feat: route `abs` on interval types through the codegen dispatcher

### LINKED ISSUE #5587: `abs` on interval types falls back to Spark
Part of #5572.

`abs` on an interval falls the whole projection back to Spark:

```scala
case _: NumericType =>
  Compatible()
case _ =>
  // Spark supports NumericType, DayTimeIntervalType, and YearMonthIntervalType
  Unsupported(Some(unsupportedReason))
```

`spark/src/main/scala/org/apache/comet/serde/math.scala:178-184`

The comment already names the gap. `docs/source/user-guide/latest/expressions.md` records it as "Interval types fall back".

Both interval families are explicitly supported by the dispatcher's type surface:

```scala
case _: YearMonthIntervalType | _: DayTimeIntervalType | CalendarIntervalType => true
```

`spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala:92`

so `canHandle` admits `abs(interval_col)` and Spark's own `doGenCode` gives the ANSI overflow behavior for `Long.MinValue` micros for free — which is worth having, since that is exactly the edge a native implementation would need to get right.

Fix: mix `CodegenDispatchFallback` into `CometAbs`.

kazantsev-maksim:

closes https://github.com/apache/datafusion-comet/issues/5587

## Rationale for this change

`abs` on interval types (`DayTimeIntervalType`, `YearMonthIntervalType`) has no native implementation, so `CometAbs` reported `Unsupported` and the entire projection fell back to Spark, even though Spark supports it.

The JVM codegen dispatcher already handles interval types (`CometBatchKernelCodegen.isSupportedDataType` admits them).

- `CometAbs` now mixes in `CodegenDispatchFallback`; interval inputs dispatch, numeric inputs keep the native path.
- Updated the misleading `unsupportedReason` (it now surfaces only when the dispatcher is disabled or rejects the tree).
- Regenerated expression compatibility docs (`abs` is now hybrid: intervals via dispatch, numerics natively).
- New `spark/src/test/resources/sql-tests/expressions/math/abs.sql`: numeric types (native), both interval families (dispatch), mixed projection; intervals covered both as inline literals (make_*_interval) and derived from numeric Parquet columns, including per-row dispatch, a null row, mixed-sign components and a mid-batch overflow (#5060 blocks reading interval columns directly). A pinned expect_fallback case documents the pre-existing year-month null literal gap (#5061).

New sql test cases added

Benchmarks (provided by @andygrove, run on this PR branch). The comparison toggles `spark.comet.exec.scalaUDF.codegen.enabled` inside a single JVM to reproduce `main`'s fallback plan — it is not a separately executed base build or a pure-Spark arm. 20M-row Parquet table, `local[*]`, noop sink, 3 warmup plus 12 measured iterations per arm:

Case | this PR | main-equivalent | ratio
-- | -- | -- | --
interval-only projection | 2701 ms | 2879 ms | 1.07x
mixed numeric and interval, 5 cols | 3318 ms | 3647 ms | 1.10x
numeric-heavy plus one interval | 2982 ms | 3250 ms | 1.09x
interval-only, repeat as a noise check | 2607 ms | 2875 ms | 1.10x


spark/src/test/resources/sql-tests/expressions/math/abs_ansi.sql
```diff
@@ -104,3 +104,8 @@ SELECT abs(v) FROM ansi_test_abs_byte
 -- literal
 query expect_error(overflow)
 SELECT abs(cast(-128 as tinyint))
+
+-- overflow: abs on Long.MinValue microseconds throws; the dispatched codegen path must
+-- propagate Spark's exception
+query expect_error(overflow)
+SELECT abs(make_dt_interval(-2147483648))
```
sunchao: [P2] Reach abs with a representable minimum interval

Could this construct `Long.MinValue` microseconds before calling `abs`? The first argument to `make_dt_interval` is days. `make_dt_interval(-2147483648)` overflows while multiplying days by microseconds per day, so evaluation never reaches `Abs`'s exact-negation check. This leaves the new path's minimum-value behavior untested. `make_dt_interval(-106751991, -4, 0, -54.775808)` constructs the actual minimum without overflowing its intermediate calculations. Please also cover `make_ym_interval(0, -2147483648)`, with ANSI both enabled and disabled, since interval `abs` checks overflow in both modes. Pair these error cases with a valid nearby interval query that requires Comet execution. The `expect_error` helper alone does not assert that the dispatcher ran.

sunchao (commented):
Reviewed `cf539183972a203ed8fe6ea8f77a0906880c6bd2` against `1e10eedd6e0303adcac4573f44d5c07ffb0fbacd`. Numeric `abs` retains its existing native path, while interval inputs use Spark's generated code. One P2 concerns the new overflow regression test, not a demonstrated production wrong result: its interval constructor throws before reaching `abs`.

Could you add a representative microbenchmark for the added dispatch path, using interval values constructed from numeric columns and a mixed numeric/interval projection? Please compare Spark, the base revision's fallback and this revision with matched input, batch size and concurrency, check equal results and confirm executed plans. This would measure whether retaining the Comet projection offsets dispatch and vector-conversion costs.

The finding and replacement boundary values are source-derived. No build, product test or benchmark was run. Current-head workflows report `action_required`, with no test results.

spark/src/test/resources/sql-tests/expressions/math/abs_ansi.sql
kazantsev-maksim: Thanks for this catch! Fixed: `abs(make_dt_interval(-106751991, -4, 0, -54.775808))` (exactly Long.MinValue µs, error now raised by `Abs` itself) and added `abs(make_ym_interval(0, -2147483648))` in both ANSI and non-ANSI files — Spark's `Abs` on interval types calls `negateExact` unconditionally, so the overflow is expected in both modes.

sunchao (approved):
Re-reviewed `e8238e4a` against `1e10eedd`. The replacement constructors address the [P2 overflow-witness finding](https://github.com/apache/datafusion-comet/pull/5622#discussion_r3910332373). The day-time construction reaches exactly `Long.MinValue` without overflowing its intermediates, and `make_ym_interval(0, -2147483648)` reaches `Int.MinValue`. Maintained Spark 3.5 and 4.0 check interval negation in both ANSI modes. I found no remaining P1/P2.

This conclusion is from pinned source and checked arithmetic, including nine independent arithmetic-model checks, not a Spark/JNI run. The existing successful non-ANSI interval queries require Comet coverage, but the error assertions alone do not prove dispatch. A nearby-boundary ANSI success query was not added. The matched benchmark request remains unanswered, so no performance improvement is verified.

All three current workflows remain `action_required`, with no head or merge check results. I did not approve or rerun CI.

spark/src/main/scala/org/apache/comet/serde/math.scala
```diff
@@ -169,9 +169,11 @@ object CometUnhex extends CometExpressionSerde[Unhex] with MathExprBase {
   }
 }
 
-object CometAbs extends CometExpressionSerde[Abs] with MathExprBase {
+object CometAbs extends CometExpressionSerde[Abs] with MathExprBase with CodegenDispatchFallback {
 
-  val unsupportedReason: String = "Only integral, floating-point, and decimal types are supported"
+  val unsupportedReason: String =
+    "Interval types are not supported natively and are handled via JVM codegen dispatch; " +
+      "this fallback only applies when the dispatcher is disabled"
```
andygrove: This string does double duty as the `Unsupported` note that shows up in EXPLAIN and as the `getUnsupportedReasons()` bullet rendered into the generated compat docs, and I think it reads awkwardly in both.

`GenerateDocs.writeExpressionCompatNotes` already prints the mechanism as a header for any `CodegenDispatchFallback` serde, so the generated `math.md` says the same thing twice:

```
The following cases have no native implementation and always run in the JVM using Spark's
code-generated implementation (inside the Comet pipeline):

- Interval types are not supported natively and are handled via JVM codegen dispatch; this
  fallback only applies when the dispatcher is disabled
```

And in EXPLAIN with the dispatcher off it contradicts the reason sitting next to it. This is what I get with `spark.comet.exec.scalaUDF.codegen.enabled=false`:

```
Project [COMET: abs: spark.comet.exec.scalaUDF.codegen.enabled=false; expression has no native path so the plan falls back to Spark, Interval types are not supported natively and are handled via JVM codegen dispatch; this fallback only applies when the dispatcher is disabled]
```

`CometRound` in `arithmetic.scala` has exactly this shape and keeps the two concerns apart. The reasons name only the case, and a class-level scaladoc explains the dispatch. Would you follow that here?

```scala
/**
 * `abs` lowers to the native `abs` kernel for numeric inputs. Interval inputs have no native
 * implementation, so `CodegenDispatchFallback` keeps them in the Comet pipeline by running
 * Spark's own `Abs.doGenCode` in the JVM codegen dispatcher, which matches Spark exactly.
 */
object CometAbs extends CometExpressionSerde[Abs] with MathExprBase with CodegenDispatchFallback {

  val unsupportedReason: String =
    "`INTERVAL YEAR TO MONTH` and `INTERVAL DAY TO SECOND` inputs"
```

spark/src/test/resources/sql-tests/expressions/math/abs.sql
```diff
@@ -27,3 +27,25 @@ SELECT abs(i), abs(l), abs(f), abs(d) FROM test_abs
...
+SELECT abs(make_dt_interval(1, 2, 3, 4.5)) AS dt_pos,
+       abs(make_dt_interval(-1, -2, -3, -4.5)) AS dt_neg,
+       abs(make_dt_interval(0, 0, 0, 0)) AS dt_zero,
+       abs(CAST(NULL AS INTERVAL DAY TO SECOND)) AS dt_null
+
+-- interval year to month: dispatched the same way; wrapped in a struct because a top-level
+-- YearMonthIntervalType column is rejected by the projection output type gate (https://github.com/apache/datafusion-comet/issues/5061)
+query
+SELECT named_struct('v', abs(make_ym_interval(1, 6))) AS ym_pos,
+       named_struct('v', abs(make_ym_interval(-1, -6))) AS ym_neg
```
andygrove: I do not think this wrapper is needed. I ran the query verbatim with `named_struct` removed, under a strict `checkSparkAnswerAndOperator`, and it stays fully native both over `OneRowRelation` and over a Parquet-backed table:

```sql
SELECT abs(make_ym_interval(1, 6)) AS ym_pos, abs(make_ym_interval(-1, -6)) AS ym_neg
```

There is no projection output-type gate for `YearMonthIntervalType`. `QueryPlanSerde.supportedDataType` does reject both ANSI interval types, but nothing consults it for a projection's output. Its callers are `CometLiteral`, `CometSink`, `CometScalarSubquery` and `hash.scala`, and `CometSink.supportedSinkDataType` admits both interval types explicitly.

The wrapper also costs coverage. With it, extended explain reports `1 native, 1 codegen dispatch`, where the native expression is the `named_struct` itself. The interval value never becomes a top-level output column, so the `IntervalYearVector` output path is never exercised. Without the wrapper you get `0 native, 1 codegen dispatch` and that path is covered.

Here is what I measured, all with `ConstantFolding` excluded and ANSI off:

| query | fully native |
|---|---|
| `SELECT abs(make_ym_interval(1, 6))` | yes |
| `SELECT abs(make_ym_interval(1, 6)), abs(make_ym_interval(-1, -6))` | yes |
| `SELECT abs(make_ym_interval(y, m)) FROM t`, Parquet, 5 rows including a null | yes |
| `SELECT abs(CAST(NULL AS INTERVAL YEAR TO MONTH))` | no |
| `SELECT CAST(NULL AS INTERVAL YEAR TO MONTH)` | no |

Those last two rows are what I suspect you actually ran into, and it is a different problem. See my next comment.

Could you drop the wrapper and both `#5061` references? Worth noting that `#5061` is the interval EPIC rather than an issue about a projection gate, so the link is misleading on its own even setting the mechanism aside.

andygrove: Separately from the wrapper, this query has no null case while the `dt` query above has `dt_null`. I would add one, because it turns up a real gap rather than just passing:

```
Project [COMET: Unsupported data type YearMonthIntervalType(0,1)]
```

`NullPropagation` folds `abs(CAST(NULL AS INTERVAL YEAR TO MONTH))` down to a bare `Literal(null, YearMonthIntervalType)`, and `CometLiteral.getSupportLevel` special-cases only `DayTimeIntervalType` (`literals.scala:63`), so the literal is `Unsupported`. `CometLiteral` does not mix in `CodegenDispatchFallback`, so there is no dispatch rescue and the whole projection falls back.

This is pre-existing rather than anything you introduced. `SELECT CAST(NULL AS INTERVAL YEAR TO MONTH)` falls back the same way with no `abs` in the query at all, and it is already a bullet in #5061. But this PR is what makes it reachable through `abs`, so I would rather pin it than have it silently avoided:

```sql
query expect_fallback(Unsupported data type YearMonthIntervalType)
SELECT abs(CAST(NULL AS INTERVAL YEAR TO MONTH))
```

That way it flips to a test failure when the literal gap gets fixed.

spark/src/test/resources/sql-tests/expressions/math/abs.sql
```diff
@@ -27,3 +27,25 @@ SELECT abs(i), abs(l), abs(f), abs(d) FROM test_abs
...
+
+-- abs() on intervals has no native impl; routed through the JVM codegen dispatcher.
+-- Interval values are built inline: native Parquet scan of interval columns is unsupported
+-- (https://github.com/apache/datafusion-comet/issues/5060), and a top-level YearMonthIntervalType projection column is still rejected by the
+-- projection type gate (https://github.com/apache/datafusion-comet/issues/5061), so the ym result is wrapped in a struct.
+query
+SELECT abs(make_dt_interval(1, 2, 3, 4.5)) AS dt_pos,
+       abs(make_dt_interval(-1, -2, -3, -4.5)) AS dt_neg,
+       abs(make_dt_interval(0, 0, 0, 0)) AS dt_zero,
+       abs(CAST(NULL AS INTERVAL DAY TO SECOND)) AS dt_null
```
andygrove, line 35: All four interval queries in this file use compile-time constants, so every row sees the same input. Excluding `ConstantFolding` does make the kernel run per row, but a bug in row indexing, the null mask, or a mid-batch overflow would still not be caught.

#5060 does block reading an interval column, but you can vary the value per row by deriving it from an integer column, which is also the shape a real query has. I ran these against the PR and they pass, so this is coverage rather than a bug:

```sql
statement
CREATE TABLE test_abs_iv(d int, h int, m int, s decimal(8,6)) USING parquet

statement
INSERT INTO test_abs_iv VALUES (1, 2, 3, 4.5), (-1, -2, -3, -4.5), (0, 0, 0, 0), (NULL, 0, 0, 0), (5, -1, 30, -0.000001)

query
SELECT abs(make_dt_interval(d, h, m, s)) FROM test_abs_iv ORDER BY d
```

A mid-batch overflow case is worth adding on the same table. I checked that one `Long.MinValue` row among valid rows raises the same message Spark does, but nothing here asserts it, since the existing overflow queries are single-row constants.

#5061 calls this trap out directly: "a single-row `VALUES (make_dt_interval(...))` gets collapsed by `ConvertToLocalRelation` into a folded literal ... Derive intervals from Parquet-backed integer columns instead.".

andygrove (commented):
Thanks for picking this up, and sorry for the slow second look.

The serde change is exactly what #5587 asked for, so no objection to the approach. I built the PR locally and probed it fairly hard: column-varying intervals over Parquet with mixed signs, zeros and nulls, `abs(ts1 - ts2)`, mixed numeric and interval expressions in one projection, both ANSI modes, and a mid-batch overflow with one `Long.MinValue` row among valid rows. Every one of those stays fully native and matches Spark exactly, including byte-identical `[ARITHMETIC_OVERFLOW]` messages. So I have no correctness concerns.

I also ran the benchmark that was asked for earlier, since it was quicker to just do it. Toggling `spark.comet.exec.scalaUDF.codegen.enabled` inside a single JVM reproduces `main`'s plan exactly (`Project` over `CometColumnarToRow` over `CometNativeScan`), which avoids rebuild and machine variance. 20M-row Parquet table, `local[*]`, noop sink, 3 warmup plus 12 measured iterations per arm:

| Case | this PR | main-equivalent | ratio |
|---|---|---|---|
| interval-only projection | 2701 ms | 2879 ms | 1.07x |
| mixed numeric and interval, 5 cols | 3318 ms | 3647 ms | 1.10x |
| numeric-heavy plus one interval | 2982 ms | 3250 ms | 1.09x |
| interval-only, repeat as a noise check | 2607 ms | 2875 ms | 1.10x |

No regression anywhere and a consistent 7 to 10 percent gain. The gain is modest because every arm still ends in a columnar-to-row transition for the sink. When the projection feeds another native operator that transition disappears and the win should be larger. Please put these numbers in the PR description.

My comments below are all about the tests and the reason string. The one I care most about is the `named_struct` wrapper, which I believe is unnecessary and is justified by a gate that does not exist.

spark/src/test/resources/sql-tests/expressions/math/abs_ansi.sql
```diff
@@ -104,3 +104,9 @@ SELECT abs(v) FROM ansi_test_abs_byte
 -- literal
 query expect_error(overflow)
 SELECT abs(cast(-128 as tinyint))
+
+query expect_error(overflow)
+SELECT abs(make_dt_interval(-106751991, -4, 0, -54.775808))
+
+query expect_error(overflow)
+SELECT abs(make_ym_interval(0, -2147483648))
```
andygrove, line 118: This file carries `MaxSparkVersion: 4.1`, so on the Spark 4.2 CI job it is skipped and `abs_ansi_spark42.sql` runs instead. Because these two cases were only added here, 4.2 and later get no ANSI interval coverage at all. Every other case in that pair exists in both files. Could you mirror these two into `abs_ansi_spark42.sql` with the loose `overflow` pattern it already uses?

One thing that might save you work. Spark's `Abs.doGenCode` sends `_: AnsiIntervalType` to `MathUtils.negateExact` unconditionally, with no `failOnError` check, so interval overflow behavior is identical in both ANSI modes and the `abs.sql` cases already assert it. Keeping these ANSI copies to pin that independence is fine by me, but then they belong in both ANSI fixtures rather than one.

sunchao (commented):
## Correctness

Discussion follow-up on `e8238e4a6009eb845f3ea26f1b3cab7df46782e1` against `1e10eedd6e0303adcac4573f44d5c07ffb0fbacd`. The source pair is unchanged from my approval. Numeric `abs` retains its native path, and interval `abs` uses Spark's generated expression inside the Comet pipeline. I found no new P1/P2 that changes that approval.

Maintained Spark 3.5/4.0 source confirms checked interval negation regardless of ANSI mode. A fresh exact-integer model confirms the corrected constructors reach `Long.MinValue` microseconds and `Int.MinValue` months without constructor overflow. This was an arithmetic check, not Spark execution. Maintained Spark 3.4/4.1 sources were unavailable.

I agree with the existing [column-varying coverage comment](https://github.com/apache/datafusion-comet/pull/5622#discussion_r3941100587): derive intervals from numeric Parquet columns, cover mixed signs/zero/null and a minimum-value row among valid rows, and pair overflow assertions with valid queries that require Comet execution in both ANSI modes. `expect_error` checks matching exceptions but does not assert the execution path. The [missing Spark 4.2 ANSI fixtures](https://github.com/apache/datafusion-comet/pull/5622#discussion_r3941100594) are real: `abs_ansi.sql` is capped at 4.1, and its 4.2 counterpart lacks both interval cases.

The [null year-month fallback](https://github.com/apache/datafusion-comet/pull/5622#discussion_r3941100581) is also source-confirmed and pre-existing: `NullPropagation` produces a typed null literal, while `CometLiteral` admits day-time intervals but not year-month intervals. A targeted expected-fallback regression is appropriate; this does not indicate broken interval `Abs` dispatch.

At the captured cutoff, `2026-09-05T21:38:58.017Z`, checks showed 64 successes and 9 skips. I inspected the [Spark 4.0 expressions job](https://github.com/apache/datafusion-comet/actions/runs/33785703974/job/101108772816): 1,296 tests succeeded, 0 failed, 3 were canceled and 12 ignored. Both applicable `abs` fixtures ran. CI checked out historical merge `4c24cc74`, containing this head and base-side `90c1dd96`. All four authored files match, but that integration includes a shared codegen null-shortcut fix absent from the standalone head. This is evidence for that integration, not full-tree equivalence. No local Spark/JNI/native test or benchmark ran.

## Performance

[Andy's reported measurements](https://github.com/apache/datafusion-comet/pull/5622#pullrequestreview-5121858367) provide useful representative evidence: 20M Parquet rows, three warmups and twelve measured iterations per arm, with reported 1.07–1.10x ratios across interval-only and mixed projections. These are third-party results. The comparison toggles dispatch in one JVM to reproduce the fallback plan; it is not a separately executed base build or a pure-Spark arm. Preserve those labels and settings when adding the table to the description. A larger gain with another downstream native operator remains unmeasured.

## Design

The existing dispatch abstraction remains appropriate. I agree with [shortening the reason string](https://github.com/apache/datafusion-comet/pull/5622#discussion_r3941100570) to identify the unsupported interval cases and moving the routing explanation into Scaladoc. Generated documentation already supplies that explanation, and dispatch can decline a tree for reasons beyond the enable flag.

## Abstraction & complexity

The [year-month `named_struct` wrapper](https://github.com/apache/datafusion-comet/pull/5622#discussion_r3941100579) is unnecessary in these successful cases: projection conversion has no such output-type gate, interval sink types are admitted, and the dispatcher has a top-level `IntervalYearVector` output path. Direct year-month outputs give better coverage and remove the misleading gate explanation. These improvements fit the current implementation without another abstraction.

kazantsev-maksim:
@sunchao @andygrove, Thank you both for the detailed reviews.

spark/src/main/scala/org/apache/comet/serde/math.scala
kazantsev-maksim: fixed

sunchao (approved):
## Correctness

Re-reviewed `e96fef85` against `7e198439`, including the changes since `e8238e4a`. I found no new or remaining P1/P2 issue. The new day-time interval queries use varying numeric Parquet columns, including null and mixed-sign inputs. The overflow table constructs exactly `Long.MinValue` microseconds, and both ANSI fixtures now pair the signed-minimum error cases with valid adjacent values. The direct year-month projection and explicit typed-null fallback also address the earlier discussion.

The maintained Spark 3.5/4.0 implementations use checked interval negation independently of ANSI mode. I reverified those sources and 26 exact-integer model checks for the fixture boundaries. The SQL harness excludes constant folding and ordinary queries require result and operator coverage. Error queries alone do not establish dispatch coverage. I did not run Spark, JNI or Comet tests locally. At 2026-09-08T17:44:37.554353+00:00, all eight current-head workflow attempts required maintainer action and exposed zero jobs, so this approval does not claim passing current-head CI. Maintained Spark 3.4/4.1 sources were unavailable. Spark 4.2 coverage here is the mirrored fixture and version-selection check.

## Performance

The numeric conversion is unchanged from both the prior revision and the assigned base. The PR body now attributes the 1.07–1.10x measurements to @andygrove and identifies them as dispatcher enabled versus disabled on the PR branch in the same JVM. Those are reported measurements, not a separately executed base comparison or pure-Spark benchmark. I did not reproduce them on this revision, and the source review does not establish a new measured speedup.

## Design

The shorter unsupported reason removes the incorrect claim that fallback occurs only when dispatch is disabled. The added Scaladoc explains the native numeric path and Spark-generated interval path, while the existing dispatcher still handles eligibility and serialization fallback. The inherited null-shortcut hardening is present in the assigned base. The updated Hybrid documentation row is consistent with this routing.

## Abstraction & complexity

Removing the year-month struct wrapper makes the test exercise the direct interval output without introducing another helper or special conversion. The typed-null case records the existing literal-support gap explicitly, and the ANSI cases are mirrored across the fixture version boundary. These changes keep the implementation small and make the test intent clearer. No additional blocking change is requested.

andygrove (approved):
Both of my inline points are addressed, and the second one landed better than I asked for.

The reason string now names only the case, `` `INTERVAL YEAR TO MONTH` and `INTERVAL DAY TO SECOND` inputs ``, with the mechanism moved into a class-level scaladoc. That matches `CometRound`'s shape, so the generated `math.md` no longer states the dispatch twice and EXPLAIN with the dispatcher off no longer prints a reason that contradicts the one beside it.

The `named_struct` wrapper is gone, so `abs(make_ym_interval(1, 6))` is a top-level output column now and the `IntervalYearVector` output path is actually exercised rather than hidden behind a native struct constructor. Nice addition on top of that: the `expect_fallback(Unsupported data type YearMonthIntervalType)` case documents the pre-existing literal gap in #5061 and is written so it turns into a failure when that is fixed, which is a better outcome than either dropping the case or leaving it silently passing for the wrong reason.

I re-ran the fixture on this head against a rebuilt native library and `CometSqlFileTestSuite abs` passes 3 of 3. Thanks for putting the benchmark numbers in the description with the methodology caveat attached; the toggle-in-one-JVM framing matters and it is stated plainly.

The branch is behind `main` but merges cleanly. Approving.

PR #5623, opened 2026-09-01
fix: skip codegen dispatcher null short-circuit when a foldable subtree can raise

### LINKED ISSUE #5608: Codegen dispatcher single-ordinal null short-circuit swallows ANSI errors from literal subtrees
### Describe the bug

The JVM codegen dispatcher's null short-circuit can swallow an ANSI error that Spark raises, returning a row where Spark fails the query.

`CometBatchKernelCodegen.canShortCircuitNulls` allows the short-circuit when the dispatched tree reads exactly one input ordinal, on the reasoning quoted in its own scaladoc:

> One ordinal: there is nothing left for Spark to evaluate ahead of that ordinal's own null check.

That is not true. A literal-only subtree between the root and the ordinal can still raise. `ConstantFolding` normally folds such a subtree away, but it deliberately leaves it in place when the subtree throws and sits inside a conditional branch (it tags the node `FAILED_TO_EVALUATE` and moves on), so the throwing expression survives into the physical plan.

This is the residual hole in #5219 / #5218. That fix added `rootChildrenAreLeaves` for the multi-ordinal case but left the single-ordinal branch unguarded.

Found by @sunchao while reviewing #5607.

### Steps to reproduce

Spark 4.1, ANSI enabled (the default on 4.x):

```sql
CREATE TABLE t (flag BOOLEAN, n INT) USING parquet;
INSERT INTO t VALUES (true, NULL), (false, NULL);

SELECT IF(flag, upper(substring('abc', CAST(1L DIV 0L AS INT), n)), NULL) FROM t;
```

Spark raises `[DIVIDE_BY_ZERO]`. Comet returns a row.

Three things have to line up, and all three are ordinary:

- `ConstantFolding` refuses to fold `1L DIV 0L` because it sits under an `If` branch, so the throwing literal reaches execution.
- `TernaryExpression.nullSafeCodeGen` emits `Substring`'s `pos` code before it tests `len`'s null, so Spark evaluates the division even though `n` is NULL.
- `Upper`, `Substring`, `Cast` and `IntegralDivide` are all null-intolerant and the dispatched tree reads exactly one ordinal, so `canShortCircuitNulls` takes the single-ordinal branch and the kernel writes NULL before `ev.code` runs.

`upper` is just a convenient witness. Any dispatched null-intolerant root with a throwing foldable subtree between it and its single input reproduces it, which is most of the ~70 expressions that route through the dispatcher. #5607 (routing `length` / `bit_length` / `octet_length` on `BinaryType` through the dispatcher) exposes three more:

```sql
SELECT IF(flag, length(substring(X'00', CAST(1L DIV 0L AS INT), n)), 0) FROM t;
SELECT IF(flag, bit_length(substring(X'00', CAST(1L DIV 0L AS INT), n)), 0) FROM t;
SELECT IF(flag, octet_length(substring(X'00', CAST(1L DIV 0L AS INT), n)), 0) FROM t;
```

### Expected behavior

Comet raises the same `[DIVIDE_BY_ZERO]` Spark does.

### Suggested fix

Tighten `canShortCircuitNulls`: in addition to the existing conditions, require that no node in the tree other than a `Literal` is `foldable`. After `ConstantFolding` has run, a surviving foldable non-`Literal` node is precisely one that threw during folding, which is exactly the dangerous case.

This keeps the existing fast paths, none of which contain a foldable non-`Literal` node: `upper(substring(s, 1, 2))`, `pmod(a, b)`, `a + b`, `conv(a, b, c)`, `make_timestamp(...)`.

### Regression test

```scala
test("single-input short-circuit does not swallow an ANSI error from a literal subtree") {
  // `canShortCircuitNulls` assumes a single input ordinal leaves Spark nothing to evaluate ahead
  // of that ordinal's null check. Not true when the tree carries a foldable subtree that throws:
  // `ConstantFolding` leaves it in place under a conditional branch, and `Substring`'s `pos` is
  // evaluated before `len`'s null is tested, so Spark raises on the (true, NULL) row while the
  // kernel short-circuits and returns NULL.
  withTable("t") {
    sql("CREATE TABLE t (flag BOOLEAN, n INT) USING parquet")
    sql("INSERT INTO t VALUES (true, NULL), (false, NULL)")
    withSQLConf(SQLConf.ANSI_ENABLED.key -> "true") {
      Seq(
        "upper(substring('abc', CAST(1L DIV 0L AS INT), n))",
        "length(substring(X'00', CAST(1L DIV 0L AS INT), n))",
        "bit_length(substring(X'00', CAST(1L DIV 0L AS INT), n))",
        "octet_length(substring(X'00', CAST(1L DIV 0L AS INT), n))").foreach { e =>
        val (sparkErr, cometErr) =
          checkSparkAnswerMaybeThrows(sql(s"SELECT IF(flag, $e, NULL) FROM t"))
        assert(sparkErr.isDefined, s"$e: Spark no longer raises, the row is not a witness")
        assert(cometErr.isDefined, s"$e: Comet returned a value where Spark raised")
        assert(cometErr.get.getMessage.contains("DIVIDE_BY_ZERO"))
      }
    }
  }
}
```

The `length` / `bit_length` / `octet_length` cases only apply once #5607 lands.

### Additional context

Verified end to end on Spark 4.1 against the #5607 branch. The `upper` case is unaffected by that PR and reproduces on `main`.

andygrove:

Closes #5608.

## Rationale for this change

`CometBatchKernelCodegen.canShortCircuitNulls` allows the pre-`ev.code` null short-circuit whenever the dispatched tree reads exactly one input ordinal, on the reasoning that a single ordinal leaves Spark nothing to evaluate ahead of that ordinal's own null check.

That is not true. A literal-only subtree between the root and the ordinal can still raise. `ConstantFolding` normally folds such a subtree away, but it deliberately leaves it in place when evaluating it throws and it sits inside a conditional branch (it tags the node `FAILED_TO_EVALUATE` and moves on), so the throwing expression survives into the physical plan. The kernel then writes NULL before `ev.code` runs, and an ANSI error Spark raises is silently swallowed:

```sql
CREATE TABLE t (flag BOOLEAN, n INT) USING parquet;
INSERT INTO t VALUES (true, NULL), (false, NULL);

SELECT IF(flag, upper(substring('abc', CAST(1L DIV 0L AS INT), n)), NULL) FROM t;
```

Spark raises `[DIVIDE_BY_ZERO]`; Comet returned a row. Three ordinary things line up: `ConstantFolding` refuses to fold `1L DIV 0L` under the `If` branch, `TernaryExpression.nullSafeCodeGen` emits `Substring`'s `pos` code before it tests `len`'s null, and `Upper` / `Substring` / `Cast` / `IntegralDivide` are all null-intolerant over a single ordinal.

This is the residual hole in #5218: that fix added `rootChildrenAreLeaves` for the multi-ordinal case but left the single-ordinal branch unguarded. `upper` is just a convenient witness; any dispatched null-intolerant root with a throwing foldable subtree between it and its single input reproduces it.

Adds a `noSurvivingFoldableSubtree` condition to `canShortCircuitNulls`: no node in the tree other than a `Literal` may be foldable. By the time the dispatcher sees the tree `ConstantFolding` has already run, so a surviving foldable non-`Literal` node is precisely one that threw during folding, which is exactly the dangerous case.

`Literal`s are exempt, so the existing fast paths are untouched: none of `upper(substring(s, 1, 2))`, `pmod(a, b)`, `a + b`, `conv(a, b, c)` or `make_timestamp(...)` contains a foldable non-`Literal` node. The scaladoc on `canShortCircuitNulls` is updated to record the new condition.

- `CometCodegenSuite`: a new end-to-end test asserting Comet raises the same `DIVIDE_BY_ZERO` Spark does for the query above, and that the codegen dispatcher actually ran for it. Verified to fail on `main` (`cometErr.isDefined was false`) and pass with the fix, on the `spark-3.4`, `spark-3.5`, `spark-4.0` and `spark-4.1` profiles.
- `CometCodegenSourceSuite`: two generated-source tests, one asserting the short-circuit is not emitted for a single-ordinal tree carrying a throwing foldable subtree, and a counterpart asserting it *is* still emitted when the only foldable nodes are `Literal`s, so the fix is not over-corrected.
- Full `CometCodegenSuite`, `CometCodegenSourceSuite`, `CometCodegenFuzzSuite` and `CometSpecializedGettersDispatchSuite` pass (177 tests).

The `length` / `bit_length` / `octet_length` witnesses listed in the issue are left out here because they only apply once #5607 lands.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
+++ b/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
@@ -437,10 +437,13 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    *   - The root is `NullIntolerant`, so a null input really does mean a null result.
    *   - Every node in the tree is null-propagating ([[allNullIntolerant]]); a `Coalesce` / `If` /
    *     `CaseWhen` anywhere would break the chain.
+   *   - No node other than a `Literal` is foldable ([[noSurvivingFoldableSubtree]]); a foldable
+   *     subtree that `ConstantFolding` left in place is one that threw while folding, and Spark
+   *     may evaluate it -- and raise -- ahead of an input's null check.
    *   - Either the tree reads exactly one ordinal, or every direct child of the root is a leaf
    *     ([[rootChildrenAreLeaves]]). Both shapes make the short-circuit exact:
-   *     - One ordinal: there is nothing left for Spark to evaluate ahead of that ordinal's own
-   *       null check.
+   *     - One ordinal: with no surviving foldable subtree, there is nothing left for Spark to
+   *       evaluate ahead of that ordinal's own null check.
    *     - Leaf-only children: the tree is one level deep, so the only code Spark runs ahead of
    *       its null checks is `BoundReference` / `Literal` reads, which cannot raise. Any error
    *       comes from the root's own logic, which runs only once every input is known non-null --
@@ -457,7 +460,33 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
    */
   private def canShortCircuitNulls(expr: Expression, inputOrdinals: Seq[Int]): Boolean =
     inputOrdinals.nonEmpty && isNullIntolerant(expr) && allNullIntolerant(expr) &&
+      noSurvivingFoldableSubtree(expr) &&
       (inputOrdinals.size == 1 || rootChildrenAreLeaves(expr))
+
+  /**
+   * True iff no node in the tree other than a `Literal` is foldable. One of the conditions on
+   * [[canShortCircuitNulls]], and what closes the single-ordinal hole in #5218 (see #5608).
+   *
+   * A single input ordinal does not by itself mean Spark has nothing to evaluate ahead of that
+   * ordinal's null check: a literal-only subtree between the root and the ordinal can still
+   * raise. `ConstantFolding` normally folds such a subtree away, but it deliberately leaves it in
+   * place when evaluating it throws and it sits inside a conditional branch (it tags the node
+   * `FAILED_TO_EVALUATE` and moves on), so the throwing expression survives into the physical
+   * plan. `IF(flag, upper(substring('abc', CAST(1L DIV 0L AS INT), n)), NULL)` is a witness:
+   * `TernaryExpression.nullSafeCodeGen` emits `Substring`'s `pos` code -- the division -- before
+   * it tests `len`'s null, so Spark raises `DIVIDE_BY_ZERO` on a row where `n` is null while the
+   * short-circuit would return null.
+   *
+   * By the time the dispatcher sees the tree `ConstantFolding` has already run, so a surviving
+   * foldable non-`Literal` node is precisely one that threw during folding -- exactly the
+   * dangerous case. `Literal`s are foldable by definition and must be exempt, or the fast path
+   * would be lost for the common shapes (`upper(substring(s, 1, 2))`, `conv(a, b, c)`, ...).
+   */
+  private def noSurvivingFoldableSubtree(expr: Expression): Boolean =
+    !expr.exists {
+      case _: Literal => false
+      case other => other.foldable
+    }
 
   /**
    * True iff every direct child of the root is a leaf (`BoundReference` or `Literal`), i.e. the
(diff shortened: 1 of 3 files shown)


sunchao (approved):
Reviewed `ae5eb02ec73c326115b9bb73a34bbbc5ee5a6c78` against `1e10eedd6e0303adcac4573f44d5c07ffb0fbacd`. No new P1/P2 findings. The guard preserves evaluation of surviving foldable subtrees before an input's null check while retaining the literal-only fast path. The regression covers conditional constant folding and verifies dispatcher use.

This was a source review, with no local test execution. Current CI has 63 successful checks, 9 skipped and 1 failed Spark 4.1 SQL core shard. I inspected the [failed job log](https://github.com/apache/datafusion-comet/actions/runs/33545831145/job/99988290560): `SQLAppStatusListenerMemoryLeakSuite`'s `no memory leak` test fails because `statusStore.listener.get.noLiveData()` is false at `SQLAppStatusListenerSuite.scala:1123`. ScalaTest reports 12,851 tests passed and 1 failed. That failure remains unresolved, and I have not established whether it is related to this PR.

andygrove:
Merged. Thanks @sunchao

PR #5627, opened 2026-09-01
fix: scope Celeborn bootstrap hooks to Comet clients

### LINKED ISSUE #5529: Native Celeborn shuffle: the installed Celeborn bootstrap hook can break client creation for the whole executor
Following up on PR #5513 (part of #5352). `CelebornTransportCallbackTracker.beginPush` appends a JDK dynamic proxy implementing `TransportClientBootstrap` to the shared `TransportClientFactory.clientBootstraps` list, so that every transport client created afterwards is routed through `FactoryHook.invoke` and into `installClient`. I think that gives a Comet-specific incompatibility the power to break all Celeborn shuffle on the executor, not just Comet's.

Celeborn treats any bootstrap failure as fatal to client creation. From `TransportClientFactory.createClient` in v0.6.0:

```java
for (TransportClientBootstrap clientBootstrap : clientBootstraps) {
  clientBootstrap.doBootstrap(client);
}
} catch (Exception e) { // catch non-RuntimeExceptions too as bootstrap may be written in Scala
  ...
  client.close();
  throw Throwables.propagate(e);
}
```

`installClient` and `installChannel` throw `IOException` in five places, plus `ReflectiveOperationException` from the `field` helper. Since `TransportClientBootstrap` declares only `void doBootstrap(TransportClient client) throws RuntimeException`, a JDK proxy wraps those checked exceptions in `UndeclaredThrowableException`, which `createClient` then propagates. So a single unexpected shape in Celeborn's internals stops every new Celeborn connection on that executor, including connections used by ordinary non-Comet Celeborn shuffle writers.

The hook is also never uninstalled. `CelebornShufflePusherFactory.releaseClient` removes the admission registry entry but restores none of the patched fields, so the proxy stays in `clientBootstraps` for the lifetime of the transport factory. That means the exposure outlives Comet's tasks entirely. An executor that ran one Comet map task early on still has the hook installed hours later when only stock Celeborn shuffle is running.

Could `installClient` swallow its own failures and fall back to the `getPushState` counting path rather than letting them escape `doBootstrap`? Comet would lose the precise ownership signal for clients it could not instrument, which seems much better than failing connection creation for everyone. And could the hook be removed from `clientBootstraps` when the last Comet client is released, so the blast radius does not persist past the tasks that needed it?

pingzh:

Closes #5529.

Follow-up to #5352.

## Rationale for this change

Comet installs a bootstrap hook in Celeborn's shared client factory to track transport ownership. Failures while installing or invoking that hook could escape into Celeborn client creation and disrupt ordinary, non-Comet Celeborn traffic. The hook also remained installed after the final Comet client released it.

- Treat reflective, I/O, and runtime hook instrumentation failures as non-fatal, disable precise transport tracking, and fall back to Celeborn `PushState` accounting.
- Reference-count the shared hook and remove the exact proxy when the final Comet client releases it, while preserving unrelated Celeborn bootstraps and their order.
- Synchronize hook release with active bootstrap calls, keep stale bootstrap snapshots inert, and restore tracking when a later Comet client reacquires the shared factory.
- Preserve already-observed transport ownership during fallback and serialize raw-push state binding so executor admission is never released early.

- Added regression coverage for bootstrap instrumentation failures, ordinary Celeborn client creation, shared hook ownership, final-owner removal, stale snapshots, reacquisition, fatal bootstrap errors, and concurrent hook release.
- Added fallback accounting coverage for recreated push states, concurrent submissions, cleanup races, retries, callbacks, and pending Netty writes.
- Ran `CelebornShufflePartitionPusherSuite` with the default Spark 4.1 profile: 53 tests passed.
- Ran Spotless and the Scala 2.12 parser check successfully.

### DIFF
--- a/spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
+++ b/spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
@@ -66,6 +66,7 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
   private final Field cryptoHandler;
   private final ExecutorShufflePushAdmission admission;
   private final CelebornTransportCallbackTracker transportCallbacks;
+  private final Object submissionLock = new Object();
   private final int shuffleId;
   private final int mapId;
   private final int encodedAttemptId;
@@ -103,11 +104,13 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
     private final int bytes;
     private ObservedPushState pushState;
     private CelebornTransportCallbackTracker.Push transportPush;
+    private Object clientPushState;
     private boolean claimed;
     private boolean submitted;
     private boolean nativeReleased;
     private boolean transportComplete;
     private boolean released;
+    private boolean pushStateFallback;
 
     private PushReservation(int bytes, boolean nativeOwned) {
       this.bytes = bytes;
@@ -118,9 +121,8 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
   private static final class ObservedPushState {
     private final LongAdder inFlightRequests;
     private final AtomicReference<?> exception;
-    private long submittedPushes;
-    private long releasedPushes;
-    private boolean terminalFailureCredited;
+    private long fallbackSubmittedPushes;
+    private long fallbackReleasedPushes;
 
     private ObservedPushState(LongAdder inFlightRequests, AtomicReference<?> exception) {
       this.inFlightRequests = inFlightRequests;
@@ -494,14 +496,34 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
     requireUnencryptedClient();
     validateFrame(partitionId, data, length);
     beginPush();
+    PushReservation reservation;
+    try {
+      // Admission can wait for an older request's completion. Never hold the submission lock
+      // while waiting; callbacks and retries must remain able to make progress.
+      reservation = claimEncodingReservation(length);
+    } catch (IOException | RuntimeException | Error failure) {
+      abortAndSuppress(failure);
+      try {
+        endPush();
+      } catch (IOException cleanupFailure) {
+        if (cleanupFailure != failure) {
+          failure.addSuppressed(cleanupFailure);
+        }
+      }
+      throw failure;
+    }
+    synchronized (submissionLock) {
+      pushClaimedPartitionData(partitionId, data, length, reservation);
+    }
+  }
 
-    PushReservation reservation = null;
+  private void pushClaimedPartitionData(
+      int partitionId, byte[] data, int length, PushReservation reservation) throws IOException {
     boolean registered = false;
     boolean submitted = false;
     boolean insideClient = false;
     Throwable failure = null;
     try {
-      reservation = claimEncodingReservation(length);
       beginClientPush();
       insideClient = true;
 
@@ -518,7 +540,7 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
           throw new IOException("Celeborn returned a null push state for this map attempt");
         }
         observed = observePushState(pushState);
-        registerPendingPush(reservation, observed);
+        registerPendingPush(reservation, pushState, observed);
         registered = true;
       }
 
@@ -549,10 +571,11 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
         if (clientPushStates != null) {
           Object current = ((Map<?, ?>) clientPushStates.get(shuffleClient)).get(mapKey());
           if (current != null && current != pushState) {
+            pushState = current;
             observed = observePushState(current);
           }
         }
-        markSubmitted(reservation, observed);
+        markSubmitted(reservation, pushState, observed);
       }
 
       int minimumAccepted = length + CELEBORN_BATCH_HEADER_BYTES;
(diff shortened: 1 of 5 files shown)


ajsquared (commented):
Reviewed 36f418e61672243ab90651f462ea7b34f7e44009. One non-blocking P2 accounting issue below; no P1 findings. Static review only.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
```diff
@@ -763,55 +814,54 @@ private void reconcileAcceptedPushes() {
...
-        while (observed.releasedPushes < completions) {
-          PushReservation reservation = smallestSubmittedReservation(observed);
-          if (reservation == null) {
+        boolean terminalFailure =
+            failure instanceof IOException
+                && !cleanedUp
+                && isTerminalRawPushFailure((IOException) failure);
+        long counterCredits = terminalFailure ? 1L : 0L;
+        long retainedRequests = Math.max(0L, observed.inFlightRequests.sum() - counterCredits);
+        long completions = Math.max(0L, observed.fallbackSubmittedPushes - retainedRequests);
```
ajsquared, line 878: [P2][non-blocking] Recover admission after a cancelled fallback request fails

If bootstrap instrumentation fails, a subsequent map abort followed by transport failure leaves this fallback reservation charged permanently. In stock Celeborn 0.6.3, `cleanup()` sets `exception` to `Cleaned Up` without resetting `totalInflightReqs`, and the raw-push failure callback returns immediately when that exception is set, without removing the batch ([callback](https://github.com/apache/celeborn/blob/v0.6.3/client/src/main/java/org/apache/celeborn/client/ShuffleClientImpl.java#L1228-L1234), [cleanup](https://github.com/apache/celeborn/blob/v0.6.3/client/src/main/java/org/apache/celeborn/client/ShuffleClientImpl.java#L1813-L1819)). With one pending push, both `fallbackSubmittedPushes` and `retainedRequests` remain 1, so this loop never releases its permits, even after all retained transport ownership completes. Enough such reservations can stall later maps sharing the client. Please retain a safe terminal-completion path for cancelled fallback pushes; the existing `failed transport after cleanup releases only its completed shared admission` case could cover this with bootstrap tracking disabled.

ajsquared (approved):
Reviewed 36f418e61672243ab90651f462ea7b34f7e44009. The previously posted P2 feedback is non-blocking.

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
ziting-openai: Confirmed this P2 with an isolated probe against unchanged head `36f418e61672243ab90651f462ea7b34f7e44009`. The client fixture follows the released Celeborn 0.6.0/0.7.0 cleanup and failed-callback behavior. After abort and terminal callback completion, the payload was gone but the counter stayed at 1, all 64 test bytes remained reserved, and the next map reached the real `ExecutorShufflePushAdmission.acquire` wait. One hundred reconciliation calls left the same state. Success and abort-then-success controls released admission. The parent revision rejected the instrumentation mismatch before publishing a raw request.

The existing `failed transport after cleanup releases only its completed shared admission` regression should cover this with bootstrap tracking disabled. A safe terminal-completion signal is still needed in that mode. Crediting `Cleaned Up` alone would release bytes before transport retirement. This probe used a fixture and a no-op logging shim on JDK 21, not a full Celeborn/Netty integration run.

_[ :robot: posted by Codex on behalf of ziting-openai using the spark-pr-review-memo skill :robot: ]_

ziting-openai (approved):
Reviewed `36f418e61672243ab90651f462ea7b34f7e44009` using `spark-pr-review-memo`. No P1 findings. One existing significant P2 remains: [fallback admission after cancellation and failed transport](https://github.com/apache/datafusion-comet/pull/5627#discussion_r3908135122). I added isolated reproduction evidence to that thread. This approval follows the requested zero-P1/fewer-than-three-P2 threshold.

Validation included unchanged current-head/parent Java probes with a Celeborn behavior fixture and logging shim, plus public JVM descriptor comparison and pinned Celeborn source review. I did not run the full Comet suite. Exact-head CI has one macOS scan JVM SIGSEGV whose connection to this patch is unestablished.

_[ :robot: posted by Codex on behalf of ziting-openai using the spark-pr-review-memo skill :robot: ]_

spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
ziting-openai: Re-reviewed `9a75f9de3e37136968694a74a5d6ad2623780385`. The isolated probe now passes the single-request case and cancellation sealed before a later downgrade. This P2 still occurs when **two requests are already precisely tracked before a global bootstrap failure downgrades both**. `awaitPushStateFallbackSlot()` cannot serialize those already submitted requests. The first failed stock callback sets the terminal exception, the second returns on that exception, and the counter remains 2. With two fallback submissions and only one counter credit, only one reservation is released.

I reproduced this against the unchanged new-head classes with the Celeborn behavior fixture: after both callbacks completed, only 64 of 128 test bytes were available, cleanup for that map remained deferred, and the next valid 112-byte reservation blocked in the actual admission wait. Please cover the two-outstanding-request downgrade sequence and preserve a terminal ownership signal for each request, or safely seal/cancel eligible precise requests before the downgrade. This was an isolated JDK 21 probe with transport fixtures and a logging shim, not a full Celeborn/Netty run.

_[ :robot: posted by Codex on behalf of ziting-openai using the spark-pr-review-memo skill :robot: ]_

ziting-openai (approved):
Re-reviewed `9a75f9de3e37136968694a74a5d6ad2623780385` using `spark-pr-review-memo`. No P1 findings. The single-request fallback cancellation and cancellation-before-downgrade cases are fixed. One existing significant P2 remains when several already submitted precise requests are downgraded together: [current-head reproduction](https://github.com/apache/datafusion-comet/pull/5627#discussion_r3908927899). This approval follows the requested zero-P1/fewer-than-three-P2 threshold.

Validation: isolated unchanged-source Java probes compared this head with the prior reviewed head, including positive controls and the remaining multiple-request leak. Public partition-pusher JVM descriptors are unchanged. The probe used transport fixtures and a logging shim on JDK 21; I did not run the full Comet/Celeborn/Netty suite. Current-head CI is still in progress with no reported failures at preflight, and the released Celeborn 0.6.0/0.7.0 reflection checks passed.

_[ :robot: posted by Codex on behalf of ziting-openai using the spark-pr-review-memo skill :robot: ]_

sunchao:
cc @andygrove

sunchao (approved):
Reviewed `9a75f9de3e37136968694a74a5d6ad2623780385` against `1e10eedd6e0303adcac4573f44d5c07ffb0fbacd`. No verified P1/P2 findings. This was a source review, including the existing fallback-accounting discussion. I did not run tests or benchmarks.

Could you add a focused BASE/HEAD native-Celeborn microbenchmark for many partitions and constrained or slow workers, comparing throughput, enqueue/blocked time and retained bytes? Matched frame sizes, admission budgets and correctness checks would help quantify the new submission serialization and fallback concurrency limit.

andygrove (approved):
LGTM, no impact on non Celeborn users

sunchao:
Thanks! merged

ziting-openai:

There are no open review threads owned by this agent here. Apologies for the confusion.

PR #5628, opened 2026-09-01
fix: make native shuffle spill metrics independent of input batching

### LINKED ISSUE #5617: Native shuffle memory_spilled_bytes charges a shared Arrow allocation once per input batch
## Describe the bug

#5370 gave the native shuffle writer a `memory_spilled_bytes` metric, which feeds Spark's `memoryBytesSpilled` task metric through the completion listener in `CometNativeShuffleWriter`. It dedups Arrow zero-copy slices so that one backing allocation is charged once rather than once per slice.

That dedup only covers slices produced **within** a single `insert_batch` call. `spill_accounted_input_buffers` is cleared at both the entry and the exit of `insert_batch` (`native/shuffle/src/partitioners/multi_partition.rs:606` and `:624`), so the `repeated_spill_buffer_bytes` subtraction at `:585` never applies across calls. `pinned_buffers` cannot compensate because `spill()` clears it at `:587`.

The uncovered case is the one the doc comment on `count_new_buffers` names as motivating (`:141`):

> Cheaper measures do not match resident memory for the batches this writer sees. A partial `HashAggregate` emits one group-values buffer sliced into batch_size chunks, and every buffered chunk shares that one allocation

Those chunks reach the writer as separate `RecordBatch`es, so each is its own `insert_batch` call. `pinned_buffers` exists precisely to dedup them for the reservation. The spill metric does not.

## To Reproduce

Against `2949fd0d2`, driving the identical 16 chunks of one 128,168-byte `Int64` allocation through `MultiPartitionShuffleRepartitioner` with `max_buffer_bytes = 8 KiB` (16 spills either way), varying only how the chunks are delivered:

| Delivery                          | `memory_spilled_bytes` | Ratio to the real allocation |
| --------------------------------- | ---------------------- | ---------------------------- |
| 16 separate `insert_batch` calls  | 2,228,224              | 17.0x                        |
| one `insert_batch`, sliced 16 ways | 262,144                | 2.0x                         |

Same allocation, same resident footprint, same spill count, 8.5x apart.

## Expected behavior

Both shapes should report the same figure, since they describe the same resident memory. The inflated number reaches users as `memoryBytesSpilled` in the Stages tab and as `memory spilled bytes` on the Exchange node.

## Additional context

#5370 states the cumulative-per-input-batch behavior is deliberate, and `max_buffer_spills_charge_shared_backing_once_per_input_batch` asserts it. The concern is that the result then depends on whether the producer emitted one batch or sixteen, which is not a distinction Spark's spill metrics carry.

A fix is not just a matter of widening the set's lifetime. Within one `insert_batch` the outer `batch` argument keeps the allocation alive, which is what makes the address-keyed set sound there. Across calls nothing holds it, so a retained address could collide with a recycled allocation. Keying on something that pins the buffer, or scoping the set to the buffers reachable from `buffered_batches`, would both need thought about the cost.

Related: #3996 (native shuffle observability), #5212 (memory pool and accounting audit sweep).

sunchao:

Closes #5617.

## Rationale for this change

Native shuffle deduplicates spilled Arrow allocations only within one outer `insert_batch` call. A producer such as partial HashAggregate can deliver the same zero-copy chunks as separate input batches. For the issue's 128 KiB allocation and sixteen spills, this reports 2,228,224 bytes for separately delivered chunks but only 262,144 bytes when one input is sliced internally. The spill boundaries and pinned allocations are identical, so caller batching should not change the metric.

Use cumulative per-spill input accounting: each spill counts the capacity of its buffered input allocations, deduplicated within that spill, plus its partition-index allocations. Buffering the same allocation for a later spill contributes again. Thus both delivery shapes report 2,228,224 bytes. This is cumulative memory released from shuffle buffering, not peak memory, globally unique task allocations, or necessarily a drop in process RSS when another operator still owns the buffer.

This also preserves the metric's input-memory meaning. Counting materialized output instead can turn a unique 1 MiB input allocation containing 1,024 live Int64 rows into only 8 KiB of reported data, or count shared dictionary/view payloads repeatedly within one spill. Four new regression tests reproduced these mismatches before the correction and pass with per-spill input accounting.

Remove the input-batch-scoped spilled-address set and repeated-buffer discount. Continue using the existing reservation, which deduplicates input allocations while they are buffered, and include the just-buffered growth if its reservation was rejected. After the spill producer drops its inputs and indices, publish the released reservation plus any rejected growth. Failed writes also count all input memory released during cleanup, regardless of how much output the writer consumed.

No payloads or allocation addresses are retained across spills for metrics. Physical memory reservations, spill triggers, shuffle output, `data_size`, and disk-byte accounting remain unchanged. The user-facing metric documentation now explains the cumulative per-spill contract.

The batching regression exercises maximum-buffer and rejected-reservation spills, verifies equal metrics and spill counts for internal/external slicing, and compares serialized shuffle output. Independently allocated small chunks correctly report less pinned memory while producing identical output. Repeated spills remain cumulative.

Additional native regressions cover oversized input allocations, aliased columns, shared string-view and dictionary payloads, heterogeneous nullable nested inputs, and cleanup on early or partially consumed writer errors. A mixed-reservation test covers a successful reservation followed by rejected growth, with both shared and independent input buffers and successful/failed writers.

Local validation on Linux with JDK 17:

- `cargo test --locked -p datafusion-comet-shuffle --lib -- --test-threads=1`: **99 passed**, none failed or ignored.
- `cargo clippy --locked -p datafusion-comet-shuffle --all-targets -- -D warnings`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.
- `make core`: passed before JVM validation.
- Root-reactor Spark **4.0.4** validation with `./mvnw -B -Pspark-4.0 test spotless:check -Dtest=none -Dsuites=org.apache.spark.sql.comet.CometTaskMetricsSuite`: **all 14 tests passed in 1 suite**, full reactor **BUILD SUCCESS**, including Spotless.

Native builds used eight jobs with dev/test debug symbols disabled. Spark ran with `SPARK_LOCAL_IP=127.0.0.1` and `SPARK_LOCAL_HOSTNAME=localhost`. Other Spark versions, live Celeborn, and benchmarks were not rerun locally.

### DIFF
--- a/docs/source/user-guide/latest/metrics.md
+++ b/docs/source/user-guide/latest/metrics.md
@@ -50,6 +50,14 @@ enabled and uncompressed when `spark.shuffle.compress=false`. Memory spill bytes
 partition-index data rather than their on-disk size. These values also appear in Spark's task
 metrics and Spark UI as `diskBytesSpilled` and `memoryBytesSpilled`, respectively.
 
+Memory spill bytes are cumulative across spills, not a peak-memory measurement or a count of
+allocations unique across the whole task. Each spill counts the full capacity of its buffered
+input allocations, deduplicating buffers shared by columns or batches in that spill, plus its
+partition-index allocations. If a later spill buffers the same backing allocation again, it
+contributes again. Whether input slices arrive in one batch or separate batches does not change
+the accounting for identical spill boundaries. Other operators may still own the same buffers,
+so this measures memory released from shuffle buffering, not necessarily a drop in process memory.
+
 ## Native Metrics
 
 Setting `spark.comet.explain.native.enabled=true` will cause native plans to be logged in each executor. Metrics are
--- a/native/shuffle/src/metrics.rs
+++ b/native/shuffle/src/metrics.rs
@@ -46,7 +46,9 @@ pub(crate) struct ShufflePartitionerMetrics {
     /// total spilled bytes during the execution of the operator
     pub(crate) spilled_bytes: Count,
 
-    /// Total in-memory bytes released by spills before compression.
+    /// Total buffer size of materialized spill batches before compression, plus the
+    /// partition-index allocations released by spills. Measured from spill output rather
+    /// than input batch boundaries; not a count of globally unique input allocations.
     pub(crate) memory_spilled_bytes: Count,
 
     /// The original size of spilled data. Different to `spilled_bytes` because of compression.
--- a/native/shuffle/src/metrics.rs
+++ b/native/shuffle/src/metrics.rs
@@ -46,9 +46,8 @@ pub(crate) struct ShufflePartitionerMetrics {
     /// total spilled bytes during the execution of the operator
     pub(crate) spilled_bytes: Count,
 
-    /// Total buffer size of materialized spill batches before compression, plus the
-    /// partition-index allocations released by spills. Measured from spill output rather
-    /// than input batch boundaries; not a count of globally unique input allocations.
+    /// Cumulative input backing-buffer and partition-index capacity released by spills.
+    /// Shared input allocations are counted once per spill, not once per input batch.
     pub(crate) memory_spilled_bytes: Count,
 
     /// The original size of spilled data. Different to `spilled_bytes` because of compression.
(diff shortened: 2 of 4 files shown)


comphead (approved):
Thanks @sunchao I checked it manually and automatically, looks strong enough

sunchao:
Thanks @comphead ! merged

PR #5630, opened 2026-09-02
fix: rebase map offsets in mapsort so sliced maps do not overrun entries

### LINKED ISSUE #5629: mapsort on a sliced map fails with "Max offset exceeds length of entries"
### Describe the bug

`mapsort` on a sliced `MapArray` fails the query instead of sorting it:

```
org.apache.comet.CometNativeException: Invalid argument error: Max offset of 40 exceeds length of entries 30
	at org.apache.comet.CometExecIterator.$anonfun$getNextBatch$2(CometExecIterator.scala:205)
	...
```

A native OFFSET slices the batch (DataFusion's limit uses `batch.slice(skip, ...)`),
and Arrow keeps a sliced `MapArray`'s original entry offsets. So `spark_map_sort`
receives a map whose first entry offset is nonzero.

`native/spark-expr/src/map_funcs/map_sort.rs` builds its `take` indices only for the
visible maps, so the sorted entries are indexed from zero, but it then passes the
*input* offsets to `MapArray::try_new`:

```rust
let sorted_entries = take(maps_arg_entries, &indices, None)?;
...
let sorted_map_arr = Arc::new(MapArray::try_new(
    Arc::clone(map_field),
    maps_arg.offsets().clone(),   // original offsets, not rebased
    sorted_map_struct.clone(),
    maps_arg.nulls().cloned(),
    *is_sorted,
)?);
```

For two two-entry maps, slicing away the first leaves offsets `[2, 4]` while `take`
produced only 2 entries, and Arrow rejects the result.

Spark 4.0+ inserts `MapSort` automatically, so no explicit `mapsort` call is needed
to reach this — `InsertMapSortInGroupingExpressions` (group-by on a map key) and
`InsertMapSortInRepartitionExpressions` (repartition on a map key) both do it.

### Steps to reproduce

Spark 4.0 or later:

```scala
withParquetTable(
  (0 until 20).map(i => (i, Map(s"b${i % 5}" -> i, s"a${i % 5}" -> (i + 1)))),
  "tbl") {
  // OFFSET below a GROUP BY on the map column
  sql("SELECT _2, count(*) FROM (SELECT * FROM tbl ORDER BY _1 LIMIT 15 OFFSET 5) GROUP BY _2")
    .collect()
}
```

The plan puts the aggregate's `_groupingmapsort` key above a native offset:

```
HashAggregate(keys=[_groupingmapsort#3400], functions=[count(1)])
+- HashAggregate(keys=[_groupingmapsort#3400], functions=[partial_count(1)])
   +- CometTakeOrderedAndProjectExec(limit=20, offset=5, orderBy=[_1 ASC NULLS FIRST])
      +- CometNativeScan parquet [_1,_2]
```

A unit-level reproduction, independent of Spark:

```rust
let full = /* two maps of two entries each */;
let sliced = full.slice(1, 1);
assert_eq!(sliced.offsets().first().copied(), Some(2));
spark_map_sort(&[ColumnarValue::Array(Arc::new(sliced))]).unwrap();
// Arrow error: Invalid argument error: Max offset of 4 exceeds length of entries 2
```

### Expected behavior

The query should sort the map entries and return the same answer as Spark, rather
than failing. `spark_map_sort` should rebase the output offsets to match the entries
it actually took.

### Additional context

Found while reviewing #5567, which admits nested types as native shuffle hash
partitioning keys and would add `InsertMapSortInRepartitionExpressions` as a second
route to this code. The bug is not caused by that change: the group-by path above
reproduces on current `main` with no configuration changes. Credit to @sunchao for
spotting it in review.

viirya:

Closes #5629.

## Rationale for this change

`spark_map_sort` builds its `take` indices only for the maps that are visible, so
the sorted entries it produces are indexed from zero. It then passed the *input*
offsets straight to `MapArray::try_new`:

```rust
let sorted_entries = take(maps_arg_entries, &indices, None)?;
...
MapArray::try_new(
    Arc::clone(map_field),
    maps_arg.offsets().clone(),   // original offsets, not rebased
    sorted_map_struct.clone(),
    ...
)
```

A sliced `MapArray` keeps its original entry offsets, so for two two-entry maps
sliced to drop the first, the offsets are `[2, 4]` against 2 taken entries and Arrow
rejects the result with `Max offset of 4 exceeds length of entries 2`.

A native OFFSET is enough to produce that slice (DataFusion's limit uses
`batch.slice(skip, ...)`), and Spark 4.0+ inserts `MapSort` on its own, so this fails
a real query today — a group-by on a map column below an OFFSET, with no explicit
`mapsort` call and no configuration changes.

- `map_sort.rs`: rebuild the output offsets from the per-map lengths rather than
  reusing the input offsets. Empty maps still contribute an offset entry, so a map
  array containing empty maps keeps the right offset count (the previous `continue`
  skipped the whole iteration).

- New Rust unit test `test_sliced_map_offsets_are_rebased`: slices a two-map array to
  drop the first map, asserts the slice really does keep the original offsets, and
  checks the sorted keys and values. `cargo test -p datafusion-comet-spark-expr
  map_sort` passes 9 tests.
- New end-to-end test in `CometMapExpressionSuite`: a group-by on a map column below
  `LIMIT ... OFFSET ...`, compared against Spark. `CometMapExpressionSuite` passes 23
  tests.

I checked that both tests actually guard the fix by reverting the offset rebase and
rebuilding: the Rust test and the end-to-end test both fail, with the Arrow error
above.

## Additional context

Found by @sunchao while reviewing #5567, which admits nested types as native shuffle
hash partitioning keys and would add `InsertMapSortInRepartitionExpressions` as a
second route into this code. This fix is independent of that PR — the group-by path
reproduces on current `main` — so it is split out here. Once this merges I will add
the repartition-on-map coverage to #5567, where that path becomes reachable.

### DIFF
--- a/native/spark-expr/src/map_funcs/map_sort.rs
+++ b/native/spark-expr/src/map_funcs/map_sort.rs
@@ -16,6 +16,7 @@
 // under the License.
 
 use arrow::array::{Array, ArrayRef, MapArray, StructArray, UInt32Array};
+use arrow::buffer::OffsetBuffer;
 use arrow::compute::{sort_to_indices, take, SortOptions};
 use arrow::datatypes::DataType;
 use datafusion::common::{exec_err, DataFusionError};
@@ -61,20 +62,26 @@ pub fn spark_map_sort(args: &[ColumnarValue]) -> Result<ColumnarValue, DataFusio
 
     // Build one global permutation over the full entries struct, respecting per-map boundaries,
     // then issue a single `take`. This avoids per-map struct copies and a final `concat`.
+    //
+    // `take` produces exactly the entries the visible maps refer to, so the result is indexed from
+    // zero. A sliced MapArray keeps its original entry offsets (a slice of two two-entry maps that
+    // drops the first has offsets `[2, 4]`), so the input offsets cannot be reused here -- they
+    // would overrun the taken entries. Rebuild them from the per-map lengths instead.
     let mut global_indices: Vec<u32> = Vec::with_capacity(maps_arg_entries.len());
+    let mut rebased_offsets: Vec<i32> = Vec::with_capacity(maps_arg.len() + 1);
+    rebased_offsets.push(0);
 
     for idx in 0..maps_arg.len() {
         let map_start = maps_arg_offsets[idx] as usize;
         let map_end = maps_arg_offsets[idx + 1] as usize;
-        if map_end == map_start {
-            continue;
+        if map_end > map_start {
+            let map_keys = maps_arg_entries
+                .column(0)
+                .slice(map_start, map_end - map_start);
+            let local_indices = sort_to_indices(&map_keys, Some(sort_options), None)?;
+            global_indices.extend(local_indices.values().iter().map(|i| map_start as u32 + *i));
         }
-
-        let map_keys = maps_arg_entries
-            .column(0)
-            .slice(map_start, map_end - map_start);
-        let local_indices = sort_to_indices(&map_keys, Some(sort_options), None)?;
-        global_indices.extend(local_indices.values().iter().map(|i| map_start as u32 + *i));
+        rebased_offsets.push(global_indices.len() as i32);
     }
 
     let indices = UInt32Array::from(global_indices);
@@ -87,7 +94,7 @@ pub fn spark_map_sort(args: &[ColumnarValue]) -> Result<ColumnarValue, DataFusio
     // Preserve the original is_sorted flag to keep schema consistent
     let sorted_map_arr = Arc::new(MapArray::try_new(
         Arc::clone(map_field),
-        maps_arg.offsets().clone(),
+        OffsetBuffer::new(rebased_offsets.into()),
         sorted_map_struct.clone(),
         maps_arg.nulls().cloned(),
         *is_sorted,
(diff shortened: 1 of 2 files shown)


sunchao (approved):
Reviewed `8eddd6d6a1172a09c112f348557342cd088171aa` against the reported base `8729f6e6adf7091e18a48670e790d4ba8fd41e51`. No actionable P1/P2 findings. Rebuilding one output boundary per visible map fixes the sliced-offset overrun while preserving empty maps, validity, and key/value pairing.

A focused Arrow 58.4.0 probe using the unchanged production function bodies passed 3 tests, including all 36 slice windows of mixed nonempty/empty/null maps and nullable values. The base function reproduces the `[2, 4]` overrun; this head returns the correctly rebased result. This is array-level validation with substituted DataFusion wrappers, not a full Comet or Spark execution. I did not run the supplied Rust/Scala suites or benchmarks.

Both [Linux](https://github.com/apache/datafusion-comet/actions/runs/33602823139/job/100160807193) and [macOS](https://github.com/apache/datafusion-comet/actions/runs/33602823139/job/100160807222) lint failures are `cargo fmt --check` differences in the new Rust test. Please format those before merging; other CI checks are still running. The Scala regression would also benefit from native OFFSET/MapSort plan assertions, since `checkSparkAnswer` alone permits fallback.

sunchao (approved):
Thanks for the formatting update. Checked `983f6fd5`: the production code is unchanged, both lint jobs now pass, and I found no new issue in the base-integration check. I reused the earlier Arrow-only probe evidence without rerunning tests. No Spark/JNI execution is claimed; the remaining CI is still pending.

viirya:
Thanks @sunchao !

PR #5634, opened 2026-09-02
feat: enable Comet's in-memory cache by default [WIP]

andygrove:

Part of #5487.

## Rationale for this change

`spark.comet.exec.inMemoryCache.enabled` has been off by default since #5051, so the native cache path only ever runs under `CometInMemoryCacheSuite` and `CometInMemoryCacheKryoSuite`, which exercise it deliberately. Nothing tells us how it behaves under the rest of the suite: the Spark SQL test diffs, the fuzz suites, the Iceberg and Delta jobs, and any test that calls `cache()`/`persist()` incidentally.

This PR flips the default so a full CI run exercises the cache format everywhere caching happens. It is opened as a **draft to collect that signal**, not as a proposal to ship the feature on by default. A clean run is evidence the format is ready for that conversation; a red one is the list of things to fix.

**This is stacked on #5543 and contains its commits.** Review #5543 first. GitHub will not let a PR against `apache/main` use a fork branch as its base, so the branch is opened against `main` to get a real CI run. Only the final commit is new here:

- `spark.comet.exec.inMemoryCache.enabled` defaults to `true`.
- `CometDriverPlugin.maybeSetCacheSerializer` read the config out of `SparkConf` with a hardcoded `false` fallback, so flipping the `ConfigEntry` alone would have left the cache serializer uninstalled for anyone who did not set the key explicitly. It now falls back to the entry's own default, the same way the plugin already reads `spark.comet.metrics.enabled`.
- The in-memory cache user guide records the new default and shows how to turn the feature off.

The feature is still described as experimental.

The point of the PR is the CI run itself. Every job now builds cached tables in Comet's Arrow format wherever a test caches anything, rather than only in the two suites that opt in.

The existing cache suites are unaffected: they set the config explicitly, including the two cases that set it to `false`. `CometInMemoryCacheSuite`'s driver-plugin test passes an explicit `true`, so it still covers the install path rather than relying on the new default.


spark/src/main/scala/org/apache/spark/Plugins.scala
```diff
@@ -104,7 +104,9 @@ object CometDriverPlugin extends Logging {
   private[apache] def maybeSetCacheSerializer(
       conf: SparkConf,
       extraConfs: ju.HashMap[String, String]): Unit = {
-    if (conf.getBoolean(CometConf.COMET_EXEC_IN_MEMORY_CACHE_ENABLED.key, false)) {
+    if (conf.getBoolean(
+        CometConf.COMET_EXEC_IN_MEMORY_CACHE_ENABLED.key,
+        CometConf.COMET_EXEC_IN_MEMORY_CACHE_ENABLED.defaultValue.get)) {
```
mbutrovich: With the default flipped, this installs Comet's serializer for every application that loads `CometPlugin`, including ones that start with `spark.comet.enabled=false` or `spark.comet.exec.enabled=false`. Those applications can never plan `CometInMemoryTableScan`, so every cached read goes through Spark operators on top of Comet's format. That is the 1.5x to 5.2x slower case in the [Limitations table](https://github.com/apache/datafusion-comet/blob/d5983c2d64784a34a5a750c851b1d0a18615251f/docs/source/user-guide/latest/in-memory-cache.md#L176-L190). Keeping the plugin in `spark.plugins` cluster-wide and switching Comet off with `spark.comet.enabled=false` is a common setup, and before this PR it left the cache format alone.

#5485 already lists this check as sound, and only calls it narrow because anyone who opted in would have execution enabled. That premise no longer holds once the feature is on by default. Could the plugin also require both configs at startup? This object already has a [`getBooleanConf`](https://github.com/apache/datafusion-comet/blob/d5983c2d64784a34a5a750c851b1d0a18615251f/spark/src/main/scala/org/apache/spark/Plugins.scala#L200-L201) helper that falls back to the entry's default, so the new read can use it too:

```suggestion
    if (getBooleanConf(conf, CometConf.COMET_ENABLED) &&
      getBooleanConf(conf, CometConf.COMET_EXEC_ENABLED) &&
      getBooleanConf(conf, CometConf.COMET_EXEC_IN_MEMORY_CACHE_ENABLED)) {
```

A session that starts with execution off and turns it on later would then keep Spark's format, and `CometExecRule` already records a fallback reason for that case. That seems like the right trade, since the serializer is fixed for the application.

The [driver-plugin test](https://github.com/apache/datafusion-comet/blob/d5983c2d64784a34a5a750c851b1d0a18615251f/spark/src/test/scala/org/apache/comet/exec/CometInMemoryCacheSuite.scala#L910-L937) sets the key to `true` explicitly, so nothing exercises the fallback to the default, which is the code change in this file. Could you add a case where the key is unset (serializer installed), plus cases for `spark.comet.enabled=false` and `spark.comet.exec.enabled=false` (serializer not installed)?

spark/src/main/scala/org/apache/comet/CometConf.scala
```diff
@@ -278,7 +278,7 @@ object CometConf extends ShimCometConf {
         "SparkContext, otherwise caching fails as soon as a block is serialized, including " +
         "the disk half of the default MEMORY_AND_DISK storage level.")
       .booleanConf
-      .createWithDefault(false)
+      .createWithDefault(true)
```
mbutrovich: #5485 treats the slower Spark-operator reads as acceptable because "the cache path is off by default ... rather than a regression in a shipped path." This PR makes it a shipped path, and the format is fixed when the relation materializes, so a user can't avoid it for one query.

With the plugin gated as suggested above, the remaining exposure is a query where the cached scan runs natively and a Spark operator above it reads through a columnar-to-row transition, and a session that turns Comet off at runtime after caching. Is there a benchmark number for the first case against Spark's own cache format? The published numbers compare against Comet off entirely. For the second case, option 1 in #5485 (a fallback reason when Spark operators read a relation stored in Comet's format) is what would tell a user to turn the feature off. Could that land before or with this PR?

docs/source/user-guide/latest/in-memory-cache.md
```diff
@@ -24,13 +24,13 @@ format that Comet operators read directly. Without it, a cached table is stored
 format and every scan of it has to convert each batch before Comet can continue, which shows up in
 the plan as a `CometSparkColumnarToColumnar` above the cache scan.
 
-This feature is **experimental and disabled by default**. Turn it on at startup, alongside the rest
-of Comet's configuration:
+This feature is **experimental and enabled by default**. To turn it off, set the config at startup,
+alongside the rest of Comet's configuration:
```
mbutrovich: This key did not exist in 1.0.0, so a user upgrading from 1.0.0 goes from Spark's cache format to Comet's without setting anything. With `spark.kryo.registrationRequired=true` and no `CometKryoRegistrator`, a `df.cache()` that spills to disk now fails with "Class is not registered" where it did not before. The plugin only [logs a warning](https://github.com/apache/datafusion-comet/blob/d5983c2d64784a34a5a750c851b1d0a18615251f/spark/src/main/scala/org/apache/spark/Plugins.scala#L134-L152) for that.

The versioning policy counts a new error under the same explicit configuration as a [behavior change](https://github.com/apache/datafusion-comet/blob/d5983c2d64784a34a5a750c851b1d0a18615251f/docs/source/about/versioning_policy.md#L166-L185). Could you add an entry to the [upgrade guide](https://github.com/apache/datafusion-comet/blob/d5983c2d64784a34a5a750c851b1d0a18615251f/docs/source/user-guide/latest/migration-guide.md) under the next release that covers the format change and the Kryo requirement? The policy asks for a `spark.comet.legacy.*` key, but `spark.comet.exec.inMemoryCache.enabled=false` already restores the old behavior, so naming that key in the entry seems enough. If you read the policy differently, it would be good to settle that here, since this is one of the first behavior changes since 1.0.0.

mbutrovich (commented):

andygrove:
I audited the in-memory cache and filed what came out of it.

#6202 is the one that bears most on this PR. With AQE on, once the table-cache stage materializes, the re-plan leaves the operators directly above it on Spark, so a plain `GROUP BY` or join over a cached table reads Comet's format through a `CometColumnarToRow` into Spark operators. That is the case @mbutrovich asks about in the review, and under AQE it is the common case rather than an edge case. The numbers in the cache guide come from a benchmark that turns AQE off, so they don't show it. It happens with the feature off too, but then the Spark operators read Spark's own format.

#6203 is about the signal this PR collects. 23 of the 53 cache test definitions check a cached query with `checkSparkAnswer`, and both sides of that comparison read the same cached bytes, so a write-path or pruning bug passes. With two bugs injected on purpose, 54 of the 57 cache tests still passed. A differential check against uncached ground truth caught both, and on unmodified `main` it matched on every write path, so the format itself looks sound.

I also added a cached-relation repro to #3079. A relation cached while Comet is on keeps native-shuffled partitions on a wide-decimal key, and a later join with Comet off returns 57 of 10000 rows. That doesn't depend on the cache format or on this PR, but caching is what carries it into queries that have Comet off.

andygrove:
Moving this back to draft. We'll ship the in-memory cache in 1.1.0 the way it is on `main`, experimental and off by default, and fix #6202, #6203 and the points from this review before turning it on by default in the next release. #3079 doesn't depend on the cache format and is being fixed separately in #6005.

spark/src/main/scala/org/apache/spark/Plugins.scala
andygrove: Agreed. With the default on, an application that can never plan the native scan shouldn't get Comet's format. I'll gate the install on `spark.comet.enabled` and `spark.comet.exec.enabled` through `getBooleanConf`, and add the three plugin cases, before this comes out of draft. It won't avoid the slow path on its own, though: with both on, AQE still puts Spark operators above the cache scan, which is #6202.

spark/src/main/scala/org/apache/comet/CometConf.scala
andygrove: The audit turned up part of the answer. Under AQE the first case is the normal outcome, not an edge case. Once the table-cache stage materializes, the re-plan leaves the operators above it on Spark, so a plain aggregate or join over a cached table reads Comet's format through a `CometColumnarToRow` (#6202). `CometInMemoryCacheBenchmark` runs with AQE off, so the published numbers don't show it. I'll fix #6202 first and then benchmark with AQE on against Spark's own format, so the number measures the path users will actually get.

Yes to option 1 from #5485 landing with the default flip. It will need to cover the #6202 path too, since nothing records a fallback reason there today.

docs/source/user-guide/latest/in-memory-cache.md
andygrove: Agreed on the upgrade guide entry, covering both the format change and the Kryo requirement. Moving the flip past 1.1.0 changes one premise, though. 1.1.0 ships this key with a default of `false`, so turning it on in the next release is a change to an existing key's default, which is the first case the policy lists. Let's settle the legacy-key question when this comes out of draft.

PR #5640, opened 2026-09-02
test: exercise the Iceberg write split-operator plan in Iceberg's own suites

### LINKED ISSUE #5259: Iceberg write split-operator plan: CI failures when enabled by default
## Describe the bug

[#4658](https://github.com/apache/datafusion-comet/pull/4658) adds Comet's split-operator plan for Iceberg V2 writes (`IcebergWriteExec` + `IcebergCommitExec`) behind `spark.comet.write.iceberg.splitOperator.enabled`, defaulting to `false`. Because the feature ships off, nothing in CI exercises it.

[#5255](https://github.com/apache/datafusion-comet/pull/5255) is a throwaway audit PR: #4658 rebased onto `main` with that default flipped to `true`, so the full matrix runs against the split-operator plan. This issue catalogues what it broke, so the work can be scoped before the feature can be enabled by default.

Audit run: https://github.com/apache/datafusion-comet/actions/runs/30920419488

8 jobs failed, and every failure falls into one of four buckets. All four are caused by the same underlying change: the write is no longer a single `V2TableWriteExec`/`AppendData` node, it is `IcebergCommit -> IcebergWrite -> <child>`.

### Failure summary

| # | Symptom | Failing tests | Jobs |
|---|---------|---------------|------|
| 1 | `Expected at least one captured plan with AppendData but got none` | `CometIcebergRewriteActionSuite`, 4 tests | Comet `[scans]` on Spark 3.4 / 3.5 / 4.0 / 4.1 |
| 2 | `ClassCastException: IcebergCommitExec cannot be cast to V2TableWriteExec` | `TestSystemFunctionPushDownInRowLevelOperations`, 7-12 tests | Iceberg extensions on 1.8 / 1.9 / 1.10 / 1.11 |
| 3 | `Snapshot summary should contain merge metric: spark.merge-into.num-target-rows-copied` | `TestCopyOnWriteMergeMetrics`, 36 parameterizations | Iceberg extensions 1.11 |
| 4 | `Cached table should reflect session schema changes` | `TestCachedTableRefresh.testCachedTableWithSessionSchemaChangeAddColumn`, 2 parameterizations | Iceberg extensions 1.11 |

Note that `PR Build (Linux) / Spark 4.2, JDK 17 [scans]` passing is not a signal: Iceberg is not on the 4.2 classpath, so all the Iceberg suites cancel there.

---

### 1. `CometIcebergRewriteActionSuite` matches on `AppendData` (4 tests, all Spark versions)

- `binPack rewrite reads each file group via CometIcebergNativeScan`
- `sort rewrite runs scan, exchange, and sort natively in Comet`
- `single-column zOrder rewrite runs scan, native exchange, and sort natively in Comet`
- `binPack rewrite applies positional and equality deletes during compaction (MOR)`

```
List() was empty Expected at least one captured plan with AppendData but got none.
--- plan 0 ---
== Physical Plan ==
IcebergCommit (4)
+- IcebergWrite (3)
   +- * CometColumnarToRow (2)
      +- CometIcebergNativeScan (1)
```

Iceberg's `rewrite_data_files` action writes its compacted output through `AppendData`, which the split operator replaces. The rewrite itself is still correct: the row-preservation and file-count assertions in these same tests pass, only the plan-shape filter finds nothing.

The plan filter is at `spark/src/test/scala/org/apache/comet/CometIcebergRewriteActionSuite.scala:207` and `:269`. It should accept either write shape rather than matching `AppendData` alone.

### 2. Iceberg's own tests cast the plan root to `V2TableWriteExec` (1.8 / 1.9 / 1.10 / 1.11)

```
java.lang.ClassCastException: class org.apache.spark.sql.comet.IcebergCommitExec cannot be cast to
class org.apache.spark.sql.execution.datasources.v2.V2TableWriteExec
    at TestSystemFunctionPushDownInRowLevelOperations.executeAndCollectFunctionCalls(...:316)
```

Affects the copy-on-write delete tests on every version (bucket-in-predicate, bucket-eq-predicate, years, months, days, hours, truncate), plus the copy-on-write update and merge variants on 1.9 / 1.10 / 1.11. 7 failures on 1.8, 12 on each of 1.9 / 1.10 / 1.11.

`executeAndCollectFunctionCalls` reaches into the executed plan and casts it to `V2TableWriteExec` to pull out the write's query. With the split operator the root is `IcebergCommitExec`, which is not a `V2TableWriteExec`.

Two possible directions:
- Patch the test in `dev/diffs/iceberg/{1.8.1,1.9.1,1.10.0,1.11.0}.diff` to unwrap `IcebergCommitExec` (none of the four diffs touch this file today).
- Or make the split-operator plan present a `V2TableWriteExec`-compatible root, so third-party code that pattern-matches on the standard Iceberg write shape keeps working. This is the more interesting question for the feature generally: any external tooling that inspects the physical plan of an Iceberg write sees a shape it does not recognize.

### 3. Merge metrics missing from the snapshot summary (1.11, 36 failures)

```
java.lang.AssertionError: [Snapshot summary should contain merge metric: spark.merge-into.num-target-rows-copied]
Expecting actual:
  {"added-data-files"="2", ..., "spark.app.id"="local-...", "total-records"="3"}
to contain key:
  "spark.merge-into.num-target-rows-copied"
    at TestMergeMetrics.assertMergeMetric(TestMergeMetrics.java:299)
```

All six `TestCopyOnWriteMergeMetrics` methods fail across all six parameterizations: `testMergeMetricsWithAllClauses`, `testMergeMetricsWithMatchedDelete`, `testMergeMetricsWithMatchedUpdate`, `testMergeMetricsWithMultipleUpdatesAndDeletes`, `testMergeMetricsWithNotMatchedBySourceDelete`, `testMergeMetricsWithNotMatchedBySourceUpdate`. The merge-on-read equivalents pass.

Unlike buckets 1 and 2 this is not a stale test expectation, it is missing behaviour. Iceberg derives the `spark.merge-into.*` snapshot summary properties from the write's driver-side metrics at commit time. Splitting write from commit appears to lose them: the commit operator does not carry the `MergeRows` metrics forward into the commit's summary properties. Anything downstream that reads snapshot summaries (auditing, incremental processing) would silently see incomplete metadata.

### 4. `TestCachedTableRefresh` (1.11, 2 failures)

```
java.lang.AssertionError: [Cached table should reflect session schema changes]
Expected size: 2 but was: 1 in:
[[1, 100]]
    at TestCachedTableRefresh.testCachedTableWithSessionSchemaChangeAddColumn(...:211)
```

Fails for `testhive`/`SparkCatalog` and `spark_catalog`/`SparkSessionCatalog`. Attribution is a little less direct than for the other three, but the Iceberg 1.11 extensions job passes on other PRs against the same `main` (e.g. https://github.com/apache/datafusion-comet/actions/runs/30915735530), so it is caused by this change rather than pre-existing. Worth a closer look, since a stale cached table after a schema change is a correctness concern, not a test-shape concern.

---

### Follow-on work

1. Relax the `AppendData` plan assertions in `CometIcebergRewriteActionSuite` to accept the split-operator shape (test-only).
2. Decide whether Iceberg's `V2TableWriteExec` cast should be patched in the Iceberg diffs, or whether the split-operator plan should preserve a compatible root node (design question, affects any third-party plan inspection).
3. Propagate merge/write metrics from `IcebergWriteExec` through to the commit so `spark.merge-into.*` snapshot summary properties are still written (behaviour gap).
4. Root-cause `TestCachedTableRefresh.testCachedTableWithSessionSchemaChangeAddColumn`.

Items 1, 3, and 4 look like prerequisites for turning the flag on by default. Item 2 needs a decision before either fix is worth writing.

### Steps to reproduce

Check out https://github.com/apache/datafusion-comet/pull/5255, which is #4658 rebased with `spark.comet.write.iceberg.splitOperator.enabled` defaulted to `true`. Alternatively, set that config to `true` on top of #4658 and run `CometIcebergRewriteActionSuite` plus the Iceberg Spark extensions suites.

### Expected behavior

The Iceberg write split-operator plan passes the existing Comet and Iceberg test suites when enabled, so the flag can eventually default to `true`.

### Additional context

Relates to #4322. Depends on #4658.

andygrove:

Closes #5259.

## Rationale for this change

#5259 catalogued four buckets of CI failures from enabling Comet's Iceberg write
split-operator plan (`IcebergCommit -> IcebergWrite`) by default. It was filed on
2026-08-04, three days before #4658 merged, so it audited an in-review state of that
PR. **All four buckets have since been fixed:**

| Bucket | Where it was fixed |
| --- | --- |
| 1. `CometIcebergRewriteActionSuite` asserts on `AppendData` | #5361 — the plan filter now accepts `AppendData` or `IcebergCommit` (`CometIcebergRewriteActionSuite.scala:282`, `:343-346`) |
| 2. Iceberg's tests cast the plan root to `V2TableWriteExec` | All four Iceberg diffs patch `TestSystemFunctionPushDownInRowLevelOperations.executeAndCollectFunctionCalls` to unwrap |
| 3. `spark.merge-into.*` missing from the snapshot summary | `IcebergWriteSummaryShim` (Spark 4.1+) collects `MergeRowsExec` metrics and passes `MergeSummaryImpl` to `BatchWrite.commit` |
| 4. `TestCachedTableRefresh` schema-change staleness | `IcebergRefreshCacheShim` (Spark 4.1+) recaches by name via `recacheTableOrView` rather than by plan |

What has not been fixed is the coverage gap the issue opens with: *"Because the feature
ships off, nothing in CI exercises it."* Comet's own suites do set the flag
(`CometIcebergWriteActionSuite`, `CometIcebergWriteDetectionSuite`,
`CometIcebergRewriteActionSuite`), which covers bucket 1. But the Iceberg diffs inject
only two Comet configs — `spark.comet.explainFallback.enabled` and
`spark.comet.scan.icebergNative.enabled` — so Iceberg's own suites still run with the
split operator off. That is exactly where buckets 2, 3, and 4 failed, and those three
fixes are currently unverified by CI.

Add `.config("spark.comet.write.iceberg.splitOperator.enabled", "true")` next to the
existing native-scan config in every Comet-configured `SparkSession` builder across the
four Iceberg test diffs — 63 sites each on 1.8.1 / 1.9.1 / 1.10.0 (which patch both
`spark/v3.4` and `spark/v3.5`) and 31 on 1.11.0.

**No user-facing default changes.** `spark.comet.write.iceberg.splitOperator.enabled`
stays `false` in `CometConf`, and remains `CATEGORY_TESTING`. Flipping the default is a
separate decision — worth taking deliberately, since bucket 2 is a live example of
third-party code that pattern-matches on the stock write plan shape and breaks.

Two things a reviewer may want to push back on:

- The config is added to the five `src/jmh` benchmark files too, not just `src/test`,
  matching how `scan.icebergNative.enabled` is already applied everywhere. Those aren't
  run in CI. The argument for including them is consistency and that someone
  benchmarking Comet Iceberg writes wants the plan Comet actually produces; happy to
  restrict to `src/test` if preferred.
- The diffs are **regenerated**, not hand-edited, per the process in
  `docs/source/contributor-guide/iceberg-spark-tests.md`. That normalises a pre-existing
  anomaly and accounts for most of the line churn: the
  `TestSystemFunctionPushDownInRowLevelOperations` hunks (bucket 2's fix) had been
  appended at the end of each diff rather than sorted into git's canonical path order.
  Generating with `--abbrev=10` keeps the `index` lines of unmodified files byte-identical,
  so index churn is limited to the files actually touched.

Also documents the new config in the diff-overview list in
`docs/source/contributor-guide/iceberg-spark-tests.md`.

CI is the test — that is the point of the change. `dev/diffs/iceberg/**` is in the path
filter for all four Iceberg jobs in `dev/ci/compute-changes.py`, so `iceberg_1_11` runs
automatically; **this PR needs the `run-iceberg-tests` label** to also cover 1.8 / 1.9 /
1.10, since the Spark 3.4 and 3.5 diffs are only exercised there.

No Comet source is touched, so Comet's own suites are unaffected.

The regenerated diffs were verified mechanically rather than by inspection. For each
version, the committed diff and the regenerated diff were each applied to a clean
checkout of the matching `apache-iceberg-<version>` tag and the resulting trees compared:

```
=== 1.8.1 ===   59 files changed, 63 insertions(+)
=== 1.9.1 ===   59 files changed, 63 insertions(+)
=== 1.10.0 ===  59 files changed, 63 insertions(+)
=== 1.11.0 ===  31 files changed, 31 insertions(+)
```

Zero removed lines, and every added line is the split-operator config — so the
regeneration is semantically identical to the previous diffs plus this change, despite
the reordering churn in the diff files themselves.

`spotlessApply` was not run against the Iceberg clones (it needs the full Gradle
toolchain). It would be a no-op here: each inserted line is a copy of the adjacent
`.config(...)` line with the same indentation, is at most 78 columns against a 100-column
limit, and all 220 insertions land mid-chain in a fluent builder (verified: none of the
anchor lines terminate with `;`).

### DIFF
--- a/dev/diffs/iceberg/1.10.0.diff
+++ b/dev/diffs/iceberg/1.10.0.diff
@@ -1,5 +1,5 @@
 diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
-index eeabe54f5f..a76d7bdcbc 100644
+index eeabe54f5f..72873ccbc9 100644
 --- a/gradle/libs.versions.toml
 +++ b/gradle/libs.versions.toml
 @@ -37,7 +37,7 @@ awssdk-s3accessgrants = "2.3.0"
@@ -24,10 +24,10 @@ index 714be0831d..dede65f213 100644
      // runtime dependencies for running Hive Catalog based integration test
      integrationRuntimeOnly project(':iceberg-hive-metastore')
 diff --git a/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/ExtensionsTestBase.java b/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/ExtensionsTestBase.java
-index 4c1a509591..6220ae2b98 100644
+index 4c1a509591..186a264a88 100644
 --- a/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/ExtensionsTestBase.java
 +++ b/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/ExtensionsTestBase.java
-@@ -59,7 +59,23 @@ public abstract class ExtensionsTestBase extends CatalogTestBase {
+@@ -59,7 +59,25 @@ public abstract class ExtensionsTestBase extends CatalogTestBase {
              .config("spark.sql.legacy.respectNullabilityInTextDatasetConversion", "true")
              .config(
                  SQLConf.ADAPTIVE_EXECUTION_ENABLED().key(), String.valueOf(RANDOM.nextBoolean()))
@@ -37,6 +37,7 @@ index 4c1a509591..6220ae2b98 100644
 +                "org.apache.spark.sql.comet.execution.shuffle.CometShuffleManager")
 +            .config("spark.comet.explainFallback.enabled", "true")
 +            .config("spark.comet.scan.icebergNative.enabled", "true")
++            .config("spark.comet.write.iceberg.splitOperator.enabled", "true")
 +            .config("spark.memory.offHeap.enabled", "true")
 +            .config("spark.memory.offHeap.size", "10g")
              .enableHiveSupport()
@@ -46,16 +47,17 @@ index 4c1a509591..6220ae2b98 100644
 +                "org.apache.spark.sql.comet.execution.shuffle.CometShuffleManager")
 +            .config("spark.comet.explainFallback.enabled", "true")
 +            .config("spark.comet.scan.icebergNative.enabled", "true")
++            .config("spark.comet.write.iceberg.splitOperator.enabled", "true")
 +            .config("spark.memory.offHeap.enabled", "true")
 +            .config("spark.memory.offHeap.size", "10g")
              .getOrCreate();
  
      TestBase.sparkContext = JavaSparkContext.fromSparkContext(spark.sparkContext());
 diff --git a/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java b/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java
-index 58d054bd05..61b307b92b 100644
+index 58d054bd05..ef1c1cb6d1 100644
 --- a/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java
 +++ b/spark/v3.4/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java
-@@ -56,6 +56,14 @@ public class TestCallStatementParser {
+@@ -56,6 +56,15 @@ public class TestCallStatementParser {
              .master("local[2]")
              .config("spark.sql.extensions", IcebergSparkSessionExtensions.class.getName())
              .config("spark.extra.prop", "value")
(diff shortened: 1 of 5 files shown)


sunchao (approved):
Thanks for closing the coverage gap. I independently applied the old and new patches to all four matching Iceberg tags and confirmed exactly 220 config insertions, with every other resulting source byte and file mode preserved. The split-plan test helper remains intact, and the user-facing default stays false. I found no actionable P1/P2 issue. This was source and mechanical validation, not a local Iceberg suite run. Runtime integration checks are successful in the supplied CI snapshot, while the core and extension suites are still running.

PR #5650, opened 2026-09-02
fix: accept dictionary encodings in remote shuffle

### LINKED ISSUE #5536: Native Celeborn shuffle: validate_remote_schema rejects dictionary shapes the shuffle writer can emit

**Component**: `native/shuffle/src/remote_schema.rs:47`

`validate_remote_schema` unwraps exactly one wire shape, a **top-level**
`Dictionary(Int32, Utf8 | Binary)`. Every other dictionary is a hard mismatch, including
`Dictionary(Int32, Int32)`, `Dictionary(Int8, Utf8)`, and any dictionary nested inside a list, map,
or struct.

The local reader is far more permissive. `unpack_dictionary` (`shuffle_scan.rs:254`) casts _any_
top-level dictionary to its value type, and `cast_and_stamp_schema` absorbs nested ones. So the two
readers disagree about what a valid frame is, and the disagreement is asymmetric: a frame that reads
fine locally becomes, on Celeborn, a `validate_remote_schema` error, which the reader converts into
a `FetchFailedException`, which triggers a map-stage rerun that reproduces the same frame, ending in
job failure at `spark.stage.maxConsecutiveAttempts`.

## This is not reachable today

I instrumented `read_ipc_compressed_impl` to log any decoded wire schema containing a `Dictionary`
at any nesting depth and ran 125 tests across `CometFuzzTestSuite`, `CometShuffleSuite`, and
`CometNativeShuffleSuite`. Zero hits. `SchemaAlignExec` aligns the writer input to the
catalyst-declared schema and `ScanExec::unpack_dictionary_type` strips top-level dictionaries before
that, so nothing dictionary-encoded currently reaches the wire.

## Why it is still worth fixing

`ShuffleBlockWriter` demonstrably encodes nested dictionaries. Both frames below were produced by
`ShuffleBlockWriter::write_batch`, not hand-built:

```
wire type      = List(Dictionary(Int32, Utf8))
spark declares = List(Utf8)
validate       = Err(Execution("Shuffle block type mismatch at column 0:
                 got List(Dictionary(Int32, Utf8)) but expected List(Utf8)"))
local cast ok  = 3 rows

Dictionary(Int32,Int32) vs Int32
               = Err(Execution("Shuffle block type mismatch at column 0:
                 got Dictionary(Int32, Int32) but expected Int32"))
```

Nothing pins the "no dictionaries on the wire" invariant. The day something upstream stops stripping
them, local shuffle keeps working and Celeborn shuffle starts failing jobs, with no compile error
and no failing test in between.

## Suggested fix

Either:

1. Recurse the dictionary unwrap through the nested `List` / `LargeList` / `FixedSizeList` / `Map` /
   `Struct` arms of `same_logical_type`, so the validator accepts whatever the encoder can emit; or
2. Add a test asserting the writer never emits a dictionary shape the reader would reject, so a
   drift in either direction breaks the build.

Option 1 alone still leaves `Dictionary(Int8, Utf8)` and `Dictionary(Int32, Int32)` rejected, so
option 2 is the one that actually closes the gap.

Verified at merge commit `98cd8c967`.

pingzh:

Closes #5536.

Follow-up to #5352, discovered during post-merge review of #5531.

## Rationale for this change

The remote shuffle validator rejects dictionary encodings that `ShuffleBlockWriter` can emit and the local reader can consume, including numeric dictionaries, alternate key widths, and dictionaries inside containers.

The current Spark pipeline normally strips these encodings before writing, but a future change could expose this mismatch and cause repeated fetch failures. Relaxing validation alone is insufficient because the JVM Arrow importer supports fewer dictionary layouts.

- Compare dictionary value types recursively while retaining strict checks for logical types, struct field identity, and container shape.
- Share remote batch decoding between native shuffle scans and JNI, validating buffers and logical types before normalizing dictionary encodings.
- Preserve null values, nested nullability reconciliation, and zero-column batch row counts.
- Add writer-to-reader regression coverage for supported dictionary layouts.

- All 99 shuffle crate tests pass.
- All 7 native shuffle-scan tests pass.
- Writer round trips cover all eight integer dictionary key types, numeric/string/binary values, nested containers, both writer modes, and all four compression codecs.
- Negative tests verify that logical-type and container-shape mismatches remain rejected.
- Clippy passes for both affected crates' libraries and tests with warnings denied.
- `cargo fmt --all -- --check` and `git diff --check` pass.

### DIFF
--- a/native/shuffle/src/remote_schema.rs
+++ b/native/shuffle/src/remote_schema.rs
@@ -16,18 +16,73 @@
 // under the License.
 
 use arrow::array::RecordBatch;
-use arrow::datatypes::DataType;
+use arrow::datatypes::{DataType, Schema};
 use datafusion::common::DataFusionError;
 use datafusion::error::Result;
+use datafusion_comet_common::cast_and_stamp_schema;
+use std::sync::Arc;
+
+/// Decode a remote shuffle batch and reconcile its encoding with Spark's declared logical types.
+/// Validate buffers and logical types before casting, so a corrupt frame cannot be made to look
+/// compatible by a value-changing cast. Dictionary keys are an encoding detail, including inside
+/// containers: decode them here before either native execution or the JVM Arrow importer sees them.
+pub fn decode_remote_shuffle_batch(
+    bytes: &[u8],
+    expected_types: &[DataType],
+) -> Result<RecordBatch> {
+    let batch = crate::read_ipc_compressed_validated(bytes)?;
+    validate_remote_schema(&batch, expected_types)?;
+    if batch
+        .columns()
+        .iter()
+        .zip(expected_types)
+        .all(|(column, expected)| column.data_type() == expected)
+    {
+        return Ok(batch);
+    }
+
+    let fields: Vec<_> = batch
+        .schema()
+        .fields()
+        .iter()
+        .zip(expected_types)
+        .map(|(field, expected)| {
+            // Non-null dictionary keys can reference null values. The declared types do not
+            // constrain top-level nullability; match ShuffleScanExec's nullable output fields.
+            field
+                .as_ref()
+                .clone()
+                .with_data_type(expected.clone())
+                .with_nullable(true)
+        })
+        .collect();
+    let schema = Arc::new(Schema::new_with_metadata(
+        fields,
+        batch.schema().metadata().clone(),
+    ));
+    // Match the local reader's unpack-then-normalize order. Taking the dictionary keys first
+    // removes unused values, which may contain nulls that cannot satisfy the expected nested
+    // nullability even though every referenced value does.
+    let columns = batch
+        .columns()
+        .iter()
+        .map(|column| match column.data_type() {
+            DataType::Dictionary(_, values) => {
+                arrow::compute::cast(column, values).map_err(DataFusionError::from)
+            }
+            _ => Ok(Arc::clone(column)),
+        })
+        .collect::<Result<Vec<_>>>()?;
+    cast_and_stamp_schema("Remote shuffle reader", &schema, columns, batch.num_rows())
+}
 
 /// Check a remotely fetched batch against the logical types declared by Spark before any cast
 /// or JVM Arrow import. Arrow buffer validation alone cannot detect a valid but incorrect IPC
 /// type, and a general cast can silently change values or turn them into nulls.
 ///
-/// Native shuffle writers align their input with Spark's types before partitioning; row writers
-/// build those types directly. Row writers may use top-level Int32-keyed string/binary
-/// dictionaries, but disable dictionaries inside containers. Do not accept other dictionary
-/// layouts just because Arrow can cast them: the JVM importer does not support all of them.
+/// Dictionary encoding does not change the logical value type, regardless of key width or nesting.
+/// Callers must decode accepted dictionaries before JVM import; use [`decode_remote_shuffle_batch`]
+/// to keep buffer validation, logical validation, and normalization in that order.
 ///
 /// Nested nullability and metadata are normalized separately. List element and map entries names
 /// are synthetic (for example, Rust uses `item` while the JVM uses `element`), but struct member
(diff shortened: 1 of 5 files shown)


native/shuffle/src/remote_schema.rs
```diff
@@ -16,18 +16,73 @@
...
+    ));
+    // Match the local reader's unpack-then-normalize order. Taking the dictionary keys first
+    // removes unused values, which may contain nulls that cannot satisfy the expected nested
+    // nullability even though every referenced value does.
+    let columns = batch
+        .columns()
+        .iter()
+        .map(|column| match column.data_type() {
+            DataType::Dictionary(_, values) => {
+                arrow::compute::cast(column, values).map_err(DataFusionError::from)
```
ajsquared: [P2][non-blocking] Unpack nested dictionaries without narrowing the outer keys

For a valid `Dictionary<UInt16, Dictionary<Int8, Int32>>` whose outer keys contain `128` and whose inner dictionary has 129 rows all referencing value `7`, recursive validation accepts the logical `Int32` type. However, `cast(column, values)` targets `Dictionary<Int8, Int32>`: Arrow 58.4's `dictionary_to_dictionary_cast` casts the outer UInt16 keys to Int8 and rejects 128, so this valid shuffle block still becomes a fetch failure. This cast does not unwrap one dictionary layer when the value type is itself a dictionary. Gather the dictionary values using the outer keys before normalization, preserving the inner key width, and cover this mixed-width nested case.

ajsquared (commented):
Remote-only review of beeae27e8e461502f6f419a983b427efdd67e133. One nonblocking dictionary-decoding edge case; no P1 finding. No local tests or CI inspection performed.

native/shuffle/src/remote_schema.rs
```diff
@@ -16,18 +16,73 @@
...
+        fields,
+        batch.schema().metadata().clone(),
+    ));
+    // Match the local reader's unpack-then-normalize order. Taking the dictionary keys first
+    // removes unused values, which may contain nulls that cannot satisfy the expected nested
+    // nullability even though every referenced value does.
+    let columns = batch
+        .columns()
+        .iter()
+        .map(|column| match column.data_type() {
```
ziting-openai: [P2] Decode dictionaries inside containers before narrowing nullability

`same_logical_type` now accepts a `List<Dictionary<Int32, Struct<value: nullable Int32>>>` when Spark expects `List<Struct<value: non-null Int32>>`, but this pre-pass unwraps a dictionary only when it is the top-level column. With keys `[0, 0]` and dictionary values `[{7}, {null}]`, the logical list contains only valid `{7}` values. `cast_and_stamp_schema` then asks Arrow to cast the list, and Arrow 58.4 casts the entire dictionary-values table to the non-null struct before applying `take(keys)`, so the unused `{null}` fails with `Found unmasked nulls for non-nullable StructArray field`. That turns a valid writer-emitted frame into a remote decode/fetch failure. Please recursively compact dictionaries inside list/struct/map containers before nullability reconciliation, and add this nested unused-value regression.

sunchao (approved):
Reviewed 8d7fea6dea8ed1d67ea3bb47ff20cf4b1629c7d2. No actionable introduced or materially worsened P1/P2 findings remain. Independently checked both existing threads: the nested unused-value case now decodes correctly, and the nested-key helper regression passes (IPC itself has only one dictionary encoding per field).

Validation on this head: 103 shuffle component tests and 7 focused native scan tests passed, plus bounded exact-source value/ownership, malformed-input and base/prior-head controls. Scan tests excluded optional HDFS support. No local Java/JNI, Spark/Celeborn service, full-suite or benchmark run. Broader CI is still running; no completed failures were visible in the reviewed snapshot.

sunchao:
There may be a useful follow-up optimization here: **preserve dictionary encoding where downstream consumers support it, and unpack only where necessary.** DataFusion can hash dictionary values once and reuse the hashes by key, Arrow filtering preserves dictionaries, and `CometDictionaryVector` supports lazy lookups for some layouts.

For the native path, this would require preserving dictionary types consistently in physical schemas and checking downstream operator support. The JVM path could retain supported layouts while decoding others. It would be worth benchmarking CPU and peak memory on low-cardinality string columns, with coverage for dictionaries changing between batches, nested types, and null dictionary values.

This could be explored separately from the compatibility fix in this PR.

sunchao:
Merged, thanks @pingzh !

ziting-openai:
Sorry for the noise. My automated review on this PR was posted by a misconfigured AI agent that should have been restricted to OpenAI-owned repositories. Please disregard that review. My inline review thread is already resolved, and this agent will not review non-OpenAI repositories again.

Issue #5649, opened 2026-09-02
[EPIC] Production-quality native Iceberg writes


andygrove:
### What is the problem the feature request solves?

Native Iceberg writes landed in three parts under #4322: the split writer/committer plan (#4658), the eligibility allowlist with fall-back reasons (#5298), and the iceberg-rust data-file writer with JVM-side manifest metrics (#5361). #4322 closed when #5361 merged, and no issue tracks what remains between "merged behind two off-by-default flags" and "production quality". The open follow-ups exist but are not linked to each other, and several gaps had no issue at all.

This epic is that plan. The goal is unchanged from #4322: an ETL job that scans, transforms, and writes Iceberg tables runs natively end to end, and falls back to iceberg-java wherever the native path cannot reproduce its outcome.

**Current state.** `spark.comet.write.iceberg.splitOperator.enabled` and `spark.comet.iceberg.write.enabled` both default to `false`. Nothing in CI runs with either flag on. Two correctness bugs are open on the native path. The user guide (`iceberg-writes.md`) still describes the writer as unimplemented in one section.

### Describe the potential solution

Phased, in priority order. Each item is a linked issue; check it here when it closes.

**Phase 1: correctness blockers**

- [ ] #5636 Bump the iceberg-rust pin past the partition-path escaping fix (a `#` in an S3 partition value commits a location iceberg-java resolves to a different key)
- [ ] #5637 Native writes drop Hadoop GCS configuration for `HadoopFileIO` + `gs://`; fail closed until the translation is complete
- [ ] #5641 Fix the stale "not yet wired in" wording and the divergence list in `iceberg-writes.md`
- [ ] #5719 `revertToSpark` erases the write node, because `CometIcebergWriteExec` and `CometNativeWriteExec` report their own child as `originalPlan` (opt-in: needs `spark.comet.exec.transitionRevert.enabled`)

**Phase 2: failure handling parity with iceberg-java**

- [ ] #5618 Clean up task-attempt data files on task failure
- [ ] #5277 Remove orphaned files when the commit fails
- [ ] #5646 Failure-injection tests that assert both the visible outcome and the storage state

**Phase 3: coverage so common workloads stay native**

- [ ] #5635 Iceberg system functions (`bucket`, `truncate`, `years`/`months`/`days`/`hours`) as native expressions, so the default hash-distribution write is native end to end (PR #5638)
- [ ] #5318 Native `MergeRowsExec`
[...]

Issue #5645, opened 2026-09-02
Adopt a pin-bump policy for the iceberg-rust dependency


andygrove:
### What is the problem the feature request solves?

`native/Cargo.toml` pins `iceberg` and `iceberg-storage-opendal` to a git revision (`3d84c81353b1b23b6e4ae8eea8f8a021cc6927a7`) rather than a release. The pin is what the native Iceberg scan and writer are tested against, but nothing decides when it moves, and #5636 is the first bug caused by it drifting behind an upstream fix (partition-path escaping, apache/iceberg-rust#2875). With the writer now producing files that iceberg-java reads back, a stale pin can turn an upstream bug fix into a Comet correctness bug.

### Describe the potential solution

Write down and follow a policy, for example:

- Bump the pin at least once per Comet release cycle, and immediately when an upstream fix affects file bytes, manifest metadata, or partition layout.
- Prefer tagged iceberg-rust releases over arbitrary revisions once a release contains everything Comet needs, so the version is visible in `Cargo.lock` and in the docs.
- Each bump runs the Iceberg suites on all four Spark/Iceberg profiles plus the manifest-metrics parity tests, since those are what catch behavior changes in the writer.
- Track "waiting on upstream" items (see the eligibility-restrictions issue) so a bump can also lift restrictions.
- Note the bump in the release notes, since users reading Comet-written tables with iceberg-java care which writer semantics they got.

### Additional context

Part of the native Iceberg writes epic, #5649. Related: #5636.

PR #5651, opened 2026-09-02
fix: Bump iceberg-rust so native Iceberg writes URL-escape partition paths

### LINKED ISSUE #5636: Native Iceberg writes use an iceberg-rust revision missing partition-path escaping
### Describe the bug

The native Iceberg write path currently pins iceberg-rust revision:

`3d84c81353b1b23b6e4ae8eea8f8a021cc6927a7`

This revision predates apache/iceberg-rust#2875, which fixed `PartitionSpec::partition_to_path` to URL-encode partition field names and values.

As a result, Comet's native Iceberg writer can generate unescaped partition paths for values containing special characters such as `#`, `?`, or `/`.

This is not only a directory-layout difference. For example, an S3 partition value containing `#` can produce a committed location such as:

```text
s3://bucket/table/data/p=a#b/file.parquet
```

Iceberg Java's `S3FileIO` treats `#` and `?` as URI fragment/query delimiters when resolving the object key, so a location produced by the older iceberg-rust behavior can resolve to a different key when later read through the standard Iceberg Java path.

Upstream iceberg-rust already fixed this in:

* apache/iceberg-rust#2874
* apache/iceberg-rust#2875

### Steps to reproduce

_No response_

### Expected behavior

Native Iceberg writes should generate partition paths compatible with Iceberg Java's `PartitionSpec.partitionToPath` behavior.

Comet should consume an iceberg-rust revision containing apache/iceberg-rust#2875 and include a native-write → JVM-read interoperability regression test for escaped partition paths.

### Additional context

This was found during follow-up review of #5361.

The upstream issue has already been fixed, so no new iceberg-rust issue is required. The remaining work is for Comet to update its pinned iceberg-rust revision and verify interoperability.

Upstream fix:

* apache/iceberg-rust#2874
* apache/iceberg-rust#2875

### LINKED ISSUE #5641: Stale native Iceberg write wording in iceberg-writes.md
### Describe the bug

The accepted-divergences section of `docs/source/user-guide/latest/iceberg-writes.md` carries an "Operational caveat" saying native Iceberg writes leave partition paths unescaped (`region=a/b` where iceberg-java writes `region=a%2Fb`), and the eligibility table's "partition spec" row points at it.

That divergence disappears once #5636 bumps the iceberg-rust pin past apache/iceberg-rust#2875, which makes `partition_to_path` use the same form-urlencoding table as iceberg-java's `PartitionSpec.partitionToPath`. The caveat and the table cross-reference should go with it.

(Earlier drafts of this issue also listed a stale "the native writer is not yet wired in" paragraph and a stale note about the metrics-mode restriction; both were already corrected on `main` by #5361, so only the escaping entry remains.)

### Steps to reproduce

Read the "Accepted divergences" section of `iceberg-writes.md` on `main`.

### Expected behavior

No divergence is listed for partition paths, and the "partition spec" eligibility row reads "any".

### Additional context

Part of the native Iceberg writes epic, #5649. Fixed together with #5636.

andygrove:

Closes #5636. Closes #5641.

## Rationale for this change

The native Iceberg writer pins iceberg-rust at `3d84c81` (2026-07-29), one day before apache/iceberg-rust#2875 merged. That fix makes `PartitionSpec::partition_to_path` form-urlencode partition field names and values, which is what iceberg-java's `PartitionSpec.partitionToPath` has always done through `URLEncoder.encode`. Without it the native writer commits raw partition directories: a value containing `#` or `?` lands in the manifest as a location that iceberg-java's `S3FileIO` parses as a URI fragment or query, so a table written by Comet resolves to a different object key when read back through the standard Java path. This is the first phase-1 item of the native Iceberg writes epic (#5649).

- `native/Cargo.toml` pins `iceberg` and `iceberg-storage-opendal` to `8adaa872f31549dd5ad8255848715758228038bc`, the merge commit of apache/iceberg-rust#2875. That revision has the same arrow (58.4), parquet (58.4), DataFusion (54.1), and opendal (0.57) versions as the current pin, so the lockfile change is confined to the two iceberg crates and no Comet code needed to change. The v0.10.1 release was considered and rejected: it sits on a release branch that does not contain the fix and pins DataFusion 53.
- `iceberg-writes.md` drops the "partition paths are not URL-escaped" operational caveat and the eligibility table's cross-reference to it; the layouts now match.

Both encoders use the application/x-www-form-urlencoded table: alphanumerics and `*-._` unescaped, space as `+`, everything else percent-encoded as UTF-8, and `null` for a null value. The commits between this revision and iceberg-rust `main` (opendal 0.58, `ObjectStorageLocationGenerator`, a GCS host fix, additional parquet writer settings) are left for the pin policy in #5645 so this change stays a targeted correctness fix.

A new test in `CometIcebergWriteActionSuite`, "partition paths are URL-escaped like iceberg-java", writes the same rows through the native writer and the JVM writer into two tables partitioned by a string column whose values contain `/`, `#`, `?`, space, `=`, `%`, `+`, the unescaped set `*-._`, multi-byte characters, and the empty string. It asserts that the set of partition directories under each table's data location is identical, that the expected encodings (`region=a%2Fb`, `region=c%23d`, `region=g+h`, `region=*-._`) are present, and that the natively written table reads back correctly through both Comet's native scan and iceberg-java's reader.

The existing Iceberg suites (native scan, write action, write detection, rewrite actions, fuzz, reflection, proto translation) were run locally against the new pin on the default Spark profile.

### DIFF
--- a/spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
@@ -1172,6 +1172,59 @@ class CometIcebergWriteActionSuite
     }
   }
 
+  // iceberg-java's `PartitionSpec.partitionToPath` runs every partition name and value through
+  // `URLEncoder.encode`, and iceberg-rust's `partition_to_path` uses the same
+  // application/x-www-form-urlencoded table (apache/iceberg-rust#2875). Readers resolve files
+  // through manifest metadata rather than paths, but the committed location still has to be one
+  // every FileIO can open: an unescaped `#` in an S3 key is a fragment delimiter to `S3FileIO`.
+  test("native acceleration: partition paths are URL-escaped like iceberg-java") {
+    assumeNativeAcceleration()
+    withIcebergCatalog { warehouseDir =>
+      createTable(warehouseDir, "escaped_native", partitionSpec = "PARTITIONED BY (region)")
+      createTable(warehouseDir, "escaped_jvm", partitionSpec = "PARTITIONED BY (region)")
+      val regions = Seq("a/b", "c#d", "e?f", "g h", "i=j", "k%l", "m+n", "*-._", "日本", "")
+      val values = regions.zipWithIndex
+        .map { case (region, i) => s"($i, '$region', $i.5)" }
+        .mkString(", ")
+
+      spark.sql(s"INSERT INTO $catalog.$ns.escaped_native VALUES $values")
+      withSQLConf(CometConf.COMET_ICEBERG_WRITE_SPLIT_OPERATOR_ENABLED.key -> "false") {
+        spark.sql(s"INSERT INTO $catalog.$ns.escaped_jvm VALUES $values")
+      }
+
+      // Every partition directory the native writer produced exists in the JVM writer's layout
+      // and vice versa, so the two tables are byte-for-byte compatible in their data locations.
+      def partitionDirs(table: String): Set[String] = {
+        val location = warehouseDir.toURI.toString.stripSuffix("/")
+        spark
+          .sql(s"SELECT file_path FROM $catalog.$ns.$table.files")
+          .collect()
+          .map(_.getString(0))
+          .map { path =>
+            val relative = path.stripPrefix(location).split("/data/", 2)(1)
+            relative.substring(0, relative.lastIndexOf('/'))
+          }
+          .toSet
+      }
+      val nativeDirs = partitionDirs("escaped_native")
+      assert(nativeDirs == partitionDirs("escaped_jvm"), s"native layout: $nativeDirs")
+      assert(nativeDirs.contains("region=a%2Fb") && nativeDirs.contains("region=c%23d"))
+      assert(nativeDirs.contains("region=g+h") && nativeDirs.contains("region=*-._"))
+
+      // Both Comet's native scan and iceberg-java's reader open the escaped locations.
+      val expected = regions.zipWithIndex.map { case (region, i) => Row(i, region, i + 0.5) }
+      Seq("true", "false").foreach { cometEnabled =>
+        withSQLConf(CometConf.COMET_ENABLED.key -> cometEnabled) {
+          val rows = spark
+            .sql(s"SELECT id, region, amount FROM $catalog.$ns.escaped_native ORDER BY id")
+            .collect()
+            .toSeq
+          assert(rows == expected, s"comet=$cometEnabled: $rows")
+        }
+      }
+    }
+  }
+
   test("native acceleration: wide primitive types keep JVM-parity values and manifest metrics") {
     assumeNativeAcceleration()
     withIcebergCatalog { _ =>
--- a/spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
@@ -1187,10 +1187,10 @@ class CometIcebergWriteActionSuite
         .map { case (region, i) => s"($i, '$region', $i.5)" }
         .mkString(", ")
 
-      spark.sql(s"INSERT INTO $catalog.$ns.escaped_native VALUES $values")
-      withSQLConf(CometConf.COMET_ICEBERG_WRITE_SPLIT_OPERATOR_ENABLED.key -> "false") {
-        spark.sql(s"INSERT INTO $catalog.$ns.escaped_jvm VALUES $values")
+      assertNativeWriteEngages("escaped_native", regions.indices) {
+        spark.sql(s"INSERT INTO $catalog.$ns.escaped_native VALUES $values")
       }
+      spark.sql(s"INSERT INTO $catalog.$ns.escaped_jvm VALUES $values")
 
       // Every partition directory the native writer produced exists in the JVM writer's layout
       // and vice versa, so the two tables are byte-for-byte compatible in their data locations.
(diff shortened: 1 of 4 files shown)


spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
```diff
@@ -1172,6 +1172,59 @@ class CometIcebergWriteActionSuite
...
+    assumeNativeAcceleration()
+    withIcebergCatalog { warehouseDir =>
+      createTable(warehouseDir, "escaped_native", partitionSpec = "PARTITIONED BY (region)")
+      createTable(warehouseDir, "escaped_jvm", partitionSpec = "PARTITIONED BY (region)")
+      val regions = Seq("a/b", "c#d", "e?f", "g h", "i=j", "k%l", "m+n", "*-._", "日本", "")
+      val values = regions.zipWithIndex
+        .map { case (region, i) => s"($i, '$region', $i.5)" }
+        .mkString(", ")
+
+      spark.sql(s"INSERT INTO $catalog.$ns.escaped_native VALUES $values")
```
unikdahal: Is this actually exercising the native writer? `COMET_ICEBERG_NATIVE_WRITE_ENABLED` is off by default, and the existing native-write tests use `assertNativeWriteEngages` / `withNativeEnabled`.

Without that, this could become JVM-vs-JVM and still pass with the old iceberg-rust revision. Could we use `assertNativeWriteEngages` here?

sunchao (commented):
Reviewed `f13e1118` against `10537e14`. The upstream encoder is a simple fit without adding a Comet-specific path rewriter. I found no additional material defect in the dependency/path change.

[P2] The [existing native-writer coverage concern](https://github.com/apache/datafusion-comet/pull/5651#discussion_r3919497917) remains. The new INSERT does not enable native writing, so the directory comparison can pass with JVM writers on both sides. Please use `assertNativeWriteEngages`. Its helper also enables the local-table-scan conversion needed by the VALUES input and asserts `CometIcebergWriteExec`. Showing the repaired test fail with the old pin and pass with the new pin would establish the regression coverage.

A focused component check using the locked Rust encoder matched Java `URLEncoder` in all 160 cases. The old raw-formatting adapter differed in 89. The lock audit found only the two Iceberg records changed, with no package-version changes. These are component/source checks, not a Spark/native-writer or storage round trip. Current CI has 55 successful, 10 running and 7 skipped checks. Successful shard metadata alone does not prove this test exercised native writing.

andygrove:
cc @anuragmantri

spark/src/test/scala/org/apache/comet/CometIcebergWriteActionSuite.scala
andygrove: You were right, and the test was not exercising the native writer. `spark.comet.iceberg.write.enabled` is off by default, so both inserts went through the JVM writer and the comparison was JVM vs JVM.

Fixed in e0bd7bfe9: the native insert is now wrapped in `assertNativeWriteEngages("escaped_native", regions.indices)`, which flips the conf, enables the local-table-scan conversion the inline `VALUES` input needs, and asserts a `CometIcebergWriteExec` in the captured plan. The JVM twin now runs on the suite defaults instead of toggling `splitOperator.enabled`, matching the other parity tests.

To confirm the repaired test is genuine regression coverage, I rebuilt the native library against both pins.

Old pin (`3d84c81`), fails:

```
- native acceleration: partition paths are URL-escaped like iceberg-java *** FAILED ***
  HashSet("region=e?f", "region=g h", "region=i=j", "region=", "region=*-._", "region=k%l",
          "region=a/b", "region=c#d", "region=日本", "region=m+n")
  did not equal
  HashSet("region=a%2Fb", "region=e%3Ff", "region=", "region=%E6%97%A5%E6%9C%AC", "region=*-._",
          "region=k%25l", "region=g+h", "region=i%3Dj", "region=c%23d", "region=m%2Bn")
  (CometIcebergWriteActionSuite.scala:1210)
Tests: succeeded 0, failed 1, canceled 0, ignored 0, pending 0
```

New pin (`8adaa87`), passes:

```
- native acceleration: partition paths are URL-escaped like iceberg-java (8 seconds, 956 milliseconds)
Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
```

The old-pin output is the failure mode this PR is about: `region=a/b` splits into a nested directory, and `region=c#d` lands raw, which is the key `S3FileIO` would resolve differently. The full `CometIcebergWriteActionSuite` on the new pin is 53 succeeded, 0 failed.

andygrove:
> Reviewed `f13e1118` against `10537e14`. The upstream encoder is a simple fit without adding a Comet-specific path rewriter. I found no additional material defect in the dependency/path change.
> 
> [P2] The [existing native-writer coverage concern](https://github.com/apache/datafusion-comet/pull/5651#discussion_r3919497917) remains. The new INSERT does not enable native writing, so the directory comparison can pass with JVM writers on both sides. Please use `assertNativeWriteEngages`. Its helper also enables the local-table-scan conversion needed by the VALUES input and asserts `CometIcebergWriteExec`. Showing the repaired test fail with the old pin and pass with the new pin would establish the regression coverage.
> 
> A focused component check using the locked Rust encoder matched Java `URLEncoder` in all 160 cases. The old raw-formatting adapter differed in 89. The lock audit found only the two Iceberg records changed, with no package-version changes. These are component/source checks, not a Spark/native-writer or storage round trip. Current CI has 55 successful, 10 running and 7 skipped checks. Successful shard metadata alone does not prove this test exercised native writing.

This is now fixed

sunchao (approved):
The existing native-writer test finding is fixed at `e0bd7bfe`. The insert now uses `assertNativeWriteEngages`, which enables the required native paths and asserts `CometIcebergWriteExec`, one new snapshot and the expected IDs. The JVM comparison runs after those native settings are cleared. No new or remaining P1/P2 findings in this increment.

The production pins and lockfile are unchanged from the previous review. This follow-up was source and retained-evidence verification, not a new Spark/JNI run. Your old-pin failure, new-pin success and 53-test suite results remain author-reported. Current CI has 17 successful checks, seven skipped and seven still running, with no failed check in this snapshot.

unikdahal:
Thank @andygrove the test concern is addressed in the latest commit. LGTM now.

unikdahal (approved):

anuragmantri (approved):
LGTM. Thanks @andygrove

andygrove:
Merged. Thanks @sunchao @unikdahal @anuragmantri

PR #5654, opened 2026-09-03
fix: match Spark's duplicate field and field id semantics in parquet field lookup

### LINKED ISSUE #5801: Duplicate field ids inside a struct are not validated when the file schema equals the requested schema and no predicate is pushed
### Describe the bug

Comet validates Parquet field id lookups in the physical expression adapter, which resolves each file's mapping once per file (#5654). DataFusion 55's Parquet opener skips the adapter entirely when the file's physical schema compares equal to the logical schema and there is no pushed predicate. In that case a file whose struct carries two children with the same field id is read positionally instead of failing with Spark's duplicate field id error.

Spark-written files always carry key-value metadata that arrow-rs folds into the physical schema, so they never compare equal and always reach the adapter. The gap is limited to files with no key-value metadata at all, read with field id matching enabled and no data filter.

### Steps to reproduce

1. Write a Parquet file without key-value metadata whose schema is `s<x: long id=1, y: long id=1>` (parquet-mr with a hand-built MessageType).
2. Read it with the native scan, `spark.sql.parquet.fieldId.read.enabled=true`, and a requested schema identical to the file schema, with no filter.
3. Spark rejects the duplicate id during schema clipping; Comet returns the rows.

### Expected behavior

The duplicate field id error, as when a cast or predicate is present.

### Additional context

Found in review of #5654, where the once-per-file validation was added. The short circuit is in DataFusion's opener, so the fix is either a Comet-side check before the opener decides, or an upstream option to always run the adapter.

dwsmith1983:

Closes #5801.

Part of the restructuring of #5365 requested in review: this extracts the duplicate field and field id matching semantics that previously traveled with the Delta contrib work, re-derived on top of the folding that #5602 added. It also absorbs the reader-factory validation that was stacked on it as #5808, since that commit cannot rebase onto main on its own and the two could not merge separately.

## Rationale for this change

Four places where the native parquet field lookup diverges from Spark:

1. `remap_physical_schema` only shields id-bearing logical fields whose id is missing from the file. When a logical field's id matches one physical field but a stray physical column carries that logical field's name, the stray column can still name-match through the expression adapter fallback and hijack the read.
2. `parquet_convert_struct_to_struct` silently resolves a requested field id that matches more than one physical field to the first match. Spark raises the duplicate field error in field id lookup mode.
3. Case sensitive exact name lookup on duplicate names resolves to the first field, while Spark builds its name map with `.toMap`, where the last entry wins.
4. DataFusion's Parquet opener creates the physical expression adapter only when a predicate is pushed or the file schema differs from the requested schema. A file with no key-value metadata whose schema equals the requested schema, read with field id matching on and no data filter, skipped the validation above and read a struct with two children sharing a requested id positionally. Spark rejects that during schema clipping. Spark-written files always carry key-value metadata that arrow-rs folds into the file schema, so they always reach the adapter; the gap is metadata-free files.

- All id-bearing logical fields are shielded from name matching, the shield runs after the name match pass so a successful match claims the field first, and fake placeholder names skip a reserved set built from both schemas so they can never collide with real columns.
- A requested field id resolving to more than one physical field raises `SparkError::DuplicateFieldByFieldId` (`_LEGACY_ERROR_TEMP_2094`). Duplicate ids that no requested field references remain harmless, at the root as well as inside nested types: the adapter records a root ambiguity per logical field and raises it from `rewrite` only for the columns a read references, the same way nested ambiguities are handled. Before this, a duplicate id on columns the read never asked for failed the query, while Spark answers it because `clipParquetSchema` only sees the requested schema.
- Exact name lookup on duplicate names resolves to the last field.
- The field mapping is resolved once per file in the adapter factory and reused for every batch; a mapping that reorders fields (ids resolved out of position) disables the metadata-only relabel shortcut, which would otherwise hand back the file's columns under the requested names with the values swapped.
- The mapping resolver, the struct-holding type walk and the array converter now share one definition of which Arrow types are lists, so a `List<Struct>` read as a `LargeList<Struct>` (or any other pair of list representations) resolves its element fields with Spark's rules before Arrow changes the list layout. This is the same shape #5681 takes for the converter, so whichever of the two lands second rebases mechanically.
- The Parquet reader factory, which fetches every file's footer regardless of what the opener decides, gains an optional field id check installed at plan time only when field id matching is on and the requested schema carries a field id at any depth, so ordinary reads pay nothing. After the footer is fetched it converts the file schema to Arrow the way the opener does and resolves the requested fields with the same mapping resolver the adapter uses, raising the same errors. A per-file memo keyed on the cached footer's identity means a file is validated once per distinct footer, not once per batch.
- The JNI error bridge walks the cause chain so a Spark error raised inside the reader factory keeps its Spark exception class instead of surfacing as a generic read failure. This is wider than the reader path: any typed `SparkError` reachable through a `Context` or `Shared` wrapper anywhere in the plan now keeps its class. An ANSI failure in a pushed-down filter is unaffected, because DataFusion stringifies it into an `ArrowError::ComputeError` and nothing typed survives the chain.

Both validation paths now validate the same scope, the columns a read requests. The footer check covers the required schema, and the adapter raises only for referenced columns. The check also catches a duplicate id on root fields and inside a `list<struct>` element, so it closes the same opener-skip gap for the root-level check, not only the nested one.

There is no equivalent gap for the case-insensitive duplicate name error: for the file and requested schemas to compare equal, the requested schema would have to hold two fields folding to the same name, and Spark rejects that at analysis with `COLUMN_ALREADY_EXISTS`. Two fields can share an id while having distinct names, which is what makes the duplicate id error reachable and the `use_field_id && schema_holds_field_ids` gate on the footer check sufficient.

Rust, all written before the change they pin and failing on the previous head unless noted:

- Adapter and conversion: stray column hijacking the remap, case insensitive sibling null-filled, duplicate id silently reading the first match, first-wins name resolution, swapped field ids bypassing the relabel shortcut, root duplicate id raised only for referenced columns (`rewrite` of an unreferenced column succeeds, of the ambiguous one errors).
- List representations: a `List<Struct>` requested as `LargeList<Struct>` resolves to a non-positional element mapping, converts by id, and null-fills a requested element field the file lacks instead of reading its neighbour.
- Scans through `DataSourceExec` on a file written without key-value metadata, asserting first that the file schema equals the requested schema so the opener skips the adapter: duplicate struct id rejected, unique ids read, check inert with field id matching off, the same through the planner's data schema plus projection wiring, duplicate id at the root, duplicate id inside a `list<struct>` element, case-insensitive duplicate name (`_LEGACY_ERROR_TEMP_2093`) rejected at footer load with the case-sensitive read of the same file succeeding, and unrequested root duplicate ids read through the adapter path.
- JNI bridge: the chain walk finds a `SparkError` behind DataFusion's `Context` and `External` wrappers; a plain Parquet error still classifies as a file read failure.

Core crate 367 tests, bridge crate 32, clippy with `-D warnings` and fmt clean. Each commit compiles and passes the parquet module tests on its own.

Scala: a `ParquetReadV1Suite` case writes the file with parquet-mr and no key-value metadata, asserts the footer's key-value map is empty and that the plan carries the native scan, and expects Spark's duplicate field id error; it reported no exception against the previous native library. `ParquetReadV1Suite`, `CometNativeReaderSuite` and `SparkErrorConverterSuite` on Spark 3.5 against the rebuilt native library: 137 succeeded, 0 failed.

### DIFF
--- a/native/core/src/parquet/cast_column.rs
+++ b/native/core/src/parquet/cast_column.rs
@@ -24,7 +24,9 @@ use arrow::{
     record_batch::RecordBatch,
 };
 
-use crate::parquet::parquet_support::{field_id, spark_parquet_convert, SparkParquetOptions};
+use crate::parquet::parquet_support::{
+    spark_parquet_convert_with_mapping, FieldMapping, SparkParquetOptions,
+};
 use datafusion::common::format::DEFAULT_CAST_OPTIONS;
 use datafusion::common::{DataFusionError, Result as DataFusionResult};
 use datafusion::logical_expr::ColumnarValue;
@@ -68,21 +70,6 @@ fn types_differ_only_in_field_names(physical: &DataType, logical: &DataType) -> 
                         && (pf.data_type() == lf.data_type()
                             || types_differ_only_in_field_names(pf.data_type(), lf.data_type()))
                 })
-        }
-        _ => false,
-    }
-}
-
-/// Returns true if any struct field in the type tree carries Parquet field-id
-/// metadata, mirroring the per-struct `should_match_by_id` check in
-/// `parquet_convert_struct_to_struct`.
-fn contains_field_id_metadata(data_type: &DataType) -> bool {
-    match data_type {
-        DataType::Struct(fields) => fields
-            .iter()
-            .any(|f| field_id(f).is_some() || contains_field_id_metadata(f.data_type())),
-        DataType::List(f) | DataType::LargeList(f) | DataType::Map(f, _) => {
-            contains_field_id_metadata(f.data_type())
         }
         _ => false,
     }
@@ -168,8 +155,11 @@ pub struct CometCastColumnExpr {
     /// Options forwarded to [`cast_column`].
     cast_options: CastOptions<'static>,
     /// Spark parquet options for complex nested type conversions.
-    /// When present, enables `spark_parquet_convert` as a fallback.
+    /// When present, enables the nested conversion as a fallback.
     parquet_options: Option<SparkParquetOptions>,
+    /// Which file field supplies each requested nested field, resolved once per file and
+    /// reused for every batch. Set together with `parquet_options`.
+    field_mapping: Option<Arc<FieldMapping>>,
 }
 
 // Manually derive `PartialEq`/`Hash` as `Arc<dyn PhysicalExpr>` does not
@@ -181,6 +171,7 @@ impl PartialEq for CometCastColumnExpr {
             && self.target_field.eq(&other.target_field)
             && self.cast_options.eq(&other.cast_options)
             && self.parquet_options.eq(&other.parquet_options)
+            && self.field_mapping.eq(&other.field_mapping)
     }
 }
 
@@ -191,6 +182,7 @@ impl Hash for CometCastColumnExpr {
         self.target_field.hash(state);
         self.cast_options.hash(state);
         self.parquet_options.hash(state);
+        self.field_mapping.hash(state);
     }
 }
 
@@ -229,12 +221,19 @@ impl CometCastColumnExpr {
             target_field,
             cast_options: cast_options.unwrap_or(DEFAULT_CAST_OPTIONS),
             parquet_options: None,
+            field_mapping: None,
         })
     }
 
-    /// Set Spark parquet options to enable complex nested type conversions.
-    pub fn with_parquet_options(mut self, options: SparkParquetOptions) -> Self {
+    /// Enable nested type conversions with Spark parquet options and the field mapping
+    /// resolved for this expression's physical and target types.
+    pub fn with_parquet_options(
+        mut self,
+        options: SparkParquetOptions,
+        field_mapping: Arc<FieldMapping>,
+    ) -> Self {
         self.parquet_options = Some(options);
+        self.field_mapping = Some(field_mapping);
         self
     }
 }
@@ -285,22 +284,20 @@ impl PhysicalExpr for CometCastColumnExpr {
         let input_physical_field = self.input_physical_field.data_type();
         let target_field = self.target_field.data_type();
 
-        // An id-based read must resolve struct fields through spark_parquet_convert,
-        // which validates the requested ids (e.g. a requested id duplicated in the
-        // file schema errors); metadata-only relabeling would skip that validation.
-        let id_based_read = self
-            .parquet_options
+        // Relabeling only swaps metadata, so it is right when every requested field reads
+        // the file field at its own position. A mapping that reorders fields (ids resolved
+        // to other positions) has to go through the nested conversion below.
+        let positional = self
+            .field_mapping
             .as_ref()
-            .is_some_and(|opts| opts.use_field_id)
-            && contains_field_id_metadata(target_field);
+            .is_none_or(|mapping| mapping.is_positional());
 
         match (input_physical_field, target_field) {
             // Nested types that differ only in field names (e.g., List element named
             // "item" vs "element", or Map entries named "key_value" vs "entries").
             // Re-label the array so the DataType metadata matches the logical schema.
-            // Skipped for id-based reads so field-id validation still runs below.
             (physical, logical)
-                if !id_based_read
+                if positional
                     && physical != logical
                     && types_differ_only_in_field_names(physical, logical) =>
             {
@@ -312,16 +309,17 @@ impl PhysicalExpr for CometCastColumnExpr {
                     other => Ok(other),
                 }
             }
-            // Fallback: use spark_parquet_convert for complex nested type conversions
-            // (e.g., List<Struct{a,b,c}> → List<Struct{a,c}>, Map field selection, etc.)
-            _ => {
-                if let Some(parquet_options) = &self.parquet_options {
-                    let converted = spark_parquet_convert(value, target_field, parquet_options)?;
-                    Ok(converted)
-                } else {
-                    Ok(value)
-                }
-            }
+            // Fallback: nested conversion through the resolved mapping
+            // (e.g., List<Struct{a,b,c}> -> List<Struct{a,c}>, Map field selection, etc.)
+            _ => match (&self.parquet_options, &self.field_mapping) {
+                (Some(parquet_options), Some(mapping)) => spark_parquet_convert_with_mapping(
+                    value,
+                    target_field,
+                    mapping,
+                    parquet_options,
+                ),
+                _ => Ok(value),
+            },
         }
     }
 
@@ -345,8 +343,8 @@ impl PhysicalExpr for CometCastColumnExpr {
             Arc::clone(&self.target_field),
             Some(self.cast_options.clone()),
         )?;
-        if let Some(opts) = &self.parquet_options {
-            new_expr = new_expr.with_parquet_options(opts.clone());
+        if let (Some(opts), Some(mapping)) = (&self.parquet_options, &self.field_mapping) {
+            new_expr = new_expr.with_parquet_options(opts.clone(), Arc::clone(mapping));
         }
         Ok(Arc::new(new_expr))
     }
@@ -359,6 +357,7 @@ impl PhysicalExpr for CometCastColumnExpr {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::parquet::parquet_support::resolve_field_mapping;
     use arrow::array::{
         Array, Int32Array, StringArray, TimestampMicrosecondArray, TimestampMillisecondArray,
     };
@@ -375,33 +374,31 @@ mod tests {
         )]))
     }
 
-    /// A file struct with two children sharing field id 1 must raise Spark's
-    /// duplicate-id error when the requested schema looks id 1 up, even though
-    /// every child name matches and the relabel shortcut would otherwise apply.
+    /// File struct `x` (id 1) = 42, `y` (id 2) = 43; requested struct names them the same
+    /// but swaps the ids. Names and types match, so only the positional gate keeps the
+    /// relabel shortcut from firing: the mapping reads by id and the result must be
+    /// `x` = 43, `y` = 42.
     #[test]
-    fn test_field_id_read_rejects_duplicate_ids_despite_matching_names() {
-        let physical_fields = Fields::from(vec![
-            int_field_with_id("x", 1),
-            int_field_with_id("y", 1),
-            int_field_with_id("z", 2),
-        ]);
-        let logical_fields = Fields::from(vec![
-            int_field_with_id("x", 1),
-            int_field_with_id("y", 3),
-            int_field_with_id("z", 2),
-        ]);
+    fn test_swapped_field_ids_bypass_relabel_shortcut() {
+        let physical_fields =
+            Fields::from(vec![int_field_with_id("x", 1), int_field_with_id("y", 2)]);
+        let logical_fields =
+            Fields::from(vec![int_field_with_id("x", 2), int_field_with_id("y", 1)]);
 
         let input_field = Arc::new(Field::new(
             "s",
             DataType::Struct(physical_fields.clone()),
             true,
         ));
-        let target_field = Arc::new(Field::new("s", DataType::Struct(logical_fields), true));
+        let target_field = Arc::new(Field::new(
+            "s",
+            DataType::Struct(logical_fields.clone()),
+            true,
+        ));
 
         let columns: Vec<ArrayRef> = vec![
             Arc::new(Int32Array::from(vec![42])),
             Arc::new(Int32Array::from(vec![43])),
-            Arc::new(Int32Array::from(vec![44])),
         ];
         let struct_arr = StructArray::new(physical_fields, columns, None);
         let schema = Schema::new(vec![Arc::clone(&input_field)]);
@@ -409,20 +406,34 @@ mod tests {
 
         let mut opts = SparkParquetOptions::new(EvalMode::Legacy, "UTC", false);
         opts.use_field_id = true;
+        let mapping = Arc::new(
+            resolve_field_mapping(input_field.data_type(), target_field.data_type(), &opts)
+                .unwrap(),
+        );
+        assert!(!mapping.is_positional());
 
         let col_expr: Arc<dyn PhysicalExpr> = Arc::new(Column::new("s", 0));
         let cast_expr = CometCastColumnExpr::try_new(col_expr, input_field, target_field, None)
             .unwrap()
-            .with_parquet_options(opts);
+            .with_parquet_options(opts, mapping);
 
-        let err = cast_expr
-            .evaluate(&batch)
-            .expect_err("requested field id 1 matches two file fields and must error");
-        let msg = err.to_string();
-        assert!(
-            msg.contains("_LEGACY_ERROR_TEMP_2094") && msg.contains("id=1"),
-            "expected duplicate field id error, got: {msg}"
-        );
+        let ColumnarValue::Array(arr) = cast_expr.evaluate(&batch).unwrap() else {
+            panic!("expected array result");
+        };
+        assert_eq!(arr.data_type(), &DataType::Struct(logical_fields));
+        let result = arr.as_any().downcast_ref::<StructArray>().unwrap();
+        let x = result
+            .column(0)
+            .as_any()
+            .downcast_ref::<Int32Array>()
+            .unwrap();
+        let y = result
+            .column(1)
+            .as_any()
+            .downcast_ref::<Int32Array>()
+            .unwrap();
+        assert_eq!(x.value(0), 43);
+        assert_eq!(y.value(0), 42);
     }
 
     /// Companion guard: without any field ids the relabel shortcut must keep
@@ -479,12 +490,17 @@ mod tests {
         .unwrap();
 
         // Enabling id read mode without any id metadata must not disable the
-        // shortcut either.
+        // shortcut either: the resolved mapping is positional.
         let mut opts = SparkParquetOptions::new(EvalMode::Legacy, "UTC", false);
         opts.use_field_id = true;
+        let mapping = Arc::new(
+            resolve_field_mapping(input_field.data_type(), target_field.data_type(), &opts)
+                .unwrap(),
+        );
+        assert!(mapping.is_positional());
         let id_mode_expr = CometCastColumnExpr::try_new(col_expr, input_field, target_field, None)
             .unwrap()
-            .with_parquet_options(opts);
+            .with_parquet_options(opts, mapping);
 
         for cast_expr in [plain_expr, id_mode_expr] {
             let result = cast_expr.evaluate(&batch).unwrap();
@@ -548,7 +564,10 @@ mod tests {
                 let expr: Arc<dyn PhysicalExpr> = Arc::new(Column::new("ts", 0));
                 let cast_expr = CometCastColumnExpr::try_new(expr, input_field, target_field, None)
                     .unwrap()
-                    .with_parquet_options(SparkParquetOptions::new(eval_mode, "UTC", false));
+                    .with_parquet_options(
+                        SparkParquetOptions::new(eval_mode, "UTC", false),
+                        Arc::new(FieldMapping::Leaf),
+                    );
 
                 let input = TimestampMillisecondArray::from(vec![Some(1_234), Some(-1_234), None])
                     .with_timezone_opt(source_tz.clone());
--- a/native/core/src/parquet/cast_column.rs
+++ b/native/core/src/parquet/cast_column.rs
@@ -24,7 +24,7 @@ use arrow::{
     record_batch::RecordBatch,
 };
 
-use crate::parquet::parquet_support::{spark_parquet_convert, SparkParquetOptions};
+use crate::parquet::parquet_support::{field_id, spark_parquet_convert, SparkParquetOptions};
 use datafusion::common::format::DEFAULT_CAST_OPTIONS;
 use datafusion::common::{DataFusionError, Result as DataFusionResult};
 use datafusion::logical_expr::ColumnarValue;
@@ -68,6 +68,21 @@ fn types_differ_only_in_field_names(physical: &DataType, logical: &DataType) -> 
                         && (pf.data_type() == lf.data_type()
                             || types_differ_only_in_field_names(pf.data_type(), lf.data_type()))
                 })
+        }
+        _ => false,
+    }
+}
+
+/// Returns true if any struct field in the type tree carries Parquet field-id
+/// metadata, mirroring the per-struct `should_match_by_id` check in
+/// `parquet_convert_struct_to_struct`.
+fn contains_field_id_metadata(data_type: &DataType) -> bool {
+    match data_type {
+        DataType::Struct(fields) => fields
+            .iter()
+            .any(|f| field_id(f).is_some() || contains_field_id_metadata(f.data_type())),
+        DataType::List(f) | DataType::LargeList(f) | DataType::Map(f, _) => {
+            contains_field_id_metadata(f.data_type())
         }
         _ => false,
     }
@@ -270,12 +285,24 @@ impl PhysicalExpr for CometCastColumnExpr {
         let input_physical_field = self.input_physical_field.data_type();
         let target_field = self.target_field.data_type();
 
+        // An id-based read must resolve struct fields through spark_parquet_convert,
+        // which validates the requested ids (e.g. a requested id duplicated in the
+        // file schema errors); metadata-only relabeling would skip that validation.
+        let id_based_read = self
+            .parquet_options
+            .as_ref()
+            .is_some_and(|opts| opts.use_field_id)
+            && contains_field_id_metadata(target_field);
+
         match (input_physical_field, target_field) {
             // Nested types that differ only in field names (e.g., List element named
             // "item" vs "element", or Map entries named "key_value" vs "entries").
             // Re-label the array so the DataType metadata matches the logical schema.
+            // Skipped for id-based reads so field-id validation still runs below.
             (physical, logical)
-                if physical != logical && types_differ_only_in_field_names(physical, logical) =>
+                if !id_based_read
+                    && physical != logical
+                    && types_differ_only_in_field_names(physical, logical) =>
             {
                 match value {
                     ColumnarValue::Array(array) => {
@@ -338,6 +365,142 @@ mod tests {
     use arrow::datatypes::{Field, Fields};
     use datafusion::physical_expr::expressions::Column;
     use datafusion_comet_spark_expr::EvalMode;
+    use parquet::arrow::PARQUET_FIELD_ID_META_KEY;
+    use std::collections::HashMap;
+
+    fn int_field_with_id(name: &str, id: i32) -> Field {
+        Field::new(name, DataType::Int32, true).with_metadata(HashMap::from([(
+            PARQUET_FIELD_ID_META_KEY.to_string(),
+            id.to_string(),
+        )]))
+    }
+
+    /// A file struct with two children sharing field id 1 must raise Spark's
+    /// duplicate-id error when the requested schema looks id 1 up, even though
+    /// every child name matches and the relabel shortcut would otherwise apply.
+    #[test]
+    fn test_field_id_read_rejects_duplicate_ids_despite_matching_names() {
+        let physical_fields = Fields::from(vec![
+            int_field_with_id("x", 1),
+            int_field_with_id("y", 1),
+            int_field_with_id("z", 2),
+        ]);
+        let logical_fields = Fields::from(vec![
+            int_field_with_id("x", 1),
+            int_field_with_id("y", 3),
+            int_field_with_id("z", 2),
+        ]);
+
+        let input_field = Arc::new(Field::new(
+            "s",
+            DataType::Struct(physical_fields.clone()),
+            true,
+        ));
+        let target_field = Arc::new(Field::new("s", DataType::Struct(logical_fields), true));
+
+        let columns: Vec<ArrayRef> = vec![
+            Arc::new(Int32Array::from(vec![42])),
+            Arc::new(Int32Array::from(vec![43])),
+            Arc::new(Int32Array::from(vec![44])),
+        ];
+        let struct_arr = StructArray::new(physical_fields, columns, None);
+        let schema = Schema::new(vec![Arc::clone(&input_field)]);
+        let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(struct_arr)]).unwrap();
+
+        let mut opts = SparkParquetOptions::new(EvalMode::Legacy, "UTC", false);
+        opts.use_field_id = true;
+
+        let col_expr: Arc<dyn PhysicalExpr> = Arc::new(Column::new("s", 0));
+        let cast_expr = CometCastColumnExpr::try_new(col_expr, input_field, target_field, None)
+            .unwrap()
+            .with_parquet_options(opts);
+
+        let err = cast_expr
+            .evaluate(&batch)
+            .expect_err("requested field id 1 matches two file fields and must error");
+        let msg = err.to_string();
+        assert!(
+            msg.contains("_LEGACY_ERROR_TEMP_2094") && msg.contains("id=1"),
+            "expected duplicate field id error, got: {msg}"
+        );
+    }
+
+    /// Companion guard: without any field ids the relabel shortcut must keep
+    /// handling name-only differences, whether or not id read mode is enabled.
+    #[test]
+    fn test_relabel_shortcut_kept_for_name_only_differences_without_ids() {
+        // Physical: s { col: List(Field("item", Int32)) }
+        // Logical:  s { col: List(Field("element", Int32)) }
+        let physical_list_field = Arc::new(Field::new("item", DataType::Int32, true));
+        let logical_list_field = Arc::new(Field::new("element", DataType::Int32, true));
+        let physical_fields = Fields::from(vec![Field::new(
+            "col",
+            DataType::List(Arc::clone(&physical_list_field)),
+            true,
+        )]);
+        let logical_fields = Fields::from(vec![Field::new(
+            "col",
+            DataType::List(logical_list_field),
+            true,
+        )]);
+
+        let input_field = Arc::new(Field::new(
+            "s",
+            DataType::Struct(physical_fields.clone()),
+            true,
+        ));
+        let target_field = Arc::new(Field::new(
+            "s",
+            DataType::Struct(logical_fields.clone()),
+            true,
+        ));
+
+        let values = Int32Array::from(vec![1, 2, 3]);
+        let list = ListArray::new(
+            physical_list_field,
+            arrow::buffer::OffsetBuffer::new(vec![0, 2, 3].into()),
+            Arc::new(values),
+            None,
+        );
+        let struct_arr = StructArray::new(physical_fields, vec![Arc::new(list) as ArrayRef], None);
+        let schema = Schema::new(vec![Arc::clone(&input_field)]);
+        let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(struct_arr)]).unwrap();
+
+        let col_expr: Arc<dyn PhysicalExpr> = Arc::new(Column::new("s", 0));
+
+        // Without parquet options the fallback arm would return the value
+        // unchanged, so a relabeled result proves the shortcut itself fired.
+        let plain_expr = CometCastColumnExpr::try_new(
+            Arc::clone(&col_expr),
+            Arc::clone(&input_field),
+            Arc::clone(&target_field),
+            None,
+        )
+        .unwrap();
+
+        // Enabling id read mode without any id metadata must not disable the
+        // shortcut either.
+        let mut opts = SparkParquetOptions::new(EvalMode::Legacy, "UTC", false);
+        opts.use_field_id = true;
+        let id_mode_expr = CometCastColumnExpr::try_new(col_expr, input_field, target_field, None)
+            .unwrap()
+            .with_parquet_options(opts);
+
+        for cast_expr in [plain_expr, id_mode_expr] {
+            let result = cast_expr.evaluate(&batch).unwrap();
+            let ColumnarValue::Array(arr) = result else {
+                panic!("expected array result");
+            };
+            assert_eq!(arr.data_type(), &DataType::Struct(logical_fields.clone()));
+            let result_struct = arr.as_any().downcast_ref::<StructArray>().unwrap();
+            let result_list = result_struct
+                .column(0)
+                .as_any()
+                .downcast_ref::<ListArray>()
+                .unwrap();
+            assert_eq!(result_list.len(), 2);
+        }
+    }
 
     #[test]
     fn test_rejects_millisecond_logical_timestamp() {
--- a/native/core/src/parquet/parquet_support.rs
+++ b/native/core/src/parquet/parquet_support.rs
@@ -22,7 +22,7 @@ use arrow::array::{
 };
 use arrow::buffer::NullBuffer;
 use arrow::compute::can_cast_types;
-use arrow::datatypes::{FieldRef, Fields};
+use arrow::datatypes::{Field, FieldRef, Fields, Schema};
 use arrow::{
     array::{
         cast::AsArray, new_null_array, types::TimestampMicrosecondType,
@@ -306,6 +306,39 @@ impl FieldMapping {
             FieldMapping::Leaf => true,
         }
     }
+}
+
+/// True when a field of `schema`, at any nesting depth, carries a Parquet field id.
+pub(crate) fn schema_holds_field_ids(schema: &Schema) -> bool {
+    schema.fields().iter().any(|f| field_holds_id(f))
+}
+
+fn field_holds_id(field: &Field) -> bool {
+    field_id(field).is_some()
+        || match field.data_type() {
+            DataType::Struct(fields) => fields.iter().any(|f| field_holds_id(f)),
+            DataType::List(f) | DataType::LargeList(f) | DataType::Map(f, _) => field_holds_id(f),
+            _ => false,
+        }
+}
+
+/// Resolve every requested root field against `file_schema` the way the expression adapter
+/// does, keeping only the ambiguity Spark reports. DataFusion's opener creates the adapter
+/// only when a predicate is pushed or the file schema differs from the requested one, so the
+/// reader factory runs this on every footer it loads to cover the files the adapter never sees.
+pub(crate) fn validate_field_mapping(
+    file_schema: &Schema,
+    requested_schema: &Schema,
+    parquet_options: &SparkParquetOptions,
+) -> Result<(), SparkError> {
+    // `ParquetMissingFieldIds` needs no counterpart here: a file with no ids differs from an
+    // id-bearing requested schema in field metadata, so the opener runs the adapter for it.
+    resolve_field_mapping(
+        &DataType::Struct(file_schema.fields().clone()),
+        &DataType::Struct(requested_schema.fields().clone()),
+        parquet_options,
+    )
+    .map(|_| ())
 }
 
 /// Resolve how `to_type` reads from `from_type`, recursing through struct, list, and map
--- a/native/core/src/parquet/parquet_support.rs
+++ b/native/core/src/parquet/parquet_support.rs
@@ -427,11 +427,14 @@ pub(crate) fn match_struct_fields(
     let should_match_by_id =
         parquet_options.use_field_id && to_fields.iter().any(|f| field_id(f).is_some());
 
-    let from_id_to_index: HashMap<i32, usize> = if should_match_by_id {
-        let mut map = HashMap::new();
+    // Keep EVERY index sharing an ID: Spark's `matchIdField` raises
+    // `foundDuplicateFieldInFieldIdLookupModeError` when a requested ID resolves to
+    // more than one file field, and only when that ID is actually requested.
+    let from_id_to_indices: HashMap<i32, Vec<usize>> = if should_match_by_id {
+        let mut map: HashMap<i32, Vec<usize>> = HashMap::new();
         for (i, field) in from_fields.iter().enumerate() {
             if let Some(id) = field_id(field) {
-                map.entry(id).or_insert(i);
+                map.entry(id).or_default().push(i);
             }
         }
         map
@@ -465,7 +468,26 @@ pub(crate) fn match_struct_fields(
             |(to_pos, to_field)| match (should_match_by_id, field_id(to_field)) {
                 // Spark treats a missing ID match as a missing column rather than
                 // falling back to name match.
-                (true, Some(id)) => Ok(from_id_to_index.get(&id).copied()),
+                (true, Some(id)) => match from_id_to_indices.get(&id) {
+                    None => Ok(None),
+                    Some(indices) if indices.len() == 1 => Ok(Some(indices[0])),
+                    // Mirror Spark's `foundDuplicateFieldInFieldIdLookupModeError`
+                    // (`_LEGACY_ERROR_TEMP_2094`): a requested ID resolving to more
+                    // than one file field is ambiguous.
+                    Some(indices) => {
+                        let matched = indices
+                            .iter()
+                            .map(|&i| from_fields[i].name().as_str())
+                            .collect::<Vec<_>>()
+                            .join(", ");
+                        Err(DataFusionError::External(Box::new(
+                            SparkError::DuplicateFieldByFieldId {
+                                required_id: id,
+                                matched_fields: matched,
+                            },
+                        )))
+                    }
+                },
                 _ => match folded_to_indices.get(to_folded[to_pos].as_str()) {
                     // Mirror Spark's `foundDuplicateFieldInCaseInsensitiveModeError`: a
                     // requested field matching more than one file field is ambiguous. Gated on
@@ -473,7 +495,8 @@ pub(crate) fn match_struct_fields(
                     // `!case_sensitive`): when case-sensitive the fold is identity, so a
                     // collision means byte-identical sibling names, and raising an error whose
                     // message says "in case-insensitive mode" would be wrong. Fall through to
-                    // the first match in that case.
+                    // the LAST match in that case, matching Spark's
+                    // `caseSensitiveParquetFieldMap` built with `.toMap` (later entry wins).
                     Some(indices) if indices.len() > 1 && !parquet_options.case_sensitive => {
                         let matched: Vec<&str> = indices
                             .iter()
@@ -483,7 +506,7 @@ pub(crate) fn match_struct_fields(
                             SparkError::duplicate_field_case_insensitive(to_field.name(), &matched),
                         )))
                     }
-                    Some(indices) => Ok(Some(indices[0])),
+                    Some(indices) => Ok(indices.last().copied()),
                     None => Ok(None),
                 },
             },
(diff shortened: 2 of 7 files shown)


native/core/src/parquet/parquet_support.rs
```diff
@@ -301,15 +304,35 @@ fn parquet_convert_struct_to_struct(
...
                     // Spark treats a missing ID match as a missing column rather than
                     // falling back to name match.
-                    (true, Some(id)) => from_id_to_index.get(&id).copied(),
+                    (true, Some(id)) => match from_id_to_indices.get(&id) {
+                        None => None,
+                        Some(indices) if indices.len() == 1 => Some(indices[0]),
+                        // Mirror Spark's `foundDuplicateFieldInFieldIdLookupModeError`
+                        // (`_LEGACY_ERROR_TEMP_2094`): a requested ID resolving to more
+                        // than one file field is ambiguous.
+                        Some(indices) => {
```
sunchao: [P2] Run duplicate-ID validation before metadata-only struct relabeling

Could you route metadata-only struct adaptations through this validation too? For file struct `s<x: int id=1, y: int id=1, z: int id=2>` and requested `s<x: int id=1, y: int id=3, z: int id=2>`, Spark rejects requested ID 1 as ambiguous. DataFusion emits a struct cast, but `CometCastColumnExpr::evaluate` takes `types_differ_only_in_field_names` and calls `relabel_array`, because that predicate ignores field-ID metadata. The new lookup never runs and leaves all three physical values in place. A focused probe using the current cast expression and a real Arrow/Parquet round trip returned `[42, 43, 44]`, while renaming requested `x` made the same input reach the duplicate-ID error. Could you guard the relabel shortcut for ID-based reads and add a cast-expression or scan regression with unchanged child names?

sunchao (commented):
This fixes stray-name matching and placeholder collisions, and adds nested duplicate-ID rejection and last-wins exact-name lookup. One gap remains: metadata-only struct relabeling bypasses the duplicate-ID check, as detailed inline.

I compared the code with maintained Spark 3.5 and 4.0 sources. Eight component-check groups passed using extracted Comet helpers with Arrow/Parquet 58.4.0 and DataFusion 54.1.0. A separate probe reproduced the cast bypass and verified a renamed-child control. These probes use limited scaffolding and are not full Comet scan, JNI or Spark query tests. The reported 246 native and 58 Spark 3.5 tests are the author's results.

At 04:51 UTC, current-head CI had 29 successful, 32 running and 7 skipped checks. Full CI validation was still pending.

native/core/src/parquet/parquet_support.rs
dwsmith1983: Good catch, the shortcut sailed right past the new validation. Fixed in 3d68f22f9: the relabel arm is now guarded so that when use_field_id is set and the requested type carries field id metadata, evaluation falls through to the struct conversion where the duplicate id lookup runs. Chose the guard at the call site rather than inside types_differ_only_in_field_names since that predicate is a pure structural comparison with no access to the parquet options. Your exact probe is now a regression test (unchanged child names, duplicate id 1, asserts the 2094 error) plus a companion pinning that the fast path survives for name only differences without ids and for the flag alone.

sunchao:
Reviewed head `58e67fee` against base `55ae4f20`. The PR is focused, but I found **one new regression, an incomplete validation fix, and avoidable batch-processing overhead**.

1. **[P2] Placeholder collisions can suppress column defaults.**  
   [schema_adapter.rs:177](https://github.com/apache/datafusion-comet/blob/58e67fee104aaf6ddd4119b96969d44d45bdbcf0/native/core/src/parquet/schema_adapter.rs#L177) reserves exact names, while missing-column detection uses case-folded names. A generated `__comet_unmatched_field_id_1` therefore collides with requested `__COMET_UNMATCHED_FIELD_ID_1`. In my reproduction, the base returns the configured default `7`; this PR returns NULL.

   Reserve names using the existing folded schema names. That change passed the reproduction. Constructing the reservation set only when shielding needs it would also avoid extra hashing on ordinary reads without field IDs.

2. **[P2] Duplicate-ID validation still depends on whether a cast occurs.**  
   The new guard in [cast_column.rs:292](https://github.com/apache/datafusion-comet/blob/58e67fee104aaf6ddd4119b96969d44d45bdbcf0/native/core/src/parquet/cast_column.rs#L292) misses identical physical/requested schemas. Such reads can omit the cast entirely or return before the guard.

   A real native Parquet scan of identical `s<x: long id=1, y: long id=1>` schemas returned `[42, 43]`; Spark 4.1.3’s schema-clipping check rejected the duplicate ID. **This also occurs on the base, so it is an incomplete fix, not a new regression.** Validation needs to cover reads that require no conversion.

3. **Avoid allocating a vector for every unique ID on every struct conversion.**  
   [parquet_support.rs:270](https://github.com/apache/datafusion-comet/blob/58e67fee104aaf6ddd4119b96969d44d45bdbcf0/native/core/src/parquet/parquet_support.rs#L270) changes the index to `HashMap<i32, Vec<usize>>`. An allocation probe using the old and new construction loops measured **8 → 264 allocations for 256 unique IDs**.

   A compact unique/duplicate entry would preserve the behavior. Collect matching field names only when reporting an ambiguity. The new `contains_field_id_metadata` predicate also depends on immutable expression state and can be computed once.

The strongest design improvement is to **resolve and validate requested fields once per file schema, then reuse the mapping across batches**. That addresses the validation bypasses and repeated lookup work together. Metadata-only relabeling remains safe when the resolved mapping is positional. A small mapping object is a useful abstraction here.

Validation: **100 native Parquet tests passed**, with default HDFS features disabled. Additional head/base probes confirmed both correctness cases. Performance evidence measures component allocations, not overall scan speed. CI snapshot: **57 passed, 7 running, 7 skipped**. Nothing was posted to GitHub.

dwsmith1983:
Thanks, all three are addressed in a69b5f5c0, following the once-per-file design you suggested.

The physical expression adapter factory already runs once per file schema, so it now resolves a small `FieldMapping` tree (struct sources, list and large list elements, map key and value, leaf) for every logical field that holds a struct, mirroring Spark's `clipParquetSchema` at each nesting level. Duplicate requested ids and ambiguous case-insensitive names are detected there, stored per logical field, and raised when the column is referenced, whether or not a cast is later emitted. The per-batch conversion receives the resolved mapping and applies it positionally, so there is no hashing or per-id allocation on the batch path; the index type is a compact entry that records an index and an ambiguity flag, and matching names are collected only when the error is built. The `contains_field_id_metadata` predicate is gone; the relabel shortcut is gated on the mapping being positional instead.

Your repros: the identical `s<x: long id=1, y: long id=1>` schema now raises the duplicate id error with no cast in the plan, pinned in Rust through the exec path and in `ParquetReadV1Suite` (the Scala case fails against the previous native library and passes now). The folded placeholder collision returns the configured default again, pinned in the adapter tests with Spark-style key-value metadata on the file.

One residual worth naming: DataFusion's opener skips the adapter entirely when the logical and physical schemas compare equal and no predicate exists. Spark-written files always carry key-value metadata that arrow-rs folds into the physical schema, so they always go through the adapter, but a file with no metadata at all and duplicated ids inside a struct would still read positionally. Happy to cover that in a follow-up if you think it matters.

dwsmith1983:
@sunchao the once-per-file mapping round covering your three findings is pushed. Ready for another look.

andygrove (commented):
`convert_struct` in `native/core/src/parquet/parquet_support.rs` around line 563 now calls `array.column(from_index)` with an index resolved at planning time against `adapted_physical_schema`, where the old code derived it from the runtime array's own `DataType`. The only guard is `sources.len() != to_fields.len()`, which checks the target side. What guarantees the struct array the reader hands back always carries the same children in the same order as the physical field the mapping was resolved against? If that can drift at all, this is an index panic in the executor rather than a `DataFusionError`. Checking `from_index` against `array.num_columns()` would bound the worst case.

The description lists last-wins exact-name resolution as one of the three fixes and `resolve_struct_mapping` does it for struct children. At the top level in case-sensitive mode with no field ids, `needs_remap` is false in `schema_adapter.rs` around line 520, so resolution falls to `DefaultPhysicalExprAdapter`, which goes through `Schema::index_of` and returns the first match. Spark builds `caseSensitiveParquetFieldMap` at the root message level with the same `.toMap` it uses for nested groups. Was the top level deliberately left out of scope?

The field-id ambiguity path is covered from several angles now. The case-insensitive name ambiguity that `resolve_struct_mapping` raises around line 388 does not appear to have a companion test in the new `struct_field_matching` module. It might be worth pinning that half too, since it is the branch that decides between an error and a silently wrong column.

On the residual you named where DataFusion skips the adapter when the two schemas compare equal and there is no predicate, I confirmed that short circuit in the 55.0.0 opener. Could you open a tracking issue and link it here so it does not get lost? The branch also conflicts with `main` right now and needs a rebase before anything meaningful runs against it.

dwsmith1983:
Rebased onto main and the three points are in the head (b438dce52).

Bounds: `convert_struct` now looks the source index up with `columns().get` and returns an error naming the requested field, the index, and the child count instead of indexing; a test feeds a struct with fewer children than the mapping expects and asserts the error. A struct column arriving with a non-struct mapping errors the same way rather than passing through. I did not add a type-equality check there, since `convert_array` dispatches on the runtime child type and strict equality would reject coercions it handles.

Root last-wins: not deliberate, the top level had simply fallen to the default adapter. With duplicate exact names at the root in case-sensitive mode the remap path now runs, the shadowed earlier fields get a placeholder name so the default adapter's `index_of` lands on the last one, and the nested resolver does the same; Spark 3.5's `clipParquetGroupFields` uses one `.toMap` for root and nested groups. Tests at the adapter level and through `DataSourceExec`, plus one that a field id match still wins over the duplicate. One thing worth knowing: parquet-mr writes two root columns named `d` into a single column chunk keyed by path, so a Spark-written file with that shape reads six interleaved values in Spark and arrow-rs alike, and no Spark-comparable end-to-end assertion exists for it; the Rust scan test uses arrow-rs to write the file.

The case-insensitive ambiguity now has its companion test in `struct_field_matching`: `A` and `a` against a requested `a` errors naming both in case-insensitive mode and reads `a` in case-sensitive mode.

The opener short circuit is tracked in #5801.

dwsmith1983:
The opener short circuit from #5801 is fixed in #5808, stacked on this branch: the reader factory validates the requested field ids from the footer with the same resolver, so metadata-free files with identical schemas and no filter now raise the same error.

andygrove (commented):
Thanks for the rework. All four points from my earlier review are addressed and I traced each one against the head rather than the summary.

`convert_struct` now resolves the child with `array.columns().get(from_index)` and returns an error naming the requested field, the index and the child count, so a struct arriving with fewer children than the mapping was resolved against is a `DataFusionError` rather than a panic in the executor. The root last-wins gap is closed properly: `has_duplicate_names` widens `needs_remap` so a case-sensitive read with duplicate root names no longer falls through to `DefaultPhysicalExprAdapter::index_of`, `shadowed_by_later_duplicate` hides the earlier ones, and I checked the interaction with the ID path in both orders. `[d(id 5), d]` against logical `d(id 5)` keeps `d` on the ID match (the ID block returns before the shadow rename) and shields the sibling, and `[d, d(id 5)]` shadows the first and keeps the second. Your note that parquet-mr keys two root `d` columns to one chunk, so there is no Spark-comparable end-to-end assertion for that shape, is worth having in the record. The case-insensitive ambiguity companion test is there, and #5801 tracks the opener short circuit with #5808 stacked on it.

One thing I would like resolved before this merges, and it is about merge order rather than the code in isolation.

### `convert_array`'s list arms collide with #5681

#5681 is approved and touches the same `match` in `parquet_convert_array_impl`. It widens the single `(List, List)` arm to the whole array family on both sides, including cross-representation pairs, specifically so a `List<Struct>` read as a `LargeList<Struct>` resolves its element fields with Spark's rules before Arrow changes the list layout. Without that, `cast_struct_to_struct` gives up on name matching and falls back to positional, so a requested field absent from the file reads the neighbouring column's values.

This branch rewrites the same arms into `convert_array` and covers `(List, List)` and `(LargeList, LargeList)` only. That is not a regression against `main` on its own. The problem is what a mechanical conflict resolution produces if #5681 lands first. `resolve_field_mapping` matches `(List, List) | (LargeList, LargeList)` and falls to `_ => Ok(FieldMapping::Leaf)` for every other pair, and `type_holds_struct` only walks `Struct`, `List`, `LargeList` and `Map`. So a `List<Struct>` read as a `LargeList<Struct>` resolves to `Leaf`, #5681's widened arm calls `mapping.list_element()?`, which returns `Ok(&Leaf)`, and the struct child then lands on the new `(Struct(_), Struct(_), other)` arm and fails with `struct column resolved to a non-struct field mapping`. A read that works on `main` today would start erroring, and it would do so from the arm you added as an internal assertion, so it would read like a planner bug rather than a missing case in the resolver.

Whichever of the two goes second has to carry the other's coverage, and the part that a conflict marker will not point at is that `resolve_field_mapping` and `type_holds_struct` need widening too, not just the `convert_array` arms. Could you and @peterxcli agree an order, and could the one that rebases extend all three places together? A cross-representation `List<Struct>` case in `struct_field_matching` would pin it.

Related and smaller: the three places that decide list handling (`type_holds_struct`, `resolve_field_mapping`, `convert_array`'s arms) currently agree only because they all default to leaf-or-passthrough for `FixedSizeList` and the two view types. That is fine today, but it is three lists that have to move together and nothing says so. Would a comment on `type_holds_struct` pointing at the other two help, or is there a way to derive one from another?

One last thing I checked and am happy with, since it is the subtle part of the new design. Gating the relabel shortcut in `CometCastColumnExpr` on `mapping.is_positional()` is a real fix, not just plumbing: swapped field IDs give names and types that satisfy `types_differ_only_in_field_names`, so before this the relabel would have returned the file's columns under the requested names and silently swapped the values. `test_swapped_field_ids_bypass_relabel_shortcut` pins exactly that, and the companion test keeps the shortcut alive for the ordinary `item`/`element` relabel with ID read mode on. Since the mapping is resolved per file in `create` and read per batch, and `spark_parquet_convert`'s public entry still resolves from the runtime type, I do not see a way for a batch to reach a mapping resolved against a different schema.

dwsmith1983:
> Is there a reason not to fold it into #5654?

No. #5808 is now the last three commits here and is closed, so this is the one reviewable change; the description covers both.

> Could you and @peterxcli agree an order, and could the one that rebases extend all three places together?

Extended all three here so the order stops mattering: `resolve_field_mapping`, `type_holds_struct` and `convert_array` share one `list_element_field` helper covering `List`, `LargeList`, `FixedSizeList` and the two view types, and the converter now converts the element values through the mapping first, rebuilds the array in the file's representation around them, and only then lets Arrow's cast change the layout, the same shape as #5681's arm. Three cases in `struct_field_matching` pin it: a `List<Struct>` requested as `LargeList<Struct>` resolves to a non-positional element mapping, converts by id with swapped values coming back swapped, and null-fills a requested element field the file lacks instead of reading its neighbour. Left @peterxcli a note on #5681 with that, so either rebase is taking one arm over the other.

> Would a comment on `type_holds_struct` pointing at the other two help, or is there a way to derive one from another?

Derived: all three call the helper, and its doc names them.

> Would you narrow the root check in `remap_physical_schema` to the requested columns, or file it as a follow-up and reword the description so the asymmetry reads the way it actually behaves?

Narrowed here, and you are right about the description; it now says what happens. `remap_physical_schema` records the ambiguity per logical field instead of failing, leaves the colliding physical fields out of the id rename so the shield fake-names them, and `rewrite` raises for referenced columns the way nested ambiguities already are. Your three-column example reads `[7]` through the adapter now (test `unrequested_duplicate_root_field_ids_read_through_adapter`, with a column the file lacks forcing the adapter), and `rewrite` of the ambiguous column still raises `_LEGACY_ERROR_TEMP_2094`. Both paths validate the same scope, the columns a read requests.

> Could you thread `case_sensitive` through the helper and cover it?

Threaded, with the case the description now explains: `s(id 10)<x, X>` in a metadata-free file fails at footer load with `_LEGACY_ERROR_TEMP_2093` in case-insensitive mode and reads in case-sensitive mode. The root duplicate id and the `list<struct>` element duplicate are pinned as well, and the description says the root check is closed by the same footer check.

The three inline items are in: the memo is a `parking_lot::Mutex` with the helper and the poison imports gone, `throw_spark_error_json` is the one throw site with the `External` arm going through `spark_error_json_in_chain` (a new test drives a `SparkError` behind `External` and `Context` across the JNI boundary and checks the class), and the opener-skip explanation lives on `FieldIdCheck` with the other two pointing at it. The description also carries the `COLUMN_ALREADY_EXISTS` reasoning and the note that the chain walk widens which typed errors keep their class.

Core crate 367 tests, bridge 32, each commit compiles on its own, and `ParquetReadV1Suite`, `CometNativeReaderSuite` and `SparkErrorConverterSuite` pass on Spark 3.5 against the rebuilt library.

andygrove (approved):
All four points from my earlier review are in, and the #5681 merge-order concern turned out fine since that landed first and this branch rebased onto a base containing it. I spot-checked the resolution rules against `ParquetReadSupport.clipParquetGroupFields` rather than taking them on trust, and they line up. Field id wins with no name fallback, a non-id requested field matches by name in the same group, `caseSensitiveParquetFieldMap` is `.toMap` so exact-name collisions resolve last-wins silently, and `caseInsensitiveParquetFieldMap` is a `groupBy` on `toLowerCase(Locale.ROOT)` that raises on more than one hit. `resolve_struct_mapping` matches all of it, including raising only for the requested id rather than validating the whole group.

One naming point. This adds `schema_holds_field_ids` in `parquet_support.rs`, which walks nested types, and it now sits one word away from `schema_has_field_ids` in `schema_adapter.rs`, which only looks at root fields. They are easy to confuse and they are not interchangeable. Could the new one be named so the difference is visible at the call site, or could the old one gain a line saying it is deliberately root only?

There is a behavior question behind that name which I think is follow-up work rather than something for this PR. `schema_has_field_ids` is what gates the `ParquetMissingFieldIds` rejection at the top of `remap_physical_schema`. Spark's equivalent in `ParquetReadSupport.getRequestedSchema` uses `containsFieldIds` on the file schema and `ParquetUtils.hasFieldIds` on the requested schema, both recursive, and it does not consult `spark.sql.parquet.fieldId.read.enabled` at all. So with the default `use_field_id` of false, a requested schema carrying `parquet.field.id` read against a file with no ids returns rows in Comet where Spark raises, and a file whose ids sit only on nested fields is rejected by Comet where Spark null-fills through `matchIdField`. Both reproduce on `main` so neither is from this change, but this is the PR that puts a correct recursive predicate one module away from the one making that decision. Could you file a tracking issue and link it here?

Separately, this needs sequencing with #5786, which I have also commented on. That one adds a `validate_field_names` footer check that errors on byte-identical sibling names, which is the opposite of the last-wins resolution here, and it hangs a second validator off the same `get_metadata` call site. `git merge-tree` reports eight conflicting hunks in `eager_page_index_reader_factory.rs` and one in `parquet_exec.rs`, and both PRs are clean against `main` on their own so neither CI run shows it. #5884 also records Spark 4.1.3 returning `{0, 100, 1}` for the nested duplicate fixture rather than last-wins, which is worth reconciling against what `clipParquetGroupFields` does, since the clip and the reader may not agree. Could you and @ErikBPF settle which behavior we want before either lands?

sunchao:
Reviewed head `1ea1be161` against base `db790673d`. **I found one P2 correctness regression to fix before merging.**

### P2: Duplicate root names now select the wrong column

The new [root-field shadowing](https://github.com/apache/datafusion-comet/blob/1ea1be161ec664b8705b109be5004593a5d7bb4f/native/core/src/parquet/schema_adapter.rs#L348) hides earlier duplicate names and selects the last column.

I tested one Parquet file containing two separate `d` columns, reading it with explicit schema `d bigint` and case sensitivity enabled:

| Reader | Result |
|---|---|
| Spark 4.1.3, vectorized | `[1, 2, 3]` |
| Base native scan harness | `[1, 2, 3]` |
| PR native scan harness | `[10, 20, 30]` |

This changes previously matching results. Spark’s schema-clipping `.toMap` behavior does not establish what its reader actually returns. Preserve the working root-column selection and add a comparison against Spark using the same file.

This supplies concrete differential evidence for the concern already discussed in [#5884](https://github.com/apache/datafusion-comet/issues/5884).

### Validation and remaining assessment

- **CI:** 56 successful, 10 skipped; `Required Checks` passed.
- **Local:** 22 component tests passed. Additional scan probes verified the earlier default-value and duplicate-ID fixes.
- **Build limit:** The registry lacks pinned DataFusion 55.1. Native probes used current changed modules with DataFusion 55.0 / Arrow 59.3, rather than a full current-head Comet build. Relevant upstream root-read paths were checked against 55.1.
- No other actionable findings emerged from mapping, projection, ownership, footer-cache, or JNI-error review.

Evidence: Spark result, base result, PR result (local reproduction logs).

Nothing was published to GitHub.

dwsmith1983:
> Preserve the working root-column selection and add a comparison against Spark using the same file.

The comparison is in: `test-data/duplicate-root-names.parquet` carries two root `d` columns, and a new `ParquetReadSuite` test reads it with `d bigint` under case sensitivity and asserts Comet's answer and plan against Spark. It passes on 3.5 and 4.0, both returning `[1, 2, 3]`, because the JVM-side projection binds the first `d` before the native adapter sees the file, so the adapter's last-wins rule only decides between columns that are both projected. Your harness drove the adapter alone, which is where `[10, 20, 30]` comes from. The end-to-end selection is unchanged from base and now pinned.

> Could the new one be named so the difference is visible at the call site, or could the old one gain a line saying it is deliberately root only?

Both: the recursive predicate is now `any_nested_field_has_id`, and `schema_has_field_ids` says it is root only on purpose and names the other.

> Could you file a tracking issue and link it here?

Will do once we settle the nested-duplicate behaviour with @ErikBPF on #5786, since the same issue should record both the field-id gating gap and the nested resolution we want; I would rather not open two overlapping ones.

ErikBPF:
Thanks—the root parity test and naming clarification address those points. My proposed order remains #5786 first, then rebase #5654.

Preserve the pre-decoder guard for selected ambiguous fields until correct physical-leaf selection is demonstrated. Unrequested nested duplicates can remain readable where structural narrowing proves they are pruned, as #5786 now covers. The root JVM projection result does not settle nested decoding.

Can we agree on that order and behavior? #5884 already tracks duplicate resolution; please explicitly track the field-ID gating gap and link both PRs.

dwsmith1983:
> Preserve the working root-column selection and add a comparison against Spark using the same file.

Correction to my earlier note: after merging today's `main`, the end-to-end read does return the last `d`, so your P2 stands on the current base. Root duplicates now resolve first-wins: a later exact duplicate takes the fake name instead of the earlier one, so the adapter binds the first file column the way Spark's reader does. The two adapter tests assert `[1, 2, 3]`, and the new `ParquetReadSuite` comparison against Spark over the two-column file is the guard that caught it.

dwsmith1983:
> Can we agree on that order and behavior? #5884 already tracks duplicate resolution; please explicitly track the field-ID gating gap and link both PRs.

The field-id gating gap is #5936, linked from both PRs. On behavior, this PR now does what Andy proposed on #5786 for the nested case rather than last-wins: a requested field that matches more than one byte-identical sibling is refused with a message naming the field, while reading the unique sibling beside them still works, both pinned in Rust and by a Scala read of a file whose struct carries `dup`, `dup`, `other`. Root exact duplicates read the first file column, as Spark's reader does, pinned end to end against Spark.

On order: since the refusal now lives in the resolver this PR owns, I would land this one first and rebase #5786 onto it for the shapes that remain, the metadata-time check included if it is still needed. The fallible name folding from #5845 is already propagated here.

sunchao:
Reviewed head `1fa518b54` against base `1d0ce5fe8`. **One P2 regression remains.**

### P2: Footer validation rejects supported duplicate-root reads

[parquet_support.rs:352](https://github.com/apache/datafusion-comet/blob/1fa518b54ea68de77577dc61be74a470e181c51b/native/core/src/parquet/parquet_support.rs#L352) treats the root schema as a nested struct, applying the new duplicate-name rejection before the adapter can select the first root column.

**Trigger:** a file contains `d`, `d`, and `a(id=7)`. Request `d` by name and `a` by ID, with case sensitivity and field-ID reads enabled.

| Reader | Result |
|---|---|
| Spark 4.1.3 vectorized | `(1,100), (2,200), (3,300)` |
| Base native scan harness | Same rows |
| PR native scan harness | Duplicate-field error for `d` |

Disabling field-ID reads makes the PR return the correct rows. The unrelated ID-bearing column activates the footer check and causes the failure.

**Fix:** preserve root first-wins behavior in footer validation while retaining nested rejection. Add a mixed ID/name regression through the reader factory.

### Validation

- **120 component tests passed**, plus Spark comparisons covering 32 read combinations.
- Reviewed mapping, projection, array lifetimes, footer caching, and JNI errors. No other introduced actionable defects found.
- **Build limitation:** the registry lacks DataFusion 55.1. Component probes used current sources with DataFusion 55.0 / Arrow 59.3; full Comet integration tests could not run.
- **CI:** 7 passed, 12 queued/running, 9 skipped.

Full review and evidence (local review report and reproduction logs). Nothing posted to GitHub.

dwsmith1983:
> **Fix:** preserve root first-wins behavior in footer validation while retaining nested rejection. Add a mixed ID/name regression through the reader factory.

Done. `validate_field_mapping` now renames every later exact duplicate at the root to a name no request can match, keeping its type and metadata, before resolving, so the footer check binds the first root column the way the adapter does while a requested nested duplicate is still refused. Two regressions: a unit test on the check itself with `d`, `d`, `a(id=7)` requested by name and by id under case sensitivity and field ids (passes), the same under case-insensitivity (Spark's ambiguity error) and the nested shape (refused); and a reader-factory test that writes that file, loads its footer and runs `FieldIdCheck.validate` on it. Both fail without the change. Merged with `main` at a8e8157ea first.

ErikBPF:
Agreed on landing this first, then integrating the remaining nested-only checks in #5786. The separate mixed-type root investigation in #5964 now has parquet-java PR apache/parquet-java#3796. Spark descriptor/filter changes have passed targeted experimental checks, but still need upstream JIRA tracking, a current-master port and an available parquet dependency. That work remains separate from these Comet PRs.

comphead (commented):
The solution might be overcomplicated here. I would expect a small check, that checks all the plan nodes and related schema and if the struct found with duplicated name, then fallback

comphead (changes requested):
Let me try to come up with more light weight solution

andygrove:
Triage note: #5786 is working the other side of duplicate Parquet field names in the native scan. It rejects ambiguous sibling names before the decoder is built, closing #5783, and explicitly leaves "Spark-compatible duplicate-name resolution" as separate work — which is this PR. You share `parquet_exec.rs`, `schema_adapter.rs` and `eager_page_index_reader_factory.rs`.

The risk is that the two disagree on the outcome for the same file: this PR resolves case-sensitive duplicates last-wins the way Spark's `.toMap` does, while #5786 would reject some of those reads before the lookup ever runs. Could you and @ErikBPF work out which cases each of you is claiming, and in what order they land?

ErikBPF:
@dwsmith1983 yours have a lot more work implemented. I see no problem in landing it first and I adapt mine after. @andygrove when do you intent to close 1.1.0? Just want to know the feasibility to have both on this release

andygrove:
> @dwsmith1983 yours have a lot more work implemented. I see no problem in landing it first and I adapt mine after. @andygrove when do you intent to close 1.1.0? Just want to know the feasibility to have both on this release

There is no hard deadline, but I was hoping to have 1.1.0 released before the end of the month.

dwsmith1983:
> Could you and @ErikBPF work out which cases each of you is claiming, and in what order they land?

Order as ErikBPF said above: this PR first, #5786 adapts after. On the split: #6004 now declines at planning any requested schema that repeats an id, so I have dropped the equal-schema rationale from the tests here and they request a single id that the file carries twice. What this PR still adds is native raising Spark's error when a requested id is ambiguous in the file, last-wins exact-name lookup like Spark's `.toMap`, the id shield against a stray same-name column, one list-type definition across the mapping paths, and typed Spark errors surviving the JNI cause chain. #5786 claims byte-identical duplicate names that the decoder cannot represent.

> I would expect a small check, that checks all the plan nodes and related schema and if the struct found with duplicated name, then fallback

That check is what #6004 landed for field ids, and it stays. The reader-side error covers the case a plan-level check cannot see, an id the requested schema names once that the file carries twice. If you would rather keep one mechanism, I can trim the footer check to that file-side case; the adapter path has to stay for the Spark error text.

comphead:
Thanks @dwsmith1983 @ErikBPF, folks, WDYT to proceed with #5786 to patch physical scans and then check what cases are still not addressed in this field and fix pursue them in this PR?

The initial reason I was slightly concerned is 2k LOC to patch an extreme scenario

dwsmith1983:
> Thanks @dwsmith1983 @ErikBPF, folks, WDYT to proceed with #5786 to patch physical scans and then check what cases are still not addressed in this field and fix pursue them in this PR?
> 
> The initial reason I was slightly concerned is 2k LOC to patch an extreme scenario

We can cut it down. The footer check's reason to exist went away with #6004, so I can drop it and its tests, which takes out the reader-factory and the Rust test bulk. What stays is the adapter raising Spark's error for an ambiguous requested id, last-wins exact names, the id shield and typed errors through the JNI cause chain, none of which #5786 covers. On order, #5786 currently does not type-check against main per @sunchao and would reject the DataFusion-written file you showed on #6004, so I would still land this first and let #5786 rebase onto the smaller diff. Does that work?

dwsmith1983:
@comphead would you like me to cut it down?

dwsmith1983:
> The initial reason I was slightly concerned is 2k LOC to patch an extreme scenario

Cut down in 24030e526, as offered above. The footer-time validator is gone with its tests, so the reader factory and the scan setup are back to main and the Rust test bulk with them. What stays is what #5786 does not cover: the resolver raising Spark's duplicate-id error for an ambiguous requested id at any nesting level, exact-name lookups that bind the way Spark's do, the id shield against a stray same-name column, one list definition across the mapping paths, and the field mapping resolved once per file instead of per batch. The JNI cause-chain walk went too: restoring main's error conversion left every kept test green, because the adapter raises its errors at the top level, so that piece only served the footer check. The description is rewritten to match, with the remaining gap stated in full: a metadata-free file whose schema equals the requested one is still read positionally because the opener skips the adapter, #6004 covers the case where the requested schema itself repeats an id, and nested duplicate names in such a file are tracked in #6136.

Against main the diff is now 6 files changed, 2102 insertions(+), 356 deletions(-), down from 9 files with 2898 insertions and 384 deletions; the reader factory, the scan setup and the JNI error conversion are back to main byte for byte.

andygrove:
@comphead could you take another look?

comphead:
Thanks @dwsmith1983 @andygrove 
#5786 is about to be merged, lets see what is still to be fixed and address it in this PR

dwsmith1983:
> #5786 is about to be merged, lets see what is still to be fixed and address it in this PR

#5786 now owns the duplicate name case at every level. `match_struct_fields` rejects a requested nested name that two siblings carry, `rewrite` rejects a referenced root name the file repeats, `id_duplicate_roots` rejects an id whose physical root name is duplicated, and `check_decoded_field_names` refuses any duplicate inside a subtree a Comet cast decodes in full. This branch sits on top of all of that. The first-wins root binding is gone, since main rejects that read and pins the rejection in `CometNativeReaderSuite`, and the branch's own duplicate-name message is gone with it. Where the resolver here replaced `match_struct_fields`, it raises main's `duplicate Parquet field name` error for byte-identical siblings in both case modes, so every path reports that shape the same way.

What this PR still fixes is field ids, which #5786 does not touch.

A requested id that two file fields carry inside a struct. File `s (id 2) <x: bigint (id 1), y: bigint (id 1)>`, read with `spark.sql.parquet.fieldId.read.enabled=true` and schema `s (id 2) <x: bigint (id 1)>`. Spark raises `_LEGACY_ERROR_TEMP_2094` from `matchIdField`. On main `match_struct_fields` keeps the first field per id (`map.entry(id).or_insert(i)`) and returns `x = 42` with no error, and none of the duplicate-name checks fire because the names differ. The resolver here raises Spark's error at any depth, and only for the ids a read references, so an unrequested duplicate id at the root no longer fails the whole scan the way `remap_physical_schema` did. `ParquetReadSuite` has this as `duplicate field id inside a struct is rejected when a requested id matches two fields`.

Swapped ids inside a struct. File `s <x (id 1), y (id 2)>`, requested `s <x (id 2), y (id 1)>`. Names and types agree, so main's relabel shortcut in `CometCastColumnExpr` hands back the file's columns under the requested names and the values come out swapped. The mapping resolved once per file is not positional for that shape, so the relabel is skipped and each field reads by id.

The id shield in `remap_physical_schema` runs after the name match. Main renames a physical column away whenever its folded name equals an id-bearing requested name, before an id-less requested field can claim it by name. File `κ (id 2)`, requested `Κ (id 1)` plus id-less `κ`, case-insensitive: main null-fills both, Spark and this branch return `(NULL, 7)`.

The rest is the once-per-file `FieldMapping` handed to every cast instead of re-resolving per batch, and one `list_element_field` helper shared by the resolver, the converter and the adapter's struct walk, so a `List<Struct>` read as `LargeList<Struct>` resolves its elements by id.

In the merge, main's `needs_remap`, the collision check in `rewrite`, `id_duplicate_roots` and the `checked_decoded_expr` wrapping stand as written. Removed from the branch: `has_duplicate_names`, `shadowed_by_earlier_duplicate`, the three root first-wins tests, the matching `ParquetReadSuite` read and `duplicate-root-names.parquet`. `issue_5783_nested_name_duplicate` is ported onto `resolve_field_mapping` because `match_struct_fields` no longer exists.

One divergence from Spark remains after the merge, and it comes from #5786 rather than from this branch. `id_duplicate_roots` in `schema_adapter.rs` rejects a requested id whenever its physical root name repeats in the file, even when the id itself matches exactly one field. Physical `[d (id 1), d (id 2)]` read as `d (id 1)` with field-id reads on: Spark's `matchIdField` in `ParquetReadSupport.clipParquetGroupFields` (branch-3.5 lines 438-450) finds one field for id 1 and reads it, and main raises `duplicate Parquet field name 'd'`. `CometNativeReaderSuite`'s `duplicate Parquet field names - root group and unprojected root duplicates` pins that rejection, so the merge keeps it and a Rust test named `known_divergence_repeated_root_name_rejects_unambiguous_field_id` records it. The one way to close it is to narrow `id_duplicate_roots` to fire only when the requested id is ambiguous, in this PR, which reverses that test.

native/core/src/parquet/cast_column.rs
```diff
@@ -270,12 +284,22 @@ impl PhysicalExpr for CometCastColumnExpr {
         let input_physical_field = self.input_physical_field.data_type();
         let target_field = self.target_field.data_type();
 
+        // Relabeling only swaps metadata, so it is right when every requested field reads
+        // the file field at its own position. A mapping that reorders fields (ids resolved
+        // to other positions) has to go through the nested conversion below.
+        let positional = self
```
andygrove: This gate fixes a silent wrong-result bug on `main`, and I filed #6192 for it. Your summary covers swapped ids, but the shape I'd expect users to actually hit is a nested column dropped and added back under the same name. A file with `s struct<x (id 1), y (id 2)>` read as `s struct<x (id 3), y (id 2)>` returns the old `x` values on `main`, where Spark returns null. The same happens inside a list element and a map value. All of those match Spark 4.1.3 on this branch, with the native scan in the plan.

Could you add `Closes #6192` to the description, plus a `ParquetReadSuite` test that runs the drop-and-re-add read through `checkSparkAnswerAndOperator` for a struct, a list element and a map value? At the moment only `test_swapped_field_ids_bypass_relabel_shortcut` covers this, at the expression level.

native/core/src/parquet/schema_adapter.rs
```diff
@@ -4170,4 +4686,88 @@ mod test {
...
+        assert_eq!(defaulted.value(0), 7);
+    }
+
+    /// File and requested schema are identical: `s` holding `x` and `y` that both carry
+    /// field id 1. No column needs conversion, so no cast is ever emitted, yet Spark's
+    /// `clipParquetSchema` rejects the read because requested id 1 resolves to two file
+    /// fields. The validation must therefore run when the file schema is mapped, not
+    /// only inside a cast.
+    #[tokio::test]
+    async fn parquet_duplicate_struct_field_id_rejected_without_cast() {
```
andygrove: #6004 now declines a requested schema that repeats an id at planning, so this test covers a read Spark can no longer hand to the native scan, and the reasoning in its doc comment no longer applies. The reachable version is @sunchao's shape from the first round, `s<x (id 1), y (id 1), z (id 2)>` read as `s<x (id 1), y (id 3), z (id 2)>`. `main` returns all three values there, Spark raises, and this branch raises too. Its test (`test_field_id_read_rejects_duplicate_ids_despite_matching_names`) was dropped in `b4549c9d2`. Could this one use that shape instead?

native/core/src/parquet/parquet_support.rs
```diff
@@ -162,43 +169,327 @@ impl SparkParquetOptions {
...
+    index: usize,
+) {
+    matches
+        .entry(key)
+        .and_modify(|m| *m = m.also(index))
+        .or_insert_with(|| FieldMatch::first(index));
+}
+
+/// Comma-joined names of the fields carrying `id`, for the duplicate-id error message.
+pub(crate) fn field_names_with_id(fields: &Fields, id: i32) -> String {
```
andygrove: Spark brackets this list in `matchIdField` with `mkString("[", ", ", "]")`, so its message reads `Found duplicate field(s) "1": [x, y] in id mapping mode` and ours reads `"1": x, y`. Now that this helper formats the list for root and nested fields alike, could it add the brackets? The Scala test could then compare the whole message with Spark's instead of just the prefix.

native/core/src/parquet/cast_column.rs
dwsmith1983: > Could you add `Closes #6192` to the description, plus a ParquetReadSuite test that runs the drop-and-re-add read through checkSparkAnswerAndOperator for a struct, a list element and a map value?

Added `Closes #6192` to the description and three ParquetReadSuite tests, `nested column dropped and re-added under the same name with a new field id: struct`, `: list element` and `: map value`. They share one file, written with field ids on, holding `struct<x (id 1), y (id 2)>` as a top level column, as a list element and as a map value, and each reads its column with x carrying id 3 and y keeping id 2. Each runs through `checkSparkAnswerAndOperator`, asserts `CometNativeScanExec` is in the plan, and pins the expected rows explicitly: x is null and y keeps its values, including a null struct inside the list and the map, empty containers, and a null column.

One thing the run turned up: Spark 3.5's own vectorized reader raises on the list and map reads (`ParquetColumnVector` rejects the `_fake_name_` field that `clipParquetSchema` generates for the unmatched id below a list or map), and Spark 4.0.4 still has that check, so for those two the comparison with Spark runs from 4.1 on and the pinned rows carry the check on 3.5 and 4.0. The struct case compares with Spark on every version.

With the gate in `cast_column.rs` forced to always take the relabel shortcut, all three fail with the old x values: the struct test returns `[1,[1,10]]` and `[2,[5,50]]` where Spark returns `[1,[null,10]]` and `[2,[null,50]]`, the list test returns `[2,20], [3,30]` and the map test `k -> [4,40]` in place of nulls. With the gate as pushed all three pass, on Spark 3.5 and 4.1.3.

native/core/src/parquet/schema_adapter.rs
dwsmith1983: > Could this one use that shape instead?

Yes. The test is now `parquet_duplicate_file_field_id_rejected_when_requested` and uses the shape from the first round: the file holds `s<x (id 1), y (id 1), z (id 2)>` and the read asks for `s<x (id 1), y (id 3), z (id 2)>`, so the duplicate sits only in the file and a planner that declines repeated ids in the requested schema still hands this read to the native scan. The scan raises the duplicate id error for requested id 1 matching `x` and `y`, where a positional read would hand back all three values. The doc comment says why the identical-schema shape is no longer reachable and what this one proves instead.

native/core/src/parquet/parquet_support.rs
dwsmith1983: > Now that this helper formats the list for root and nested fields alike, could it add the brackets?

Done. `field_names_with_id` now returns the list bracketed and comma joined the way `matchIdField` renders it, so the message Comet hands to `foundDuplicateFieldInFieldIdLookupModeError` reads `Found duplicate field(s) "1": [x, y] in id mapping mode` on both sides. The Rust display string dropped its own brackets so the list is not wrapped twice, and the Rust assertions pin `id=1 matches [x, y]` rather than only the id.

On the Scala side, `multiple id matches` and `duplicate field id inside a struct is rejected when a requested id matches two fields` run the same read with Comet off, take the duplicate id message Spark raises, check that it lists the expected fields (`"1": [a, rand2]` for the root case, `"1": [x, y]` for the struct), and assert that Comet's message equals it in full.

native/core/src/parquet/parquet_support.rs
```diff
@@ -162,43 +169,330 @@ impl SparkParquetOptions {
...
+                // other path raises for a duplicate the decoder cannot represent.
+                Some(m) if m.ambiguous => {
+                    let matched: Vec<&str> = from_folded
+                        .iter()
+                        .zip(from_fields.iter())
+                        .filter(|(folded, _)| *folded == &to_folded[to_pos])
+                        .map(|(_, f)| f.name().as_str())
+                        .collect();
+                    if parquet_options.case_sensitive {
+                        return Err(SparkError::Internal(duplicate_parquet_field_message(
```
andygrove: This turns main's error into an `INTERNAL_ERROR`. `SparkError::Internal` goes through `SparkErrorConverter` as `SparkException.internalError`, so a requested nested duplicate now reaches the user as `[INTERNAL_ERROR] Found duplicate Parquet field name 'dup' SQLSTATE: XX000`. That tells them they've hit a bug in the engine, for a limitation `scans.md` documents. On `main` the same read raises `CometNativeException` with the plain message. On this branch the root check and the non-pruning path still do, so the same limitation surfaces as two exception classes depending on nesting. In the test log, all ten `duplicate Parquet field names fail clearly` cases, the schema-merge test and the fixture test take the `INTERNAL_ERROR` path. The tests don't notice because they only match the message text. Could the resolver raise the same error main does here? And would you add an assertion on the exception class in one of those tests, so the paths can't drift apart again?

spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala
```diff
@@ -2109,6 +2221,76 @@ abstract class ParquetReadSuite extends CometTestBase {
...
           cause.isInstanceOf[RuntimeException] &&
             cause.getMessage.contains("Found duplicate field(s)"))
+        checkDuplicateFieldIdMessage(
+          spark.read.schema(readSchema).parquet(dir.getCanonicalPath),
+          """"1": [a, rand2]""")
+      }
+    }
+  }
+
+  test("duplicate exact nested names are refused when requested and skipped otherwise") {
```
andygrove: Is there something about `duplicate-nested-names.parquet` that a Spark-written file can't reproduce? It was written by pyarrow 25.0.1, and nothing in the repo says how to regenerate it. The two reads it covers, `s.other` succeeding and `s.dup` refusing, are the pair `CometNativeReaderSuite` already covers on `main` with a file written from `named_struct('dup', id, 'dup', id + 100, 'other', id + 900)`. If the pyarrow writer is the point, could the test say why? Otherwise I'd drop the test and the binary, which also takes a little off the diff comphead was concerned about.

andygrove (commented):
Thanks for the quick turnaround on today's comments. The brackets, the drop-and-re-add tests and sunchao's shape all look right, and I checked the 3.5 and 4.0 gating against `ParquetColumnVector` in 3.5.9 and 4.0.4, which do have that check. Since your runs were on 3.5, I ran everything on the default Spark 4.1 profile, and it's all green. I also disabled each of the new guards in turn, the relabel gate, the nested and root duplicate-id errors, and the nested raise in `rewrite`, and every one of them has a test that fails without it.

On narrowing `id_duplicate_roots`, I'd keep it as it is. I measured the shape on a file with roots `a (id 1) = 1`, `a (id 2) = 2` and `b (id 3) = 3`, in both case modes. With the guard removed, Comet returns `[1, 2]` for a read of id 1 or id 2, which is two rows from a one-row file. Spark doesn't give a usable answer either. Its vectorized reader returns `[1]` for both ids, so id 2 comes back with id 1's value, and the parquet-mr reader fails outright. That's the same name-based leaf lookup #5964 describes. So the rejection protects the decoder rather than diverging from a Spark behavior we could match. Could the doc comment on `known_divergence_repeated_root_name_rejects_unambiguous_field_id` say that, and link #5964? As written it reads as a gap waiting to be closed, and closing it would bring #5783 back at the root.

comphead:
Thanks @dwsmith1983 checking this now

native/core/src/parquet/cast_column.rs
comphead: This gate is the whole #6192 fix, and it does not need a `FieldMapping`. Checking ids in the positional walk that `types_differ_only_in_field_names` already does is enough, with `use_field_id` read from `parquet_options`:

```rust
// Struct arm, per (pf, lf) pair
&& (!use_field_id || field_id(lf).is_none() || field_id(lf) == field_id(pf))
```

Separately, `positional` and `types_differ_only_in_field_names` depend only on fields fixed at construction, yet both walk the type tree on every batch. One `bool` computed in `with_parquet_options` would cover both. And since `field_mapping` is always set together with `parquet_options`, the two could be one `Option`.

native/core/src/parquet/parquet_support.rs
comphead: Returning `SparkError` from the resolver, so errors can be cached per field, changes two error classes. Here, the case-sensitive nested duplicate becomes `SparkError::Internal`, which `ShimSparkErrorConverter` turns into `SparkException.internalError`, Spark's `INTERNAL_ERROR` (SQLSTATE `XX000`). Spark reserves that for engine bugs, and `main` raises an execution error with the same text. At L426, a JNI failure while folding names is flattened into a string, so the Java exception that #5845 made propagate is lost. The tests only match message substrings, so neither shows up. Keeping `DataFusionResult` and not caching errors avoids both.

native/core/src/parquet/parquet_support.rs
```diff
@@ -162,43 +169,330 @@ impl SparkParquetOptions {
...
+pub(crate) fn spark_error(error: SparkError) -> DataFusionError {
+    DataFusionError::External(Box::new(error))
+}
+
+/// Outcome of matching one requested id or name against a struct's file fields: the last
+/// file field that matched and whether more than one did. A plain `Copy` value, so resolving
+/// a wide struct allocates nothing per id or per name; the matched names are only gathered
+/// when an ambiguity is reported.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct FieldMatch {
```
comphead: The later index that `also` records is never read, because both ambiguous branches return an error. An `Option<usize>` entry does the same in one pass without this struct, `record_field_match` or `field_match_records_ambiguity_without_allocating`. It is also all the nested duplicate-id fix needs in `match_struct_fields` on `main`:

```rust
// None marks an id that more than one file field carries
map.entry(id).and_modify(|m| *m = None).or_insert(Some(i));
// ...
(true, Some(id)) => match from_id_to_index.get(&id) {
    Some(None) => Err(/* DuplicateFieldByFieldId, matched fields as "[x, y]" */),
    index => Ok(index.copied().flatten()),
},
```

native/core/src/parquet/parquet_support.rs
```diff
@@ -162,43 +169,330 @@ impl SparkParquetOptions {
...
             )?;
             Ok(ColumnarValue::Scalar(scalar))
         }
     }
 }
 
-fn parquet_convert_array(
-    array: ArrayRef,
+/// Wrap a [`SparkError`] the way every native operator surfaces it to the JVM.
+pub(crate) fn spark_error(error: SparkError) -> DataFusionError {
```
comphead: `impl From<SparkError> for DataFusionError` in `native/common/src/error.rs` already does this, so the call sites can use `.into()`. In the same vein, `field_id` becomes `pub(crate)` here while `schema_adapter.rs` keeps its identical `parse_field_id`. One of the two could go.

native/core/src/parquet/parquet_support.rs
```diff
@@ -162,43 +169,330 @@ impl SparkParquetOptions {
...
+            None => FieldMapping::Leaf,
+        };
+        sources.push(StructFieldSource { from_index, nested });
+    }
+    Ok(FieldMapping::Struct(sources))
+}
+
+/// Convert `array` to `to_type` through its resolved `mapping`. `parent_nulls` masks the rows
+/// hidden beneath null ancestors, so only values Spark reads are checked for overflow.
+fn convert_array(
```
comphead: The renames (`parquet_convert_array_impl` to `convert_array`, `parquet_convert_struct_to_struct` to `convert_struct`, while `parquet_convert_map_to_map` keeps its name), the `convert_list` extraction, the `has_timestamp_unit` rewrite and the reflowed arms do not change behavior, but they add a lot of diff. Resolving once per file does save the per-batch `match_struct_fields` work on `main` (folded names and a `Vec` per name, though ASCII names never reach the JVM). That saving is not measured yet. Could it be its own PR, with a case added to `native/core/benches/parquet_timestamp_conversion.rs`?

native/core/src/parquet/schema_adapter.rs
```diff
@@ -1009,10 +1083,84 @@ impl PhysicalExprAdapterFactory for SparkPhysicalExprAdapterFactory {
...
+        DataType::Map(f, _) => type_holds_struct(f.data_type()),
+        _ => false,
+    }
+}
+
+/// Resolve the nested mapping of every logical field whose type holds a struct and that has
+/// a physical counterpart. Returns `None` when no field qualifies, so flat reads build
+/// nothing here. Ambiguities are kept per field rather than raised: Spark validates only
+/// the fields a read requests, and `rewrite` sees which ones those are.
+fn resolve_nested_mappings(
```
comphead: This resolves each mapping by exact name on the adapted schema, while the cast picks its physical field by `col.index()` in `replace_with_spark_cast` and by folded position in `wrap_all_type_mismatches`. The two agree only because repeated root names are rejected first. Building the mapping where `CometCastColumnExpr` is built, from its own `(input_field, target_field)`, keeps it once per file, since `rewrite` runs once per file. That removes `NestedMappings`, this function, `type_holds_struct`, `field_mapping_for`, `check_opaque_decode`, the nested block in `rewrite`, and `Leaf` standing in for lists and maps. Resolving eagerly is not needed for errors either. With #6004 declining repeated requested ids, an ambiguous read always has differing types, so it always gets a cast.

native/core/src/parquet/schema_adapter.rs
```diff
@@ -345,14 +356,37 @@ fn remap_physical_schema(
                             .with_metadata(field.metadata().clone()),
                         );
                     }
+                    return Arc::clone(field);
                 }
             }
 
+            // Shield: any remaining physical field whose name would hit an ID-bearing
```
comphead: I don't think this ordering is reachable from Spark. It only matters when an id-bearing and an id-less requested field fold to the same name, and `DataSource.resolveRelation` rejects that read schema with `COLUMN_ALREADY_EXISTS` (`checkSchemaColumnNameDuplication`, in 3.5.9, 4.0.4 and 4.1.3). In case-sensitive mode the shield only fires on identical names, which Spark rejects in any mode. By my reading, the case-sensitive Kappa read already returns `(NULL, 7)` on `main`. I would keep `main`'s order and drop the case-insensitive Kappa and placeholder tests.

native/core/src/parquet/schema_adapter.rs
```diff
@@ -1009,10 +1083,84 @@ impl PhysicalExprAdapterFactory for SparkPhysicalExprAdapterFactory {
...
     }
 }
 
+/// Per logical field name, the mapping of its nested type against its physical counterpart,
+/// or the ambiguity Spark reports for it. Only fields whose type holds a struct are listed.
+type NestedMappings = HashMap<String, Result<Arc<FieldMapping>, SparkError>>;
+
+/// Per logical field name, the `_LEGACY_ERROR_TEMP_2094` ambiguity of a root field whose id
+/// matches more than one physical root field. Only ambiguous fields are listed.
+type RootIdAmbiguities = HashMap<String, SparkError>;
```
comphead: This parallels `id_duplicate_roots`: both map a logical name to a root error that `rewrite` raises when the column is referenced. One map would do, and `rewrite` could then check each referenced column in one pass instead of three loops over `col_names`.

comphead (commented):
Thanks @dwsmith1983. I reproduced #6192 end to end on `main` at `64e98918ab` (Spark 4.1.3, native scan). A struct, an `array<struct>` element and a map value all come back by position, both when `x` is dropped and re-added and when the ids are swapped. A root-level swap and a nested swap with one extra requested field both match Spark, so the relabel shortcut is the whole cause, and the `positional` gate here fixes it.

The fix itself is a small part of the diff, and much of the rest has no user-visible effect on `main`, so I would like to split it. Only the first row below comes from a run. The others are from reading the code.

| Change | On `main` | Suggestion |
|---|---|---|
| Relabel shortcut ignores field ids (#6192) | Wrong results | Keep, as a small PR (sketch inline) |
| A requested nested id matches two file fields | Reads the first match where Spark raises `_LEGACY_ERROR_TEMP_2094` | Keep, as a small PR (sketch inline) |
| `[x, y]` brackets in the `2094` message | The shims pass `matchedFields` verbatim, so Spark's message loses them | Keep, with the previous row |
| Root duplicate id raised only for referenced columns | Fails the whole scan, even when that column is not selected | Optional follow-up that reuses `id_duplicate_roots` |
| Shield after the name match | Not reachable from Spark (inline) | Drop |
| `FieldMapping` once per file, `list_element_field`, `convert_list`, renames | No behavior change, #5681 already resolves list elements by id | Separate PR with a benchmark |

A local sketch of the first three rows is +60/-14 across `cast_column.rs`, `parquet_support.rs`, `schema_adapter.rs` and `error.rs`. It compiles and passes clippy. I have not run tests on it. It is enough because:
- `check_conversion` already walks the requested tree with `match_struct_fields` once per file, before every non-Variant `CometCastColumnExpr`, so the duplicate-id error fires at plan time.
- #6004 declines requested schemas that repeat an id. An ambiguous read therefore always has differing physical and logical types, so it always gets a cast.

<details><summary>Tests: keep, already covered on main, or tied to the new machinery</summary>

- **Keep** (these fail on `main`): `test_swapped_field_ids_bypass_relabel_shortcut`, the three `nested column dropped and re-added…` tests (merged, see inline), `requested_duplicate_field_id_errors` together with `unrequested_duplicate_field_id_reads_fine`, `parquet_duplicate_file_field_id_rejected_when_requested`, `duplicate field id inside a struct…`, the message check in `multiple id matches`, and `root_duplicate_id_raised_only_for_referenced_columns` if the root change stays.
- **Already covered on `main`**:
  - `duplicate_exact_names_are_rejected_when_requested` by `issue_5783_nested_name_duplicate`.
  - `unrequested_duplicate_siblings_do_not_block_the_read` and the Scala `duplicate exact nested names…` test by `CometNativeReaderSuite` (inline).
  - `case_insensitive_ambiguous_names_error_but_exact_match_reads` and `case_insensitive_byte_identical_children_raise_spark_duplicate_error` by #5751's `native reader duplicate nested struct fields: * (case-insensitive)`.
  - `convert_list_to_large_list_null_fills_missing_element_field` by `list_representations_preserve_missing_struct_fields`.
  - `known_divergence_repeated_root_name_rejects_unambiguous_field_id` by the `CometNativeReaderSuite` pin its own doc names. Its rationale could live in a comment on `id_duplicate_roots`.
  - `parquet_field_id_miss_null_fills_but_exact_name_sibling_still_reads` and `parquet_field_id_match_beats_stray_column_with_requested_name`, which read the same on `main` as far as I can tell.
- **Tied to the new machinery**, so they go with it: `field_match_records_ambiguity_without_allocating`, the four `resolve_mapping_*` tests, `convert_list_to_large_list_reads_elements_by_id`, `leaf_mapping_converts_list_and_map_elements`, `mapping_index_beyond_struct_children_errors`, `test_relabel_shortcut_kept_for_name_only_differences_without_ids`, `parquet_field_id_miss_case_insensitive_sibling_still_reads` and `parquet_shield_placeholder_never_folds_onto_requested_column`.

None of these can move to SQL file tests, because Spark SQL DDL cannot attach `parquet.field.id` metadata.
</details>

For context, Spark 3.5.9, 4.0.4 and 4.1.3 share the same field-id logic in `clipParquetGroupFields`. [DuckDB](https://github.com/duckdb/duckdb/blob/7bcaeeb2da0657458679bcf86f14e35379438ec4/src/common/multi_file/multi_file_column_mapper.cpp#L94-L146), [StarRocks](https://github.com/StarRocks/starrocks/blob/f132aa4ca115a75625bbc0e1228c08a87f47b323/be/src/formats/parquet/column_reader_factory.cpp#L642-L703) and iceberg-rust's `RecordBatchTransformer` also match nested fields by id at every level and resolve outside the batch loop, so resolving once per file is a sound direction for its own PR. DuckDB and StarRocks silently keep the first duplicate id, so the `2094` error is only about matching Spark.

native/core/src/parquet/schema_adapter.rs
```diff
@@ -3518,11 +3856,187 @@ mod test {
...
+            columns,
+            required_schema,
+            spark_parquet_options,
+            None,
+        )
+        .await
+    }
+
+    /// `scan_with_adapter` with column defaults for fields missing from the file.
+    async fn scan_with_defaults(
```
comphead: This repeats `scan_parquet` apart from the defaults and the writer properties. Could `scan_parquet` take those as optional arguments instead? Likewise, `field_with_id` here and in `parquet_support.rs`, and `int_field_with_id` in `cast_column.rs`, overlap the existing `id_meta` and `struct_type_with_field_id` helpers.

spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala
comphead: `CometNativeReaderSuite` already covers both halves on `main`. `duplicate Parquet field names fail clearly - *` covers the refused read (struct, array element and map value, at two batch sizes), and `duplicate Parquet field names - exact-name projection works in both resolver modes` covers the unique sibling. Those tests write the file with `named_struct`, so I think this test and `duplicate-nested-names.parquet` can both go.

spark/src/test/scala/org/apache/comet/parquet/ParquetReadSuite.scala
```diff
@@ -2079,10 +2079,122 @@ abstract class ParquetReadSuite extends CometTestBase {
...
+  // The shape a schema evolution leaves behind: a nested column dropped and added back under
+  // its old name gets a fresh field id, so the file holds `struct<x (id 1), y (id 2)>` while
+  // the table reads `struct<x (id 3), y (id 2)>`. The names line up at every position, which
+  // is exactly what the metadata-only relabel shortcut in the native cast looks for, so without
+  // the field-mapping gate the scan hands back the old x values. Spark returns null for x and
+  // keeps y. One test per nesting, so a wrong answer names the level that produced it.
+  // Before Spark 4.1 the vectorized reader raises on this read below a list or map, since its
+  // column vector rejects the placeholder field the clipped schema carries for the unmatched
+  // id, so the comparison with Spark runs from 4.1 on. The pinned rows hold everywhere.
+  private def checkDroppedAndReAddedFieldId(
```
comphead: These pin #6192 well. Could the three tests become one table-driven test over `s`, `l` and `m`, with a swapped-id row (`x (id 2), y (id 1)`) added? The swapped case is the other shape in the issue, and right now only Rust covers it.

PR #5665, opened 2026-09-03
fix(celeborn): reject unsafe native push completion tracking

### LINKED ISSUE #5528: Native Celeborn shuffle: reflectively replacing Celeborn's final fields can release push admission while payloads are in flight
Following up on PR #5513 (part of #5352). `CelebornTransportCallbackTracker` installs its push-ownership instrumentation by reflectively assigning four fields on live Celeborn objects, and all four are declared `private final` in the Celeborn releases the PR targets.

The four are `TransportClientFactory.clientBootstraps`, `TransportClient.channel`, `TransportResponseHandler.outstandingPushes`, and `ShuffleClientImpl.pushDataRetryPool`. I checked the declarations in both v0.6.0 and v0.7.0 and they match in each:

```
TransportClientFactory.java:79   private final List<TransportClientBootstrap> clientBootstraps;
TransportClient.java:81          private final Channel channel;
TransportResponseHandler.java:60 private final ConcurrentHashMap<Long, PushRequestInfo> outstandingPushes;
ShuffleClientImpl.java:131       private final ExecutorService pushDataRetryPool;
```

`setAccessible(true)` makes the write legal for a non-static final field, so this is not an access problem. The issue is visibility. JLS 17.5.3 gives no guarantee that a final field modified after construction becomes visible to another thread unless that thread synchronizes with the writer. Comet writes under the transport factory monitor, the response handler monitor, or the shuffle client monitor, and Celeborn's own reads of these fields on the Netty event loops and on other tasks' pushing threads take none of those. A thread that has no happens-before edge, or a JIT that has already trusted and folded the final field load, keeps using the original channel, the original `outstandingPushes` map, and the original retry pool.

What concerns me is the direction the failure takes. If a push's callback is never wrapped because some thread still sees the original request map, then no `Lease` is retained for it, `Push.owners` reaches zero at `close()`, `isComplete()` returns true, and the pusher releases its admission bytes while the payload is still in flight. That is precisely the over-release the tracker was added to prevent, and it happens silently in the direction of memory overcommit rather than of a stall.

I do not think this is reachable by any test currently in the tree. The 46 tests in `CelebornShufflePartitionPusherSuite` all drive hand-written single-threaded Scala stand-ins, so there is never a concurrently running Celeborn reader racing the reflective write.

Is there a way to get the same push-completion signal without mutating another project's final fields? Two ideas that come to mind are wrapping the client object that Comet hands to the native writer so Comet controls the push entry point, or asking Celeborn upstream for a supported push-completion listener hook so this does not depend on private internals at all. If neither is workable soon, would it be worth documenting the reliance explicitly and pinning the supported Celeborn versions?

pingzh:

Closes #5528.

## Rationale for this change

Replacing final fields on live Celeborn clients does not guarantee that transport and retry threads observe Comet's ownership hooks. Admission can then be released while payloads remain in flight. Push-state counters alone do not prove that callbacks, retries, and timed-out writes have released those payloads.

- Require volatile instance fields for shared transport hook replacements, and reject incompatible clients before native submission or admission registration.
- Keep unsupported clients on ordinary Spark/Celeborn shuffle during planning, including collect-limit and top-K. Released Celeborn 0.6.0 and 0.7.0 now take this path even when native shuffle is requested; preserving native support requires a safe completion API.
- Add rejection and fallback regressions, model safely published fields in the ownership test doubles, and document the limitation.

- 135 tests passed across `CelebornShufflePartitionPusherSuite`, `CometCelebornShuffleManagerSuite`, and `CometCelebornShufflePlanningSuite` in a root reactor build with Spark 4.1.3 / Scala 2.13.17. Coverage includes retained references on another thread and Spark fallback queries with AQE enabled and disabled.
- `CelebornReflectionCompatibilitySuite` passed all 5 tests against each released 0.6.0 and 0.7.0 JAR. These suites were compiled and run separately, including production Java compilation with `--release 11`.
- Native build, root Maven packaging for Spark 4.1, Spotless formatting/checks, and `git diff --check` passed.

### DIFF
--- a/docs/source/user-guide/latest/tuning.md
+++ b/docs/source/user-guide/latest/tuning.md
@@ -254,8 +254,14 @@ even when both its parent and child are non-Comet operators.
 
 ### Remote Shuffle with Celeborn
 
-Applications using Apache Celeborn can opt into Comet's native shuffle writer and reader with
-the composite shuffle manager:
+Applications using Apache Celeborn can use Comet's composite shuffle manager to retain ordinary
+Spark/Celeborn shuffle while accelerating other operators with Comet.
+
+Native shuffle requires reliable completion tracking for in-flight payloads. Released Celeborn
+0.6.0 and 0.7.0 clients do not provide the required guarantee, so these versions retain ordinary
+Spark/Celeborn shuffle even when `spark.comet.shuffle.mode=native`. Native shuffle support for
+these clients requires a safe Celeborn push-completion API. The following settings request
+native shuffle when the client passes Comet's compatibility checks:
 
 ```properties
 spark.shuffle.manager=org.apache.spark.sql.comet.execution.shuffle.CometCelebornShuffleManager
--- a/spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
+++ b/spark/src/main/java/org/apache/comet/shuffle/CelebornShufflePartitionPusher.java
@@ -234,6 +234,13 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
           "Celeborn in-flight byte limit must fit all copies of one Comet frame");
     }
 
+    String unavailable = nativePushCompletionUnavailableReason(shuffleClient.getClass());
+    if (unavailable != null) {
+      // Do not silently use PushState counters: callbacks, retries and timed-out writes can
+      // still retain payloads after those counters report completion.
+      throw new UnsupportedOperationException(unavailable);
+    }
+
     final Method pushMethod;
     try {
       pushMethod =
@@ -345,8 +352,8 @@ public final class CelebornShufflePartitionPusher implements ShufflePartitionPus
         // Wrappers need not provide the optional side-effect-free state lookup.
       }
     } catch (NoSuchMethodException missing) {
-      // Compatibility-only clients can still push synchronously. Stock Celeborn 0.6 and 0.7
-      // expose getPushState and use completion-backed admission.
+      // Compatibility-only clients can still push synchronously. Asynchronous clients must
+      // expose getPushState in addition to safely published transport hooks.
       pushStateMethod = null;
       pushStatesField = null;
       trackerField = null;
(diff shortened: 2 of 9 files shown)


sunchao (approved):
Reviewed `6540fa0aa77575e92caa9f846a41a1eb02a4fc82` against `90c1dd96c89895bd9732d64055e4919b81a8635d`. No actionable P1/P2 findings.

### Design and compatibility

Rejecting unsupported completion tracking before admission registration addresses #5528 without treating push-state counters as proof that payloads are released. The planner uses the same capability decision for ordinary exchanges, collect-limit and top-K, preserving ordinary Spark/Celeborn dependencies on fallback. Actual Celeborn 0.6.0 and 0.7.0 clients have final, non-volatile hook fields and are rejected. This does not certify arbitrary future client implementations solely because their fields are volatile.

### Performance

Capability detection is captured per manager, with no new per-row or per-frame reflection loop. Released 0.6.0/0.7.0 clients intentionally lose native shuffle acceleration, including when native mode is requested. That is a documented safety tradeoff. No throughput or memory benchmark was run, and no speedup is claimed.

### Abstraction and complexity

Reusing one rejection reason and the existing planner gate keeps the decision consistent without introducing a new completion framework. The injected capability supplier supports testing the manager boundary. A safe upstream completion API remains the route to restoring native support for released clients.

### Validation and CI

The focused review passed 53 selected pusher tests, five compatibility tests against each released JAR, and 12 field-shape rejection checks. I independently reran the 12 admission checks. The new rejection regression fails on the base and passes on this head. Four Spark-dependent tests were omitted from the local selection. Separately, observed Spark 3.5 merge CI passed 383 tests, including the new AQE-on/off fallback queries.

No full local Spark/JNI run, live Celeborn service or Java 11 runtime/API validation was performed. Local compilation used Java 21 with source/target 11.

CI is not complete: the latest snapshot has 48 successful checks, seven skipped, 16 running and one failed. The [TPC-DS verification job](https://github.com/apache/datafusion-comet/actions/runs/33786100414/job/100762118143) stopped while downloading Maven with HTTP 429, before result verification. I did not rerun or approve CI.

sunchao:
Merged, thanks @pingzh !

Issue #4576, opened 2026-06-02
Investigate adopting DataFusion's allocator-level memory accounting to replace manual memory tuning


andygrove:
## What is the problem the feature request solves?

Comet's native memory accounting relies on DataFusion's `MemoryPool`, which is *voluntary*: it only counts bytes that an operator explicitly reserves via `reserve()` / `try_grow()`. Allocations that bypass the pool (Arrow buffers built during predicate evaluation, intermediate scratch in joins, allocations inside expression kernels) are invisible to the accounting. Comet's `CometUnifiedMemoryPool` forwards these reservations across JNI to Spark's `TaskMemoryManager`, so the undercount propagates up to Spark as well.

This is a known, documented limitation. From the tuning guide:

> Comet's memory accounting isn't 100% accurate and this can result in Comet using more memory than it reserves, leading to out-of-memory exceptions. To work around this issue, it is possible to set `spark.comet.exec.memoryPool.fraction` to a value less than `1.0`...

`spark.comet.exec.memoryPool.fraction` is effectively a manual fudge factor that compensates for an unknown undercount margin, tuned per workload. We would like to move away from this manual and unreliable accounting toward something backed by real, measured memory usage.

The Comet roadmap already calls this out:

> Improving memory accounting, reservation strategies, and spill integration will reduce out-of-memory errors and allow Comet to make better use of available resources, especially in multi-query and multi-task environments.

## Describe the potential solution

Two complementary mechanisms, both backed by **real allocator stats** rather than voluntary reservations:

### 1. Online accounting / observability

DataFusion PR [apache/datafusion#22626](https://github.com/apache/datafusion/pull/22626) introduces allocator-level accounting that measures actual heap usage independent of voluntary pool reservations. Comet already does the *offline* version of exactly this: `native/common/src/bin/analyze_trace.rs` "compares jemalloc usage against the sum of per-thread Comet memory pool reservations [and] reports any points where jemalloc exceeds the total pool size." The proposal is to make this an online signal so the divergence can drive behavior (and ultimately retire the `memoryPool.fraction` heuristic).

### 2. A process-level RSS circuit breaker (OomGuard)

Wrap the global allocator to maintain a running estimate of memory headroom, periodically resynced from real resident memory. The proposed design has a few notable properties:

[...]

PR #5666, opened 2026-09-03
feat: experiment with `RealUsagePool`

comphead:

Experiment for #4576 .

Adopted from #4582

### DIFF
--- a/docs/source/user-guide/latest/tuning.md
+++ b/docs/source/user-guide/latest/tuning.md
@@ -60,8 +60,9 @@ The valid pool types are:
 
 - `fair_unified` (default when `spark.memory.offHeap.enabled=true` is set)
 - `greedy_unified`
+- `unbounded`
 
-Both pool types are shared across all native execution contexts within the same Spark task. When
+The two `unified` pool types are shared across all native execution contexts within the same Spark task. When
 Comet executes a shuffle, it runs two native execution contexts concurrently (e.g. one for
 pre-shuffle operators and one for the shuffle writer). The shared pool ensures that the combined
 memory usage stays within the per-task limit.
@@ -73,6 +74,12 @@ when there is sufficient memory in order to leave enough memory for other operat
 
 The `greedy_unified` pool type implements a greedy first-come first-serve limit. This pool works well for queries that do not
 need to spill or have a single spillable operator.
+
+The `unbounded` pool does no accounting of its own and imposes no limit, so Comet's native memory is capped only by
+the experimental `spark.comet.exec.memoryGuard.enabled`. Enabling that guard in off-heap mode selects `unbounded`
+automatically and ignores this setting: growth is then gated on real allocator usage against the off-heap budget
+rather than on Spark's per-task reservations. Note that Spark's `TaskMemoryManager` no longer sees Comet's off-heap
+usage in that mode, so Spark cannot ask Comet to spill on behalf of its own operators.
 
 [shuffle]: #shuffle
 [Advanced Memory Tuning]: #advanced-memory-tuning
--- a/native/core/Cargo.toml
+++ b/native/core/Cargo.toml
@@ -103,9 +103,16 @@ datafusion-functions-nested = { version = "55.0.0" }
 
 [features]
 backtrace = ["datafusion/backtrace"]
-default = ["hdfs-opendal"]
+default = ["hdfs-opendal", "oom-guard"]
 hdfs-opendal = ["opendal", "object_store_opendal", "hdfs-sys"]
 jemalloc = ["tikv-jemallocator", "tikv-jemalloc-ctl"]
+
+# Allocator-level OOM circuit breaker. Wraps the global allocator to track real
+# allocated bytes and gate/abort over-budget query-worker threads. Enabled by default
+# so `spark.comet.exec.memoryGuard.*` and the `real_usage` memory pool work without a
+# special build; an idle guard is near-free (tracking stays off until a task arms it).
+# Drop it from `default` for a bare allocator.
+oom-guard = []
 # Delta Lake integration. When enabled, links the `comet-contrib-delta` crate
 # into `libcomet` and activates the `OpStruct::DeltaScan` dispatcher arm.
 # Default builds carry zero Delta surface.
--- a/native/core/Cargo.toml
+++ b/native/core/Cargo.toml
@@ -109,8 +109,8 @@ jemalloc = ["tikv-jemallocator", "tikv-jemalloc-ctl"]
 
 # Allocator-level OOM circuit breaker. Wraps the global allocator to track real
 # allocated bytes and gate/abort over-budget query-worker threads. Enabled by default
-# so `spark.comet.exec.memoryGuard.*` and the `real_usage` memory pool work without a
-# special build; an idle guard is near-free (tracking stays off until a task arms it).
+# so `spark.comet.exec.memoryGuard.*` works without a special build; an idle guard is
+# near-free (tracking stays off until a task arms it).
 # Drop it from `default` for a bare allocator.
 oom-guard = []
 # Delta Lake integration. When enabled, links the `comet-contrib-delta` crate
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -110,7 +110,9 @@ use crate::execution::tracing::{
 
 use crate::execution::memory_pools::logging_pool::LoggingMemoryPool;
 #[cfg(feature = "oom-guard")]
-use crate::execution::memory_pools::{oom_guard, MemoryPoolType, RealUsagePool};
+use crate::execution::memory_pools::{
+    oom_guard, MemoryPoolConfig, MemoryPoolType, RealUsageMemoryPool,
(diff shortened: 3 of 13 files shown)


comphead:
Tested on 
```
/opt/homebrew/Cellar/apache-spark/3.5.5/bin/spark-submit  --master="local[*]"    --conf spark.driver.memory=8G     --conf spark.executor.instances=1     --conf spark.executor.cores=8     --conf spark.cores.max=8     --conf spark.executor.memory=16g     --conf spark.memory.offHeap.enabled=true     --conf spark.memory.offHeap.size=16g --conf spark.comet.explainFallback.enabled=true --conf spark.comet.logFallbackReasons.enabled=true --conf spark.eventLog.enabled=false     --jars $COMET_JAR     --driver-class-path $COMET_JAR     --conf spark.driver.extraClassPath=$COMET_JAR     --conf spark.executor.extraClassPath=$COMET_JAR     --conf spark.plugins=org.apache.spark.CometPlugin     --conf spark.shuffle.manager=org.apache.spark.sql.comet.execution.shuffle.CometShuffleManager     --conf spark.comet.enabled=true     --conf spark.comet.exec.shuffle.enableFastEncoding=true      --conf spark.comet.cast.allowIncompatible=true   --conf spark.comet.exec.memoryPool=real_usage ../datafusion-benchmarks/runners/datafusion-comet/tpcbench.py     --benchmark tpcds     --data ../datafusion-benchmarks/tpcds/data_parquet/     --queries ../datafusion-benchmarks/tpcds/queries-spark  --query 67 --output . --iterations 1
```

Before PR the Q67 failed on 
```
26/09/02 09:12:13 WARN TaskSetManager: Lost task 110.0 in stage 29.0 (TID 261) (192.168.4.142 executor driver): TaskKilled (Stage cancelled: Job aborted due to stage failure: Task 109 in stage 29.0 failed 1 times, most recent failure: Lost task 109.0 in stage 29.0 (TID 260) (192.168.4.142 executor driver): org.apache.comet.CometNativeException: Additional allocation failed for ExternalSorterMerge[0] with top memory consumers (across reservations) as:
  ExternalSorter[0]#1106(can spill: true) consumed 1024.0 MB, peak 1058.2 MB,
  ExternalSorterMerge[0]#1107(can spill: false) consumed 456.0 B, peak 10.0 MB,
  ShuffleRepartitioner[0]#1109(can spill: true) consumed 0.0 B, peak 0.0 B.
Error: Failed to acquire 228842 bytes, only got 11829 bytes. Reserved: 1073729995 bytes
```

After PR the query passed

andygrove (commented):
Thanks for picking this up @comphead. I spent some time with this and with the allocator accounting in #5934, and my overall take is that this is a large and complex PR and I think we need to step back and build this incrementally. Perhaps we could start with adding the accounting allocator as proposed in #5934, and then layer the gate and any enforcement on top of that as separate PRs? That would give us one balance and one wrapper to reason about, and it would let us settle a few design questions below one at a time rather than all at once.

Detailed notes follow. Some of them come directly from what I learned validating #5934 on TPC-H SF100, so I have tried to include the numbers.

## The breaker panics inside the allocator

`oom_guard::track` calls `panic_any` from inside `GlobalAlloc::alloc`, `alloc_zeroed` and `realloc`. The `GlobalAlloc` docs in `core/src/alloc/global.rs` say:

> It's undefined behavior if global allocators unwind. This restriction may be lifted in the future, but currently a panic from any of these functions may lead to memory unsafety.

The `catch_unwind` scaffolding in `jni_api.rs` and the `UNWINDING` flag make the unwind land somewhere sensible when it works, but the "failed to initiate panic, exit 133" abort described in the comment above the `ARMED` CAS looks like this contract being hit. The CAS reduces how often it happens rather than making it sound. I think the enforcement point has to move out of the allocator. The gate in `try_grow` is already a safe enforcement point, and the `executePlan` poll loop already has an every-100-polls hook where a balance check could return a `ResourcesExhausted` error for the current task without unwinding through arbitrary code. That loses the "any allocation anywhere trips it" property, but it keeps the process alive, which is the whole point.

## What the gate can and cannot see

From the SF100 TPC-H traces on #5934, pool reservations summed to near zero at points where the balance was over 1 GB. Q8 was 2011 MB real against 428 MB reserved, Q16 was 1192 MB against 0, Q17 was 1292 MB against 34 MB. Most native bytes at peak are scan buffers, shuffle and kernel intermediates that never call `try_grow`. Since `RealUsageMemoryPool` can only reject reserving operators, a task pushed over budget by unreserved allocation gets its sort or aggregate rejected, spills, retries, and is rejected again because spilling freed little. That ends in `ResourcesExhausted` for a task that was not the one consuming the memory, and the breaker does the same on an even more arbitrary thread. The Q67 result is encouraging, but it is one query that happens to have a large spillable sort. I think the design needs to say what the expected outcome is in the common case where the over-budget bytes are not behind a reservation.

## Dealloc ordering

`dealloc` in `oom_guard.rs` calls `inner.dealloc` first and subtracts afterwards. jemalloc drops its `stats.allocated` at the start of a large free and then eagerly unmaps anything over its 8 MiB oversize threshold, which takes milliseconds for a 100+ MB block, so the balance carries a block the allocator has already released for that whole window. This was the root cause of the `native_allocated > jemalloc_allocated` samples on #5934 (up to 160 MB high, about 2% of samples, all during task teardown) and the same code is here. On #5934 it was cosmetic. Here the balance is enforced, so another thread's `try_grow` can be rejected, or the breaker can trip, on a value that is 100+ MB high. A free cannot fail, so the subtraction can move before the delegation, which is what ce9b13a did on #5934. The reasoning in the `realloc` comment about accounting before delegating is right for realloc and does not apply to dealloc.

## Thread-exit drift

`LOCAL_DRIFT` is a bare `Cell<isize>`, so up to 64 KiB per dying thread is discarded, in either direction. Tokio's blocking pool churns on its idle timeout, so on a long-lived executor the enforced balance acquires a slow bias. #5934 handles this with a `ThreadDrift` destructor plus the re-entrancy flag and `try_with` needed to make a destructor-bearing thread-local safe inside an allocator. That carries over directly if this builds on that module.

## Overlap with #5934 and the default feature

Both PRs install a wrapper around the same allocator selection in `lib.rs` with different feature names and different cfgs, and they will conflict on merge. #5934's `alloc_accounting` module is the same accounting with the two fixes above already applied, and its `current_balance()` is exactly what `BalanceSource::Live` needs. This PR also adds `oom-guard` to the `default` feature, so every build carries the wrapper with a relaxed load per allocation when idle. That cost is probably small but it is unmeasured, and the #5934 branch has an `alloc_overhead` bench that could answer it. For reference, with tracking fully on the cost on TPC-H SF100 was 1.4% overall and about 3% on Q21. #5934 is opt-in, this PR is always-on, and I would rather make that decision once.

## Ceiling semantics

The gate's ceiling and the breaker's default limit are both `memory_limit`, which in off-heap mode is `spark.memory.offHeap.size` scaled by `spark.comet.exec.memoryPool.fraction`. The `memoryGuard.size` doc says it defaults to the off-heap size, which is only true when the fraction is 1.0. Separately, the balance excludes everything the JVM allocates off-heap through `Unsafe`, and with `unbounded` Spark's `TaskMemoryManager` no longer sees Comet's usage. Both consumers can then each approach the full off-heap size independently, so the executor's real off-heap footprint is bounded by roughly twice the budget rather than once. The tuning.md paragraph hints at the second half of this but not the consequence. Is the intent that operators set `memoryGuard.size` to a container limit? If so the docs should say that.

## Fair share

`has_per_task_budget` returns false for `Unbounded`, and in off-heap mode the guard always forces `Unbounded`, so `fair_share` is always `None` and the gate is first-come. The comments in `createPlan` and `try_grow` describe fair share as if it applies in that path. It only applies in on-heap mode, where the divisor falls back to `executor_cores` because the greedy per-task pools keep no registry. Could the comments be scoped accordingly, or is there a plan to make it apply?

## Smaller items

- Exposing `unbounded` as a user-selectable off-heap pool type in `tuning.md` and the `memoryPool` doc means a user can remove all limits by setting one config without the guard. If it exists only so the guard can force it, it does not need to be a documented choice.
- `arm` runs on every `createPlan` and overwrites the process-wide `LIMIT`. After a trip, `ARMED` stays false until the next `createPlan` anywhere, so the breaker is off for every other task in that window.
- `stamp_current_thread` covers tokio workers and the JNI caller thread. Allocations on the blocking pool are tracked but never enforced. The prototype listed that as a follow-up and it is worth stating here too.
- The repro in the comment above uses `spark.comet.exec.memoryPool=real_usage`, which is not a value this diff accepts. The tested configuration in the description should match the code.

## Tests

It might be worth adding a JVM-level test that enables the guard with a small `memoryGuard.size` and runs a sort or aggregate to confirm the spill-and-succeed path, and one with an impossible budget to confirm a retriable failure rather than a hang or a process abort. A unit test for the `4g` size resolution added to `serializeCometSQLConfs` would also help. The unit tests that are here cover the pure helpers well.

comphead:
Thanks @andygrove I'll keep this PR draft if you wanna refer to it when adopting native allocator accounting in parallel track

PR #5667, opened 2026-09-03
perf: slice the child instead of gathering it when unnesting

### LINKED ISSUE #5685: Comet's explode is slower than Spark when the element is a struct, worst on nested rows

`CometExplodeExec` is slower than Spark's `GenerateExec` whenever the exploded element is a struct, and the gap is largest on the deeply nested, wide row shapes that motivate a generator in the first place.

Measured with `CometExplodeBenchmark` (added in #5381) on an Apple M3 Max, JDK 17, `local[1]`, Spark 4.1 / Scala 2.13, release native build. These are whole-query times including the Parquet scan and the counting aggregate, so they understate the operator's share (Best ms):

| Case | Spark | Comet | Relative |
| ---- | ----- | ----- | -------- |
| `explode array<struct<a bigint, b string>>[10]` | 61 | 86 | 0.7X |
| `explode array<struct>[4]` at depth 1 | 69 | 156 | 0.4X |
| `explode array<struct>[4]` at depth 8 | 84 | 167 | 0.5X |
| `explode array<struct>[4]` at depth 8, then its inner array | 96 | 181 | 0.5X |

For contrast, over `array<bigint>` at fan-out 10 Comet is 1.2X faster, and at fan-out 2 it is 1.5X faster. So this is not a general generator problem — it tracks the element type.

The nesting itself is cheap. Going from depth 1 to depth 8 costs Comet 11ms and Spark 15ms, and nested schema pruning narrows the scan to exactly the events path either way, so both depths read the same leaf payload. What costs is the element being a wide struct: the depth-8 element here is `struct<platform string, entries array<struct<type string, ts bigint, page string, source string>>>`, and the plain `array<struct<bigint, string>>` case shows the same direction at a smaller magnitude.

The nested cases are the most reproducible numbers in the whole benchmark — across five runs Comet's arm varied by a few ms while Spark's swung under machine contention.

`ExplodeExec` is Comet's fork of DataFusion's `UnnestExec`, and it gathers the unnested column through a take-index buffer sized to the output. For a struct element that gather recurses into every field, which is where the cost concentrates.

#5667 is in flight and replaces the gather with a slice of the child when the take indices would be a contiguous run, which covers plain `explode` and both columns of `posexplode`. Its Rust-level benchmark shows `explode_element_type/struct` down 91%, so it should close a large part of this. This issue is for whatever remains after that lands, and for the cases #5667 deliberately does not reach:

- `explode_outer` / `posexplode_outer` still gather, because a NULL or empty row is padded with a NULL that no slice of the child contains.
- Carried passthrough columns are still a real gather.
- Whether a struct element benefits from slicing as much as a primitive one needs measuring rather than assuming, since `ListArray::slice` leaves `values` whole and a struct's field arrays are sliced independently.

## Additional context

The benchmark's nested group deliberately counts the whole event struct rather than one of its fields, so nested schema pruning cannot narrow the exploded element and leave the case measuring a two-column gather.

A generator over a `map` cannot be measured against Comet at all yet — see #2837 — so the shape a real schema would use for this data (a map keyed by platform) is an `array<struct<platform, entries>>` in the benchmark.

andygrove:

Closes #5685. That issue reports `explode` over a struct element being slower than Spark's
`GenerateExec`, worst on nested rows. Its four cases now run at 1.1X to 1.3X. Getting there took
two things: #5262 (DataFusion 55) fixed the nested Parquet scan that dominated the nested cases,
and this closes what was left in the operator. Neither does it alone: on merged `main` with #5262
in, the four cases still sit at 1.1X, 0.8X, 1.0X and 0.7X. See the measurements below.

## Rationale for this change

`ExplodeExec` is a fork of DataFusion's `UnnestExec`, and it inherited kernels written to be
generic over every list type DataFusion supports. Comet plans exactly one shape: a `List`,
depth 1, one column for `explode` and two for `posexplode`. Three costs in the general path are
avoidable at that shape.

The largest is in `unnest_list_array`. Unnesting a single list column pads nothing, so the take
indices it builds: one `i64` per output element, are the contiguous run
`offsets.first()..offsets.last()`, and the gather through them reads the child straight through
in order. Rows of a `ListArray` are adjacent by construction, since row `i` is
`[offsets[i], offsets[i + 1])`, so the answer is already sitting in the child as one slice. The
operator was allocating an index buffer the size of its output and copying every element to
reproduce it.

Two smaller ones. `predict_output_lens` derives per-row lengths through `find_longest_length`,
which chains `length`, `cast`, `is_not_null` and `zip` to stay generic: four allocating passes
to subtract adjacent offsets, once per input batch, because `length` returns `Int32` for `List`
and NULL rows need substituting. And `create_take_indices` appended the repeat indices one
element at a time through a builder that has no validity to track.

`unnest_list_array` returns a slice of the child when the gather would be a contiguous run.
Comet takes this for plain `explode`, and for both columns of `posexplode`, whose position array
is built with the same per-row lengths and null mask. `explode_outer` still gathers: a NULL or
empty row is padded with a NULL that no slice of the child contains.

The guard is not just arithmetic. Checking that the offset span equals the capacity is not
sufficient, because Arrow permits a NULL list slot to span elements: the gather skips those and
a slice would include them, and such a row can cancel out padding elsewhere and leave the totals
agreeing with a run that is not the one to take. `populated_null_row_falls_back_even_when_the_totals_agree`
pins that case. The view types are rejected outright, since their per-row offsets are independent
and need not be ordered.

Worth noting for the guard's reach: it accepts NULL rows whose range is empty, which is what
builders and the Parquet reader emit. That matters because `InferFiltersFromGenerate` puts
`size(arr) > 0 AND arr IS NOT NULL` below every non-outer generator over a column, so a plain
`explode` in production is usually handed an array column with the NULL rows already filtered
out, but when it is not, the rows still contribute nothing and the run still holds.

`predict_output_lens` computes lengths in one pass for the single-`List` case and keeps
`find_longest_length` as the fallback. `create_take_indices` fills the buffer per run.

The module header said a change to the forked region "either belongs upstream or does not belong
at all". This PR deliberately reverses that: the specializations are for the shape Comet plans,
not upstream's, and the header now says so and says what retiring the fork would cost.

### Trade-off

Output batches now alias the input's child buffer rather than owning a compacted copy, and
`ListArray::slice` leaves `values` whole, so the alias is to the child of the whole input batch,
not of the chunk. All chunks from one input batch share that buffer and between them fill it, so
there is no amplification when they are all retained; a downstream operator that keeps only some
of them pins all of it. That is bounded by one input batch's expansion, which `pending_input`
already holds materialized, and `BatchSplitStream` above it already hands out slices of the
input.

Eight new unit tests in `explode.rs` cover the fast path: child aliasing, a non-zero offset base
from slicing, empty and dropped-NULL rows, the padded fallback, the populated-NULL-row case
above, and view input, plus equivalence of the fused length computation against
`find_longest_length` across NULL handling, empty batches, and slicing. The existing chunking
tests already assert that chunked output matches unchunked output exactly, which is what would
catch a fast path that fired when it should not have.

`CometGenerateExecSuite` (37 tests) and `CometExecSuite` (144 tests) pass.

### Measurements

`cargo bench --bench explode`, added in the first commit, on an Apple M3 Ultra:

| Case | Change |
| ---- | ------ |
| `explode_fan_out/2` | −72% |
| `explode_fan_out/10` | −80% |
| `explode_fan_out/100` | −86% |
| `explode_element_type/bigint` | −80% |
| `explode_element_type/string` | −90% |
| `explode_element_type/struct` | −91% |
| `explode_carried_columns/0` | −80% |
| `explode_carried_columns/3` | −59% |
| `explode_outer_with_nulls/bigint` | −12% |
| `explode_outer_with_nulls/string` | −5% |

`CometExplodeBenchmark` from #5381, this branch against merged `main`, three runs of each arm
interleaved on one machine: Apple M3 Ultra, JDK 17, `local[1]`, Spark 4.1 / Scala 2.13, release
native build. These are whole-query times including the Parquet scan and the counting aggregate,
so they are the diluted view (best of the three Best ms):

| Case | Spark before / after | Comet before / after | Relative before / after |
| ---- | -------------------- | -------------------- | ----------------------- |
| fan-out 2 | 30 / 29 | 18 / 17 | 1.7X / 1.7X |
| fan-out 10 | 38 / 37 | 29 / 24 | 1.3X / 1.5X |
| fan-out 100 | 125 / 105 | 148 / 110 | 0.8X / 1.0X |
| explode | 38 / 36 | 27 / 24 | 1.4X / 1.5X |
| posexplode | 38 / 37 | 33 / 26 | 1.2X / 1.4X |
| explode_outer | 39 / 38 | 32 / 31 | 1.2X / 1.2X |
| posexplode_outer | 41 / 41 | 41 / 41 | 1.0X / 1.0X |
| element bigint | 37 / 35 | 27 / 23 | 1.4X / 1.5X |
| element string | 51 / 50 | 38 / 29 | 1.3X / 1.7X |
| element struct | 57 / 57 | 53 / 43 | 1.1X / 1.3X |
| nested, depth 1 | 63 / 63 | 77 / 59 | 0.8X / 1.1X |
| nested, depth 8 | 77 / 76 | 79 / 60 | 1.0X / 1.3X |
| nested, depth 8 then inner array | 82 / 81 | 112 / 74 | 0.7X / 1.1X |
| explode alone | 46 / 47 | 34 / 30 | 1.4X / 1.6X |
| plus 3 carried columns | 48 / 48 | 44 / 40 | 1.1X / 1.2X |

The Spark arm is the control. Across all six runs its per-case spread is at most 4ms except
`fan-out 100`, where one branch run came in at 105 against 125 to 131 everywhere else, and
`element string`, which is bimodal at 50 or 58 in both arms. Comet's spread is at most 4ms in
every case, so the movement in the table is the change and not the machine. Read the `fan-out 100`
row as Comet 148 to 110 against a Spark arm that is somewhere in 105 to 131: it goes from clearly
behind to level, and the exact multiple is the noisy part.

No case is behind Spark after this. The four cases #5685 names go from 1.1X, 0.8X, 1.0X and 0.7X
to 1.3X, 1.1X, 1.3X and 1.1X.

The three nested cases are the ones #5262 also moved: Comet's arm on them was 114, 117 and 155ms
before that landed, so the scan fix took them to 77, 79 and 112, and this takes them to 59, 60
and 74. Both were needed. What is left in those cases is the carried-column gather and the
aggregate over the expanded batch, neither of which is now the reason they trail anything.

### DIFF
--- a/native/core/Cargo.toml
+++ b/native/core/Cargo.toml
@@ -122,3 +122,7 @@ crate-type = ["cdylib", "rlib"]
 [[bench]]
 name = "array_element_append"
 harness = false
+
+[[bench]]
+name = "explode"
+harness = false
--- a/native/core/benches/explode.rs
+++ b/native/core/benches/explode.rs
@@ -1,0 +1,222 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Micro-benchmarks for `ExplodeExec`, the operator behind Spark's `explode` and `posexplode`.
+//!
+//! `CometExplodeBenchmark` on the JVM side measures the same operator end to end, where the
+//! Parquet scan and the counting aggregate are a large share of the total. This one runs the
+//! operator over in-memory batches so a change to the unnesting kernels shows up undiluted.
+//!
+//! The dimensions are the ones that drive its cost: how far each row fans out, the element type
+//! being unnested, how many columns are replicated alongside the generated one, and whether the
+//! input has the NULL rows that force outer semantics.
+
+use std::sync::Arc;
+
+use arrow::array::{Array, ArrayRef, Int64Array, ListArray, StringArray, StructArray};
+use arrow::buffer::{NullBuffer, OffsetBuffer};
+use arrow::datatypes::{DataType, Field, Fields, Schema, SchemaRef};
+use arrow::record_batch::RecordBatch;
+use comet::execution::operators::ExplodeExec;
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use datafusion::common::UnnestOptions;
+use datafusion::datasource::memory::MemorySourceConfig;
+use datafusion::execution::TaskContext;
+use datafusion::physical_plan::unnest::ListUnnest;
+use datafusion::physical_plan::{common::collect, ExecutionPlan};
+use datafusion::prelude::SessionConfig;
+use tokio::runtime::Runtime;
+
+/// Input rows per batch, and batches per run. 8192 is DataFusion's default `batch_size`, so the
+/// operator chunks the input rather than seeing it whole.
+const ROWS_PER_BATCH: usize = 8192;
+const BATCHES: usize = 8;
+
+/// The element types worth distinguishing: a fixed-width primitive, a variable-width type whose
+/// gather has to rebuild offsets and copy bytes, and a nested type that gathers per field.
+#[derive(Clone, Copy, PartialEq)]
+enum Element {
+    Int64,
+    Utf8,
+    Struct,
+}
+
+impl Element {
+    fn name(self) -> &'static str {
+        match self {
+            Element::Int64 => "bigint",
+            Element::Utf8 => "string",
+            Element::Struct => "struct",
+        }
+    }
+
+    fn data_type(self) -> DataType {
+        match self {
+            Element::Int64 => DataType::Int64,
+            Element::Utf8 => DataType::Utf8,
+            Element::Struct => DataType::Struct(self.struct_fields()),
+        }
+    }
+
+    fn struct_fields(self) -> Fields {
+        Fields::from(vec![
+            Field::new("a", DataType::Int64, true),
+            Field::new("b", DataType::Utf8, true),
+        ])
+    }
+
+    /// A flat child array of `count` elements, which the list offsets then carve into rows.
+    fn values(self, count: usize) -> ArrayRef {
+        let ints = Int64Array::from_iter_values((0..count).map(|i| i as i64));
+        match self {
+            Element::Int64 => Arc::new(ints),
+            Element::Utf8 => Arc::new(StringArray::from_iter_values(
+                (0..count).map(|i| format!("str_{i}")),
+            )),
+            Element::Struct => {
+                let strings = StringArray::from_iter_values((0..count).map(|i| format!("str_{i}")));
+                Arc::new(StructArray::new(
+                    self.struct_fields(),
+                    vec![Arc::new(ints), Arc::new(strings)],
+                    None,
+                ))
+            }
+        }
+    }
+}
+
+/// One input batch: a `List` column of `fan_out`-element rows, plus `carried` passthrough
+/// columns that unnesting has to replicate.
+///
+/// With `nulls`, every tenth row is a NULL list. That is the shape `explode_outer` sees, and it
+/// is also what decides whether the unnested column can be sliced out of the child or has to be
+/// gathered: a NULL row under outer semantics is padded, which breaks the run.
+fn input_batch(element: Element, fan_out: usize, carried: usize, nulls: bool) -> RecordBatch {
+    let total = ROWS_PER_BATCH * fan_out;
+    let offsets: Vec<i32> = (0..=ROWS_PER_BATCH).map(|r| (r * fan_out) as i32).collect();
+    let null_buffer =
+        nulls.then(|| NullBuffer::from_iter((0..ROWS_PER_BATCH).map(|row| row % 10 != 0)));
+
+    let list = ListArray::new(
+        Arc::new(Field::new("item", element.data_type(), true)),
+        OffsetBuffer::new(offsets.into()),
+        element.values(total),
+        null_buffer,
+    );
+
+    let mut fields = vec![Field::new("arr", list.data_type().clone(), true)];
+    let mut columns: Vec<ArrayRef> = vec![Arc::new(list)];
+    for c in 0..carried {
+        fields.push(Field::new(format!("k{c}"), DataType::Int64, true));
+        columns.push(Arc::new(Int64Array::from_iter_values(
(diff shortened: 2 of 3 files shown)


sunchao (approved):
## Correctness

Reviewed `634d2a8f0125fe985c87e28018fddf32794e8a00` against `0fc677962ab64f3884b95db8fb4b9723ef5870bf`. I found no actionable P1/P2 issue in the three authored files. Previously, unnesting copied a child through gather indices even when the result was already contiguous. The new path preserves the same rows by requiring consecutive physical ranges, an offset span equal to the requested output count, and no populated null parent slot. The last check is essential: hidden values beneath a null parent can otherwise cancel padding elsewhere. Because each caller's target length is at least the corresponding logical list length, equality of the totals then proves that no row needs padding.

I traced sliced input, element nulls, carried-column repetition, parallel position/value lists and outer generation against the planner and maintained Spark 3.5/4.0 generator implementations. Slicing starts at the actual first offset; element validity remains in the child; and the existing empty-to-null wrapper makes empty/null outer rows use the padded path. Multi-column and recursive length calculation retain the general kernel, while the fused single-List calculation preserves the chunking counts. View arrays remain on the gather path. The Arrow import and fresh-buffer conversion paths also support retaining child slices after advancing the input.

The [native CI job](https://github.com/apache/datafusion-comet/actions/runs/33995526137/job/101385588854) passed all eight added tests and all 1,175 executed native tests, with four skipped. The [Spark 4.0/JDK 21 execution job](https://github.com/apache/datafusion-comet/actions/runs/33995526137/job/101389068255) passed all 37 generator-suite cases, including nullable elements, structs, offset slices, outer rows and batch boundaries; its full shard had 810 successes, three cancellations and five ignored tests. Both checked out merge `7500cf992c716b4e26d0f2e43548218e49b41809`, whose parents are this base and head; the reviewed implementation and test paths match the head. An independent index-level model found no slice/gather mismatch across 1,038,603 finite cases. This model is not Rust/Arrow execution. No local product build or benchmark ran, and maintained Spark 3.4/4.1 sources were unavailable. At the final CI check on 2026-09-05 23:51:55 UTC, 62 checks had succeeded, seven were skipped and four remained in progress. No failing check was reported at that cutoff.

## Performance

The main saving follows directly from the code: eligible output avoids an index buffer and child gather, which is particularly useful for strings and nested values. The fused length pass removes intermediate arrays for the single-List case, and repeat-index construction fills each run directly. Carried columns still require repetition and gathering, while padded outer chunks retain the gather path, so their expected gains are smaller.

The new benchmark separates fan-out, element type, carried columns and outer-null cases. Its inputs are prepared outside the timed operator execution. I treat the reported speedups as author measurements; I did not reproduce them. The documented memory trade-off is real: retaining a small output slice can retain the whole input child's backing buffers, including after that input batch has drained. The benchmark collects all output batches and therefore does not measure selective downstream retention.

## Design

The specialization is placed at the unnest kernel, where the existing target lengths and physical layout can prove equivalence. It therefore benefits both `explode` and `posexplode` without adding planner branches or changing Spark admission rules. Keeping the generic gather and length kernels provides a clear fallback for padding, view layouts and other supported operator shapes. The benchmark is registered separately and does not alter runtime configuration.

## Abstraction & complexity

The contiguous-layout trait method and small guard express the required invariant without introducing another execution operator or duplicate unnest implementation. The module documentation now makes the deliberate divergence from upstream explicit and records what a future replacement must preserve. I found no additional abstraction or complexity issue requiring a change.

PR #5679, opened 2026-09-04
fix: fall back for concat_ws with array arguments instead of failing natively

### LINKED ISSUE #5675: concat_ws with array<string> arguments fails natively instead of flattening like Spark
### Describe the bug

Spark's `ConcatWs` accepts both `StringType` and `ArrayType(StringType)` children and flattens array arguments: `concat_ws(',', array('a','b'), 'c')` = `a,b,c`. `CometConcatWs.getSupportLevel` (`spark/src/main/scala/org/apache/comet/serde/strings.scala:291-298`) never inspects child types and lowers every non-foldable call to DataFusion's `concat_ws`, which accepts only `Utf8` / `LargeUtf8` / `Utf8View` and fails with `Input was List(...) which is not a supported datatype for concat_ws function` (datafusion-functions 54.1 `concat_ws.rs`). The query fails at native execution where Spark returns a result. There is no Comet test with an array argument.

### Steps to reproduce

```sql
CREATE TABLE t USING parquet AS SELECT * FROM VALUES (array('a','b'), 'c') AS t(arr, s);
SELECT concat_ws(',', arr, s) FROM t;
-- also: SELECT concat_ws(',', split(s, ' ')) FROM t
```

Spark: `a,b,c`. Comet: native error `Input was List(Field { name: "item", data_type: Utf8, ... }) which is not a supported datatype for concat_ws function`.

### Expected behavior

`a,b,c` — either flatten array arguments natively like Spark, or fall back to Spark for them.

### Proposed solution

Short term: in `getSupportLevel`, return `Unsupported` (with a reason) when any child's `dataType` is an `ArrayType`, so the expression falls back (or goes through the codegen dispatcher). Longer term: a Spark-compatible native `concat_ws` that flattens list arguments (skipping null elements like Spark). Add a test with an array argument.

### Additional context

Only the all-foldable case is declined today (and Spark constant-folds that anyway). Related: #3339 (null-separator crash, fixed by #3542 with a null-literal branch but no argument-type gate).

peterxcli:

Closes #5675.

## Rationale for this change

Spark's `ConcatWs` accepts `array<string>` arguments after the separator and flattens their elements into the strings to join (`concat_ws(',', array('a','b'), 'c')` = `a,b,c`, null elements skipped). `CometConcatWs.getSupportLevel` never inspected child types and lowered every non-foldable call to DataFusion's `concat_ws`, which accepts only string arguments, so `concat_ws(',', arr, s)` failed at native execution with `Input was List(...) which is not a supported datatype for concat_ws function` where Spark returns a result.

- `CometConcatWs` mixes in `CodegenDispatchFallback` and returns `Unsupported` (with a documented reason) whenever any argument has an `ArrayType`. These calls now run through the JVM codegen dispatcher (Spark's own `ConcatWs.doGenCode` inside the Comet pipeline) and fall back to Spark with that reason only when `spark.comet.exec.scalaUDF.codegen.enabled=false`. The NULL-separator shortcut and the native path for plain string arguments are unchanged.
- `expressions.md`: `concat_ws` is now `Hybrid` (what `GenerateDocs` emits for a `CodegenDispatchFallback` serde) with a note; the compatibility page picks up the new unsupported reason automatically.
- Follow-up (not in this PR): a Spark-compatible native `concat_ws` kernel that flattens list arguments (skipping null elements like Spark) so these calls can leave the dispatcher.

- New `CometStringExpressionSuite` test `concat_ws with array<string> arguments`: over a Parquet table with an `array<string>` column it runs `concat_ws(',', arr, s)`, `concat_ws(',', s, arr)`, `concat_ws(',', arr)`, `concat_ws('-', arr, s, arr)`, `concat_ws(',', split(s, ' '))` and `concat_ws(',', split(s, ' '), arr)` and asserts Spark's answer with the whole plan in Comet and the dispatcher actually running; with the dispatcher disabled it asserts the Spark answer plus the serde's fallback reason; it also asserts a NULL separator still runs natively and that plain string arguments keep the native path.
- `sql-tests/expressions/string/concat_ws.sql`: a new `array<string>` table and eight query blocks (column, mixed, single, repeated, `split`-produced and literal array arguments, NULL separator), all asserting Spark-identical results.
- Ran `CometStringExpressionSuite` and `CometSqlFileTestSuite concat_ws` (Spark 4.1 profile): 35/35 passed. Native code is untouched.

### DIFF
--- a/docs/source/user-guide/latest/expressions.md
+++ b/docs/source/user-guide/latest/expressions.md
@@ -561,7 +561,7 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `chr` | ✅ | Native |  |
 | `collate` | 🔜 | — | Spark collation (umbrella [#2190](https://github.com/apache/datafusion-comet/issues/2190)) |
 | `collation` | ✅ | — | Constant-folded to a literal (Spark 4.0+) |
-| `concat_ws` | ✅ | Native |  |
+| `concat_ws` | ✅ | Hybrid | Array arguments route through the JVM codegen dispatcher; string arguments run natively ([details](compatibility/expressions/string.md)) |
 | `contains` | ✅ | — |  |
 | `decode` | ✅ | — |  |
 | `elt` | ✅ | Codegen dispatch |  |
--- a/spark/src/main/scala/org/apache/comet/serde/strings.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/strings.scala
@@ -20,7 +20,7 @@
 package org.apache.comet.serde
 
 import org.apache.spark.sql.catalyst.expressions.{Attribute, Base64, BitLength, Cast, Concat, ConcatWs, Contains, Elt, Empty2Null, EndsWith, Expression, FindInSet, FormatNumber, FormatString, GetJsonObject, InitCap, Left, Length, Levenshtein, Like, Literal, Lower, Mask, OctetLength, Overlay, RegExpExtract, RegExpExtractAll, RegExpInStr, RegExpReplace, Right, RLike, SoundEx, StartsWith, StringLocate, StringLPad, StringRepeat, StringReplace, StringRPad, StringSplit, StringTranslate, Substring, SubstringIndex, ToCharacter, ToNumber, TryToNumber, UnBase64, Upper}
-import org.apache.spark.sql.types.{BinaryType, DataTypes, IntegerType, LongType, StringType}
+import org.apache.spark.sql.types.{ArrayType, BinaryType, DataTypes, IntegerType, LongType, StringType}
 
 import org.apache.comet.CometConf
 import org.apache.comet.serde.ExprOuterClass.Expr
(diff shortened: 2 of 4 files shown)


sunchao (approved):
## Summary

`concat_ws` with array arguments now uses Spark's generated implementation through the existing JVM codegen dispatcher instead of reaching a native function that rejects lists. Ordinary string-column calls and the literal-NULL separator shortcut retain their native paths. I found no actionable P1/P2 issue in the reviewed changes.

## Correctness and compatibility

I checked the maintained Spark 3.5 and 4.0 implementations, including array flattening, null arrays and elements, empty inputs, argument order and separator handling. The new input-shape guard and shared fallback path preserve those semantics. The SQL cases exercise the results, while the Scala assertions verify dispatcher use, disabled-dispatcher fallback and the unchanged string-only path.

## Validation and CI

The new array regression and `concat_ws.sql` passed in the [Spark 3.5 Linux job](https://github.com/apache/datafusion-comet/actions/runs/33873119971/job/101032485436) and [Spark 4.0 macOS job](https://github.com/apache/datafusion-comet/actions/runs/33873119971/job/101025782850). These results are from CI merge `db59f299ef692c7276285bfb16f9400243d8e33a`, whose relevant implementation and test sources match this head, not a local bare-head execution. At 14:11 UTC, 63 checks had succeeded, eight were skipped and two were still running. I did not run a local Spark build or benchmark.

## Performance

The array case pays the existing cached codegen dispatch and Spark concatenation costs. The patch does not add a second array-flattening implementation or new work to the ordinary string-column path. This is a correctness fallback, not a measured speedup. Native array support remains separately tracked in #5675.

## Design

Reusing `CodegenDispatchFallback` keeps support detection, dispatcher admission and disabled-dispatcher behavior in the established flow. The shared unsupported reason also feeds the generated compatibility documentation, while the hand-edited expression list accurately describes the hybrid implementation.

## Abstraction & complexity

The change is limited to the existing trait, a shared reason and an array-type guard. The extra Scala coverage is justified by the dispatcher counters and fallback assertions that the SQL cases cannot express directly. I do not see an additional abstraction or simpler alternative needed before merge.

sunchao:
Merged, thanks @peterxcli !

Maybe we should also implement native array support as separate follow-up items.

peterxcli:
> Maybe we should also implement native array support as separate follow-up items.

filed https://github.com/apache/datafusion-comet/issues/5687, thanks for the suggestion!

Issue #5389, opened 2026-08-18
FuzzDataGenerator silently drops nulls for Boolean/Byte/Short/Integer columns


Smallfu666:
### Describe the bug

`DataGenOptions.allowNull` defaults to `true`, but a `Boolean`, `Byte`, `Short` or `Integer` column
declared `nullable = true` never receives a single null.

Those four derive their values from the `LongType` generator and then unbox each element
(`spark/src/main/scala/org/apache/comet/testing/FuzzDataGenerator.scala:200-212`):

```scala
case DataTypes.BooleanType =>
  generateColumn(r, DataTypes.LongType, numRows, options)
    .map(_.asInstanceOf[Long].toShort)
    .map(s => s % 2 == 0)
case DataTypes.ByteType =>
  generateColumn(r, DataTypes.LongType, numRows, options)
    .map(_.asInstanceOf[Long].toByte)
case DataTypes.ShortType =>
  generateColumn(r, DataTypes.LongType, numRows, options)
    .map(_.asInstanceOf[Long].toShort)
case DataTypes.IntegerType =>
  generateColumn(r, DataTypes.LongType, numRows, options)
    .map(_.asInstanceOf[Long].toInt)
```

The `LongType` generator does emit nulls, one in fifty:

```scala
case DataTypes.LongType =>
  Range(0, numRows).map(_ => {
    r.nextInt(50) match {
      case 0 if options.allowNull => null
      ...
```

but the elements are typed `Any`, so `_.asInstanceOf[Long]` compiles to
`scala.runtime.BoxesRunTime.unboxToLong`, which returns `0` for a null reference rather than
throwing. Every generated null becomes an ordinary value:

| type | a generated `null` becomes |
| --- | --- |
| `ByteType` | `0` |
| `ShortType` | `0` |
| `IntegerType` | `0` |
| `BooleanType` | `true`, since `0 % 2 == 0` |

`Long`, `Float`, `Double`, `String` and `Array` keep their nulls, and `Binary` keeps them too
because it derives from `String` through a `case _ => null` arm rather than an unboxing cast. So the
gap is silent and type-dependent: a suite fuzzing a nullable `Long` column really is testing nulls,
and the identical-looking `Int` column beside it is not.

### This has already let a native bug through an end-to-end test

Not a theoretical coverage concern. PR #2630 (`00922cfc4`, "Fallback to Spark for lpad/rpad for
unsupported arguments & fix negative length handling") introduced, **in the same commit**:

[...]

PR #5680, opened 2026-09-04
fix: return NULL from rpad/lpad when the length column is NULL instead of panicking

### LINKED ISSUE #5672: rpad/lpad panic (Option::unwrap on None) when the length column contains NULL
### Describe the bug

`spark_read_side_padding_internal` (`native/spark-expr/src/static_invoke/char_varchar_utils/read_side_padding.rs:225`) iterates the length array with `let length = length.unwrap();`, so any NULL in the length column panics with `called Option::unwrap() on a None value` and fails the task with a `CometNativeException`. Spark's `StringRPad` / `StringLPad` are `NullIntolerant` and return NULL for that row.

The path is default-on: `CometStringRPad` / `CometStringLPad` only require the pad to be a literal, so `rpad(s, len_col)` runs natively. Both the 2-arg `[Array, Array]` arm (lines 144-162) and the 3-arg `[Array, Array, Scalar]` arm (163-183) reach the unwrap.

### Steps to reproduce

```sql
CREATE TABLE t USING parquet AS SELECT * FROM VALUES ('abc', 5), ('abc', CAST(NULL AS INT)) AS t(s, l);
SELECT rpad(s, l) FROM t;      -- same for lpad(s, l) and rpad(s, l, 'x')
```

Spark: `abc  `, `NULL`. Comet: task fails — `called Option::unwrap() on a None value` at `read_side_padding.rs:225`.

Unit level: `spark_rpad(["abc","abc"], lengths=[Some(4), None], " ")` panics.

### Expected behavior

NULL for every row whose length is NULL, like Spark.

### Proposed solution

Replace the unwrap with a match that appends null (and continues) when either the string or the length is null, in both arms. Add a unit test with a NULL length and a NULL-length row to `string_rpad.sql` / `testStringPadding`.

### Additional context

The fuzz test cannot catch this because `FuzzDataGenerator` never generates NULL integers (#5389). The unwrap was introduced in #2630; #2096 was a different, earlier `rpad` panic (non-literal length hitting the unsupported-arguments arm).

peterxcli:

Closes #5672.

## Rationale for this change

`rpad(s, len)` / `lpad(s, len)` (and the 3-arg forms with a literal pad) run natively by default, and `spark_read_side_padding_internal` unwrapped every value of the length array. Any NULL in the length column panicked with `called Option::unwrap() on a None value` and failed the task with a `CometNativeException`. Spark's `StringRPad` / `StringLPad` are null-intolerant and return NULL for that row.

- `read_side_padding.rs`: in the array-length path (used by both the `[Array, Array]` and `[Array, Array, Scalar]` arms) a NULL length now yields a NULL row, like a NULL string. The output-buffer sizing pass only looks at non-null lengths, since values under null slots are unspecified. Non-null rows are unchanged. The scalar-length and dictionary paths only ever receive `Int32(Some(_))` (Spark folds NULL-literal arguments to NULL before Comet sees them) and are untouched.
- Rust unit tests: 2-arg and 3-arg `rpad`/`lpad` (and `read_side_padding`) with a length array containing `None` → NULL for that row, other rows unchanged; null string + null length; all-null lengths.
- `string_rpad.sql` / `string_lpad.sql`: rows `('hi', NULL, 'x')` and `(NULL, NULL, 'x')` in the Parquet-backed table, plus a `column + column + literal` query (`rpad(s, len, 'x')`) so the 3-arg array-length arm is covered natively.
- `CometStringExpressionSuite`: `lpad/rpad with NULL length` builds the rows explicitly (`FuzzDataGenerator` cannot generate NULL integers, #5389) and checks Spark vs Comet with `checkSparkAnswerAndOperator` for the 2-arg and 3-arg forms.

- `cd native && cargo test -p datafusion-comet-spark-expr read_side_padding`: 12 passed (4 new). The 4 new tests fail on `main` with the panic from the issue.
- `cargo fmt --all -- --check` and `cargo clippy --all-targets --workspace -- -D warnings`: clean.
- `./mvnw spotless:check test -Dtest=none -Dsuites="org.apache.comet.CometStringExpressionSuite,org.apache.comet.CometSqlFileTestSuite pad.sql"` (Spark 4.1.3): 2 suites, 36 tests, 0 failures, including `lpad/rpad with NULL length`, `sql-file: expressions/string/string_lpad.sql` and `sql-file: expressions/string/string_rpad.sql`.

### DIFF
--- a/native/spark-expr/src/static_invoke/char_varchar_utils/read_side_padding.rs
+++ b/native/spark-expr/src/static_invoke/char_varchar_utils/read_side_padding.rs
@@ -202,11 +202,13 @@ fn spark_read_side_padding_internal<T: OffsetSizeTrait>(
 
             // Every row is padded to its target length, so the sum of the target
             // lengths sizes the output (exactly, for ASCII input), except when a
-            // row is longer than its target and passes through untruncated.
+            // row is longer than its target and passes through untruncated. Null
+            // lengths produce null rows and are skipped: the values under null
+            // slots are unspecified and must not size the output.
             let mut data_capacity = 0usize;
             let mut max_length = 0usize;
-            for length in int_pad_array.values() {
-                let length = (*length).max(0) as usize;
+            for length in int_pad_array.iter().flatten() {
+                let length = length.max(0) as usize;
                 data_capacity = data_capacity.saturating_add(length);
                 max_length = max_length.max(length);
             }
@@ -222,15 +224,16 @@ fn spark_read_side_padding_internal<T: OffsetSizeTrait>(
             };
 
             for (string, length) in string_array.iter().zip(int_pad_array) {
-                let length = length.unwrap();
-                match string {
-                    Some(string) => {
+                match (string, length) {
+                    (Some(string), Some(length)) => {
                         if length >= 0 {
                             padder.append(&mut builder, string, length as usize);
                         } else {
                             builder.append_value("");
                         }
                     }
+                    // Spark's StringRPad/StringLPad are null-intolerant: a null
+                    // string or a null length yields a null row.
                     _ => builder.append_null(),
                 }
             }
@@ -399,6 +402,10 @@ mod tests {
         ColumnarValue::Scalar(ScalarValue::Utf8(Some(pad.to_string())))
     }
 
+    fn len_array(lengths: &[Option<i32>]) -> ColumnarValue {
+        ColumnarValue::Array(Arc::new(Int32Array::from(lengths.to_vec())) as ArrayRef)
+    }
+
     #[test]
     fn rpad_default_padding() {
         let args = vec![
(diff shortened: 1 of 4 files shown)


sunchao (approved):
## PR summary

Reviewed `6fc59544` against PR base `81d637b9`. No actionable P1/P2 findings. The change fixes NULL length propagation in the shared padding routine and adds column-backed regressions for `lpad` and `rpad`.

The sizing pass now ignores physical values beneath invalid length slots. The output loop emits NULL when either the string or length is NULL, matching the maintained Spark 3.5 and 4.0 expression sources. Existing non-null padding and truncation logic is unchanged.

## Validation and CI

Local focused validation passed 28 component checks: 12 head production tests, 8 base tests and 8 reviewer checks, including a deliberately expected original null-length panic. These compiled exact production modules against pinned Arrow/DataFusion types. They cover mixed and all-null lengths, invalid-slot storage, empty arrays, Unicode/LargeUtf8 input and unchanged argument-shape boundaries. This was not a full native workspace build or a local Spark/JNI run.

The existing [Spark 3.5/Linux expression job](https://github.com/apache/datafusion-comet/actions/runs/33873798695/job/101034672441) and [Spark 4.0/macOS expression job](https://github.com/apache/datafusion-comet/actions/runs/33873798695/job/101027561382) explicitly passed the new NULL-length regression and both padding SQL files. The [Rust test job](https://github.com/apache/datafusion-comet/actions/runs/33873798695/job/101026558095) also succeeded. They ran on synthetic merge `2ced5b41`, with the reviewed base/head as parents. The relevant implementation, serde, tests and lockfile match the head, but full-tree equivalence is not claimed.

At 13:54 UTC, 56 checks had succeeded, 9 were still running and 7 were skipped, with no failures.

## Performance

The fix keeps the existing two passes over input rows and introduces no additional array materialization. Skipping invalid lengths avoids reserving space from unspecified values. The component buffer check is not a throughput or process-memory benchmark, so no measured speedup is claimed.

## Design

Null propagation stays in the existing shared append loop. The supported array-length paths receive the same behavior without separate implementations for each function.

## Abstraction and complexity

No new production abstraction, dependency or public interface is introduced. The nullable-length helper is small and confined to tests, and the SQL regressions extend the existing test files.

sunchao:
Merged, thanks @peterxcli !

PR #5682, opened 2026-09-04
fix: align string to timestamp parsing with Spark's segment rules

### LINKED ISSUE #5674: String-to-timestamp cast diverges from Spark's parseTimestampString: rejects 1-digit segments and empty fractions, accepts zone suffixes on date-only strings and 7-digit years
### Describe the bug

Comet's string→timestamp parser is a fixed list of anchored regexes (`native/spark-expr/src/conversion_funcs/string.rs:1646-1660`) plus an offset-suffix fallback (`:1471-1475`, `:1567-1635`). Spark's `SparkDateTimeUtils.parseTimestampString` is a segment scanner with per-segment digit rules (`isValidDigits`). They disagree in four ways, all on the default-on path, for both TIMESTAMP and TIMESTAMP_NTZ (LEGACY returns NULL, ANSI raises `CAST_INVALID_INPUT`):

1. Month/day/hour/minute/second must be exactly 2 digits in Comet; Spark accepts 1–2 digits: `'2020-1-1'`, `'2020-01-01 12:34:5'` → Spark valid, Comet NULL / error.
2. Empty fraction: `'2020-01-01 12:34:56.'` → Spark valid, Comet NULL (`RE_MICROSECOND` requires `\.\d+`).
3. Zone suffix on a date-only string: `'2020-10-01Z'`, `'2020-01-01+05:30'`, `'2020-10-01 UTC'` → Spark NULL (a zone is only allowed after a time segment), Comet accepts them (midnight in that zone).
4. 7-digit years: Spark's timestamp `maxDigitsYear` is 6 (only `stringToDate` allows 7); `'0002020-01-01 00:00:00'` → Spark NULL, Comet `2020-01-01 00:00:00`.

### Steps to reproduce

```sql
CREATE TABLE t USING parquet AS SELECT * FROM VALUES
  ('2020-1-1'), ('2020-01-01 12:34:5'), ('2020-01-01 12:34:56.'), ('2020-10-01Z'), ('0002020-01-01 00:00:00') AS t(s);
SELECT s, CAST(s AS TIMESTAMP) FROM t;
```

| input | Spark (UTC) | Comet |
|---|---|---|
| `2020-1-1` | 2020-01-01 00:00:00 | NULL |
| `2020-01-01 12:34:5` | 2020-01-01 12:34:05 | NULL |
| `2020-01-01 12:34:56.` | 2020-01-01 12:34:56 | NULL |
| `2020-10-01Z` | NULL | 2020-10-01 00:00:00 |
| `0002020-01-01 00:00:00` | NULL | 2020-01-01 00:00:00 |

With `spark.sql.ansi.enabled=true` the first three fail the whole query in Comet.

### Expected behavior

The same accept/reject set as Spark.

### Proposed solution

Port Spark's byte-scanning segment machine (segments 0–8 with per-segment digit-count validation; a zone suffix only after a time segment) for both the TZ and NTZ parsers. At minimum: relax month/day/hour/minute/second to `\d{1,2}`, allow an empty fraction, cap timestamp years at 6 digits, and only attempt suffix extraction when the remainder ends after a time segment. Add these strings to the cast suite — the fuzz alphabet `0123456789/:T` contains neither `-` nor `.`, so it cannot generate them.

### Additional context

Open PR #5130 re-implements the same 14 shapes as a byte classifier with bit-identical output, so it preserves these divergences. #5165 tracks sibling trimming divergences in the same parser; #3776 asks to port `DateTimeUtilsSuite`, which would surface some of these.

peterxcli:

Closes #5674.

## Rationale for this change

Comet's string-to-timestamp regexes disagreed with Spark's segment scanner for both TIMESTAMP and TIMESTAMP_NTZ. Besides rejecting valid inputs, the fallback could silently return the wrong timestamp: in UTC, `2020-10-1` and `2020-12-1` were interpreted with a trailing -01:00 offset, producing 01:00 instead of midnight. It also rejects date-only zone suffixes: for example, `CAST('2020-01-01Z' AS TIMESTAMP)` incorrectly returned midnight on main and now returns NULL, matching Spark. Unicode fractions such as `2020-01-01 12:34:56.1٢٢٢` and `T1:2:3.1٢٢٢` could panic when truncated at a non-character byte boundary.

The ASCII digit classes also improve throughput. In [andygrove's latest review benchmarks](https://github.com/apache/datafusion-comet/pull/5682#issuecomment-5561675333) at `89e5ff5cc` against the merge base, canonical, microseconds, and offset_suffix times decreased by **36.6%, 44.1%, and 18.6%**, respectively. The offset figure supersedes the earlier 25% measurement. The reviewer also reports **43.4% less time for `spark4_legacy/padded`** after replacing four anchored regex checks with `starts_with('T')`.

The `date_only_zone` shape takes 2.6% more time in UTC and 3.2% more in non-UTC, because it now validates that zones follow seconds and rejects inputs that previously returned wrong values. This small cost buys a correctness fix. These are reviewer-reported results at the stated revision, before the zone-whitespace follow-up; they are not new measurements of this follow-up. `[0-9]` avoids the more expensive Unicode `Nd` class compiled by `\d` with regex's default features.

- Match Spark's ASCII segment rules: 4–6 year digits, 1–2 digits for other components, and an optional empty fraction. DATE retains its separate 7-digit-year grammar.
- Accept zone suffixes only after seconds/fractions in both timestamp parsers. Match Java `String.trim` at all three zone-stripping sites (`c <= U+0020`), accepting ASCII control separators while rejecting non-ASCII spaces and DELETE. This accepts `2021-11-22 10:54:27 +08:00` while still rejecting date-only zones.
- Use checked fraction slices in both decoders so a future regex change cannot restore the UTF-8 boundary panic.
- Extend Rust and Parquet-backed Scala regressions with two-digit months/one-digit days, negative years with offsets, Unicode panic inputs, valid spaced zone suffixes, malformed partial timestamps with offsets, and ASCII/Unicode zone separators. Check every whitespace case under legacy, TRY and ANSI for TIMESTAMP/NTZ in UTC and America/Los_Angeles. Widen the timestamp fuzz alphabet with `-`, `.`, `+`, and `Z`.
- The wider fuzz corpus exposed explicit positive years being rejected by Comet. Track this separately in #5716 and the compatibility guide; exclude only bare `+[0-9]{4,6}` years from this fuzz comparison until fixed.
- Add benchmark shapes for one-digit segments, empty fractions, and rejected date-only zones; skip the invalid-only shape under ANSI. Clarify that the `spark4_legacy` padded/mixed cases measure guard overhead without taking the rejection branch.
- Keep the broader padded-whitespace mismatch (#5149), positive years (#5716), and the pre-existing NTZ ANSI-error asymmetry (#5749) outside this fix. Move the #5716 documentation bullet into the existing divergence list.

Fresh validation for the zone-whitespace review follow-up:

- The new native regression fails on the previous parser at `89e5ff5cc`: U+0001 before an offset incorrectly returned NULL. With the fix, all **663 spark-expr unit tests pass**. The regression covers ASCII controls, Unicode spaces, DELETE, numeric/named/Z suffixes, fractions, legacy/TRY/ANSI, both Spark-version flags, and NTZ with zones allowed or disallowed.
- Spark 4.1.3 `CometNativeCastSuite Timestamp`: **27 passed, zero failures, 7 existing ignored tests**. This includes all 16 requested additional segment fixtures and the new Parquet-backed whitespace regression. Each whitespace input is checked separately for TIMESTAMP and NTZ in UTC and America/Los_Angeles, including ANSI behavior.
- Matching native JNI build, JVM test compilation, **77 timestamp benchmark smoke cases**, package-wide Clippy with `-D warnings`, Rust formatting, Maven Spotless/scalastyle, and `git diff --check` passed.
- The benchmark smoke run checks execution, not performance. The revision-qualified throughput figures above are the reviewer's A/B results; no new timed A/B is claimed for this follow-up.

Reproduction:

```sh
# From native/
cargo test -p datafusion-comet-spark-expr --lib
cargo build -p datafusion-comet
cargo test -p datafusion-comet-spark-expr --bench cast_string_to_timestamp -- --test
cargo clippy -p datafusion-comet-spark-expr --all-targets -- -D warnings
```

```sh
# From the repository root
./mvnw test -Dtest=none '-Dsuites=org.apache.comet.CometNativeCastSuite Timestamp' -Djni.dir="$PWD/native/target/debug"
```


native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1636,34 +1641,52 @@ fn extract_offset_suffix(value: &str) -> Option<(&str, Tz)> {
...
+// These shapes transcribe the per-segment digit rules of Spark's
+// `SparkDateTimeUtils.parseTimestampString` (`isValidDigits`): the year takes 4-6 digits
+// (`maxDigitsYear = 6`, so "0002020-01-01" is malformed for a timestamp even though
+// `stringToDate`, ported by `date_parser`, allows 7), month/day/hour/minute/second take 1-2
+// digits each, and the fraction takes any number of digits including none ("12:34:56." is
+// valid), of which only the first six are kept.
 static RE_YEAR: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^-?\d{4,6}$").unwrap());
-static RE_MONTH: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^-?\d{4,7}-\d{2}$").unwrap());
-static RE_DAY: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^-?\d{4,7}-\d{2}-\d{2}$").unwrap());
+static RE_MONTH: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^-?\d{4,6}-\d{1,2}$").unwrap());
```
sunchao: ### Correctness

[P2] Restrict newly accepted timestamp segments to ASCII digits

Could we make the broadened digit classes ASCII-only and add a column-input regression? Rust's `\d` also matches Unicode decimal digits. For example, `2020-٢` (U+0662) now matches `RE_MONTH`, then `parse_to_timestamp_info` substitutes month 1 when `parse::<u32>()` fails. The production Arrow entry points return `1577836800000000` for both TIMESTAMP and TIMESTAMP_NTZ in Legacy, Try and ANSI at this head. The base returns null in Legacy/Try and an error in ANSI. Spark 3.5 and 4.0 reject the input because their scanner accepts only ASCII `0` through `9`.

The newly accepted single-digit day, minute and second forms have the same problem. Using `[0-9]` for these shapes preserves the intended one-digit support without silently converting malformed values to valid timestamps. Please exercise the regression with column input so Spark constant folding cannot bypass the native cast. This finding excludes the pre-existing two-digit Unicode and hour-only cases.

sunchao (commented):
## Correctness

This change brings the timestamp grammar closer to Spark 3.5 and 4.0. It permits one-digit month, day and time segments and an empty fractional segment, restricts years to four through six digits, and accepts a timezone suffix only after seconds or fractional seconds. The separate DATE parser is unchanged. I compared the TIMESTAMP and TIMESTAMP_NTZ paths, the Legacy/ANSI/Try wrappers, null propagation, timezone handling and the native array dispatch against Spark's parser and cast implementation.

There is one new [P2] correctness issue in the broadened digit classes, detailed inline. A newly accepted non-ASCII digit can pass the regex and then be replaced by a default component when integer parsing fails. That turns malformed column input into a real timestamp, including in ANSI mode, instead of following Spark's rejection semantics. I separated this regression from the older two-digit Unicode behavior and did not report the older behavior as introduced here.

### Validation and CI

The focused Rust probe compiled the exact head/base production parser blocks, helper graph, error definitions and Arrow array entry points. It covered 117 main inputs plus separate Unicode and empty/error sets across Legacy, ANSI and Try and both Spark-version flags. The 2,040 main/Unicode evidence assertions and 60 empty/error checks include assertions confirming the reported defect. They are not a claim that every input matched Spark. Column-array checks are important here because literal casts can be folded by Spark before native execution.

At the final CI/discussion refresh, there were 65 successful and nine skipped checks. I also verified the tested merge's exact parents for the inspected Rust and Spark jobs. Those logs report 1,114 Rust tests passed with four skipped, 1,296 Spark 3.5 tests passed with 11 cancelled and 12 ignored, and 1,303 Spark 4.0 tests passed with three cancelled and 12 ignored. No failures were reported in those inspected test summaries. The custom Unicode input was exercised through native Arrow entry points, not through a local Spark/JNI run. I did not run a local full Comet workspace suite or a distributed query.

## Performance

The new suffix predicate does not add a heap allocation, but zoned inputs can now be classified once by that predicate and again by the main regex dispatch. I checked this with two alternating-order component benchmark runs using 8,192-row arrays, nulls, repeated kernel calls and nine samples. Ordinary date and ISO cases were roughly unchanged. Zulu-suffix cases measured about 1.16–1.21 times the previous time, and numeric offsets about 1.07–1.10 times. Named-zone results varied more.

These are shared-host parser measurements without CPU pinning, not a Spark-versus-Comet or end-to-end throughput result. They identify the extra classification cost but do not establish a query-level regression. The existing regex-dispatch design remains the main source of repeated work. I did not infer a performance improvement from the expanded correctness coverage.

## Design

Keeping the timezone-suffix rule in one predicate makes the intended Spark rule easier to inspect across TIMESTAMP and TIMESTAMP_NTZ. Preserving the separate DATE path also limits the behavioral scope. The year bound, empty-fraction behavior and suffix placement are understandable local changes rather than a new parsing architecture.

The important design invariant is that the shape recognizer and numeric decoder must accept the same alphabet. The inline fix can restore that invariant locally while retaining the intended one-digit support. A broader parser rewrite is not needed to address the demonstrated regression.

## Abstraction & complexity

The patch reuses the existing parser table, conversion helpers and cast dispatch rather than adding a new public interface, dependency or configuration. The shared suffix helper represents a concrete rule used by both timestamp paths, so its level of abstraction is appropriate.

The remaining complexity comes from having regex recognition and component decoding as separate stages. I traced both stages instead of treating a regex match as proof that decoding succeeds. The existing fallback defaults make that separation observable for malformed digits, which is why the new grammar needs the ASCII constraint described inline.

native/spark-expr/src/conversion_funcs/string.rs
peterxcli: Fixed in bcd380371. The newly broadened timestamp segment classes now use ASCII digit ranges, and the Spark regression covers Unicode month, day, minute, and second values through Parquet-backed column input for both TIMESTAMP and TIMESTAMP_NTZ across Legacy/Try/ANSI behavior. Locally, the three focused Rust segment-rule tests, cargo fmt --check, git diff --check, and ./mvnw spotless:check pass. The local native JNI link did not finish, so I left the full Spark-suite validation to CI.

sunchao (commented):
## Correctness

Follow-up on `bcd3803717ec3e6d24d7a30637251d2669307591` against base `81d637b9bf40a5be6f4f0c65ad6f497b34746e69`. The four reported month/day/minute/second examples are fixed. The added cases use Parquet-backed columns, with each malformed input evaluated separately under ANSI mode.

**[P2] The existing Unicode-digit finding remains partially unresolved.** In the newly accepted short-date forms, `2020-1-1T٢` still becomes `2020-01-01 00:00:00`, and `2020-1-1T1:2:3.٢` becomes `2020-01-01 01:02:03` (UTC; `٢` is U+0662). Both produce values for TIMESTAMP and TIMESTAMP_NTZ in Legacy, TRY and ANSI modes. The authoritative base and Spark return NULL in Legacy/TRY and reject them in ANSI. This continues [the existing finding](https://github.com/apache/datafusion-comet/pull/5682#discussion_r3935745419), rather than adding a duplicate inline comment.

The hour class in [RE_HOUR](https://github.com/apache/datafusion-comet/blob/bcd3803717ec3e6d24d7a30637251d2669307591/native/spark-expr/src/conversion_funcs/string.rs#L1655) and the fraction class in [RE_MICROSECOND](https://github.com/apache/datafusion-comet/blob/bcd3803717ec3e6d24d7a30637251d2669307591/native/spark-expr/src/conversion_funcs/string.rs#L1663) still use Unicode-aware `\d`; failed numeric conversion then defaults to zero. The maintained Spark 3.5 and 4.0 scanners accept only ASCII digits in these segments. Please make the remaining numeric classes consistent with that rule and add these mixed short-date cases to the column regressions.

### Validation

Exact production parser/Arrow cast blocks from the base, previous head and current head reproduced both failures. Spark 4.0.4 `Cast(BoundReference)` and generated projections agreed on rejection. The 180-input probe produced 9,720 native result records and 1,080 Spark evaluations through each path; 2,250 ASCII/null comparisons were unchanged by this update. This was component validation; the broader corpus also contains existing mismatches and is not an all-pass compatibility claim. No local full Comet JNI suite or Spark 3.4/3.5/4.1 runtime validation was performed. Canonical Spark 3.4/4.1 source refs were unavailable.

CI logs confirm the Rust segment tests and both timestamp segment suites in the Linux Spark 3.5/4.0 expression jobs passed. They executed merge `48846aee`, combining this head with `719cba11`, not the authoritative base above. The fresh head-check read at 21:37 UTC reports 64 successful and nine skipped checks, with no failing or unfinished check in that list. That does not change the merge/base qualification above.

## Performance

The update narrows existing regex character classes without adding another per-row scan, allocation or parser pass. No new performance concern was identified in this delta. The correctness probe is not a benchmark; no throughput improvement is claimed.

## Design

Restricting the numeric grammar is a small, appropriate fix, but it needs to cover whole accepted shapes: shortening an ASCII month/day can currently expose an unchanged Unicode hour or fraction. Applying the same digit rule throughout those patterns closes this gap without changing timezone handling or adding a separate parser.

## Abstraction & complexity

No new production abstraction is introduced. The existing shared malformed-input fixtures remain suitable for the additional cases, and separate ANSI evaluations ensure a failure in the first input does not hide later regressions.

native/spark-expr/src/conversion_funcs/string.rs
peterxcli: Fixed in 0a1fc76a0. All numeric classes in the timestamp patterns now use `[0-9]`, including years, the hour-only shape, fractions, and time-only forms. The shared Rust and Parquet-backed Spark fixtures include both `2020-1-1T٢` and `2020-1-1T1:2:3.٢`, plus a zoned fraction and Unicode year/time-only cases. Both timestamp types are checked in Legacy, TRY, and ANSI modes, with separate Spark queries for each malformed input.

The expanded parser tests failed against bcd380371 on `2020-1-1T٢` for both TIMESTAMP and TIMESTAMP_NTZ before the fix. Afterward, all three focused Rust segment-rule tests passed. `make core` succeeded, and the two focused `CometNativeCastSuite` segment-rule tests passed locally on Spark 4.1.3 (2 tests, 0 failures). Formatting, Scalastyle, and `git diff --check` also passed.

sunchao (approved):
## Correctness

Follow-up on `0a1fc76a02ae14228a6c521d6dda1949f8afc312` against base `81d637b9bf40a5be6f4f0c65ad6f497b34746e69`. The remaining cases in [the existing P2](https://github.com/apache/datafusion-comet/pull/5682#discussion_r3935745419) are fixed. Both `2020-1-1T٢` and `2020-1-1T1:2:3.٢` now follow the malformed-input path for TIMESTAMP and TIMESTAMP_NTZ. No remaining or new P1/P2 was found in this update.

All 14 timestamp patterns now use ASCII digits, matching the maintained Spark 3.5 and 4.0 byte scanners. I traced the year, hour, fraction and time-only classes through direct matching, timezone-suffix classification and both inner dispatch tables. Numeric Unicode cannot pass those gates and then become a default component. The date parser, numeric decoders, dispatch and dependencies are unchanged. For ASCII input, the regex languages are equivalent to the previous head, preserving the existing widths, empty fractions, timezone handling and microsecond calculation.

### Validation

The expanded Rust parser fixtures and Parquet-backed Scala fixtures include the remaining examples, Unicode years, zoned fractions and time-only forms. Each malformed Scala input is evaluated separately for both timestamp types, so an earlier ANSI failure cannot hide later inputs. The helper explicitly checks native operators for Legacy and compares TRY results and ANSI exceptions.

CI logs confirm the three Rust segment tests and both timestamp segment suites passed on [Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/33936305616/job/101228628379) and [Spark 4.1](https://github.com/apache/datafusion-comet/actions/runs/33936305616/job/101228628442). Those jobs tested merge `7db8334c`, combining this head with `719cba11`. The relevant parser, dependency and timestamp-test sections match the head; this is not a claim that CI tested the authoritative base above.

The local Spark 4.0.4 component probe covered 457 inputs with 2,742 evaluations through each of the interpreted and generated paths. All 266 focused Unicode inputs returned NULL in Legacy/TRY and CAST_INVALID_INPUT in ANSI for both types. Source comparison additionally verified all 14 ASCII grammars. No new local native parser build or full Comet JNI suite was run. The earlier native corpus remains historical and includes known pre-existing mismatches; maintained Spark 3.4/4.1 source gaps remain recorded.

At `2026-09-05T02:39:14.744Z`, current-head checks reported 53 successful, 11 running and 7 skipped, with no failures. Full CI was still running at that cutoff.

## Performance

The update narrows the existing regex classes without adding a scan, allocation or parser pass. ASCII inputs retain the same classification decisions. No new performance concern was found in this delta, and no throughput improvement is claimed; the correctness checks were not benchmarks.

## Design

Applying the same digit alphabet to every accepted timestamp shape closes the gap between recognition and numeric decoding. The shared predicates now enforce that rule consistently across direct and zoned input, while preserving the separate DATE grammar. This is an appropriate local fix.

## Abstraction & complexity

No new production abstraction, dependency or configuration is introduced. Extending the shared malformed-input fixtures keeps the change small, and the existing per-input ANSI queries provide useful regression coverage without another testing framework.

andygrove:
I dug into this one because the regex edits looked like they might cost something, and it turns out the opposite is true. Benchmarking `cast_string_to_timestamp` at 81d637b9b against 0a1fc76a0, the cast gets a lot faster: -37% on the `canonical` shape, -44% on `microseconds`, -25% on `offset_suffix`, and comparable numbers across the NTZ and non-UTC groups. I did a base-vs-base run first to get a noise floor and those two shapes sit under 1%, so it isn't measurement noise. The cause is the `\d` to `[0-9]` swap rather than the segment rules, since `regex` comes in with default features and `\d` compiles to the Unicode `Nd` class, which is much more expensive to match than one ASCII range. Could you add that to the description? As written this reads as a pure strictness change, and the next person to touch these patterns has no way to know the ASCII classes are load-bearing for throughput as well as correctness. It would also be worth adding cases to `cast_string_to_timestamp.rs` for the shapes you have made reachable, since none of the six existing ones use 1-digit segments, an empty fraction, or a zone after a date-only value.

The description also undersells the fix. On main, `CAST('2020-10-1' AS TIMESTAMP)` returns 2020-10-01 01:00:00 in a UTC session rather than NULL, because `RE_DAY` needed a 2-digit day, so `extract_offset_suffix` read the trailing `-1` as a -01:00 offset and `RE_MONTH` then matched the remaining `2020-10`. `2020-12-1` does the same thing. Those are silently wrong values rather than NULLs, which makes them the most valuable regression tests in the whole set, and neither one is in `SPARK_SEGMENT_RULE_VALID` or `sparkSegmentRuleTimestamps`. Every valid case you listed is a 1-digit-month shape, and those only ever returned NULL. Could you add the 2-digit-month with 1-digit-day shape to both corpora? While you are in there, neither Scala list has a negative-year case, valid or malformed, even though `-?[0-9]{4,6}` is one of the edited patterns and the Rust lists do cover it.

Related, and also worth putting in the description: on main `CAST('2020-01-01 12:34:56.1٢٢٢' AS TIMESTAMP)` panics with "end byte index 6 is not a char boundary", because Unicode `\d` let `RE_MICROSECOND` match a fraction of multi-byte digits and `parse_to_timestamp_info` then byte-slices a `&str`. `T1:2:3.1٢٢٢` reaches the same bug through `parse_str_to_time_only_timestamp`. Your change makes both unreachable, which is a bigger deal than stricter parsing. Since those two slices are now safe only because no pattern can hand them non-ASCII, and that invariant lives a few hundred lines away at the regex definitions, could we make them char-boundary-safe on their own so that a future pattern edit cannot resurrect the panic?

On the `'2021-11-22 10:54:27 +08:00'` case you left out of scope, I think it belongs in this PR. Adding `let stripped = stripped.trim_end();` before the `ends_with_seconds_segment` call in `timestamp_parser` returns the Spark value, leaves `2020-10-01Z` at NULL, and passes all 662 spark-expr tests. It is the same `trim_end()` that `timestamp_ntz_parser` already applies. As it stands the two paths return different things for that input, a NULL and a value, which cuts against the point of giving them one shared zone rule. If you would rather keep it separate that is fine, but then it needs a filed issue and a bullet under "Known result-value divergences" in `docs/source/user-guide/latest/compatibility/index.md`, next to the #5149 whitespace entry, or it will not get picked up.

For what it is worth, I checked `parseTimestampString` and `isValidDigits` at v3.4.4, v3.5.9, v4.0.1, v4.1.1 and master, and all five agree on every rule you encoded, including the timestamp year bound of 6 against 7 for `stringToDate`, the empty fraction, and the zone id only being captured inside the seconds or fraction segment. No compatibility objection from me on the behavior itself.

Last thing, and I think it is the real story here. Your note about the fuzz alphabet is the important one: `timestampPattern` being `"0123456789/:T" + whitespaceChars` with no `-`, `.` or `+` means it cannot generate a plausible date, which is why this whole family of mismatches sat behind a passing fuzz test. Any appetite for widening it? If it surfaces more mismatches, those are issues to file rather than a reason to leave the alphabet alone.

peterxcli:
Addressed the follow-up:

- Added the ASCII-class throughput explanation to the description, attributing the benchmark numbers to your measurements, and documented the invariant beside the regexes.
- Added `2020-10-1` / `2020-12-1` to the Rust and Scala corpora, negative-year Scala cases, and both Unicode fraction panic examples. Checked slices in both decoders now prevent the boundary panic independently of the regexes; direct decoder tests cover that.
- Included `trim_end()` and exact TZ/NTZ regression values for the spaced offset.
- Added benchmark shapes for single-digit segments, empty fractions, and rejected date-only zones.
- Widened the fuzz alphabet with `-`, `.`, `+`, and `Z`. This exposed a separate positive-year mismatch (`+7528`, etc.), now filed as #5716 and documented in the compatibility guide. Only bare positive years are excluded pending that fix; the wider alphabet stays.

Validation: 662 Rust tests passed; `make core`, benchmark compilation, all 77 benchmark smoke cases, Clippy (`-D warnings`), and formatting/style checks passed. The Spark 4.1.3 full cast suite had 169 passes and the one positive-year failure; after excluding that tracked case, all 7 string-to-timestamp tests passed, including both segment-rule suites and the widened fuzz test.

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1470,7 +1474,13 @@ fn timestamp_parser<T: TimeZone>(
 
     if !has_direct_match {
         if let Some((stripped, suffix_tz)) = extract_offset_suffix(value) {
-            return timestamp_parser_with_tz(stripped, eval_mode, &suffix_tz);
+            let stripped = stripped.trim_end();
```
sunchao: ### Correctness

[P2] Preserve the Spark 4 leading-whitespace check on the new suffix path

For a STRING column containing `\tT1:2:3 +08:00` (a leading TAB), Spark 4 returns NULL in Legacy/TRY and CAST_INVALID_INPUT in ANSI. The guard above checks the whole trimmed value, so its anchored time-only patterns do not match while ` +08:00` is still present. This new trim then turns the extracted `T1:2:3 ` into `T1:2:3` and sends it directly to `timestamp_parser_with_tz`, which produces a value without another Spark-4 check. At `0a1fc76a`, the same prefix failed the seconds-segment gate and the input was rejected. Please preserve/reapply the raw-leading-whitespace restriction after suffix extraction and add this column case in all three modes. This is specific to timezone-aware TIMESTAMP on Spark 4; maintained Spark 3.5 accepts the leading whitespace, while NTZ rejects time-only strings. The Spark result was confirmed with 4.0.4 Cast evaluation and generated projections; the Comet result follows from the changed source path.

sunchao (commented):
## Correctness

Follow-up at `7c449fd9` against `81d637b9`, covering the changes since `0a1fc76a`. One new P2 remains: trimming the extracted suffix prefix newly admits a leading-whitespace T-time-only input on Spark 4 without reapplying its existing rejection guard. The inline comment traces the exact input and Legacy/TRY/ANSI behavior.

The two fraction `get(..)` guards are safe, and all 14 ASCII digit patterns remain unchanged, so the previous Unicode-digit P2 stays fixed. The added month10/12-day1, negative-year and spaced-offset fixtures improve coverage. Null handling, fractional truncation, component bounds and overflow paths are unchanged. The known positive bare-year mismatch is explicitly tracked by [#5716](https://github.com/apache/datafusion-comet/issues/5716); the narrow fuzz exclusion and compatibility entry do not fix it. Broader whitespace-set differences, including Unicode separators before a zone, remain covered by [#5149](https://github.com/apache/datafusion-comet/issues/5149); they are not duplicated as another inline.

I compared the maintained Spark 3.5 and 4.0 sources: 3.5 trims bytes before scanning, while 4.0 retains the original byte index and accepts a leading T only at index zero. NTZ rejects time-only input entirely. Maintained Spark 3.4/4.1 source refs were unavailable, which remains a coverage limitation.

### Validation

The local Spark 4.0.4 oracle exercised 251 non-foldable column inputs through both Cast evaluation and generated projections: 1,506 results per route, covering both timestamp types and all three modes. The primary counterexample returns NULL in Legacy/TRY and CAST_INVALID_INPUT in ANSI. Comet's changed behavior is established by the source path; I did not run a new local Comet native/JNI build. A separate Rust standard-library trim probe is only whitespace-component evidence.

Current-head Rust CI and Spark 3.5/4.0/4.1 expression jobs passed the segment-rule tests. CI tested merge `6f3dc8a5` (parents `7190df63` and this head), not the authoritative base pair. The parser, dispatch and benchmark blobs, plus relevant Scala timestamp fixtures/helpers, match the head; the merge also contains unrelated numeric tests, package dependency edges and benchmark registrations. At 2026-09-05 21:23:07 UTC, the refreshed head checks report 8 skipped, 65 successful, 1 failed. All checks had completed at that cutoff. The [failed macOS scans job](https://github.com/apache/datafusion-comet/actions/runs/33978976233/job/101345763014) records a JVM SIGSEGV; its cause and relationship to this change are unproven.

## Performance

The benchmark adds single-digit segments, an empty fraction and an invalid date-only zone. It evaluates columns in prebuilt 8,192-row batches, retains null coverage, and appropriately skips invalid batches under ANSI. The safe fraction access adds no allocation, and the new trim runs only after successful suffix extraction.

The [peer benchmark report](https://github.com/apache/datafusion-comet/pull/5682#issuecomment-5552876461) measured base `81d637b9` versus the previous head `0a1fc76a`: roughly 37% lower canonical time, 44% lower microseconds time and 25% lower offset-suffix time, with similar NTZ/non-UTC results. These are historical peer measurements; I did not independently benchmark `7c449fd9`, and benchmark compilation/smoke success is not a timing result. No additional material performance finding in this follow-up.

## Design

The local defensive fraction guards and expanded fixtures fit the existing parser design. The suffix normalization correctly enables legal space-before-offset datetimes, but that path must also preserve the wrapper's version-specific admission checks. Reapplying the Spark 4 leading-whitespace restriction to the extracted T-time-only prefix, with a column regression fixture, is a narrow fix.

## Abstraction & complexity

No new abstraction, dependency or serialization layer is introduced. The existing regex recognition and component decoders remain easy to follow, and the benchmark extends the existing batch generator. No separate complexity finding.

native/spark-expr/src/conversion_funcs/string.rs
peterxcli: Fixed in 89e5ff5cc. The Spark 4 raw-leading-whitespace guard now checks the T prefix before suffix extraction, rather than requiring a base-pattern match that a timezone suffix can hide. Spark 3.x keeps its existing trim-first path.

Added the exact TAB-prefixed column case, a fractional-seconds case, and a no-leading-whitespace control through the Parquet-backed cast helper. Each is compared with Spark in Legacy, TRY, and ANSI modes for TIMESTAMP and TIMESTAMP_NTZ. Extended the native regression for Legacy/TRY/ANSI rejection and Spark 3.x acceptance.

The native regression failed on the reported input before the fix. Afterward: all 662 spark-expr unit tests and all 8 selected Spark 4.1.3 timestamp-cast tests passed. Native build, Rust formatting, Spotless, and diff whitespace checks passed. Spark 3.x compatibility was checked through the native version flag; I did not run a separate Spark 3.5 JVM suite.

sunchao (approved):
## Correctness

Follow-up at `89e5ff5c` against `81d637b9`, reviewing the delta from `7c449fd9`. The [leading-whitespace P2](https://github.com/apache/datafusion-comet/pull/5682#discussion_r3942101881) is fixed. The wrapper now rejects a leading-whitespace `T` prefix before any suffix extraction, so a timezone cannot hide the restriction. The reported TAB-prefixed input and fractional variant produce NULL in Legacy/TRY and an error in ANSI on Spark 4. Spark 3.5 retains its trim-first behavior, and NTZ still rejects time-only input.

The maintained Spark 3.5 and 4.0 sources support this distinction: 3.5 starts scanning after trimming the bytes, while 4.0 recognizes a leading `T` only at original byte index zero. The new guard cannot admit anything rejected by the old guard; it broadens rejection only to leading-whitespace `T` forms that Spark 4 rejects. I verified that all other production parser bytes are unchanged, including the 14 ASCII patterns, fraction guards, offsets, null handling and overflow paths. No new or remaining P1/P2 was found in this update.

The added Parquet-backed tests evaluate the reported input, a fractional case and a valid control separately for both timestamp types. The helper checks the native operator in Legacy and compares TRY results and ANSI errors. Current CI logs confirm the new test passed on [Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/33993614015/job/101380982285), [Spark 4.0](https://github.com/apache/datafusion-comet/actions/runs/33993614015/job/101380982335) and [Spark 4.1](https://github.com/apache/datafusion-comet/actions/runs/33993614015/job/101380982313); the [native whitespace regression](https://github.com/apache/datafusion-comet/actions/runs/33993614015/job/101380448898) also passed.

Those jobs tested merge `d9b1ef12`, combining this head with `c348f775`, rather than the authoritative base above. Its complete parser blob, timestamp dispatch arms and Scala timestamp fixtures/helpers match the head. Its dependency graph is newer, including Arrow 59.3 versus 58.4 and DataFusion 55 versus 54.1; the regex, chrono and chrono-tz package blocks match. This qualifies the CI evidence.

Locally, 232 Spark 4.0.4 inputs produced 1,392 matching results through each of Cast evaluation and generated projections. A separate standard-library guard probe rejected 105 inputs on the Spark-4 flag, all consistent with Spark's rejection, and rejected none on the Spark-3 flag. These are component checks; I did not run a new local Comet native/JNI build. Maintained Spark 3.4/4.1 source refs remain unavailable. The final current-head CI refresh at `2026-09-05T22:49:34Z` reported 62 successful, three running and seven skipped checks, with no failed check at that cutoff.

## Performance

This delta replaces up to four anchored regex checks with a prefix check on the existing leading-whitespace path. It adds no allocation or parser pass. The benchmark code is unchanged from the previous review. No material performance concern was found; no new timing or end-to-end speedup is claimed.

## Design

Checking the raw-input restriction in the outer wrapper is appropriate: it covers direct and zoned time-only forms before either parsing route can bypass the rule. The existing version flag keeps Spark 3.5 behavior intact, and the paired valid control checks that the fix does not reject ordinary zoned time-only input.

## Abstraction & complexity

The prefix predicate simplifies the existing condition without introducing an abstraction, dependency or configuration. Reusing the column-cast helper and extending the native fixture provides focused coverage for the actual bypass, including separate invalid ANSI evaluations.

andygrove:
I re-probed and re-benchmarked at `89e5ff5cc`. Everything I asked for last time is in, and the `starts_with('T')` rewrite is better than what I had in mind. I traced `parseTimestampString` at v4.0.4 and v4.1.3 to check it: once `getTrimmedStart(bytes) > 0` and `bytes(j) == 'T'`, every branch falls to `return (Array.empty, None, false)`, so there is no leading-whitespace T-prefixed input Spark accepts and your predicate is a strict superset of the four regexes it replaced. I probed `\x01T2:30`, `\x7FT2:30`, `\x0BT2:30` and `" Tgarbage"` under both version flags and Comet agrees with Spark on all of them.

The `trim_end()` I asked you to add turned out to carry an edge case I had not thought through when I suggested it. Spark builds the zone string as `new String(bytes, j, strEndTrimmed - j)` and then resolves it with `getZoneId(zoneName.trim)`. Java's `String.trim` only strips chars `<= U+0020`, so Spark keeps `U+00A0` and the other non-ASCII spaces, which makes `ZoneId.of` throw and the whole cast return NULL, and it strips `0x01` through `0x1F`, which makes the value parse. Rust's `trim_end` gets both of those backwards. I ran seven forms against Spark 4.1.3 through Parquet-backed columns and they all diverge:

| input (`WS` is the character named) | Spark | Comet |
| --- | --- | --- |
| `2020-01-01 12:34:56` `WS=U+00A0` `+08:00` | `null` | `2019-12-31 20:34:56` |
| `2020-01-01 12:34:56` `WS=U+3000` `+08:00` | `null` | `2019-12-31 20:34:56` |
| `2020-01-01 12:34:56` `WS=U+2009` `+08:00` | `null` | `2019-12-31 20:34:56` |
| `2020-01-01 12:34:56` `WS=U+00A0` `UTC` | `null` | `2020-01-01 04:34:56` |
| `2020-01-01 12:34:56` `WS=U+00A0` `Z` | `null` | `2020-01-01 04:34:56` |
| `2020-01-01 12:34:56.123` `WS=U+00A0` `+08:00` | `null` | `2019-12-31 20:34:56.123` |
| `2020-01-01 12:34:56` `WS=0x01` `+08:00` | `2019-12-31 20:34:56` | `null` |

The first six are silent wrong values, and under ANSI Comet returns a row where Spark raises. They are new here, because before this PR there was no `trim_end()` on the TIMESTAMP path and all six fell through to NULL. The NTZ path already had one, so those are older.

I know @sunchao routed the whitespace-set mismatch to #5149, and that is a fair call in general. The epic does cover the class and the compatibility guide already documents it for the padded case. What tips it for me here is that the fix is one line on a line this PR is adding, and that the PR moves the divergence from a NULL to a wrong value. Would you be willing to take:

```rust
let stripped = stripped.trim_end_matches(|c: char| c <= '\u{20}');
```

at `string.rs:1471`, and the same at the two NTZ sites at `:1823` and `:1835`? I applied exactly that and all eleven inputs I probed then match Spark, including `0x7F` staying rejected and `0x0B` and `0x0C` staying accepted, which is what Spark's split between `getTrimmedEnd` and `String.trim` produces. 662 spark-expr tests and all 26 `CometNativeCastSuite` timestamp tests still pass, so it does not disturb anything you added.

While you are in there, `2021-11-22 10:54:27 +08:00` is currently the only fixture for the whole family that `trim_end()` makes reachable, and there is no negative case in the same shape. The one I would most want covered is a zone after a non-seconds segment with whitespace in front, since that is precisely what `trim_end()` could have broken. `2020-01-01 12:34 UTC` covers the named form but not the bare-offset form. I ran these against Spark 4.1.3 and all sixteen pass on your head, so they are free regression coverage:

```scala
// sparkSegmentRuleTimestamps
"2020-01-01 12:34:56 Z",
"2020-01-01 12:34:56\t+08:00",
"2020-01-01 12:34:56\n+08:00",
"2020-01-01 12:34:56.123 +08:00",
"2020-01-01 12:34:56. +08:00",
"2020-01-01 12:34:56  +08:00",
"2020-01-01 12:34:56 -08:00",
"2020-01-01 12:34:56 Europe/Moscow",
"-0001-01-01T12:34:56 +08:00",
"-0001-01-01T12:34:56-08:00",

// sparkSegmentRuleMalformedTimestamps
"2020-01-01 12:34 +08:00",
"2020-01-01 12 +08:00",
"2020-01-01 +08:00",
"2020-01-01 Z",
"2020-01 +08:00",
"2020 +08:00",
```

The last two valid ones also cover the interaction between a leading `-` year sign and `extract_offset_suffix` picking the rightmost sign, which nothing tests today.

On the docs, the #5716 bullet landed between the `## Known result-value divergences` heading and the "The following native paths silently return values that differ from Spark" paragraph that introduces that section's list, so it renders as an orphan bullet above its own introduction. Could you move it down into the list next to the #5149 entry? That entry already describes a NULL-in-Comet direction, so it fits there.

I re-ran `cast_string_to_timestamp` at your head against the merge base, three interleaved rounds with run-to-run spread of 1.5% or less on your side. `canonical` at -36.6% and `microseconds` at -44.1% still match what I measured at `0a1fc76a0`, but `offset_suffix` is now -18.6% rather than -25%, which makes sense given `7c449fd9c` added a `trim_end()` and an `ends_with_seconds_segment()` call to that exact path. Worth updating that one number so the description matches the head. Two things it undersells, though. `89e5ff5cc` is a real speedup in its own right: swapping the four anchored regex matches for `starts_with('T')` gives -43.4% on `spark4_legacy/padded`. And the only shape that got slower is `date_only_zone` at +2.6% on UTC and +3.2% on non-UTC, which is the shape that went from returning a wrong value on main to returning NULL, so the extra `ends_with_seconds_segment` pass is what buys the correctness fix. That is worth saying explicitly, because it is the one number a future reader might otherwise try to optimise away. Related, `ends_with_seconds_segment` also fixes a silent wrong value the description does not mention at all: on main `CAST('2020-01-01Z' AS TIMESTAMP)` returns `2020-01-01 00:00:00`, and it is NULL now.

Two comments in the benchmark drifted out of date. The one above `batches` lists the shapes starting at `canonical`, but the three new ones went in above it, so it now describes only part of the array. And the `spark4_legacy` comment says it measures "the inputs where that check can fire", but the guard needs leading whitespace and a `T` prefix together, and neither `padded` nor `mixed` has both, so it measures the check's cost without ever taking its rejection branch. The cost is the interesting part given it just moved 43%, so it is worth saying that instead.

Last thing, and not something to hold this PR for. While probing your new fraction guards I noticed they both return `Ok(None)`, which on the NTZ side lands on `timestamp_ntz_parser_inner`'s `None => Ok(None)` arm at `string.rs:1871` and returns NULL even under ANSI. So if a future pattern edit ever does make those guards reachable, the invariant they exist to protect comes back as a silently swallowed ANSI error rather than a panic. That asymmetry is already reachable today through the year-range guard at `:1147`, where `CAST('294249-01-01' AS TIMESTAMP_NTZ)` under ANSI raises `CAST_INVALID_INPUT` in Spark and returns NULL in Comet while the TIMESTAMP version raises correctly. It predates this PR so I have filed it separately as #5749.

sunchao (approved):
## Correctness

Follow-up at `578b9c8b` against authoritative base `81d637b9`, covering the update from `89e5ff5c`. The zone-separator mismatch in [the latest discussion](https://github.com/apache/datafusion-comet/pull/5682#issuecomment-5561675333) is fixed. All three suffix-prefix trimming sites now use `c <= U+0020`, matching the Java `String.trim` applied to zone names by the maintained Spark 3.5 and 4.0 parsers. ASCII controls before a supported zone become valid; DELETE, NBSP, thin space and ideographic space remain invalid. Invalid inputs follow NULL in Legacy/TRY and the existing error path in ANSI.

The seconds-segment gate still runs after trimming, so a zone after a date, hour or minute remains rejected. The Spark 4 raw-leading-whitespace check still runs before suffix extraction; the previous P2 stays fixed. Mechanical comparison confirms that the production delta consists only of these three trim replacements, apart from formatting and the explanatory comment. The ASCII grammars, fraction guards, offsets, dispatch, null handling and overflow paths are unchanged. No new or remaining P1/P2 was found in this update. The broader padded-whitespace, positive-year and NTZ ANSI asymmetry issues remain outside this fix.

The new native regression covers numeric, UTC and Z suffixes, fractional seconds, all three modes, both Spark-version flags and NTZ with zones allowed or disallowed. The Parquet-backed Scala test evaluates each separator separately for both timestamp types in UTC and America/Los_Angeles. Its helper asserts native execution for Legacy and compares TRY results and ANSI outcomes. The additional valid and malformed fixtures cover the requested partial-segment and negative-year interactions.

CI logs confirm the native regression passed and the new column test passed on [Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/34057140767/job/101555362378), [Spark 4.0](https://github.com/apache/datafusion-comet/actions/runs/34057140767/job/101555362432) and [Spark 4.1](https://github.com/apache/datafusion-comet/actions/runs/34057140767/job/101555362449). These jobs tested merge `7a499adf`, combining this head with `bc74cc79`, rather than the authoritative base above. Its parser blob, timestamp dispatch and relevant Scala fixtures/helpers match the head. Arrow/DataFusion are newer in that merge; regex, chrono and chrono-tz package blocks match. At `2026-09-08T02:16:20.765Z`, all current-head checks had completed: 66 successful and eight skipped.

Locally, 264 Spark 4.0.4 component inputs produced 3,168 matching results through each of interpreted Cast evaluation and generated projections, covering every character through U+0020 plus DELETE and selected Unicode spaces, both types, three modes and two session zones. This verifies Spark's boundary; it is not a new local Comet/JNI run. Maintained Spark 3.4/4.1 source refs remain unavailable.

## Performance

The update replaces the predicate in existing suffix trims without adding an allocation or another parser pass. The NTZ path retains its existing two trims. Benchmark executable code is unchanged; the corrected comments now accurately describe its input shapes and that the padded/mixed cases measure guard cost without taking the rejection branch. No material performance concern was found in this delta.

The description now attributes the revised timing figures to the peer's `89e5ff5c` measurements and explicitly separates them from this follow-up. Those historical results and the author's benchmark smoke run are not new timing evidence for `578b9c8b`; I did not run a new timed benchmark.

## Design

Applying the same character boundary at the timezone-aware site and both NTZ sites keeps recognition and subsequent parsing consistent. The fix preserves the existing suffix-placement and Spark-version checks. Moving the unchanged #5716 bullet into the compatibility list also resolves the documentation placement issue without suggesting that positive-year parsing was fixed.

## Abstraction & complexity

The three local predicate substitutions introduce no new abstraction, dependency or configuration. Extending the existing native loops and column-cast helper provides focused coverage, including separate invalid ANSI evaluations. No additional design or complexity change is needed. Approving this revision.

sunchao (approved):
## Correctness

Follow-up on `a4c89b34d93958b0a2bceee7b650791ee34d1844` against `bb9e74020adc228e486f6f4d0fa68292b30bff31`. This merges the [previously approved revision](https://github.com/apache/datafusion-comet/pull/5682#pullrequestreview-5136700021) with the assigned base. The ordered additions and removals in all four PR files are unchanged from that approved contribution. I found no new or remaining verified P1/P2 issue in the reviewed contribution and integration.

The 14 ASCII patterns, safe fraction slices, seconds-only zone gate, three Java-style suffix trims and Spark 4 raw-leading-whitespace guard remain intact. I rechecked them against the maintained Spark 3.5 and 4.0 scanners and cast wrappers. The zone-separator boundary remains `<= U+0020`, with DELETE and the tested non-ASCII separators rejected. TIMESTAMP applies a valid zone, while TIMESTAMP_NTZ discards it when allowed and rejects time-only input. The array null/error handling, cast dispatch, Scala support policy and existing column-test helper are unchanged from the prior head.

The integration also preserves the base's NTZ ANSI-error fix. A rejected decoder result, including an out-of-range year, now reaches the ANSI error path instead of returning NULL early. That inner function is byte-identical to the assigned base and agrees with the maintained Spark ANSI wrapper. Legacy/TRY retain NULL for those rejected values. The broader padded-whitespace and positive-year limitations remain outside the authored change.

### Validation

The current [Rust job](https://github.com/apache/datafusion-comet/actions/runs/34211920985/job/102015709809) passed **1,197 tests with four skipped**, including the segment parser/array tests, timezone-whitespace regression, leading-whitespace guard and NTZ out-of-range-year regression. It checked out `f065bc25ce3886cad2973549dff62b347f2d7889`, whose parents are the assigned base and head. Its entire tree, `6c324438b41259c51bc33059bb16bd254ffeaa2b`, equals the reviewed head, including the integrated dependency versions.

At **2026-09-08 10:43 UTC**, the refreshed checks show 25 successful, 32 running, four queued and six skipped, with no failures. The remaining CI is still pending. I am not crediting a completed current JVM expression suite or native-artifact consumption at that cutoff. The previous Spark/JVM results and Spark 4.0.4 component oracle remain historical evidence. Local validation in this round was source equivalence, maintained-source/test inspection, whitespace and formatting checks. Maintained Spark 3.4/4.1 sources remain unavailable.

## Performance

The timestamp benchmark file and its string-batch generator are unchanged from the approved head. The authored parser paths add no new allocation, scan or repeated parsing in this merge. The base updates Arrow/DataFusion and `regex-automata`, so the peer's earlier measurements remain specific to their recorded revision and dependency graph. Current [benchmark compilation and lint](https://github.com/apache/datafusion-comet/actions/runs/34211920985/job/102015498633) passed, but that is not a new execution-time measurement. No new material performance issue was established.

## Design

The merge preserves the existing shared segment policy and combines it with the base's consistent NTZ error handling. It introduces no competing parser or normalization route. I found no integration conflict that requires a different design.

## Abstraction & complexity

There is no new authored helper, dependency, configuration or abstraction since the previous approval. The existing fixtures and cast helper remain in place, and the base's extra benchmark helpers do not change this benchmark's input construction. No further complexity change is needed for this revision.

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1636,34 +1641,58 @@ fn extract_offset_suffix(value: &str) -> Option<(&str, Tz)> {
...
+// `SparkDateTimeUtils.parseTimestampString` (`isValidDigits`): the year takes 4-6 digits
+// (`maxDigitsYear = 6`, so "0002020-01-01" is malformed for a timestamp even though
+// `stringToDate`, ported by `date_parser`, allows 7), month/day/hour/minute/second take 1-2
+// digits each, and the fraction takes any number of digits including none ("12:34:56." is
+// valid), of which only the first six are kept. All digits must be ASCII, matching Spark's
+// byte scanner and the numeric parsers used after shape recognition.
+// Keep the ASCII ranges: Unicode `\d` also costs substantially more to match on valid input.
+static RE_YEAR: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^-?[0-9]{4,6}$").unwrap());
+static RE_MONTH: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^-?[0-9]{4,6}-[0-9]{1,2}$").unwrap());
```
sunchao, line 1657: [P1] Guard offset parsing against mixed-width Unicode digits

`2020-1१` (ASCII `1` followed by U+0967, a three-byte digit) matched the base's `\d{2}` month pattern, but no longer matches this ASCII pattern. Both timestamp parsers then try the date's `-` as an offset sign. `parse_sign_offset` sees four bytes after it and `rest[..2]` splits the Unicode character, panicking before Legacy/TRY can return NULL or ANSI can raise CAST_INVALID_INPUT. The exact head array cast panics for both TIMESTAMP and TIMESTAMP_NTZ in all three modes; Spark 4.1.3 rejects the input. Please reject non-ASCII offset bytes before fixed-width slicing (or use checked ranges), and add Parquet-backed column regressions for mixed ASCII/three-byte Unicode month and day digits. The current two-byte Unicode fixtures do not exercise this path.

native/spark-expr/src/conversion_funcs/string.rs
```diff
@@ -1470,7 +1468,14 @@ fn timestamp_parser<T: TimeZone>(
 
     if !has_direct_match {
         if let Some((stripped, suffix_tz)) = extract_offset_suffix(value) {
-            return timestamp_parser_with_tz(stripped, eval_mode, &suffix_tz);
+            // Spark applies Java String.trim to the zone, not Unicode whitespace trimming.
+            let stripped = stripped.trim_end_matches(|c: char| c <= '\u{20}');
```
sunchao, line 1472: [P2] Reject malformed numeric offsets newly admitted by the prefix trim

`2020-01-01 12:34:56 +08:000` now has its space removed here, passes `ends_with_seconds_segment`, and returns `1577853296000000` as TIMESTAMP. The base returns NULL (or an ANSI error), as does Spark 4.1.3; `+008:00` and `+18:01` have the same new behavior. `parse_sign_offset` parses colon-separated parts without a 1–2 digit limit and allows 18 hours plus nonzero minutes before normalizing to an Arrow offset. Could we validate the numeric zone's field widths and ±18:00 bound before accepting it, with separate column inputs under Legacy, TRY and ANSI? This newly accepted spaced TIMESTAMP case is the regression here; unspaced malformed offsets and NTZ acceptance already existed on the base.

sunchao (commented):
## Correctness

Follow-up on `a4c89b34d93958b0a2bceee7b650791ee34d1844` against `bb9e74020adc228e486f6f4d0fa68292b30bff31`. I found two inputs that my September 8 approval did not cover. Both produce new behavior that disagrees with Spark; the reproductions and suggested fixes are inline. I would address them before merging.

The first is a native panic for a mixed ASCII/three-byte Unicode month or day digit. The old parser returned a value for this malformed input, which was already incompatible with Spark, but the new ASCII pattern routes it through an unchecked byte slice and makes the entire cast fail, including in TRY mode. The second is specific to timezone-aware TIMESTAMP with whitespace before a malformed numeric offset: the new trim turns inputs that the base and Spark reject into timestamps. The pre-existing acceptance of unspaced malformed offsets and the NTZ variant is outside this new regression; tightening the shared offset parser can fix those too.

### Validation and CI

I compiled the exact base and head production Arrow array cast functions, their shared batch macro, the full timestamp decoders and offset helpers in focused Rust probes using the cached Arrow 59.3.0, chrono and regex dependencies. Each input was evaluated separately for TIMESTAMP and TIMESTAMP_NTZ in Legacy, TRY and ANSI modes. The mixed Unicode cases change from a value at base to a panic at head. The spaced `+08:000`, `+008:00` and `+18:01` cases change from NULL/ANSI error to a timestamp at head for TIMESTAMP. A valid spaced `+08:00` succeeds at head as intended. Separately, 204 Spark 4.1.3 production `Cast.eval` row evaluations across both timestamp types and all three modes reject the malformed cases (NULL in Legacy/TRY, CAST_INVALID_INPUT in ANSI).

Current CI has 65 successful and eight skipped checks, with no failure or pending check. The Rust job passed 1,197 tests; the Spark 3.5, 4.0 and 4.1 expression jobs passed their segment and zone-whitespace tests. Those cases do not include the inputs above. My local Rust probes execute the production array functions from source, but are not a full JNI/Spark Comet query. A separate full local Cargo test could not start because the configured dependency mirror lacked the locked `aws-smithy-runtime-api 1.16.0`; this is a dependency-resolution limit, not a test failure.

## Performance

The ASCII digit ranges remain valuable: Andy's measurements on the earlier `89e5ff5cc` dependency graph show reduced parser time for common inputs. They are not current-head or whole-query measurements. The suffix check adds recognition work only on that path; I found no new allocation or independently established performance blocker.

## Design and abstraction

Keeping the segment rules and zone placement checks in the existing parser is sensible. The two fixes can stay local: make offset parsing safe for UTF-8 and validate the numeric offset's width and bounds before normalizing it. No new public interface or parser architecture is needed.

sunchao (commented):
## Correctness

Follow-up on `a4c89b34d93958b0a2bceee7b650791ee34d1844` against `bb9e74020adc228e486f6f4d0fa68292b30bff31`. Independent inspection supports both findings in the [new review](https://github.com/apache/datafusion-comet/pull/5682#pullrequestreview-5149633669). My [previous approval](https://github.com/apache/datafusion-comet/pull/5682#pullrequestreview-5140674621) missed these paths. I no longer recommend merging this revision until they are fixed. The head is unchanged, and I am referring to the existing inlines rather than adding duplicates.

- **[P1] The Unicode offset-parsing finding remains.** The changed ASCII patterns can send malformed non-ASCII date segments into the existing offset helper, whose unchecked string slicing can panic. Both TIMESTAMP and TIMESTAMP_NTZ reach that helper before normal null/error handling, so the mode-specific handling does not protect Legacy or TRY. The helper itself is unchanged from the base. The pattern change exposes this path for inputs previously recognized directly. The maintained Spark 3.5 and 4.0 byte scanners reject these malformed date segments, with NULL in Legacy/TRY and a cast error in ANSI. See the [existing P1 inline](https://github.com/apache/datafusion-comet/pull/5682#discussion_r3964548535).
- **[P2] The spaced numeric-offset finding remains.** Removing whitespace from the datetime prefix makes malformed offset widths and offsets exceeding the ±18:00 boundary reach successful TIMESTAMP conversion. The offset helper checks numeric values without fully enforcing the accepted spelling or the boundary, and normalization loses the original spelling. The assigned base rejected these spaced TIMESTAMP forms. Its acceptance of the corresponding unspaced forms and of these NTZ forms is pre-existing, so those are not additional regressions attributed to this PR. Both maintained Spark branches validate the original zone spelling through `getZoneId` before TIMESTAMP conversion or NTZ zone discard. See the [existing P2 inline](https://github.com/apache/datafusion-comet/pull/5682#discussion_r3964548538).

### Validation

This adjudication used exact base/head source and call-flow inspection, maintained Spark 3.5/4.0 scanners and cast wrappers, and the locked Arrow/chrono timezone behavior. A 16-case isolated Java zone-validation oracle passed on JDK 26.0.1, including valid boundary controls and rejection of the reported malformed offset categories. That oracle is not Spark `Cast.eval`. No local Comet/JNI query or panic reproduction was run. Maintained Spark 3.4/4.1 sources remain unavailable.

At **2026-09-09 04:47 UTC**, the current CI snapshot has **65 successful and eight skipped checks**. Refetched logs show [Rust](https://github.com/apache/datafusion-comet/actions/runs/34211920985/job/102015709809), [Spark 3.5 expressions](https://github.com/apache/datafusion-comet/actions/runs/34211920985/job/102027224365), and [Spark 4.0 expressions](https://github.com/apache/datafusion-comet/actions/runs/34211920985/job/102027224406) passed 1,197, 1,326, and 1,334 tests respectively. All three checked out `f065bc25ce3886cad2973549dff62b347f2d7889`, whose parents are the assigned base/head and whose whole tree equals the reviewed head. The current moving merge ref uses a newer base. These results apply to the recorded merge. Existing segment and whitespace tests passed but omit the newly reported categories.

## Performance

There is no source change since the prior approval and no new timed benchmark in this adjudication. Earlier benchmark results remain specific to their recorded revision and dependency graph. The required offset validation can stay in the shared suffix helper. No wider parser redesign is needed to address these findings.

## Design

Validate the original offset safely before normalizing it. Preserve Spark's accepted numeric spellings, enforce the ±18:00 boundary, and return an ordinary malformed-input result for unsupported characters. This retains the useful ASCII segment rules and whitespace compatibility changes while closing the two gaps.

## Abstraction & complexity

Both fixes belong in the existing shared offset parser. Extend the existing column-based tests so each invalid input is checked separately under Legacy, TRY, and ANSI, for both timestamp types, alongside valid offset controls. A second parser or another normalization layer would add unnecessary complexity.

sunchao (approved):
## Correctness

Follow-up at `24803ed4`, including the changes since `a4c89b34`. Both findings from the [previous review](https://github.com/apache/datafusion-comet/pull/5682#pullrequestreview-5149873775) are fixed, and I found no new correctness issue in this update.

- The [P1 offset-parser boundary issue](https://github.com/apache/datafusion-comet/pull/5682#discussion_r3964548535) is addressed by validating the complete signed remainder as ASCII before fixed-width slicing. Both TIMESTAMP and TIMESTAMP_NTZ use this shared helper, so the check applies before their error/null handling.
- The [P2 numeric-offset validation issue](https://github.com/apache/datafusion-comet/pull/5682#discussion_r3964548538) is addressed by checking the original component widths and the absolute ±18:00 limit before normalization. Valid one-digit components, compact offsets and the exact endpoints remain accepted. Rejected suffixes still have to pass the existing complete timestamp grammar and seconds/fraction prefix gate, so the alternate suffix paths do not undo the validation.

The new Parquet-backed regression checks each value separately for both timestamp types in UTC and America/Los_Angeles. It compares legacy CAST, TRY with ANSI both disabled and enabled, and ANSI results/errors. The legacy CAST branch also asserts Comet execution. The new regression and the existing segment/zone-whitespace tests pass in the current [Spark 3.5.9 expression job](https://github.com/apache/datafusion-comet/actions/runs/34352145853/job/102478839403) and [Spark 4.2.0 expression job](https://github.com/apache/datafusion-comet/actions/runs/34352145853/job/102478839480). The [Linux Rust job](https://github.com/apache/datafusion-comet/actions/runs/34352145853/job/102470504588) passes all 1,263 executed tests, including the offset helper and both timestamp parser regressions.

These jobs checked out merge `501e1bfa`, which contains this head but has a different first parent from the assigned base. I verified that all four PR files, the native timestamp cast dispatch and the checked build files are identical to the reviewed head. This is not whole-tree qualification of the assigned head/base pair. At **2026-09-09 13:49 UTC**, 48 checks had succeeded, 12 were running, one was queued and six were skipped. No current check had failed. The remaining CI is still pending. I ran no local runtime test. Canonical source comparison used the maintained Spark 3.5 and 4.0 branches. The required 3.4 and 4.1 branches were unavailable.

## Performance

The update adds one ASCII byte scan over a candidate offset and constant component-width/range checks, with no new allocation. The direct timestamp-shape path and benchmark cases are unchanged from the previous head. This is reasonable work at the validation boundary. The [benchmark compile/lint job](https://github.com/apache/datafusion-comet/actions/runs/34352145853/job/102470243547) passes, but it is not timing evidence. The speedups already discussed on this PR are measurements from an earlier revision. I did not reproduce them or measure this validation increment.

## Design

Checking the original spelling before formatting the normalized timezone is the right place for this repair. It prevents normalization from erasing invalid widths or bounds, and gives the prefixed and bare numeric-offset paths the same rules. The existing separation between timestamp-shape recognition, suffix extraction and conversion remains clear. The new tests cover valid forms alongside the rejected cases, which protects compatibility while tightening validation.

## Abstraction & complexity

The repair stays within the existing shared helper and adds focused tests without introducing another parser abstraction or mode-specific implementation. Both timestamp types receive the same validation policy, while their existing timezone and error behavior remains in their callers. No additional abstraction or refactoring is needed for this follow-up.

sunchao:
Merged, thanks @peterxcli !

peterxcli:
Thanks!

PR #4278, opened 2026-05-10
test: enable nested array cast coverage

manuzhang:
Closes #3906

## Summary
- Enable the nested `ArrayType` cast matrix in `CometCastSuite`.
- Add bounded nested array inputs to avoid recursively generated huge task results.
- Exclude `DecimalType(38,18)` from the nested matrix for the same known decimal-to-floating ULP mismatch as the one-dimensional matrix.
- Replace deprecated `JavaConverters` usage with `scala.jdk.CollectionConverters`.

## Root Cause
The previous nested test used `generateArrays(100, ArrayType(dt))`, which let Spark random data generation build very large nested arrays. The `array<array<string>>` to `array<array<boolean>>` case produced a task result around 401 MiB and failed deserialization with `unread block data`. After bounding the nested inputs, the remaining failure was the existing decimal38-to-floating precision mismatch, so the nested matrix now mirrors the one-dimensional exclusion.

## Tests
- `make core`
- `./mvnw test -Pspark-4.1 -Dtest=none -Dsuites="org.apache.comet.CometCastSuite cast nested ArrayType to nested ArrayType"`

---

Co-authored-by: @codex

PR #5684, opened 2026-09-04
fix: correctly rounded decimal to double/float cast matching BigDecimal.doubleValue/floatValue

### LINKED ISSUE #5670: Decimal to double/float cast is not correctly rounded: last-digit-wrong results for ordinary DECIMAL(38,18) values
### Describe the bug

`CAST(decimal AS DOUBLE)` / `CAST(decimal AS FLOAT)` is delegated to arrow-cast, which computes `(unscaled as f64) / 10^scale` — two roundings for DOUBLE and three for FLOAT (via an f64 intermediate). Spark uses `BigDecimal.doubleValue` / `floatValue`, which round the exact decimal once. Whenever `|unscaled| >= 2^53` and `scale >= 1` the two can differ in the last ulp; for `DECIMAL(38,18)` that is essentially every `|value| >= 0.009` (about 26% of random values differ in a model of arrow's formula). The wrong double is visible in string output, equality filters/joins on the cast value, and sums.

Location: `native/spark-expr/src/conversion_funcs/numeric.rs:64-85` — `is_df_cast_from_decimal_spark_compatible` lists `Float32`/`Float64`, so `cast.rs:402-407` falls through to arrow's cast. Affects LEGACY, ANSI and TRY alike.

### Steps to reproduce

```scala
val p = "/tmp/dec_to_double"
Seq("12345.6789", "123456.789012", "76543.21").toDF("s")
  .selectExpr("CAST(s AS DECIMAL(38,18)) AS c").write.mode("overwrite").parquet(p)
spark.read.parquet(p).selectExpr("CAST(c AS DOUBLE)").show(false)
```

| value | Spark | Comet |
|---|---|---|
| 12345.6789 | 12345.6789 | 12345.678899999999 |
| 123456.789012 | 123456.789012 | 123456.78901200001 |
| 76543.21 | 76543.21 | 76543.20999999999 |

FLOAT: `SELECT CAST(CAST('16777217.0000000001' AS DECIMAL(38,10)) AS FLOAT)` → Spark `1.6777218E7`, Comet `1.6777216E7` (the f64 intermediate is exactly 16777217.0, then ties-to-even).

### Expected behavior

Match Spark, i.e. a correctly rounded conversion of the exact decimal value (`BigDecimal.doubleValue` / `floatValue`).

### Proposed solution

Remove `Float32`/`Float64` from `is_df_cast_from_decimal_spark_compatible` and add a correctly rounded `Decimal128 -> f64/f32` kernel in `numeric.rs`:

- fast path when `|unscaled| < 2^53` and `scale <= 22`: `unscaled as f64 / 10^scale` is then a single correctly rounded operation (exactly Java's fast path);
- otherwise format `<unscaled>e-<scale>` into a stack buffer and parse with Rust's correctly rounded `f64::from_str`; for FLOAT parse the same string with `f32::from_str` directly (never through f64).

Add `DECIMAL(38,18)` values such as `12345.6789` and `76543.21` to `CometNativeCastSuite`; the existing DECIMAL(38,18) cases only use values whose double happens to coincide, which is why CI is green.

### Additional context

PR #4278 excluded DECIMAL(38,18) from the nested cast matrix citing "the same known decimal-to-floating ULP mismatch", but no issue was filed. #3948 added the DECIMAL(38,18) → Float/Double scalar tests with non-diverging values. #286 listed "Decimal to Floating-point — seems correct but needs tests to confirm".

peterxcli:

Closes #5670.

## Rationale for this change

`CAST(decimal AS DOUBLE/FLOAT)` was delegated to arrow-cast, which computes `(unscaled as f64) / 10^scale` and rounds twice (three times for FLOAT, via the f64 intermediate). Spark's `Decimal.toDouble/toFloat` use `BigDecimal.doubleValue()/floatValue()`, which round the exact decimal once, so results differed in the last ulp whenever `|unscaled| > 2^53` — for `DECIMAL(38,18)` that is essentially every value of magnitude >= 0.01 (`12345.6789` became `12345.678899999999`; `CAST(16777217.0000000001 AS FLOAT)` gave `16777216` instead of `16777218`).

- New `Decimal128 -> Float64/Float32` kernels in `numeric.rs` mirroring Java's algorithm: a single IEEE division/multiplication when the unscaled value and the power of ten are both exact in the target type (`|unscaled| <= 2^53` and `scale <= 22` for f64, `<= 2^24` / `<= 10` for f32), otherwise a correctly rounded parse of the exact decimal `<unscaled>e<-scale>` from a stack buffer straight into the target width (never via f64 for FLOAT). Negative scales and out-of-range floats (`±Infinity`) behave like `BigDecimal`.
- `cast.rs` routes `(Decimal128, Float32|Float64)` to the kernels for LEGACY, ANSI and TRY (the conversion cannot fail); `Float32/Float64` are removed from `is_df_cast_from_decimal_spark_compatible`.

- Rust unit tests: the issue's values, ties-to-even cases (`2^53+1`, `2^24+1`), 38-digit extremes, `i128::MIN/MAX`, zero, negative scales, null preservation in every eval mode, and a 30k-value randomized check that the fast paths match the correctly rounded exact path; the tests carry `assert_ne!` guards showing the old `(unscaled as f64)/10^scale` formula fails on the issue's values. Reference values were cross-checked with `jshell` (`BigDecimal.doubleValue()/floatValue()`). `cargo test -p datafusion-comet-spark-expr`: 663 passed; `cargo clippy --all-targets --workspace -- -D warnings` clean.
- `CometNativeCastSuite`: `DECIMAL(38,18)` values from the issue added to the Float/Double tests, new `DECIMAL(38,10) -> Float/Double` tests (including `16777217.0000000001`), and `DecimalType(38,18)` re-enabled in the array and nested-array cast matrices that #4278 had excluded for this mismatch. Suite passes on Spark 4.1.3 (170/170, 8 pre-existing ignores).

### DIFF
--- a/native/spark-expr/src/conversion_funcs/numeric.rs
+++ b/native/spark-expr/src/conversion_funcs/numeric.rs
@@ -890,12 +890,15 @@ const F32_EXACT_INT_LIMIT: u128 = 1 << 24;
 /// magnitude `>= 0.01`) the two roundings can land one ulp away from the correctly rounded
 /// result, e.g. `12345.6789` becomes `12345.678899999999`.
 pub(crate) fn decimal128_to_f64(unscaled: i128, scale: i8) -> f64 {
+    if scale == 0 {
+        return unscaled as f64;
+    }
+
     // Fast path (also Java's): when the unscaled value and the power of ten are both exact
     // doubles, a single IEEE division or multiplication is correctly rounded.
     if unscaled.unsigned_abs() <= F64_EXACT_INT_LIMIT {
         let m = unscaled as f64;
         match scale {
-            0 => return m,
             1..=22 => return m / F64_EXACT_POW10[scale as usize],
             -22..=-1 => return m * F64_EXACT_POW10[scale.unsigned_abs() as usize],
             _ => {}
@@ -911,10 +914,13 @@ pub(crate) fn decimal128_to_f64(unscaled: i128, scale: i8) -> f64 {
 /// rounding twice is observable: `16777217.0000000001` is `16777218` as a float, but narrowing
 /// the nearest `f64` (`16777217.0`, a tie) gives `16777216`.
 pub(crate) fn decimal128_to_f32(unscaled: i128, scale: i8) -> f32 {
+    if scale == 0 {
+        return unscaled as f32;
+    }
+
     if unscaled.unsigned_abs() <= F32_EXACT_INT_LIMIT {
         let m = unscaled as f32;
         match scale {
-            0 => return m,
             1..=10 => return m / F32_EXACT_POW10[scale as usize],
             -10..=-1 => return m * F32_EXACT_POW10[scale.unsigned_abs() as usize],
             _ => {}
--- a/native/spark-expr/src/conversion_funcs/numeric.rs
+++ b/native/spark-expr/src/conversion_funcs/numeric.rs
@@ -68,9 +68,11 @@ pub(crate) fn is_df_cast_from_decimal_spark_compatible(to_type: &DataType) -> bo
             | DataType::Int16
             | DataType::Int32
             | DataType::Int64
-            | DataType::Float32 // DataFusion divides i128 by 10^scale in f64, then narrows to
-            | DataType::Float64 // f32 if needed; empirically matches Spark's BigDecimal.doubleValue
-                                // / floatValue for all tested values
+            // Float32 / Float64 are intentionally absent: DataFusion computes
+            // `(unscaled as f64) / 10^scale`, which rounds twice (three times for f32, via the
+            // f64 intermediate) and can differ from Spark's correctly rounded
+            // BigDecimal.doubleValue() / floatValue() by one ulp once |unscaled| > 2^53.
+            // cast.rs routes Decimal128 sources to cast_decimal128_to_float64 / _float32.
             | DataType::Decimal128(_, _)
             | DataType::Decimal256(_, _)
             // DataFusion's Decimal128→Utf8 cast uses plain notation (toPlainString semantics),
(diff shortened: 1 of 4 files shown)


sunchao (commented):
## Correctness

The prior Arrow path converts the coefficient to `f64` before applying the decimal scale, and narrows through `f64` for FLOAT. Those intermediate roundings can differ from Spark's decimal conversion. This change gives Decimal128-to-DOUBLE/FLOAT dedicated kernels: arithmetic when both operands are exact in the target width, otherwise direct parsing of the exact decimal value. The new dispatch also reaches decimal values inside recursive casts, and the JVM tests restore high-precision decimal coverage.

I compared the implementation with the maintained Spark 3.5 and 4.0 branches (`5947fd6e74a1` and `03f28fc43180`). Both interpreted and generated casts reach `Decimal.toDouble` / `toFloat`, which use BigDecimal. The separate target-width conversions avoid the FLOAT double-rounding problem. The arithmetic bounds are sufficient for one correctly rounded operation, and the parser preserves the coefficient and scale exactly. No correctness finding remained after checking signs, zero, ties, precision boundaries, nulls, empty/sliced arrays, and the scalar and recursive dispatch paths.

LEGACY, ANSI and TRY share these decimal-to-floating semantics; floating overflow produces infinity rather than a decimal overflow error. Negative scales are supported when `spark.sql.legacy.allowNegativeScaleOfDecimal` is enabled, so FLOAT infinity is reachable there. FLOAT subnormals are reachable at scale 38. DOUBLE overflow/subnormals are outside the Decimal128/i8-scale range. Other source/target fallback decisions are unchanged; removing FLOAT/DOUBLE from the Arrow-compatibility matrix accompanies the dedicated dispatch.

### Validation and scope

Reviewed HEAD `f977e8f4a2ec74bd7a598f02586d5974e8afadd5` against BASE `81d637b9bf40a5be6f4f0c65ad6f497b34746e69`. An isolated harness containing the unchanged new kernels matched Java BigDecimal 26.0.1 and an independent integer-rational reference for 367,682 inputs at both widths, with zero bit mismatches; 128 Arrow array checks passed. Cases distinguish valid Spark decimals from extended i128/i8 stress inputs. This was component validation using Arrow 58.4.0 and its own lockfile, not a local full Comet or Spark build.

CI ran merge `b561865f30998ca62872747195d7af25510e94de`, whose parents are the exact base/head above. The changed files and native lockfile are identical to HEAD. The [Rust job](https://github.com/apache/datafusion-comet/actions/runs/33876959697/job/101037053646) passed all four added tests (1,115 passed, 4 skipped overall). Successful [Spark 3.5/JDK 17](https://github.com/apache/datafusion-comet/actions/runs/33876959697/job/101045907576) and [Spark 4.0/JDK 21](https://github.com/apache/datafusion-comet/actions/runs/33876959697/job/101045907799) logs include the decimal-to-FLOAT/DOUBLE cases and array-cast suite. The review snapshot at 2026-09-04T17:24:25.103Z contains 65 successful and 9 skipped checks.

## Performance

One P2 finding: handle scale zero before the exact-integer magnitude guard in both scalar helpers. Larger `DECIMAL(p,0)` coefficients currently incur formatting and parsing for each value even though a direct cast to the target float already performs the required single rounding. The inline comment identifies the branch and proposed change.

Matched Arrow-array benchmarks measured the following median nanoseconds per input slot, including output allocation:

| Cast, no nulls | Prior Arrow | PR kernel | Scale-zero direct-cast candidate |
| --- | ---: | ---: | ---: |
| DECIMAL(18,0) → DOUBLE | 1.53 | 33.49 | 1.53 |
| DECIMAL(38,0) → DOUBLE | 1.51 | 59.27 | 1.52 |
| DECIMAL(12,0) → FLOAT | 1.53 | 35.01 | 1.63 |

The candidate remained about 18× and 31× faster for the two DOUBLE datasets with approximately 20% nulls. Direct target casts matched the PR, Java and the rational reference on 162,083 scale-zero inputs, including midpoint cases. FLOAT must cast directly from `i128` to `f32`; using the old `f64` intermediate can reintroduce rounding differences.

These are component timings on Apple M5 Max, rustc 1.97.1 release and Arrow 58.4.0: 8,192-element arrays, nine alternating-order rounds, varied signed inputs, black-boxed inputs/outputs, and at least 25 ms calibration per strategy. They are not end-to-end Spark speedups. The candidate keeps nonzero-scale behavior unchanged, with similar control timings. Parsing on the general nonzero-scale path also costs substantially more than the prior arithmetic, but that arithmetic is not a correctness-preserving replacement.

## Design

The dedicated kernels are a sensible boundary for Spark-specific rounding semantics. Keeping FLOAT separate from DOUBLE is necessary, and the exact-arithmetic fast path plus exact-decimal fallback is straightforward to audit. The scale-zero change is a small local improvement within this design; it requires no new dependency or conversion framework.

The fixed 48-byte buffer is sufficient: an i128 sign and 39 digits, exponent separator, and four exponent characters require at most 45 bytes. Widening the scale before negation handles `i8::MIN`. Valid ASCII decimal literals make the private parser's error expectations defensible. Negative scales work even though one comment describes only the default nonnegative-scale case.

## Abstraction & complexity

The two small Arrow wrappers share one private formatter/parser and reuse Arrow's null-preserving unary operation. This keeps the Spark-specific behavior close to the conversion dispatch without introducing wider state or indirection. The generic parser has only the two intended float instantiations, which earns the small amount of abstraction. No additional design or abstraction finding was identified.

native/spark-expr/src/conversion_funcs/numeric.rs
```diff
@@ -863,6 +865,103 @@ pub(crate) fn spark_cast_decimal_to_boolean(array: &dyn Array) -> SparkResult<Ar
...
+/// matching Spark's `Decimal.toDouble`, i.e. `java.math.BigDecimal.doubleValue()`.
+///
+/// DataFusion's kernel computes `(unscaled as f64) / 10^scale`, which rounds the unscaled value
+/// first and the quotient second; once `|unscaled| > 2^53` (any `DECIMAL(38,18)` value of
+/// magnitude `>= 0.01`) the two roundings can land one ulp away from the correctly rounded
+/// result, e.g. `12345.6789` becomes `12345.678899999999`.
+pub(crate) fn decimal128_to_f64(unscaled: i128, scale: i8) -> f64 {
+    // Fast path (also Java's): when the unscaled value and the power of ten are both exact
+    // doubles, a single IEEE division or multiplication is correctly rounded.
+    if unscaled.unsigned_abs() <= F64_EXACT_INT_LIMIT {
```
sunchao, line 899: ### Performance

[P2] Handle scale zero before the exact-integer guard

Please handle `scale == 0` before this magnitude guard in both helpers. Above `2^53` (`2^24` for FLOAT), `DECIMAL(p,0)` currently formats and parses every value, although a direct `i128 as f64` / `i128 as f32` already performs the required [single, correctly rounded conversion](https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast). In matched 8,192-element Arrow 58.4.0 array benchmarks (M5 Max, rustc 1.97.1 release), `DECIMAL(18,0) → DOUBLE` rose from 1.53 ns/input slot in Arrow to 33.49 ns here; handling scale zero first took 1.53 ns. For `DECIMAL(38,0)`, this was 59.27 ns versus 1.52 ns, with the gain persisting with nulls. Direct target casts matched HEAD, Java BigDecimal and an exact rational reference on 162,083 scale-zero inputs, including ties. This avoids the regression while retaining the exact nonzero-scale path. FLOAT should cast directly to `f32`, without an `f64` intermediate. These are component timings, not end-to-end Spark speedups. Please add scale-zero FLOAT/DOUBLE microbenchmark cases, including nulls, to keep this fast path covered.

peterxcli: Addressed in d410f61d8. Both helpers now handle scale zero before the magnitude guard with direct target-width casts (`i128 as f64` and `i128 as f32`). I also added `cast_numeric` Criterion cases for DECIMAL(18,0) to DOUBLE, DECIMAL(38,0) to DOUBLE, and DECIMAL(12,0) to FLOAT, each with and without nulls. A local release run measured 1.71-1.99 ns/input slot across the six cases. `cargo test -p datafusion-comet-spark-expr --lib` passed 663 tests, and package Clippy with all targets passed with warnings denied.

sunchao (approved):
## Correctness

The [existing P2](https://github.com/apache/datafusion-comet/pull/5684#discussion_r3936454700) is addressed at `d410f61d`. On maintained Spark 3.5 and 4.0, decimal-to-FLOAT/DOUBLE delegates to BigDecimal's target-width conversion. At scale zero the coefficient already is the value, and the new direct `i128 as f32` / `i128 as f64` returns round once. FLOAT does not pass through DOUBLE. Both signs, zero and precision boundaries are preserved. Valid scale-zero Decimal128 values remain finite. Nonzero-scale conversion, null handling, LEGACY/ANSI/TRY dispatch and fallback eligibility are unchanged.

A fresh dependency-free check of the exact new scalar helpers matched Java 17 BigDecimal and an independent integer-rational reference for all 367,682 inputs at both widths, including 162,083 scale-zero cases. The corpus includes midpoint boundaries and separately identified inputs beyond Spark's decimal domain. This is component validation, not a local full Spark/Comet suite. Canonical-source comparison covers maintained 3.5/4.0. Maintained 3.4/4.1 refs were unavailable.

The [Rust CI job](https://github.com/apache/datafusion-comet/actions/runs/33914166009/job/101158110235) passed 1,133 tests, with four skipped, including decimal rounding and null-preservation tests in every eval mode. At the September 4, 20:45:28 UTC snapshot, benchmark compile/lint had passed and broader CI was running (24 successful checks, 41 running, six skipped). That job checked out merge `a17ce159`, whose base parent is `719cba11`, while the reviewed API base remains `81d637b9`. The decimal helper bodies and benchmark match this head, but the complete numeric module and JVM suite differ. This is qualified integration evidence, not an exact head/base full-suite result. No new correctness finding.

## Performance

The scale-zero path now avoids decimal formatting and parsing for large coefficients. It is equivalent to the candidate measured during the earlier review: no-null Decimal(18,0)/Decimal(38,0)-to-DOUBLE medians were approximately 33.49/59.27 ns per input slot before the shortcut and 1.53/1.52 ns with it. Decimal(12,0)-to-FLOAT was 35.01 versus 1.63 ns. These are reused matched component measurements, not fresh timings of this commit or end-to-end Spark gains.

The six added Criterion cases cover those three conversions with no nulls and approximately 20% nulls. They evaluate the actual `Cast` expression, include result allocation, and black-box both input and result. Dataset setup stays outside the timed loop. This supplies the requested durable benchmark coverage without changing the nonzero-scale algorithm.

## Design

Moving the scale check ahead of the magnitude guard is the appropriate local fix: the exact-integer restriction is needed for multiplication/division by powers of ten, while scale zero needs only one target-width conversion. Removing the now-unreachable zero match arms keeps that distinction clear without changing dispatch or configuration.

## Abstraction & complexity

No new production abstraction or dependency is introduced. The two existing scalar helpers gain simple early returns, and the benchmark uses a small shared batch constructor with valid declared precisions. The additional code stays proportional to the regression it covers.

sunchao:
Merged, thanks @peterxcli !

Issue #5572, opened 2026-08-31
[EPIC] Codegen-dispatch coverage audit: expressions that fall back to Spark where the JVM dispatcher would work


andygrove:
## What / Why

Comet has a JVM codegen dispatcher (`CometScalaUDF.emitJvmCodegenDispatch`, `spark/src/main/scala/org/apache/comet/serde/CometScalaUDF.scala:70`) that compiles a Spark expression's own `doGenCode` output into a per-batch kernel reading and writing Arrow vectors directly. When a serde has no native path for some input, routing through the dispatcher keeps the whole operator inside Comet and still matches Spark byte for byte, instead of failing the enclosing projection back to Spark.

A serde opts in by mixing in `CodegenDispatchFallback` (`spark/src/main/scala/org/apache/comet/serde/CometExpressionSerde.scala:125`). `QueryPlanSerde` then tries the dispatcher for that serde's `Unsupported` and non-opt-in `Incompatible` results (`QueryPlanSerde.scala:941` and `:968`) before giving up.

Adoption is uneven. I swept all 261 expression serdes under `spark/src/main/scala/org/apache/comet/serde/` plus the version shims. 98 reach the dispatcher; 63 decline at least one case without one. This EPIC collects the cases where the mixin is applicable and worth adding, and — just as importantly — records the cases where it is *not* applicable, so they aren't re-litigated.

Several of the gaps are plain asymmetries against an already-dispatched sibling: `bround` dispatches but `round` on a double falls back; `to_unix_timestamp` dispatches but `unix_timestamp` on a string falls back; `to_json`, `from_csv` and `schema_of_csv` dispatch but `to_csv` never runs in Comet by default at all.

## What gates a dispatch

`CometBatchKernelCodegen.canHandle` (`spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala:119`) rejects `AggregateFunction`, `Generator`, `Unevaluable`, and any bound reference or output type outside `isSupportedDataType` (`:85`) — notably `NullType`, `ObjectType` and Variant. Everything else is admitted, including `CodegenFallback`, nondeterministic and stateful expressions, `HigherOrderFunction`, and subquery expressions.

Every item below was checked against that gate.

## Prerequisites

Both should land before the catch-all in the `StaticInvoke` item below.

- #5573 — guard the closure-serialize step in `emitJvmCodegenDispatch` so a non-serializable tree falls back cleanly instead of throwing at plan time
[...]

PR #5692, opened 2026-09-04
feat: route unrecognized StaticInvoke and Invoke through the codegen dispatcher

### LINKED ISSUE #5573: Codegen dispatcher: guard the closure-serialize step so a non-serializable tree falls back cleanly instead of throwing at plan time
Part of #5572.

`CometScalaUDF.emitJvmCodegenDispatch` closure-serializes the bound expression with no guard:

```scala
val serializer = SparkEnv.get.closureSerializer.newInstance()
val buffer = serializer.serialize(boundExpr)
```

`spark/src/main/scala/org/apache/comet/serde/CometScalaUDF.scala:111-112`

Every other failure mode in that method degrades cleanly — the disabled-config check, `CometBatchKernelCodegen.canHandle`, the literal-payload conversion, the data-arg conversion and the return-type conversion all call `withFallbackReason` and return `None`, which the caller treats as a clean Spark fallback. The serialize call is the one step that can throw instead, and a `NotSerializableException` there escapes during **planning**, which is a much worse failure than falling back.

Today's callers are all curated expression types where this is very unlikely to bite. It matters because of where the sweep wants to go next: routing an unrecognized `StaticInvoke` or `Invoke` through the dispatcher as a catch-all hands it arbitrary trees, including ones holding references the closure serializer will refuse. So this should land before that change.

The fix is small — wrap the serialize in a `try` / `catch`, call `withFallbackReason` with the exception message and return `None`, matching the shape of the checks around it. A unit test can force it with an expression capturing a deliberately non-serializable object.

### LINKED ISSUE #5575: Route unrecognized `StaticInvoke` and `Invoke` through the codegen dispatcher instead of falling back
Part of #5572.

`CometStaticInvoke` dispatches on an allowlist of `(functionName, staticObject)` pairs and falls the whole operator back to Spark for anything not in it:

```scala
case None =>
  withFallbackReason(
    expr,
    s"Static invoke expression: ${expr.functionName} is not supported")
  None
```

`spark/src/main/scala/org/apache/comet/serde/statics.scala:61-65`

`StaticInvoke` is a codegen-friendly node — `doGenCode` emits a static method call — and `CometStaticInvokeCodegenDispatch` (`statics.scala:120`) already exists and is wired for `aesEncrypt`, `aesDecrypt` and `StringDecode.decode`. Making it the `case None` default turns a hard fallback into in-pipeline execution for every lowering nobody has gotten to yet.

The compatibility guide already records the symptom. `docs/source/user-guide/latest/expressions.md` lists `encode` as "🔜 Lowers to `StaticInvoke(encode)` (not allowlisted); falls back", and `to_binary` as "Hex form accelerated; other formats fall back". Both would be covered by the default.

The same applies to `Invoke`, which has no entry in the serde map at all. `Spark4xCometExprShim.sparkVersionSpecificExprToProtoInternal` recognizes `StructsToJsonEvaluator` and `ParseUrlEvaluator` and returns `None` for everything else (`spark/src/main/spark-4.x/org/apache/comet/shims/Spark4xCometExprShim.scala:94`), after which the generic path reports "not supported". Spark 4.x lowers a growing number of `RuntimeReplaceable` expressions to evaluator-backed `Invoke` nodes, so the same catch-all is worth having there.

This is safe by construction for the encoder and deserializer trees that make up most `StaticInvoke` usage in typed Dataset operations: their arguments and return types are `ObjectType`, which is outside `CometBatchKernelCodegen.isSupportedDataType`, so `canHandle` rejects them and they fall back exactly as they do now.

Depends on the closure-serialize guard, since this is the change that starts handing the dispatcher arbitrary trees.

andygrove:

Closes #5573.
Closes #5575.

Both are children of the codegen-dispatch coverage audit, #5572.

## Rationale for this change

`CometStaticInvoke` dispatches on an allowlist of `(functionName, staticObject)` pairs and fails the whole operator back to Spark for anything else. `Invoke` has no entry in the serde map at all. Both nodes are codegen-friendly — `StaticInvoke.doGenCode` emits a static method call, `Invoke.doGenCode` emits a method call on the target object — so routing them through the JVM codegen dispatcher runs Spark's own implementation inside the Comet pipeline and matches Spark by construction, instead of losing the whole projection for every lowering nobody has allowlisted yet.

Spark keeps adding these. Three concrete cases that were fallbacks before this PR:

- `lpad` / `rpad` on **binary** input lower to `StaticInvoke(ByteArray, funcName, ...)` on every supported Spark version (3.4 through 4.1).
- Spark 4.1's `to_time(str, fmt)` lowers to an evaluator-backed `Invoke`.
- `encode`, and the `utf-8` form of `to_binary`, lower to `StaticInvoke(Encode.encode, ...)` on Spark 4.0+. The compatibility guide already recorded both as falling back for exactly this reason.

#5573 is the prerequisite: the closure-serialize in `emitJvmCodegenDispatch` was the one failure mode in that method that threw during planning rather than degrading to a Spark fallback, and a catch-all is what starts handing the dispatcher arbitrary trees.

**#5573** — `CometScalaUDF.emitJvmCodegenDispatch` wraps the closure-serialize in `try` / `catch NonFatal`, tags a fallback reason and returns `None`, matching the shape of every other check around it. `NonFatal` rather than `NotSerializableException` because Java serialization reports an unserializable object graph as one of several exception types depending on where it trips, and a custom `writeObject` can throw anything. The scaladoc now states that the method never throws.

**#5575** — `CometStaticInvoke.convert`'s `case None` routes to `CometStaticInvokeCodegenDispatch`; if the dispatcher declines (config off, or `canHandle` refuses the tree), it still tags a fallback reason naming the function and the declaring class, so the diagnostic that existed before is preserved. New `CometInvoke extends CometCodegenDispatch[Invoke]`, registered as `classOf[Invoke]` in `QueryPlanSerde.miscExpressions`.

**Docs** — `expressions.md` rows for `encode` and `to_binary`, and the `lpad` / `rpad` known-limitation notes in the string expression audit.

### One deliberate deviation from #5572

#5572 frames these items as "add the `CodegenDispatchFallback` mixin". I implemented that variant first for `CometStaticInvoke` and it **produces wrong answers**, so this PR scopes the catch-all to functions with no handler at all.

The mixin also rescues a *handler's* `Unsupported`. `CometIcebergTruncate` declines a decimal because Iceberg's `truncate` can return a value wider than the column's declared precision, which Spark turns into null only when the row is materialized. The dispatcher writes into the same Arrow `Decimal128(p, s)` vector a native kernel would, so the rescued plan returned `-100000000000000.0000` for a row Spark nulls, in `CometIcebergSystemFunctionSuite`'s "truncate on a decimal falls back to Spark".

The mixin's stated contract is that the case "must be something `Expression.doGenCode` can compile". That does not cover a limit living at the **Arrow output boundary**, which the dispatcher and a native kernel share. Worth checking each remaining child issue's `Unsupported` reason against that distinction. The reasoning is recorded in `CometStaticInvoke`'s scaladoc and in the Iceberg test so it is not re-litigated.

### Residual risk, not addressed here

The dispatcher Janino-compiles at execute time with no recovery, so a `StaticInvoke` whose target class is not public would fail the task where Spark's whole-stage codegen falls back to interpreted evaluation. I could not exhibit one — Spark's own lowerings all use public utility classes, and the DSv2 magic-method contract effectively requires a public class since Spark's own WSCG emits the same call. A plan-time guard would have to replicate `StaticInvoke.doGenCode`'s name mangling, and getting that approximation wrong would reject valid dispatches, which is the worse failure. Flagging it rather than guessing.

Encoder and deserializer trees — which are most `StaticInvoke` / `Invoke` usage in typed Dataset operations — are unaffected either way: their arguments are `ObjectType`, which `CometBatchKernelCodegen.isSupportedDataType` rejects, so `canHandle` declines them and they fall back exactly as before.

New tests in `CometCodegenSuite`:

- `dispatch falls back cleanly when the bound tree cannot be closure-serialized (#5573)` — an `Invoke` holding a deliberately non-`Serializable` target. `canHandle` greenlights the tree (string in, string out), so the serializer is the step that refuses, and the assertion is on the fallback reason rather than on an escaping exception.
- `unrecognized StaticInvoke routes through the dispatcher instead of falling back (#5575)` — `lpad` on a binary column: asserts the dispatcher ran and the operator stayed native, then asserts a clean fallback with the dispatcher disabled.
- `Invoke routes through the codegen dispatcher (#5575)` — pins the serde-map registration by going through `QueryPlanSerde.exprToProto`, then runs the compiled kernel to check the emitted method call evaluates.

Existing tests updated to the new behaviour:

- `CometStringExpressionSuite`'s binary `lpad` / `rpad` flip from `checkSparkAnswerAndFallbackReason` to `checkSparkAnswerAndOperator`.
- `expressions/datetime/to_time.sql` flips two `expect_fallback(invoke is not supported)` blocks to plain `query`.
- `CometIcebergSystemFunctionSuite`'s unlisted-static-invoke test now asserts the dispatch proto, and that the declaring-class diagnostic survives when the dispatcher is off.

Regression runs, all on this branch:

| Scope | Result |
| --- | --- |
| `CometExpressionSuite`, `CometCodegenFuzzSuite`, `CometSqlFileTestSuite`, `CometExecSuite` (4.1) | 782 passed |
| `CometCodegenSuite`, `CometStringExpressionSuite`, `CometIcebergSystemFunctionSuite`, `CometSqlFileTestSuite` (4.1) | 602 passed |
| `CometStringExpressionSuite`, `CometCodegenSuite` (3.4 and 3.5) | 121 passed each |
| `CometStringExpressionSuite`, `CometCodegenSuite`, `CometSqlFileTestSuite` (4.0) | 590 passed |

### Benchmark

`CometCodegenDispatchBenchmark` (added in response to review) times every case this PR newly dispatches against the operator fallback it replaces: three arms per case (codegen dispatch, `spark.comet.exec.scalaUDF.codegen.enabled=false`, Comet off), at 1024 and 1,048,576 rows, with the rows compared across all three arms before anything is timed and a repeat of the baseline closing each table as a noise floor. First-use compilation gets its own table so it does not contaminate the steady-state numbers.

At 1M rows a projection containing nothing but the dispatched call runs between parity and ~20% slower than falling the projection back (`lpad` 51 -> 61 ms is the worst); a projection that mixes it with native expressions is 1.4x (137 -> 101 ms), and a grouped aggregate over it is 1.1x (73 -> 65 ms) because the exchange and both aggregates stay native. At 1024 rows every arm is inside the noise floor. Full numbers are in the review thread.

Cross-compiled against Spark 3.4 / 3.5 / 4.0 / 4.1; spotless and scalastyle clean; `mvn package` regenerates the docs with no churn beyond the note edits above.

### DIFF
--- a/spark/src/main/scala/org/apache/comet/serde/statics.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/statics.scala
@@ -20,7 +20,7 @@
 package org.apache.comet.serde
 
 import org.apache.spark.sql.catalyst.expressions.{Attribute, Base64, ExpressionImplUtils, Literal, StringDecode, TryEval, UrlCodec}
-import org.apache.spark.sql.catalyst.expressions.objects.StaticInvoke
+import org.apache.spark.sql.catalyst.expressions.objects.{Invoke, StaticInvoke}
 import org.apache.spark.sql.catalyst.util.CharVarcharCodegenUtils
 import org.apache.spark.sql.types.StringType
 
@@ -59,6 +59,23 @@ object CometStaticInvoke extends CometExpressionSerde[StaticInvoke] {
   private def handlerFor(expr: StaticInvoke): Option[CometExpressionSerde[StaticInvoke]] =
     staticInvokeExpressions.get((expr.functionName, expr.staticObject.getName))
 
+  /** Every Iceberg system function is named `invoke`, so name the declaring class too. */
+  private def noNativePathNote(expr: StaticInvoke): String =
+    s"Static invoke expression: ${expr.functionName} has no native implementation and the " +
+      s"codegen dispatcher declined it (declared on ${expr.staticObject.getName})"
+
+  /**
+   * A `StaticInvoke` outside the allowlist reports `Compatible` and is handled in [[convert]],
+   * which routes it through the codegen dispatcher. Deliberately not `Unsupported` +
+   * [[CodegenDispatchFallback]]: that would also route a *handler's* `Unsupported` through the
+   * dispatcher, and at least one of those is not dispatchable. `CometIcebergTruncate` declines a
+   * decimal because Iceberg's `truncate` can return a value wider than the column's declared
+   * precision, which Spark nulls only when the row is materialized; the dispatcher writes into an
+   * Arrow `Decimal128(precision, scale)` vector just like a native kernel does, so it produces
+   * the out-of-range value instead of a null. The mixin's contract ("the case must be something
+   * `doGenCode` can compile") does not cover a limit that lives at the Arrow output boundary, so
+   * enrollment stays with the individual handlers.
+   */
   override def getSupportLevel(expr: StaticInvoke): SupportLevel =
     handlerFor(expr).map(_.getSupportLevel(expr)).getOrElse(Compatible())
 
@@ -78,15 +95,37 @@ object CometStaticInvoke extends CometExpressionSerde[StaticInvoke] {
       case Some(handler) =>
         handler.convert(expr, inputs, binding)
       case None =>
-        // Every Iceberg system function is named `invoke`, so name the declaring class too.
-        withFallbackReason(
-          expr,
-          s"Static invoke expression: ${expr.functionName} is not supported " +
-            s"(declared on ${expr.staticObject.getName})")
-        None
+        // Nothing in the allowlist covers this lowering, so run Spark's own implementation inside
+        // the Comet pipeline rather than failing the whole operator back to Spark.
+        // `StaticInvoke.doGenCode` emits a static method call, so the kernel matches Spark by
+        // construction. Spark 4.x keeps lowering more `RuntimeReplaceable` functions this way
+        // (`encode`, `is_valid_utf8`, the `TIME` family, ...) and `lpad` / `rpad` on binary has
+        // lowered to `StaticInvoke(ByteArray, ...)` since Spark 3.4.
+        //
+        // The encoder and deserializer trees that make up most `StaticInvoke` usage in typed
+        // Dataset operations are unaffected: their arguments are `ObjectType`, which
+        // `CometBatchKernelCodegen.isSupportedDataType` rejects, so the dispatcher declines them
+        // and they fall back exactly as before.
+        CometStaticInvokeCodegenDispatch.convert(expr, inputs, binding).orElse {
+          // The dispatcher tags its own reason, but not which static invoke it was.
+          withFallbackReason(expr, noNativePathNote(expr))
+          None
+        }
     }
   }
 }
+
+/**
+ * Catch-all for `Invoke`, which has no allowlist at all. Spark 4.x lowers a growing number of
+ * `RuntimeReplaceable` expressions to evaluator-backed `Invoke` nodes; the Spark 4.x shim
+ * reconstructs the handful it recognizes and everything else lands here. `Invoke.doGenCode` emits
+ * a method call on the target object, so the dispatcher runs Spark's own implementation inside
+ * the Comet pipeline rather than failing the operator back to Spark.
+ *
+ * As with [[CometStaticInvoke]], the object-typed `Invoke` nodes in encoder / deserializer trees
+ * are rejected by `CometBatchKernelCodegen.canHandle` and fall back as before.
+ */
+object CometInvoke extends CometCodegenDispatch[Invoke]
 
 object CometUrlEncodeStaticInvoke extends CometExpressionSerde[StaticInvoke] {
   override def convert(
--- a/docs/source/contributor-guide/expression-audits/string_funcs.md
+++ b/docs/source/contributor-guide/expression-audits/string_funcs.md
@@ -137,7 +137,7 @@
 - Spark 3.4.3 (audited 2026-05-27): identical to 3.5.8.
 - Spark 3.5.8 (audited 2026-05-27): baseline. `StringLPad(str, len, pad) -> StringType`; `len <= 0` returns the empty string, empty `pad` returns `str` unchanged, NULL inputs propagate. Comet serde requires `str` to be a column and `pad` to be a literal; otherwise falls back.
 - Spark 4.0.1 (audited 2026-05-27): `NullIntolerant` trait replaced by `override def nullIntolerant: Boolean = true`; `inputTypes` widened to `StringTypeWithCollation(supportsTrimCollation = true)`. Semantics unchanged for `UTF8_BINARY`. Non-default collations not honoured by Comet ([#4496](https://github.com/apache/datafusion-comet/issues/4496)).
-- Known limitation: `lpad(<binary>, ...)` is rewritten by Spark to `BinaryPad / StaticInvoke(ByteArray.lpad)` before serde runs and always falls back to Spark.
+- `lpad(<binary>, ...)` is rewritten by Spark to `BinaryPad / StaticInvoke(ByteArray.lpad)` before serde runs. There is no native path, so `CometStaticInvoke` routes it through the JVM codegen dispatcher and the projection stays in the Comet pipeline; it falls back to Spark only when the dispatcher is disabled.
 
 ## ltrim
 
@@ -182,7 +182,7 @@
 - Spark 3.4.3 (audited 2026-05-27): identical to 3.5.8.
 - Spark 3.5.8 (audited 2026-05-27): baseline. `StringRPad(str, len, pad) -> StringType`; same edge-case behaviour as `lpad` (negative len, empty pad, NULL propagation). Comet serde requires column `str` and literal `pad`.
 - Spark 4.0.1 (audited 2026-05-27): same evolution as `lpad`; default-pad literal type tightened; semantics unchanged for `UTF8_BINARY`. Non-default collations not honoured by Comet ([#4496](https://github.com/apache/datafusion-comet/issues/4496)).
-- Known limitation: same `BinaryPad / StaticInvoke` rewrite as `lpad` causes `rpad(<binary>, ...)` to fall back.
+- Same `BinaryPad / StaticInvoke` rewrite as `lpad`, so `rpad(<binary>, ...)` also runs through the codegen dispatcher rather than falling back.
 
 ## rtrim
 
--- a/docs/source/user-guide/latest/expressions.md
+++ b/docs/source/user-guide/latest/expressions.md
@@ -565,7 +565,7 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `contains` | ✅ | — |  |
 | `decode` | ✅ | — |  |
 | `elt` | ✅ | Codegen dispatch |  |
-| `encode` | 🔜 | — | Lowers to `StaticInvoke(encode)` (not allowlisted); falls back |
+| `encode` | ✅ | — | Spark 4.0+ lowers to `StaticInvoke(Encode.encode)`, which runs via codegen dispatch; on Spark 3.x it falls back |
 | `endswith` | ✅ | — |  |
 | `find_in_set` | ✅ | Codegen dispatch |  |
 | `format_number` | ✅ | Codegen dispatch |  |
@@ -606,7 +606,7 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `substr` | ✅ | Native |  |
 | `substring` | ✅ | Native |  |
 | `substring_index` | ✅ | Native |  |
-| `to_binary` | ✅ | — | Hex form accelerated; other formats fall back |
+| `to_binary` | ✅ | — | Hex and base64 forms accelerated natively; the `utf-8` form rewrites to `encode` and runs via codegen dispatch on Spark 4.0+ |
 | `to_char` | ✅ | Codegen dispatch |  |
 | `to_number` | ✅ | Codegen dispatch |  |
 | `to_varchar` | ✅ | Codegen dispatch |  |
--- a/spark/src/main/scala/org/apache/comet/serde/CometScalaUDF.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/CometScalaUDF.scala
@@ -19,6 +19,8 @@
 
 package org.apache.comet.serde
 
+import scala.util.control.NonFatal
+
 import org.apache.spark.SparkEnv
 import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, AttributeSeq, BindReferences, Expression, Literal, RuntimeReplaceable, ScalaUDF}
 import org.apache.spark.sql.types.BinaryType
@@ -63,9 +65,9 @@ object CometScalaUDF extends CometExpressionSerde[ScalaUDF] {
    * batch kernel on first invocation per task.
    *
    * Returns `None` (with `withFallbackReason` tagging the reason) when the dispatcher is disabled
-   * via [[CometConf.COMET_SCALA_UDF_CODEGEN_ENABLED]] or when
-   * [[CometBatchKernelCodegen.canHandle]] refuses the expression tree. Callers should treat
-   * `None` as a clean Spark-fallback signal.
+   * via [[CometConf.COMET_SCALA_UDF_CODEGEN_ENABLED]], when [[CometBatchKernelCodegen.canHandle]]
+   * refuses the expression tree, or when the bound tree cannot be closure-serialized. Callers
+   * should treat `None` as a clean Spark-fallback signal; this method never throws.
    */
   def emitJvmCodegenDispatch(
       expr: Expression,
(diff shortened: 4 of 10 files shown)


sunchao (commented):
## Correctness

Reviewed head `517c46687d4e421cee70cf88065b8c5a85571886` against base `719cba11076e4237d6030925c49b1c1ffcac6f8e`. No correctness defect was verified within the scope below. The performance-validation request is described in the inline P2.

Previously, an unrecognized `StaticInvoke` fell back and `Invoke` had no general serde. The change sends these expressions through the existing JVM codegen dispatcher. Comparing the maintained Spark 3.5 and 4.0 implementations, retaining Spark's generated calls preserves argument evaluation, primitive boxing, target-null checks, and nullable results. Binary padding and Spark 4.0 `encode` use this static-call path. UTF-8 `to_binary` lowers through `Encode`.

The distinction between an unknown function and a known handler's decline is necessary. `CometStaticInvoke` does not acquire `CodegenDispatchFallback`, so the outer support gate still honors explicit `Unsupported` decisions, including Iceberg decimal truncation. Reusing Spark's generated call alone does not reproduce Spark's decimal materialization checks at the Arrow output boundary. Disabled dispatch and serialization refusal retain a Spark-fallback result. The added `NonFatal` guard covers the serialization operation without swallowing fatal JVM errors.

Independent component checks on Java 17 compiled the exact-head kernel sources and matched 60 padding/encoding row-or-error comparisons on Spark 4.0.4, plus three rows through a serialized public receiver. On Spark 4.1.3, serialized `to_time` evaluator kernels matched 12 rows with literal/changing format, null input/null format, and fractional seconds. These are component checks using public Spark runtimes and some existing Comet infrastructure, not a full Comet SQL/native test run. Canonical source comparisons cover maintained 3.5/4.0. Maintained 3.4/4.1 sources were unavailable. The PR's new planning, native-plan, and disabled-dispatch tests were inspected, not independently run locally.

CI at `2026-09-04T20:03:46.453Z` remains incomplete: 51 successful checks, 7 skipped, 13 running, and one failure. The [Iceberg extensions job](https://github.com/apache/datafusion-comet/actions/runs/33907019200/job/101148049297) failed on an HTTP 502 fetching a Maven site-plugin dependency, before compiling Comet modules or running Iceberg tests. The log pins synthetic merge `c2b14eed`, whose tree differs from the standalone head. This is infrastructure-failure evidence, not a failing expression test or a complete CI pass.

## Performance

The added work is concentrated in expressions that previously forced fallback. Known native handlers retain their direct path, and disabled dispatch returns before binding and serialization. The generic route still pays planning-time tree serialization, payload transport, first-use Janino compilation, per-batch argument/schema work, and Arrow output allocation. The existing per-task kernel cache amortizes compilation across batches. This PR does not add another cache or per-row reflection.

No matched performance results accompany these newly enabled paths, and the correctness probes establish no speedup. The inline P2 asks for small and multi-batch comparisons against the existing Spark fallback so we can assess whether avoiding operator fallback outweighs the bridge costs for inexpensive functions. This is a validation gap, not a measured-regression claim.

## Design

Keeping the native handler map and adding a catch-all only for absent handlers is a clear, conservative extension. Reusing the existing dispatcher avoids duplicating Spark's function implementations while preserving explicit compatibility restrictions. Function-and-class fallback diagnostics are useful because unrelated Iceberg functions share the name `invoke`.

Serialization success remains an admission check, not a guarantee that task-time Janino compilation or deserialization will succeed. That boundary should remain explicit. I did not establish a reachable Spark-working visibility regression. The targeted tests cover registration, serialization refusal, execution, and disabling dispatch. The independent component probe also checked a serialized and deserialized public receiver, while the full task bridge remains outside that probe.

## Abstraction & complexity

The new `CometInvoke` object uses the existing dispatch abstraction and introduces no separate evaluator, protocol, or lifecycle. The small diagnostic helper keeps the static handler's decision flow readable. The explicit unknown-handler branch earns its extra code because a broader fallback mixin would erase important handler-level declines. I found no unnecessary new abstraction or actionable complexity issue.

spark/src/main/scala/org/apache/comet/serde/statics.scala
```diff
@@ -78,16 +95,38 @@ object CometStaticInvoke extends CometExpressionSerde[StaticInvoke] {
...
+        // `StaticInvoke.doGenCode` emits a static method call, so the kernel matches Spark by
+        // construction. Spark 4.x keeps lowering more `RuntimeReplaceable` functions this way
+        // (`encode`, `is_valid_utf8`, the `TIME` family, ...) and `lpad` / `rpad` on binary has
+        // lowered to `StaticInvoke(ByteArray, ...)` since Spark 3.4.
+        //
+        // The encoder and deserializer trees that make up most `StaticInvoke` usage in typed
+        // Dataset operations are unaffected: their arguments are `ObjectType`, which
+        // `CometBatchKernelCodegen.isSupportedDataType` rejects, so the dispatcher declines them
+        // and they fall back exactly as before.
+        CometStaticInvokeCodegenDispatch.convert(expr, inputs, binding).orElse {
```
sunchao, line 109: ### Performance

[P2] Could you add matched microbenchmark results for the newly enabled static-call and receiver-call paths? This catch-all changes existing binary padding, encoding and `to_time` queries from Spark operator fallback to the JVM dispatcher. The function itself still runs Spark code, while dispatch adds expression transport, first-use compilation, per-batch bridge work and Arrow output materialization. The benefit from keeping surrounding operators in Comet needs to be measured against those costs, especially for short binary/string values and small inputs.

Please compare dispatch enabled with the existing dispatcher-disabled fallback on identical nonconstant inputs, covering `lpad`/`rpad`, `encode` or UTF-8 `to_binary`, and one evaluator-backed `Invoke` such as `to_time`. Include small and multi-batch inputs plus a mixed native projection, verify equal results, and state the Spark/JDK versions, batch size and warmup. Separate first-use compilation from steady-state timings. No slowdown has been measured here. These results are needed to validate the new default execution-path tradeoff.

andygrove: Added `CometCodegenDispatchBenchmark` in 9f8fade and ran it.

Three arms per case: codegen dispatch, `spark.comet.exec.scalaUDF.codegen.enabled=false` (which is exactly the operator fallback that shipped before this PR), and Comet off as a reference. Same corpus for all three, at 1024 rows (sub-batch) and 1,048,576 rows (~128 batches). All three arms are run and their rows compared before anything is timed, and every table repeats its baseline at the end so the noise floor sits next to the arms rather than in a footnote.

Spark 4.1.3, OpenJDK 17.0.10, Scala 2.13.17, `local[1]`, Apple M3 Max on macOS 26.6.2, `spark.comet.batchSize=8192`, whole-stage codegen on. Warmup is Spark's `Benchmark` default: 2s untimed per case, then at least two iterations and at least 2s of timed iterations. Best times in ms at 1,048,576 rows:

| case | dispatch off (pre-PR) | codegen dispatch | Comet off | baseline repeated |
| `lpad(binary)` | 51 | 61 | 71 | 53 |
| `rpad(binary)` | 64 | 61 | 79 | 67 |
| `encode(utf-8)` | 79 | 82 | 99 | 79 |
| `to_binary(utf-8)` | 76 | 79 | 97 | 76 |
| `to_time(fmt)` | 261 | 261 | 274 | 260 |
| mixed projection | 137 | 101 | 226 | 141 |
| group by dispatch | 73 | 65 | 61 | 73 |

The last column is the same work as the first, so ±4ms is noise here.

Where the projection is nothing but the one dispatched call, dispatch runs between parity and about 20% slower than letting the projection fall back. `lpad` is the worst of them at 51 → 61ms; `rpad` and `to_time` are parity; `encode` and `to_binary` are ~4% slower. That is the bridge cost you were asking about and it is real — for a bare `SELECT f(col) FROM t` there is nothing else in the projection for staying in Comet to save, so the transport and the Arrow output are pure addition.

It goes the other way as soon as the plan has anything else in it. The mixed projection is `length(c_str), c_long + 1, substring(c_str, 1, 4), lpad(c_bin, 24, c_pad)`, and dispatch is 1.4x there (137 → 101ms) because the three native expressions no longer follow the fourth out to Spark. The grouped case is `count(*)` over `lpad(c_pad, 8, c_pad)`, 1.1x (73 → 65ms), because the partial aggregate, the exchange and the final aggregate all stay native instead of following the projection. Comet-off is nominally best on that last row, but its average is 84ms against dispatch's 69ms with a 44ms stdev, so I would not read the 61 as a real win.

At 1024 rows nothing is resolvable. Every Comet arm lands in 4–11ms, fixed job cost dominates, and the baseline's own repeat moves by as much as the arms differ from each other. I would not read a direction into that table in either direction.

First use is separated out into its own table (1024 rows, taken before anything else has executed these queries, and after warming the dispatcher on an unrelated expression so that what is left is mostly the one kernel). 1st run / 2nd run / difference, in ms: `lpad` 85 / 29 / 56, `rpad` 44 / 27 / 17, `encode` 55 / 29 / 26, `to_binary` 33 / 23 / 10, `to_time` 69 / 25 / 44, mixed 103 / 39 / 64, group by 214 / 38 / 175. The dispatcher reported exactly one compile in each first run. These are upper bounds rather than compile times: the grouped case's 175ms is mostly the first shuffle in that JVM, and `to_binary` looks cheap only because `encode` ran before it and the two compile to the same source, which `CodeGenerator` caches JVM-wide. Net of that it is tens of ms per distinct kernel source per JVM, paid once and amortised by the per-task cache after.

Two things about how the benchmark is built, because a case that quietly measures the wrong thing is worse than no case. Each table asserts before timing that the dispatch arm is fully Comet native, that the dispatch-off arm is not, and that the dispatcher actually ran, and writes a warning into the results file if any of that fails. The first two shapes I tried for "the operator above stays native" — a filter, and an ungrouped `sum` — both tripped it, because I had written `length()` over binary and that has no native Comet path, so the operator was falling back for a reason that has nothing to do with this PR. Both would have printed a clean 1.0x.

Where that leaves me: the shapes that lose are the ones where falling back was cheapest to begin with, and they lose single digits to 20% on a query that is mostly scan; the shapes that win are the ones where falling back was expensive. I think that is the right trade, and I would rather not gate the catch-all on something like "only dispatch when the projection has other convertible expressions" until a real query regresses, since that gate would also have to be right about what is downstream of the projection and getting it wrong costs more than the 10ms.

sunchao (approved):
## Correctness

Re-reviewed `97108557177b0209b2a3e0717544b637f73543a6` against base `719cba11076e4237d6030925c49b1c1ffcac6f8e`, focusing on the two commits since the previous review. The only added file is `CometCodegenDispatchBenchmark`; all 2,084 previously tracked entries, including production code and dependencies, are unchanged. The maintained Spark 3.5/4.0 comparison therefore still applies: Spark-generated static/receiver calls retain their argument and null handling, while disabled dispatch, serialization refusal and known-handler declines preserve fallback. This update adds no new type, ANSI, overflow or error-handling behavior.

The benchmark compares complete result multisets against Comet-disabled Spark, normalizing binary values, for the same materialized nonconstant corpus at both sizes before steady-state timing. It uses the existing consuming `noop` action, not a count that could discard the projection. These are valid-input comparisons; they do not extend the earlier null/error boundary coverage. I reverified source equivalence for the prior component probes: 60 padding/encoding row-or-error comparisons, a serialized public receiver, and 12 `to_time` rows. Those were public Spark 4.0.4/4.1.3 component executions, not a full native Comet SQL run, and were not rerun for this benchmark-only update. Maintained 3.4/4.1 source gaps remain.

Current-head checks at `2026-09-04T22:54:48.090Z` were 21 successful, 30 running and 6 skipped, with no failed check observed. The [benchmark compile and lint check](https://github.com/apache/datafusion-comet/actions/runs/33924795983/job/101191249277) was successful; the wider CI workflow remained in progress. I inspected the added benchmark and checked the incremental diff; I did not independently compile or execute it, reproduce its timings, or establish a complete CI pass.

## Performance

The benchmark and [reported results](https://github.com/apache/datafusion-comet/pull/5692#discussion_r3938192745) address the [earlier measurement request](https://github.com/apache/datafusion-comet/pull/5692#discussion_r3937536823): dispatch on/off and Comet off, small/multi-batch inputs, static and receiver calls, mixed projection, explicit versions/batch size/warmup, and a repeated fallback baseline. The numbers are author-reported on Spark 4.1.3/JDK 17, not independently reproduced measurements.

The standalone `lpad` regression is real within that report: 51–53 ms for fallback versus 61 ms with dispatch, around 15–20% slower. It should not be dismissed as noise. The mixed projection improves from 137–141 ms to 101 ms, while grouping improves from 73 ms to 65 ms. These demonstrate a query-shape tradeoff rather than a general speedup. Small-input differences are unresolved, and one repeated baseline does not establish a universal ±4 ms noise bound.

The separate first-use table is useful as end-to-end startup evidence, but it does not isolate Janino time: query/shuffle setup and shared source-cache hits remain, and `compileCount` counts per-task dispatcher misses. The supplied reply includes selected results rather than the full output. Plan checks emit warnings and continue, so a warned table must not be treated as a validated dispatch-versus-fallback measurement. With these limits, I consider the existing evidence request addressed; I found no additional P1/P2 to duplicate or add.

## Design

Toggling the existing dispatcher flag provides the relevant prior-behavior comparison for these cases, while Comet-disabled Spark supplies a separate reference. The mixed and grouped cases explain why a gate based only on whether a projection contains another native expression would be incomplete: retaining native downstream operators also matters. This corpus does not establish a reliable rule for adding such a gate. Native implementations could remove bridge overhead, but would require their own Spark-compatibility work.

## Abstraction & complexity

The case descriptor and shared configuration helpers keep all arms on one corpus and reuse the existing benchmark framework. They add no production abstraction or hot-path work. The warning-based plan diagnostics and first-use counters are useful inspection aids with the limits above; they are not an automated performance-regression gate. No new actionable complexity issue was identified.

PR #4810, opened 2026-07-03
feat: Runtime dynamic filter pushdown for native hash joins

schenksj:
> _Disclosure: this issue was drafted with the help of an AI assistant ._

## Which issue does this PR close?

Closes #4807.

Follow-up issues for IO-level pushdown into kernel scans: #4808 (Iceberg), #4809 (Delta).

## Rationale for this change

DataFusion 54 ships runtime dynamic filtering for hash joins (`HashJoinExec` + `DynamicFilterPhysicalExpr` + `SharedBuildAccumulator`), but the consumer-side wiring lives in DataFusion's physical optimizer, which Comet does not run — so Comet never benefits. Build-side runtime filtering is one of the largest TPC-DS levers in comparable engines (Velox `HashProbe::pushdownDynamicFilters`; Databricks DFP). This PR wires it up on Comet's side, scan-agnostically, so all probe-side scans (Parquet `DataSourceExec`, JVM-fed `ScanExec`, Iceberg, and future Delta kernel scans) inherit row-level filtering.

## What changes are included in this PR?

- New `CometDynamicFilterExec` operator (`native/core/src/execution/operators/dynamic_filter.rs`): evaluates the join's shared `DynamicFilterPhysicalExpr` against probe-side batches before the hash probe, with
  - a pass-through fast path while the filter is the constant-`true` placeholder (correctness never depends on population),
  - a selectivity guard that permanently disables evaluation on a partition stream when it keeps > 95% of rows after 64K rows observed,
  - metrics: `dynamic_filter_rows_pruned` + baseline elapsed-compute/output-rows.
- `attach_join_dynamic_filter` helper: rewires an eligible `HashJoinExec` (including the `ProjectionExec`-over-join shape produced by `swap_inputs`) so the join and the new probe-side wrapper share one filter, via the public `HashJoinExecBuilder` + `with_dynamic_filter_expr` APIs.
- Eligibility gate mirrors DataFusion's own: probe side must be preserved under the ON clause (`JoinType::on_lr_is_preserved().1` — Inner, LeftOuter, LeftSemi/RightSemi, LeftAnti, LeftMark); null-aware anti joins are excluded.
- Planner integration in both HashJoin branches (build-left and swap), proto field `HashJoin.dynamic_filter_enabled`, and a new config `spark.comet.exec.join.dynamicFilter.enabled` (default `false`, experimental).

[...]

Issue #4807, opened 2026-07-03
feat: Runtime dynamic filter pushdown for native hash joins (build side -> probe side)


schenksj:
> _Disclosure: this issue was drafted with the help of an AI assistant ._

## What

Wire up runtime dynamic filter pushdown (a.k.a. runtime filtering / sideways information passing) for Comet's native hash joins: once the build side completes, apply a build-key-derived predicate to probe-side batches *before* the hash probe.

DataFusion 54 already ships the machinery — `HashJoinExec` supports an attached `DynamicFilterPhysicalExpr` (placeholder `lit(true)`, updated after the build with `min/max` bounds plus `InList` or hash-table-lookup membership via `SharedBuildAccumulator`), and `with_dynamic_filter_expr` is public. Comet never benefits because the consumer-side wiring lives in DataFusion's physical optimizer (`FilterPushdown` rule), and Comet constructs its physical plan directly without running optimizer rules.

This class of optimization is one of the largest TPC-DS levers in comparable engines (Velox `HashProbe::pushdownDynamicFilters`; Databricks DFP reports up to 8x on selective star joins).

## Proposed design (generic, scan-agnostic phase)

In the planner's `OpStruct::HashJoin` arm, when eligible:

1. Create `DynamicFilterPhysicalExpr::new(probe_keys, lit(true))` from the final (post-`swap_inputs`) join's `on()` probe-side keys — always valid against the probe child schema, no column remapping.
2. Attach it to the join via `with_dynamic_filter_expr` (rebuild through `HashJoinExecBuilder`).
3. Wrap the join's probe child in a new `CometDynamicFilterExec` operator evaluating the *same* `Arc`:
   - pass-through fast path while the filter is the scalar-`true` placeholder (correctness never depends on population),
   - `filter_record_batch` once populated,
   - per-stream selectivity guard (auto-disable when observed selectivity > ~0.95 after 64k rows),
   - metrics: `dynamic_filter_rows_pruned`, engaged flag.

Eligibility gate (mirrors DataFusion's own): `join_type.on_lr_is_preserved().1` (Inner, LeftOuter, LeftSemi, RightSemi, LeftAnti, LeftMark), not null-aware anti join, and a new opt-in config `spark.comet.exec.join.dynamicFilter.enabled` (default false initially) carried in the `HashJoin` proto message.

[...]

PR #5699, opened 2026-09-04
feat: native dynamic filter pushdown for hash join into Parquet scans

### LINKED ISSUE #5695: Push native join runtime filters into scans to reduce I/O and decoding

Selective joins can discard most fact-table rows, but filtering is much less valuable when those rows have already been read, decoded, and converted into Arrow. We should let eligible native joins pass runtime information back to native readers so that work can be avoided at the scan.

For example:

```sql
SELECT SUM(s.ss_net_paid)
FROM store_sales s
JOIN item i ON s.ss_item_sk = i.i_item_sk
WHERE i.i_category = 'Books';
```

Suppose the filtered build side contains item IDs `{10, 20, 30}`. After that build is complete, its key domain can provide a safe predicate for the fact scan. A reader may then skip row groups whose statistics exclude those IDs, or reject rows during decoding before materializing other columns. This does not require `ss_item_sk` to be a table partition column. The actual join remains responsible for producing the result.

In the successful-query subset of our latest 10 TB comparison, Spark-generated Bloom filters already executed natively in 27 of 79 queries. However, the observed path was **Spark scan → Spark-to-Arrow conversion → native Bloom filter → downstream operators**. Those filters cannot undo the preceding I/O or conversion. Native scans were deliberately disabled for that shuffle comparison, so it does not measure the benefit proposed here.

DataFusion already has dynamic-filter producers and consumers, but Comet directly constructs and executes physical plans rather than running the physical optimizer pass that normally connects them. This is an integration gap, not simply a disabled DataFusion option.

Connect build-side runtime filters to eligible native probe-side scans, starting with a narrow hash-join path. Reuse DataFusion's existing filter machinery where appropriate, and preserve Comet's Spark-facing execution and metrics contracts.

There are two distinct milestones:

1. **Producer and row-filter integration:** populate a shared filter from the completed join build and consume it on probe batches. This is useful infrastructure, but a filter immediately before the hash probe can only save later work.
2. **Reader integration — the goal of this issue:** propagate or safely translate that filter into the native reader before relevant reads/decoding occur. Demonstrate real row-group skipping and/or reduced decoding, rather than treating a post-scan row filter as scan pushdown.

Start with native Parquet and the shared Parquet machinery used by our internal Spark/Delta-authoritative reader. Keep Spark and Delta authoritative for snapshots, selected files, splits, deletion vectors, and logical-to-physical column mapping. Executor-local pruning should only eliminate work within that authorized input; it must not introduce a second Delta snapshot planner or change Spark task partitioning.

Do not assume that running every DataFusion optimizer rule is safe: rewrites can change operator structure, break native-to-Spark metric mapping, or violate Spark partition/aggregation contracts. Evaluate a targeted pass or explicit producer/consumer wiring, and document where pushdown must stop.

### Additional context

_No response_

pingzh:

Closes #5695.

## Rationale for this change

Selective joins can reject most probe rows after those rows have already been read and decoded. Connect completed native hash-join build domains to eligible Parquet readers so they can skip row groups, while retaining the original join and residual predicates.

- Add the default-off `spark.comet.exec.join.dynamicFilter.enabled` option for inner joins with one direct signed integer key and one native partition per input.
- Create fresh producer/consumer state for each execution, attach the live predicate to eligible native Parquet readers, and filter decoded probe batches. Preserve build-side swaps, Spark operator structure, and join/scan/filter metric ownership.
- Support scan projection remapping and direct-column null-check conjunctions. Stop reader attachment at standalone projections, other predicates, limits, and unsupported readers; do not transport filters across Spark exchanges or JVM/Arrow boundaries.
- Add dedicated runtime-filter metrics and user documentation. The original join remains responsible for exact matching, and existing defaults and dependency versions remain unchanged.

This includes native Parquet reader integration beyond the probe-batch filtering proposed in #4807 and #4810. Iceberg and Delta kernel reader integration remain separate work.

- Native runtime-filter and planner tests: 14 passed, covering completed and delayed builds, swaps, null/empty/duplicate keys, unsupported shapes, execution isolation and cleanup, reader/filter limits, schema remapping, metric attribution, and clustered versus unclustered Parquet pruning with byte-read assertions.
- Native build and clean Spark 4.0 JVM package build passed.
- Full Spark 4.0 `CometJoinSuite`: 44 passed, including 13 new runtime-filter regressions with native-plan assertions, both Spark build sides, broadcast/shuffle strategies, AQE, reader pruning, and seeded-randomness boundaries.
- Rustfmt, Scalafix, Spotless, Scalastyle, and workspace Clippy with warnings denied passed.

### Performance Evidence

The original fork was benchmarked on TPC-DS q52 at SF1000 using five counterbalanced enabled/disabled pairs. Only `spark.comet.exec.join.dynamicFilter.enabled` differed between arms.

| Spark SQL execution time | Disabled | Enabled | Reduction |
|---|---:|---:|---:|
| Mean | 38.4 s | 27.3 s | 28.9% |
| Median | 38.8 s | 27.4 s | 29.4% |

Every pair improved, with SQL execution-time reductions ranging from 18.8% to 35.3%. All ten runs succeeded and produced matching canonical result hashes.

A separate reader-level preflight showed:

| Metric | Disabled | Enabled | Reduction |
|---|---:|---:|---:|
| store_sales scan-output rows | 2,879,966,589 | 95,743,561 | 96.68% |
| Bytes scanned | 15.2 GiB | 524.7 MiB | 96.63% |

The enabled preflight pruned 858 of 1,147 row groups.

These are original-fork results using a native Delta scan path, not measurements of this Apache PR. This was a feature-isolation experiment: DPP, DFP, and runtime Bloom filtering were disabled, and InferFiltersFromConstraints was excluded. Cluster placement varied, and compile-time/runtime Spark builds differed. The results demonstrate potential on this selective query, not a production-default or full-suite speedup.


sunchao (approved):
## Correctness

Reviewed `44e56bb871b099e0dec5006e8497ef11880e2ce5` against `719cba11076e4237d6030925c49b1c1ffcac6f8e`. Previously, a native hash join could consume probe rows that its completed build keys would later reject. This adds a default-off connection from DataFusion's completed build domain to an eligible native Parquet reader and a residual probe filter, while retaining the original hash join for exact matching.

The eligible path is limited to inner joins with ordinary equality, one matching direct signed-integer key and one native partition on each side. That fits the inspected maintained Spark 3.5/4.0 semantics: null keys do not match ordinary equality, duplicate build keys retain all matches, and the original join condition still determines output. The prefilter does not introduce new key coercion or arithmetic/ANSI behavior. Outer, semi, anti, null-aware and unsupported key shapes retain their existing execution paths.

The join completes its build domain before probing. Reader traversal permits direct null checks and their conjunctions, handles scan projection remapping, and stops at other filters, standalone projections, limits and unsupported readers. Those boundaries protect stateful expressions such as seeded randomness. A fresh dynamic predicate and join state are created for each execution, and the stream wrapper releases that state at exhaustion or error. I found no verified false-negative pruning or stale-domain issue in the reviewed paths.

The latest commit changes only the fallback test's hint selection. RIGHT JOIN now builds alias `p`, while the other seven cases retain `b`. All eight query clauses, native-plan assertions and zero-filter metric checks remain, and the separate null-aware `NOT IN` block is unchanged. The RIGHT JOIN query no longer repeats the fixed right-side hint that caused the earlier Spark 3.4 failure. I checked the remaining aliases and join shapes against the maintained Spark 3.5/4.0 hint/build-side rules and found no additional issue. This source assessment does not establish that the complete loop now passes on Spark 3.4.

Production code, native tests, dependency lockfiles and every file outside this Scala test are byte-identical to the previously reviewed head `8cc4602bb64db7a85bdb404b30509a7ac1143ecd`. Its CI merge had a tree identical to that head. The prior Rust job passed 1,143 tests with four skipped, including all 14 new native/planner tests. The prior Spark 4.0 job passed 800 tests, including all 13 new runtime-filter cases. These remain evidence for the unchanged native implementation and the seven unchanged loop queries. They are not a passing result for the changed RIGHT JOIN query. The previous Spark 3.4 loop aborted at RIGHT JOIN, so its later FULL, SEMI, ANTI, null-safe, computed-key, string-key and `NOT IN` cases remain unvalidated on that version by those old results. [Prior Rust results](https://github.com/apache/datafusion-comet/actions/runs/33916991257/job/101167341424), [prior Spark 4.0 results](https://github.com/apache/datafusion-comet/actions/runs/33916991257/job/101174540081).

The current CI merge is `5cea8f90509f4a3340f56e70fdab3913af6f2c97`, with the exact reviewed head/base parents and a tree identical to the current head. At 2026-09-04T22:22:50.469305+00:00, current-head checks were 15 successful, 6 running and 7 skipped, with no failures. The Spark 3.4/JDK 11 Java-lint job passed on that merge, which establishes lint success, not execution of the fallback test. Current-head Spark 3.4 runtime validation is still pending. [Current CI run](https://github.com/apache/datafusion-comet/actions/runs/33923968311), [current Spark 3.4 lint job](https://github.com/apache/datafusion-comet/actions/runs/33923968311/job/101189022742). The current-head Rust job was still running when its metadata was collected.

Maintained Spark 3.4/4.1 source refs are unavailable, so no canonical-source coverage is claimed for those versions. The earlier local component harness never built or ran after offline dependency resolution stopped at the configured registry index, and it was not retried for this revision. No local runtime or timing result is implied. No verified P1/P2 finding remains in this review.

## Performance

The disabled path does not install a row filter or reader wrapper; its added work is configuration/planning and the feature flag. When enabled, the residual filter adds predicate evaluation and potentially batch filtering/copying even when little is pruned. Reusing DataFusion's existing hash map, and locally disabling per-row IN-list materialization, avoids building another literal collection for duplicate-heavy keys. The filter and reader metrics distinguish attachment, residual row reduction, and scan savings, which is useful because attachment alone can save no I/O.

Please add a matched elapsed-time benchmark for **disabled**, **residual-only**, and **reader-plus-residual** execution. Include selective and non-selective joins, narrow and wide probe rows, clustered and unclustered Parquet layouts, and a late/expensive build case; use an eligible end-to-end query where possible. Keep data, resources, cache conditions, native plan and settings matched, report at least three repetitions and medians, and preserve result fingerprints. Record exact revisions, scan bytes/row-group pruning, residual rows evaluated/pruned/bypassed, and elapsed time separately. The correctness tests demonstrate pruning mechanics, but neither fewer rows nor fewer bytes proves an end-to-end speedup. No measured regression or speedup is established by this review.

## Design

The division of responsibility is appropriate: DataFusion supplies the completed domain, shared dynamic expression, projection remapping and Parquet pruning; Comet supplies the explicit wiring needed by its planner, eligibility checks, execution lifetime and Spark metrics. Leaving the original hash join authoritative keeps this an optional prefilter. Treating failed or unsupported reader attachment as a fallback, while retaining the residual consumer, keeps reader capability separate from join correctness.

The permanent wrapper retains an unexecuted join template rather than completed filter state. Rebuilding the predicate and reader view for each execution, then dropping the execution stream at EOF/error, addresses reuse and retained-memory risks. Applying the attachment after build-side swapping and preserving existing metric handles are also necessary details. The documented boundaries around Spark exchanges, JVM readers and other table formats keep the feature's initial scope understandable.

## Abstraction & complexity

The two adapters have distinct jobs: the join wrapper owns execution-local filter state, while the filter adapter preserves Spark's metric identity as its execution input is rebuilt. That extra code has a concrete purpose; replacing it with a shared completed predicate would make plan reuse and cancellation harder to reason about. Reusing upstream domain/pruning machinery avoids a parallel range or membership implementation in Comet.

The narrow null-check traversal and explicit eligibility checks are preferable to broad recursive pushdown with scattered exceptions. The added lifecycle, reader-remapping, fallback, nondeterministic-expression and duplicate-heavy tests support these boundaries. The test now records its build alias beside each join clause. This makes the required side explicit without introducing a version-specific production path. Current-head Spark 3.4 execution is still needed to confirm the entire fallback loop.

native/core/src/execution/operators/dynamic_filter.rs
```diff
@@ -0,0 +1,1744 @@
...
+        let bypassed =
+            MetricBuilder::new(&self.metrics).counter("dynamic_filter_rows_bypassed", partition);
+        // Only dedicated metrics: merging this helper into the Spark join must not
+        // add its input/output counts or elapsed time to the join's existing metrics.
+        let eval_time =
+            MetricBuilder::new(&self.metrics).subset_time("dynamic_filter_eval_time", partition);
+        let stream = input.map(move |batch| {
+            let batch = batch?;
+            let _timer = eval_time.timer();
+            match predicate.evaluate(&batch)? {
```
sunchao: **[P2] Evaluate the runtime predicate using only the join-key column**

The predicate receives the entire probe batch here. Its completed-build expression is `bounds AND hash_lookup`. When the bounds produce a non-null mask retaining some rows, but at most 20% of the batch, DataFusion 54.1 first filters the entire batch before evaluating membership. That copies every payload column into a temporary batch. The `filter_record_batch` at line 159 then filters the original batch again, copying the surviving payload a second time.

I reproduced this using the actual completed-build predicate with 8,192 rows, 1,024 selected rows, and 32 BIGINT payload columns: **256 KiB of discarded payload copying per batch**. The probe used build keys `0..1024` and a permutation of probe keys `0..8192` (`(i * 641) % 8192`). Observing the input to the real membership expression confirmed distinct original, temporary, and final payload buffers. Evaluating the same predicate on a key-only projection produced the identical Boolean mask without the temporary payload copy.

Could we use the existing dynamic-expression remapping mechanism to evaluate a zero-copy, key-only batch, then apply its mask to the original batch once? Eligibility already guarantees one direct join-key column. This avoids work that scales with the probe payload width while retaining the current producer and reader behavior. Whole-query performance impact remains unmeasured.

sunchao (commented):
Reviewed `44e56bb871b099e0dec5006e8497ef11880e2ce5` against `719cba11076e4237d6030925c49b1c1ffcac6f8e`. One P2 performance finding is documented inline. No confirmed correctness defect.

### Review conclusions

- **Correctness:** Build completion, nulls, duplicates, build-side swaps, execution isolation, and cleanup look sound. Reader boundaries preserve limits and stateful-expression evaluation.
- **Performance:** Selective runtime-filter evaluation can copy payload columns into a temporary batch and then copy the surviving payload again. The inline finding includes a buffer-copy reproduction and a key-only evaluation alternative. Clustered Parquet pruning tests demonstrate avoided reads. Benefits depend on layout and reader eligibility, and shuffled inputs cannot propagate filters back across Spark exchanges. Whole-query performance impact remains unmeasured.
- **Design:** Execution-scoped producer/consumer state is justified and avoids retaining completed build state.
- **Abstraction & complexity:** The join wrapper and filter metric adapter each address a concrete lifetime or metric-ownership requirement. I found no separate complexity issue.

### Validation

- All **14 focused native tests passed** on this head using `cargo test -p datafusion-comet --no-default-features dynamic_filter`.
- **24 additional combinations passed**, covering integer extremes, type promotion, partition keys, defaults, and missing keys, with runtime filtering and Parquet row filtering independently enabled/disabled.
- The buffer-copy reproduction passed: 8,192 input rows, 1,024 selected rows, and 32 BIGINT payload columns caused **256 KiB of discarded payload copying per batch**. Original, temporary, and final payload buffers were distinct. Evaluating the same completed-build predicate on a key-only projection produced an identical mask with no temporary payload copies. This establishes unnecessary copying; whole-query timings were not measured.
- Full Spark/Comet JVM suites were not rerun locally.
- Current-head CI at publication preflight: **21 successful, 29 running, 7 skipped, no failures**. [PR checks](https://github.com/apache/datafusion-comet/pull/5699/checks).

sunchao (approved):
## Correctness

Re-reviewed `509bd139568d9f877245f9bd94ec5a4ab7fdf3da` against base `719cba11076e4237d6030925c49b1c1ffcac6f8e`, covering the full change since `44e56bb8`. Only `dynamic_filter.rs` changed. The implementation addresses the [existing P2](https://github.com/apache/datafusion-comet/pull/5699#discussion_r3938370027), and I found no verified new or remaining P1/P2.

The consumer reads the actual probe-key index and derives a live predicate view with that key mapped to index zero. DataFusion remaps every matching leaf in the completed expression, including both bounds and membership. The producer and reader retain their original coordinates, and the membership rewrite preserves the hash seed and shared map. Arrow's one-column projection retains the original key values, validity, type, row count and order.

Scalar true still returns the original batch, false/null returns its empty slice, and an array result filters the original full batch once. The original join remains responsible for duplicate matches and its join condition, consistent with the maintained Spark 3.5/4.0 implementations. No new coercion or arithmetic behavior is introduced. Evaluated/pruned/bypassed metrics still use original rows, and the existing reset and stream-release paths are unchanged.

The new regression uses key index 17, 32 BIGINT payload columns and the real completed bounds/membership predicate. It checks one-column evaluation, a shared key-value buffer and the complete filtered output. The updated placeholder test changes the producer after stream creation using original key index 1, then checks filtering, metrics, reset and error propagation.

Current CI uses merge `4703d2088231d51a28f8925886f0c22d7f292704`, whose exact head/base parents and full-tree equality to the reviewed head were verified. The [current Rust job](https://github.com/apache/datafusion-comet/actions/runs/33929840124/job/101206626026) passed 1,144 tests with four skipped, including both `completed_filter_evaluates_only_the_shared_probe_key` and the modified `placeholder_updates_and_errors_are_not_hidden`. I independently read the complete log and verified its checkout. At `2026-09-05T00:12:43.042Z`, current-head checks were 20 successful, 30 running and 7 skipped, with no failures. This is not a complete CI pass or current-head Spark/JNI validation. No local build, harness execution or benchmark was run. Maintained Spark 3.4/4.1 source refs remain unavailable.

## Performance

The conditional AND shortcut can now copy only the key into its temporary batch, eliminating the payload-width-dependent copying described in the P2. This is a source-verified removal of that work, not a measured whole-query speedup. The derived view is created once per execution, but projection metadata is allocated per batch and the live expression is remapped during evaluation. Scalar placeholders also incur the projection step. The disabled path and reader pushdown behavior are unchanged.

Please retain the matched timing comparison, including narrow/wide payloads, selective bounds versus sparse membership within broad bounds, nullable and all/none cases, and repeated end-to-end timings with identical plans/settings and result fingerprints. Those controls distinguish avoided payload copies from the added metadata/remapping cost.

## Design

The fix stays in the residual consumer and uses the existing dynamic-expression view mechanism. That preserves live producer updates without a copied domain or a stale completed-expression snapshot. Keeping output filtering on the original batch also avoids reconstructing payload columns or changing Spark's output schema. The new assertions enforce the existing one-direct-key contract rather than extending eligibility.

## Abstraction & complexity

The production change is small and reuses Arrow projection and DataFusion remapping. The larger addition is test-only instrumentation that observes the actual predicate input and retained key buffer, including a nonzero original key index. It provides targeted coverage of the reported copy path without adding a production abstraction. No duplicate inline finding is needed.

peterxcli:
@pingzh curious if there are any performance benchmarks for this, mainly want to see which TPC-H queries or custom workloads the dynamic filter improves the most.

peterxcli (commented):
I would suggest renaming the PR title to:

feat: native **dynamic** filter pushdown for **hash join** into Parquet scans

Two reasons:
1. "Runtime filter" implies generating a bloom filter in the context of Spark.
2. DataFusion's dynamic filter also supports agg and topk exec, so I think we can emphasize it in the title and then open a follow-up issue to track dynamic filter support for other operators. (I think I can help land the follow-up issue.)

native/core/src/execution/operators/dynamic_filter.rs
```diff
@@ -0,0 +1,1929 @@
...
+    }
+
+    fn partition_statistics(&self, partition: Option<usize>) -> Result<Arc<Statistics>> {
+        self.template.partition_statistics(partition)
+    }
+}
+
+/// Attach after choosing the final build side, including the projection emitted
+/// by `swap_inputs`. The wrapper owns both join and dedicated filter metrics.
+pub(crate) fn attach_join_dynamic_filter(
```
peterxcli: this whole function looks like a optimizer rule, I think we should move this to planner module?

native/core/src/execution/operators/dynamic_filter.rs
```diff
@@ -0,0 +1,1929 @@
...
+            build_type,
+            DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64
+        )
+    {
+        return Ok(Some("requires matching signed integer keys"));
+    }
+    Ok(None)
+}
+
+#[cfg(test)]
```
peterxcli, line 571: Tests occupy roughly 3/4 of the file, consider to move test to another file?

pingzh: @peterxcli Agreed. I’ll move the tests into dynamic_filter/tests.rs

pingzh:
> @pingzh curious if there are any performance benchmarks for this, mainly want to see which TPC-H queries or custom workloads the dynamic filter improves the most.

hi @peterxcli thanks for checking. I just updated the PR description to include some results.

native/core/src/execution/operators/dynamic_filter.rs
```diff
@@ -0,0 +1,525 @@
...
+    }
+}
+
+impl DisplayAs for DynamicFilterExec {
+    fn fmt_as(&self, _t: DisplayFormatType, f: &mut Formatter) -> std::fmt::Result {
+        write!(f, "CometDynamicFilterExec")
+    }
+}
+
+impl ExecutionPlan for DynamicFilterExec {
```
sunchao, line 82: ### Correctness

**[P1] Adapt the wrappers and tests to the current DataFusion integration**

Could we update these implementations for DataFusion 55 before merging? The [current Rust job](https://github.com/apache/datafusion-comet/actions/runs/33995365059/job/101385223231) actually checks out `512ddb6ad3fedaba3d1874fb4b86cc2e619a3cdf`, with parents `c348f775` and `fb4f75c6`. It reports missing `apply_expressions` implementations for `DynamicFilterExec`, `DynamicFilterJoinExec` and `CometFilterExec`. Clippy also rejects obsolete API calls, including the moved planner call, and the reader test still uses unavailable `inner()` and `remapped_children()` accessors. The job exits before runtime tests.

The dynamic-filter files and `filter.rs` are byte-identical between this head and the executed merge, and the moved planner helper and its call sites also match. These implementations still need the API migration for this integration to build. The assigned `719cba11` pair retains DataFusion 54.1.0, so older passing results do not qualify the newer-base DataFusion 55 build.

sunchao (commented):
## Correctness

Re-reviewed `fb4f75c6f3243911cb3d9e8de8d5cd7a53d6f2ee` against `719cba11076e4237d6030925c49b1c1ffcac6f8e`, covering the full planner-refactor increment and the earlier test extraction since the published `509bd139` approval. Attachment still follows the final build-side decision. `try_new` retains the same eligibility checks, and `with_new_children` rebuilds the join before rechecking them. Unsupported replacements return the rebuilt ordinary join. Producer, reader and consumer state remain per execution.

The original 14 child tests retain their assertions after routing seven calls through the planner. Two added tests cover supported versus unsupported replacement key types, reset behavior and replacement of either side with multiple native partitions. The module remains private and `#[cfg(test)]`-gated. These are source checks, not passing runtime results. Eligibility still limits filtering to ordinary inner equality on matching signed-integer keys, preserving the maintained Spark 3.5/4.0 analysis of nulls, duplicates and residual conditions. No new coercion, overflow, ANSI or fallback semantics are introduced.

### Current CI integration

The [current Rust job](https://github.com/apache/datafusion-comet/actions/runs/33995365059/job/101385223231) confirms that the DataFusion 55 integration blocker remains. Its actual checkout is `512ddb6a`, with parents `c348f775` and `fb4f75c6`. That differs from the assigned base and tree. The dynamic-filter files, registration and `filter.rs` match the head exactly, as do the moved planner helper and its call sites. Clippy still reports missing `apply_expressions` implementations, obsolete APIs and unavailable dynamic-expression accessors. The runtime-test step is skipped. The inline P1 uses this new evidence.

At `2026-09-05T22:35:11Z`, checks were 14 successful, five running, two failed and seven skipped. The [default-build gate](https://github.com/apache/datafusion-comet/actions/runs/33995364804/job/101384867146) also failed with exit 101 on the same merge. Its underlying cause is not established here. The assigned head retains DataFusion 54.1.0, while the executed integration uses 55.0.0. Historical passing CI does not qualify that integration or the new tests. No local build, Spark/JNI run or benchmark was performed. Maintained Spark 3.4/4.1 source refs remain unavailable.

## Performance

Runtime filtering and key-only predicate evaluation are unchanged. The planner now obtains a session-config snapshot before the disabled guard, adding planning work even when the feature is off. This shares the options through an `Arc` and copies the extension map, rather than deep-copying all options. No material timing regression is established.

The author's original-fork native-Delta q52 report remains separate from this Apache implementation. Its repeated timings and reported result hashes do not complete the existing current-PR disabled/residual-only/reader-plus-residual comparison. Placement, build-version and separate-preflight limitations remain, without another inline benchmark request or a suite-wide speedup claim.

## Design

The requested planner move is applied. Plan-shape traversal now belongs to `PhysicalPlanner`, while the wrapper owns eligibility and execution-local state. The crate-private `try_new` entry point lets both planning and child replacement enforce that contract without a production operator-to-planner dependency. The title and test extraction are already applied as well.

## Abstraction & complexity

The new constructor returns an optional wrapper instead of exposing its fields or duplicating eligibility rules. The planner preserves the original plan when attachment is unsupported, and the two added tests target the child-replacement boundary. This is a reasonable split of responsibilities. The remaining integration failure requires API adaptation, not another module reorganization.

sunchao (approved):
## Correctness

Re-reviewed `245552fded92412e2bc06980fcc3e378668299fd` against `f97fb4519cad2a4f6fe435d66aa17eacbd919a9b`. The five earlier commits are patch-identical after rebasing. The new four-file migration addresses the [published DataFusion 55 P1](https://github.com/apache/datafusion-comet/pull/5699#discussion_r3942277415). I found no verified new or remaining P1/P2.

The consumer exposes its live predicate through `apply_expressions`, while the join and filter adapters delegate expression visitation to their underlying operators. Delegating the complete distribution requirement preserves DataFusion's co-partitioning constraint. Both adapters forward child-statistics requests and statistics calculation, retaining overall and per-partition estimates. The tests now inspect dynamic expressions through their public IDs and children.

Child replacement still resets execution state and rechecks eligibility. DataFusion 55's join builder invalidates cached properties when the replacement children have different property identities, so the wrapper does not retain stale schema or partitioning in those cases. The filter adapter forwards replacement options. Build completion, key-only residual evaluation, reader boundaries, null/duplicate handling and stream cleanup are unchanged by the migration.

The [current Rust job](https://github.com/apache/datafusion-comet/actions/runs/33997285374/job/101390179322) passed Clippy with warnings denied and **1,188 tests, with four skipped**. This includes all 18 dynamic-filter tests, both new filter-adapter tests and the planner metrics test. Its actual checkout is `d13ab8ba`, with the exact assigned base and head as parents. The entire native tree and all PR-authored files match the reviewed head. The only head-to-merge difference is the base branch's columnar-to-row benchmark test. These results directly qualify the migrated native code and the previously unexecuted child-replacement tests.

At the final complete snapshot on `2026-09-05T23:54:24.360Z`, 57 checks had succeeded, seven remained in progress and seven were skipped, with no failing checks reported. This is not a complete CI or Spark suite pass. No local build, runtime test or benchmark was run. The maintained Spark 3.5/4.0 analysis remains applicable to unchanged join eligibility and semantics. Maintained Spark 3.4/4.1 source refs remain unavailable.

## Performance

The migration leaves Comet's key-only residual evaluation and reader attachment unchanged. It adds API delegation and focused tests without introducing a new row-processing pass. No measured regression or speedup is established by this review.

The original-fork native-Delta q52 results remain separate from this Apache PR. Their reported timing and result-hash evidence does not complete the existing matched disabled/residual-only/reader-plus-residual comparison. The documented placement, build-version and separate-preflight limitations still apply, without a duplicate benchmark request.

## Design

The adapters now use DataFusion 55's expression, statistics and distribution interfaces while preserving execution-local producer/consumer ownership. Delegating the underlying operators' behavior avoids duplicating their statistics or distribution logic. Public expression inspection also removes the tests' dependence on unavailable private accessors.

## Abstraction & complexity

The migration stays within the existing wrappers and planner helper. Four new tests cover visitor recursion, predicate identity, partition statistics and co-partitioning requirements. The earlier planner move and test extraction remain intact. No additional abstraction or module reorganization is needed, and no duplicate inline finding is proposed.

peterxcli:
> hi @peterxcli thanks for checking. I just updated the PR description to include some results.

number looks great! thanks!

sunchao (commented):
## Correctness

Re-reviewed `0d206f31` against `7e198439`. All six commits are patch-identical to the previously approved series ending at `245552fd`. The native core, reader integration, planner, dependency lockfile, Scala join serialization and join tests are unchanged. The DataFusion 55 migration remains intact, including expression visitation, distribution requirements, statistics delegation and child replacement. I found no new or remaining P1/P2. This update preserves the existing [approval](https://github.com/apache/datafusion-comet/pull/5699#pullrequestreview-5123498403), which GitHub now associates with this head. All four existing review threads are resolved.

The new base adds an independent, default-off option for nested native-shuffle keys. It does not broaden runtime-filter eligibility: filtering still requires an ordinary inner join with one matching direct signed-integer key and one native partition per input. Float/NaN, decimal, string and nested keys remain excluded. Null handling, duplicate matches and the original residual join condition retain the maintained Spark 3.5/4.0 semantics checked previously and rechecked here. No new coercion, overflow or ANSI behavior is introduced. Attachment still follows the build-side swap, and reader traversal stops at Spark/JVM/shuffle boundaries, limits and unsupported expressions. Producer publication, execution isolation and stream cleanup are unchanged. Maintained Spark 3.4/4.1 sources remain unavailable.

The [current Linux lint job](https://github.com/apache/datafusion-comet/actions/runs/34011011813/job/101426904889) passed on `9a527913`, whose parents are the exact assigned base/head and whose complete tree matches the reviewed head. The current Rust job was still running at `2026-09-06T04:21:08Z`. The previously verified [1,188 passing native tests, with four skipped](https://github.com/apache/datafusion-comet/actions/runs/33997285374/job/101390179322), remain evidence for the unchanged native implementation and its DataFusion 55 integration. They are not a current-run or new-base Spark/JNI result. At the complete snapshot cutoff `2026-09-06T04:28:28.027Z`, 17 checks had succeeded, 29 were running and seven were skipped, with no reported failures. No local build or runtime test was run, and complete CI remains unqualified.

## Performance

The rebase leaves key-only residual evaluation, reader attachment and execution-local filter construction unchanged. It adds no new runtime-filter allocation or row-processing pass. The PR's benchmark text is also unchanged: the original-fork native-Delta q52 experiment does not establish performance for this Apache revision or complete the existing disabled/residual-only/reader-plus-residual comparison. No new timing claim or duplicate benchmark request is proposed.

## Design

The new base does not alter the ownership or attachment model. The planner still applies filtering after the final build orientation, and each execution creates its own producer/consumer state. Nested shuffle support remains a separate capability with its own flag, while the filter's narrower eligibility and reader boundaries stay explicit.

## Abstraction & complexity

No additional runtime-filter wrapper, abstraction or test indirection was introduced by the rebase. The planner helper, private test module and DataFusion 55 delegation are unchanged. The source-equivalence checks support retaining the previous review conclusions without another restructuring request or duplicate inline finding.

native/core/src/execution/operators/dynamic_filter.rs
```diff
@@ -0,0 +1,572 @@
...
+        return binary.op() == &Operator::And
+            && is_direct_column_null_checks(binary.left())
+            && is_direct_column_null_checks(binary.right());
+    }
+    predicate
+        .downcast_ref::<IsNotNullExpr>()
+        .is_some_and(|is_not_null| is_not_null.arg().is::<Column>())
+}
+
+fn try_attach_parquet_reader_filter(
```
peterxcli, line 246: sorry I didnt see this before, but I think this could be move to planner, too.
and maybe is time to split the planner.rs into multiple rule files.

peterxcli (approved):
@pingzh thanks for the update! 
other LGTM, the inline comment can be address in followup

sunchao:
Merged, thanks @pingzh !!

PR #5700, opened 2026-09-04
chore: add benches array functions

coderfender:

Part of https://github.com/apache/datafusion-comet/issues/5396

benches for array functions (custom comet kernels)


native/spark-expr/benches/common/mod.rs
```diff
@@ -21,17 +21,21 @@
 #![allow(dead_code)]
 
 use arrow::array::{
-    builder::StringBuilder, ArrayRef, Float64Array, Int64Array, RecordBatch, StringArray,
-    TimestampMicrosecondArray,
+    builder::{BooleanBuilder, ListBuilder, StringBuilder},
+    ArrayRef, Float64Array, Int64Array, ListArray, RecordBatch,
```
sunchao: ### Correctness

[P2] Restore the existing array-type imports

Could you retain `StringArray` and `TimestampMicrosecondArray` in this import list? The unchanged `string_array` and `timestamp_micros_array` helpers still reference them at lines 76 and 97. Rust type-checks those helpers even when a particular benchmark does not call them, so this breaks both the seven new targets and existing users of `common/mod.rs`. The benchmark-check and Rust CI jobs on merge `013199ba` both fail with E0425 for these exact types. Restoring the two imports makes the shared helpers resolvable again.

sunchao (commented):
## Correctness

This adds seven Criterion targets for Comet's array kernels and shared typed-list builders. It fills part of the native benchmark coverage catalog without changing production expressions, serde or Spark compatibility gates. Six targets span ten element types, three row counts and three null ratios. Struct-field extraction adds three field projections, giving 567 distinct cases.

**[P2] Restore the two removed imports in the shared benchmark module.** `StringArray` and `TimestampMicrosecondArray` are still used by `string_array` and `timestamp_micros_array`. Removing their imports makes every target that includes `common/mod.rs` fail type checking, including existing benchmarks. Both the [benchmark check](https://github.com/apache/datafusion-comet/actions/runs/33927004194/job/101197916980) and [Rust job](https://github.com/apache/datafusion-comet/actions/runs/33927004194/job/101198042376) report those exact E0425 errors.

The selected cases are consistent with the maintained Spark 3.5/4.0 semantics: insertion at position 3, extraction at one-based index 2, slicing from position 2 for four elements, first-match array positions, paired zip fields and flattening four inner lists. Slice's Int64 native arguments are intentional because Comet's serde widens Spark's Int32 arguments. Decimal list values and their search/insert scalar both use precision 38 and scale 10. The projection ordinals match the fixed-width, string and nested-list fields. Null rows are represented separately from empty arrays.

### Validation and CI

I reviewed all nine changed files, the seven kernel paths, their serde contracts and maintained Spark implementations/tests. Static fixture analysis confirmed unique benchmark IDs, deterministic null/search distributions and bounded offsets. This was source and fixture analysis, not execution of the new benchmarks or a Spark/JNI compatibility suite.

At `2026-09-04T23:45:24.213Z`, current-head checks were 48 successful, 3 failed, 6 skipped and 15 running. The two Rust compilation failures above are PR-caused. The [Spark 3.5 expression job](https://github.com/apache/datafusion-comet/actions/runs/33927004194/job/101202351790) failed fetching `apache-rat-plugin` because Maven Central was unreachable, before tests started. All three checked out merge `013199ba`, whose parents are the reviewed base/head and whose tree matches the head. Maintained Spark 3.4/4.1 sources were unavailable.

## Performance

The benchmarks construct data and expressions before timing, invoke native kernels directly, and consume returned values with `black_box`. Spark constant folding and plan pruning are therefore outside this path. Timings include output destruction and, for UDF calls, argument cloning, scalar expansion and construction of invocation metadata. They measure that native invocation path rather than pure inner-loop cost or end-to-end Spark throughput.

The current workload scope is specific: eight-element lists, four-by-four flatten inputs, mostly unsuccessful numeric/string searches, and early successful Boolean searches. Zip uses the same array for both inputs, so lengths and nulls coincide. Struct-field projection exercises its buffer-sharing path because the structs and their fields are non-null. These are usable baseline cases after compilation is fixed, but results should not be generalized to long/empty lists, unequal zip lengths, null search values or null-bitmap merging. No speed comparison or new benchmark results are claimed.

## Design

Direct native calls fit the stated goal of isolating Comet expression cost from scans and JNI transfer. The setup uses real Arrow arrays and column expressions where required. Constant positions and search values remain arguments to the evaluated kernel. The three direct UDFs do not consult the placeholder return fields, so those placeholders do not turn these cases into null-only work.

No competing implementation is timed, so there is no invalid Spark-versus-DataFusion comparator to reconcile. The benchmarks provide repeatable individual workloads, not a cross-engine performance conclusion.

## Abstraction & complexity

The primitive, Boolean and string list builders remove substantial setup duplication while keeping each benchmark's arguments visible. The existing Criterion registration pattern remains easy to follow, and the small batch helpers do not introduce a new framework. Keeping the existing shared-module imports resolves the verified regression without requiring a larger redesign.

native/spark-expr/benches/common/mod.rs
```diff
@@ -20,18 +20,19 @@
 //! auto-discovery, which only looks at `benches/*.rs`, does not treat it as a bench target.
 #![allow(dead_code)]
 
-use arrow::array::{
-    builder::StringBuilder, ArrayRef, Float64Array, Int64Array, RecordBatch, StringArray,
-    TimestampMicrosecondArray,
+use arrow::array::{builder::{BooleanBuilder, ListBuilder, StringBuilder}, ArrayRef, Float64Array, Int64Array, ListArray, RecordBatch, StringArray, TimestampMicrosecondArray};
```
sunchao: ### Correctness

[P2] Format the restored import block to unblock validation

Could you apply the existing Rust formatter to this import block while retaining both restored names? The missing-import failure is fixed, but this replacement line makes Linux lint, macOS lint and benchmark validation fail their formatting checks. The benchmark job then skips Clippy and benchmark compilation. All three jobs checked out merge ebb2626b with this exact shared-module blob, and a local read-only rustfmt check using the head configuration produces the same multiline-import diff. This needs to pass before those validation stages can run.

sunchao (commented):
## Correctness

Follow-up at `183003a4` against `719cba11`, covering the update from `9fc41a9d`. The [previous missing-import P2](https://github.com/apache/datafusion-comet/pull/5700#discussion_r3938614934) is fixed in source: `StringArray` and `TimestampMicrosecondArray` are back in scope for the unchanged helpers.

**[P2] Format the restored import block so the validation gates can run.** The replacement puts the entire import on one line. Current [Linux lint](https://github.com/apache/datafusion-comet/actions/runs/33997640939/job/101391006972), [macOS lint](https://github.com/apache/datafusion-comet/actions/runs/33997640939/job/101391006884) and [benchmark validation](https://github.com/apache/datafusion-comet/actions/runs/33997640939/job/101391006932) all fail their formatting step on that exact line. The benchmark job skips Clippy and benchmark compilation after this failure. A read-only local `rustfmt --check` with the head's configuration produces the same import-block diff. Please apply the formatter's layout while retaining both restored names.

The inspected jobs checked out merge `ebb2626b`, whose parents are `f97fb451` and this head. Its shared benchmark module, all seven benchmark bodies and formatter configuration match the reviewed head. This is not CI execution against the authoritative base pair, and the failed formatting gate provides no successful benchmark-compilation or benchmark-runtime evidence. The final current-head CI refresh at `2026-09-05T23:35:15Z` reported seven successful, three failed, 16 skipped and five running checks.

I verified 30 previously reviewed Comet files against this head. Only the import block differs, so production kernels, serde, fixture values and the earlier maintained Spark 3.5/4.0 semantic assessment are unchanged. No additional behavioral regression was found in this delta. Maintained Spark 3.4/4.1 source gaps remain. No local native benchmark or Spark/JNI suite was executed.

## Performance

The seven benchmark bodies, deterministic input generators and Criterion loops are byte-identical to the previous review. Their workload coverage and setup/timing boundaries therefore have not changed. One timed iteration still invokes the kernel for the whole labeled batch and consumes the returned value with `black_box`, including output destruction and applicable invocation overhead. It is not a per-element timing. No new timing results or speedup claims are supplied, and compilation checks would not establish performance even after the formatting gate passes.

## Design

Restoring the existing imports is the appropriate local repair for the shared helper's unresolved names. It preserves the native column/array invocation paths and introduces no optimizer or constant-folding route. The remaining repair is to keep the required formatting around those imports so the existing validation pipeline can proceed.

## Abstraction & complexity

No helper, public interface, dependency or abstraction changed in this follow-up. The existing shared builders and benchmark registrations remain intact. No separate complexity finding.

coderfender:
fixed formatting

sunchao (approved):
## Correctness

Follow-up at `8bc52a3c13cd3876d529ee8087cb1111149ab392`, against `719cba11076e4237d6030925c49b1c1ffcac6f8e`, covering the update from `183003a4`. Both previous P2 findings are fixed. The required `StringArray` and `TimestampMicrosecondArray` imports remain present, and the [formatting finding](https://github.com/apache/datafusion-comet/pull/5700#discussion_r3942401778) is addressed by the multiline layout. The exact captured source passes a read-only local formatter check.

The update changes only import whitespace and an optional trailing comma. I rechecked 30 previously reviewed Comet files. The shared generator functions, all seven benchmark bodies, production kernels and serde are unchanged, so the earlier maintained Spark 3.5/4.0 semantic assessment still applies. No new or remaining verified P1/P2 finding in this delta. Maintained Spark 3.4/4.1 source gaps remain.

### Validation and CI

At the final check on **2026-09-06 02:49:06 UTC**, current-head checks are **65 successful / 8 skipped**, with none failing or pending. The [benchmark job](https://github.com/apache/datafusion-comet/actions/runs/33999498924/job/101395898626) passes Cargo formatting, workspace/all-targets Clippy and `cargo check --benches`. Its manifest registers all seven new targets without feature restrictions, and the workspace selection includes their crate. The [Linux formatting job](https://github.com/apache/datafusion-comet/actions/runs/33999498924/job/101395898523) also passes.

Both inspected jobs checked out merge `bcfc6e8d`, with parents `f97fb451` and this head. The benchmark bodies, shared module and formatter configuration match the reviewed head. Compilation evidence is limited to that verified merge environment: Rust 1.98.1, DataFusion 55.0.0 and Arrow 59.3.0. The authoritative head retains DataFusion 54.1.0 and Arrow 58.4.0. The repository selects `stable`, and the CI check does not use `--locked`, so an exact authoritative-head locked build remains unverified. No local native build, benchmark execution or Spark/JNI suite was run.

## Performance

The update preserves the existing inputs, setup placement and Criterion loops. Input construction remains outside the timed closure, and each iteration measures a whole native invocation with output consumption and destruction. The earlier workload limits remain unchanged. Passing compilation provides no new timing result or speedup claim.

## Design

This is a local formatting repair that retains the restored names and lets the existing validation stages proceed. It changes no expression construction, kernel dispatch or benchmark measurement boundary. No additional design issue in the update.

## Abstraction & complexity

No helper, interface, dependency or abstraction changes relative to the previous reviewed head. The existing shared generators and registrations remain intact, with no new indirection or complexity finding.

PR #5714, opened 2026-09-05
feat: fuse the typed Dataset map sandwich into a Comet projection

### LINKED ISSUE #5710: Fuse the SerializeFromObject / MapElements / DeserializeToObject sandwich into a Comet projection instead of falling back
Part of #5572.

`ds.map(f)` drops a three-operator island into the middle of an otherwise-native plan, and every operator in it falls back. Reproduced on `81d637b9b` against a Parquet table with a `case class ScratchRec(a: Int, b: String)`:

```
*(1) SerializeFromObject [invoke(knownnotnull(assertnotnull(input[0, ScratchRec, true])).a()) AS a#21, \
                          static_invoke(UTF8String.fromString(invoke(...).b())) AS b#22]
+- *(1) MapElements <lambda>, obj#18: ScratchRec
   +- *(1) DeserializeToObject newInstance(class ScratchRec), obj#15: ScratchRec
      +- *(1) CometColumnarToRow
         +- CometProject [a#4, b#5], [_1#2 AS a#4, _2#3 AS b#5]
            +- CometNativeScan parquet [_1#2,_2#3]
```

Only `DeserializeToObject` carries a fallback reason, because it is the lowest node whose children are all native; the two above it just inherit the island. Neither operator has an entry in `CometExecRule.nativeExecs` (`spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala:75`), so both take the generic "`X` is not supported" path at `:507`.

It cascades past the island. `ds.map(f).groupBy("b").count()` loses the aggregate and the exchange too:

```
 HashAggregate [COMET: Comet aggregate that merges intermediate buffers requires a Comet child aggregate ...]
+- AQEShuffleRead
   +- Exchange
      +- HashAggregate
         +- SerializeFromObject
            +- MapElements
               +-  DeserializeToObject [COMET: DeserializeToObject is not supported]
                  +- CometColumnarToRow
                     +- CometProject
                        +- CometNativeScan parquet

Comet accelerated 2 out of 8 eligible operators (25%).
```

## Why the mixin is not the answer here

`DeserializeToObjectExec.output` is exactly one attribute of `ObjectType(cls)` (Spark's `ObjectProducerExec.output`), and `SerializeFromObjectExec`'s input is the same. `ObjectType` is absent from `QueryPlanSerde.supportedDataType` (`QueryPlanSerde.scala:555`) and `serializeDataType` returns `None` for it (`:612`), and it is outside `CometBatchKernelCodegen.isSupportedDataType` (`CometBatchKernelCodegen.scala:85`) as the EPIC already notes. A JVM object reference cannot live in an Arrow `FieldVector`. Each of these operators has an object on one side of it, so neither can be a kernel in isolation — the kernel contract is Arrow-in, Arrow-out.

`CodegenDispatchFallback`'s self-type is also `CometExpressionSerde[_]` (`CometExpressionSerde.scala:125`), so there is no operator-level dispatch hook to mix it into. This is the same structural wall the EPIC records for the aggregate serdes.

## Why the fused sandwich is

Two things make the fused form legal, and neither needs new native code.

`canHandle` only inspects the **root** `dataType` (`CometBatchKernelCodegen.scala:120`) and every **`BoundReference`** `dataType` (`:171`). Intermediate nodes are never type-checked. So an `ObjectType` that exists strictly *inside* the tree passes the gate — it is only the boundary that has to be Arrow-representable, and for this sandwich the outer boundary is ordinary SQL data on both sides.

Spark also already builds the fused expression itself. `MapElementsExec.doConsume` constructs

```scala
val funcObj = Literal.create(func, ObjectType(funcClass))
val callFunc = Invoke(funcObj, funcName, outputObjectType, child.output, propagateNull = false)
```

so the user closure has a first-class Catalyst representation and we do not have to invent one. `DeserializeToObjectExec.doConsume` is `bindReference(deserializer, child.output).genCode(ctx)` and `SerializeFromObjectExec.doConsume` is the same over each serializer. The fused tree is therefore just: substitute `Invoke(func, deserializer)` into each serializer's `BoundReference(0)` — exactly what whole-stage codegen produces by chaining the three `doConsume`s today.

That means the shape of the change is a **plan rewrite into `ProjectExec`**, not a new operator serde. Recognize the sandwich, emit a `ProjectExec` whose expressions are the fused trees, and let the existing `CometProjectExec` plus `CometScalaUDF.emitJvmCodegenDispatch` path take it from there. No proto change, no native change.

#5692 landed `Invoke` as a `CometCodegenDispatch[Invoke]`, which is the piece that makes the middle of the sandwich dispatchable. Note that #5575 explicitly scoped *out* the encoder and deserializer trees on the grounds that their `ObjectType` boundary makes `canHandle` reject them — correct for a bare `Invoke`, and this issue is the follow-up that removes the boundary by fusing.

## The one design problem worth deciding up front

`emitJvmCodegenDispatch` emits one `JvmScalarUdf` and gets one output vector back. `SerializeFromObject` has N serializers that all share the *same* `Invoke`. Dispatching them as N separate expressions means N closure invocations per row: N times the cost, and observably wrong for a closure with side effects, since Spark's fused loop calls it once.

The fix is to make the kernel's single root a `CreateNamedStruct(serializer)` and project the N fields off it natively with `GetStructField`. Struct output is already supported — `CometBatchKernelCodegenOutput.outputVectorClass` maps `StructType` to `StructVector` (`CometBatchKernelCodegenOutput.scala:180`) — and both `CometCreateNamedStruct` and `CometGetStructField` already exist (`serde/structs.scala:34`, `:73`). `Invoke` is deterministic in Catalyst (`InvokeLike` does not override `deterministic`, and both children are deterministic), so the `doSubexpressionElimination = true` path in `generateSource` (`CometBatchKernelCodegen.scala:261`) hoists the shared `Invoke` into a single per-row call.

## Scope

| Shape | Middle operator | Fusable |
| --- | --- | --- |
| `ds.map(f)`, `ds.map(MapFunction)` | `MapElementsExec` | yes, per-row |
| `ds.mapPartitions`, `ds.flatMap` | `MapPartitionsExec(Iterator[Any] => Iterator[Any])` | no — arbitrary row count, no per-row expression exists |
| `groupByKey(...).mapGroups` / `cogroup` | `FlatMapGroupsExec`, `CoGroupExec` | no — group-at-a-time |
| `groupByKey` key extraction | `AppendColumnsExec` | per-row but widens the schema to `child.output ++ serializer`; a Project with passthrough, worth a follow-up rather than this issue |
| `ds.filter(func)` | none — `EliminateSerialization` and `TypedFilter` collapse to a plain `FilterExec` with an `Invoke` condition | already covered by #5692 |

So the target is `MapElementsExec`, i.e. `ds.map`. `ds.mapPartitions` was confirmed in the same run to produce the identical sandwich with `MapPartitions` in the middle, and it is not reachable this way.

## Payoff, and where it could regress

The closure does not get faster. It is the same `Invoke` on the same JVM lambda, once per row, in a Janino-compiled kernel loop instead of a whole-stage-codegen loop. The win is deleting the island, and it is worth having only when the typed operation sits *between* native operators — the `groupBy` case above, or a typed map feeding a join.

When the typed operation is at the top of the plan, which `ds.map(f).collect()` is, the gain is roughly nil: we would trade a `CometColumnarToRow` before the map for one after it, and it could come out slightly behind, since the kernel writes into Arrow only for something to immediately read rows back out. Worth checking whether `RevertNativeForTransitionHeavyStages` already covers that shape before assuming a new heuristic is needed.

## Risk to check before writing code

The lesson from #5575 applies directly: the mixin's contract covers whether `doGenCode` *compiles*, not limits that live at the **Arrow output boundary**. That is where the Iceberg `truncate(w, decimal)` regression came from.

Encoder serializers for decimals are the case I would check first. A `BigDecimal` encoder declares `DecimalType(38, 18)`, and the `StaticInvoke(Decimal.fromDecimal)` in the serializer can produce a value wider than the declared precision — Spark nulls it at row materialization, but the Arrow `Decimal128(p, s)` writer the kernel targets may not. Same failure shape as the Iceberg bug, so it needs a fixture before this lands.

Two smaller ones: `NewInstance` for a nested (non-top-level) case class carries an `outerPointer` closure over the enclosing instance, which closure-serialization would drag along — `ScalaUDF` has the same hazard, but it is worth a test. And any encoder that touches a type outside `isSupportedDataType` (`NullType`, UDTs, Variant) has to fall back cleanly at plan time rather than fail the Janino compile at execute time, which is what the `canHandle` gate is for.

## Verification note

The plan shapes and fallback reasons above were reproduced on `81d637b9b` with a throwaway suite; the fused-kernel design has not been prototyped. The `DecimalType` concern in particular is a hypothesis drawn from the #5575 postmortem, not a reproduced bug.

andygrove:

Closes #5710. Part of #5572.

## Rationale for this change

`ds.map(f)` drops a three-operator island into the middle of an otherwise native plan, and every operator in it falls back:

```
*(1) SerializeFromObject [invoke(knownnotnull(assertnotnull(input[0, TypedRec, true])).a()) AS a#21, ...]
+- *(1) MapElements <lambda>, obj#18: TypedRec
   +- *(1) DeserializeToObject newInstance(class TypedRec), obj#15: TypedRec
      +- *(1) CometColumnarToRow
         +- CometProject [a#4, b#5], [_1#2 AS a#4, _2#3 AS b#5]
            +- CometNativeScan parquet [_1#2,_2#3]
```

It cascades past the island. `ds.map(f).groupBy("b").count()` loses the aggregate and the exchange too — 2 of 8 eligible operators accelerated.

Neither object operator can be handled on its own. `DeserializeToObjectExec` outputs a single `ObjectType` attribute and `SerializeFromObjectExec` consumes one; `ObjectType` is outside `QueryPlanSerde.supportedDataType` and outside `CometBatchKernelCodegen.isSupportedDataType`, because a JVM object reference cannot live in an Arrow vector. `CodegenDispatchFallback`'s self-type is `CometExpressionSerde[_]`, so there is no operator-level dispatch hook to mix in either.

Fusing the sandwich works, though, for two reasons. `canHandle` only type-checks the **root** `dataType` and every **`BoundReference`** — intermediate nodes are never inspected — so the object may exist strictly _inside_ the tree while the outer boundary stays ordinary SQL data. And Spark already builds the fused expression: `MapElementsExec.doConsume` constructs `Invoke(Literal.create(func, ObjectType(funcClass)), funcName, outputObjectType, child.output, propagateNull = false)`, so the closure has a first-class Catalyst representation. The rule just does statically what whole-stage codegen does by chaining the three `doConsume`s.

`RewriteTypedDatasetMap` (new), run as a pre-pass in `CometExecRule._apply` before `transform`, since the bottom-up conversion would otherwise reach `DeserializeToObjectExec` and fall the island back before anything could fuse it. It rewrites the sandwich into a projection over the deserializer's child. The projection then converts through the ordinary `CometProjectExec` path and the fused tree routes through the JVM codegen dispatcher — **no proto change and no native change**.

**Multiple output columns.** `emitJvmCodegenDispatch` emits one `JvmScalarUdf` per expression and gets one Arrow vector back, but `SerializeFromObject` has N serializers that all share one `Invoke`. Converting them separately would produce N kernels and call the user closure N times per row where Spark calls it once — a real behavioural difference for a side-effecting closure, not just a slowdown. So for N > 1 the rule emits two stacked projections: an inner one producing a single `CreateNamedStruct`, and an outer one extracting the fields with `GetStructField`. Both node types already have serdes, and `CometBatchKernelCodegenOutput` already maps `StructType` to `StructVector`.

**`CometScalaUDF.FORCE_DISPATCH`.** The struct needs to compile into _one_ kernel, but `CreateNamedStruct` has a perfectly good native serde that would convert each field independently and put us back at N kernels. A `TreeNodeTag` checked at the top of `exprToProtoInternal` forces whole-subtree dispatch. I used a tag rather than a Comet-specific `Expression` subclass deliberately: the rewritten plan stays built entirely from stock Spark expressions, so it still executes correctly if the enclosing operator falls back to Spark for an unrelated reason.

**Scope.** Only `MapElementsExec`, including a chain of them (`ds.map(f).map(g)` leaves two adjacent `MapElements` under one Serialize/Deserialize pair). `MapPartitionsExec`, `FlatMapGroupsExec` and `CoGroupExec` consume iterators or groups, so no per-row expression exists. `AppendColumnsExec` is per-row but widens the schema; left for a follow-up. Typed filters never produce this sandwich — Catalyst lowers them to a `FilterExec` over an `Invoke`, which #5692 already dispatches.

**Declines rather than guesses** when the serializer has a non-`Alias` element, reads an unexpected bound reference, produces a dangling attribute reference, `canHandle` refuses the tree, the dispatcher is disabled, or (for N > 1) subexpression elimination is off. Each records a fallback reason so `EXPLAIN` says why instead of showing the bare "not supported".

**Off by default** (`spark.comet.exec.typedDatasetMap.enabled`). The rewrite pays off when the typed operation sits between native operators; when it is at the top of the plan (`ds.map(f).collect()`) the gain is roughly nil and could be slightly negative, since the kernel writes into Arrow only for something to read rows straight back out. Flipping the default needs benchmarks, and it is worth checking whether `RevertNativeForTransitionHeavyStages` already covers that shape.

New `CometTypedDatasetSuite`, 16 tests, registered in both `pr_build_linux.yml` and `pr_build_macos.yml`. Green on the default profile and on `-Pspark-3.5`; main and test code compile on `spark-3.4`, `spark-3.5`, `spark-4.0` and the default. No regressions in `CometCodegenSuite` (86), `CometCodegenSourceSuite` (60), `CometExpressionSuite` (141), `CometExecRuleSuite` (29) or `CometCoverageStatsSuite`.

Beyond `checkSparkAnswerAndOperator` on single-column, multi-column, wide (decimal / string / `Option`), nested-struct-and-array, chained-map and Java `MapFunction` shapes, the tests that carry the most weight:

- **`closure runs exactly once per row with multiple output columns`** — a JVM-static counter asserts 50 calls for 50 rows. This is the test that would catch the N-kernel regression the struct wrapper exists to prevent.
- **`fusion unblocks the aggregate and shuffle above it`** — asserts the `groupBy().count()` case is fully native, i.e. the cascade is actually fixed.
- **`output schema is unchanged by the rewrite`** — compares the schema tree with the flag on and off, guarding the nullability reasoning behind `GetStructField` over a non-nullable `CreateNamedStruct`.
- **`AssertNotNull inside the fused kernel still raises like Spark`** — returning null for a non-nullable product is an error in Spark; the serializer's `assertnotnull` has to survive the fuse or Comet would silently emit a null row. The stack trace confirms it fires from `SpecificCometBatchKernel.subExpr_0$`, which incidentally also confirms CSE hoisted the shared `Invoke`.
- **`decimal overflow in the serializer is caught before the Arrow write`** — I raised this on #5710 as the #5575-shaped risk: an encoder-declared `decimal(38,18)` receiving a wider value that Spark nulls at row materialization but the kernel's `DecimalVector` write might not. **It does not apply.** The encoder's serializer already wraps the value in `CheckOverflow`, so the fused tree raises under ANSI and nulls under non-ANSI exactly where Spark does, ahead of the write. The test pins both directions. I verified the same is true on the pre-existing plain-`ScalaUDF` decimal path.
- Negative tests for each decline path: off by default, dispatcher disabled, CSE disabled, and `mapPartitions` left alone.

Not covered, and worth a reviewer's attention: no benchmark numbers yet, which is why the flag is off. Nested (non-top-level) case classes carry a `NewInstance` `outerPointer` closure over the enclosing instance that closure serialization would drag along — `ScalaUDF` has the same exposure, but I have not written a test for it.

### DIFF
--- a/.github/workflows/pr_build_linux.yml
+++ b/.github/workflows/pr_build_linux.yml
@@ -471,6 +471,7 @@ jobs:
               org.apache.comet.CometUuidExpressionSuite
               org.apache.comet.serde.CometScalarFunctionSuite
               org.apache.comet.CometFallbackInvarianceSuite
+              org.apache.comet.CometTypedDatasetSuite
       fail-fast: false
     name: ${{ matrix.profile.name }} [${{ matrix.suite.name }}]
     runs-on: ubuntu-24.04
--- a/.github/workflows/pr_build_macos.yml
+++ b/.github/workflows/pr_build_macos.yml
@@ -244,6 +244,7 @@ jobs:
               org.apache.comet.CometUuidExpressionSuite
               org.apache.comet.serde.CometScalarFunctionSuite
               org.apache.comet.CometFallbackInvarianceSuite
+              org.apache.comet.CometTypedDatasetSuite
 
       fail-fast: false
     name: ${{ matrix.os }}/${{ matrix.profile.name }} [${{ matrix.suite.name }}]
--- a/docs/source/user-guide/latest/operators.md
+++ b/docs/source/user-guide/latest/operators.md
@@ -120,6 +120,23 @@ omitted from the tables below and may be reconsidered based on demand:
 | ------------------------ | ------ | ----------------------------------------------------------------- |
 | `DataWritingCommandExec` | ⚠️     | Experimental native Parquet writes, disabled by default (opt-in). |
 
+## Typed Dataset operations
+
+Typed `Dataset` operations bracket a JVM object in a `DeserializeToObject` / `SerializeFromObject`
+pair. Neither operator can run natively on its own, because its input or output is a raw JVM object
+reference that has no Arrow representation. What Comet can do is fuse a whole sandwich back into a
+projection when everything between the pair is per-row, so the typed operation no longer forces a
+Spark fallback island in the middle of an otherwise native plan. The user closure still runs on the
+JVM, once per row, inside Comet's codegen dispatcher, so results match Spark exactly.
+
+| Operator                                                                     | Status | Notes                                                                                                                                                                                                                                                               |
+| ---------------------------------------------------------------------------- | ------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `SerializeFromObjectExec` + `MapElementsExec` + `DeserializeToObjectExec`    | ⚠️     | The `ds.map(f)` sandwich is fused into a projection when `spark.comet.exec.typedDatasetMap.enabled=true` (off by default). Worth enabling when the typed operation sits between native operators ([#5710](https://github.com/apache/datafusion-comet/issues/5710)). |
+| `MapPartitionsExec`, `FlatMapGroupsExec`, `CoGroupExec`, `AppendColumnsExec` | 🔜     | These consume iterators or groups rather than rows, so there is no per-row expression to fuse. `AppendColumnsExec` is per-row but widens the schema and is not handled yet.                                                                                         |
+
+Typed filters (`ds.filter(func)`) do not produce this sandwich at all: Catalyst lowers them to an
+ordinary `FilterExec` whose condition is an `Invoke` of the closure, which Comet already dispatches.
+
 ## Python and UDF
 
 | Operator                                                                                | Status | Notes                                                                                                                        |
--- a/spark/src/main/scala/org/apache/comet/CometConf.scala
+++ b/spark/src/main/scala/org/apache/comet/CometConf.scala
@@ -396,6 +396,20 @@ object CometConf extends ShimCometConf {
       .booleanConf
       .createWithDefault(true)
 
+  val COMET_EXEC_TYPED_DATASET_MAP_ENABLED: ConfigEntry[Boolean] =
+    conf("spark.comet.exec.typedDatasetMap.enabled")
+      .category(CATEGORY_EXEC)
+      .doc("Experimental. Whether to fuse the `SerializeFromObject` / `MapElements` / " +
+        "`DeserializeToObject` operator sandwich that a typed `Dataset.map` produces into a " +
+        "Comet projection, so the typed operation no longer forces a Spark fallback island in " +
+        "the middle of an otherwise native plan. The user closure still runs on the JVM, once " +
+        "per row, inside Comet's Arrow-direct codegen dispatcher, so results match Spark. Off " +
+        "by default: the rewrite only pays off when the typed operation sits between native " +
+        "operators, and can cost a little when it is at the top of the plan. Requires " +
+        s"${COMET_SCALA_UDF_CODEGEN_ENABLED.key}=true.")
+      .booleanConf
+      .createWithDefault(false)
+
   val COMET_SHUFFLE_NATIVE_HASH_PARTITIONING_ENABLED: ConfigEntry[Boolean] =
     conf("spark.comet.shuffle.native.partitioning.hash.enabled")
       .withAlternative("spark.comet.native.shuffle.partitioning.hash.enabled")
--- a/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
+++ b/spark/src/main/scala/org/apache/comet/codegen/CometBatchKernelCodegen.scala
@@ -124,9 +124,22 @@ object CometBatchKernelCodegen extends Logging with CometExprTraitShim with Come
     // typed input field count, the typed-getter switch, and the constant pool. Refuse here so the
     // operator falls back to Spark cleanly rather than tripping a Janino compile failure
     // mid-execution (Comet has no recovery for that).
+    //
+    // Count each input ordinal once. WSCG gates on the operator's *schema*
+    // (`plan.schema.map(_.dataType).map(numOfNestedFields).sum`), so a column read more than once
+    // contributes once; the kernel likewise emits one typed field and one getter per ordinal, not
+    // per occurrence. Counting occurrences instead would scale with how often the tree happens to
+    // repeat a column, which is what `RewriteTypedDatasetMap` does deliberately: it puts the same
+    // shared subtree in every struct field so subexpression elimination can collapse it. A
+    // 12-column typed map reads 12 ordinals in each of 12 fields, which counted per occurrence
+    // came to 156 and refused a kernel that is really 24 fields wide.
     val maxFields = SQLConf.get.wholeStageMaxNumFields
-    val totalFields = numOfNestedFields(boundExpr.dataType) +
-      boundExpr.collect { case b: BoundReference => numOfNestedFields(b.dataType) }.sum
+    val inputFields = boundExpr
+      .collect { case b: BoundReference => b.ordinal -> b.dataType }
+      .distinct
+      .map { case (_, dt) => numOfNestedFields(dt) }
+      .sum
+    val totalFields = numOfNestedFields(boundExpr.dataType) + inputFields
     if (totalFields > maxFields) {
       return Some(
         s"codegen dispatch: too many nested fields ($totalFields > " +
--- a/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
+++ b/spark/src/main/scala/org/apache/comet/rules/CometExecRule.scala
@@ -707,13 +707,26 @@ case class CometExecRule(session: SparkSession)
         normalizedPlan
       }
 
+      // Collapse the SerializeFromObject / MapElements / DeserializeToObject sandwich that a typed
+      // `Dataset.map` produces into a projection, before the bottom-up conversion sees the
+      // individual operators. `transform()` visits children first, so by the time it reached
+      // `DeserializeToObjectExec` the island would already have fallen back.
+      val planWithTypedMapFused =
+        if (CometConf.COMET_EXEC_TYPED_DATASET_MAP_ENABLED.get(conf)) {
+          planWithJoinRewritten.transformUp { case p =>
+            RewriteTypedDatasetMap.rewrite(p)
+          }
+        } else {
+          planWithJoinRewritten
+        }
+
       // Tag Partial aggregates that must not be converted to Comet because a
       // corresponding Final or PartialMerge cannot be converted and the intermediate buffer
       // formats are incompatible. This runs before transform() so the tags are checked
       // during the bottom-up conversion. Tags persist through AQE stage creation.
-      tagUnsafePartialAggregates(planWithJoinRewritten)
+      tagUnsafePartialAggregates(planWithTypedMapFused)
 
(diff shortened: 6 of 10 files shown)


andygrove:
Pushed a review pass. One finding was a real limitation rather than a cleanup, so flagging it here:

**`canHandle`'s `spark.sql.codegen.maxFields` gate counted `BoundReference` occurrences, not distinct ordinals.** The fused struct puts the shared deserializer in every field deliberately, so a 12-column typed map counted 12 + 12*12 = 156 against the default cap of 100 and silently declined to fuse. The feature capped out around 9 columns and nothing caught it, since the widest record in the suite had 4 fields. Fixed at the root: WSCG gates on the operator's schema, where a column read twice contributes once, and the kernel likewise emits one typed field and one getter per ordinal. Now pinned by a 12-column test that I confirmed fails without the fix. Note this touches shared code, so it also slightly relaxes the gate for `ScalaUDF` — in the correct direction.

Also worth a reviewer's attention, and deliberately **not** changed here:

**This rewrite could live in `convertNode` instead of as an `_apply` pre-pass, and probably should.** My original comment justified the pre-pass by claiming the bottom-up walk would have already fallen the island back — that is wrong, and I have corrected it. `convertNode` only tags `DeserializeToObjectExec` with a fallback reason; it never replaces it, so the sandwich is structurally intact when the walk reaches `SerializeFromObjectExec`. Doing the fusion there would let the rule see whether `deserialize.child` actually converted to a `CometNativeExec` and decline when it did not — which is exactly the profitability signal this rewrite currently lacks and the reason the config is off by default. That is a ~55 line change and a behavior change (it would decline more often), so I left it out of a cleanup pass, but it is the thing most likely to let the flag default to `true`.

The rest is cleanup: extracted `CometScalaUDF.canDispatch` (the rule was re-deriving the bind-then-`canHandle` gate and had already drifted — it omitted the `RuntimeReplaceable` unwrap); moved `FORCE_DISPATCH` onto `QueryPlanSerde` which reads it, matching what every other behavior tag in the codebase does, and documented that the tag bypasses the per-expression policy layer; moved the tag check ahead of the version shim so the override is unconditional; moved the pre-pass above `normalizePlan` so the synthesized projections get the same NaN / -0.0 treatment as every other `ProjectExec`.

On tests: one assertion was vacuous — `Dataset.schema` comes from the analyzed plan, which a physical rule cannot touch — so it now compares the executed plan's output attributes instead. Both decline paths now pin their message via `checkSparkAnswerAndFallbackReason` rather than just asserting the sandwich survived.

17 tests, green on the default profile and `-Pspark-3.5`; compiles on 3.4/3.5/4.0/default. No regressions in `CometCodegenSuite` (86), `CometCodegenSourceSuite` (60), `CometCodegenFuzzSuite` (28), `CometExpressionSuite` (141), `CometExecRuleSuite` (29).

PR #5720, opened 2026-09-05
feat: route next_day and levenshtein collated input through the codegen dispatcher

### LINKED ISSUE #5591: `next_day` and `levenshtein` fall back to Spark on collated strings
Part of #5572.

Two datetime and string serdes decline non-default collations and fall the whole projection back to Spark:

- `CometNextDay` returns `Incompatible(collationReason)` (`spark/src/main/scala/org/apache/comet/serde/datetime.scala:451`)
- `CometLevenshtein` returns `Unsupported` (`spark/src/main/scala/org/apache/comet/serde/strings.scala:128-130`)

Neither mixes in `CodegenDispatchFallback`. Both are plain codegen expressions over `StringType`, which `CometBatchKernelCodegen.isSupportedDataType` accepts regardless of collation ID, so the dispatcher can run Spark's own collation-aware code inside the pipeline.

This is the same treatment the collation-declining serdes elsewhere already get — `CometFromUTCTimestamp` (`datetime.scala:366`), `CometToUTCTimestamp` (`:386`), `CometConvertTimezone` (`:406`), `CometTruncDate` (`:533`), `CometTruncTimestamp` (`:587`), `CometDateFormat` (`:691`) and `CometConcat` (`strings.scala:256`) all carry the mixin.

`CometUnixTimestamp` has the same collation arm at `:314` but is covered by its own issue, since its bigger problem is string input.

Fix: mix `CodegenDispatchFallback` into `CometNextDay` and `CometLevenshtein`.

Worth a sanity check first that collated string columns actually reach Comet operators in the first place, rather than being rejected earlier at the scan or schema-support layer — if they never get that far, these two are dead code and the issue should be closed rather than fixed. `CometCollationSuite` is the place to confirm.

0lai0:

Closes #5591. Part of #5572.

## Rationale for this change

`CometNextDay` and `CometLevenshtein` decline non-UTF8_BINARY collated arguments but do not mix in `CodegenDispatchFallback`, so a collated argument fails the whole projection back to Spark. Both are ordinary codegen expressions, so the dispatcher can run Spark's own `doGenCode` inside the Comet pipeline instead.

Collated values do reach Comet operators, but only from the query text (`col COLLATE x` or `CAST(col AS STRING COLLATE UTF8_LCASE)`). A table schema that declares a collated column is rejected at scan (`Unsupported schema ... StringType(UTF8_LCASE)`). The guards are live.

Neither Spark implementation consults collation (`getDayOfWeekFromString` / `UTF8String.levenshteinDistance` take no collation id on 4.0.1 or 4.1.1, neither appears in `CollationSupport`). That suggested dropping the guards and running the native kernels. I implemented that variant and benchmarked it. It lost on every axis, so this PR does what the issue originally proposed.

- Mix `CodegenDispatchFallback` into `CometNextDay` and `CometLevenshtein`.
- Collapse `CometLevenshtein`'s two collation reason strings into one `private val`, reworded so it does not contradict the `GenerateDocs` "runs in the JVM" header for mixin serdes.
- `expressions.md`: both rows `Native` → `Hybrid`.
- `understanding-comet-plans.md`: `levenshtein` was wrongly listed as `CometCodegenDispatch` with no native kernel. The example now uses `pmod`.
- Collated cases on `CometStringExpressionBenchmark` and `CometDatetimeExpressionBenchmark` (`isSpark40Plus`; `COLLATE` does not parse on 3.4/3.5). `next_day` had no benchmark at all, so the default-collation case is added too.

## Benchmark
1,048,576 rows, `spark-4.0`, `local[1]`. "guard removed" is the native-path alternative above; it is not in this PR.

| Case | Spark | guard removed | this PR |
| --- | --- | --- | --- |
| `next_day`, UTF8_LCASE | 89 ms | 125 ms (0.7X) | **70 ms (1.3X)** |
| `levenshtein`, UTF8_LCASE | 378 ms | 489 ms (0.8X) | **341 ms (1.1X)** |
| `next_day`, default (control) | 90 ms | 104 ms | 104 ms |
| `levenshtein`, default (control) | 380 ms | 439 ms | 435 ms |

Controls match between variants, so only the collated path differs. Dispatching is also the only variant that covers `col COLLATE x`: `Collate` has no serde, and a `Compatible` expression never reaches the dispatcher.

Added benchmark cases on a second profile / data shape: collated `next_day` 1.4X on `spark-4.1`; collated `levenshtein` 2.6X / 3.2X on `spark-4.0`. Default-collation `next_day` is 1.0X.

New Spark 4.0+ SQL fixtures:

- `datetime/next_day_collation.sql`, `datetime/next_day_ansi_collation.sql`
- `string/levenshtein_collation.sql`

Column and literal arguments, `UTF8_LCASE` / `UNICODE_CI` / RTRIM collations, non-ASCII, bad day names, ANSI errors, three-argument `levenshtein`, and NULL in data. NULL literals are omitted:
both expressions are `nullIntolerant`, so `NullPropagation` folds them before Comet sees them.

The load-bearing case is `levenshtein('HELLO' COLLATE UTF8_LCASE, 'hello' COLLATE UTF8_LCASE)` returning 5, not 0. That query fails first if Spark ever makes `levenshtein` collation-aware.

`CometCollationSuite` (4.0 and 4.1) adds a `next_day` dispatcher test and an `allowIncompatible=true` native-path test, matching `date_format`. The latter is the only coverage of the native kernel on collated input. `CometStringExpressionSuite` asserts the `levenshtein` path through `assertCodegenRan`.

Local runs, no style checks skipped: `spark-4.0` 566/0/0, `spark-4.1` 561/0/0, `spark-3.5` 539/0/1 (`assume(isSpark40Plus)` skip). Suites: `CometCollationSuite`, `CometSqlFileTestSuite`, `CometTemporalExpressionSuite`, `CometStringExpressionSuite`.

## Known gap
`col COLLATE utf8_lcase` works here only because the dispatcher compiles the whole bound subtree, including `Collate`. Comet has no serde for `Collate` (`QueryPlanSerde.scala:1009`), so any `Compatible()` expression over a `Collate` child still fails the projection, e.g.`length(s COLLATE utf8_lcase)`. That is wider than this issue and predates this PR.

### DIFF
--- a/docs/source/user-guide/latest/expressions.md
+++ b/docs/source/user-guide/latest/expressions.md
@@ -287,7 +287,7 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `month` | ✅ | Native |  |
 | `monthname` | ✅ | — | Abbreviated month name (Spark 4.0+) |
 | `months_between` | ✅ | Codegen dispatch |  |
-| `next_day` | ✅ | Native |  |
+| `next_day` | ✅ | Hybrid | Non-UTF8_BINARY collated `dayOfWeek` routes through the JVM codegen dispatcher; other input runs natively |
 | `now` | ✅ | — | Constant-folded to a literal (alias of `current_timestamp`) |
 | `quarter` | ✅ | Native |  |
 | `second` | ✅ | Native |  |
@@ -576,7 +576,7 @@ The type-name conversion functions (`bigint`, `binary`, `boolean`, `date`, `deci
 | `left` | ✅ | Native |  |
 | `len` | ✅ | Native |  |
 | `length` | ✅ | Native |  |
-| `levenshtein` | ✅ | Native |  |
+| `levenshtein` | ✅ | Hybrid | Non-UTF8_BINARY collated input routes through the JVM codegen dispatcher; other input runs natively |
 | `locate` | ✅ | Codegen dispatch |  |
 | `lower` | ✅ | Hybrid |  |
 | `lpad` | ✅ | — |  |
--- a/docs/source/user-guide/latest/understanding-comet-plans.md
+++ b/docs/source/user-guide/latest/understanding-comet-plans.md
@@ -122,7 +122,7 @@ natively in DataFusion" from "runs Spark's generated code inside a Comet
 kernel". The annotation only appears for expressions Comet actually routed
 through the dispatcher on this plan — either because no native DataFusion
 implementation exists (`CometCodegenDispatch` serdes such as `hypot`,
-`levenshtein`, `pmod`), or because a native path exists but declined this
+`pmod`, `months_between`), or because a native path exists but declined this
 specific input (`CodegenDispatchFallback` mixins). It is not a general "this
 expression ran here" marker.
 
@@ -132,7 +132,7 @@ Example:
 spark.conf.set("spark.comet.exec.scalaUDF.codegen.enabled", "true")
 spark.conf.set("spark.comet.explain.codegen.enabled", "true")
 
-val df = spark.sql("SELECT hypot(a, b), levenshtein(s1, s2) FROM t")
+val df = spark.sql("SELECT hypot(a, b), pmod(a, b) FROM t")
 println(new org.apache.comet.ExtendedExplainInfo()
   .generateExtendedInfo(df.queryExecution.executedPlan))
 ```
@@ -141,14 +141,14 @@ Output:
 
 ```
 CometColumnarToRow
-+- CometProject [COMET-INFO: JVM codegen dispatcher: hypot, levenshtein]
++- CometProject [COMET-INFO: JVM codegen dispatcher: hypot, pmod]
    +- CometNativeScan parquet spark_catalog.default.t
 
 Comet accelerated 2 out of 2 eligible operators (100%). Final plan contains 1 transitions between Spark and Comet. Accelerated expressions: 0 native, 2 codegen dispatch.
 ```
 
 Note that the operator is still `CometProject` (Comet-accelerated); only the
-per-expression evaluation for `hypot` and `levenshtein` takes the JVM codegen
+per-expression evaluation for `hypot` and `pmod` takes the JVM codegen
 path.
 
 ### `spark.comet.explain.format`
--- a/spark/src/main/scala/org/apache/comet/serde/datetime.scala
+++ b/spark/src/main/scala/org/apache/comet/serde/datetime.scala
@@ -433,13 +433,16 @@ object CometConvertTimezone
   }
 }
 
-object CometNextDay extends CometExpressionSerde[NextDay] {
+/**
+ * The native `next_day` kernel reads its `dayOfWeek` argument as raw bytes, so a non-UTF8_BINARY
+ * collation is reported as `Incompatible` and `CodegenDispatchFallback` routes it through the JVM
+ * codegen dispatcher instead of failing the projection back to Spark. Dispatching is not only the
+ * compatible answer here, it is also the faster one: over 1M rows a collated `next_day` measured
+ * 70ms dispatched against 89ms for Spark. See
+ * https://github.com/apache/datafusion-comet/issues/5591.
+ */
+object CometNextDay extends CometExpressionSerde[NextDay] with CodegenDispatchFallback {
 
-  /**
-   * `failOnError` mirrors `spark.sql.ansi.enabled`: under ANSI, Spark throws on a malformed
-   * `dayOfWeek` rather than returning NULL. The resolved flag is passed to native via the
-   * `ScalarFunc.fail_on_error` field.
-   */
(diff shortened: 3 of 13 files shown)


sunchao (approved):
## Correctness

Reviewed head `2a21edfc526a1921886b65445146e5a2b9fabb65` against base `7190df631afe3795914839203c7afe57ea23903c`. I found no new actionable P1/P2 in the reviewed execution paths.

Previously, the non-UTF8_BINARY guards on `next_day` and `levenshtein` could make an otherwise supported projection fall back to Spark. Adding `CodegenDispatchFallback` enrolls those cases in the existing dispatcher. It binds and serializes the complete expression, including a query-level `Collate`, and runs Spark's generated code over Arrow inputs. The default UTF8_BINARY converters and native kernels are unchanged. Disabling expression support or the dispatcher preserves fallback. `next_day` also retains its existing `allowIncompatible` native opt-in, while `levenshtein`'s `Unsupported` case always tries the dispatcher.

Against the maintained Spark 3.5 and 4.0 sources, `next_day` returns a nullable date, selects a strictly later matching weekday, uppercases names with `Locale.ROOT`, and does not trim padded names. Spark 4.0 accepts collated weekday strings, including RTRIM collations, but its weekday helper does not use a collation key. The dispatched expression retains `failOnError`, child casts and their resolved timezone state. Invalid names therefore follow Spark's generated ANSI error/non-ANSI null branches. Null operands and the output null flag are preserved.

`levenshtein` returns an integer and accepts two strings plus an optional integer threshold. Spark's helper compares UTF-8 code-point sequences without case folding or collation-key normalization, so `HELLO` versus `hello` remains five edits and a trailing space remains an edit under RTRIM. Dispatch preserves the two generated branches, null checks, negative-threshold result of `-1`, and zero/exact/large-threshold handling. The explicit return types also remain correct: the dispatched `JvmScalarUdf` avoids DataFusion's scalar-function signature inference, and the unchanged native converter already supplies `IntegerType` for the three-argument form.

The new fixtures cover column and literal strings, null input rows, multiple collations, Unicode, invalid weekdays, ANSI errors and positive thresholds. Constant folding is disabled in the SQL harness. Levenshtein also asserts dispatcher activity, while NextDay checks the Comet operator and retains a dispatcher-disabled fallback test. The ANSI sentinel establishes that a valid collated case stays in Comet, although an `expect_error` query alone does not establish its execution path. No nullable/nonconstant-threshold or interpreter-mode query was run in this review. The compatibility assessment follows Spark's generated execution and does not assert parity with the existing nullable-threshold discrepancy in Spark's internal `NO_CODEGEN` testing mode. Maintained Spark 3.4/4.1 sources were unavailable, so those versions are not source-qualified here.

At the authenticated cutoff of 2026-09-05 21:35:04 UTC, there were no reviews or discussion comments and no existing current-head approval. CI, CodeQL, Check PR Title and Delta Contrib Build Gate all require action, with zero head or merge check runs. The current merge `bff8564dc7187dd99f913252762e7f43e63d9eb8` has the expected base/head parents and exactly the head's tree. The changed paths select the relevant build/test filters, but no execution is credited. Spark 4.0's PR test job additionally requires its label. The whitespace check passed. I ran no local product build, Spark/JNI test or benchmark. The suite totals in the description remain author reports.

## Performance

The added benchmarks exercise column inputs and include a default-collation NextDay control and both Levenshtein arities. Their shared harness disables constant folding for both engines and checks the physical plan, although fallback produces a warning rather than a failed benchmark. The checked-in NextDay case uses 1,048,576 rows, while the string benchmark extends the existing 1,024-row data shape. These should remain distinct from the separate million-row experiment in the description.

The author reports 70 ms versus 89 ms for collated NextDay and 341 ms versus 378 ms for collated Levenshtein. I have not independently reproduced those timings or validated raw benchmark logs. The reported default-collation Comet controls are slower than Spark, but this PR leaves those kernels and routing unchanged. The dispatcher reuses its per-task kernel and Spark bytecode caches, with the existing JNI, allocation and synchronization costs per batch. No new material overhead on the unchanged native path was identified, and the supplied focused benchmarks address the performance question for this enrollment.

## Design

Reusing the dispatcher is a small, coherent change. Compiling the whole expression handles the `Collate` wrapper without requiring a new native serde for it, and keeps the established configuration and admission boundaries. Collated table schemas still follow the existing scan fallback. Removing the guards alone would not cover an unfolded `Collate` child, so the broader alternative would require additional work. The expression table now describes both functions as hybrid, and the plan-explanation example correctly uses expressions that always dispatch.

## Abstraction & complexity

The two mixins reuse an existing extension point without adding a dispatcher, UDF registry entry, protocol type or kernel. Consolidating Levenshtein's reason string removes duplicated metadata. Binding, closure serialization, Arrow representation and cache lifetime remain shared. No additional abstraction or broader refactoring is needed for this change.

andygrove (approved):
Thanks @0lai0

0lai0:
Thanks @sunchao and @andygrove

PR #5722, opened 2026-09-05
perf: cache expected schemas for remote shuffle decoding

### LINKED ISSUE #5535: Native Celeborn shuffle: the remote read path costs up to 4.5x local decode and is unbenchmarked

**Component**: `native/shuffle/src/ipc.rs` (`read_single_batch`),
`native/core/src/execution/jni_api.rs:1265` (`decodeShuffleBlockWithValidation`)

`read_ipc_compressed_validated` drops `with_skip_validation(true)`, so every remotely fetched batch
pays full Arrow buffer and offset validation. That is the right call for untrusted bytes, but the
cost is significant and is not recorded anywhere. Since the justification for a native Celeborn
reader is that it beats Celeborn's own row decoder, this number should be measured before 8/n
enables the path end to end.

## Measurement

Release build, one `ShuffleBlockWriter` frame of 20 columns (10 `Int32`, 10 `Utf8`) x 8192 rows,
300 iterations after 20 warm-up, Apple silicon. "local" is `read_ipc_compressed`; "validated" is
`read_ipc_compressed_validated` + `validate_remote_schema`, i.e. what the reducer runs.

| codec    | frame bytes | local (us/blk) | validated (us/blk) | ratio |
| -------- | ----------- | -------------- | ------------------ | ----- |
| None     | 1,570,252   | 38.8           | 175.2              | 4.52x |
| Lz4Frame | 752,215     | 225.8          | 356.6              | 1.58x |
| Snappy   | 821,025     | 558.5          | 696.5              | 1.25x |
| Zstd(1)  | 235,724     | 636.9          | 773.1              | 1.21x |

For reference, the local path itself did not regress in #5531: against the pre-PR implementation it
measured -8.8% (None), +0.8% (Lz4), +1.1% (Snappy), -1.8% (Zstd). The table above is purely the
price of the new remote path.

## Two hoistable costs on the same path

- `decodeShuffleBlockWithValidation` calls `ShuffleScan::decode` and rebuilds `expected_types` via
  `to_arrow_datatype` on **every block**. Both are constant per reducer.
- `env.convert_byte_array(expected_schema)` copies the serialized schema into a fresh `Vec` per
  block.

Caching the decoded `Vec<DataType>` once per reducer, e.g. behind a native handle, would remove
both.

## Ask

Publish a benchmark for the remote read path in the 8/n description, and decide whether the
validation cost is acceptable or should be made configurable.

pingzh:

Closes #5535.

## Rationale for this change

The JVM remote shuffle reader copies and parses the same expected schema on every block. Retain the parsed Spark types for the iterator's lifetime so subsequent blocks avoid JNI schema copies and repeated protobuf/type conversion.

- Lazily create a native decoder on the first remote block and reuse it across subsequent blocks. Release it once on iterator close, including early completion and cleanup failures. Empty and local streams allocate no decoder.
- Coordinate decoder creation, decoding, Arrow import, and batch ownership with concurrent task cleanup. Keep transport reads and failure-reporting RPCs outside the monitor so cleanup can proceed while they are blocked.
- Keep Arrow buffer/offset and logical type validation mandatory. Decoder initialization failures remain outside remote corruption reporting.

### Remote decode measurements

Linux x86_64, AMD EPYC 9V74, Rust 1.97.1, DataFusion 55.0.0, Arrow 59.3.0; default release profile, 100 samples, 3-second warm-up, 5-second target measurement time. Each block contains 8192 rows and 20 columns (10 Int32, 10 Utf8). Times are Criterion's estimated mean in microseconds per block.

| Encoding | Codec | Frame bytes | Local | Remote | Remote/local |
| --- | --- | ---: | ---: | ---: | ---: |
| plain | None | 2,075,868 | 106.7 | 290.5 | 2.72x |
| plain | LZ4 | 704,026 | 410.4 | 602.9 | 1.47x |
| plain | Snappy | 785,894 | 577.1 | 774.1 | 1.34x |
| plain | Zstd(1) | 248,249 | 1577.5 | 1776.2 | 1.13x |
| dictionary | None | 709,660 | 67.0 | 751.5 | 11.21x |
| dictionary | LZ4 | 342,473 | 191.3 | 877.8 | 4.59x |
| dictionary | Snappy | 370,097 | 259.6 | 943.4 | 3.63x |
| dictionary | Zstd(1) | 205,611 | 359.3 | 1053.5 | 2.93x |

The remote dictionary path also expands values for the JVM importer, so that gap includes normalization. Measurements include decompression, IPC decoding, validation/normalization, and batch destruction; they exclude transport, JNI, Arrow FFI, and schema initialization. They do not measure the schema-cache speedup or establish an end-to-end advantage over Celeborn's row decoder. Fixtures differ from the issue's original measurements. Validation remains required despite its cost.

- `CometCelebornShuffleReaderSuite` on Spark 4.1 and Spark 3.4.3: 69 passed on each against the new release JNI library, including multi-frame reuse, later-frame schema mismatch, initialization failure classification, and cleanup lifecycle coverage.
- Five new latch-based concurrency regressions failed before the lifetime fix and passed afterward. They cover cleanup during decoder creation, active decoding (including interrupted cleanup), Arrow import and batch publication, transport reads, and failure reporting.
- `cargo test --release -p datafusion-comet-shuffle --lib`: 118 passed.
- Native release build, full JVM reactor test compilation, Spotless, Scalastyle, `cargo fmt --check`, and `git diff --check` passed.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/NativeBatchDecoderIterator.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/NativeBatchDecoderIterator.scala
@@ -45,11 +45,14 @@ case class NativeBatchDecoderIterator(
     expectedSchema: Option[Array[Byte]] = None)
     extends Iterator[ColumnarBatch] {
 
+  // One consumer reads this iterator, while task completion may close it from another thread.
+  // The monitor protects decoder and batch ownership; transport reads stay outside it.
   private var isClosed = false
   private val longBuf = ByteBuffer.allocate(8).order(ByteOrder.LITTLE_ENDIAN)
   private var currentBatch: ColumnarBatch = null
   private var batch: Option[ColumnarBatch] = None
   private val validateRemoteFrames = in.isInstanceOf[CometShuffleReadFailureHandler]
+  private var remoteDecoderHandle = 0L
 
   require(
     !validateRemoteFrames || expectedSchema.exists(_ != null),
@@ -64,25 +67,23 @@ case class NativeBatchDecoderIterator(
   }
 
   override def hasNext: Boolean = {
-    if (channel == null || isClosed) {
-      return false
-    }
-    if (batch.isDefined) {
-      return true
+    synchronized {
+      if (channel == null || isClosed) {
+        return false
+      }
+      if (batch.isDefined) {
+        return true
+      }
+
+      // Clear ownership before cleanup so a failed close cannot release this batch twice.
+      if (currentBatch != null) {
+        val previous = currentBatch
+        currentBatch = null
+        previous.close()
+      }
     }
 
-    // Release the previous batch.
-    if (currentBatch != null) {
-      currentBatch.close()
-      currentBatch = null
-    }
-
-    batch = fetchNext()
-    if (batch.isEmpty) {
-      close()
-      return false
-    }
-    true
+    fetchNext()
   }
 
   def next(): ColumnarBatch = {
@@ -90,52 +91,76 @@ case class NativeBatchDecoderIterator(
       throw new NoSuchElementException
     }
 
-    val nextBatch = batch.get
-
-    currentBatch = nextBatch
-    batch = None
-    currentBatch
-  }
-
-  private def fetchNext(): Option[ColumnarBatch] = {
-    // The remote input owns metadata, transport, and frame-boundary failure classification. Do
-    // not turn deliberately unreported metadata timeouts or stream-close failures into corruption.
-    val block = readNextBlock()
-    block.flatMap { case (fieldCount, dataBuf, bytesToRead) =>
-      // Allocation and Arrow import failures are not evidence of a corrupt remote shuffle. Only
-      // forward failures from native codec/IPC decoding and logical type validation.
-      val startTime = System.nanoTime()
-      val decoded = nativeUtil.getNextBatch(
-        fieldCount,
-        (arrayAddrs, schemaAddrs) => {
-          handleReadFailure {
-            if (validateRemoteFrames) {
-              nativeLib.decodeShuffleBlockWithValidation(
-                dataBuf,
-                bytesToRead,
-                arrayAddrs,
-                schemaAddrs,
-                tracingEnabled,
-                expectedSchema.get)
-            } else {
-              nativeLib.decodeShuffleBlock(
-                dataBuf,
-                bytesToRead,
-                arrayAddrs,
-                schemaAddrs,
-                tracingEnabled)
-            }
-          }
-        })
-      decodeTime.add(System.nanoTime() - startTime)
-      decoded
+    synchronized {
+      // Completion may have closed the iterator after hasNext() returned.
+      if (isClosed) {
+        throw new NoSuchElementException
+      }
+      currentBatch = batch.get
+      batch = None
+      currentBatch
     }
   }
 
-  private def handleReadFailure[T](read: => T): T = {
-    try read
-    catch {
-      case NonFatal(failure) =>
+  private def fetchNext(): Boolean = {
+    // The remote input owns metadata, transport, and frame-boundary failure classification. Do
+    // not turn deliberately unreported metadata timeouts or stream-close failures into corruption.
+    // Read outside the monitor so close() can unblock the underlying stream.
+    val block = readNextBlock()
+    var nativeFailure: Throwable = null
+    try {
+      synchronized {
+        // Cleanup may have finished during the read. Do not create a decoder or import a batch
+        // after close(), and publish each decoded batch before cleanup can inspect ownership.
+        if (isClosed) {
+          return false
+        }
+        batch = block.flatMap { case (fieldCount, dataBuf, bytesToRead) =>
+          val startTime = System.nanoTime()
+          // Invalid expected schemas are setup failures, not evidence of corrupt persisted data.
+          if (validateRemoteFrames && remoteDecoderHandle == 0L) {
+            remoteDecoderHandle = nativeLib.createRemoteShuffleDecoder(expectedSchema.get)
+          }
+          val decoded = nativeUtil.getNextBatch(
+            fieldCount,
+            (arrayAddrs, schemaAddrs) => {
+              try {
+                if (validateRemoteFrames) {
+                  nativeLib.decodeShuffleBlockWithValidation(
+                    dataBuf,
+                    bytesToRead,
+                    arrayAddrs,
+                    schemaAddrs,
+                    tracingEnabled,
+                    remoteDecoderHandle)
+                } else {
+                  nativeLib.decodeShuffleBlock(
+                    dataBuf,
+                    bytesToRead,
+                    arrayAddrs,
+                    schemaAddrs,
+                    tracingEnabled)
+                }
+              } catch {
+                case NonFatal(failure) =>
+                  // Record only native decode failures; creation, allocation and import failures
+                  // must not be reported as corrupt remote data. Let NativeUtil clean up first.
+                  nativeFailure = failure
+                  throw failure
+              }
+            })
+          decodeTime.add(System.nanoTime() - startTime)
+          decoded
+        }
+        if (batch.isEmpty) {
+          close()
+        }
+        batch.isDefined
+      }
+    } catch {
+      case NonFatal(failure) if failure eq nativeFailure =>
+        // Reporting can make an RPC. It must run after Arrow cleanup and outside the monitor so
+        // task completion can release the decoder while that reporting is in progress.
         in match {
           case handler: CometShuffleReadFailureHandler => handler.onShuffleReadFailure(failure)
           case _ =>
@@ -145,10 +170,6 @@ case class NativeBatchDecoderIterator(
   }
 
   private def readNextBlock(): Option[(Int, ByteBuffer, Int)] = {
-    if (channel == null || isClosed) {
-      return None
-    }
-
     // read compressed batch size from header
     longBuf.clear()
     while (longBuf.hasRemaining && channel.read(longBuf) >= 0) {}
@@ -212,6 +233,8 @@ case class NativeBatchDecoderIterator(
         currentBatch = null
         val prefetched = batch
         batch = None
+        val decoderHandle = remoteDecoderHandle
+        remoteDecoderHandle = 0L
 
         var failure: Throwable = null
         def release(resource: => Unit): Unit = {
@@ -225,6 +248,7 @@ case class NativeBatchDecoderIterator(
 
         if (previous != null) release(previous.close())
         prefetched.filterNot(_ eq previous).foreach(pending => release(pending.close()))
+        if (decoderHandle != 0L) release(nativeLib.releaseRemoteShuffleDecoder(decoderHandle))
         if (in != null) release(in.close())
         release(resetDataBuf())
         if (failure != null) throw failure
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -1253,9 +1253,55 @@ pub unsafe extern "system" fn Java_org_apache_comet_Native_decodeShuffleBlock(
 }
 
 #[no_mangle]
+/// Parse the expected schema once for a remote shuffle iterator.
+///
+/// The iterator owns the returned decoder and releases it when the input is closed.
+pub extern "system" fn Java_org_apache_comet_Native_createRemoteShuffleDecoder(
+    e: EnvUnowned,
+    _class: JClass,
+    expected_schema: JByteArray,
+) -> jlong {
+    try_unwrap_or_throw(&e, |env| {
+        let bytes = env.convert_byte_array(expected_schema)?;
+        let schema = ShuffleScan::decode(bytes.as_slice()).map_err(|error| {
+            CometError::Internal(format!("Invalid expected remote shuffle schema: {error}"))
+        })?;
+        let decoder = RemoteShuffleDecoder {
+            expected_types: schema.fields.iter().map(to_arrow_datatype).collect(),
+        };
+        Ok(Box::into_raw(Box::new(decoder)) as jlong)
+    })
+}
+
+/// Immutable decoding state owned by one JVM remote shuffle iterator, not shared across tasks.
+struct RemoteShuffleDecoder {
+    expected_types: Vec<ArrowDataType>,
+}
+
+#[no_mangle]
+/// Release a remote shuffle iterator's decoder.
+///
+/// # Safety
+/// A nonzero handle must have been returned by `createRemoteShuffleDecoder`, must not have
+/// been released, and must not be in use by a concurrent decode call.
+pub unsafe extern "system" fn Java_org_apache_comet_Native_releaseRemoteShuffleDecoder(
+    e: EnvUnowned,
+    _class: JClass,
+    decoder_handle: jlong,
+) {
+    try_unwrap_or_throw(&e, |_| {
+        if decoder_handle != 0 {
+            drop(unsafe { Box::from_raw(decoder_handle as *mut RemoteShuffleDecoder) });
+        }
+        Ok(())
+    })
+}
+
+#[no_mangle]
 /// Decode a remote native shuffle block with Arrow array and logical type validation enabled.
 /// # Safety
-/// This function is inherently unsafe since it deals with raw pointers passed from JNI.
+/// Buffer and output pointers must be valid. The decoder handle must have been returned by
+/// `createRemoteShuffleDecoder` and must remain alive for the duration of this call.
 pub unsafe extern "system" fn Java_org_apache_comet_Native_decodeShuffleBlockWithValidation(
     e: EnvUnowned,
     _class: JClass,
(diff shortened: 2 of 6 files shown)


sunchao:
## Review result

**No actionable defects found** in [PR #5722](https://github.com/apache/datafusion-comet/pull/5722) at `cedc3bd99`, against base `75fdddc92`.

### Correctness

The cache preserves the existing Spark type conversion and per-block validation. It retains expected types, so later blocks still undergo logical-type and Arrow validation.

I traced normal exhaustion, early task completion, initialization failure, decode failure, and cleanup exceptions. Ownership is cleared before release, remaining cleanup proceeds after failures, and the production callers decode synchronously. I found no reachable release race or introduced leak. [Iterator implementation](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/NativeBatchDecoderIterator.scala#L104)

### Performance

For an iterator containing **N blocks**, schema copying, protobuf parsing, and type construction decrease from **N times to once**. The cost is two additional lifecycle JNI calls and schema metadata retained until close. Single-block iterators cannot amortize that overhead.

The expected benefit is strongest with many small blocks and wide or nested schemas. **The actual speedup remains unmeasured:** the PR’s benchmark explicitly excludes JNI and schema initialization. It measures remote decoding cost, not this cache’s benefit or whole-query performance. A before/after benchmark including those costs would establish the net gain.

### Design

One immutable native object owned by one iterator is an appropriate lifetime. Lazy creation avoids allocation for empty or unused streams. Keeping initialization failures outside remote-corruption reporting also preserves the right failure classification. [Native implementation](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/native/core/src/execution/jni_api.rs#L1255)

### Abstraction & complexity

The wrapper is small and serves a clear ownership purpose. I see no unnecessary framework or abstraction, and no evidence that a global cache or combined initialization/decode API would justify additional complexity.

### Validation

- **64/64 reader tests passed locally** on Spark 4.1, including reuse, later-frame validation, and cleanup cases.
- JVM reactor build, formatting checks, and Rust formatting passed.
- Local native compilation was blocked by the dependency mirror. Tests used a checksum-verified CI library whose source tree matches this commit.
- Latest CI: **59 passed, 6 pending, 7 skipped; no failures**. [CI run](https://github.com/apache/datafusion-comet/actions/runs/33989024299)

I would not request code changes on this revision.

sunchao (approved):
Approved the reviewed commit cedc3bd99ec4580226ffe77154a18fc7058a1547. The full review is posted at https://github.com/apache/datafusion-comet/pull/5722#issuecomment-5554808754.

CI refreshed before publication: 60 passed, 5 pending, 7 skipped, with no failures.

sunchao (commented):
## Correctness

No additional verified P1/P2 finding from this follow-up at `cedc3bd99ec4580226ffe77154a18fc7058a1547`, against `75fdddc9285ec61c0cd326977c61dd41fca39a8b`. The [existing review and approval](https://github.com/apache/datafusion-comet/pull/5722#pullrequestreview-5123046452) already cover this revision. The maintained Spark 3.5 and 4.0 task-completion and interruption paths support the reviewed iterator cleanup behavior.

### Updated validation

The exact-head CI refresh at **2026-09-05 21:46 UTC** reports **65 successful / 9 skipped** checks, with none pending or failing. I verified that the [Spark 3.5 shuffle job](https://github.com/apache/datafusion-comet/actions/runs/33989024299/job/101368681791) and [Spark 4.0 shuffle job](https://github.com/apache/datafusion-comet/actions/runs/33989024299/job/101368681824) each passed all **64 reader-suite tests**, including the new reuse, later-frame mismatch, and lifecycle cases. Their complete shards passed 426 and 469 tests respectively.

Both jobs checked out merge `da40304adc8455b91ce0803f66c27fcd6c592751`. Its parents are this base and head, and its entire source tree matches the reviewed head. The native library build succeeded on that merge using the CI profile. These reader tests execute Spark and JNI with a recording Celeborn client and in-memory frames. They do not exercise a deployed Celeborn service. I ran no local build or benchmark. Maintained Spark 3.4 and 4.1 sources remain unavailable for independent source qualification.

## Performance

No additional performance finding. The existing review already identifies the missing measurement of the cache's net benefit. The supplied remote/local decode timings exclude JNI and schema initialization, so this follow-up adds no cache-speedup or whole-query performance claim.

## Design

No new design concern after checking the production callers, task-completion order, and initialization-failure classification. There has been no code revision since the existing approval that needs a separate design summary.

## Abstraction & complexity

No additional abstraction or complexity finding in this unchanged revision. No repeated inline feedback or duplicate approval is proposed.

spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/NativeBatchDecoderIterator.scala
```diff
@@ -225,6 +233,7 @@ case class NativeBatchDecoderIterator(
 
         if (previous != null) release(previous.close())
         prefetched.filterNot(_ eq previous).foreach(pending => release(pending.close()))
+        if (decoderHandle != 0L) release(nativeLib.releaseRemoteShuffleDecoder(decoderHandle))
```
jayanth86, line 251: [P1] Keep the decoder alive while a read is in flight

Could creation and decoding be coordinated with this release? On supported Spark 3.4.3, a Python UDF downstream of a Celeborn exchange consumes this iterator on PythonRunner's writer thread. Its completion listener [interrupts and joins that writer](https://github.com/apache/spark/blob/1eb558c3a6fbdd59e5a305bc3ab12ce748f6511f/core/src/main/scala/org/apache/spark/api/python/PythonRunner.scala#L251-L260), but task cancellation with `interruptThread=true` can interrupt the join. [TaskContextImpl catches the listener exception and continues cleanup](https://github.com/apache/spark/blob/1eb558c3a6fbdd59e5a305bc3ab12ce748f6511f/core/src/main/scala/org/apache/spark/TaskContextImpl.scala#L194-L250), so Comet can reach this release while the writer is still in native decompression.

`fetchNext()` does not take the `close()` lock. This release therefore frees the boxed decoder while [the active JNI call still borrows its `expected_types`](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/native/core/src/execution/jni_api.rs#L1317-L1327), which are read after decompression in `decode_remote_shuffle_batch`. That introduces a use-after-free capable of crashing the executor. Previously, the types were owned locally by the JNI invocation and could not be freed by task cleanup.

Please protect the handle across creation, active decoding, and release, and add coverage for interrupted cleanup overlapping a read. A latch-based test using this exact iterator and mocked JNI reproduces release while decoding is still in flight. This is source-verified reachability plus a lifecycle reproduction, not a reproduced native crash.

jayanth86 (commented):
Reviewed `cedc3bd99ec4580226ffe77154a18fc7058a1547` against `75fdddc9285ec61c0cd326977c61dd41fca39a8b`. I found one additional native-lifetime issue in the supported Spark 3.4.3 Python-consumer cancellation path, detailed inline. The per-frame validation and per-iterator schema scope otherwise remain intact.

Validation included independent source review and three local latch-based probes of the unmodified iterator with mocked JNI. The probes reproduce in-flight release, late handle publication after close, and release following an interrupted join. They do not execute Spark or native Rust.

I also verified that [CI](https://github.com/apache/datafusion-comet/actions/runs/33989024299) built the same source tree as this head: all 64 reader-suite tests passed on Spark 3.5, 4.0, and 4.1, and the Rust job passed 1,163 tests with 4 skipped. Those tests do not cover this cancellation interleaving. A local native build was not run because the required dependencies were not cached.

sunchao (commented):
## Correctness

**[P1] Protect the cached decoder's lifetime during cancellation.**

This follow-up at `cedc3bd99ec4580226ffe77154a18fc7058a1547`, against `75fdddc9285ec61c0cd326977c61dd41fca39a8b`, supports the ownership concern in the [existing P1 thread](https://github.com/apache/datafusion-comet/pull/5722#discussion_r3942256302). My [earlier cleanup assessment](https://github.com/apache/datafusion-comet/pull/5722#pullrequestreview-5123146435) was too broad: synchronous decoding within the consumer does not establish exclusive access throughout task cleanup.

The [iterator](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/NativeBatchDecoderIterator.scala) does not coordinate decoder creation and use with closure. The [new native ownership contract](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/native/core/src/execution/jni_api.rs#L1276-L1304) therefore depends on caller guarantees that are not sufficient under cancellation. Clearing the stored handle before release makes repeated closure idempotent; it does not establish that active use has finished or that creation cannot publish state after closure. Before this change, the expected types belonged to the individual JNI invocation.

Independent inspection of the maintained Spark 3.5 branch also found a threaded Python input consumer and an interruptible cleanup wait. That requires the lifetime protection discussed in the existing thread. The maintained 4.0 branch instead consumes Python input on the task thread; I am not extending that specific threaded-consumer conclusion to 4.0. Maintained 3.4 and 4.1 sources were unavailable, so I have not independently qualified those versions or reproduced the external reviewer's probes.

The final GitHub CI check at **2026-09-06 00:01:47 UTC** has **65 successful / 9 skipped** checks, with no pending or failing check. Source inspection shows sequential decoder lifecycle coverage and separate concurrent raw-stream-opening tests; neither establishes the new decoder's lifetime during cancellation. This follow-up ran no local Spark, JNI, native build, or concurrency reproduction. No duplicate inline is proposed.

## Performance

No additional performance finding from this lifecycle review. A lifetime fix should preserve cancellation responsiveness and avoid unnecessary contention during normal reading. Those effects need validation once the fix exists; this follow-up adds no timing claim.

## Design

The iterator needs one explicit ownership rule covering initialization, publication, active native use, and final release. That rule must remain valid when completion callbacks fail or cleanup is interrupted. The native safety preconditions should be enforced by the owning layer, rather than relying solely on normal callback order. Input shutdown must still be able to unblock transport reads.

## Abstraction & complexity

Keep the lifetime mechanism local to the iterator and make its relationship with batch and `NativeUtil` cleanup clear. Merely adding visibility to the handle or another close-only guard would leave the ownership gap. Focused regression coverage should verify the lifetime invariants across cancellation and cleanup failures, while retaining the existing allocation and idempotence checks.

spark/src/main/scala/org/apache/spark/sql/comet/execution/shuffle/NativeBatchDecoderIterator.scala
andygrove: The P1 lifetime concern looks right to me. I reproduced the JVM half of it locally with a latch-based probe on this iterator and a mocked `Native`: `releaseRemoteShuffleDecoder` runs to completion while `decodeShuffleBlockWithValidation` is still inside the JNI call, so nothing in the iterator serializes the release against an active decode. `remoteDecoderHandle` is published from `fetchNext()` without the monitor that `close()` takes, and `decode_remote_shuffle_batch` reads `expected_types` after `read_ipc_compressed_validated` has already decompressed the block, so the borrow outlives the free in that ordering. The mirror-image ordering is a leak rather than a crash, since `close()` reads `0L` and the box is never freed.

What I do not think is true is that this PR introduces the hazard. `CometExecIterator` has had the same shape for a long time, with a wider window on a much hotter path. It registers `close()` as a task completion listener at construction ([`CometExecIterator.scala:189`](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/spark/src/main/scala/org/apache/comet/CometExecIterator.scala#L189)), `close()` is `synchronized` and calls `nativeLib.releasePlan(plan)` ([`:275`](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/spark/src/main/scala/org/apache/comet/CometExecIterator.scala#L275) and [`:305`](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/spark/src/main/scala/org/apache/comet/CometExecIterator.scala#L305)), and `getNextBatch` calls `nativeLib.executePlan(..., plan, ...)` outside that monitor ([`:205`](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/spark/src/main/scala/org/apache/comet/CometExecIterator.scala#L205)). On the native side `releasePlan` is a `Box::from_raw` drop of the context ([`jni_api.rs:1051`](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/native/core/src/execution/jni_api.rs#L1051)) while `executePlan` holds `get_execution_context(exec_context)`, an unbounded `&'a mut ExecutionContext` derived from the raw handle, for the duration of an entire plan execution ([`jni_api.rs:848`](https://github.com/apache/datafusion-comet/blob/cedc3bd99ec4580226ffe77154a18fc7058a1547/native/core/src/execution/jni_api.rs#L848)). So if the threaded Python consumer plus an interrupted cleanup wait reaches the shuffle decoder, it already reaches every Comet native plan today.

Given that, I would rather not hold this PR for it. Could we file one issue covering JVM-owned native handle lifetime for both `NativeBatchDecoderIterator` and `CometExecIterator`, and fix them the same way? Fixing only the new handle here would leave the wider window untouched and leave us with two different ownership conventions for the same problem.

For what it is worth, the minimal fix for this iterator looks like taking the monitor around just the `nativeUtil.getNextBatch` call and leaving `readNextBlock()` outside it, so `in.close()` can still unblock a transport read. Decoding one block is bounded work, so cleanup cannot hang waiting on the monitor. Whatever we settle on should apply to `releasePlan` as well.

pingzh: Codex acting for the user: Fixed the `NativeBatchDecoderIterator` lifetime race in [92fa68c5b](https://github.com/apache/datafusion-comet/commit/92fa68c5b62894e758406d0d0ae91ef4a16f4d75).

Decoder creation, native decoding, Arrow import, and batch publication now share the monitor used by `close()`. `next()` also transfers batch ownership under that monitor. A transport read that finishes after cleanup rechecks `isClosed` before creating or using a decoder.

`readNextBlock()` stays outside the monitor so `in.close()` can unblock it. Native decode failures unwind Arrow cleanup before reporting outside the monitor, since reporting may perform an RPC.

Added five latch-based regressions covering creation, active decoding (including interrupted cleanup), import/publication, blocked transport reads, and failure reporting. All five failed before the fix; all 69 reader-suite tests now pass on both Spark 4.1 and Spark 3.4.3. Spotless and Scalastyle also pass.

The broader `CometExecIterator` lifetime concern remains a separate follow-up.

sunchao (approved):
## Correctness

The lifetime fix at `92fa68c5b62894e758406d0d0ae91ef4a16f4d75`, against `bc74cc79fcf5eeed9610d9a20937375652cb74df`, addresses the [existing P1](https://github.com/apache/datafusion-comet/pull/5722#discussion_r3942256302). I found no new or remaining verified P1/P2 in this update. Decoder creation, native use, Arrow import and batch publication now share the cleanup monitor; a completed transport read rechecks closure before acquiring native state. This enforces the cached decoder's lifetime independently of whether Spark 3.5's Python-writer cleanup wait completes normally. The native output owns its arrays and schema metadata, so it retains no borrow of the cached expected types after decoding returns.

Native decode failures unwind `NativeUtil` cleanup before reporting outside the monitor. Initialization, allocation, import and stream-cleanup failures retain their separate classification. The production reader closes `NativeUtil` after decoder cleanup, which now waits for an active import. These conclusions concern this iterator's ownership protocol; they do not establish safe arbitrary use of a delivered batch after task completion.

The **2026-09-08 01:17:35 UTC** snapshot has **65 successful / 9 skipped** checks. I verified **69 reader-suite tests passed in each** of the [Spark 3.5](https://github.com/apache/datafusion-comet/actions/runs/34055099583/job/101549058261) and [Spark 4.0](https://github.com/apache/datafusion-comet/actions/runs/34055099583/job/101549058304) jobs, including the five new concurrency cases. The full shards passed 438 and 487 tests respectively; the 3.5 shard also canceled six tests outside this suite. The new concurrency cases use mocked native calls. They verify JVM coordination, not an actual native crash or an end-to-end Python cancellation scenario.

Those jobs and the [native build](https://github.com/apache/datafusion-comet/actions/runs/34055099583/job/101545828442) used merge `24323fd3f6c6f29b15ee18b64bf06a73f1377754`, whose parents are the assigned base and head. All six PR files and the inspected lifecycle dependencies match the head; the complete trees differ elsewhere. Both test jobs verified the same native artifact digest. This credits the tested merge, not a standalone head build. I ran no local build or runtime reproduction. Maintained Spark 3.4 and 4.1 source qualification remains unavailable.

## Performance

The additional synchronization is local to one iterator. Transport reads and failure-reporting RPCs remain outside the monitor; cleanup waits for active decoding and import. This avoids introducing a shared cache or contention across tasks. Neither these tests nor the existing remote/local decode measurements quantify the synchronization cost or the cache's net speedup, so this follow-up adds no performance claim.

## Design

The update covers initialization and ownership transfer as well as the native call itself. That is the necessary scope for coordinating cleanup, and the failure-reporting boundary keeps temporary Arrow resources out of a potentially slow RPC. The [broader `CometExecIterator` discussion](https://github.com/apache/datafusion-comet/pull/5722#discussion_r3944263483) remains separate: that implementation is unchanged from the merge base and this fix does not establish its lifetime guarantees.

## Abstraction & complexity

One existing monitor and an explicitly documented single-consumer contract keep the production mechanism small. The five focused concurrency cases cover its main ownership boundaries while the existing lifecycle tests retain cleanup and failure-classification coverage. I see no additional abstraction or simplification needed for this update, and propose no duplicate inline finding.

sunchao:
Merged, thanks @pingzh !

PR #788, opened 2024-08-06
chore: Upgrade to DataFusion rev bddb641 and disable "skip partial aggregates" feature

andygrove:
## Which issue does this PR close?

N/A

## Rationale for this change

Upgrade to latest DF

## What changes are included in this PR?

- Upgrade to latest DF
- Disable new "skip partial aggregates" feature that was causing TPC-DS test failures
- Other changes needed due to DF upgrade

PR #5723, opened 2026-09-05
feat: Enable adaptive partial aggregation for eligible native shuffle plans

### LINKED ISSUE #5717: Re-enable adaptive skip-partial aggregation for supported high-cardinality aggregates

Comet disables DataFusion's adaptive skip-partial aggregation by setting the probe ratio threshold to `1.1`. High-cardinality partial aggregation can spend substantial CPU building a hash table while emitting almost as many rows as it consumes.

Release-mode Spark A/B measurements on Comet `75fdddc9285ec61c0cd326977c61dd41fca39a8b` (DataFusion 55.0.0, merged in #5262) confirm the opportunity:

| Fixture | Disabled (`1.1`), median wall | Enabled (`0.8`), median wall | Result |
| --- | ---: | ---: | --- |
| High-cardinality COUNT | 2,060 ms | 1,380 ms | 33.0% faster |
| High-cardinality COUNT, fresh JVM/reversed initial order | 1,501 ms | 1,027 ms | 31.6% faster |
| Low-cardinality control | 292 ms | 267 ms | Mixed pairs/warmup effects; not a claimed gain |
| Low-cardinality control, fresh JVM/reversed initial order | 1,370 ms | 1,360 ms | Effectively neutral under contention |

Each JVM used 16,777,216 Parquet rows, 8 writers, 32 reducers, `local[8]`, a 10 GiB heap, Spark 4.1.3, AQE/exchange reuse disabled, native direct-read shuffle, one warmup per mode, and seven alternating measured pairs. The high-cardinality query groups by a unique integer key, counts each group, then sums the counts; the control groups into 1,024 keys. All results were checked exactly. Skipping won all 14 high-cardinality pairs. Map-stage medians were 1,248 -> 569 ms and 917 -> 413 ms. These are workstation measurements with concurrent desktop activity, not a controlled comparison between DataFusion versions or a universal no-regression guarantee.

The experiment used `spark.comet.exec.respectDataFusionConfigs=true`, toggled `spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold`, and held `spark.comet.datafusion.execution.coalesce_batches=false` in both modes. The older DataFusion 54.1.0 fixture at `81d637b9bf40a5be6f4f0c65ad6f497b34746e69` measured 2,125 -> 1,560 ms (26.6%); the table above is the new 55.0.0 evidence.

Implement a guarded reintroduction for explicitly supported accumulators and actual Spark aggregate modes, **not a global threshold reduction**.

1. **Preserve grouping-only DISTINCT semantics.** Distribution-required grouping-only stages must fully deduplicate their keys. Serialize these empty-aggregate stages as native `Final`, rather than leaving the protobuf default `Partial`. On unmodified 55.0.0 main, enabling skipping made both `SELECT count(*) FROM (SELECT DISTINCT k ...)` and `SELECT count(DISTINCT k) ...` return 2,400,000 instead of 1,980,000. The local mode fix restores both results at 2.4 million input rows, and its focused ScalaTest passes (1 test covering both shapes).
2. **Use explicit accumulator/mode eligibility.** Start narrowly with grouping-only/COUNT cases. Exclude Spark `PartialMerge` and mixed-mode stages even when represented as native `Partial`; their inputs are already aggregate states. Unsupported accumulators must retain ordinary aggregation, including when users opt into skipping. Grouped integer SUM and mixed DISTINCT/COUNT both reproduce `Input batch conversion to state not implemented` under unrestricted skipping on 55.0.0, even after the DISTINCT fix. Do not admit SUM, custom accumulators, or variable-width aggregate states without independent conversion, memory, and spill validation.
3. **Keep the supported path covered by runtime regressions.** After the mode fix, nine supported SQL cases (COUNT, nullable/filtered COUNT, both DISTINCT shapes, empty grouped/global input, null grouping keys, all-null COUNT values) match Spark with skipping off/on. Together with the two unsupported cases' disabled controls this gives 20 matching results; the two enabled unsupported cases still fail as expected until eligibility is implemented. Add regressions that require those excluded cases to remain correct without skipping, and prove eligible cases actually skip.
4. **Validate bounded memory and controls.** The same 16,777,216-row fixture with `spark.comet.memoryOverhead=128m` passed all eight executions (two warmups plus three measured pairs), with 64 aggregate spills and approximately 273 MB spilled per query in both modes. Default-memory runs had zero spills. This validates one spilling workload, not all memory pools, AQE, distributed deployment, or variable-width states. Preserve low-cardinality controls and rerun the guarded implementation.

### Guarded implementation validation

A local implementation on the same DataFusion 55 main base now combines the DISTINCT fix with a conservative fused-plan eligibility gate. It admits only native shuffle-writer plans whose partial aggregates are grouping-only or single-argument COUNT. Any unsupported partial accumulator or PartialMerge/mixed mode disables skipping throughout that fused plan. Non-native-shuffle roots remain disabled, and testing overrides cannot bypass eligibility. A new `rows bypassing partial aggregation` SQL metric verifies actual skipping.

- Native eligibility regression: 1 passed. Full Spark aggregate suite: 93 passed, zero failures, 2 ignored. The two new focused tests also passed independently.
- The 2.4-million-row correctness harness now has **22 exact matches against Spark and zero conversion errors**, including both previously excluded error cases, which retain ordinary aggregation.
- Guarded high-cardinality A/B, seven alternating pairs: **664 -> 479 ms median wall time (27.9% faster)**; enabled won all seven pairs. Enabled queries bypassed 15,925,248 rows each; disabled queries bypassed zero.
- Low-cardinality control, seven pairs: 199 -> 185 ms; both modes bypassed zero rows. No optimization gain is claimed for this control.
- The 128 MiB native-memory run passed all eight executions (two warmups plus three pairs), with 64 aggregate spills and approximately 273 MB spilled per query. Enabled queries still bypassed 15,925,248 rows, so skipping and spilling were exercised together. Median wall times were 945 -> 858 ms, a limited performance sample.

These use the same 16,777,216-row fixture and settings described above. Compilation and test processes had finished before timing. Compare modes within each run, not absolute times across the earlier and guarded runs. The implementation is local; these results do not imply a published or merged PR, nor coverage of every memory pool, AQE, or distributed deployment.

### Additional context

DataFusion 55.0.0 is already integrated, but it was **not a prerequisite for COUNT skipping**: 54.1.0 already supported that path. The earlier expectation of a new capability guard was incorrect. DataFusion 55 removes `GroupsAccumulator::supports_convert_to_state()` and makes `convert_to_state()` required. Comet #5262 supplies error-returning placeholders for several custom accumulators, including integer SUM and `MergeAsPartialGroupsAccumulator`; required trait methods do not imply working conversion support.

Skip-partial was disabled in #788 after correctness failures; apache/datafusion#11850 records the historical failure. The proposal is a guarded implementation with explicit correctness boundaries, not unrestricted enablement.

peterxcli:

Closes #5717.

## Rationale for this change

High-cardinality partial aggregation can spend substantial CPU maintaining a hash table while emitting almost as many rows as it consumes. Re-enable DataFusion's adaptive bypass only for supported native-shuffle plans, rather than lowering the threshold globally.

DataFusion 55 requires `convert_to_state`, but several Comet accumulators still return an unsupported-operation error. Spark's grouping-only DISTINCT stages also need an explicit mode to preserve full deduplication.

- Serialize distribution-required grouping-only aggregates as native `Final` to preserve DISTINCT semantics.
- Admit adaptive skipping only for native shuffle-writer plans whose partial aggregates are grouping-only or single-argument COUNT. An unsupported partial accumulator, PartialMerge, or mixed mode disables skipping throughout the fused plan. Non-native-shuffle roots remain disabled.
- Apply the eligibility guard after DataFusion configuration pass-through, so testing overrides cannot bypass safety. Eligible plans retain the adaptive default and explicit disablement remains available.
- Expose `skipped_aggregation_rows` as a SQL metric and document the conservative scope.

No new configuration, dependency, or execution wrapper is introduced.

- Full `CometAggregateSuite`: 93 passed, zero failures, 2 ignored. The two new focused regressions also passed independently.
- Native `skip_partial_eligibility_is_fail_closed`: 1 passed.
- Native release build, Rust formatting check, Scala Spotless formatting, and diff whitespace check passed.
- Separate 2.4-million-row correctness fixture: 22 exact comparisons against Spark, zero errors, covering COUNT, nullable/filtered COUNT, DISTINCT, empty/null inputs, SUM, and mixed PartialMerge.
- The checked-in Spark regression asserts actual skipping for eligible high-cardinality COUNT and no skipping for low cardinality, explicit disablement, unsupported accumulators/modes, and JVM shuffle.

Release-mode Spark A/B on the DataFusion 55 base (`75fdddc9285ec61c0cd326977c61dd41fca39a8b`): 16,777,216 Parquet rows, 8 writers, 32 reducers, local[8], Spark 4.1.3, native direct-read shuffle, AQE/exchange reuse disabled, one warmup per mode, alternating pairs. Build/test processes had finished before timing.

| Workload | Pairs | Disabled median | Guarded enabled median | Speedup (wall-time reduction) |
| --- | ---: | ---: | ---: | --- |
| High-cardinality COUNT | 7 | 664 ms | 479 ms | 27.9% |
| Low-cardinality control | 7 | 199 ms | 185 ms | 7.0% observed; no skipping, no optimization gain claimed |
| High-cardinality COUNT, 128 MiB native memory | 3 | 945 ms | 858 ms | 9.2% observed; limited sample |

Wall-time reduction is `(disabled median - enabled median) / disabled median × 100%`.

High-cardinality COUNT improved 27.9%, winning all seven pairs; enabled runs bypassed 15,925,248 rows each, versus zero when disabled. The low-cardinality control bypassed zero rows in both modes, so no optimization gain is claimed there.

All eight constrained-memory executions (including warmups) returned correct results with 64 aggregate spills and approximately 273 MB spilled each. Enabled executions still bypassed 15,925,248 rows, exercising skipping and spilling together. These are workstation measurements, not a universal speedup or coverage of every memory pool, AQE, or distributed deployment. Additional benchmark context is recorded in #5717.

Focused reproduction from the repository root:

```sh
./mvnw test -Dtest=none '-Dsuites=org.apache.comet.exec.CometAggregateSuite skip partial aggregation' -Dscalastyle.skip=true -Djni.dir="$PWD/native/target/release"
```

Native regression from `native/` (on macOS, expose the installed JVM's `lib/server` via `DYLD_LIBRARY_PATH`):

```sh
RUSTFLAGS='-Ctarget-cpu=native' cargo test --release --locked -p datafusion-comet skip_partial_eligibility --lib
```

### DIFF
--- a/docs/source/user-guide/latest/tuning.md
+++ b/docs/source/user-guide/latest/tuning.md
@@ -31,6 +31,20 @@ It is recommended that `COMET_WORKER_THREADS` be set to the number of executor c
 in some environments, such as Kubernetes, where the number of cores allocated to a pod will already be equal to the
 number of executor cores.
 
+## Adaptive Partial Aggregation
+
+For high-cardinality grouping, Comet can bypass partial hash aggregation when it is not
+reducing the number of rows enough. This currently applies only to fused native shuffle-writer
+plans whose partial aggregates are grouping-only or single-argument `COUNT`. Low-cardinality
+inputs continue to aggregate normally. The SQL metric `rows bypassing partial aggregation`
+shows whether skipping occurred.
+
+Eligibility is conservative for the whole fused native plan: any unsupported partial accumulator,
+Spark `PartialMerge`, or mixed-mode aggregate disables skipping in that plan. Multi-argument
+`COUNT` and other accumulators are not admitted. Distribution-required grouping-only stages
+still fully deduplicate, and non-native-shuffle plans retain ordinary aggregation.
+The DataFusion testing configuration override does not bypass these safety checks.
+
 ## Memory Tuning
 
 It is necessary to specify how much memory Comet can use in addition to memory already allocated to Spark. In some
--- a/docs/source/user-guide/latest/tuning.md
+++ b/docs/source/user-guide/latest/tuning.md
@@ -45,6 +45,32 @@ Spark `PartialMerge`, or mixed-mode aggregate disables skipping in that plan. Mu
 still fully deduplicate, and non-native-shuffle plans retain ordinary aggregation.
 The DataFusion testing configuration override does not bypass these safety checks.
 
+DataFusion 55 defaults to probing after 100,000 input rows per partial aggregation
+partition and skipping when the number of groups divided by input rows exceeds `0.8`.
+To experiment with these thresholds, enable `spark.comet.exec.respectDataFusionConfigs`,
+a development and testing option that defaults to `false`. For example, the following
+SQL settings pass through the default threshold values, which you can adjust:
+
+```sql
+SET spark.comet.exec.respectDataFusionConfigs=true;
+SET spark.comet.datafusion.execution.skip_partial_aggregation_probe_rows_threshold=100000;
+SET spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold=0.8;
+```
+
+A lower row threshold allows an earlier decision; a lower ratio threshold makes
+skipping more likely. Skipping can increase the number of partial states emitted
+and the amount of shuffle data, so measure the effect on your workload.
+
+To disable skipping, keep `spark.comet.exec.respectDataFusionConfigs=true` and set
+the ratio threshold above the maximum possible groups/input-rows ratio:
+
+```sql
+SET spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold=1.1;
+```
+
+These settings only tune eligible plans. Unsupported accumulators and modes remain
+disabled even when configuration overrides are enabled.
+
 ## Memory Tuning
 
 It is necessary to specify how much memory Comet can use in addition to memory already allocated to Spark. In some
--- a/spark/src/main/scala/org/apache/spark/sql/comet/operators.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/comet/operators.scala
@@ -1735,6 +1735,15 @@ trait CometBaseAggregate {
     if (aggregateExpressions.isEmpty) {
       val hashAggBuilder = OperatorOuterClass.HashAggregate.newBuilder()
       hashAggBuilder.addAllGroupingExprs(groupingExprs.map(_.get).asJava)
+      // Spark has no expression mode to serialize here. An empty aggregate with a required child
+      // distribution must fully deduplicate its keys (Final, or a pre-distinct PartialMerge), so
+      // use native Final to keep skip-partial disabled.
+      val mode = if (aggregate.requiredChildDistributionExpressions.isDefined) {
+        CometAggregateMode.Final
+      } else {
+        CometAggregateMode.Partial
+      }
+      hashAggBuilder.setModeValue(mode.getNumber)
       buildAggOp(
         builder,
         hashAggBuilder,
--- a/spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala
@@ -1650,7 +1650,7 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
             "spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold" -> "0.8") {
             for (expression <- Seq("sum(v)", "count(v, w)", "count(*) + sum(v)")) {
               val result = aggregates(
-                s"SELECT sum(n) FROM " +
+                "SELECT sum(n) FROM " +
                   s"(SELECT k, $expression n FROM skip_partial_eligibility GROUP BY k)")
               assert(result.map(skipped).sum == 0L, s"Unexpected skipping for $expression")
             }
--- a/spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala
+++ b/spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala
@@ -28,7 +28,7 @@ import org.apache.spark.{CometListenerBusUtils, SparkConf}
 import org.apache.spark.scheduler.{SparkListener, SparkListenerTaskEnd}
 import org.apache.spark.sql.{CometTestBase, DataFrame, Row}
 import org.apache.spark.sql.catalyst.expressions.Cast
-import org.apache.spark.sql.catalyst.expressions.aggregate.{Final, Partial}
+import org.apache.spark.sql.catalyst.expressions.aggregate.{Final, Partial, PartialMerge}
 import org.apache.spark.sql.catalyst.optimizer.EliminateSorts
 import org.apache.spark.sql.catalyst.plans.physical.RangePartitioning
 import org.apache.spark.sql.comet.CometHashAggregateExec
@@ -1559,6 +1559,108 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
                   "SELECT COUNT(col2), MIN(col2), COUNT(DISTINCT col2), SUM(col2)," +
                     s" SUM(DISTINCT col2), COUNT(DISTINCT col2), col1 FROM $table group by col1")
               }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  test("skip partial aggregation preserves post-shuffle distinct") {
+    val writers = 8
+    val rowsPerWriter = 300L
+    val overlap = 60L
+    val rows = writers * rowsPerWriter
+
+    withTempDir { dir =>
+      val path = new Path(dir.toURI.toString, "input")
+      withSQLConf(CometConf.COMET_ENABLED.key -> "false") {
+        spark
+          .range(0L, rows, 1L, writers)
+          .selectExpr(s"id - spark_partition_id() * $overlap AS k")
+          .write
+          .parquet(path.toUri.toString)
+      }
+
+      withParquetTable(path.toUri.toString, "skip_partial_distinct") {
+        withSQLConf(
+          SQLConf.ADAPTIVE_EXECUTION_ENABLED.key -> "false",
+          SQLConf.SHUFFLE_PARTITIONS.key -> writers.toString,
+          CometConf.COMET_SHUFFLE_ENABLED.key -> "true",
+          CometConf.COMET_SHUFFLE_MODE.key -> "native",
+          CometConf.COMET_RESPECT_DATAFUSION_CONFIGS.key -> "true",
+          "spark.comet.datafusion.execution.skip_partial_aggregation_probe_rows_threshold" -> "100",
+          "spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold" -> "0.8") {
+          checkSparkAnswerAndOperator(
+            "SELECT count(*) FROM (SELECT DISTINCT k FROM skip_partial_distinct)")
+          checkSparkAnswerAndOperator("SELECT count(DISTINCT k) FROM skip_partial_distinct")
+        }
+      }
+    }
+  }
+
+  test("skip partial aggregation admits only supported native shuffle plans") {
+    withTempDir { dir =>
+      val path = new Path(dir.toURI.toString, "input").toUri.toString
+      withSQLConf(CometConf.COMET_ENABLED.key -> "false") {
+        spark
+          .range(0L, 16384L, 1L, 8)
+          .selectExpr(
+            "id AS k",
+            "CASE WHEN id % 5 = 0 THEN NULL ELSE id % 17 END AS v",
+            "CASE WHEN id % 7 = 0 THEN NULL ELSE id % 11 END AS w")
+          .write
+          .parquet(path)
+      }
+      withParquetTable(path, "skip_partial_eligibility") {
+        withSQLConf(
+          SQLConf.ADAPTIVE_EXECUTION_ENABLED.key -> "false",
+          SQLConf.SHUFFLE_PARTITIONS.key -> "8",
+          CometConf.COMET_BATCH_SIZE.key -> "128",
+          CometConf.COMET_SHUFFLE_ENABLED.key -> "true",
+          CometConf.COMET_SHUFFLE_MODE.key -> "native",
+          CometConf.COMET_RESPECT_DATAFUSION_CONFIGS.key -> "true",
+          "spark.comet.datafusion.execution.skip_partial_aggregation_probe_rows_threshold" -> "100") {
+          def aggregates(query: String): Seq[CometHashAggregateExec] = {
+            val (_, plan) = checkSparkAnswerAndOperator(query)
+            val result = stripAQEPlan(plan).collect { case aggregate: CometHashAggregateExec =>
+              aggregate
+            }
+            assert(result.nonEmpty)
+            result
+          }
+          def skipped(aggregate: CometHashAggregateExec): Long =
+            aggregate.metrics.get("skipped_aggregation_rows").map(_.value).getOrElse(0L)
+
+          val countQuery = "SELECT sum(n) FROM " +
+            "(SELECT k, count(*) n FROM skip_partial_eligibility GROUP BY k)"
+          // No ratio override: the eligible plan uses DataFusion's adaptive default.
+          assert(aggregates(countQuery).map(skipped).sum > 0L)
+          assert(
+            aggregates("SELECT sum(n) FROM " +
+              "(SELECT k % 2, count(*) n FROM skip_partial_eligibility GROUP BY k % 2)")
+              .map(skipped)
+              .sum == 0L)
+
+          withSQLConf(
+            "spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold" -> "1.1") {
+            assert(aggregates(countQuery).map(skipped).sum == 0L)
+          }
+          withSQLConf(
+            "spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold" -> "0.8") {
+            for (expression <- Seq("sum(v)", "count(v, w)", "count(*) + sum(v)")) {
+              val result = aggregates(
+                s"SELECT sum(n) FROM " +
+                  s"(SELECT k, $expression n FROM skip_partial_eligibility GROUP BY k)")
+              assert(result.map(skipped).sum == 0L, s"Unexpected skipping for $expression")
+            }
+            val mixed =
+              aggregates("SELECT count(DISTINCT k), count(*) FROM skip_partial_eligibility")
+                .filter(_.modes.contains(PartialMerge))
+            assert(mixed.nonEmpty, "Expected a PartialMerge stage in the DISTINCT plan")
+            assert(mixed.map(skipped).sum == 0L)
+            withSQLConf(CometConf.COMET_SHUFFLE_MODE.key -> "jvm") {
+              assert(aggregates(countQuery).map(skipped).sum == 0L)
             }
           }
         }
--- a/native/core/src/execution/jni_api.rs
+++ b/native/core/src/execution/jni_api.rs
@@ -31,7 +31,7 @@ use std::collections::HashSet;
 use arrow::array::{Array, RecordBatch, UInt32Array};
 use arrow::compute::{take, TakeOptions};
 use arrow::datatypes::DataType as ArrowDataType;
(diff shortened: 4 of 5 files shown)


sunchao (approved):
## Correctness

Reviewed `8d915424c1cd91e2f53c23a8e183f4fd028c3adb` against `75fdddc9285ec61c0cd326977c61dd41fca39a8b`.

The previous session configuration disabled adaptive partial aggregation globally. This change enables it only for native shuffle-writer roots when every partial aggregate in the fused plan is grouping-only or single-argument `COUNT`. The recursive guard rejects unsupported accumulators, PartialMerge, mixed modes and unknown modes, and runs after configuration pass-through so an override cannot enable unsupported state conversion.

The DISTINCT mode change matches the maintained Spark 3.5 and 4.0 planning paths: local grouping-only stages have no required child distribution, while the final and pre-distinct merge stages require distribution and must fully deduplicate. Serializing those latter stages as native `Final` preserves that obligation, including when Spark can satisfy distribution without inserting another exchange. COUNT's locked DataFusion 55 implementation emits non-null Int64 0/1 states using both argument validity and FILTER validity. Downstream count-state merging matches Spark's additive buffer semantics. Accumulated groups drain before later input bypasses aggregation. Ordered and global aggregation paths do not gain this bypass. I found no additional correctness issue in those paths. Maintained Spark 3.4 and 4.1 sources were unavailable, so this source comparison does not qualify those versions.

The redundant string interpolation in the new regression test is fixed. The refreshed current-head checks at `2026-09-05T22:08:12.363Z` show 15 successes, seven skips and six checks still running, with no failures. The [Scala syntactic lint job](https://github.com/apache/datafusion-comet/actions/runs/33994220605/job/101381944058) passed. The completed syntactic lint log confirms checkout of synthetic merge `b27feaae6ecffe8c3f3e373139f1f67ff0f1da97`, which has parents `c348f775d5b20d91d1989756b72f62f366d67d32` and the reviewed head. Its four additional cast files make it a different tree, while all five files changed here, Cargo.lock and the lint configuration match the reviewed head. The native eligibility test passed on the preceding CI merge with identical native code and dependencies. That is historical supporting evidence, not a completed current-head native run. The full JVM suite and 22 Spark comparisons remain author-reported. I did not run local product tests and found no remaining actionable P1/P2 issue.

## Performance

The new plan walk occurs once during native session creation. It does not add a per-row Comet wrapper. Eligible unordered partials retain DataFusion's default probe at 100,000 input rows and a groups/rows ratio greater than 0.8. A threshold of 1.1 creates no probe, preserving explicit disablement and the unsupported path. On memory pressure the unordered partial stream emits states early. This should be distinguished from final-stage disk spills.

The [author's guarded workload](https://github.com/apache/datafusion-comet/issues/5717) reports seven alternating high-cardinality pairs improving median wall time from 664 to 479 ms, with 15,925,248 bypassed rows. The low-cardinality control bypassed zero rows, and the 128 MiB case supplied a smaller spilling sample. These are useful bounded workstation measurements, not independently reproduced or distributed/AQE qualification.

## Design

A single conservative decision for the fused plan is easy to follow and avoids assigning different session settings to individual operators. Inspecting children even below Final prevents an unsupported nested partial from being hidden by its parent. The grouping-only mode correction addresses the semantic reason global enablement was unsafe, while the post-override guard makes accumulator support an invariant. The current boundary is intentionally narrow. Broadening it should require working state conversion and runtime evidence for each additional accumulator and execution context.

## Abstraction & complexity

The helper, explicit mode selection, metric and focused admission tests fit the existing planner and metric interfaces. There is no new execution layer or configuration surface to maintain. The tests distinguish actual bypass from merely setting a threshold, and exercise unsupported modes and JVM shuffle. I found no unnecessary abstraction or actionable simplification.

sunchao (commented):
## Correctness

Follow-up on `a8f5969d3caf97338828212baf733cee2bddddaf` against `17f54da8ca5cb0ad5dbe8357b6e037fef8a0db2c`, after [review 5123205955](https://github.com/apache/datafusion-comet/pull/5723#pullrequestreview-5123205955). All five authored patches match the previously reviewed contribution. The 148-file increment matches main's advancement, and the eligibility guard, session configuration, aggregate planner, Scala aggregate serialization and aggregate metrics blocks are unchanged. I found no new or remaining actionable P1/P2 issue in this update and the relevant inherited interactions.

The maintained Spark 3.5/4.0 planning contracts still support the grouping-only `Final` mapping for stages requiring distribution. The locked DataFusion 55 COUNT state conversion still combines argument/FILTER validity into non-null Int64 states, and the partial stream drains accumulated states before forwarding later input. Unsupported accumulators and PartialMerge remain excluded after configuration overrides. The native writer still supplies the ShuffleWriter root. Scan plan-data injection preserves it. Inherited AQE input traversal and nested shuffle hashing do not alter these eligibility or grouping-mode decisions. Maintained Spark 3.4/4.1 branches remain unavailable, so the source comparison does not qualify those versions.

Both new aggregation regressions passed in the current [Spark 3.4](https://github.com/apache/datafusion-comet/actions/runs/34308307511/job/102332110269), [3.5](https://github.com/apache/datafusion-comet/actions/runs/34308307511/job/102332110267), [4.0](https://github.com/apache/datafusion-comet/actions/runs/34308307511/job/102332110380) and [4.1](https://github.com/apache/datafusion-comet/actions/runs/34308307511/job/102332110355) execution jobs. These check post-shuffle DISTINCT correctness and actual bypass/admission, including the low-cardinality, explicit-disable, unsupported aggregate, mixed-mode and JVM-shuffle controls. The [native fail-closed eligibility test](https://github.com/apache/datafusion-comet/actions/runs/34308307511/job/102330937214) also passed. That run reports 1,252 passed and five skipped tests. All inspected jobs ran merge `bd0cc3640207384b11ba11994f606cd820675efd`, whose parents are the assigned BASE/HEAD and whose tracked tree equals HEAD. Producer, artifact metadata and the four Spark consumers agree on native artifact `10087734700` and its SHA256 digest.

At `2026-09-09T05:54:26.032357+00:00`, checks were 64 successful, nine skipped and one failed, with none running. The [failed Spark SQL 4.1 job](https://github.com/apache/datafusion-comet/actions/runs/34308307511/job/102333893090) reports `SQLAppStatusListenerMemoryLeakSuite`'s `noLiveData()` assertion. Immediately preceding activity from that test logs that the Comet extension is disabled because of the shuffle-manager configuration. This does not demonstrate a failure in adaptive native aggregation, but the listener failure remains unresolved and CI is still red. No local product tests or builds were run. GitHub reports the existing approval attached to this head. This COMMENT preserves that approval.

## Performance

The merge leaves the one-time plan walk and locked adaptive probe/state-conversion paths unchanged. I found no added per-row Comet allocation or repeated work in this update. The [author's guarded benchmark](https://github.com/apache/datafusion-comet/issues/5717) remains the available measurement: seven alternating high-cardinality COUNT pairs, median 664 to 479 ms, with 15,925,248 bypassed rows. The low-cardinality control bypassed zero rows, and the constrained-memory run supplied three measured pairs. Those are historical workstation results. Main also advanced runtime, shuffle and other components, so feature-block equivalence does not establish a fresh speedup or an identical benchmark environment. No new timing, distributed/AQE or deployed-build qualification is claimed.

## Design

The update preserves the conservative whole-plan decision and the independent grouping-only mode correction. Final nodes still allow recursive inspection of nested partials, and configuration overrides cannot enlarge accumulator eligibility. The inherited writer and scan-data changes preserve the native plan boundary used by the guard. I found no changed design issue requiring an additional layer or a broader allowlist.

## Abstraction & complexity

The helper, metric and admission tests are unchanged and remain within existing planner interfaces. The merge introduces no additional abstraction for this feature. I found no actionable simplification in the reviewed update.

peterxcli:
> reports `SQLAppStatusListenerMemoryLeakSuite`'s `noLiveData()` assertio ... This COMMENT preserves that approval.

@sunchao could you please help me to retry the ci? thanks! the failed job must share the same reason as https://github.com/apache/datafusion-comet/issues/5023

docs/source/user-guide/latest/tuning.md
```diff
@@ -31,6 +31,20 @@ It is recommended that `COMET_WORKER_THREADS` be set to the number of executor c
 in some environments, such as Kubernetes, where the number of cores allocated to a pod will already be equal to the
 number of executor cores.
 
+## Adaptive Partial Aggregation
```
sunchao, line 34: nit: can we also document how to tune the threshold (via the DF config) or even disable this?

sunchao (commented):
## Correctness

Discussion follow-up on `a8f5969d3caf97338828212baf733cee2bddddaf` against `17f54da8ca5cb0ad5dbe8357b6e037fef8a0db2c`, after [review 5150301288](https://github.com/apache/datafusion-comet/pull/5723#pullrequestreview-5150301288). Head and base are unchanged. I re-read the full five-file contribution and verified its source against the prior packet. The incremental diff is empty. I found no new or remaining verified P1/P2. The existing [approval](https://github.com/apache/datafusion-comet/pull/5723#pullrequestreview-5123205955) stands.

I checked the controls behind the [documentation comment](https://github.com/apache/datafusion-comet/pull/5723#discussion_r3970239102). Overrides require `spark.comet.exec.respectDataFusionConfigs=true`, a development/testing option that defaults to false. The keys are `spark.comet.datafusion.execution.skip_partial_aggregation_probe_rows_threshold` and `spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold`. Locked DataFusion 55 defaults are 100,000 input rows per partial-aggregation partition and a groups/input-rows ratio greater than 0.8. Setting the ratio to 1.1 disables bypass. Both aggregate implementations omit the probe for ratios at least 1.0. Eligible plans preserve explicit overrides, while unsupported plans force 1.1 after pass-through. The tuning page currently documents eligibility and the metric but omits these controls. That omission is already covered by the existing thread.

The refreshed maintained Spark 3.5/4.0 references still support the unchanged grouping-only `Final` mapping for distribution-required stages and COUNT's zero-initialized, additive Long state. DataFusion's unchanged COUNT conversion includes input/FILTER validity and emits non-null 0/1 states. The update adds no type, null, ANSI/overflow, error or fallback change. Maintained Spark 3.4/4.1 source gaps remain.

### Validation and CI

At `2026-09-09T18:56:35Z`, checks remain 64 successful, nine skipped and one failed. The same completed jobs and run attempt retain the passing native eligibility test and both aggregation regressions on Spark 3.4/3.5/4.0/4.1. I reconciled their job identities and saved log provenance. Merge `bd0cc3640207384b11ba11994f606cd820675efd` has the assigned base/head parents and the same tracked tree as HEAD. This reuses the previous CI execution evidence, with no new local build, test or benchmark.

The [failed listener job](https://github.com/apache/datafusion-comet/actions/runs/34308307511/job/102333893090) still reports `SQLAppStatusListenerMemoryLeakSuite`'s `noLiveData()` assertion, with Comet disabled in the preceding test activity. The author's [retry request](https://github.com/apache/datafusion-comet/pull/5723#issuecomment-5601814040) links [#5023](https://github.com/apache/datafusion-comet/issues/5023), which reports a macOS scan SIGSEGV and has a comment attributing it to libhdfs JNI lifetime handling. That report does not establish a common cause for this Linux listener assertion. CI remains red and the listener failure remains unresolved.

## Performance

There is no execution change since the previous review. The eligibility walk remains once per native context. Lower probe thresholds can trade hash-table work for more emitted partial states, so the existing author measurements do not establish a universal tuning value. No fresh timing, distributed/AQE or deployed-build benefit is claimed.

## Design

The documented controls should reflect the existing post-override safety boundary: tuning can change when an eligible plan skips, but cannot admit unsupported accumulators or modes. I verified that boundary against the configuration serialization and native session setup. No changed design issue requires another control or execution path.

## Abstraction & complexity

The helper, metric, mode mapping and admission tests are unchanged. The existing DataFusion pass-through already supplies the tuning mechanism. I found no new abstraction issue or actionable simplification in this follow-up.

sunchao (approved):
Reviewed `7ec7cfcd` against `17f54da8`. The documentation now addresses the [tuning and disablement request](https://github.com/apache/datafusion-comet/pull/5723#discussion_r3970239102): it explains the configuration gate, the 100,000-row and 0.8 defaults, and disabling bypass with 1.1. I checked the examples against the configuration path and confirmed that eligibility checks still run after overrides.

This update changes only the documentation; execution code, regression tests, and dependencies are unchanged from `a8f5969d`. No new or remaining verified P1/P2 findings.

Current check metadata reports 56 successes and 10 skips. Current job logs were not reverified; the fetched merge preview includes newer main changes and DataFusion 55.1, while the reviewed head locks 55.0. Earlier verified native and Spark 3.4–4.1 regression runs support the unchanged implementation. No local tests or fresh performance benchmark were run.

spark/src/test/scala/org/apache/comet/exec/CometAggregateSuite.scala
```diff
@@ -1566,6 +1566,108 @@ class CometAggregateSuite extends CometTestBase with AdaptiveSparkPlanHelper {
     }
   }
 
+  test("skip partial aggregation preserves post-shuffle distinct") {
```
andygrove, line 1569: This test compares results but does not confirm that the stage 1 partial actually bypassed aggregation. If the probe never fires, the test passes without exercising the DISTINCT dedup path it is meant to protect. It might be worth collecting the `CometHashAggregateExec` nodes and asserting `skipped_aggregation_rows` is greater than zero on the Partial stage, the same way the eligibility test does.

docs/source/user-guide/latest/tuning.md
```diff
@@ -31,6 +31,46 @@ It is recommended that `COMET_WORKER_THREADS` be set to the number of executor c
...
+SET spark.comet.exec.respectDataFusionConfigs=true;
+SET spark.comet.datafusion.execution.skip_partial_aggregation_probe_rows_threshold=100000;
+SET spark.comet.datafusion.execution.skip_partial_aggregation_probe_ratio_threshold=0.8;
+```
+
+A lower row threshold allows an earlier decision; a lower ratio threshold makes
+skipping more likely. Skipping can increase the number of partial states emitted
+and the amount of shuffle data, so measure the effect on your workload.
+
+To disable skipping, keep `spark.comet.exec.respectDataFusionConfigs=true` and set
```
andygrove, line 64: The only documented way to turn this off is `spark.comet.exec.respectDataFusionConfigs`, which is a TESTING category option, plus a raw DataFusion key. Since this feature was disabled once before for a TPC-DS wrong-result (#788, apache/datafusion#11850) and is now on by default, would you consider a first-class boolean such as `spark.comet.exec.aggregate.skipPartial.enabled` (default true) that the guard in `configure_skip_partial_aggregation` checks alongside the plan walk? That gives operators a supported kill switch without pointing them at a testing flag.

spark/src/main/scala/org/apache/spark/sql/comet/operators.scala
```diff
@@ -1734,6 +1734,15 @@ trait CometBaseAggregate {
     if (aggregateExpressions.isEmpty) {
       val hashAggBuilder = OperatorOuterClass.HashAggregate.newBuilder()
       hashAggBuilder.addAllGroupingExprs(groupingExprs.map(_.get).asJava)
+      // Spark has no expression mode to serialize here. An empty aggregate with a required child
+      // distribution must fully deduplicate its keys (Final, or a pre-distinct PartialMerge), so
```
andygrove, line 1738: This comment says a distribution-required stage must fully deduplicate its keys, including the pre-distinct `PartialMerge` case. For the non-empty `PartialMerge` stage, Comet still plans a DataFusion `Partial` with `MergeAsPartial` accumulators, and DataFusion's Partial mode uses `EmitEarly` under memory pressure, which can re-emit a key that was already emitted. Does that mean stage 3's distinct `count(k)` can overcount when stage 2 spills? I think this is pre-existing and separate from this PR, and the guard correctly refuses to skip in that shape, but could you open a tracking issue and link it from this comment so the obligation is recorded?

andygrove (approved):
Thanks for the careful approach here. Mapping the distribution-required grouping-only stage to native `Final` addresses the actual cause of the old q97 corruption, and the guard in `configure_skip_partial_aggregation` is fail-closed. I traced the plan shapes it admits (grouping-only and single-argument `Count` partials under a native `ShuffleWriter`) and could not construct one that produces wrong results, including when the Spark `Final` aggregate falls back while the shuffle stays native. CI is green across all Spark versions.

A few comments inline. The test hardening and the config question are worth addressing before merge. The third is about a pre-existing gap that this PR's own comment documents, so I would just like a tracking issue for it.

